PR #675, opened 2014-05-09
Use handlebars-runtime in production.

xtian:


Kuzirashi:
I still get an error if I try to remove following workaround:
~~~~
var index = app.legacyFilesToAppend.indexOf('bower_components/handlebars/handlebars.runtime.js');
if(index) {
  app.legacyFilesToAppend[index] = 'bower_components/handlebars/handlebars.js';
}
~~~~
Any update on this? CLI 0.1.9.

rwjblue:
@Kuzirashi - That is not how to update a vendored asset.  See http://www.ember-cli.com/#customizing-a-built-in-asset for details.

Kuzirashi:
@rwjblue Thanks, I've changed this code. To clarify - has this something to do with my plugins which require full version of Handlebars?

PR #803, opened 2014-05-23
More detailed docs on using non-AMD assets with ember-cli

josh-padnick:
Not having worked with javascript dependencies before, I found the current ember-cli docs confusing since they spoke of the admittedly amazing ES6 imports, except that they aren't available yet!

Hopefully this will clarify for other noobs how to handle dependencies immediately.


abuiles:
Tag #3307 we might eventually add this once the new docs come in.

PR #899, opened 2014-06-05
Display message to user if an ember-cli update is available (and add `ember update` command)

andycrum:
This is a first pass at checking for ember-cli updates and displaying a message to the user if an update is available (fixes https://github.com/stefanpenner/ember-cli/issues/335).
### TODOs
- [x] Persistence via .ember-cli
- [x] Option to disable via .ember-cli
- [x] `ember update` command
- [x] Tests


devinrhode2:
Why isn't `ember update` mentioned in the site documentation http://www.ember-cli.com/#upgrading?

PR #1024, opened 2014-06-13
Allow passing customHash in fingerprint options

chnn:
This adds a `customHash` option that can be passed to the fingerprint configuration:

```
var app = new EmberApp({
  fingerprint: {
    enabled: true,
    customHash: 'custom123'
  }
});
```

The result is that fingerprinted files will use the `customHash` string instead of a MD5 hash of the file. (e.g. "assets/vendor-5429f75166356b28c3e42dd401abaf36.js" becomes "assets/vendor-custom123.js").

The motivation behind this is that it's difficult to get the names of fingerprinted files if you aren't using the ember-cli generated dist/index.html file. By using `customHash`, you can set the fingerprint to be something like the current commit's SHA if you are working in a git repo or a version number from a package.json file. Those are easier to predict and setup in a non ember-cli generated index.html template elsewhere.

It relies on some new functionality released in broccoli-asset-rev 0.0.6.


raycohen:
let's get this handy feature documented in http://www.ember-cli.com/user-guide/#fingerprinting-and-cdn-urls

PR #1242, opened 2014-07-03
[DOC] Using canary build instead of release

pixelhandler:
Adds instructions on editing bower.json to use ember canary build.

Based on https://github.com/rjackson/_____ember-cli-test/commit/e4a9413b0b6c6d87f325a58b6115326f852741cc


shaunc:
I notice that ember-data is in npm as well as bower. Does that also need update; if so, how to point to "canary"? (do we need to point to "master" at github?)

PR #1369, opened 2014-07-13
Add git init to `ember new`

rondale-sc:


joostdevries:
@stefanpenner Any reason we're always showing `Could not find watchman, falling back to NodeWatcher for file system events` but we're not showing `Could not find git, no repository initialized` here?

PR #2122, opened 2014-09-26
Add `before` and `after` options to Blueprint#insertIntoFile

bantic:
Adds the ability to specify the string in a file to insert contents before or after, similar to the [same method in Thor](https://github.com/erikhuda/thor/blob/master/lib/thor/actions/inject_into_file.rb#L17).

 Example:

```
 // .jshintrc
 "predef": [
   "window",
   "document"
 ]

 insertIntoFile('.jshintrc', '  "newGlobal",', {after:'"predef": [\n'});

 // new .jshintrc
  "predef": [
    "newGlobal",
    "window",
    "document"
 ]
```

### DIFF
--- a/tests/unit/models/blueprint-test.js
+++ b/tests/unit/models/blueprint-test.js
@@ -612,6 +612,126 @@ describe('Blueprint', function() {
           assert(result.inserted, 'inserted should indicate that the file was not modified');
         });
     });
+
+    it('will insert into the file after a specified string if options.after is specified', function(){
+      var toInsert = 'blahzorz blammo';
+      var line1 = 'line1 is here';
+      var line2 = 'line2 here';
+      var line3 = 'line3';
+      var originalContent = [line1, line2, line3].join(EOL);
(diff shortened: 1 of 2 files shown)


tests/unit/models/blueprint-test.js
```diff
@@ -612,6 +612,126 @@ describe('Blueprint', function() {
...
+
+          assert.equal(contents, [line1, line2, toInsert, line2, line3].join(EOL),
+                       'inserted contents should be inserted after the `after` value');
+          assert.equal(result.originalContents, originalContent, 'returned object should contain original contents');
+          assert(result.inserted, 'inserted should indicate that the file was modified');
+        });
+    });
+
+    it('will insert into the file before a specified string if options.before is specified', function(){
+      var toInsert = 'blahzorz blammo';
```
jameelahuq, line 659: Inserting an EOL in the "before" statment causes formatting issues. Can we just insert "before" starting immediately before the statement? And leave it to the insertion statment to add the EOL?

PR #2164, opened 2014-10-02
Enable easier customization of production environments.

rwjblue:
The following is all that is needed to enable building `staging` as another `production` type environment:

``` javascript
var app = new EmberApp({
  productionEnvironments: [ 'production', 'staging' ]
});
```

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -55,12 +55,14 @@ function EmberApp(options) {
     this.project.configPath = function() { return options.configPath; };
   }
 
-  this.env  = process.env.EMBER_ENV || 'development';
+  this.env  = process.env.EMBER_ENV || options.environment || 'development';
   this.name = options.name || this.project.name();
 
   this.registry = options.registry || p.setupRegistry(this);
 
-  var isProduction = this.env === 'production';
+  this.productionEnvironments = options.productionEnvironments || ['production'];
+
+  var isProduction = this.productionEnvironments.indexOf(this.env) > -1;
 
   this.tests   = options.hasOwnProperty('tests')   ? options.tests   : !isProduction;
   this.hinting = options.hasOwnProperty('hinting') ? options.hinting : !isProduction;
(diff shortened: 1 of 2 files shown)


stefanpenner:
@devinus nope, we have not worked on the configuration story yet

skylar:
Ha, I just wrote something very similar to what @rwjblue wrote for my app. It might be worth updating the core docs under "environment" to make it clear that different environments aren't supported and that you should stick to the core dev/test/prod.

I very much like the idea of a set of config flags that could be imported/reused by other configurations. For example, we'd take a "production" config and modify it slightly for staging. Ping me if someone else is working on this currently and could use some help.

KaiFuglem:
@skylar would you be able to share how you achieved this? I haven't been able to get the code above working for me.  Thanks in advance!

skylar:
@KaiFuglem ha - yes, I forgot I had done this. In any case, I put my `ember-cli-build.js` in a just for you.  let me know if that helps:

https://gist.github.com/skylar/e1ca3ea34dbca53d63ff

devinus:
Why couldn't we get something like this until this huge refactor lands?

skylar:
@devinus do you have a PR?

devinus:
@skylar We're on the PR.

KaiFuglem:
Thanks @skylar! That worked for me, but for those still having issues implementing this, I also had to modify an app.import:

```javascript
  app.import({
    development: 'xxx-debug.js',
    production: 'xxx.js',
    staging: 'xxx.js'
  });
```

blimmer:
I really would love to see some kind of support like @rwjblue mentioned way back at the start of this issue. 

```
var app = new EmberApp({
  productionEnvironments: [ 'production', 'staging' ]
});
```

I've run into **3** subtle problems today that caused deployment headaches because of differences between my staging and production environments.

In my workflow, staging should be **just like** production so I can catch bugs before I deploy to prod. Since I use ember-cli-deploy-redis the new versions aren't activated until I explicitly say so, however it's really unsettling to test everything/accept stories on staging and then find (multiple) bugs in production caused by build differences.

I currently have 11 instances of flipping the switch based on `isProductionLikeBuild` in my `ember-cli-build.js` file, which is about 10 too many, IMO.

stefanpenner:
I would recommend, people stick to the 3 main envs, and use ENV vars (Accessible inside the builld and the config/environment.js to configure further.

This allows addons to understand the intent of a given ENV. I plan to also the concept of --target (ios/embedded/electron etc.etc.) the 3x3 matrix + ENV var channel should handle most cases.

2468ben:
@blimmer I have/had the same pain point with ember-cli & ember-cli-deploy and was using the switches until I ran into a couple plugins that internally based things on ```env === 'production'```. I recently moved over to some ENV vars similar to @stefanpenner's suggestion; here's some relevant bits of my files for how it works. But this way ember-cli only sees 'production', and ember-cli-deploy + app code gets production and staging env vars to work with.

It feels confusing to type out, but I'd be glad to talk on slack/wherever if you want more code or a better-thought-out explanation.

```
//.env.staging
EMBER_ENV=production
DEPLOY_ENV=staging
PRODUCTION_LIKE_BUILD=true


//config/deploy.js
require('dotenv').load();  //loads .env
require('dotenv').config({path: '.env.'+deployTarget});  //loads .env.staging
var ENV = {
    build: {
      environment: deployTarget  //returns 'staging' 
    }
}


// config/environment.js
var ENV = {
  ...
  environment: environment, //returns 'production'
  EmberENV: {
    PRODUCTION_LIKE_BUILD: process.env['PRODUCTION_LIKE_BUILD']  //returns true
    environment: process.env['DEPLOY_ENV']  //returns 'staging'
  }
}
```

blimmer:
@stefanpenner I ran into this again today with a bug report we should have caught in staging. are there any updates wrt this problem? should I submit an RFC to get more feedback? I would really like to help figure out a solution for this issue.

PR #2298, opened 2014-10-12
Add `storeConfigInMeta` option.

rwjblue:
Allows opting out of using `<meta>` tag in `head` section for configuration.
- If set to `true` the calculated config will be stored in a `<meta>` tag in `index.html`.
- If set to `false` the calculated config will be stored in the final `my-app-name.js` file.

Defaults to `true`.

Closes https://github.com/stefanpenner/ember-cli/issues/2184.

### DIFF
--- a/lib/broccoli/app-config-from-meta.js
+++ b/lib/broccoli/app-config-from-meta.js
@@ -1,0 +1,14 @@
+/* jshint ignore:start */
+
+try {
+  var metaName = prefix + '/config/environment';
+  var rawConfig = Ember['default'].$('meta[name="' + metaName + '"]').attr('content');
+  var config = JSON.parse(unescape(rawConfig));
+
+  return { 'default': config };
+}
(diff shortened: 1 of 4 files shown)


oskarrough:
How about setting the default to false? Perhaps you can enlighten me but I don't see why you'd want it in index.html instead of the final .js file.

stefanpenner:
Equality between builds where only the config changes. Allows staging and production builds to be easily compared.

jmurphyau:
Also I found that in development when running ember server, navigating to `/tests` and having acceptance tests that navigate to other URLs, if `storeConfigInMeta` is `false` the test actually initiates a transition to that URL and you end up exiting from the test page

stefanpenner:
Yes. We plan to address that with the packager work

dejongnj:
Awesome! I know I'm late to the party, but this PR keeps giving :)

Issue #2354, opened 2014-10-20
Preprocessing addons doesn't work


davewasmer:
I've created a [demonstration repo](https://github.com/davewasmer/ember-addon-coffeescript-test) to illustrate the issue. It contains an empty Ember app, freshly generated from the CLI (v0.1.2), with an addon (also generated with v0.1.2) that contains three files:

* `/app/controllers/bar.coffee`, which imports the other two files;
* `addon/bat.js` which exports a simple POJO as a basic test
* `addon/foo.coffee` which exports a simple POJO as a basic test

After a whole bunch of logging and tracing, it looks like the Registry for the Addon doesn't pick up the Coffeescript compiler - the glob that is built to look for files in the `/addon/` directory only checks for `.js` and `.em`.

I'm continuing to dig, but there's a lot of code to familiarize myself with, so I figured I'd post here in case it was an obvious fix.

**Steps to reproduce**

Run the demonstration repo (copy/pastable):

```sh
git clone git@github.com:davewasmer/ember-addon-coffeescript-test.git && cd ember-addon-coffeescript-test
npm install
bower install
ember s
```

**Expected Result**

The built app (`/dist`) should contain a controller called `bar` which imports from `addon-using-cs/bat` and `addon-using-cs/foo`.

**Actual Result**

Build fails:

```sh
ENOENT, no such file or directory '.../ember-addon-coffeescript-test/tmp/tree_merger-tmp_dest_dir-uy3fD5ba.tmp/addon-using-cs/foo.js'
```

And looking at the output in that tmp directory, it appears that the JavaScript file (`/addon/bat.js`) was picked up correctly, but not the Coffeescript one (`/addon/foo.coffee`):

```
$ cat vendor/addons.js
eval("define(\"addon-using-cs/bat\", \n  [\"exports\"],\n  function(__exports__) {\n    \"use strict\";\n    bat = {\n      fizz: \'bat\'\n    };\n\n    __exports__[\"default\"] = bat;\n  });//# sourceURL=addon-using-cs/bat.js");

;eval("define(\"addon-using-cs\", [\"addon-using-cs/index\",\"exports\"], function(__index__, __exports__) {\n  \"use strict\";\n  Object.keys(__index__).forEach(function(key){\n    __exports__[key] = __index__[key];\n  });\n});\n//# sourceURL=__reexport.js");
```

aceofspades:
+1

chadhietala:
[...]

PR #2381, opened 2014-10-26
Enable properly nested addons

davewasmer:
See #2354 for context.

This adds a `parentAddon` context for any nested addons included in a project. The `parentAddon` is supplied to each child addon on the addon instance as well as passed into the `included` hook.

The child addon is then free to determine if it should modify the `parentAddon` or the `app` from within the `included` hook.

This enables addons to have their own isolated dependencies which don't interfere with the root app's preprocessors. I.e. I can write my addon in Coffeescript, and the parent app doesn't need to depend on the Coffeescript addon directly.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -236,18 +236,18 @@ EmberApp.prototype.concatFiles = function(tree, options) {
   @private
 */
 EmberApp.prototype._notifyAddonIncluded = function() {
+  var self = this;
   this.initializeAddons();
-  this.project.addons = this.project.addons.filter(function(addon) {
-    addon.app = this;
 
-    if (!addon.isEnabled || addon.isEnabled()) {
-      if (addon.included) {
-        addon.included(this);
-      }
+  this.project.addonGraph.topsort(function(vertex) {
+    var addon        = vertex.value.addon;
+    var parentAddon  = vertex.value.parentAddon;
+    var target       = parentAddon || self;
 
-      return addon;
+    if (addon.included) {
+      addon.included(target);
     }
-  }, this);
+  });
 };
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Applications
 ### Addons
+* [ENHANCEMENT] Addons can now use nested addons. This enables things like addons using their own preprocessors. [#2354](https://github.com/stefanpenner/ember-cli/issues/2354)
+
 ### Blueprints
 
 ### 0.1.7
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -20,7 +20,7 @@ function Project(root, pkg) {
   debug('init root: %s', root);
   this.root          = root;
   this.pkg           = pkg;
-  this.addonPackages = {};
+  this.addonGraph    = new DAG();
   this.liveReloadFilterPatterns = [];
   this.setupBowerDirectory();
 }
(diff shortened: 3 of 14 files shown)


lib/broccoli/ember-app.js
```diff
@@ -190,12 +190,22 @@ function EmberApp(options) {
...
-  this.project.addons.forEach(function(addon) {
+
+  this.project.addonGraph.topsort(function(vertex) {
+    var addon        = vertex.value.addon;
+    var parentAddon  = vertex.value.parentAddon;
+    var target       = parentAddon || self;
+
     if (addon.included) {
-      addon.included(this);
+      if (addon.name === 'broccoli-asset-rev') {
```
stefanpenner: ping?

stefanpenner:
@davewasmer got time to rebase this bad boy. I believe this will be a good base for our efforts here.

Sorry it has taken me so much time to get back to you.

stefanpenner:
This addon currently begins the process of combining `app` and `addon` this union makes sense, unfortunately an `app` has various methods like `options` or `import` that and add-on does not have. 

what does `addon.import` mean? For today, we should likely justdelegate that to app.import.

This also exposes complexities in our current add-on development harness, were its `dummy` app is really just "merged" into the add-on, and is barely a second entity.

One approach, is to split todays concept of `app`, one part is the app/* tree we are used to today, and the other part is the "runner" for example the index.html, webworker or node.

I am hopeful we can sort these problems out, but we have much work todo.

stefanpenner:
i suspect this really needs a fleshed out RFC to go along with it. It is tricky to reason about the many aspects of this problem.

stefanpenner:
@davewasmer / @chadhietala do either of you have time this weekend for a hangout? I feel a high bandwidth conversation may help us figure out the path forward.

chadhietala:
@stefanpenner Yea I can probably make time.

davewasmer:
@stefanpenner @chadhietala same, I can probably find some time.

I rebased, but it was a bit of a blind rebase - looks like a few changes happened to some of the areas of code I touched, and I didn't spend the time to fully grasp them. I didn't test anything out afterwards, since this is probably just serving as a strawman to get things started (at best).

@stefanpenner honestly, I think an RFC would be a good way to go, since I think doing addons the right way ends up unifying "addons" and "apps" into the same concept, where the "app" is just the top level.

Certainly open to other approaches, that's just my reaction after digging through this (and writing numerous addons for myself).

stefanpenner:
@chadhietala @davewasmer does 1pm EST saturday work for you guys?

davewasmer:
Sure, sounds good to me.

stefanpenner:
Ah, this weekend wont work, maybe early next week? or the following weekend?

krisselden:
@davewasmer I agree, does seem to be likely need one model that works for app and addon where the app just is the top level, I feel like the index stuff should like be part of step ending package step.

stefanpenner:
> I feel like the index stuff should like be part of step ending package step.

+1

davewasmer:
@stefanpenner next week could work - during the day, or evenings?

Would it make sense for me to start drafting a rough RFC? Or would you rather wait until we have time to chat?

chadhietala:
@stefanpenner @davewasmer I can do anytime during the week or Saturday. No go for next Sunday.

davewasmer:
@stefanpenner any idea what your availability is to work on this? I've got some time this Saturday, if that works for you.

davewasmer:
ping @stefanpenner @rwjblue any updates on this? I know you guys are super busy, just need to decide whether to hack around this or hold out a bit longer for some kind of fix / movement.

rwjblue:
This will be replaced with https://github.com/ember-cli/ember-cli/pull/3166 I think

rwjblue:
Please review https://github.com/ember-cli/ember-cli/pull/3166, I am still working on more tests + docs...

davewasmer:
Ah. Didn't know there was already a feature complete solution to this. I guess I'll follow that thread to see if there is any way to help.

rwjblue:
@davewasmer - Definitely could use your thoughts on that approach....

stefanpenner:
I'm moving ny -> ca this weekend. I will not be around 

@lukemelia and @rwjblue have been making good progress. You should check Luke's most recent pr

PR #2465, opened 2014-11-08
[SPIKE] Detect and propagate file changes on rebuild.

rwjblue:
This should make it possible to live-reload only changed output files.

Needs review + tests.

/cc @stefanpenner

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -40,6 +40,7 @@ module.exports = Task.extend({
 
     this.mkdirp   = require('mkdirp');
     this.walkSync = require('walk-sync');
+    this._previousBuildStats = null;
   },
 
   canDeleteOutputPath: function(outputPath) {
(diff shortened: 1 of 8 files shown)


aexmachina:
I'm having another go at this and I'm seeing some strange results. When I run `mocha tests/**/*-slow.js` I get errors saying `Testem Server Error: listen EADDRINUSE`, but I've made sure to kill all node processes before running the tests.

aexmachina:
The other failure is `AssertionError: tmp/ should be empty after `ember` tasks. Contained: caching-writer-dest-dir_1ML0Mr.tmp/` followed by a huge list of files. Strange results... any ideas?

stefanpenner:
this has gone stale. We will re-visit when the – now actively being developed – pre-packager/packager work lands. 
cc @chadhietala

PR #2624, opened 2014-12-02
Add test helper blueprint

ryanto:
```
ember g test-helper routes-to
```

Will create a ``routesTo`` test helper.

---

This is my first cli PR. Let me know if I should be adding something
else in here. 

Do you guys unit test blueprints? I can add one, but it didn't look
like it was common/required. Also, by default this creates an async test
helper, which I think is right. However I can change/flag if required.


stefanpenner:
@ryanto got time for that rebase good sir?

stefanpenner:
rebased https://github.com/ember-cli/ember-cli/pull/3049

ryanto:
Ah thanks Stefan! I missed your original message last month.

Issue #2640, opened 2014-12-03
Add `.npmignore` to Addon Blueprint


rwjblue:
With something like:

```
tests/
bower_components/

.travis.yml
.npmignore
Brocfile.js
.editorconfig
testem.json
.ember-cli
bower.json
.bowerrc
**/.gitkeep
```

Thanks to @pogopaule for bringing it up in https://github.com/dockyard/ember-cli-i18n/pull/28.

chadhietala:
Closed by #2669.

PR #2669, opened 2014-12-04
add .npmignore to addon blueprint

pogopaule:
Having an `.npmignore` in your addons directory can reduce the size of the packaged and published node module considerably. For example: In case of the [ember-cli-i18n addon](https://github.com/dockyard/ember-cli-i18n), the size of the tgz file build with the `.npmignore` in this PR is only 1.5% of the one build without the `.npmignore`.

I'm not sure if calling `tar` (line 204 in the test) without any checks is such a good idea...The test probably fails on Windows. What do you think about this?

This resolves issue #2640

### DIFF
--- a/blueprints/addon/files/.npmignore
+++ b/blueprints/addon/files/.npmignore
@@ -1,0 +1,12 @@
+bower_components/
+tests/
+
+.bowerrc
+.editorconfig
+.ember-cli
+.travis.yml
+.npmignore
+**/.gitkeep
+bower.json
+Brocfile.js
+testem.json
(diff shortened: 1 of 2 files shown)


bcardarella:
@stefanpenner @rwjblue is the test for this necessary? Seems a little much. I think we can kill the test and just bring in the `npmignore`

rwjblue:
@bcardarella - Test is nice, and seems to pass just fine now.  Looks like this needs a rebase though...

pogopaule:
@rwjblue rebase done

bcardarella:
@pogopaule thank you!. Closing related issue in ember-cli-i18n

PR #2722, opened 2014-12-10
Give anonymous modules named variables in files

indirect:
Now that pods are officially the “best” way to organize Ember projects,
it can get really confusing to have 25 files named “model.js” open. This
pull request adds variable names to the component, controller, mixin,
route, service, and view blueprints so that simply having the file open
reveals the name of the module that you are editing.


deepan83:
Could we generate a comment in the files specifying the path and name as a measure?

PR #2730, opened 2014-12-11
Add datetime to build successful message

knownasilya:
Making it clear when something ran

screen shot 2014-12-11 at 11 30 02 am


stefanpenner:
appears stale, will revisit if a fresh PR with passing tests is presented.

PR #2732, opened 2014-12-12
Configure broccoli to use os.tmpdir

raytiley:
Not ready yet, obviously needs tests and such, but want to get feedback on approach.  This PR depends on: 
https://github.com/broccolijs/broccoli/pull/224

An example of how a plugin can be modified to make use of the provided tmpDir is here:

https://github.com/rwjblue/broccoli-funnel/pull/7

The nice thing about this is its not all or nothing. Nothing changes unless a plugin opts in.


stefanpenner:
Will accept a PR once the broccoli side settles.

PR #2808, opened 2014-12-21
Task for uninstalling Bower packages

chrmod:
Managing bower components versions from ember-addons is currently not possible as ember-cli offers no way to alter once installed version. This PR introduces simple task for uninstalling bower dependencies (and removing them from bower.json) so they can be easly installed in another version (using bower-install task).

Using this task will allow authros of ember-addons to help their users to upgrade bower dependencies.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -835,6 +835,19 @@ Blueprint.prototype.addBowerPackagesToProject = function(packages, installOption
   });
 };
 
+Blueprint.prototype.removeBowerPackageFromProject = function(packageName) {
+  return this.removeBowerPackagesFromProject([packageName]);
+};
+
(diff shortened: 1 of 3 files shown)


stefanpenner:
We can revisit this someday if my above concerns are addressed, or someone is able to convince me that is it reasonable. But today i don't believe either to be true.

Thanks for the PR, but i don't feel it is appropriate at this time.

broerse:
I try to switch to npm versions for ember-pouch here: https://github.com/nolanlawson/ember-pouch/pull/162/commits/17ff52e660fb64c02921a7c5a73cf52434c6f71e

Is there some code example, I can look at, that warns for unused bower modules? Is this perhaps added in the move from Bower to npm in ember-cli?

broerse:
Solved it with warnings: https://github.com/nolanlawson/ember-pouch/blob/master/index.js#L10

PR #2866, opened 2015-01-01
Add .npmignore for ember-cli addons

bcardarella:
When packaging the node module it likely does not need all of these
files. We can disucss if anything here should be included.

This was originally submitted to ember-cli-i18n by @pogopaule and we
requested that he submit PR to ember-cli but to my knowledge that has
not happened so I'm taking the initiative.


bcardarella:
Closing in favor of #2669

PR #2867, opened 2015-01-01
Add location for storeConfigInMeta flag

mitchlloyd:
It took me a minute to figure out where this flag was supposed to go even though Brocfile.js was mentioned at the top of this section.


rwjblue:
Thanks!!

PR #2885, opened 2015-01-04
[FEATURE] NPM should use save-exact flags

chadhietala:
NPM chooses not to do this by default. I think is a mistake because of issues with fuzzy matching dependencies.  We have an opportunity to lock dependencies down to real versions in Ember CLI by default.

### DIFF
--- a/lib/commands/install-addon.js
+++ b/lib/commands/install-addon.js
@@ -31,7 +31,8 @@ module.exports = Command.extend({
 
     return npmInstall.run({
       packages: rawArgs,
-      'save-dev': true
+      'save-dev': true,
+      'save-exact': true
     }).then(function() {
       return this.project.reloadAddons();
     }.bind(this)).then(function() {
--- a/lib/commands/install-npm.js
+++ b/lib/commands/install-npm.js
@@ -21,7 +21,8 @@ module.exports = Command.extend({
 
     return npmInstall.run({
       packages: rawArgs,
-      'save-dev': true
+      'save-dev': true,
+      'save-exact': true
     });
   }
 });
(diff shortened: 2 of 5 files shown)


rwjblue:
Can you describe what this flag is for, and why we should use it?

chadhietala:
> --save-exact: Saved dependencies will be configured with an exact version rather than using npm's default semver range operator.

So instead of npm flavored semver where they add range operators, you will get a definitive version. For example if you do `npm i glob --save --save-exact` it will download `4.3.2` to node_modules and place `glob: "4.3.2"` into the package.json. It is now guaranteed that your application will get version `4.3.2`.

When you just run `npm i glob --save` you open your self up to minor releases because what goes into your package.json is `glob: ^4.3.2`. This has the potential of having unexpected breaking changes in your CI box or staging environment.

Basically having dependencies that mutate is probably a bad idea and is harder to reason about then just a install version that is immutable.

stefanpenner:
since npm lacks a usable "lock" file (shrinkwrap is deficient) I am for this.

@rwjblue thoughts?

rwjblue:
My concern is that bugfixe releases from a semver bound package (> 1.0.0) would then require app developers to manually research and update. 

Using exact versions for < 1.0.0 packages absolutely makes sense, but for > 1.0.0 I personally would prefer ^1.0.0. 

I'm happy to give this change a chance though, if yall feel strongly about it.

chadhietala:
I think there is the possibility of doing a periodic top level `npm outdated` check during either `ember serve` or when the developer does `ember init` or `ember docter` (TBD).

The Ember ecosystem does an awesome job at semver and have a very structured and no surprises way of bumping versions. That guarantee is not a constant outside of the Ember community though. I think Ember CLI can be the abstraction to mitigate some of the NPM craziness by introduce more sane defaults.

__strong tag for strong opinion__ :smiley:

rwjblue:
Then lets give it a go.

PR #2887, opened 2015-01-04
I don’t think we need this anymore.

stefanpenner:
maybe we do..https://github.com/ember-cli/ember-cli/commit/3a439e1ea9c1f07c91a03d6cbb6703ada9c9afc1#diff-f1e1923c0440a385f51d2a81c0e53b64R63

But I would like to get a path forward, were we don't need this.

If this is merely going to be a breaking change for add-ons then we will doit, and bump the version accordingly.

Also if this is something we want to support, we need proper test coverage.


rwjblue:
I think its fine to remove, as long as we have a `[BREAKING]` tag in the changelog (under the addon section).  I do not think this warrants a 0.2.x version bump.

PR #2901, opened 2015-01-06
improve boot by 300ms -> 400ms

stefanpenner:


jakecraige:
Would it be possible to add a test for boot time perf? That would be a great way to watch for regressions

stefanpenner:
> Would it be possible to add a test for boot time perf? That would be a great way to watch for regressions

CI server is pretty crappy re: perf. But maybe a test we can run only locally?

jakecraige:
Yeah I guess the system perf affects the test :/ But yeah, some way to easily check it / keep an eye on it in a consistent format would be nice

stefanpenner:
apparently i can't program, will investigate these failures after work... "(unless someone has time to look earlier)"

stefanpenner:
> Yeah I guess the system perf affects the test :/ But yeah, some way to easily check it / keep an eye on it in a consistent format would be nice

i gave this some thought, how about a test that only run when `CI=false` ?

jakecraige:
Sounds fine to me :+1:

PR #2909, opened 2015-01-06
Use lib/ext/promise instead of RSVP directly

zeppelin:
Similarly how it's done in other files under `lib/`


jakecraige:
:+1: when green

stefanpenner:
:+1:

PR #2910, opened 2015-01-06
Update org references to ember-cli

zeppelin:
"Destefanpennerize" would make a good word :)


stefanpenner:
> "Destefanpennerize" would make a good word :)

:+1:

PR #2936, opened 2015-01-09
Fixes style preprocessor behavior in addons

pzuraq:
Currently when a style preprocessor is used in an Ember-CLI app, it will attempt to preprocess styles in any addons which include a `/addon/styles` folder. The style preprocessor assumes that `outputPaths` will be passed along with the options, and fails since addons do not.

It also fails in the `getExt` function because addons do not pass the full directory to the preprocessor. Unfortunately, because the tree passed by the addons resolves to the styles directory, there isn't a simple way to get the full path of the styles directory, so the current solution I've implemented is to pass it in as an additional option. 

Finally, the preprocessor will get called on any addon, regardless of whether or not the addon actually uses a preprocessor. To solve this each addon's registry is passed the dependencies of the addon rather than the dependencies of the parent project. This way the addon will only invoke the preprocessor if it is also required by the addon (i.e. the addon uses the preprocessor, and we can expect an app.[sass/scss/less/styl] there).

TODO: 
* This will not support addons which use a preprocessor that is not included in the parent app. In order to support that, the dynamics of nested addons need to be established and the behavior of  `require-local` will have to be changed.
* Find a more elegant solution to the `getExt` problem.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -332,6 +332,7 @@ Addon.prototype.compileStyles = function(tree) {
   if (tree) {
     var self = this;
     var processedStyles = preprocessCss(tree, '/', '/', {
+      stylesPath: path.join(this.root, '/addon/styles'),
       registry: this.registry
     });
 
@@ -530,6 +531,16 @@ Addon.prototype.config = function (env, baseConfig) {
 
     return configGenerator(env, baseConfig);
   }
+};
+
+/**
+  @public
+  @method dependencies
+  @return {Object} The addon's dependencies based on the addon's package.json
+*/
+Addon.prototype.dependencies = function() {
+  var pkg = this.pkg || {};
+  return assign({}, pkg['devDependencies'], pkg['dependencies']);
 };
 
 /**
--- a/tests/acceptance/preprocessor-smoke-test-slow.js
+++ b/tests/acceptance/preprocessor-smoke-test-slow.js
@@ -1,0 +1,97 @@
+'use strict';
+
+var path       = require('path');
+var fs         = require('fs');
+var expect     = require('chai').expect;
+
+var runCommand          = require('../helpers/run-command');
+var acceptance          = require('../helpers/acceptance');
+var copyFixtureFiles    = require('../helpers/copy-fixture-files');
+var assertDirEmpty      = require('../helpers/assert-dir-empty');
+var createTestTargets   = acceptance.createTestTargets;
+var teardownTestTargets = acceptance.teardownTestTargets;
+var linkDependencies    = acceptance.linkDependencies;
+var cleanupRun          = acceptance.cleanupRun;
+
+var appName  = 'some-cool-app';
+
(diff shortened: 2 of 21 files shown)


lib/models/addon.js
```diff
@@ -335,6 +336,11 @@ Addon.prototype.config = function (env, baseConfig) {
   }
 };
 
+Addon.prototype.dependencies = function(pkg) {
+  pkg = pkg || this.pkg || {};
```
stefanpenner: we try to not re-use argument variable names, as it makes debugging annoying.

stefanpenner:
this looks good, can we get some test coverage?

pzuraq:
@stefanpenner I didn't think to add tests since it's mostly refactor, but now that you mention it I can definitely add one to verify that the addons only activate if the style preprocessor is a dependency. I'll see if I can think of any others to add and get them in tonight.

lib/models/addon.js
pzuraq: Ok, should it be something along the lines of `var package = pkg || this.pkg || {}`? I lifted this method from the EmberApp class and left it otherwise unchanged.

stefanpenner:
@pzuraq awesome that would be great.

pzuraq:
@stefanpenner Tests added. Took a while for me to figure out the acceptance test system and then I had to rebase to include the latest changes. I added a new slow test file for running tests against preprocessors specifically, as I am planning on doing some more work on them over the weekend.

tests/acceptance/preprocessor-smoke-test-slow.js
```diff
@@ -0,0 +1,127 @@
...
+var expect     = require('chai').expect;
+
+var runCommand       = require('../helpers/run-command');
+var buildApp         = require('../helpers/build-app');
+var copyFixtureFiles = require('../helpers/copy-fixture-files');
+
+var appName  = 'some-cool-app';
+
+describe('Acceptance: preprocessor-smoke-test', function() {
+  before(function() {
```
chadhietala: Probably should use the new pattern for "slow" tests that was introduced in #3013.

chadhietala:
Probably going to need a rebase. Also see my comment about acceptance tests. I think it will be good to merge after that.

/cc @stefanpenner

pzuraq:
@chadhietala @stefanpenner Rebased and switched to the new pattern for acceptance tests, which is much better by the way (thanks @chadhietala!)

pzuraq:
@chadhietala Tests are failing due to an assertion in the new acceptance test pattern. I think it's because the assertion to check if `tmp/` is empty is called before tmp is actually emptied, but they are both async calls so sometimes it works and sometimes it doesn't. I submitted a PR #3048 and will resubmit this one once it's resolved.

stefanpenner:
@pzuraq master is now green, try a rebase if you have time.

pzuraq:
@stefanpenner Rebased and passing tests, think it's finally ready to merge!

PR #2937, opened 2015-01-09
[BUGFIX] only include dependencies(not devDependencies) of included addons

### LINKED ISSUE #2935: Using an addon that uses ember-qunit fails in projects that use ember-mocha
I'm creating an addon, which has ember-qunit by default. When I install this in our project, which uses ember-mocha, I get an error message `Path or pattern "passed.png" did not match any files`. I'm not sure where the issue is coming from - ember-cli, ember-qunit or ember-mocha. Should I create an issue for this on one of these projects?

The following makes the issue go away, but then the QUnit UI is also displayed after the mocha UI.

```
npm install --save-dev ember-cli-qunit
bower install ember-qunit-notifications ember-qunit qunit
```

jakecraige:
It used to include devDependencies which causes issues when developing locally.
It can also lead to a hard to debug error where something works locally because
you addon is linked and works but when installing from npm it doesn't.

closes #2935

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -130,9 +130,15 @@ Project.prototype.require = function(file) {
 
 Project.prototype.emberCLIVersion = emberCLIVersion;
 
-Project.prototype.dependencies = function(pkg) {
+Project.prototype.dependencies = function(pkg, excludeDevDeps) {
   pkg = pkg || this.pkg || {};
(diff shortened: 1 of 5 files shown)


aexmachina:
Any idea when this is going to be merged?

rwjblue:
@aexmachina - Yep

PR #2939, opened 2015-01-09
Add a hook for postprocessing tests tree

ef4:
This adds a postprocessTree hook for the tests tree, analogous to the one we already have for app js.

With this, ember-browserify will support importing npm dependencies within the tests tree.


rwjblue:
:+1: - Can we have a test for this also?

ef4:
Unit test added. It just verifies that every flavor of the addon postprocessTree hook gets invoked -- currently 'all', 'js', and 'test'.

rwjblue:
Perfect, thank you!

PR #2947, opened 2015-01-09
Removes broccoli-writer

twokul:


rwjblue:
Why?

twokul:
@rwjblue why do we need to different writers?

PR #2951, opened 2015-01-10
Fix for aliases for hyphenated options

### LINKED ISSUE #2943: Alias for --output-path isn't working for the build command in 0.1.5
When trying to build using `ember build -e foo -o dist-foo/` I get the following error:

```
The option '--outputPath' is not supported by the build command. Run `ember build --help` for a list of supported options.
Build failed.
Arguments to path.resolve must be strings
```

It works fine with the full argument `--output-path`

marcioj:
For instance when we run `ember build -o dist-foo`, inside of `Command#parseAlias` method `option.key` will return `outputPath`, and pass something like this to `nopt`:

`nopt({ 'output-path': path }, { o: [ '--outputPath' ] }, [ '-o', 'dist-foo' ], 0)`
which returns
`{ outputPath: true,
  argv:
   { remain: [ 'dist-foo' ],
     cooked: [ '--outputPath', 'dist-foo' ],
     original: [ '-o', 'dist-foo' ] } }`

Since `--outputPath` is different of `--output-path`. `nopt` doesn't get what we want and return a boolean instead `{ outputPath: true }`

Using `option.name` will use the original option name with hyphen and satisfy the `nopt` parsing:

`nopt({ 'output-path': path }, { o: [ '--output-path' ] }, [ '-o', 'dist-foo' ], 0)`
which returns
`{ 'output-path': '/path/to/dist-foo',
  argv:
   { remain: [],
     cooked: [ '--output-path', 'dist-foo' ],
     original: [ '-o', 'dist-foo' ] } } `

Fixes #2943

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -256,10 +256,10 @@ Command.prototype.parseAlias = function(option, alias) {
   if (isValidAlias(alias, option.type)) {
     if (aliasType === 'string') {
       key = alias;
-      value = ['--' + option.key];
+      value = ['--' + option.name];
     } else if (aliasType === 'object') {
       key = Object.keys(alias)[0];
-      value = ['--' + option.key, alias[key]];
+      value = ['--' + option.name, alias[key]];
     }
   } else {
     if (Array.isArray(alias)) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thanks!!

PR #2952, opened 2015-01-10
Fixes validation of shorthand commandOptions

trabus:
The command options validation is incorrectly warning on dasherized options when the shorthand form is used (as reported in #2943).

`ember init -d`
results  in a warning:
`The option '--dryRun' is not supported by the init command. Run `ember init --help` for a list of supported options.`

The `--dry-run` option is still picked up, however:
```
installing
You specified the dry-run flag, so no changes will be written.
```

The command option validation now uses `hasOwnProperty` instead of `_.findKey`, and unknown (or misspelled) options are added to command options after a warning. Also reworded the warning:
`The option '--mispelledoption' is not registered with the serve command. Run `ember serve --help` for a list of supported options.`

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -18,7 +18,6 @@ var reject        = require('lodash-node/modern/collections/reject');
 var EOL           = require('os').EOL;
 var CoreObject    = require('core-object');
 var debug         = require('debug')('ember-cli:command');
-var findKey       = require('lodash-node/modern/objects/findKey');
 var Watcher       = require('../models/watcher');
 var SilentError   = require('../errors/silent');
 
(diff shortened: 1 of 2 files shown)


lib/models/command.js
```diff
@@ -18,7 +18,7 @@ var reject        = require('lodash-node/modern/collections/reject');
 var EOL           = require('os').EOL;
 var CoreObject    = require('core-object');
 var debug         = require('debug')('ember-cli:command');
-var findKey       = require('lodash-node/modern/objects/findKey');
+// var findKey       = require('lodash-node/modern/objects/findKey');
```
jakecraige: let's just remove it if it's not going to be used anymore

trabus: Crap, I thought I did.

rwjblue:
Awesome, thank you!

PR #2953, opened 2015-01-11
Clarify create command

samselikoff:
It looked like the command was missing, because of the :


samselikoff:
Nvm, missed the command from the above section

PR #2960, opened 2015-01-12
Move from esnext to 6to5

stefanpenner:
esnext and 6to5 efforts have combined into a super pack of future es syntax goodness!

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -24,7 +24,7 @@
     "ember-cli": "<%= emberCLIVersion %>",
     "ember-cli-content-security-policy": "0.3.0",
     "ember-cli-dependency-checker": "0.0.7",
-    "ember-cli-esnext": "0.1.1",
+    "ember-cli-6to5": "0.2.1",
     "ember-cli-ic-ajax": "0.1.1",
     "ember-cli-inject-live-reload": "^1.3.0",
     "ember-cli-qunit": "0.1.2",


stefanpenner:
bower appears down...

stefanpenner:
so far the recent perf work but @sebmck look quite nice!

chadhietala:
:+1:

chadhietala:
I'm assuming we should remove the module transpiler as well and then browserify? Or is there perf concerns with that?

stefanpenner:
> I'm assuming we could remove the module transpiler since 6to5 compiles modules to CJS and then we would just browserify?

i wonder what the perf of browserify concat is like, compared to what we currently have?

chadhietala:
The `npm:foo` syntax is fantastic.  I wish there was a `es6:foo` that did the same thing but transpiled as well. Perhaps this is something I can work on.

stefanpenner:
loader.js has streams now, the 1.x branch doesn't support resolving relative paths during require, rather for dependencies only. 2.x supports that, but announces itself as an AMD compliant loader, which depends on a r.js esq step.

cibernox:
I've tried yesterday this, and:

1) Everything worked like a charm (I don't use a ton of ES6 feature though, just rest args and arrow functions)
2) Build times of the 6to5 tree are much faster than es-next!

sebmck:
@cibernox Should be hoping to boost performance even more!

cibernox:
Something I would love to have is to have sourcemaps that show me the ES6 source instead of the resulting JS.

sebmck:
@cibernox 6to5 (and previously esnext) had source map support.

cibernox:
In that case something is not working for me, because I see the transpiled code in dev toos. I have JS sourcemaps enabled in chrome. Do I need to configure anything else?

stefanpenner:
Our concat step doesn't merge inline sources yet. Hopefully shortly it will.

I'm slightly nervous about the cost of that during the concat phase, I'm sure we can keep it to a minimum.

stefanpenner:
fixes: https://github.com/ember-cli/ember-cli/pull/2960

QuantumInformation:
Great! Looking forward to arrow functions etc al

et:
Can we confirm whether the `ember test` works with this?
I tried using a native `Set` and it works fine with `ember serve` but I am getting the following error in the test suite: `Can't find variable: Set`

et:
Disregard my last comment. It's broken in PhantomJS but works in Chrome. Seems like it's not an ember-cli issue.

PR #2965, opened 2015-01-12
Remove Hardcoded test-loader

chadhietala:
This is a backwards compatible change and removes the hardcoded `destDir` for the test-loader.


chadhietala:
I think appveyor needs a :facepunch: or something. Appears to be queuing everyone's PRs.

PR #2975, opened 2015-01-13
0.1.7 changelog.

abuiles:


CHANGELOG.md
```diff
@@ -2,13 +2,37 @@
...
-
 ### Applications
 ### Addons
 ### Blueprints
 
+### 0.1.7
+
+- Users
+  + Upgrade your project's ember-cli version - [docs](http://www.ember-cli.com/#project-update)
+  + Replace `ember-cli-esnext` with `ember-cli-6to5`:
```
jakecraige: This could be changed to something like:

Uninstall `ember-cli-esnext`. It has been replaced by `ember-cli-6to5` which should be added to your project from the upgrade steps.

stefanpenner: we should also tell people that if the module transpile step is stil es3 safe. Aspects of 6to5 are fine on es3, but if they are concerned about the temptation they can remove it package.jso entirely..

abuiles: @stefanpenner great, adding that as a note 

#### Using 6to5.

The module transiple is stil es3 safe,  if you are concern about using
6to5 you can remove it from package.json.

PR #2983, opened 2015-01-13
Update QUnit version

wagenet:


jakecraige:
We should also update this [here](https://github.com/ember-cli/ember-cli-qunit/blob/master/blueprints/ember-cli-qunit/index.js) when merging

rwjblue:
Upstream PR: https://github.com/ember-cli/ember-cli-qunit/pull/27

PR #3007, opened 2015-01-15
[Enhancement] - adds a watcher option to the build command

rauhryan:
Allows you to specify the watcher type from the
build command, this is useful for running build inside
a vagrant box where file watching doesn't work very
well

Specifically when using ember-cli with ember-cli-rails
which compiles the ember app automatically using build
instead of serve


stefanpenner:
:+1:

stefanpenner:
thanks!

PR #3009, opened 2015-01-16
Tweak error message for executing `ember unknownCommand`

artfuldodger:
The English syntax is more correct, and the JavaScript
was updated a bit.


stefanpenner:
awesome, thanks for the cleanup PR. It looks much better now.

PR #3011, opened 2015-01-16
Add folders referenced in documentation to blueprint.

patricksrobertson:
In the website (http://www.ember-cli.com/#naming-conventions)
there's mention of adapters/initializers/mixins/serializers/transfors/
utils but they don't ship with the `ember new` command. Perhaps
they should though!


jgwhite:
Thanks for the PR.

We discussed including these directories by default but came to the conclusion that not every project will need them. We’d rather introduce a minimal set of directories than overwhelm new users.

Check out the conversation in #263 for some history.

patricksrobertson:
I think as a new user -- it's more of a pain in my side to go to the docs and figure out that the Ember-CLI convention for naming my utilities folder is 'utils' and then doing it. At the very least it's so routine that I've had to reopen the `ApplicationAdapter` or build a new adapter for a specifically named model that I feel like that makes the most sense to include the adapters directory by default.

I'm unsure how a folder overwhelms a new user, but will defer to y'all on it as its your library.

jgwhite:
Good points. Few other things to bear in mind:

- Some users may want to use the pods structure
- Not every project will be using ember-data
- `ember generate` will create the correct folder structure when you need it

Out of interest: do you tend to use `ember generate adapter` and `ember generate util` or create the files by hand?

patricksrobertson:
I generate by hand, but it seems like the best practice is to use the generator commands :).

Perhaps using generator commands should be included in the naming conventions section to outline how easy it is to keep Ember kosher with the generators.

patricksrobertson:
regards to pods (this is bigger than a folder): but as a newer convert having two conventions is pretty lame. It should be all pods, all the time, or no pods.

jgwhite:
I think the plan is pods by default. Still gathering consensus from the community but everyone seems in favour.

jgwhite:
Also: good call on referencing the generators in the naming conventions section. Would welcome a PR to that effect :smile:

patricksrobertson:
Alright, closing this one then.

PR #3014, opened 2015-01-16
Enhancement: Move app preview window to be at the bottom of the page so ...

TayHobbs:
...that it won't cover the tests when run in the browser.

After a certain number of tests, the preview window will cover half of the test description, this would fix that by keeping the preview window below all of the descriptions.

### DIFF
--- a/blueprints/app/files/tests/index.html
+++ b/blueprints/app/files/tests/index.html
@@ -15,7 +15,7 @@
     <link rel="stylesheet" href="assets/test-support.css">
     <style>
       #ember-testing-container {
-        position: absolute;
+        position: relative;
         background: white;
         bottom: 0;
         right: 0;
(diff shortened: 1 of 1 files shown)


rwjblue:
I am in favor, but we need to actually change the CSS in ember-cli-qunit [here](https://github.com/ember-cli/ember-cli-qunit/blob/master/vendor/ember-cli-qunit/test-container-styles.css).  Then update ember-cli itself to simply not use the inline styles at all (they cause CSP errors).

stefanpenner:
@TayHobbs got time to work on @rwjblue's suggestion?

TayHobbs:
Yeah, I'll look into it

stefanpenner:
thanks, closing this PR as with his suggestions it is currently invalid.

Let us know if u run into any issues.

Issue #2207, opened 2014-10-05
Please allow "ember-cli" to be in "dependencies"


vladikoff:
Currently with this code:

```
Project.prototype.isEmberCLIProject = function() {
  return this.pkg.devDependencies && 'ember-cli' in this.pkg.devDependencies;
};
```

if `ember-cli` is in dependencies then the project fails to build with 

```
You have to be inside an ember-cli project in order to use the build command.
```

PR #3015, opened 2015-01-16
Enhancement: Allow `ember-cli` to be included in `dependencies` in

TayHobbs:
package.json rather than only being allowed in `devDependencies`.

Addresses issue #2207

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -66,7 +66,12 @@ Project.prototype.name = function() {
 };
 
 Project.prototype.isEmberCLIProject = function() {
-  return this.pkg.devDependencies && 'ember-cli' in this.pkg.devDependencies;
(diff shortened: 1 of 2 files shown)


lib/models/project.js
```diff
@@ -66,7 +66,12 @@ Project.prototype.name = function() {
 };
 
 Project.prototype.isEmberCLIProject = function() {
-  return this.pkg.devDependencies && 'ember-cli' in this.pkg.devDependencies;
+  if (this.pkg.devDependencies && 'ember-cli' in this.pkg.devDependencies) {
```
rwjblue: I believe that this can use the `dependencies` function from below, which already merges the devDeps and deps.

```javascript
return 'ember-cli' in this.dependencies();
```

stefanpenner:
@rwjblue indicates we already do this.

PR #3018, opened 2015-01-17
Remove ember-data from package.json

tonywok:
Noticed ember-data was in both the `package.json` and the `bower.json`. Verified that my app uses the version included by bower.

Please close if it's there for a reason I'm not aware of.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -28,7 +28,6 @@
     "ember-cli-ic-ajax": "0.1.1",
     "ember-cli-inject-live-reload": "^1.3.0",
     "ember-cli-qunit": "0.1.2",
-    "ember-data": "1.0.0-beta.12",
     "ember-export-application-global": "^1.0.0",
     "express": "^4.8.5",
     "glob": "^4.0.5"


quaertym:
Ember Data is also an ember cli addon and addons are discovered by the package.json as far as I know. I think it should stay.

zeppelin:
Shouldn't these lines supposed to take care of the module shims & importing ED to the project? https://github.com/emberjs/data/blob/master/lib/ember-addon/index.js#L14-L29

tonywok:
@zeppelin That's what I thought, but I don't have enough knowledge at the moment to be sure.

quaertym:
@zeppelin @tonywok This is how addons are discovered: https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L169-183

zeppelin:
@tonywok in the light of the evidents above, I believe the issue should be closed ;)

jakecraige:
As @quaertym said, this is intentional. It's in the `bower.json` by default so that we don't have to run the ember-data blueprint as well when creating an app.

tonywok:
Ah, okay. Good to know. Thanks :)

PR #3029, opened 2015-01-19
Add a simple example of how to disable minifying

leandrocp:
It´s very useful to debug production errors (like Handlebars.compile undefined), although it´s not very clear for beginners how to disable minify.


chadhietala:
Thanks!

PR #3031, opened 2015-01-19
Use symlink-or-copy so acceptance slow tests don't fail on windows

chadhietala:
Hopefully this fixes the Appveyor failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,7 @@
     "nock": "^0.51.0",
     "node-require-timings": "0.0.2",
     "supertest": "0.14.0",
+    "symlink-or-copy": "^1.0.1",
     "tmp-sync": "^1.0.1",
     "yuidocjs": "0.3.50"
   }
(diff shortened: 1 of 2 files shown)


chadhietala:
@jakecraige want to review?

stefanpenner:
@chadhietala seems like a good addition, except this means our windows builds will not benefit from the fast symlinking. It is likely worth seeing if we can enable the windows linking code on CI (like http://www.ember-cli.com/#symlinks-on-windows) Maybe the appveyor folks can point us in the right direction.

stefanpenner:
this is :+1: from me, although ^^ would be awesome as time permits.

rwjblue:
Symlink or copy does symlink if possible on windows

stefanpenner:
@rwjblue ya, we need to see if we can enable that mode on appveyor, by pulling the correct windows levers.

PR #3035, opened 2015-01-20
Fix url for proxies on Windows

### LINKED ISSUE #2796: path.join in proxy blueprint fails on windows
In the proxy blueprint we have

    req.url = path.join(proxyPath, req.url);

But this would seem to be incorrect. `path.join` does a normalization which under Windows changes slashes to backslashes (ouch!). `path.join` should not be used for URLs, only for file system paths.

This was introduced as part of commit https://github.com/stefanpenner/ember-cli/commit/431e6986d583f2a18522c67529058495d246259f made by @tstirrat.

chadhietala:
Should close #2796

### DIFF
--- a/blueprints/http-proxy/files/server/proxies/__name__.js
+++ b/blueprints/http-proxy/files/server/proxies/__name__.js
@@ -4,7 +4,6 @@ module.exports = function(app) {
   // For options, see:
   // https://github.com/nodejitsu/node-http-proxy
   var proxy = require('http-proxy').createProxyServer({});
-  var path = require('path');
 
   proxy.on('error', function(err, req) {
     console.error(err, req.url);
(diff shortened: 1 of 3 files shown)


chadhietala:
@stefanpenner :+1: or :-1: ?

blueprints/http-proxy/files/server/proxies/__name__.js
```diff
@@ -12,7 +12,7 @@ module.exports = function(app) {
 
   app.use(proxyPath, function(req, res, next){
     // include root path in proxied request
-    req.url = path.join(proxyPath, req.url);
+    req.url = urlJoin(proxyPath, req.url);
```
stefanpenner: im curious why we can just `proxyPath +  '/' + req.url`

chadhietala: Nope just over doing it to the nth degree.

chadhietala:
Not really sure what is going on with CI.

stefanpenner:
many CI things are fixed, can you rebase from master and lets see what happens.

stefanpenner:
pulled in another simpler PR that does this.

Issue #1300, opened 2014-07-07
0.0.39 "file in the root of `vendor/`" error messages for _everything_


ccoenen:
After upgrading to 0.0.39, i now get these messages at startup. Is this a known issue or is it a problem with my setup? In case i'm missing something really simple, I'd be glad for any pointers.

```
$ ember s
version: 0.0.39
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/loader/loader.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/jquery/dist/jquery.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/handlebars/handlebars.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember/ember.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-cli-shims/app-shims.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-resolver/dist/modules/ember-resolver.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-load-initializers/ember-load-initializers.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ic-ajax/dist/named-amd/main.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-data/ember-data.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/d3/d3.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/jqzabbix/jqzabbix.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-qunit/dist/named-amd/main.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `vendor/ember-cli-shims/test-shims.js` into a subdirectory.
Livereload server on port 35729
Serving on http://0.0.0.0:4200
```

Issue #1261, opened 2014-07-04
Incorrect warnings about vendor files not in root


basepack:
After upgrading to 0.0.39 I got a bunch of warnings:

```
version: 0.0.39
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/loader/loader.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/jquery/dist/jquery.js` into a subdirectory
.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/handlebars/handlebars.js` into a subdirect
ory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember/ember.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-cli-shims/app-shims.js` into a subdi
rectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-resolver/dist/modules/ember-resolver
.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-load-initializers/ember-load-initial
izers.js` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-data/ember-data.js` into a subdirect
ory.
app.import(vendor/ember-data/ember-data.js) - Passing modules object is deprecat
ed. Please pass an option object with modules as export key (see http://git.io/H
1GsPw for more info).
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ic-ajax/dist/named-amd/main.js` into a sub
directory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-simple-auth/simple-auth.amd.js` into
 a subdirectory.
app.import(vendor/ember-simple-auth/simple-auth.amd.js) - Passing modules object
 is deprecated. Please pass an option object with modules as export key (see htt
p://git.io/H1GsPw for more info).
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `vendor/ember-simple-auth/simple-auth-oauth2.amd.j
s` into a subdirectory.
[...]

PR #3036, opened 2015-01-20
[BUGFIX] Fix platform dependent path separator, close #2954

### LINKED ISSUE #2954: Incorrect warning about files not being in sub directories (Windows)
When I upgraded my project to cli@0.1.6 I got warnings like this

version: 0.1.6

```javascript
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `bower_components\jquery-ui\jquery-ui.js` into a s
ubdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `bower_components\jquery-ui\themes\base\accordion.
css` into a subdirectory.
Using `app.import` with a file in the root of `vendor/` causes a significant per
formance penalty. Please move `bower_components\jquery-ui\themes\base\all.css` i
nto a subdirectory.
```
Files are clearly in the sub directories.

I think this issue may be linked to platform. I am using Windows 7 (32 bit).

KarimBaaba:
@stefanpenner here is a quick PR you asked for #2954 

I did a quick grep on the project and noticed other potential bugs from using hardcoded separators. Mind if I take a stab at it and amend this?

Thanks!

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -5,6 +5,7 @@
 @module ember-cli
 */
 var fs    = require('fs');
+var path  = require('path');
 var p     = require('../preprocessors');
 var chalk = require('chalk');
 var EOL   = require('os').EOL;
@@ -943,6 +944,8 @@ EmberApp.prototype.import = function(asset, options) {
   if (!assetPath) {
     return;
   }
+
+  assetPath = assetPath.replace(path.sep, '/');
 
   if (assetPath.split('/').length < 2) {
     console.log(chalk.red('Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `'+ assetPath + '` into a subdirectory.'));


rwjblue:
We previously had this using `path.sep`, but had to change it due to many reports of breakage on Windows.  

References:

https://github.com/ember-cli/ember-cli/issues/1261
https://github.com/ember-cli/ember-cli/pull/1264
https://github.com/ember-cli/ember-cli/issues/1300

stefanpenner:
hmmm

stefanpenner:
@rwjblue https://github.com/ember-cli/ember-cli/issues/2954 appears to be the exact opposite

KarimBaaba:
These changes don't seem to be the cause of tests failing( installed latest node/npm from chocolatey on win7 & ran npm test). Tests seem to also fail on this commit's parent aka `HEAD^`

I have a feeling that the original issue is unrelated to path separators since Windows typically accepts forward slashes then converts them... I'll rebase this and update it, build should pass then.

rwjblue:
app.import should always use forward slashes (likely that is the original source of confusion).

rwjblue:
I realized that was quite terse (sorry), a bit more explanation:

The vast majority of addons use forward slashes, which is why we can't use path.sep, and node knows how to properly handle forward AND backward slashes (which is why we haven't had to change any 'require' calls to use path.sep). The main issue with forward vs backward path comes from dealing with absolute paths (which is not the case here).

lib/broccoli/ember-app.js
```diff
@@ -874,7 +875,7 @@ EmberApp.prototype.import = function(asset, options) {
     return;
   }
```
stefanpenner: as node internally normalizes most `\`  or `/` to platform specific separator, we should do the same

```js
assetPath = assetPath.replace(path.sep, '/')
```
before that conditional, i believe will do the trick

rwjblue: :+1:

KarimBaaba: Done. 

Tiny caveat, if users are adding paths that include whitespaces and try to cleverly escape them then this will break. Node is capable of handling the following:
```
app.import('bower_components/some lib/some file.js');
``` 
But this will break:
```
app.import('bower_components/some\ lib/some\ file.js');
``` 
I doubt that anyone is doing this but food for thought :information_desk_person:

rwjblue:
Thank you!

PR #3039, opened 2015-01-20
Move static file check earlier so it only affects the default value

krisselden:
This is determining the best default value, by moving it earlier, it allows vendorFiles to refer to an import from the vendor broccoli tree, not just statically available in the root.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -126,6 +126,13 @@ function EmberApp(options) {
     extensions: ['js']
   }, defaults);
 
+  // in Ember 1.10 and higher `ember.js` is deprecated in favor of
+  // the more aptly named `ember.debug.js`.
+  var defaultDevelopmentEmber = this.bowerDirectory + '/ember/ember.debug.js';
+  if (!fs.existsSync(path.join(this.project.root, defaultDevelopmentEmber))) {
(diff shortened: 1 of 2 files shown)


stefanpenner:
@krisselden can you add a test?

krisselden:
@stefanpenner the original check wasn't tested, looking into it now

stefanpenner:
@krisselden thanks, i know its a pain, but this way we don't regress in future refactorings.

krisselden:
@stefanpenner @rwjblue  added the missing tests for the fallback and for this change and fixed an unrelated vendorFiles test.

stefanpenner:
re-running appveyor looks like a timeout

krisselden:
@stefanpenner because I tried to write and unlink a file synchronously which windows isn't happy about.  I just pushed a fix, but I'm thinking likely should be just a project fixture with the default ember.debug.js file it is looking for.

krisselden:
@stefanpenner @rwjblue I replaced the setup and teardown with a fixture, this should be good to go after CI has run.

krisselden:
@stefanpenner can you restart? I'm pretty sure this last failure has nothing to do with my tests.

rwjblue:
LGTM, pending appveyor only at this point.

stefanpenner:
:+1:

PR #3048, opened 2015-01-21
Fix acceptance slow test helper by moving assertion

pzuraq:
The new acceptance test pattern typically passes an assertion to the teardown helper asserting that the `tmp/` directory should be empty after each test. Because the assertion is called before the folder is actually torn down, sometimes the test fails. This fixes the issue by calling the assertion after the promise for destroying the `tmp/` directory has resolved.

### DIFF
--- a/tests/helpers/acceptance.js
+++ b/tests/helpers/acceptance.js
@@ -163,13 +163,9 @@ function linkDependencies(projectName) {
 
 /**
  * Clean a test run and optionally assert.
- * @param  {Function} [assertion] An assertion that gets ran at the end of a run
  * @return {Promise}
  */
-function cleanupRun(assertion) {
-  if (assertion) {
-    assertion();
-  }
+function cleanupRun() {
   return tmp.teardown('./tmp');
 }
 
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -41,7 +41,7 @@ describe('Acceptance: addon-smoke-test', function() {
 
   afterEach(function() {
     this.timeout(15000);
-    return cleanupRun(function() {
+    return cleanupRun().then(function() {
       assertDirEmpty('tmp');
     });
   });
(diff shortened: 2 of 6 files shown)


tests/helpers/acceptance.js
```diff
@@ -167,10 +167,11 @@ function linkDependencies(projectName) {
  * @return {Promise}
  */
 function cleanupRun(assertion) {
-  if (assertion) {
-    assertion();
-  }
-  return tmp.teardown('./tmp');
+  return tmp.teardown('./tmp').then(function() {
```
stefanpenner: :+1: good catch sir.

stefanpenner:
the most recent failure looks related to invalid symbol usage in the module transpiler,

https://github.com/ef4/broccoli-es6modules/pull/4 would fix it, but we are pending some work by @ef4 i believe.

chadhietala:
@pzuraq Good catch. I figured there was a race condition somewhere because I could not replicate locally.

@stefanpenner Yea symbol usage seems to be the issue.

stefanpenner:
> @stefanpenner Yea symbol usage seems to be the issue.

seems like a recent version of node `0.11.x` got symbols, and the transpilers traceur isn't using them correctly.

chadhietala:
@stefanpenner Yea got to love those ranged versions :trollface:.

stefanpenner:
@chadhietala nono, its the fact that real Symbols are in iojs and node 0.11.x and the old version of traceur doesn't support it.

jakecraige:
Since we're now returning a promise, can we just chain the assertion in the implementation rather than using a callback? 

```js
return cleanupRun(function() {
  assertDirEmpty('tmp');
});
// to
return cleanupRun().then(function() {
  assertDirEmpty('tmp');
});
```


affected lines [here](https://github.com/ember-cli/ember-cli/blob/1c45f4651ffe77c9d42c2fcdcc3a0a28ab1b827a/tests/acceptance/express-server-restart-slow.js#L44) and [here](https://github.com/ember-cli/ember-cli/blob/1c45f4651ffe77c9d42c2fcdcc3a0a28ab1b827a/tests/acceptance/smoke-test-slow.js#L40) it looks like

pzuraq:
@jakecraige good point, I'll update the fix accordingly

stefanpenner:
test failure is broken traceur in our module transpiler in the latest node 0.11.x. I am merging this because it does fix other issues.

will address the traceur issue shortly

PR #3049, opened 2015-01-21
Test helper blueprint

stefanpenner:


chadhietala:
:+1:

stefanpenner:
ya, we are being trolled by that Symbol issue.

i am about to get some sleep, it is possible that a more recent 0.3x version works alright: https://github.com/esnext/es6-module-transpiler/tree/v0.3.x might be a stopgap until the esperanto effort is complete

cc @eventualbuddha

rwjblue:
IMHO, we should be wary of adding new blueprints without tests (for the app).  Specifically, how do you test your test helper here?  Just manually (maybe that is fine?), some other demo app, whatever?

@stefanpenner - thoughts?

stefanpenner:
@rwjblue I agree, we should add a smoke test, or addd it to an existing smoke test.

PR #3053, opened 2015-01-21
Remove forward slash from image path.

dinebennett:
We found that the slash prevented the image from being displayed in a phonegap application.


stefanpenner:
thanks!

PR #3056, opened 2015-01-21
[WIP] Update core object

jgwhite:
**Current status:** still testing against addons.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -51,574 +51,657 @@ var walkSync   = require('walk-sync');
   @param {Project or Addon} parent The project or addon that directly depends on this addon
   @param {Project} project The current project (deprecated)
 */
-function Addon(parent, project) {
-  this.parent = parent;
-  this.project = project;
-  this.ui = project && project.ui;
-  this.addonPackages = {};
-  this.addons = [];
-  this.addonDiscovery = new AddonDiscovery(this.ui);
-  this.addonsFactory = new AddonsFactory(this, this.project);
-  this.registry = p.defaultRegistry(this);
-  this._didRequiredBuildPackages = false;
+var Addon = CoreObject.extend({
+  init: function(parent, project) {
+    this.parent = parent;
+    this.project = project;
+    this.ui = project && project.ui;
+    this.addonPackages = {};
+    this.addons = [];
+    this.addonDiscovery = new AddonDiscovery(this.ui);
+    this.addonsFactory = new AddonsFactory(this, this.project);
+    this.registry = p.defaultRegistry(this);
+    this._didRequiredBuildPackages = false;
 
-  if (!this.root) {
-    throw new Error('Addon classes must be instantiated with the `root` property');
-  }
-  this.nodeModulesPath = nodeModulesPath(this.root);
+    if (!this.root) {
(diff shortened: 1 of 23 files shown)


lib/models/addon.js
```diff
@@ -9,7 +9,6 @@ var SilentError = require('../errors/silent');
 var reexport    = require('../utilities/reexport');
 var escapeRegExp= require('../utilities/escape-regexp');
 
-
 var p                   = require('../preprocessors');
```
chadhietala: Can we rename this to preprocessors?

stefanpenner: :+1:

jgwhite: Totally.

krisselden:
@jgwhite after working with the old core object, I'm a big fan of this.

jgwhite:
This now passes but the changes are extensive. :grimacing: 

Making the switch without breaking our existing crop of addons involves two awkward contortions:

- [Project detects and uses the appropriate constructor signature for addons](https://github.com/jgwhite/ember-cli/commit/684c52169c3db80386e4a20f319d5f7f11cbd00f)
- [A modified version of core-object that makes the `_super` available as `_super.someMethod`](https://github.com/jgwhite/core-object/commit/029e3fbdf06159460f3f35951e17ffa1d39926e3)

Other concerns:

- Omitting the `this._super.apply(this, arguments)` in `init` will almost always break things
- The difference between `CoreObject#init` and `Ember.CoreObject#init` may cause confusion

stefanpenner:
> Omitting the this._super.apply(this, arguments) in init will almost always break things

we can `fn.toString().indexOf('._super')`

trabus:
Wow, awesome work! :clap:

stefanpenner:
@jgwhite im nervous that this is getting way out of sync. Ping me, I would like to be sure to unblock you, or take the reigns before we diverge entirely, I would love to get this  in asap.

jgwhite:
@stefanpenner I’ve been rebasing fairly regularly and I believe this is in a mergeable state right now.

jgwhite:
Oops. Spoke too soon!

stefanpenner:
@jgwhite we should sync up, ping me when you are free. Would love to unblock you :)

stefanpenner:
@jgwhite looks like lots of progress was made again :) Care to share the current blockers?

jgwhite:
@stefanpenner I’ve been trying to find a way to roll out the new CoreObject without changing exsting libraries but, the more I look at it, the more I think this will need a symlink-style unilateral upgrade. It’s mainly variations in expectations around `init` that cause trouble.

broccoli-caching-writer is a particularly off-beat culprit. What do you think of releasing a new-core-object-friendly version: https://github.com/ember-cli/broccoli-caching-writer/compare/ember-cli:master...jgwhite:upgrade-core-object

stefanpenner:
@jgwhite im not quite sure what that external lib was causing grief. But it does not need to change (and also no longer uses core object itself)

stefanpenner:
@jgwhite do you have interest in this, or should we close and move on?

jgwhite:
@stefanpenner much as it pains me to say it — let’s close and revisit later.

stefanpenner:
@jgwhite NP, future work :)

PR #3058, opened 2015-01-21
Prefer `EmberEnv.env()` to `process.env`

yaymukund:
`EmberEnv.env()` was built for this purpose and defaults to `'development'` which is nice.


jakecraige:
nice

PR #3030, opened 2015-01-19
Fix StylePlugin to readTree(inputTree) and getExt to take inputTreeRoot.

krisselden:
This is a first step in making my-addon/addon/styles and my-addon/app/styles work with style plugins, next step is to actually correctly setup options given to preprocessCSS.

PR #3063, opened 2015-01-22
Revert "Fixes style preprocessor behavior in addons"

krisselden:
The options that are passed to the StylePlugin via preprocessCss in the Addon model should include outputPaths instead of adding a new options.fileName.  The getExt check should run after reading the input tree #3030


pzuraq:
@krisselden Yes I absolutely agree, I thought your method for getting the input path was far superior to mine and was planning on integrating it after my PR was merged (didn't want to take away your PR!). I also think we should pass an outputPaths option, I didn't want to change too much so I followed the convention but it makes more sense to change the addons to pass the outputPaths.

I think some things from #2936 could be included and we could build on them with your suggestions. In particular:

* Adding dependencies functions to Addons so that addons only use a preprocessor if they require the preprocessor (else we will get errors from addons that do not have app.sass, etc.)
* The preprocessor slow tests, which set up a model for testing nested addons which will be useful in the near future.

If you'd like I can resubmit the PR with your changes integrated, like I said I didn't want to step on toes.

krisselden:
@pzuraq sounds good, just want to be careful of abstraction creep, it is a problem when being to fearful of being too invasive, but end up proliferating extra layers that cause more problems down the road.  I will look at your tests.

pzuraq:
@krisselden Totally. I'm going to resubmit the parts of my PR that weren't related to the style plugin separately and leave the style plugin to you.

PR #3064, opened 2015-01-22
Revert "Fixes style preprocessor behavior in addons"

chadhietala:
Reverts ember-cli/ember-cli#2936 

@pzuraq We unfortunately have to revert this and fix the root cause of the problem. See #3063


pzuraq:
No worries @chadhietala. I already posted a response in #3063, completely agree with the decision.

PR #3068, opened 2015-01-22
Hide passed tests by default.

rwjblue:
* Add `?hidepassed` to `testem.json` by default (which tells QUnit to hide passed tests).
* Fix an issue with using a `testem.json` `test_page` value that already contains a question mark along with the `--filter` option.
* Update tests to make assertions on the full `test_page` instead of just doing `indexOf` checks.


rwjblue:
@stefanpenner - r?

PR #3069, opened 2015-01-23
Fix style preprocessors for included addons - v2

pzuraq:
Updated version of PR #2936 with @krisselden's improved strategy for getting the file extension and passing output paths from PR #3030.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -330,20 +330,9 @@ Addon.prototype.compileStyles = function(tree) {
   this._requireBuildPackages();
 
   if (tree) {
-    var self = this;
-    var processedStyles = preprocessCss(tree, '/', '/', {
+    return preprocessCss(tree, '/', '/', {
+      outputPaths: { 'addon': this.name + '.css' },
       registry: this.registry
-    });
-
-    return new this.Funnel(processedStyles, {
-      getDestinationPath: function(relativePath) {
-        if (relativePath === 'app.css') {
-          return self.name + '.css';
-        }
-
-        return relativePath;
-      },
-      description: 'Funnel: Addon CSS'
     });
   }
 };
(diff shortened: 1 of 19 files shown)


rwjblue:
Looks good to me. @krisselden can you review since this supersedes #2936?

lib/models/addon.js
```diff
@@ -332,6 +332,7 @@ Addon.prototype.compileStyles = function(tree) {
   if (tree) {
     var self = this;
     var processedStyles = preprocessCss(tree, '/', '/', {
+      outputPaths: { 'addon': this.name + '.css' },
```
krisselden: the funnel below this that looks for app.css and renames it to this.name is unneeded

pzuraq:
Removed the funnel and am now returning preprocessCss directly.

tests/fixtures/preprocessor-tests/app-with-addon-without-preprocessors/node_modules/broccoli-sass/index.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var Writer = require('broccoli-writer');
+
+function copyPreserveSync (src, dest) {
```
stefanpenner: seems duplicated with the above ? or is something included incorrectly

pzuraq: @stefanpenner This is a dummy SASS preprocessor that's used by the tests, each test included it  in `node_modules/`. I was thinking it would be best to abstract it out somehow, since the new acceptance helper links against modules that already exist, but I wasn't sure what the best way to do it would be. For now I removed the second copy since it'll already be in `node_modules-tmp/`, will try to think of a better way to abstract out shared modules for tests in the future.

rwjblue:
Thank you!

sandstrom:
Awesome! :boat:

PR #3082, opened 2015-01-23
[DOCS] contentFor only receives type and config

chadhietala:
Writing an addon... verifying docs along the way.

### DIFF
--- a/ADDON_HOOKS.md
+++ b/ADDON_HOOKS.md
@@ -280,9 +280,8 @@ Allow addons to implement contentFor method to add string output into the associ
 
 **Received arguments:**
 
-- config
-- match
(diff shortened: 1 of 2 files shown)


rwjblue:
:+1: - Merge or wait and add more as you go, up to you.

chadhietala:
I'll keep it open for now just in case. Will merge in an hour or 2.

joostdevries:
Lgtm. The doc probably copied the signature from EmberApp.

joostdevries:
If you could add a jsdoc block in https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js too that'd be completely awesome by the way

Issue #3073, opened 2015-01-23
I am creating a ember-cli addon and like to include the css for that addon. Can anyone help me?


vinodhrajamuralidharan:
I am creating a ember-cli addon and like to include the css/less for that addon (don't know where and how I have to add the stylesheet). and it has to be injected on the ember-cli application where I use... Can anyone help me?

zeppelin:
Um, this is clearly a question suitable for asking on StackOverflow, could you please ask these over there first? Or even try googling before: http://hashrocket.com/blog/posts/building-ember-addons ;)

vinodhrajamuralidharan:
I checked the stack overflow and the blog you sent as well... But could not find the solution, that's why I posted the query here...

stefanpenner:
@bcardarella you were instrumental in designing this, can you share some thoughts?

stefanpenner:
@vinodhrajamuralidharan just putting in in app/styles or vendor/styles should be sufficient

stefanpenner:
anyways, please open a SO post and cross link here. As SO is a better venue for this sort of QA

bcardarella:
@vinodhrajamuralidharan you should be able to put your styles into app/styles in your addon and they'll be scooped up into your app's css file. It's currently "all or nothing".

vinodhrajamuralidharan:
As you suggested I tried like app/styles in my addon but its not merging with the existing css which is available in the ember-cli app where I am going to use the addon

bcardarella:
That's a bug then

deepan83:
Workaround would be to place the file within the vendor directory and import within the included hook?

raytiley:
I looked at the code today (need to start understanding how addons work). And where the addon builder looks for styles is in `addon/styles` not `app/styles`. @stefanpenner @bcardarella  does this sound right? I can do a PR to change it to `app` if that is deemed correct.

bcardarella:
@raytiley it should also pull styles from `app/styles` both should work

rwwagner90:
I am using ember-cli-less and I tried including all my styles in addon/styles and app/styles and neither seemed to work. @bcardarella @stefanpenner any ideas? Do I need to do anything special or should it be included automatically? My app.less in the dummy app is still empty after running ember serve.

stefanpenner:
> @bcardarella @stefanpenner any ideas?

No need to @ mention me, I review every issue on these repos as time permits.

bartocc:
[...]

PR #3088, opened 2015-01-24
[BUGFIX] - addons include app css

raytiley:
Fix for #3073. Still needs a test. @bcardarella mind spot checking this? I'll get to writing the tests later.


bcardarella:
This is incorrect as it will compile the styles from `app/styles` in an addon into `vendor.css`. We want styles from `app/styles` to compile into the application's stylesheet instead.

PR #3094, opened 2015-01-25
Make linting pluggable

ef4:
This adds a new ~~`postprocessTree` hook type~~ `lintTree` hook that allows addons to do arbitrary linting. I've moved out out-of-the-box jshinting into ember-cli-qunit (because it generates qunit-specific output).

See corresponding changes to ember-cli-qunit: https://github.com/ember-cli/ember-cli-qunit/pull/28
And ember-cli-mocha: https://github.com/ef4/ember-cli-mocha/compare/switchfly:master...master


blueprints/app/files/package.json
```diff
@@ -27,7 +27,7 @@
     "ember-cli-dependency-checker": "0.0.7",
     "ember-cli-ic-ajax": "0.1.1",
     "ember-cli-inject-live-reload": "^1.3.0",
-    "ember-cli-qunit": "0.1.2",
+    "ember-cli-qunit": "git://github.com/ef4/ember-cli-qunit#hinting",
```
bcardarella: this this supposed to be "linting" branch?

rwjblue: No the branch seems correct: https://github.com/ef4/ember-cli-qunit/tree/hinting.  This will be updated to use a released version once https://github.com/ember-cli/ember-cli-qunit/pull/28 lands and is released.

ef4: Nope. But this line is only here so we can see everything run in CI. Before we merge this we'll want to release ember-cli-qunit.

ef4:
I found a problem. This doesn't work if you have no hinters configured, and it doesn't let hinters stack correctly.

ef4:
Ok, I addressed the above issue by switching to a dedicated `lintTree` hook. 

We can't reuse `postprocessTree` here because we need the linters to run in parallel, not in series.

ef4:
For reference, the ember-cli-qunit PR that this depends on is now https://github.com/ember-cli/ember-cli-qunit/pull/29

stefanpenner:
@ef4 awesome, this seems like its craving us to introduce addon <-> ember-cli dependencies.

From reading, both this and ember-cli-qunit need to release a new version in lock-step, is this correct or can this PR support legacy versions of ember-cli-qunit?

rwjblue:
@ef4 - Can you add the new hook to https://github.com/ember-cli/ember-cli/blob/master/ADDON_HOOKS.md?

rwjblue:
@stefanpenner - You are correct, when we remove embedded support, ember-cli-qunit (and ember-cli-mocha) will need to be doing the work.  But ember-cli-qunit can be released before ember-cli (since it is just going to add a hook that will not be used until ember-cli is updated).

I suppose we could also potentially do some hackery to confirm that the version of ember-cli-qunit that is being used has a `lintTree` method, and if not then use the legacy path with a warning. I'm not convinced we need that much hoop jumping though (as an ember-cli upgrade should prompt users to update ember-cli-qunit anyways).

stefanpenner:
> I suppose we could also potentially do some hackery to confirm that the version of ember-cli-qunit t

ya, i think (outside of this PR) we should expand our addon improvements to include version checks.

rwjblue:
Just reviewed with @stefanpenner, and we are both :+1: on this.

A few bits of cleanup will be needed before merging:

- [ ] Add to ADDON_HOOKS.md
- [ ] Release ember-cli-qunit

ef4:
I added docs. @rwjblue, it looks like you can release ember-cli-qunit.

rwjblue:
Bumped ember-cli-qunit to 0.2.0 and submitted https://github.com/ember-cli/ember-cli/pull/3119 to pull it in.  Will merge that PR once green, then you can rebase this one and we can get it all out the door....

chadhietala:
@ef4 #3119 is now merged.

rwjblue:
Rebased and re-submitted in #3122.

topaxi:
I tried to switch to ember-cli-eslint, it seems work pretty well. But as I'm using async/await now, I get jshint errors all over the place, is there a way to disable jshint?

Relevant: https://github.com/jonathanKingston/ember-cli-eslint/issues/7

rwjblue:
Yes, I am working on making ember-cli-qunit / ember-cli-mocha use pluggable linter (I want to use eslint for the same reasons).

taylon:
Hey @rwjblue what is the status of it?

I tried to figure out looking at the many PRs related to it and the ember-cli documentation but couldn't figure it out for sure.

Is this still working in progress?

PR #3101, opened 2015-01-26
No need for rimraf or mkdirp if fs-extra is already a dependency

chadhietala:


rwjblue:
@chadhietala: Can you double check if fs-extras `remove` has the same retry semantics as `rimraf` on windows?  We are relying on that for better Windows support and I just want to make sure it does have the same properties before removing `rimraf`. /cc @raytiley

rwjblue:
Heh, just did a bit of digging on this myself, and it appears that `fs-extra` is just using `rimraf` for `fs.remove`, so we should be fine.  

https://github.com/jprichardson/node-fs-extra/blob/master/lib/remove.js

stefanpenner:
restarting appveyor

PR #3102, opened 2015-01-26
Updated tiny-lr

johanneswuerbach:
tiny-lr uses livereload.js now directly from https://github.com/livereload/livereload-js instead of maintaining (an outdated) copy of it

All changes https://github.com/mklabs/tiny-lr/compare/v0.1.4...v0.1.5


rwjblue:
Thanks!

PR #3042, opened 2015-01-20
Create editors section

BryanHunt:
The information on Sublime Text was moved to this new section and information on Atom was added.

PR #3110, opened 2015-01-26
[DOCS] Generalize Sublime Text issues to both versions 2 and 3

cbrock:
Using Sublime Text 2, I experienced significantly slower build times before realizing that the same issue (already described specifically for Sublime Text 3) was the cause. The same solution - excluding the tmp directory - also resolved my issue.

I think it would be helpful to point out that both Sublime Text 2 and Sublime Text 3 can potentially cause issues.

*Note that these changes had been previously merged with https://github.com/ember-cli/ember-cli/pull/2385, but look to have since been overwritten with https://github.com/ember-cli/ember-cli/pull/3042*

### DIFF
--- a/_posts/2013-04-02-editors.md
+++ b/_posts/2013-04-02-editors.md
@@ -15,11 +15,11 @@ Search for [ember-cli-helper](https://atom.io/packages/ember-cli-helper)
 
 `Install`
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks!

Issue #1451, opened 2014-07-23
Ember Test: No Tests Are Ran


agentVargas2012R:
I searched the issues and did not find anything matched this scenario exactly so I am creating an issues. 

**Background**
C:\opt\workspaces\CallCenter>ember test 
version: 0.0.39       

**From the docs**
ember test
Run tests with Testem on CI mode. You can pass any options to Testem through testem.json, by default we’ll search for it under your project’s root or you can specify config-file.

**Built project successfully. Stored in "C:\opt\workspaces\yomoma\tmp\class-tests_dist-qZh4xARd.tmp".**

1..0
 tests 0
 pass  0
 fail  0

ok

**directory structure:**
tests/*_/_-test.js

Is there a manual step that needs to be done in order to register the tests to execute inside the test folder?

ccarterc:
Fixed it by adding:

`C:\Users\blessenm\AppData\Roaming\npm\node_modules\phantomjs\lib\phantom` to path variable.  

Do NOT add:

`C:\Users\blessenm\AppData\Roaming\npm\node_modules\phantomjs\lib\phantom\phantomjs.exe`

Also make sure to replace `blessenm` with your appropriate user name.

KurtRMueller:
Hmmm.... I'm running into this same problem. I installed phantomjs via the ` npm install -g phantomjs` command. I get an error where 0 tests are ran.

I'm on a Mac however. Should I open up a new issue?

vilhelmjosander:
Same problem here on mac. Terminal is just blank after building application, then when pressing ctrl+C I just get the following:

1..1
# tests 1
# pass  0
# fail  1

stefanpenner:
@KurtRMueller @vilhelmjosander if you open a new issue be sure to give actionable info

* what version of node / npm / ember-cli / os / phantom
* does it work with a vanilla app etc.

KurtRMueller:
Some of that info I gave here: https://github.com/ember-cli/ember-cli/issues/3977#issuecomment-97285168

My current version of phantom js is listed below. Of course, after `brew upgrade`ing and the like, it might be different than the version I attempted to run the tests with. Next time I will try to include more information.

```
phantomjs -v                                                                                                                                                                                     127 ↵
1.9.8
```

I am running OS/X Yosemite.

steffenbrem:
I have the same as @vilhelmjosander described. I can't test my ember apps on my CI server because of this problem.

stefanpenner:
[...]

PR #3115, opened 2015-01-27
Document common windows PhantomJS issue

corsen2000:
I ran into #1451 when using ember-cli on windows.  I was able to get it working by following the suggestions on the issue.  Since the documentation has a "common issues" section I thought it might be useful to add something about this.


Dhaulagiri:
Nice!  This one has bitten me before and I wasn't sure how to fix.

PR #3116, opened 2015-01-27
[Bugfix] files named app/styles... or app/templates... should also be

stefanpenner:
included

- [x] tests
- [x] ~~confirm the use of `path.sep` is what windows wants~~ (it is not).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -384,8 +384,10 @@ EmberApp.prototype._filterAppTree = function() {
 
   var podPatterns = this._podTemplatePatterns();
   var excludePatterns = podPatterns.concat([
-    /^styles/,
-    /^templates/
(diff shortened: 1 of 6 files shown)


rwjblue:
@stefanpenner - Updated with tests, and an inline comment on why we are not using `path.sep`.

lib/broccoli/ember-app.js
```diff
@@ -384,8 +384,10 @@ EmberApp.prototype._filterAppTree = function() {
 
   var podPatterns = this._podTemplatePatterns();
   var excludePatterns = podPatterns.concat([
-    /^styles/,
-    /^templates/
+    // note: do not use path.sep here Funnel uses
+    // walk-sync which always joins with `/` (not path.sep)
```
stefanpenner: perfect

stefanpenner:
LGTM, lets see what CI says

Issue #3103, opened 2015-01-26
[Feature Request] Defeatureifying application code


jkarsrud:
I have been trying to build an [addon](https://github.com/jkarsrud/ember-cli-defeatureify) that gives users support for [defeatureify](https://github.com/thomasboyt/defeatureify)ing their application code using `broccoli-defeatureify`. During development of the addon I met some difficult problems that I have discussed partly with @rwjblue and @rondale-sc on IRC.

The issues I met during development were mostly related to when `broccoli-defeatureify` should run in the build;  it needs to run *after* es6-transpilation, but *before* minification in order to prevent defeatureify to "freak out" about the minified code. Right now there really isn't a good way to do this from an addon, as there aren't any hooks that allows for this to happen. One solution that @rondale-sc and I discussed on IRC is to somehow "halt" minifcation by storing the minifyJS options, and running the minifcation manually afterwards, but to my that seems like an unwanted hack. I'd prefer if the addon didn't interfere with the rest of the build pipeline at all, unless it's run from a specific hook. The other issue with running the addon in its current state is speed, as I haven't been able to make it run on other than `'all'` in the `postprocessTree`-hook.

As there is no hook in the build pipeline where we can safely run defeatureify without disabling/halting other steps, I was wondering if perhaps this should be included in the CLI itself, as that would enable full control of when this step happens. The defeatureify step would of course only happen when running a production build, and would need to support embedding the application feature flags like [emberjs-build](https://github.com/emberjs/emberjs-build) does it for dev builds. 

Does this sound like too much of an edge-case to be included in the CLI itself, or do you think this is something that more people would find useful? Perhaps just exposing another hook would be enough for this to be useful as an addon? I'm open for discussions, and can probably submit a PR for this to the build process if it's actually considered useful for a lot of people.

rwjblue:
https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L777-783 needs to be made into an addon that uses `postprocessTree`.  Then your addon could simply specify that it needs to be run before the uglify addon (via normal addon ordering mechanism).

jkarsrud:
@rwjblue I've made [ember-cli-uglify](https://github.com/jkarsrud/ember-cli-uglify) now, and removed the minification steps from ember-cli. I guess the next step is to add the addon to the app blueprint and submit a PR?

rwjblue:
@jkarsrud - Yep.

rondale-sc:
[...]

PR #3117, opened 2015-01-27
Replace builtin minification with ember-cli-uglify

jkarsrud:
This replaces the builtin broccoli-uglify-sourcemap minification with
ember-cli-uglify. This allows for other addons to run before the
minification step, which was not possible earlier.

This is the result of a discussion in #3103 about when to run certain types of addons.


jkarsrud:
Hmm, I'm not sure if this has done something to fingerprinting or not, but that test seems to fail... Any ideas?

rwjblue:
@jkarsrud - fingerprinting needs to be done after minification (it is happening too early and the fingerprint does not match the files hash).  I opened an issue upstream with more details: https://github.com/jkarsrud/ember-cli-uglify/issues/1.

jkarsrud:
@rwjblue Thanks for the feedback! I've updated the package and all tests are good locally, so just need CI to confirm :)

rwjblue:
Looks great!  I'm going to wait to land this until after 0.1.10 is shipped (hopefully later tonight), then pull it in.

rwjblue:
Nevermind, gonna pull this into 0.1.10 after all.

eljusto:
How to perform minification in my app?
Before update I've just built my project with `-e production` but now it doesn't work.
So what should I do to use ember-cli-uglify?

rwjblue:
@eljusto - All you need to do is add `ember-cli-uglify` to your project.  `npm install --save-dev ember-cli-uglify` should do the trick.

eljusto:
Oh, thanks a lot. 
I've read the changelog and thought that `ember-cli-uglify` has been installed with`ember init` command.

rwjblue:
@eljusto - Only if you accepted that change when reviewing the diffs (it is really easy to miss).

PR #3118, opened 2015-01-27
BUGFIX serve command host alias was conflicting with help alias

taddeimania:
any variant of -ho, -hos, and -host seem to work as well.  This is more of a documentation fix than anything.

### DIFF
--- a/tests/unit/commands/server-test.js
+++ b/tests/unit/commands/server-test.js
@@ -126,4 +126,16 @@ describe('server command', function() {
       expect(ops.baseURL).to.equal('test', 'Uses the correct environment.');
     });
   });
+
+  it('host alias does not conflict with help alias', function() {
+    return new ServeCommand(options).validateAndRun([
+      '-H', 'hostname'
+    ]).then(function() {
+      var serveRun = tasks.Serve.prototype.run;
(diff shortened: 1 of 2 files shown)


tests/unit/commands/server-test.js
```diff
@@ -126,4 +126,16 @@ describe('server command', function() {
       expect(ops.baseURL).to.equal('test', 'Uses the correct environment.');
     });
   });
+
+  it('host alias does not conflict with help alias', function() {
+    return new ServeCommand(options).validateAndRun([
+      '-ho', 'hostname'
```
stefanpenner: having a two letter short form seems funky, maybe -H ?

taddeimania: If that's the direction that aliases are going to go then I can absolutely do that.  I didn't notice any other commands with a capital letter alias and didn't want to break any unspoken convention.

stefanpenner: no convention that i know of, i think its alright.


@all thoughts?

taddeimania:
changed alias to -H to provide a non conflicting shortcut for --host

stefanpenner:
thanks :)

PR #3130, opened 2015-01-28
Touchup Tomster

johnnyshields:
No '  `  or " chars present


stefanpenner:
does it commit?

johnnyshields:
So--this works fine on my Windows machine as you can see in the PR commit. I'd appreciate if someone can double check it.

chadhietala:
Confirmed commit happens and Tomster is indeed intact.

PR #3145, opened 2015-01-29
Allow beforeInstall hook in the blueprint model to access and modify locals.

mattmarcum:
This is the only hook that can update locals asynchronously.  If you need to generate `locals` or `fileMapTokens` asynchronously right now, you can't.  Those hooks are called synchronously.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -393,7 +393,7 @@ Blueprint.prototype.install = function(options) {
   var locals = this._locals(options);
 
   return Promise.resolve()
-    .then(this.beforeInstall.bind(this, options))
+    .then(this.beforeInstall.bind(this, options, locals))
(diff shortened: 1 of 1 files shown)


jgwhite:
This looks good to me. Could you also update the documentation (both inline and on the website)?

jgwhite:
Example usage would be super valuable too.

I'm imagining you'd mutate `locals` in `beforeInstall`. We should probably add a test that the extra values definitely make it to the following hooks.

mattmarcum:
Ok, I'll take a look at those after work

stefanpenner:
@trabus yes :+1:  on tests

mattmarcum:
Didn't get to this last night, but def will this weekend

trabus:
No need to close this while you're working on tests, this PR should stay visible. Let me know if you need any assistance with the tests, I would be happy to help.

Issue #3142, opened 2015-01-28
Encountering error with globPattern when installing addons with a blueprint


johnotander:
When installing an ember addon with `ember install:addon addon-name`, I encounter the following error

```
The globPattern "undefined" did not match any files, so no file updates will be made.
```

Here's an example of the error in action:

```
ember install:addon ember-cli-gravatar          
version: 0.1.11
Installed packages for tooling via npm.
installing
The globPattern "undefined" did not match any files, so no file updates will be made.
Installing browser packages via Bower...
  cached git://github.com/blueimp/JavaScript-MD5.git#1.1.0
Installed browser packages via Bower.
Installed addon package.
```

It seems to be occurring when the generator is run. This is also reproduced with `npm i --save-dev addon-name && ember g addon-name`.

My environment:

```
ember -v
version: 0.1.11
node: 0.10.35
npm: 2.1.8
```

I've gone through installing a collection of addons to see which result in an error, and which don't. It appears that addons that have a blueprint for adding a bower dependency are the ones raising this error. What I'm not sure about is whether this is something that the addons are missing, or if it is a problem with the ember-cli. Whichever the case is, we can probably narrow it down and provide a more descriptive message. Anyone have any idea what's going on here?

I'll keep digging. Thanks for your time.

chadhietala:
Looking into this. Trying to find the root cause to avoid unnecessary branching in the code.

johnotander:
Great, thanks!

trabus:
The fix to prevent the warning was merged, thanks @chadhietala.

The ember-cli-gravatar addon blueprint doesn't install any files, it just adds a bower package. So that would explain why the addon install runs the blueprint without processing any files (the reason the warning was being thrown).

Mystery solved, this issue can be closed!

PR #3155, opened 2015-01-29
Guard against rawArgs being undefined

chadhietala:
I'm not sure if this is the right fix for #3142. I'm not even sure `rawArgs` is used for since I ran a bunch of generators to see what it what it's value was. It seemed to always be `undefined`. @rwjblue, @stefanpenner, or @bcardarella can you drop some knowledge bombs on this?

This makes the message go away but I'm not sure what is the root cause. Traced it all the way back to `_locals`.  It only effects bower.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -595,7 +595,7 @@ Blueprint.prototype.processFiles = function(intoDir, templateVariables) {
 
   var fileInfos = files.map(this.buildFileInfo.bind(this, destPath, templateVariables));
 
-  if (fileInfos.filter(isFilePath).length < 1) {
+  if (fileInfos.filter(isFilePath).length < 1 && templateVariables.rawArgs) {
     this.ui.writeLine(chalk.yellow('The globPattern \"' + templateVariables.rawArgs +
       '\" did not match any files, so no file updates will be made.'));
   }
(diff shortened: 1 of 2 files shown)


cibernox:
I faced this too. My main suspect what the `normalizeEntityName: function() {},` of a blueprints of an addon I was installing.

Does that makes any sense?

trabus:
I think this is related to some work on `ember init` that I did a while back. To allow a glob pattern to be used to define which files to init, a `targetFiles` array is passed in, in addition to the `rawArgs` [here](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/init.js#L68) so we can use it for the message you're seeing.

I'm not sure why an addon install is coming up with 0 files to process, but that would definitely be the reason the message comes up. Checking for the existence of `rawArgs` is probably the easiest path to fixing it, but I'd definitely like to understand why an addon install is behaving that way.

stefanpenner:
cc @jgwhite

jgwhite:
Pretty sure @trabus is right and this is the correct fix. Merging.

trabus:
I just figured out why the blueprint was being processed with no files: [it didn't have any](https://github.com/johnotander/ember-cli-gravatar/tree/master/blueprints/ember-cli-gravatar) to process. The blueprint simply installed a bower package.

PR #3156, opened 2015-01-30
Serve existing files from `/test` if they exist

### LINKED ISSUE #3150: Test Middleware: Incorrect Handling of Static Assets
In the [TestServerAddon](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/middleware/tests-server/index.js) we are not properly handling any static assets in `tests/` directory.  All requests are re-written to `/tests/index.html` regardless of whether they are for an existing file in the broccoli builds output directory.

A few things need to be updated to handle this properly:

- [ ] Only rewrite the request URL **if** the requested asset does not already exist. The HistorySupport middleware does this properly, and we can completely steal its implementation :).  [See here](https://github.com/ember-cli/ember-cli/blob/b509978e1c82b067a1547254b611723d14a15bf2/lib/tasks/server/middleware/history-support/index.js#L31-L32).
- [ ] Add tests to confirm that any files in `tests/` other than `index.html` can be retrieved.  A nice base test to start with is [here](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/server/express-server-test.js#L378-L398), and it needs to be tweaked to request a static asset in the output dir (like [this test](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/server/express-server-test.js#L450-L460)) and confirm its contents.

---

This definitely affects ember-cli apps that want to have other files in `tests/` be requested from their `tests/index.html`, but where this was noticed primarily is in third party libraries using ember-cli simply for building.  Specifically:

*  https://github.com/tildeio/htmlbars - This used to use `test/` to avoid the bizarre issues here (before they were understood), but this issue surfaced recently when `test/` was moved to `tests/`.  Reproduction: run `ember serve` in HTMLBars repo, and navigating to `localhost:4200/tests/index.html`. You will see that tests do not run properly (because other configuration files are loaded from `tests/`, but this middleware is rewriting these requests to `tests/index.html`).

trek:
Closes #3150

### DIFF
--- a/lib/tasks/server/middleware/tests-server/index.js
+++ b/lib/tasks/server/middleware/tests-server/index.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var cleanBaseURL = require('../../../../utilities/clean-base-url');
+var fs = require('fs');
+var path = require('path');
 
 function TestsServerAddon(project) {
   this.project = project;
@@ -14,10 +16,9 @@ TestsServerAddon.prototype.serverMiddleware = function(config) {
 
   var baseURL = cleanBaseURL(options.baseURL);
   var testsRegexp = new RegExp('^' + baseURL + 'tests');
-  var baseURLRegexp = new RegExp('^' + baseURL);
 
   app.use(function(req, res, next) {
-    watcher.then(function() {
+    watcher.then(function(results) {
 
       var acceptHeaders = req.headers.accept || [];
       var hasHTMLHeader = acceptHeaders.indexOf('text/html') !== -1;
@@ -26,10 +27,11 @@ TestsServerAddon.prototype.serverMiddleware = function(config) {
       var isForTests = testsRegexp.test(req.path);
 
       if (isForTests && (hasHTMLHeader || hasWildcardHeader) && req.method === 'GET') {
-        if (baseURLRegexp.test(req.path)) {
-          req.url = baseURL + 'tests/index.html';
-        } else {
-          req.url = '/tests/index.html';
+        var assetPath = req.path.slice(baseURL.length);
+        var filePath = path.join(results.directory, assetPath);
+
+        if(!fs.existsSync(filePath) || !fs.lstatSync(filePath).isFile()) {
+          req.url = baseURL + '/tests/index.html';
         }
       }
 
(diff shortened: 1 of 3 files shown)


lib/tasks/server/middleware/tests-server/index.js
```diff
@@ -14,10 +16,9 @@ TestsServerAddon.prototype.serverMiddleware = function(config) {
 
   var baseURL = cleanBaseURL(options.baseURL);
```
trek: baseUrl is either `/test` or `/`, so I got rid of checking whether it matched `test`. This code path was never tested anyway (I verified by throwing inside the `else`.

rwjblue:
:+1:

tests/unit/tasks/server/express-server-test.js
```diff
@@ -414,6 +414,41 @@ describe('express-server', function() {
...
+      });
+
+      it('GET /tests/an-existing-file.tla serves tests/an-existing-file.tla if it is found', function(done) {
+        return startServer()
+          .then(function() {
+            request(subject.app)
+              .get('/tests/test-file.txt')
+              .set('accept', 'text/html')
+              .expect(200)
+              .expect('some contents\n')
```
rwjblue: This needs to use `path.sep`.  AppVeyor failed with the following message:

```
1) express-server behaviour without proxy GET /tests/an-existing-file.tla serves tests/an-existing-file.tla if it is found:
     Error: expected 'some contents\n' response body, got 'some contents\r\n'
      at error (C:\projects\ember-cli\node_modules\supertest\lib\test.js:235:13)
      at Test.assert (C:\projects\ember-cli\node_modules\supertest\lib\test.js:180:21)
      at Server.assert (C:\projects\ember-cli\node_modules\supertest\lib\test.js:132:12)
      at Server.g (events.js:180:16)
      at Server.emit (events.js:92:17)
      at net.js:1276:10
      at process._tickCallback (node.js:419:13)
```

trek: I'll just remove the newline

rwjblue:
Will fix the underlying problem in https://github.com/tildeio/htmlbars/issues/272 when it is landed.

rwjblue:
Thank you!!!

seanpdoyle:
:clap:

PR #3160, opened 2015-01-30
[BUGFIX] Don't override the request's path

dmathieu:
I am writing an add-on which needs to get access to some URLs server-side (to perform OAuth authentication).
The thing is, the request's path is being rewritten to display the default file content.

Displaying the default file when nothing else can be rendered makes sense. But overriding the path and preventing middlewares from reading doesn't.

### DIFF
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -30,7 +30,7 @@ HistorySupportAddon.prototype.serverMiddleware = function(config) {
         if (hasHTMLHeader && isForBaseURL && req.method === 'GET') {
           var assetPath = req.path.slice(baseURL.length);
           if (!fs.existsSync(path.join(results.directory, assetPath))) {
-            req.url = baseURL + 'index.html';
+            req.serveUrl = baseURL + 'index.html';
           }
         }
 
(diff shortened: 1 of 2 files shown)


chadhietala:
Restarting Appveyor.

stefanpenner:
tests?

dmathieu:
@stefanpenner: I have nothing against writing tests. Unfortunately, after spending a bit of time on it, I feel like writing this test would need refactoring all of add-ons.
I intended to write something like this:

```
      it('allows addons to render on a path', function(done) {
        return subject.start({
          host:  '0.0.0.0',
          port: '1337',
          baseURL: '/'
        }).then(function() {
          request(subject.app)
            .get('/addon.foobar')
            .set('accept', 'text/html')
            .expect(200)
            .expect(function(res) {
              expect(res.text).to.equal('rendered by addon');
            })
            .end(function(err) {
              if (err) {
                return done(err);
              }
              done();
            });
        });
```

Unfortunately, at [line 561](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/server/express-server-test.js#L561), the add-ons test resets the initializeAddons method, which removes all built-in add-ons.
I believe this prevents from properly testing them, as a built-in add-on can introduce something breaking other ones further in the stack (which is the case here).

I took a quick look at the possibility of not resetting that method, but it looks like that would require quite a lot of setup.

Where/how would you add this test?

dmathieu:
Thank you @stefanpenner

stefanpenner:
Thank you

PR #3164, opened 2015-01-31
[ENHANCEMENT] Enable asynchronous addon#serverMiddleware hooks 

taras:
This PR makes startHttpServer process addon & app middlewares asynchronously. 

Currently, History Support middleware clobbers any serverMiddleware that has to perform asynchronous operations before applying the middleware. This code makes it possible to return a promise from a serverMiddleware hook. processAddonMiddlewares will wait for promise to be resolved before moving onto the next addon.

**Tests**

- [x] That the addons promises are processed in the proper order (should work perfect due to mapSeries, but I'd like a test for it)
- [x] If the addon does not return a promise (all works well)
- [x] If the addon returns a promise that rejects/throws (error is not swallowed, but is printed or bubbled up)


rwjblue:
I'm generally :+1:, but we need tests for this. To check the following scenarios:

* That the addons promises are processed in the proper order (should work perfect due to `mapSeries`, but I'd like a test for it)
* If the addon does not return a promise (all works well)
* If the addon returns a promise that rejects/throws (error is not swallowed, but is printed or bubbled up)

lib/tasks/server/express-server.js
```diff
@@ -58,9 +59,9 @@ module.exports = Task.extend({
 
   processAddonMiddlewares: function(options) {
     this.project.initializeAddons();
-    this.project.addons.forEach(function(addon) {
+    mapSeries(this.project.addons, function(addon) {
```
rwjblue: Needs to be returned

lib/tasks/server/express-server.js
```diff
@@ -161,13 +162,19 @@ module.exports = Task.extend({
...
-    this.processAppMiddlewares(options);
-    this.processAddonMiddlewares(options);
-
-    return this.listen(options.port, options.host)
-      .catch(function() {
-        throw new SilentError('Could not serve on http://' + options.host + ':' + options.port + '. It is either in use or you do not have permission.');
-      });
+    return Promise.resolve()
+      .then(function(){
+        this.processAppMiddlewares(options);
```
rwjblue: needs to return (so promises are properly chained)

taras:
@rwjblue thank you for the test suggestions, that's exactly what I was looking for :) I'm on it

lib/tasks/server/express-server.js
```diff
@@ -161,13 +162,19 @@ module.exports = Task.extend({
...
+      }.bind(this))
+      .then(function(){
+        this.processAddonMiddlewares(options);
```
rwjblue: needs to return (so promises are properly chained)

rwjblue:
We should also allow in-app middlewares to be async (so the two API's are still in sync).

taras:
Return statements added, now the tests.

taras:
> We should also allow in-app middlewares to be async (so the two API's are still in sync).

Should we add return statement and call mocks & proxies with mapSeries?

rwjblue:
@taras - We should update `processAppMiddlewares` with the same `mapSeries` logic as `processAddonMiddlewares`, but we do not need to change the blueprints (returning is not required at all, its just nice to have the ability).

taras:
I already added a return statement to processAppMiddlewares. `processAppMiddlewares` only has 1 call, so it doesn't need a mapSeries. It should work as it is now.

rwjblue:
@taras - Yep, makes sense (didn't really think about the fact that there was only one `server/index.js` so we didn't need a `mapSeries`).

taras:
@rwjblue added the tests, squashed and it's green :)

taras:
I made a minor change to test descriptions.

taras:
@rwjblue ping

rwjblue:
Thank you!

Issue #2652, opened 2014-12-03
how should we handle config.usePodsByDefault when pods are enabled by default?


EndangeredMassa:
When making pods the default project structure in ember-cli, how should we handle the config option `usePodsByDefault`? We have a couple of options.

1. switch to a different config option like `disablePodsByDefault`; (I don't know what to call the opposite of pods.)
2. just change the app blueprint to set `usePodsByDefault: true` 
3. deprecate the flag and assume pods going forward

I think that Option 2 is fine, but would love to hear what other people think about it.

---

/cc @trabus @stefanpenner

stefanpenner:
better question, why is the podConfig in the app/config ? seems like a build tool concern.

cc @rwjblue

trabus:
@stefanpenner This was my fault. I misunderstood something when I was adding it and actually wasn't aware of .ember-cli

stefanpenner:
> @stefanpenner This was my fault. I misunderstood something when I was adding it and actually wasn't aware of .ember-cli

got time to deprecate and move? ;)

stefanpenner:
I'm trying to re-start the pod fleshout effort this weekend.

example project -> https://github.com/stefanpenner/ember-jobs

trabus:
Awesome. Yeah, I'll take a crack at it, I'm at a resort with my family on a Girl Scouts outing, but they're all asleep so I'll see how far I can get tonight. ;)

trabus:
Looking for an example of how to show a deprecation warning in ember-cli. [This](https://github.com/ember-cli/ember-cli/blob/2935d10f6e12e99906ed5d27e0bdf883555219ab/lib/models/addon.js#L550) isn't working for some reason (inside the blueprint model).

trabus:
Going with `usePods` to replace `usePodsByDefault`.  I had to do some tweaks to get the environment settings to be available to the blueprint, but I have it working. It should probably be easier to get `.ember-cli` properties inside ember-cli, but passing in the settings through the command and task is working for now.

trabus:
Just a note, this issue isn't resolved with moving the config property to `.ember-cli`. We still need to derive some sort of antonym for `--pod` that people can use if they want to generate (or destroy) modules in the type based format.

This actually goes back to the initial addition of pod support, where it was proposed as a `--structure=pod` flag. We could refactor to use that instead, using shorthands for the generate and destroy commands to set the flag. Generate available options would look like:
```
availableOptions: [
[...]

PR #3167, opened 2015-01-31
`usePodsByDefault` in app config deprecated in favor of `usePods` in .ember-cli

trabus:
As per #2652
`usePodsByDefault` in the `config/environment.js` file is now deprecated in favor of `usePods` in `.ember-cli`
Added environment settings to generate and destroy commands and tasks to make the `usePods` property available.

- [x] update documentation
- [x] add tests

### DIFF
--- a/lib/commands/destroy.js
+++ b/lib/commands/destroy.js
@@ -42,7 +42,8 @@ module.exports = Command.extend({
     var task = new Task({
       ui: this.ui,
       analytics: this.analytics,
-      project: this.project
+      project: this.project,
+      settings: this.settings
     });
 
     var taskArgs = {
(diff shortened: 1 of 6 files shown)


stefanpenner:
:+1: 

@rwjblue r?

trabus:
I'm unavailable this morning, but should be able to add some tests and documentation changes this afternoon.

trabus:
I've temporarily stalled on this after investigating how to properly fix the flag. This change could still go out without fixing the way the flag works, and is probably the best route, I just wanted to get my thoughts out on structures before proceeding further on this. Will likely resume on adding tests later this week.

trabus:
Added acceptance tests for `usePods` flag, should I make one for the deprecation warning?

PR #3170, opened 2015-02-01
[ENHANCEMENT] Update ember-qunit to 0.2.0.

rwjblue:
The update to ember-qunit 0.2.0 has a few caveats but is largely compatible with 0.1.x.  A few of the main changes are:

* Rewritten to be a thin wrapper around ember-test-helpers, which are now shared between ember-qunit and ember-mocha.
* Deprecated `this.append()` in favor of `this.render()` in `moduleForComponent`.
* `setup` callback no longer receives the `container` as its first argument. This requires a largely mechanical refactor to use `this.container` in the callback (instead of using the first argument).

Users that update ember-cli-qunit, but not ember-qunit will see no changes (ember-cli-qunit itself is compatible with both versions).


chadhietala:
:+1:

abuiles:
Thanks!

Issue #2959, opened 2015-01-12
How to use multiple CSS preprocessors in one project?


tsabolov:
I prefer Stylus by myself to write styles. Also I use Foundation which is written in SCSS, and Selectize whose styles are in LESS. What is the correct way (if any) of using multiple preprocessors together?

stefanpenner:
specific preprocessors should detect based on the file extension

tsabolov:
But how different files should be compiled into one app.css? Maybe I missed something in docs?

stefanpenner:
hmm, i suspect they would be kept as independent trees.

pzuraq:
@stefanpenner I just ran a test case and it looks like the trees aren't merging properly. I think the expected behavior should be including both an `app.[scss/sass]` file and an `app.styl` file as entry points for the respective preprocessors (and allowing the outputPath to be specified in config as usual). Currently it fails if an app has both preprocessors and both files, I'm going to dig in and figure out why.

tsabolov:
Am I getting it right that by design both `app.[scss/sass]` and `app.styl` should be merged into one output file `app.css`? If it's the case then in what order merge will be done?

pzuraq:
@tsabolov Yes they should be merged into one file, but I was not thinking about order. Ideally you should be able to specify the order that they are merged in, as well as names of the input files.

knownasilya:
Was this ever resolved? Related issues https://github.com/gdub22/ember-cli-less/issues/27 and https://github.com/jeffjewiss/ember-cli-postcss/issues/4

msalahz:
Any updates about this ?

webark:
@pzuraq any luck with this..?

pzuraq:
@webark If I remember correctly I had a solution working, but the core issue I came up against was if you had multiple style files with the same name and different preprocessors (e.g. `app.scss` and `app.styl`). The respective broccoli plugins would both produce an `app.css` which would cause a conflict and throw an error, so this kind of became a question of what to do in that scenario.

A lot has changed since then, I'll try to take a look at it sometime but I haven't been working on CLI stuff for a while.

knownasilya:
Postcss might be a good option, since it has plugins for sass and less. Not sure if those work together though.

PR #3171, opened 2015-02-01
WIP Fixes behavior for apps with multiple style preprocessors

pzuraq:
Apps that have multiple preprocessors (such as the one mentioned in #2959). This PR currently allows for multiple preprocessors to output to separate files by customizing an app's outputPaths. This allows users to include the output in their app in whatever order they choose, solving dependency issues.

`mergeTrees()` will throw an error if an app contains two or more files with the same name under different preprocessors (e.g. `app.scss` and `app.styl`). I am thinking that default behavior should be to concatenate same name files in arbitrary order (whatever order they were registered in the registry). This will require a custom version of broccoli-merge-trees, or a broccoli plugin that provides this functionality. Thoughts?

### DIFF
--- a/lib/preprocessors.js
+++ b/lib/preprocessors.js
@@ -3,6 +3,8 @@
 var path         = require('path');
 var Registry     = require('./preprocessors/registry');
 var requireLocal = require('./utilities/require-local');
+var Funnel       = require('broccoli-funnel');
+var mergeTrees   = require('broccoli-merge-trees');
 
 module.exports.setupRegistry = function(app) {
   var registry = new Registry(app.dependencies(), app);
@@ -50,7 +52,6 @@ module.exports.preprocessCss = function(tree, inputPath, outputPath, options) {
   var plugins = options.registry.load('css');
 
   if (plugins.length === 0) {
-    var Funnel = require('broccoli-funnel');
 
     return new Funnel(tree, {
       srcDir: inputPath,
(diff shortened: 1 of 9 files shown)


pzuraq:
Missed the fact that javascript plugins process sequentially, passing the output of the previous plugin as input into the next one. Javascript plugins use n-to-n semantics, whereas style plugins use n-to-one semantics. I think it should be ok for javascript plugins to process on separate trees rather than sequentially, though the files for the plugin should be picked.

pzuraq:
Decided to decouple style and template/js preprocessing, as both js and template preprocessors may chain. The current pattern used by broccoli-*-single plugins doesn't allow for chaining.

stefanpenner:
@krisselden would like your eye on this

pzuraq:
@krisselden Any word on this?

krisselden:
You had the WIP label on it and I've been busy, I just look it over it looks good, though I'd really like to rethink StylePlugins altogether and just have a way of adding input trees to the final css packaging step but that is a larger refactor. @rwjblue

pzuraq:
Yeah, the WIP is for the last consideration which is default behavior of stylesheets with different preprocessors but same names. I think they should append to one final output file, but I would have to write a new broccoli plugin to get that functionality.

stefanpenner:
@pzuraq can we get a rebase?

stefanpenner:
@rwjblue r?

pzuraq:
Having trouble with the tests on my machine, will get this pushed soon.

stefanpenner:
> Having trouble with the tests on my machine, will get this pushed soon.

:+1:

pzuraq:
Going to close this PR for now and reopen when I am able to fix my tests and rebase. See #3524.

PR #3182, opened 2015-02-02
Update ember-router-generator (single quote all the things)

abuiles:
Now we can specify quote or single quote for strings, default back to
single quotes.


rwjblue:
Awesome!

PR #3154, opened 2015-01-29
[BUGFIX] Replaces instances of dasherizedPackageName with relative paths. 

stavarotti:
Replaces instances of dasherizedPackageName with relative paths for the following generators:

- helper-test
- initializer-test
- mixin-test
- util-test

This closes #3143.

Issue #3143, opened 2015-01-29
Addons - Failing test for newly created initializer


stavarotti:
A newly created initializer test fails, with the following message:

```
Uncaught Error: Could not find module my-button/initializers/awesome-init

Uncaught Error: 
Could not find module my-button/initializers/awesome-initloader.js:58 
requireModuleloader.js:41 
reifyloader.js:69 
requireModuletest-loader.js:14 (anonymous function)
```
This is after running:

```
ember g initializer awesome-init
```
in a newly created addon and then navigating to localhost:4200/tests

Environment details:

```
ember-cli - 0.1.5
node - 0.10.36
npm - 1.4.28
Windows 7 64 Bit
```

package.json:

```
{
  "name": "my-botton",
  "version": "0.0.3",
  "description": "A nice addon button.",
  "directories": {
    "doc": "doc",
    "test": "tests"
  },
  "scripts": {
    "start": "ember server",
    "build": "ember build",
    "test": "ember test"
  },
  "repository": "",
  "engines": {
    "node": ">= 0.10.0"
  },
  "author": "",
  "license": "MIT",
  "dependencies": {
    "broccoli-merge-trees": "^0.2.1",
    "broccoli-static-compiler": "^0.2.1"
  },
  "devDependencies": {
    "broccoli-asset-rev": "^2.0.0",
    "broccoli-ember-hbs-template-compiler": "^1.6.1",
    "ember-cli": "0.1.5",
    "ember-cli-6to5": "0.2.1",
    "ember-cli-content-security-policy": "0.3.0",
    "ember-cli-dependency-checker": "0.0.7",
    "ember-cli-ic-ajax": "0.1.1",
    "ember-cli-inject-live-reload": "^1.3.0",
    "ember-cli-qunit": "0.1.2",
    "ember-data": "1.0.0-beta.12",
    "ember-export-application-global": "^1.0.0",
    "express": "^4.8.5",
    "glob": "^4.0.5"
  },
  "keywords": [
    "ember-addon"
  ],
  "ember-addon": {
    "configPath": "tests/dummy/config"
  }
}
```

jakecraige:
There is an open issue by me about generators not working in addons. This is that

jakecraige:
If you create it manually and you still have problems feel free to reopen this though

stavarotti:
Is this specific to initializers?  Components and helpers seems to work fine.

jakecraige:
It may happen with others, here's what's happening.

[...]

PR #3186, opened 2015-02-03
allow generated acceptance tests to be in directories

koriroys:
An example of what currently doesn't work:

    ember g acceptance-test projects/create

Opening this to start a discussion. Probably you guys know a better solution.

### DIFF
--- a/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import startApp from '../helpers/start-app';
+import startApp from '<%= dasherizedPackageName %>/tests/helpers/start-app';
 
 var application;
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
sounds good, but you'll also need to update the tests

koriroys:
@stefanpenner :heavy_check_mark:

jakecraige:
Some other tests were just recently changed to be relative for issues with generating in addons - #3154 . This doesn't reverse any of those changes, but it seems like it may cause the same issues reported in #3143 I'd want to make sure this isn't the case before merging. Pretty much, acceptance test generation should work within an addon

stefanpenner:
@jakecraige ok, this is waiting you +1 then. Merge when you feel good, or leave a comment detailing why not :)

koriroys:
@jakecraige do/can addons generate acceptance tests? I admit I'm not familiar with addons or how they work yet.

stefanpenner:
i rebased and made some formatting changes

PR #3188, opened 2015-02-03
Bump ember-data

abuiles:


rwjblue:
Need to bump bower.json value also.

stefanpenner:
> Need to bump bower.json value also.

cries the tears of 1000 rivers

abuiles:
@rwjblue fixed, @stefanpenner +1 :cry:

Issue #3146, opened 2015-01-29
Can't generate `locals` or `fileMapTokens` asynchronously in current blueprints


mattmarcum:
I'm trying to create a complex blueprint for an addon, and some of the lookups it needs to do for both the `locals` and `fileMapTokens` must be done asynchronously (via a promise).  Unfortunately both these hooks are run synchronously and the `beforeInstall` hook doesn't have access to the `locals` var in order to modify it.  

I create a pr that just adds the `locals` var as a second param to the beforeInstall hook.
https://github.com/ember-cli/ember-cli/pull/3199

trabus:
Closing, as #3199 was merged.

PR #3199, opened 2015-02-04
added locals to beforeInstall/beforeUninstall

mattmarcum:
updated the docs to document this, and added unit tests with updated basicBlueprint fixtures


mattmarcum:
This addresses https://github.com/ember-cli/ember-cli/issues/3146

at the time I created this master branch was not passing tests locally for me, but ci says it is...

jgwhite:
This is looking good. I’m a little hesitant about mutating `locals` in-place (rather than returning a new set) mainly because it’s a pattern we’re not using elsewhere. That said, it looks like the most practical solution in this case.

@trabus can we get your eyes on this too?

tests/unit/models/blueprint-test.js
```diff
@@ -225,7 +226,7 @@ describe('Blueprint', function() {
         .then(function() {
           var actualFiles = walkSync(tmpdir).sort();
           var output = ui.output.trim().split(EOL);
-
+
```
jgwhite: Bit of trailing whitespace here :P

mattmarcum:
I had started to create the pattern where locals were returned like this:

```javascript
return Promise.resolve()
    .then(function(){ 
      return this.locals(options); 
    }.bind(this))
    .then(function(customLocals){ 
      return this._locals(options, customLocals);
    }.bind(this))
    .then(function(locals){
      var promise = this.beforeInstall(options);
      if(promise && promise.then){
        return promise.then(function(){
          return locals;
        });
      }else{
        return locals;
      }
    }.bind(this))
    .then(function(locals){
      return this.processFiles(intoDir, locals);
    }.bind(this))
    .map(commit)
    .then(this.afterInstall.bind(this, options));
```
But this doesn't help if you want to do async work in fileMapTokens too.  The call to that function is several functions deep off of _locals.  It would take a major refactor to be able to handle a promise returned from fileMapTokens.  I think with the unit test here, we would know if `locals` ever gets removed from the scope of the before(Un)Install functions.

mattmarcum:
The smoke test failed because of a timeoute...they were passing earlier and then I removed some whitespace.  I bet if they are run again they will pass.

mattmarcum:
slow-tests are still timing out, but now on a different test

stefanpenner:
rebuilding...

trabus:
I think this is okay for now, but I like the idea of making the locals and fileMapTokens hooks promise aware as a longterm fix.

mattmarcum:
Does anything need to be changed on the pr before it can be merged to master?

jgwhite:
I’m ready to push the button on this. @mattmarcum could you work up a PR on `gh-pages` updating the docs for `beforeInstall`?

jgwhite:
Note to self: make sure pull requests are squashed before merge :)

Issue #3208, opened 2015-02-04
ember http-mock request.body is undefined


user9821:
not sure if this a bug or the api has changed.

just migrated to ember-cli 0.1.12 and now http-mock is not working for me.

Its working fine for get request . request.query returns all the query parameters.

However for POST request i cannot get the request parameters as i get request.body is undefined in the mocks defined in the server routes.

Can some one let me know how to access the request body in ember http-mock?
There is hardly any documentation i can find for this.

joostdevries:
@user9821 Did you regenerate the server? Because in #2568 the bodyparser (which gives you a POST/PUT body) has been removed. You'll have to add it to package.json and server/index.js again yourself.

PR #3211, opened 2015-02-05
Added inline docs regarding the removal of body-parser

joostdevries:
refs #3208 

Adds the following docblock to eg. `server/mocks/foo-bar`:

```javascript
/**
  The POST and PUT call will not contain a request body
  because the body-parser is not included by default.
  To use req.body, run:

     npm install --save-dev body-parser

  After installing, you need to `use` the body-parser for
  this mock by appending:

     app.use('/api/foo-bar', require('body-parser'));

*/  
```

### DIFF
--- a/blueprints/http-mock/files/server/mocks/__name__.js
+++ b/blueprints/http-mock/files/server/mocks/__name__.js
@@ -8,16 +8,29 @@ module.exports = function(app) {
     });
   });
 
-  <%= camelizedModuleName %>Router.post('/', function(req, res) {
-    res.status(201).end();
-  });
-
   <%= camelizedModuleName %>Router.get('/:id', function(req, res) {
     res.send({
       '<%= dasherizedModuleName %>': {
         id: req.params.id
       }
     });
+  });
+
+/**
+  The POST and PUT call will not contain a request body
+  because the body-parser is not included by default.
+  To use req.body, run:
+
+     npm install --save-dev body-parser
+
+  After installing, you need to `use` the body-parser for
+  this mock by appending:
+
+     app.use('/api/<%= decamelizedModuleName %>', require('body-parser'));
+
+*/   
+  <%= camelizedModuleName %>Router.post('/', function(req, res) {
+    res.status(201).end();
   });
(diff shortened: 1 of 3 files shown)


lessless:
@joostdevries in 0.1.15 there was a deprecation error ```body-parser deprecated bodyParser: use individual json/urlencoded middlewares node_modules/ember-cli/node_modules/express/lib/router/layer.js:82:5```

so it became  ```app.use('/api/users', require('body-parser').json(), usersRouter);```

joostdevries:
@lessless thanks!

Samsinite:
@joostdevries thanks. Ran across this today while mocking/stubing an api. Was this PR intentionally reverted without any intention to re-add it back to the blueprint? Seems like a pretty helpful comment to have.

joostdevries:
@Samsinite No! PR to re-add in the works now!

joostdevries:
#4235

Samsinite:
:+1: I was just happy to find the solution that you helped provide in this issue :).

ianwdunlop:
I added the code from @lessless but still cannot see the request body for a POST when saving a record. I can log the request in the server console and the js console in the browser claims that the JSON is being sent over but there doesn't seem to be a req.body available. For a GET request req.params works ok. Are there any good docs/resources for http-mock.

Samsinite:
@ianwdunlop this worked for me:
```js
module.exports = function(app) {
  var express = require('express');
  var paymentMethodsRouter = express.Router()

  ...

  paymentMethodsRouter.post('/', function(req, res) {
    var paymentMethod = createPaymentMethod(req.body.payment_method);

    __paymentMethods.push(paymentMethod);
    res.status(201).send({
      payment_method: paymentMethod
    });
  });

  ...

  app.use('/api/payment_methods', require('body-parser').json(), paymentMethodsRouter);
}
```

ianwdunlop:
Thanks for the help. Got it all working now. Would be good if this found its way in to the Ember/CLI docs (I know, it's open source, add it yourself etc ;) )

PR #3213, opened 2015-02-05
Revert "Added inline docs regarding the removal of body-parser"

joostdevries:
Did not realize this would generate an ugly diff on upgrade. Will figure out another way to document this.


stefanpenner:
LGTM

Issue #3215, opened 2015-02-06
Regression in vendorFiles: { 'something.js': null }


martndemus:
~~Before 0.1.12~~ A long time ago (before 0.0.38) this worked:

```javascript
// Brocfile.js
var app = new EmberApp({
  vendorFiles: {
    'jquery.js': {
       development: 'path/to/jquery.js',
       production:  null
    }
  }
})
```

But now it doesn't. This way it will now include the development jQuery. This caused a very subtle bug for us, because we now had two jQuery files in our production app. ('Mothership' app supplies jQuery for us)

I worked around it by pointing to an empty file for production, but would rather have this fixed!

This should work, because it then would be consistent with:
```javascript
var app = new EmberApp({
  vendorFiles: {
    'jquery.js': null
  }
})
```

martndemus:
Working on a fix, PR incoming soon.

PR #3216, opened 2015-02-06
Do not default to development asset

martndemus:
I removed default to development asset, this is not consistent with `app.import(null)`.
See #3215


stefanpenner:
this should like to and `in` check instead

martndemus:
Rebases, will add suggested check.

martndemus:
Added an update, which use in to check if current env is set.

(5 blueprint related tests fail, these also fail if I roll back to before my patches.)

rwjblue:
@martndemus - Thank you!

martndemus:
@rwjblue NP, this is way better than having to use an empty file.

Issue #694, opened 2014-05-12
karma + ember-cli


vire:
Hi, I'm struggling a bit with configuring the karma (version 0.12.16) and ember-cli (0.0.25).

Reason - I would like to leverage the test running/watching on karma. Now i get errors when I try to load the ./tests/test-helper/loader.js and the test-cases saying it can't resolve the import keyword (parse error)

**Question** is how to setup the karma.conf.js, especially  the files section.

My conf looks now like:

```
files: [
      'vendor/loader/loader.js',
      'vendor/jquery/jquery.min.js',
      'vendor/handlebars/handlebars.js',
      'vendor/ember/ember.js',
      'vendor/ember-resolver/dist/ember-resolver.js',
      'vendor/ember-load-initializers/ember-load-initializers.js',
      'tmp/output/assets/app.js',
      'tests/unit/routes/index-test.js'
    ],
```

amercier:
+1 @mattjmorrison as well.

Adding coverage to Karma is [very easy](https://github.com/karma-runner/karma-coverage). See [a configuration example](https://github.com/amercier/todomvc-testing/blob/master/tests/vanillajs/mocha/karma.conf.js).
Example of generated coverage report [here](http://amercier.github.io/todomvc-testing/tests/vanillajs/qunit/coverage/html/index.html).

corsen2000:
I'd also like to see karma support.  The primary reason being karma has built in selenium support to use browsers on a grid.  I have not found a way to do that with testem, but I imagine its possible since there is "edge" support for BrowserStack and Sauce.

Is there an open issue for making the test runner pluggable?

onmission:
+1 for Karma because of the it's reporters, So I can use karma test reports with TFS 2015 builds.

PR #3226, opened 2015-02-08
[ENHANCEMENT] Extract test command to plugin

mattjmorrison:
This is what I hope will be the first step to make the test runner pluggable as mentioned in Issue #694. 

This PR moves all of the test runner code into http://github.com/mattjmorrison/ember-cli-testem along with the appropriate tests. There are still a few acceptance tests that need some work though. I did publish ember-cli-testem to npm as well, for testing purposes.

I consider the ember-cli-testem repo that I created as being a temporary home for that code - since I did not write it I just moved it from ember-cli to an addon - that repo should probably live in the ember-cli organization. 

My hope is that by doing this we will have a clear path to create an addon for the karma test runner.


mattjmorrison:
It looks like there was an acceptance test that failed on TravisCI that didn't fail for me locally. I will look into that and see what's going on.

stefanpenner:
this is a pretty cool effort, but it seems to have gone stale. Is there something I can do to help move it forward?

mattjmorrison:
@stefanpenner I haven't spent much time on this the past few weeks. I am a bit stuck on getting this one last test to pass. I have a general idea of what the root of the problem is, and there are probably a few different discussions that need to happen once I have that issue resolved. I'm going to be spending some time working on this today, I will post an update this evening with any progress or lack thereof that I've made. Thanks for checking in!

mattjmorrison:
I think I finally made some progress on the failing test... which, for the record is `ember test --environment=production` in `smoke-test-slow.js`. Currently, when I run that test it will just hang, but removing the '--environment=production' will allow the test to run and pass. I believe this has to do with testem not being able to find `testem.json` (since when that file is missing the test will hang). The current home for `testem.json` is inside of `ember-cli-testem` (which is one of the topics that needs to be discussed - but for now in just getting everything to work it is inside of the addon).

Could the problem be something with this:
```javascript
    return pickFiles(testemPath, {
        files: ['testem.js'],
        srcDir: '/',
        destDir: '/'
      });
```
https://github.com/mattjmorrison/ember-cli-testem/blob/wip/index.js#L27

where it might do something different when running with `--environment=production` or something like that?  Any thoughts that might point me in the right direction?

johanneswuerbach:
The `testem.json` is something, people generally need to modify. I think it should be generated by a blueprint on install time (as today).

mattjmorrison:
My thinking was this: generally I never modify the `testem.json` file that currently ships with ember-cli, but I do realize that it is possible that others would want to modify it. By default, I was thinking that ember-cli-testem would contain the default `testem.json` file and if someone would want to modify it there would be a generator that they could run that would add the `testem.json` file to the root of their project. Otherwise, if someone does not need to modify the file from the default, they just wouldn't need it. I think the future ember-cli-karma addon would work the same way. No configuration needed by default, but a generator would create a file that you can modify if you choose to.

mattjmorrison:
Any suggestions as to why `--environment=production` would cause testem to hang in this test?

stefanpenner:
> Any suggestions as to why --environment=production would cause testem to hang in this test?
>  Hide all checks

something invalid in the compiled output?

stefanpenner:
I think we should likely transition ember-cli-testem to be part of the initial install, because that is currently what the community needs.

I suspect our blueprinting story is also slightly dificient, as it would be nice if this addon, could add something like the testem.json to the default `app` and `addon` blueprint. Thoughts @trabus @jgwhite ?

mattjmorrison:
I apologize in advance for the novel...

My thinking, based on my experience with ember-cli, is that the `testem.json` file rarely if ever needs to be modified. I have actually never modified `testem.json` for an ember-cli app, and based on asking around to a number of other Ember developers they have had the same experience. Based on that experience it seems unnecessary to have it in the root of every ember-cli project so why not just have the majority use case baked into the addon. Then the addon can have a generator for those rare situations when you do need to modify it for some reason.

My hope is that as soon as I can get this addon polished and working in a way that is acceptable to everyone, I will begin working on `ember-cli-karma`. There are a few specific things that are missing for me in testem which karma provides.

1. The ability to measure test coverage
2. [The ability to test different screen sizes](http://www.mattjmorrison.com/today-i-learned/2014/09/24/learned.html)

Both of those are possible (and fairly straightforward) using karma, so that is my motivation. 

My intention is that `ember-cli-testem` will work out of the box with any existing ember-cli app using the project's `testem.json` file. Then for new projects it will not be needed, until the app creator has a situation that they need to customize, in which case they can use a generator provided by the `ember-cli-testem` addon. `ember-cli-karma` will work the same way. There will be no need for a default `karma.conf.js` file until you need to make customizations, in which case a generator will create the default configuration file for you that can be modified.

tschoartschi:
@mattjmorrison Hey, any news on ember-cli-karma? I'm very interested on this topic. I have some knowledge about karma but not how to integrate it with ember-cli. Just let me know how far you are, maybe I could help you somehow ;)

mattjmorrison:
@tschoartschi That would be great. I haven't started on that yet. I wanted to get ember-cli-testem extracted from ember-cli first and hopefully arrive an an API that both ember-cli-testem and ember-cli-karam could implement so that switching between the 2 will be easy.

tschoartschi:
@mattjmorrison hey, any news on the topic "ember-cli-karma" ;-)

mattjmorrison:
@tschoartschi Nope, sorry. I'm still stuck on getting `ember-cli-testem` working seamlessly. I want to make sure that is solid before getting too gung-ho on `ember-cli-karam`.

stefanpenner:
I would love for this to land, but it appears to have gone stale.

Will gladly accept something that accomplishes this.

PR #3166, opened 2015-01-31
Addon discovery and isolation

lukemelia:
The goal of this PR is to address #2714 by moving to a model where addons are more isolated, and addons that have addons don't pollute the project with the dependent addons.

Feedback welcome.

- [x] extract discovery of addons from Project to it's own AddonDiscovery object
- [x] extract initialization of addons from Project to it's own AddonsFactory object
- [x] give each addon it's own AddonDiscovery instance, to look for child addons
- [x] give each addon it's own AddonsFactory instance, to instantiate any child addons
- [x] add acceptance test verifying addon isolation
- [x] add new addon hook: `setupPreprocessorRegistry(type, registry)`
- [x] move `registry` to `models/`
- [x] add more acceptance tests exercising nested plugins
- [x] implement support for nested addons contributing to built output, and handling for addons appearing more than once in the graph, including two different versions of the same addon (*keeping current semantics of one addon per build output, and last added "wins"*)
- [x] unit tests for `setupPreprocessorRegistry` hook
- [x] unit tests for AddonsDiscovery
- [ ] unit tests for AddonsFactory
- [x] inline docs for `setupPreprocessorRegistry` hook
- [x] inline docs for AddonsDiscovery
- [ ] inline docs for AddonsFactory
- [x] verify with addon that has HTMLBars templates in the addon (requires ember-cli-htmlbars 0.7.4 or above): [Demo App](https://github.com/rwjblue/templates-in-nested-addons/)
- [ ] document changes necessary to addons to take advantage of this.
- [x] Update to Ember 1.10 by default.

PR #3231, opened 2015-02-09
Remove extra Addon build steps.

rwjblue:
This is in support of https://github.com/ember-cli/ember-cli/pull/3166.  When that lands we will need to ensure that we:

1) Keep output as individual files as long as possible.
2) Preserve the concept that the last instance of a given addon is used.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -588,10 +588,10 @@ EmberApp.prototype._processedVendorTree = function() {
     return this._cachedVendorTree;
   }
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
woop

runspired:
Just pinging to keep notified.

lukemelia:
@runspired you can subscribe to a thread without commenting. Check the right sidebar. screengrab

runspired:
@lukemelia I know, but I subscribe to everything in Ember, Ember-cli, and ember-data by default to keep aware of what's changing.  Adding a comment then bumps it into my "participating" notifications and lets me know I should check it out a little closer ;)

PR #3236, opened 2015-02-09
Remove module transpilation from Addon model.

rwjblue:
The long term goal is to keep ES6 as the lingua franca for JS modules and to extract a "Packager" role to handle the final bundling.


rwjblue:
More internal support refactors to help with https://github.com/ember-cli/ember-cli/pull/3166.

PR #3239, opened 2015-02-10
ENHANCEMENT: Added `--test-port`/`testPort` option to configure test port

### LINKED ISSUE #3022: [Feature Request] Config: Set different test server port to app server port
When setting `port` in the `.ember-cli` config file, it sets the same port for both the regular app and the testing server. This means you have to run one of them with the `--port` flag if running them simultaneously.

Out-of-the-box Ember CLI sets two different ports for the app (4200) and the testing server (7357), it would be nice if you could also set the the app and testing ports separately in the config file.

If this is a change you guys would like to make, I'd be happy to put together the PR for it.

patocallaghan:
Fixes #3022 

/cc @stefanpenner 

~~I hope I've gone about this the right way. It kind of seems weird to me that you can pass `--port` (will output `port+1`) and `--test-port` through the CLI but I needed to do that for my tests to pass. Any pointers/feedback appreciated.~~

Removes `--port` and adds `--test-port` to the `test` command. This is to stop clashes with the port in the app v test server when customizing it in `.ember-cli`

* [x] No port passed defaults to 7357
* [x] Only `port` passed becomes `port` + 1
* [x] `testPort` or `--test-port` becomes the test port
* [x] tests
* [ ] Update docs?

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -7,6 +7,8 @@ var Builder   = require('../models/builder');
 var fs = require('fs');
 var path = require('path');
 
+var defaultPortNumber = 7357;
+
 module.exports = Command.extend({
   name: 'test',
   aliases: ['test', 't'],
@@ -17,7 +19,7 @@ module.exports = Command.extend({
     { name: 'config-file', type: String,  default: './testem.json', aliases: ['c', 'cf'] },
     { name: 'server',      type: Boolean, default: false, aliases: ['s'] },
     { name: 'host',        type: String,  aliases: ['H'] },
-    { name: 'port',        type: Number,  default: 7357, description: 'The port to use when running with --server.', aliases: ['p'] },
+    { name: 'test-port',   type: Number,  default: defaultPortNumber, description: 'The test port to use when running with --server.', aliases: ['tp'] },
     { name: 'filter',      type: String,  description: 'A string to filter tests to run', aliases: ['f'] },
     { name: 'module',      type: String,  description: 'The name of a test module to run', aliases: ['m'] },
     { name: 'watcher',     type: String,  default: 'events', aliases: ['w'] },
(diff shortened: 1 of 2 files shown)


stefanpenner:
> It kind of seems weird to me that you can pass --port (will output port+1)

ya it is kinda strange, but any suggestions?

otherwise this looks good (other then the test failures)

patocallaghan:
We could drop `--port` for the `ember test` command and just use `--test-port`?

Downsides to that are:
1) It's a breaking change for the `ember test` command 
2) it will mean we can't test this functionality in `commands/test-test.js` as it relies on the ability to pass in the different flags to set the option state. Any ideas how it could be tested then?

patocallaghan:
@stefanpenner any thoughts on my last comment?

stefanpenner:
> 1) It's a breaking change for the ember test command

i think this is fine



> it will mean we can't test this functionality in commands/test-test.js as it relies on the ability to pass in the different flags to set the option state. Any ideas how it could be tested then?

unsure, i'll try and noodle on it

stefanpenner:
> it will mean we can't test this functionality in commands/test-test.js as it relies on the ability to pass in the different flags to set the option state. Any ideas how it could be tested then?
> unsure, i'll try and noodle on it

we could add some extra indirection and the mock out those aspects? We can also easily spin up a subprocess like in author tests.

stefanpenner:
@patocallaghan are you still interested in moving this forward?

patocallaghan:
Yep, sorry been slammed at work. Although I'm still a bit unsure how the mocking should work. Could you point me to the tests you mention above?

stefanpenner:
@patocallaghan friendly ping.

patocallaghan:
@stefanpenner Yes, I'd love to bring this to completion. I'm still unsure of how the mocking should work though. You mention:

> We can also easily spin up a subprocess like in author tests.

Could you point me in the direction of those tests as a search didn't bring up anything for me.

patocallaghan:
@stefanpenner this looks like it's good to go. Anything else I need to do?

stefanpenner:
> @stefanpenner this looks like it's good to go. Anything else I need to do?

nope, I was just slow. Thanks :)

PR #3242, opened 2015-02-10
[DOCS] Add isDevelopingAddon

matthiasleitner:


sohara:
:+1: We ran into this issue today :)

stefanpenner:
my dream it to watch everything by default, once broccoli removes tmp from the project, it will be a closer reality.

PR #3244, opened 2015-02-10
Moved reverse() to after filter() so we can keep consistent order

nathanpalmer:
I have a project that has multiple "overrides" for blueprints. However the paths for the main blueprint and the test blueprint were coming up in an inconsistent order. That means I would get the main blueprint for one project and the test blueprint for another.

To help illustrate the problem I have this

1. ember-cli
2. ember-coffeescript
3. in-repo addon

When I run this command

```
ember g adapter users --dry-run
```

It chooses the adapter blueprint from `in-repo addon` and the adapter-test blueprint from `ember-coffeescript`. Even-though both blueprints exist in the `in-repo addon`. When I traced it down it seemed to work everytime as long as we reverse after the filter.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -464,13 +464,13 @@ Project.prototype.blueprintLookupPaths = function() {
   @return {Array} List of paths
  */
 Project.prototype.addonBlueprintLookupPaths = function() {
-  var addonPaths = this.addons.reverse().map(function(addon) {
+  var addonPaths = this.addons.map(function(addon) {
     if (addon.blueprintsPath) {
       return addon.blueprintsPath();
     }
   }, this);
 
-  return addonPaths.filter(Boolean);
+  return addonPaths.filter(Boolean).reverse();
 };
 
 /**
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -222,7 +222,7 @@ describe('models/project.js', function() {
       expect(project.localBlueprintLookupPath()).to.equal(expected);
     });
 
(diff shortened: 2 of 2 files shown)


trabus:
Do you think it would be much work to add a test for this? Would probably be a good thing to check for regression.

stefanpenner:
@trabus is correct, this detail should be tested.

nathanpalmer:
I added a test that shows that it fails when we call this method twice. This is what happens when we get the lookupPaths for the main blueprint and then a second time for the test blueprint

stefanpenner:
@nathanpalmer any luck on getting the tests passing though?

nathanpalmer:
@stefanpenner The test I added and the test I modified are passing just fine.

screenshot 2015-02-10 17 11 25

The master build however wasn't passing when I forked. It seems only related to the iojs build starting with [this commit](https://github.com/ember-cli/ember-cli/commit/e6b477f45ae577c870fbce0f1032f3b14ba78c6a) and [this build](https://travis-ci.org/ember-cli/ember-cli/builds/50207706).

trabus:
It's actually been failing since [this was merged](https://github.com/ember-cli/ember-cli/pull/3233/files).

As far as the tests go, lgtm :+1:

trabus:
If you rebase after https://github.com/ember-cli/ember-cli/pull/3250 gets merged in, we can be sure.

rwjblue:
#3250 is merged, can you rebase?

nathanpalmer:
Sure, I'll rebase when I get home.

rwjblue:
Thanks!

PR #3248, opened 2015-02-10
closes #3195 enable editing diffs on windows

### LINKED ISSUE #3195: ember init doesn't allow editing diffs on windows when upgrading
Is there a reason for [this](https://github.com/ember-cli/ember-cli/blob/602a1c4558fa3af8fef21b75341945099dc9f82a/lib/models/file-info.js#L50) line disallowing windows from editing diffs?

    function canEdit() {
      return ! /^win/.test(process.platform) && process.env.EDITOR !== undefined;
    }

I'm willing to help if it needs more work to support windows here.

jesse-black:
After successfully editing diffs with this on windows I don't see any reason for excluding windows from editing diffs. It doesn't appear to require any special support for windows.

### DIFF
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -47,7 +47,7 @@ FileInfo.prototype.confirmOverwrite = function(path) {
 };
 
 function canEdit() {
-  return ! /^win/.test(process.platform) && process.env.EDITOR !== undefined;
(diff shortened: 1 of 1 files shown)


stefanpenner:
I believe we disabled it, because when it was implemented no-one had time to confirm it works on windows. But if you say it works, then fantastic. :)

PR #3278, opened 2015-02-13
Bump Core Object and remove it from bundled dependencies

chadhietala:


chadhietala:
Closing for now, I think there actually needs to be a lot refactoring before this can happen.

PR #3279, opened 2015-02-13
Passing a git url as the blueprint is supported but not documented

chadhietala:
In the implementation of init blue print is shoveled off into `installBlueprint`. `installBlueprint` determines if it was passed a git url or a regular string.

This PR just documents that this is possible.


rwjblue:
What does the help output look like with this (curious)?

chadhietala:
It's the same as `ember new --help`

```
--blueprint (gitUrl, Path) (Default: app)
aliases: -b <value>
```

joostdevries:
it's a shame that nopt doesnt support a bit more descriptive typedefs but other than that I'm :cool:

chadhietala:
Will merge if @rwjblue is cool with this.

PR #3283, opened 2015-02-13
Add clarification for several steps in the addon creation tutorial.

endash:
What it says on the tin. This closely follows my haphazard process of discovery by looking at other addons, so if anything can be streamlined, is ultimately inaccurate, or otherwise could be changed for the better that's no prob.


joostdevries:
:+1: awesome thanks!

endash:
I missed some ember addon/addon dependency name confusions.

PR #3288, opened 2015-02-13
Override default port with `PORT` env var

knownasilya:
`PORT=3000 npm start` will run the server on port 3000.

Where `npm start` is `ember serve`.

If this sounds good, I can add a test. Also, might be nice to add the override for when the port is set. Maybe it needs to be done at a different level..

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -12,7 +12,7 @@ module.exports = Command.extend({
   aliases: ['server', 's'],
 
   availableOptions: [
-    { name: 'port', type: Number, default: 4200, aliases: ['p'] },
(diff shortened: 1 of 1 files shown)


knownasilya:
Maybe here instead? https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/serve.js#L15

joostdevries:
I'm not directly seeing a use case for this, did you have something in mind? Also I'd use a more explicit ENV var, like `EMBER_CLI_PORT` or even `EMBER_CLI_HTTP_PORT`

knownasilya:
`PORT` is the standard env variable that most node.js applications use, so it's appropriate.

The use case is to be able to use `npm start` and being able to change the port without having to change the script for `npm start`.

PR #3289, opened 2015-02-13
ember-cli is a invalid project name

flecno:
I tried to build an example project which is named `ember-cli`. It seems there is a problem with the project name:

```
~ % cd /tmp
/tmp % ember new ember-cli
version: 0.1.12
installing
[...]
Installed browser packages via Bower.
Successfully initialized git.
ember new ember-cli  18.09s user 2.20s system 92% cpu 21.951 total
/tmp % cd ember-cli 
/tmp/ember-cli % ember build
version: 0.1.12
You have to be inside an ember-cli project in order to use the build command.
```

Also I see this for names like `ember`, so I would extend the tests with names like `ember-cli`. I simply add the exception to `lib/utilities/valid-project-name.js` but if I run `npm test` it doesn't work. It seems the tests use my cloned directory `ember-cli` as project name for other tests? But I have no idea how to fix it.

Can you help me? What's the reason for the name exceptions? Is there is a problem with projects named `ember` etc. in resolving of ES6 modules? Is there any idea how to resolve this?

### DIFF
--- a/tests/unit/commands/init-test.js
+++ b/tests/unit/commands/init-test.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var fs            = require('fs');
+var os            = require('os');
 var path          = require('path');
 var expect        = require('chai').expect;
 var MockUI        = require('../../helpers/mock-ui');
@@ -91,12 +93,23 @@ describe('init command', function() {
 
 
   it('Uses process.cwd if no package is found when calling installBlueprint', function() {
+    // change the working dir so `process.cwd` can't be a invalid path for base directories 
+    // named `ember-cli`.
+
+    var tmpDir = os.tmpdir();
+    var workingDir = tmpDir + '/ember-cli-test-project';
+    var currentWorkingDir = process.cwd();
+
+    fs.mkdirSync(workingDir);
+    process.chdir(workingDir);
+
     tasks.InstallBlueprint = Task.extend({
       run: function(blueprintOpts) {
         expect(blueprintOpts.rawName).to.equal(path.basename(process.cwd()));
         return Promise.reject('Called run');
       }
     });
+
     var command = new InitCommand({
       ui: ui,
       analytics: analytics,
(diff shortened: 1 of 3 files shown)


joostdevries:
@flecno not entirely sure why the tests are failing but we're prohibiting project names that collide with predefined namespaces (like `ember`, `test` and `vendor`). While `ember-cli` is not something the resolver would have issues with I think it's a sensible idea to prohibit it as a project name because of https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L542.

When we get the tests passing we'll merge this.

joostdevries:
Ok so the failure is related to the fact that in one of the unit tests for `ember init` we're apparently trying to create a project called `ember-cli` (https://github.com/ember-cli/ember-cli/blob/master/tests/unit/commands/init-test.js#L103). @rwjblue could we fix the init test to use a tmpdir?

rwjblue:
Yes, changing that test makes sense.

flecno:
@joostdevries thanks for your tips. I changed the init test but I'm not happy with the fix. I think it doesn't fit into the coding conventions if I look around the tests. (beginning with importing `fs` and `os`).

I would be happy for suggestions for build this better.

stefanpenner:
kicking appveyor to rebuild.

tests/unit/commands/init-test.js
```diff
@@ -91,12 +93,23 @@ describe('init command', function() {
...
   it('Uses process.cwd if no package is found when calling installBlueprint', function() {
+    // change the working dir so `process.cwd` can't be a invalid path for base directories 
+    // named `ember-cli`.
+
+    var tmpDir = os.tmpdir();
+    var workingDir = tmpDir + '/ember-cli-test-project';
+    var currentWorkingDir = process.cwd();
+
+    fs.mkdirSync(workingDir);
+    process.chdir(workingDir);
```
stefanpenner: im not sure if this works correctly in windows. If i recall something does go side-ways here.

PR #3297, opened 2015-02-15
[bugfix] Use ember-cli-babel instead of ember-cli-6to5

fivetanley:
The 6to5 project has been renamed to Babel. This commit *depends on*
https://github.com/babel/ember-cli-6to5/pull/14 being merged and
a 4.0.0 version package being released. This should also be called out
in the release notes so that people won't be confused when 6to5
disappears when running `ember init`.


fivetanley:
I think after travis/appveyor gets restarted this can be merged. all the work on ember-cli-6to5 has gone through and is now ember-cli-babel

stefanpenner:
triggering rebuild

PR #3311, opened 2015-02-17
Fix failing pods destroy test.

rwjblue:
Because we were not returning the promise these tests would randomly pass when they should have been failing (they were expecting the files to be in an incorrect location).


trabus:
D'oh. Sorry about this!

Issue #3319, opened 2015-02-19
consider min-node version of 0.12.x


stefanpenner:
@ember-cli/owners thoughts?

- [x] deprecate `0.10.x` https://github.com/ember-cli/ember-cli/pull/3326
- [ ] c-extensions that don't work yet?
  - [ ] node-sass ?

trabus:
Are we talking as of ^0.2.1? I'm assuming this is primarily so we can start using more ES6 in ember-cli itself? 

I will need to have a conversation with our ops team to get 0.12 onto our build server, and I imagine that will be the case for a lot of slower moving companies. I imagine there are still lots of companies on 0.10.x out there.

Personally, I won't mind as long as we can get the 0.12 builds stop failing from timeouts on our slower tests.

stefanpenner:
> I imagine there are still lots of companies on 0.10.x out there.

its trivial to change this...



> Are we talking as of ^0.2.1? I'm assuming this is primarily so we can start using more ES6 in ember-cli itself?

my main motivation is actually `execSync` being built in and no longer requiring a c-extension.

trabus:
> its trivial to change this... is this honestly something ops people have trouble doing? Isn't this their job?
I'm sure it is, but getting anything done in a corporate setting will always be met with some pushback from somewhere.

stefanpenner:
> I'm sure it is, but getting anything done in a corporate setting will always be met with some pushback from somewhere.

:cry:

raytiley:
> its trivial to change this... is this honestly something ops people have trouble doing? Isn't this their job?

lol. Haven't been at yahoo long I guess. Unless it provides some huge win that can't be gained any other way it seems like a pretty stiff requirement. 

At my company the only reason anyone know any node, myself included is because of ember-cli.

stefanpenner:
> Unless it provides some huge win that can't be gained any other way it seems like a pretty stiff requirement.

the `huge` win, is that ember-cli works.. :+1:

trabus:
I honestly don't think that it's a huge issue to require it, it's only needed for the build server, and even then there are ways around needing different versions of node. I think I'm just trained to expect caution when asking to upgrade software. Where I've worked it has always been a slow cautious process due to upgrades and things breaking often being associated with each other. I imagine I'm not the only one with a (overly) cautious work culture.

Btw, I just spoke to our ops guy and he had no problem with it, I just need to make a ticket for it. :+1:

stefanpenner:
[...]

PR #3326, opened 2015-02-20
Print deprecation warning for Node 0.10.

rwjblue:
Adds warning when running on older Node versions as mentioned in https://github.com/ember-cli/ember-cli/issues/3319#issuecomment-75230978.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -7,6 +7,8 @@ var UpdateChecker   = require('../models/update-checker');
 var getOptionArgs   = require('../utilities/get-option-args');
 var debug           = require('debug')('ember-cli:cli');
 
+var validPlatformVersion = require('../utilities/valid-platform-version');
+
 function CLI(options) {
   this.ui   = options.ui;
   this.analytics = options.analytics;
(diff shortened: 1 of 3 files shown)


stefanpenner:
:+1:

govidat:
I was in Node 0.10 and I got this warning.

I updated to current Node version 0.13.0-pre. I still get this warning when I do "ember server" :
Future versions of Ember CLI will not support v0.13.0-pre. Please update to Node 0.12 or io.js.

intuitivepixel:
@govidat why not use node 0.12.0 then?

govidat:
Hi thanks.
I did not select the version of 0.12.0 and just took the latest, which
happens to be 0.13. I presume that this higher node version should be ok.

Or do you suggest that I should go to the lower version of 0.12.

intuitivepixel:
@govidat no, not lower than 0.12 just stable 0.12 should work perfectly, the check is for node 0.12.x that's why it's failing , but version 0.13 will most probably become io.js anyway.

rwjblue:
Please beware that odd versions of Node are considered "unstable", but it should be allowed so I submitted https://github.com/ember-cli/ember-cli/pull/3601 to allow node 0.13.

PR #3333, opened 2015-02-20
rename variables to be clear

stefanpenner:


lib/broccoli/ember-app.js
```diff
@@ -839,7 +839,7 @@ EmberApp.prototype.javascript = function() {
   var legacyFilesToAppend = this.legacyFilesToAppend;
   var appOutputPath       = this.options.outputPaths.app.js;
 
-  var es6 = new transpileES6(
+  var app = new ES6Modules(
```
rwjblue: It isn't the whole app, maybe `appJs`?

stefanpenner: :+1:

PR #3336, opened 2015-02-22
Nested addons should be overrideable from parent.

rwjblue:
The initial logic added to call `treeFor` on all of a given addons
nested children had the following issues:

* The trees returned could not be overridden by the including addon.
* The final merging of those trees did not use the `overwrite: true`
  flag, which meant that any shared addons would throw an error instead
  of simply using the last one (which is the Ember CLI 0.1.x behavior).

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -248,7 +248,7 @@ Addon.prototype.treeFor = function treeFor(name) {
   this._requireBuildPackages();
 
   var tree;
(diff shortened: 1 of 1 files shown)


rwjblue:
Should address an issue pointed out in IRC by @lukemelia.

lukemelia:
Thanks @rwjblue. This should cover what @raycohen and I ran into today.

PR #3339, opened 2015-02-22
Move preprocessor packages to dependencies in addons

teddyzeenny:
We can later add more addons to the list (such as babel) when they start supporting ember-cli 0.2.

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -53,6 +53,8 @@ describe('Acceptance: addon-smoke-test', function() {
     expect(packageContents.private).to.be.an('undefined');
     expect(packageContents.keywords).to.deep.equal([ 'ember-addon' ]);
     expect(packageContents['ember-addon']).to.deep.equal({ 'configPath': 'tests/dummy/config' });
+    expect(packageContents.dependencies).to.be.an('object');
+    expect(packageContents.dependencies['ember-cli-htmlbars']).to.be.a('string');
 
     var bowerContents = JSON.parse(fs.readFileSync('bower.json', { encoding: 'utf8' }));
 
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -13,6 +13,15 @@ module.exports = {
     var packagePath = path.join(this._appBlueprint.path, 'files', 'package.json');
     var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
 
+    // preprocessors should be under dependencies
+    var preprocessors = ['ember-cli-htmlbars'];
+    preprocessors.forEach(function(name) {
+      var version = contents.devDependencies[name];
+      delete contents.devDependencies[name];
+      contents.dependencies = contents.dependencies || {};
+      contents.dependencies[name] = version;
+    });
+
     delete contents.private;
     contents.name = this.project.name();
     contents.description = this.description;


lukemelia:
Looks good.

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -53,6 +53,8 @@ describe('Acceptance: addon-smoke-test', function() {
     expect(packageContents.private).to.be.an('undefined');
     expect(packageContents.keywords).to.deep.equal([ 'ember-addon' ]);
     expect(packageContents['ember-addon']).to.deep.equal({ 'configPath': 'tests/dummy/config' });
+    expect(packageContents.dependencies).to.exist;
+    expect(packageContents.dependencies['ember-cli-htmlbars']).to.exist;
```
chadhietala: JSHint tests are failing. Perhaps this should change to

```
expect(packageContents.dependencies).to.an('object');
expect(packageContents.dependencies['ember-cli-htmlbars']).to.equal('0.15.0')
```

rwjblue:
ember-cli-htmlbars is only needed as a dep if you have templates in `addon/`, do we want all addons created by default to bring a dep that is generally unused?

teddyzeenny:
@rwjblue good point. I lost some time on this so I thought I'd help others avoid it. It may not be a good idea after all.

I assumed however that in `0.2` addon builds will be completely isolated and even files in `app/` would use the addon's own preprocessors. Is that not the case? It seems strange to use different registries for two folders in the same addon. For example I'm always fighting with jscs linting the `app/` folder of addons I don't own.

lukemelia:
@rwjblue just thinking about real-world use cases here. Imagine two scenarios for an addon which has templates in app/ or addon/:

1) the templates use Handlebars-incompatible htmlbars features and you would need the app to run with the htmlbars runtime. Precompiling could happen either by the app or addon, it wouldn't matter.
2) the templates are Handlebars-compatible, and you would want the templates to be compiled either by htmlbars or handlebars, depending on what the app was using.

Is my thinking off base here?

rwjblue:
> I assumed however that in 0.2 addon builds will be completely isolated and even files in app/ would use the addon's own preprocessors. Is that not the case?

No, that is not the case. The idea of the `app/` tree is is that it is merged into the final app, exactly the same as any actual code the user has in `app/`.  This means that anything in `app/` tree will use the apps preprocessors (since it is merged directly with the app before any preprocessing happens).

> For example I'm always fighting with jscs linting the app/ folder of addons I don't own.

That is a different issue: linting as a preprocessor is no longer needed, I'll PR upstream in broccoli-jscs.

rwjblue:
My suggestion is, and has generally always been, that nearly all code in an addon should be in the `addon/` tree and imported/re-exported in `app/` tree for those things that need to be merged with the app (mostly just components, helpers, and initializers for lookup purposes).

For example, if your addon provides a component with a template:

Template:

```hbs
// addon/templates/foo-bar.hbs
{{#if someStuff}}
  {{yield}}
{{/if}}
```

Component:

```javascript
// addon/components/foo-bar.js

import Ember from 'ember';
import layout from '../templates/foo-bar';

  layout: layout
});
```

Re-export:

```javascript
import FooBar from 'my-addon-name/components/foo-bar';

export default FooBar;
```

In the above case, the template would be processed by the addon's preprocessors (so it could use Emblem or custom AST walkers or whatever) and simply be available (and overridable) in the app.

rwjblue:
> Is my thinking off base here?

I'm not really sure if I addressed your question, did I?

rwjblue:
Submitted https://github.com/kellyselden/broccoli-jscs/pull/17 to address the JSCS issue mentioned by @teddyzeenny above.

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -46,13 +46,15 @@ describe('Acceptance: addon-smoke-test', function() {
     });
   });
 
-  it('generates package.json and bower.json with proper metadata', function() {
+  it.only('generates package.json and bower.json with proper metadata', function() {
```
stefanpenner: oops?

stefanpenner:
> Component:
> 
> // addon/components/foo-bar.js
> 
> import Ember from 'ember';
> import layout from '../templates/foo-bar';
> 
> export default Ember.Component.extend({
>   layout: layout
> });

we need generators in an addon do to this.

stefanpenner:
> he idea of the app/ tree is is that it is merged into the final app, exactly the same as any actual code the user has in app/. This means that anything in app/ tree will use the apps preprocessors (since it is merged directly with the app before any preprocessing happens).

I was not aware of this. I assumed this would be pre-processed by the addons preprocessors but in a 1:1 way. Emitting ES5 safe code with the exceptions of es6 modules.

stefanpenner:
> 1) the templates use Handlebars-incompatible htmlbars features and you would need the app to run with the htmlbars runtime. Precompiling could happen either by the app or addon, it wouldn't matter.

my understanding, which may be wrong. HTMLBars has no need for a runtime, as ember provides those hooks.

stefanpenner:
This thread is illuminating some mis-understandings.

@rwjblue should I compile an thorough RFC for these interactions, and we can use that as a baseline for discussions?

trabus:
I think what we should do is break the addon rexport code into its own blueprint (`component-addon` for instance), similar to [what we do for tests](https://github.com/ember-cli/ember-cli/tree/master/blueprints/component-test), and [conditionally install](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/generate-from-blueprint.js#L44) them from the `generate-from-blueprint.js` task. 

We can add one more token for the root folder, so in `/files` of your blueprint, you'd have: `__root__/__path__/__file__.js` 
For instance when generating a component blueprint into an addon, the `__root__` token would be "addon", but otherwise it would be "app" for a regular project.

The addon rexport blueprint would be appended with `-addon`, so we could look it up in the same manner we do for test blueprints. We'd add a check in the task for `addonBlueprint`, and install them if they exist just like we do for test blueprints.

rwjblue:
@trabus - That sounds like an awesome plan to get the blueprints working like we want inside an addon project.  I opened https://github.com/ember-cli/ember-cli/issues/3354 to track this.

teddyzeenny:
Closing this to avoid adding deps to all addons.

Issue #3331, opened 2015-02-20
Improve blueprint help output method (markdown support)


trabus:
As per discussion in [#3316](https://github.com/ember-cli/ember-cli/pull/3316#issuecomment-75205628), we should simplify the way blueprint help content is added. With the addition of #3316 we can add a `printDetailedHelp` hook to print help output. The [current way the output is built](https://github.com/ember-cli/ember-cli/blob/master/blueprints/model/index.js#L51) is pretty primitive and not a pattern I'd like to endorse moving forward.

I'm thinking that we can simplify this by adding a default printDetailedHelp method that will look for a `README.md` or `HELP.md` in the blueprint root folder then parse (likely a specific section of it) and return the output formatted for the terminal. 

Writing the help text in a format that can be parsed into color text may be a bit challenging, but I figured we could write a lexer for it. The [marked](https://github.com/chjj/marked) library looks pretty good, and [ought to do the job](https://github.com/chjj/marked#access-to-lexer-and-parser).

trabus:
Looks like [marked-terminal](https://github.com/mikaelbr/marked-terminal) may do most of what I need out of the box. Though custom color markup (and a lexer to recognize it) will need to be defined. I was thinking something like `^color^` could work for marking the different colors. So `^cyan^text goes here^cyan^` would be for cyan, `^yellow^` for yellow, and so on.

stefanpenner:
This is a great idea

trabus:
So after some investigation, it appears that marked it isn't as easily extensible as I'd hoped. You can override existing tokens, but adding new ones is a whole different story. 

What I'm intending on doing instead is parsing the `README.md` with marked first, then doing a string replace afterwards to support the custom color tokens. Something like:
```
var tokens = {
  cyan: /(?:\^cyan\^)(.*?)(?:\^cyan\^)/,
  green: /(?:\^green\^)(.*?)(?:\^green\^)/
};
var renderer = {
  cyan: function cyan(match, p1){
    return chalk.cyan(p1);
  },
  green: function green(match, p1){
    return chalk.green(p1);
  }
};

// then later
marked(input).replace(tokens.cyan,renderer.cyan);
```

I'll probably get an array of the tokens used in the string, then loop through them to run the replacement. I'm probably going to add something to look for [these patterns too](https://github.com/ember-cli/ember-cli/blob/master/ARCHITECTURE.md#formatting-colors), that would occur before the `^color^` tokens so they could be overridden if needed.

[...]

PR #3341, opened 2015-02-22
Improve blueprint help output method (markdown support)

trabus:
This PR fulfills #3331 by adding a new utility for parsing markdown into color terminal output. Additional tokens have been added, as well as recognized patterns such as `--option` (should be cyan) and `<argument>` (should be yellow) as laid out [here](https://github.com/ember-cli/ember-cli/blob/master/ARCHITECTURE.md#formatting-colors).
The current implementation is as simple as:
```
var markdownColor = new MarkdownColor();
markdownColor.renderFile('./HELP.md');
```
Example output:
(the first block is output after being parsed with marked, the second is after parsing with markdown-color)
screen shot 2015-02-22 at 1 36 28 pm

`MarkdownColor` also exposes a `render(string)` method to allow manipulation to the input before rendering.

Internally, tokens are defined as a regex pattern and the chalk style[s] to apply to the matched text. For instance:
```
var tokens = {
  cyan: {
    token: '<cyan>', // example of what the token looks like
    pattern:/(?:<cyan>)(.*?)(?:<\/cyan>)/g, // pattern to match token
    render: renderStylesFactory('cyan') // chalk style to render matched text with
  },
  option: {
    token: '--option',
    pattern: /((--\w*\b)|(<options>))/g,
    render: renderStylesFactory(['cyan', 'underline]) //renders both cyan and underlined
  }
};
```
The currently available color tokens are as follows:
* red:` <red></red>`
* green:` <green></green>`
* blue:` <blue></green>`
* cyan:` <cyan></green>`
* magenta:` <magenta></magenta>`
* yellow:` <yellow></yellow>`
* black:` <black></black>`
* grey:` <grey></grey>`
* gray:` <gray></gray>`
* bgRed:` <bgRed></bgRed>`
* bgGreen:` <bgGreen></bgGreen>`
* bgBlue:` <bgBlue></bgGreen>`
* bgCyan:` <bgCyan></bgCyan>`
* bgMagenta:` <bgMagenta></bgMagenta>`
* bgYellow:` <bgYellow></bgYellow>`
* bgBlack:` <bgBlack></bgBlack>`

The tokens can then be used in your markdown like so:
```
// file.md
<cyan>This text is cyan <red>but this is red</red> this back to cyan though</cyan>
```

Some additional work needs to be done:
- [x] fix code highlighting (more involved than this PR, will have to be resolved later)
- [x] clean up some of the ember-cli patterns
- [x] add tokens and renderers for the other styles and background colors
- [x] determine if any additional preset styles should be added
- [x] add optional indent value
- [x] unit tests for `markdown-color.js`
- [x] ~~acceptance tests for `markdown-color.js`~~ (stdout eats ansi, so this would just be a duplicate of the unit test)

### DIFF
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,6 +1,4 @@
-      model <name> <attr:type>
-        ^gr^Generates an ember-data model.^gr^
-        ^gr^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^gr^
+         ^gr^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^gr^
           <attr-name>
           <attr-name>:array
           <attr-name>:boolean
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,15 +1,15 @@
-^grey^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^grey^
-  <attr-name>
-  <attr-name>:array
-  <attr-name>:boolean
-  <attr-name>:date
-  <attr-name>:object
-  <attr-name>:number
-  <attr-name>:string
-  <attr-name>:belongs-to:<model-name>
-  <attr-name>:has-many:<model-name>
+<grey>You may generate models with as many attrs as you would like to pass. The following attribute types are supported:</grey>
+  <yellow><attr-name></yellow>
+  <yellow><attr-name></yellow>:array
+  <yellow><attr-name></yellow>:boolean
+  <yellow><attr-name></yellow>:date
+  <yellow><attr-name></yellow>:object
+  <yellow><attr-name></yellow>:number
+  <yellow><attr-name></yellow>:string
+  <yellow><attr-name></yellow>:belongs-to:<yellow><model-name></yellow>
+  <yellow><attr-name></yellow>:has-many:<yellow><model-name></yellow>
 
-For instance: ^green^\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`^green^
+For instance: <green>\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`</green>
 would result in the following model:
 
 ```js
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,4 +1,4 @@
-^gr^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^gr^
+^grey^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^grey^
   <attr-name>
   <attr-name>:array
   <attr-name>:boolean
@@ -9,16 +9,16 @@
   <attr-name>:belongs-to:<model-name>
   <attr-name>:has-many:<model-name>
 
-For instance: ^g^\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`^g^
+For instance: ^green^\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`^green^
 would result in the following model:
 
 ```js
 import DS from 'ember-data';
 export default DS.Model.extend({
   filling: DS.belongsTo('protein'),
-  toppings: DS.hasMany('topping')
+  toppings: DS.hasMany('topping'),
   name: DS.attr('string'),
-  price: DS.attr('number')
+  price: DS.attr('number'),
   misc: DS.attr()
 });
 ```
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,0 +1,25 @@
+      model <name> <attr:type>
+        ^gr^Generates an ember-data model.^gr^
+        ^gr^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^gr^
+          <attr-name>
+          <attr-name>:array
+          <attr-name>:boolean
+          <attr-name>:date
+          <attr-name>:object
+          <attr-name>:number
+          <attr-name>:string
+          <attr-name>:belongs-to:<model-name>
+          <attr-name>:has-many:<model-name>
+
+        For instance: ^g^`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc`^g^
+        would result in the following model:
+```js
+        import DS from 'ember-data';
+        export default DS.Model.extend({
+          filling: DS.belongsTo('protein'),
+          toppings: DS.hasMany('topping')
+          name: DS.attr('string'),
+          price: DS.attr('number')
+          misc: DS.attr()
+        });
+```
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,23 +1,24 @@
-         ^gr^You may generate models with as many attrs as you would like to pass. The following attribute types are supported:^gr^
-          <attr-name>
-          <attr-name>:array
-          <attr-name>:boolean
-          <attr-name>:date
-          <attr-name>:object
-          <attr-name>:number
-          <attr-name>:string
-          <attr-name>:belongs-to:<model-name>
-          <attr-name>:has-many:<model-name>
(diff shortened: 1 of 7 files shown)


trabus:
I've updated it so the `printDetailHelp` has a default for blueprints that looks for a specific file (`HELP.md`), but it won't fail if the markdown file isn't present.

I've got some process ordering issues I need to figure out with nesting the styles, but aside from that, it seems pretty close. I've added the ability to extend the tokens upon instantiation of a markdown-color instance.

I still need to add some more tests, and I'd like to tighten up the regex patterns a bit to make sure I'm properly capturing the tokens and the values contained between. My next step however will be to figure out why the code highlighting is not properly functioning.

@jgwhite @stefanpenner @rwjblue @joostdevries Would love some input on my approach and whether the tokens need to change, or if they're okay as is.

joostdevries:
@trabus I could probably look into this further tomorrow. From an initial glance this looks good. Two initial questions:
- Why is the markdown indented?
- Is it wise to give that much formatting power? While very awesome it could also result in more diversity in how the help texts look.

trabus:
> Why is the markdown indented?

The blueprint help output is [currently](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/generate.js#L134) [indented](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/generate.js#L152). The markdown was indented to match it, but I'm open to ideas on whether we should add indenting post parse, or whether we should have it at all.

> Is it wise to give that much formatting power? While very awesome it could also result in more diversity in how the help texts look.

This definitely could get out of hand in addon-land if we were to add the feature without guidelines and recommendations. I think some documentation should be added to the README.md of the addon blueprint, as well as some documentation in the website docs. It might be good to limit the tags to those we use though, I just went through and added ones for all the styles available in chalk.

stefanpenner:
awesome, some requests:

* color cheat-sheet
* steps/tools to quickly preview

stefanpenner:
just an idea, what if instead of `^r^` we used a tag `<green> asdfadsf </green>` ? or something..

trabus:
> just an idea, what if instead of ^r^ we used a tag <green> asdfadsf </green> ? or something..

I can easily change the tokens, I just wanted to make sure I wasn't going to clobber anything with them. Using the carat seemed like the easiest and obscure enough to not conflict. I'm absolutely open to whatever token we'd like to use collectively though.

`<color></color>` sounds good to me personally, but I'd like to investigate cases where it could collide.

stefanpenner:
> <color></color> sounds good to me personally, but I'd like to investigate cases where it could collide.
> Hide all checks

im fine with whatever is your preference here. I am merely adding another option, I will be content with either. The markdown approach makes me excited :)

trabus:
Seems that my tests don't fly on Windows.
screen shot 2015-02-23 at 4 54 43 pm

@raytiley any suggestions on how to handle this?

raytiley:
I'll pull it down and see what I can figure out.

raytiley:
@trabus not sure whats going on here. Your tests pass on my windows 8 machine. Its possible this might have something to do w/ the windows CI environment. I'll kick the build and bring up to appveyor to see if they have any insights.

ember-cli-win-dev 2015-02-23 21-24-46

trabus:
@raytiley Thank you for looking into this! Let me know if there is anything I can do to help from my end.

trabus:
@raytiley Are you using cmdr or something other than cmd.exe? I'm guessing that the reason it's failing is because chalk uses ansi [which isn't supported for cmd.exe](https://github.com/sindresorhus/chalk#windows).

Looks like I'll need to figure out another way to test it. :\

raytiley:
@trabus ahh... I wasn't thinking of that. I was using powershell in the screenshot. I usually use cmd but I thought appveyor was using powershell. I just submitted a PR to switch to powershell.. kinda curious if it works. I feel like when I got the test suite passing I switched to cmd for some other bug. Not sure what that was.

trabus:
@ember-cli/owners So in that case should we recommend people use powershell or cmdr for ember-cli?

twokul:
@trabus sounds like a good idea

stefanpenner:
we need to support both, but we can recommend one.

trabus:
I figure I can check if ascii is supported and have different expected results for when it's not supported. Not great, but it would probably work.

raytiley:
I'll work on the powershell stuff over the weekend. What's confusing me is looking at these tests were not actually hitting the shell are we?

trabus:
So the ability to call this.skip() from within a test was [added recently](https://github.com/mochajs/mocha/pull/946), but has not made it into a tagged version yet. I'm currently using a sort of hack to skip if ansi isn't supported.

```
function isAnsiSupported() {
  // when ansi is supported this should be '\u001b[0m\u001b[31ma\u001b[39m\u001b[0m\n\n'
  return chalk.red('a') !== 'a';
}
(isAnsiSupported() ? describe : describe.skip)('MarkdownColor', function() {
...
```

trabus:
> I'll work on the powershell stuff over the weekend. What's confusing me is looking at these tests were not actually hitting the shell are we?

I don't have an answer for that, but it does seem that the environment of where the chalk command is being run for the test has bearing on the output.

trabus:
So after looking into the code highlighting, there are some issues with [cardinal.js](https://github.com/thlorenz/cardinal) parsing es6 syntax, so it falls back to plain highlighting after encountering an error. For instance when the `import` statement is encountered:
```
Error: Unable to perform highlight. The code contained syntax errors: Line 1: Unexpected reserved word
    at throwError (/Users/trabus/Development/cardinal-test/node_modules/cardinal/node_modules/redeyed/node_modules/esprima/esprima.js:1156:21)
    at throwUnexpected (/Users/trabus/Development/cardinal-test/node_modules/cardinal/node_modules/redeyed/node_modules/esprima/esprima.js:1204:17)
``` 
I figure this is a minor issue, that we can tackle later.

trabus:
I got code highlighting working locally, after swapping out esprima with esprima-fb. I'll make an effort to reach out and see if I can make a PR to add optional ES6 support for [cardinal.js](https://github.com/thlorenz/cardinal) and [redeyed.js](https://github.com/thlorenz/redeyed), then see if marked-terminal can upgrade to use the newer cardinal.js.

trabus:
[Working with](https://github.com/thlorenz/redeyed/pull/2) @thlorenz to get a new version of redeyed.js out that supports ES6. Will try to get cardinal.js and marked-terminal upgraded as soon as it's released.

trabus:
So I think we can go out without the code highlighting support, but as soon as the ES6 updates hit cardinal and marked-terminal, I'll bump the version for the support. Until then code will look like this:

screen shot 2015-03-03 at 2 28 16 pm

After we update, it will look like this:
screen shot 2015-03-03 at 2 47 58 pm

I'm going to write a couple acceptance tests for it, and then I think it's good to go. :shipit:

rwjblue:
This looks pretty good to me at the moment, @trabus what additional tests are you planning to add?

twokul:
@rwjblue we briefly talked with @trabus about adding integration tests for it but it turns out that `stdout` strips ANSI chars out and returns a string so I don't think there's a way to verify string + color tags (I might be wrong). Unit tests LGTM.

trabus:
@rwjblue I'm going to add an acceptance test where I intercept the output from MockUI rather than passing them through, so I can see the raw output before the ansi codes get parsed out.

I'm also swapping the tokens, instead of `^color^` we'll use `<color></color>`. It required the removal of the `<args, options>` token, but that can be manually tagged.

I'm also going to try and make the renderer pluggable, so we can drop in a regular markdown renderer + color parsing or other transforms, which will open the door for using the markdown-color utility for parsing into docs for the website.

trabus:
@stefanpenner @rwjblue @jgwhite This is ready for review. 
I've updated the tokens to use `<color></color>` format, and add the ability to pass in a markdown renderer as well as a function to handle render styles (like chalk).

Next step is to add docs to gh-pages, referring to `HELP.md` and how to add tokens, or use a custom renderer, or even how to override printDetailedHelp and use curl to pull in the markdown to parse and output in the terminal (as suggested by @rwjblue).

rwjblue:
:+1:

PR #3353, opened 2015-02-23
Update pr 3186

stefanpenner:
[closes #3186]


rwjblue:
I'm :+1: assuming that the addon case (where absolute paths are an issue) will be addressed separately (in https://github.com/ember-cli/ember-cli/issues/3354).

PR #3355, opened 2015-02-24
Add `ui` to `Project` and `Addon` instances.

rwjblue:
/cc @stefanpenner


rwjblue:
There is an issue with AppVeyor installing phantom at the moment, will need to restart the build in a few hours (hopefully fixed by then).

stefanpenner:
this should be fine on windows. I'll deal with fallout if it happens.

PR #3361, opened 2015-02-25
bump loader

stefanpenner:
now includes `alias`  @rwjblue / @ef4 this should be nicer then the current re-export thing we got going on.


rwjblue:
I'll update the re-exporter tomorrow.

PR #3378, opened 2015-02-26
Check which addon we are developing

teddyzeenny:
`isDevelopingAddon` now only returns `true` if it's the addon we're developing (instead of returning true for all addons). This will fix issues such as linting all addons instead of just the addon itself. 

I think a simpler `if(this.parent.name() === this.name)` would work instead of going through the addon discovery object, but I wasn't sure if we wanted to make this assumption (that the addon and project name will always be the same).

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -21,6 +21,7 @@ var preprocessTemplates = p.preprocessTemplates;
 var AddonDiscovery  = require('../models/addon-discovery');
 var AddonsFactory   = require('../models/addons-factory');
 var CoreObject = require('core-object');
+var Project = require('./project');
 
 /**
   Root class for an Addon. If your addon module exports an Object this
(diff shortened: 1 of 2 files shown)


lib/models/addon.js
```diff
@@ -154,7 +155,13 @@ Addon.prototype._unwatchedTreeGenerator = function unwatchedTree(dir) {
   @return {Boolean}
 */
 Addon.prototype.isDevelopingAddon = function() {
-  return process.env.EMBER_ADDON_ENV === 'development';
+  if (process.env.EMBER_ADDON_ENV === 'development' && (this.parent instanceof Project)) {
+    var addon = this.addonDiscovery.discoverFromProjectItself(this.parent);
```
rwjblue: `parent` is not a project for nested addons

rwjblue: You would have to do `this.project` to get the `project` instance reliably.

rwjblue:
Would be great to get this fixed, thanks for tackling it.

lib/models/addon.js
teddyzeenny: Yeah that's why I'm checking `this.parent instanceof Project`, the `project` param is marked as deprecated at the top of the addon, so I wasn't sure if I could reliably use it.

stefanpenner:
woohoo!

PR #3383, opened 2015-02-27
Use Ember.HTMLBars by default in new helpers.

maxwerr:
I assume this is the preferred way to go since Ember.Handlebars.makeBoundHelper is deprecated.

http://emberjs.com/api/classes/Ember.Handlebars.html#method_makeBoundHelper
http://emberjs.com/api/classes/Ember.HTMLBars.html#method_makeBoundHelper

### DIFF
--- a/blueprints/helper/files/app/helpers/__name__.js
+++ b/blueprints/helper/files/app/helpers/__name__.js
@@ -4,4 +4,4 @@ export function <%= camelizedModuleName %>(input) {
   return input;
 }
 
-export default Ember.Handlebars.makeBoundHelper(<%= camelizedModuleName %>);
+export default Ember.HTMLBars.makeBoundHelper(<%= camelizedModuleName %>);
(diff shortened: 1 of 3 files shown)


rwjblue:
:+1: on the change, but j think you will need to update some tests to pass the build.

maxwerr:
Ah, yes. Ill take a crack at that soon. tests for anything with 'ember generate helper foo-bar' or 'ember new' would Probably make sense to update.

maxwerr:
At a bit of a loss--everything passed locally running node 0.12.0. Maybe travis doesn't like me?

rwjblue:
Likely a transient failure.  I restarted the build.

maxwerr:
Woot! thanks

rwjblue:
Thank you!

jonnii:
@maxwerr thank you for this. it is beautiful.

maxwerr:
If it isn't the 9th best enterprise developer in all of NYC, @jonnii. Well, at 3am, when the sink is leaking, the plumber is the artist. So this is art, and you're welcome.

jonnii:
@maxwerr :heart: :heart:

stefanpenner:
> If it isn't the 9th best enterprise developer in all of NYC, @jonnii. Well, at 3am, when the sink is leaking, the plumber is the artist. So this is art, and you're welcome.

rofl

PR #3384, opened 2015-02-27
Test powershell for appveyor builds

raytiley:
If this doesn't screw anything else up it would allow us to test #3341 correctly.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
 test_script:
   # Output useful info for debugging.
   - npm version
-  - cmd: npm run-script test-all
+  - ps: npm run-script test-all
 
 # Don't actually build.
 build: off


stefanpenner:
should we test both? or.. just recommend ps

raytiley:
Should probably test both since both are included on a default system. I'll work on getting this going and fix the power shell tests. Seems like just a funky CI thing with paths.

abuiles:
@raytiley  ping

raytiley:
retrying this on appveyor just to see how it behaves.

stefanpenner:
@raytiley should we rest both ps and cmd ? or..

raytiley:
I'm not sure if it really matters. I would probably be good to test both because the environments differ in how they can format output (the ansi color codes comes to mind). But i'm hesitant to double our windows CI time until we can get the tests a bit faster.

PR #3386, opened 2015-02-27
Display localhost in console instead of 0.0.0.0.

rwjblue:


stefanpenner:
:shipit:

wmlele:
Sure I am missing something, but why display 'localhost' when in fact is binding to 0.0.0.0 (at least it does for me)?

`Serving on http://localhost:4200/`

`tcp        0      0 0.0.0.0:4200            0.0.0.0:*               LISTEN      23360/ember ` 

Might give the idea it's safe to leave it there, while in fact it is reachable from outside.

stefanpenner:
> Might give the idea it's safe to leave it there, while in fact it is reachable from outside.

firewalls are your friend

PR #3387, opened 2015-02-27
[Enhancement] Adding ember publish command

thoov:
Gives the ability to run ember publish which will publish your addon to npm.

### DIFF
--- a/lib/commands/publish.js
+++ b/lib/commands/publish.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var Command = require('../models/command');
+
+module.exports = Command.extend({
+  name: 'publish',
+  description: 'Publishes your ember-cli addon to npm.',
+  works: 'insideProject',
+
+  anonymousOptions: [],
+
+  run: function(commandOptions, rawArgs) {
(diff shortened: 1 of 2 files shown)


rwjblue:
What does this give us over just running `npm publish`?

thoov:
@rwjblue Right now nothing. This is just the same as `ember install:addon <addon_name>` is basically `npm i --save-dev <addon_name>`. I would like to make this command take a version number such as:
`ember publish 0.0.1` and it will update package.json, do a version bump commit, and publish in one step. Not sure if thats doing to much. What are your thoughts?

rwjblue:
`ember install:addon <addon_name>` does:

* `npm install --save-dev addon-name`
* `ember generate addon-name`

rwjblue:
I'm not opposed to your idea though, I just think it should be an RFC so we can flesh out what the features should be...

thoov:
@rwjblue makes sense. I will close this and we can discuss this in another place.

rwjblue:
@thoov - https://github.com/ember-cli/rfcs is perfect for this

PR #3398, opened 2015-03-02
proxy requires hash based routing

baijum:
I am not sure whether changing the documentation or setting the default `locationType` to `hash` is better.


rwjblue:
You should sill be able to use auto or history location with proxying. 

If you cannot, it is a bug.

baijum:
The proxy is not working here without changing the `locationType` value from `auto` to `hash`. Do you want me to report this issue and close this pull request ?

stefanpenner:
Are you proxying / ?  If so its working as expected.  Typically we proxy only sub paths like Api/*

baijum:
When I give the value as `hash` and `none` the the proxy works. But if I give the value as `history` or `auto` (tested from Chrome 40+ browser) it doesn't work.

PR #3401, opened 2015-03-02
[fixes #3395] Fixes accidental global pollution.

### LINKED ISSUE #3395: SilentError breaks handlebars error messages
This line in handlebars:

https://github.com/wycats/handlebars.js/blob/64ab232d723f5935e21b0e470d04dcf369986172/lib/handlebars/exception.js#L15

expects to be able to `call` the `Error` constructor and get back a value. But it looks like ember-cli is replacing `Error` with `SilentError`, and the above line ends up with `tmp` undefined, resulting in a new exception that obscures the underlying exception.

To reproduce, just drop this template into your application:

````handlebars
{{#misspelled-componen}}
{{/misspelled-component}}
````

stefanpenner:
Oh JavaScript…

screen shot 2015-03-02 at 12 06 50 pm

### DIFF
--- a/lib/errors/silent.js
+++ b/lib/errors/silent.js
@@ -12,7 +12,7 @@ function SilentError(message) {
   }
 }
 
-SilentError.prototype = Error.prototype;
+SilentError.prototype = Object.create(Error.prototype);
 SilentError.prototype.constructor = SilentError;
 
 module.exports = SilentError;


stefanpenner:
@ef4 thanks for catching this!

rwjblue:
Ugh, good catch

chadhietala:
Restarting build.

PR #3402, opened 2015-03-02
Pave the way for pod styles.

rwjblue:
Prior to this, adding `foo.css` in `app/pods/my-foo/` would throw an error, this prevents it and allows an addon to make an intelligent decision on how to deal with these files (via a registry preprocessor).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -411,13 +411,16 @@ EmberApp.prototype._filterAppTree = function() {
     return this._cachedFilterAppTree;
   }
 
-  var podPatterns = this._podTemplatePatterns();
-  var excludePatterns = podPatterns.concat([
-    // note: do not use path.sep here Funnel uses
-    // walk-sync which always joins with `/` (not path.sep)
-    new RegExp('^styles/'),
-    new RegExp('^templates/'),
-  ]);
+  var excludePatterns = [].concat(
+    this._podTemplatePatterns(),
+    this._podStylePatterns(),
+    [
+      // note: do not use path.sep here Funnel uses
+      // walk-sync which always joins with `/` (not path.sep)
+      new RegExp('^styles/'),
+      new RegExp('^templates/'),
+    ]
+  );
 
   return this._cachedFilterAppTree = new Funnel(this.trees.app, {
     exclude: excludePatterns,
@@ -543,6 +546,17 @@ EmberApp.prototype._processedTemplatesTree = function() {
 EmberApp.prototype._podTemplatePatterns = function() {
   return this.registry.extensionsForType('template').map(function(extension) {
     return new RegExp('template.' + extension + '$');
+  });
+};
+
+/**
+  @private
+  @method _podStylePatterns
+  @returns Array An array of regexp's for each known style extension.
+*/
+EmberApp.prototype._podStylePatterns = function() {
+  return this.registry.extensionsForType('css').map(function(extension) {
+    return new RegExp(extension + '$');
   });
 };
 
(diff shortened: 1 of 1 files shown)


ebryn:
Thanks @rwjblue. I'll take it over from here via an add-on! :dancers:

dfreeman:
:tada:

rtm:
Putting `style.css` in a pod directory does not throw an error for me. Who would be throwing that error for what reason? Granted there is room for additional machinery to streamline putting styles in pods, is there any downside at the moment to simply putting `@import '../pods/resource/style.scss';` in `app/styles/app.scss` if that's how you like to roll?

lib/broccoli/ember-app.js
```diff
@@ -548,6 +551,17 @@ EmberApp.prototype._podTemplatePatterns = function() {
 
 /**
   @private
+  @method _podStylePatterns
+  @returns Array An array of regexp's for each known style extension.
+*/
+EmberApp.prototype._podStylePatterns = function() {
+  return this.registry.extensionsForType('css').map(function(extension) {
+    return new RegExp(extension + '$');
```
stefanpenner: should this be `'\.' + extension + '$'`

dschmidt:
What is holding this back? Is there anything one can do to bring this over the finish line?

stefanpenner:
> What is holding this back? Is there anything one can do to bring this over the finish line?

I left a comment, it should be addressed.

Additionally, this will slow down builds yet again without further internals refactoring so I am hesitant to lob this on top.

knownasilya:
Update?

dfreeman:
Does any of the recent performance work that's been going on help mitigate the hit here?
Obviously @stefanpenner's comment on the extension regex is also still hanging out, but it would be great to see https://github.com/ebryn/ember-component-css/issues/67 close.

stefanpenner:
> Does any of the recent performance work that's been going on help mitigate the hit here?

it begins to, my current effort will negate it nearly entirely.

In addition to that, @rwjblue is working to transition the build to actually take advantage of broccoli's pipelining power. With both of these tasks together we can likely get this in.

My hope is to have this (or something that supports this) in by months end, and if time permits sooner.

rtm:
Excuse my ignorance, but I'm not getting why this is a performance or pipeline issue. Is the notion that somehow the css files in the pods need to be automatically picked up and imported in the right order? Personally, I'm find with writing my imports and actually prefer it since it gives me more control and visibility over what's going on. In `styles/app.scss`, I'd write `@import 'pods/foo/index.scss'` (or just `@import 'pods/foo/'`), and in `pods/foo/index.scss` I'd write other imports that I need; in that way, it seems that a straightforward scss compile run would do everything it needed to.

buschtoens:
@rtm Regarding import order and automatic imports: pods should be independent of each other. This implies that the order in which they are imported doesn't actually matter as each pod style should only style its respective pod and have no side-effects.

However, this is irrelevant to ember-cli, as this PR instructs the build pipeline to ignore these files and let addons take care of them.

(If I understood eveything correctly, that is.)

@stefanpenner @rwjblue What is the status on this?

Will pod styles remain part of the addon land (e.g. ember-component-css) and if so will these addons need to be updated besides the monkey patching removal?

stefanpenner:
> Will pod styles remain part

we have plans to make pod-styles part of the official pods when they land.

stefanpenner:
We will revisit once we finalize pods. i know @rwjblue and @dgeb are working to push flesh out the details. At which point, we can see how this fits in and reopen as needed.

PR #3410, opened 2015-03-03
Use correct bound helper params

jbrown:


blueprints/helper/files/app/helpers/__name__.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
-export function <%= camelizedModuleName %>(input) {
-  return input;
+export function <%= camelizedModuleName %>(params, hash, options, env) {
```
rwjblue: Bound helpers only get `params` and `hash` (the first two args).

rwjblue: You should likely put the args in comments (otherwise the JSHint tests will blow up after generating):

```javascript
export function <%= camelizedModuleName %>(/* params, hash, options, env */) {
}
```

And I think that the generate tests will also need to be updated.

jbrown: @rwjblue I'm a little confused, all the helpers I see in core have 4 params. Can you point me to some documentation on this?

rwjblue: Helpers in core are not using `Ember.HTMLBars.makeBoundHelper`, they are not bound helpers.

Documentation is here:

https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/system/make_bound_helper.js#L18-L60

jbrown:
@rwjblue Does it look okay now?

kimroen:
I think this will make the default helper test not pass: https://github.com/ember-cli/ember-cli/blob/master/blueprints/helper-test/files/tests/unit/helpers/__name__-test.js

rwjblue:
@kimroen - You are correct, mind submitting a PR?

kimroen:
@rwjblue Done :)

PR #3423, opened 2015-03-05
Update documentation to support integration tests

null-null-null:
The documentation describes using `import { module, test } from 'ember-qunit'`, but this won't work, as `ember-qunit` doesn't have a `module` function. The syntax has been updated to use plain `qunit` instead.


joostdevries:
awesome, thanks!

PR #3436, opened 2015-03-06
Update Broccoli to 0.13.6 to provide errors on new API.

rwjblue:
We are preparing to update to the next major revision of Broccoli, and 0.13.6 provides an error if we discover a `rebuild` method.

Shortly after releasing 0.2.0 of ember-cli, we need to update to use Broccoli 0.15.0, which enables the new API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "abbrev": "^1.0.5",
     "bower": "^1.3.5",
     "bower-config": "0.5.2",
-    "broccoli": "0.13.3",
+    "broccoli": "0.13.6",
     "broccoli-caching-writer": "0.5.1",
     "broccoli-clean-css": "0.2.0",
     "broccoli-es3-safe-recast": "1.0.0",


stefanpenner:
@rwjblue whats blocking us from moving to 0.15, or is the  "break all the addons" thing ?

rwjblue:
@stefanpenner - 0.15.0 adds a bunch of code to allow usage of both `read` and `rebuild`, and has not been terribly "battle tested". We need to ensure that 0.2.0 is stable (but provides good feedback to folks using addons that only have a `rebuild` function), then give some time to deal with any issues with 0.15.0 on master.

stefanpenner:
> @stefanpenner - 0.15.0 adds a bunch of code to allow usage of both read and rebuild, and has not been terribly "battle tested". We need to ensure that 0.2.0 is stable (but provides good feedback to folks using addons that only have a rebuild function), then give some time to deal with any issues with 0.15.0 on master.

sounds good.

PR #3438, opened 2015-03-06
Ensure nested addon registry matches addon order.

### LINKED ISSUE #3396: Registry ordering issue
@code0100fun and myself found a possible issue with the registry in EmberCLI not respecting the order defined in the package.json correctly.  The steps to reproduce are as follows:

1. pull down
  - [ember-cli-emblem-hbs-printer](https://github.com/rondale-sc/ember-cli-emblem-hbs-printer)
  - [ember-cli-emblem-test](https://github.com/hashrocketeer/ember-cli-emblem-test)
  - ember-cli
- `cd` into `ember-cli-emblem-hbs-printer` pull in branch named `handle_0.1.x_compat`
- npm install e'erything
- `npm link` both `ember-cli-emblem-hbs-printer` and `ember-cli` to `ember-cli-emblem-test`
- Finally, load the page.  

The test app should display the header "Winning"  but instead displays nothing.  

Things we've learned while trying to tackle this one.  

First, when we place a log the output from [registry.extensionsForType](https://github.com/ember-cli/ember-cli/blob/master/lib/models/registry.js#L35) we see that the `extensions` var is set out of order with `hbs` before `embl` and `emblem` (which shouldn't happen because of the before in the package.json)

Second, changing [registry.add](https://github.com/ember-cli/ember-cli/blob/master/lib/models/registry.js#L72) from `unshift` to `push` fixes the issue.  Which is what makes us think that this is a problem with the registry ordering. (though it'd be very unlikely for this to be the actual fix, it is just illustrative)

 It doesn't seem as if the DAG is being respected when you manually call add on the registry.

We also found that whenever we had access to `project` or `app` and could call `addons` we'd see that their order was correct despite the registry not being ordered correctly.

(we imported pluck from lodash

This is all from inside `registry.extensionsForType`  

 

With `ember-cli-emblem-hbs-printer` occurring before `ember-cli-htmlbars`.

If I can be any help with finding a resolution just let me know.  Thanks!

rwjblue:
Fixes #3396.

### DIFF
--- a/lib/models/registry.js
+++ b/lib/models/registry.js
@@ -69,7 +69,7 @@ Registry.prototype.add = function(type, name, extension, options) {
     plugin = new PluginType(name, extension, options);
   }
 
-  registered.unshift(plugin);
+  registered.push(plugin);
 };
 
 Registry.prototype.remove = function(type, name) {
--- a/lib/preprocessors.js
+++ b/lib/preprocessors.js
@@ -43,6 +43,8 @@ module.exports.setupRegistry = function(appOrAddon) {
     appOrAddon.setupPreprocessorRegistry('self', registry);
   }
   setupRegistryForEachAddon(registry, appOrAddon);
+
+  addLegacyPreprocessors(registry);
 };
 
 /**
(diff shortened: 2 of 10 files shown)


buschtoens:
Awesome, when's the next beta due? :smile:

rwjblue:
@buschtoens - It will be 0.2.0 final

buschtoens:
i m ok with this

PR #3439, opened 2015-03-06
Clarify the description of --filter option

avdv:
The given value is not interpreted as an regex, but a case-insensitive
substring search in `module.name + ": " + testName` is used.

You might negate the match using an exclamation mark as the first
letter, e.g. `--filter '!foo'` will run all tests *except* those
containing the word "foo".


avdv:
Needless to say, throwing regexes onto that option almost made me mad...

Took me quite some time looking through qunit's code to figure that out -- I haven't found any documentation for this on the QUnit site.

stefanpenner:
> Needless to say, throwing regexes onto that option almost made me mad...

ah, sorry about that. Good catch.

stefanpenner:
re-triggering appveyor, seems like a small race caused it to fail

avdv:
> ah, sorry about that. Good catch.

no worries, problem solved. Thanks!

PR #3440, opened 2015-03-06
Enforcing newlines in template files results in unwanted Nodes

jclem:
The default `.editorconfig` of Ember CLI tells an editor to append new lines to the end of files on write. This can cause surprising behavior when I do not see a new line at the end of my template file, but my rendered component has an extra `Text` node containing a new line at the end of it.

Appending a newline seems fine for any file that's not a template file, but it can be problematic when applied to templates.

### DIFF
--- a/blueprints/app/files/.editorconfig
+++ b/blueprints/app/files/.editorconfig
@@ -18,6 +18,7 @@ indent_style = space
 indent_size = 2
 
 [*.hbs]
+insert_final_newline = false
 indent_style = space
 indent_size = 2
 


lukemelia:
Maybe we should strip trailing whitespace in template compilation?

jclem:
@lukemelia That would break cases where a component's `white-space` is `pre`. If I put newlines in the template, good, but we should not have the editor append one just to make Unix happy.

gorner:
@jclem Not to impose but I have been curious whether you or anyone else can reproduce this text-nodes issue with modern versions of Ember. 

One of my colleagues has been frustrated by this setting* and has been insistent that he can't reproduce the original issue with our current version of Ember (i.e., if there's an `\n` at the end of a template it does not create an extra text node. But it does add such a node if it he adds multiple `\n`s). 

*Even if we disable the EditorConfig rule locally, it's still enforced by TemplateLint + Prettier in compliance with the blueprint default, as was noted in https://github.com/prettier/prettier/pull/6243.

NullVoxPopuli:
do you think it's worth proposing no empty newline?

or a setting on `<template>` to remove unneeded invisibles? (such as via https://github.com/mainmatter/ember-hbs-minifier)

using `<template>` in general gets you around the problem of extra new lines -- but does still have you end up with extra text nodes due to the indentation.

I've been using mainmatter's minifier on my projects, and I'm very happy with it -- I think we should built it in to `@glimmer/syntax`

gorner:
@NullVoxPopuli Thanks for the insight. We may be a ways away from fully getting on the `<template>` bandwagon (we're just now upgrading off 3.28!) but I'd forgotten about that and agree it should ultimately obviate this as a concern. And I'm sure those tools will be useful when we get there.

NullVoxPopuli:
ember-hbs-minifier supports 3.20+, so that could perhaps work for your existing stuff today?

gorner:
@NullVoxPopuli On reread, either I'm not quite grokking your solution or you've misunderstood our issue.

There has been – since this 2015 PR – a rule that already causes Prettier to complain if there's a newline at the end of a .hbs (which is contrary to how Prettier handles every other file). We think that rule is based on an outdated concern about text nodes (we can't reproduce the issue @jclem originally complained about when he submitted this PR). 

I see how a minifier would help with cleaning up empty text nodes... which would help if we were seeing these end-of-template text nodes, but we're not, even if we disable Prettier and add a closing `\n`.

Ultimately though, moving to the `<template>` tag means everything will go into the .gjs files and at that point this special Prettier handling for .hbs will no longer apply.

(For context: we'd previously disabled Prettier from the app blueprint, we recently started looking at applying it and were surprised it was trimming ending newlines only for .hbs, and enforcing them everywhere else.)

jclem:
I'm no longer in the Ember ecosystem, so I'm sorry that I can't help out here, any longer! I hope all is well in Ember-world, though 😄

PR #3446, opened 2015-03-06
Remove accidentally committed `.only`.

rwjblue:


chadhietala:
Bouncing CI.

tests/acceptance/preprocessor-smoke-test-slow.js
```diff
@@ -67,7 +67,7 @@ describe('Acceptance: preprocessor-smoke-test', function() {
       });
   });
 
-  it.only('addon registry entries are added in the proper order', function() {
+  it('addon registry entries are added in the proper order', function() {
```
chadhietala: I think line 83 needs to be:

```
return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'build', '--silent');
```

But this test still keeps exiting with 1 when ran end to end.

chadhietala: Further investigation this is exiting with 1 because build is looking for a css preprocessor.  Changing `tests/fixtures/preprocessor-tests/app-registry-ordering/app/styles/app.css` to `app.scss` allows the tests to pass.

PR #3459, opened 2015-03-07
Fixes failing preprocessor-smoke-test

chadhietala:
In the acceptance tests we broccoli-sass installed so all `app/styles` need to be scss. This is a known issue that you cannot have both `.css` and `.scss` in a project.  

Also added the `--silent` flag to be consistent with other tests and to suppress any warnings. This should fix the failing tests in #3446.


rwjblue:
:heart: - I broke the build, then promptly went offline for 24 hours.  Thank you for fixing this!

PR #3460, opened 2015-03-08
[ENHANCEMENT]: set filter param toLowerCase

williamsbdev:
When I use a filter, I have to make it all lower case as I'm assuming testem will lower case my test name from "Bar foo" to "bar foo". When I am attempting to filter and copy the test name "Bar foo", it will not find the test.

Please let me know if there is a better place or a different way to handle this scenario.

Thanks!

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -68,7 +68,7 @@ module.exports = Command.extend({
     }
 
     if (options.filter) {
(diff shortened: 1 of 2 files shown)


toranb:
+1

chadhietala:
Thanks!

PR #3463, opened 2015-03-08
disable visual progress effect in dumb terminals

jesse-black:
Disables visual progress indicators in dumb terminals such as editor buffers which don't support terminal cursor movement. They configure the environment with `TERM=dumb` and in the case of emacs it configures the environment with `TERM=emacs` in compilation mode.

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -66,7 +66,7 @@ function cli(options) {
   var ui = new UI({
     inputStream:  options.inputStream,
     outputStream: options.outputStream,
-    ci:           process.env.CI,
+    ci:           process.env.CI || /^(dumb|emacs)$/.test(process.env.TERM),
     writeLevel:   ~process.argv.indexOf('--silent') ? 'ERROR' : undefined
   });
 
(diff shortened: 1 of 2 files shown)


lib/cli/index.js
```diff
@@ -66,7 +66,7 @@ function cli(options) {
   var ui = new UI({
     inputStream:  options.inputStream,
     outputStream: options.outputStream,
-    ci:           process.env.CI,
+    ci:           process.env.CI || /^(dumb|emacs)$/.test(process.env.TERM),
```
rwjblue: Should we use `process.stdout.isTTY` instead of `process.env.TERM`?

https://nodejs.org/api/process.html#process_process_stdout
https://nodejs.org/api/tty.html#tty_tty

jesse-black: It looks like `process.stdout.isTTY` has more to do with the process pipe type than with the terminal's capabilities, because it behaves differently across platforms. On Windows `process.stdout.isTTY` is true inside the Windows console and false inside emacs eshell and compilation-mode like we might expect for it to work here, but on OSX it is true inside all of them: emacs eshell, compilation-mode, as well as a real terminal.

stefanpenner: sounds like TERM might be the more resilient solution

@rwjblue thoughts?

rwjblue: Happy to give it a go as `process.env.TERM`...

stefanpenner:
#yolo

PR #3465, opened 2015-03-08
Updates SSL protocols

danmcclain:
SSL 3 is vulnerable to POODLE attacks: https://community.qualys.com/blogs/securitylabs/2014/10/15/ssl-3-is-dead-killed-by-the-poodle-attack

Follows the advice linked to in the above article for SSL protocols: https://scotthelme.co.uk/sslv3-goes-to-the-dogs-poodle-kills-off-protocol/


rwjblue:
Thank you!

PR #3470, opened 2015-03-09
Removes SSL configuration and links to resource for configuring SSL

danmcclain:
By providing recommendations on SSL configuration (protocols and
schemas), ember-cli would need to stay on top of best practices to keep
the information up to date to provide the best possible configuration
without being vulnerable. This removes that information and links to a
great resource instead, since ember-cli is not in the business of
security


rwjblue:
Thank you!!

PR #3476, opened 2015-03-09
Removes initializer mention from service generator help text

corpulentcoffee:
As of 2ea101c, an initializer is no longer automatically created when running the `service` generator.


kimroen:
:+1:

rwjblue:
Thank you!

PR #3481, opened 2015-03-09
BUGFIX Fixes #3472 Check for 'usePods' instead of 'pod'.

### LINKED ISSUE #3472: [0.2.0] Component blueprint imports layout module regardless of pod structure
Firstly, I have set ` "usePods": "true"` in my `.ember-cli` file.

If I use `ember generate component x-foo`, it correctly generates the files in `/app/pods/components/x-foo/`, but the component class tries to import the layout from the "pod-less" structure.

```javascript
// app/pods/components/x-foo

import Ember from 'ember';
import layout from '../templates/components/x-foo`;

  layout: layout

  columns: {
    
  }
});
```

This, in turn, throws the error 
```
Uncaught Error: Could not find module `<project>/pods/components/templates/components/x-foo` imported from `<project>/pods/components/x-foo/component`
```

jankrueger:
Fixes #3472 by checking for 'options.usePods' instead of 'options.pod'

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -44,7 +44,7 @@ module.exports = {
   locals: function(options) {
     var templatePath = '../templates/components/' + stringUtil.dasherize(options.entity.name);
 
-    if (options.pod) {
+    if (options.usePods) {
       templatePath = './template';
     }
     return {
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -44,7 +44,7 @@ module.exports = {
   locals: function(options) {
     var templatePath = '../templates/components/' + stringUtil.dasherize(options.entity.name);
 
-    if (options.usePods) {
+    if (options.pod) {
       templatePath = './template';
     }
     return {
--- a/lib/commands/destroy.js
+++ b/lib/commands/destroy.js
@@ -50,6 +50,10 @@ module.exports = Command.extend({
       args: rawArgs
     };
 
+    if(this.settings && this.settings.usePods) {
+      commandOptions.pod = !commandOptions.pod;
+    }
+
     var taskOptions = merge(taskArgs, commandOptions || {});
 
     if (this.project.initializeAddons) {
--- a/lib/commands/generate.js
+++ b/lib/commands/generate.js
@@ -65,6 +65,10 @@ module.exports = Command.extend({
       args: rawArgs
     };
 
+    if(this.settings && this.settings.usePods) {
+      commandOptions.pod = !commandOptions.pod;
+    }
+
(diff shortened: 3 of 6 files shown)


rwjblue:
@trabus - :+1: / :-1: ?

trabus:
`usePods` is the wrong value to be checking for. The `options.pod` value is passed on from the generate command, and should be present in locals, if it's not that's a bug that needs to be investigated. Also looking at the issue this is fixing, the blueprint importing the layout in a regular project is wrong (I believe I misunderstood the initial spec when implementing), and should be fixed as part of the issue.

trabus:
@jankrueger So the fix is actually just as simple, but belongs in the blueprint model [here](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L341). When the value for `usePods` is present, we're inverting the value of `options.pod`, but not setting the value back on `options.pod`. We need to fix the `options.pod` value to be the correct value when `usePods` is true.

We have two options.
First would be to mutate the `options.pod` value from the blueprint model (quickest option)
```js
// lib/models/blueprint.js install and destroy methods
var pod = (options.settings && options.settings.usePods) ? !options.pod : options.pod;
this.pod = options.pod = pod;
```

The second would be to check for `usePods` earlier in the generate and destroy commands, and set the value there. (cleaner option, handled upstream, preferred)
```js
// lib/commands/generate.js and lib/commands/destroy.js
 if(this.settings && this.settings.usePods) {
   commandOptions.pod = !commandOptions.pod;
 }
 var taskOptions = merge(taskArgs, commandOptions || {});

// lib/models/blueprint.js install and destroy methods
this.pod     = options.pod;
```

jankrueger:
@trabus Thanks for clarification! 
There seems to be a test in place that is failing. I will fix that hopefully tonight with this pr.

trabus:
@jankrueger The existing tests should be fine after this fix, just be sure to revert the change in the component blueprint. I was suggesting we add an additional generate test with `usePods` set.

Regarding additional tests, you should be able to copy [this one](https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/pods-generate-test.js#L109), and change the values being passed. You'll just need to adjust what the expected output should be.

jankrueger:
@trabus I implemented your preferred version and changed the commandOptions in the generate/destroy command. The failing test is now passing and i added two more tests as suggested.

trabus:
@jankrueger Looking good! :+1: 

Can @rwjblue or @stefanpenner review, please?

rwjblue:
LGTM

rwjblue:
@jankrueger / @trabus - Thanks for working together on this!

PR #3488, opened 2015-03-10
ENHANCEMENT: update-checker.js should use environment http_proxy if detected

xomaczar:
I am working behind corporate proxy and every command I provide to ember-cli (except test & update) result in an error without ability to proceed/recover.

### DIFF
--- a/lib/models/update-checker.js
+++ b/lib/models/update-checker.js
@@ -66,7 +66,7 @@ UpdateChecker.prototype.doCheck = function() {
 UpdateChecker.prototype.checkNPM = function() {
   // use npm to determine the currently availabe ember-cli version
   var loadNPM = Promise.denodeify(require('npm').load);
-  var getEmberCliVersion = function(npm){
+  var fetchEmberCliVersion = function(npm){
     return Promise.denodeify(npm.commands.view)(['ember-cli', 'version']);
   };
   var extractVersion = function(data) { return Object.keys(data)[0]; };
@@ -77,9 +77,14 @@ UpdateChecker.prototype.checkNPM = function() {
       'cache-min': 1,
       'cache-max': 0
     })
-    .then(getEmberCliVersion)
+    .then(fetchEmberCliVersion)
     .then(extractVersion)
-    .then(this.saveVersionInformation.bind(this));
+    .then(this.saveVersionInformation.bind(this))
+    .catch(function(){
+      this.ui.writeLine(chalk.red('An error occurred while checking for Ember CLI updates. ' +
+        'Please verify your internet connectivity and npm configurations.'));
+      return false;
+    }.bind(this));
 };
 
 UpdateChecker.prototype.saveVersionInformation = function(version) {
--- a/lib/models/update-checker.js
+++ b/lib/models/update-checker.js
@@ -64,25 +64,32 @@ UpdateChecker.prototype.doCheck = function() {
 };
 
 UpdateChecker.prototype.checkNPM = function() {
-  // make an http call to npm to get the latest version
-  var http = require('http');
-  var concat = require('concat-stream');
+  // use npm to determine the currently availabe ember-cli version
+  var npm = require('npm'),
+      npmViewConfig = {
+        'loglevel': 'silent',
+        'fetch-retries': 1,
+        'cache-min': 1,
+        'cache-max': 0
+      };
 
   return new Promise(function(resolve, reject) {
-    http.get('http://registry.npmjs.org/ember-cli/latest', function(res) {
-      res.setEncoding('utf8');
-
-      res.pipe(concat(function (data) {
-        try {
-          resolve(JSON.parse(data).version);
-        } catch(error) {
-          reject(error);
-        }
-      }));
-
-      res.on('error', reject);
+    npm.load(npmViewConfig, function(err) {
+      if (!err) {
+        npm.commands.view(['ember-cli', 'version'], function(err, data) {
+          if (!err) {
+            resolve(Object.keys(data)[0]);
+          }
+          else {
+            reject(err);
+          }
+        });
+      }
+      else {
+        reject(err);
+      }
     });
-  }).then(function(version) {
+  }.then(function(version) {
     // we only want to save the version information when we check NPM
     return this.saveVersionInformation(version);
   }.bind(this)).catch(function(error) {
--- a/lib/models/update-checker.js
+++ b/lib/models/update-checker.js
@@ -65,37 +65,28 @@ UpdateChecker.prototype.doCheck = function() {
 
 UpdateChecker.prototype.checkNPM = function() {
   // use npm to determine the currently availabe ember-cli version
-  var npm = require('npm'),
-      npmViewConfig = {
-        'loglevel': 'silent',
-        'fetch-retries': 1,
-        'cache-min': 1,
-        'cache-max': 0
-      };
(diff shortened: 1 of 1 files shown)


lib/cli/index.js
```diff
@@ -53,7 +53,8 @@ function cli(options) {
   };
 
   var defaultUpdateCheckerOptions = {
-    checkForUpdates: true
+    checkForUpdates: true,
+    httpProxy: process.env['http_proxy']
```
stefanpenner: our convention is all caps for env vars.

@rwjblue r?

lib/models/update-checker.js
```diff
@@ -69,24 +69,47 @@ UpdateChecker.prototype.checkNPM = function() {
...
         try {
+
           resolve(JSON.parse(data).version);
         } catch(error) {
           reject(error);
         }
       }));
-
       res.on('error', reject);
+    }).on('error', function(error){
```
stefanpenner: the enclosing callback is not needed, `on('error', resolve)` should be sufficient.

lib/cli/index.js
stefanpenner: although the npm settings seem to be lower-case.  Is it possible for us to read these settings via npm as a lib?

stefanpenner:
@xomaczar CI seems to not agree with this PR yet.

lib/cli/index.js
xomaczar: Yes, agree. I'll check into that.

lib/models/update-checker.js
xomaczar: +1

lib/models/update-checker.js
```diff
@@ -69,24 +69,47 @@ UpdateChecker.prototype.checkNPM = function() {
...
+        reject(error);
     });
-  }).then(function(version) {
+
+  }.bind(this)).then(function(version) {
```
xomaczar: this one? I am accessing this.settings.httpProxy, although it can be moved out where bind(this) will not be needed.

lib/models/update-checker.js
stefanpenner: on('error')

lib/models/update-checker.js
xomaczar: @stefanpenner I ran `npm test` as described in docs before making this PR. All tests were passing.

stefanpenner: `npm test` is only the fast tests, `npm run test-all` includes the slow acceptance tests.

stefanpenner: updated docs: https://github.com/ember-cli/ember-cli/blob/master/CONTRIBUTING.md

xomaczar: Great

xomaczar:
@stefanpenner I need help, I ran npm run test-all, everything passed, except 2 pending. I squashed my changes, but PR does not reflect that.

rwjblue:
This is failing because the latest testem version (0.7.4) is missing a dependency.  I submitted https://github.com/airportyh/testem/pull/515 to fix this upstream.

johanneswuerbach:
Testem 0.7.5 includes the fix, sorry about that.

stefanpenner:
> missing }).then(....

where?

lib/models/update-checker.js
```diff
@@ -64,23 +64,30 @@ UpdateChecker.prototype.doCheck = function() {
...
-      res.pipe(concat(function (data) {
-        try {
-          resolve(JSON.parse(data).version);
-        } catch(error) {
-          reject(error);
-        }
-      }));
-
-      res.on('error', reject);
+    npm.load(npmViewConfig, function(err) {
```
stefanpenner: we can use `denodeify` https://github.com/ember-cli/ember-cli/blob/master/lib/ext/promise.js#L13-L21
docs: https://github.com/tildeio/rsvp.js/blob/master/lib/rsvp/node.js#L73-L201

xomaczar: Promisey it is. Update to follow

stefanpenner: <3

stefanpenner: let me know :)

xomaczar:
@stefanpenner would u like for me to squash those 3 commits

lib/models/update-checker.js
```diff
@@ -64,31 +64,29 @@ UpdateChecker.prototype.doCheck = function() {
...
+      'cache-max': 0
+    })
+    .then(getEmberCliVersion)
+    .then(extractVersion)
+    .then(function(version) {
+      // we only want to save the version information when we check NPM
+      return this.saveVersionInformation(version);
+    }.bind(this))
+    .catch(function(error) {
+      this.ui.writeLine('There was an error checking NPM for an update: ' + error);
```
stefanpenner: we have a this.ui.writeError, but the throw should handle that already, i suspect only the error.message is needed her.

stefanpenner:
@xomaczar not a big deal, i left one suggestion, then i think we are good to go.

lib/models/update-checker.js
xomaczar: @stefanpenner you mean just this.ui.writeLine(e.message)? I haven't looked at the calling code, but according to your statement we're catching all rejection upstream, do we even need catch here?

stefanpenner: https://github.com/ember-cli/ember-cli/blob/master/lib/ui/index.js#L84-L93

xomaczar: Why are we trying to catching those errors/rejects at the level https://github.com/xomaczar/ember-cli/blob/master/lib/models/update-checker.js#L86-L88.
The only way any of the those promises get rejected if an error gets thrown, we don't explicitly reject() anything in any then(...). Let's just https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L98 catch and log it.

stefanpenner: yes lets doit!

xomaczar:
@stefanpenner There is one case where updatechecker will still throw exception, therefore stopping any command execution with `getaddrinfo` error. 
if settings.checkForUpdates && configstore/ember-cli-version:lastVersionCheckAt > 1day && npm cache clean && no internet connection.
updateChecker should fail in that case, report the issue to the user "Failed to check for updates" and continue execution. Updates notification is nice to have, but when it block people from running their app, it become a nuisance! Thoughts?

stefanpenner:
@xomaczar so just log and message and more on if that is the error we see? Actually, maybe this should never re-throw and always just log and error but handle the rejection. It does seem silly that updateChecker could stop someone from developing.

stefanpenner:
cc @twokul

xomaczar:
Agree just catch log, but don't re-throw, tne promise gets resolved, user sees the issue but able to continue...Hopefully last change

stefanpenner:
@xomaczar thanks for working through all these changes with us :)

twokul:
@xomaczar @stefanpenner I assume this got resolved?

xomaczar:
@stefanpenner happy to be involved. Try to push it in tonight.
@twokul one more commit and we should be good to go

stefanpenner:
> @stefanpenner happy to be involved. Try to push it in tonight.

:+1: we are happy to have you involved, feel free to share feedback to help us all make contributing better.

xomaczar:
@stefanpenner ready for another review
cc @twokul

stefanpenner:
:+1:

PR #3500, opened 2015-03-11
Document `/public` directory behavior

mattconstantine:
Ember CLI very helpfully includes anything in `/public` in the consuming application's tree. However, this behavior isn't documented anywhere that I could find. After much fiddling with the relatively undocumented `treeForPublic` hook, I finally found this. Let's make sure it's super clear to everyone how to properly include static files.

Related, `ember addon` doesn't include `/public`, but that's another issue for another time perhaps.


stefanpenner:
:+1:

PR #3508, opened 2015-03-11
Update ember-cli-app-version

btecu:
Fixes 'Library "App" is already registered with Ember' https://github.com/emberjs/ember.js/issues/10600


rwjblue:
:+1:

rwjblue:
The issue with testem was causing the build to fail here, I just restarted them.

PR #3515, opened 2015-03-12
[BUGFIX] Fixes nested component generation in addons with correct relative path for template import

### LINKED ISSUE #3514: Issue with Component Generator using a Nested Path with Ember Addons
Ember CLI isn't smart about paths when generating components in addons and generating the import statement for the layout template, it seems. 

When you do, for example, `ember g component something/other-thing`, it will generate `import layout from '../templates/components/something/other-thing';`, however the actual relative path to the layout file is `'../../templates/components/something/other-thing';`

Ember CLI version 0.2.0, with Ember 1.11.0-beta.4

trabus:
This PR fixes #3514

Generating nested components inside of an addon was not taking into account the path depth when generating the template import path.

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -47,9 +47,14 @@ module.exports = {
     var contents       = '';
     // if we're in an addon, build import statement
     if (options.project.isEmberCLIAddon()) {
-      templatePath = options.pod ? './template' : '../templates/components/' + stringUtil.dasherize(options.entity.name);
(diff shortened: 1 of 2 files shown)


trabus:
I have no idea why Appveyor hates me.

rwjblue:
You need to rebase, it will update AppVeyor to use Node 0.12.

trabus:
@rwjblue Appveyor feels like a slot machine sometimes. Finally got it to run without timing out on a random test.

stefanpenner:
:+1:

PR #3518, opened 2015-03-12
[BUGFIX] Add missing bind when server already in use

bdvholmes:
See 43a8a6d4dbf82a72ca574e4d53b49edca602d888


bdvholmes:
Fixes #3519

rwjblue:
ugh, thank you! Why didn't our tests catch this?

rwjblue:
I thought https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/server/express-server-test.js#L122-L139 was testing this directly...

bartocc:
I just stumbled on this on as well.
+1 for the missing bind

PR #3538, opened 2015-03-15
Fix dumb typo

felixrieseberg:
I seriously had a typo in the installation instructions.


rwjblue:
Thanks!

stefanpenner:
@felixrieseberg any thoughts on integrating this with [ember-cli-deploy](github.com/ember-cli/ember-cli-deploy) ?

felixrieseberg:
You really only need this for Azure Websites (it's a simple deploy runtime for Node/Python/NET apps and intended for people who just want to quickly set up a small website). The script enables you to just connect your Azure Website with your GitHub repo and always have the latest version up. In a way, it's for people who don't want to configure or think about deployments too much.

ember-cli-deploy is awesome (and has a neat Azure provider plugin), but I think it's for a more sophisticated audience... If you're thinking about CDNs and asset pipelines, you're probably going with the "bigger" Azure offerings ;)

Thoughts?

lukemelia:
I would like ultimately for ember-cli-deploy to be simple enough to use that you would use it to deploy in the overwhelming majority of cases. I guess this section would become more of a listing of ember-cli-deploy plugins.

PR #3539, opened 2015-03-15
add tmp dir to npmignore

ahmadsoe:
I see some developer publish ember-cli addon with their tmp directory, make the addon's size pretty big. This PR will prevent that.


rwjblue:
:+1:

PR #3542, opened 2015-03-15
use local bower for CI, rather then installing it twice

stefanpenner:


rwjblue:
I think we need to test this in a real app, as our tests do not directly use this file...

stefanpenner:
it works

stefanpenner:
maybe not: https://travis-ci.org/stefanpenner/ember-jobs/jobs/54480422 looking shortly

stefanpenner:
ug, thet validations prevented it..

Issue #3458, opened 2015-03-07
Improve watchman error text


felixbuenemann:
ember-cli complains about watchman not being installed, without providing helpful steps to fix this:
```bash
ember -v
version: 0.2.0-beta.1
Could not find watchman, falling back to NodeWatcher for file system events
node: 1.4.3
npm: 2.1.8
```

As an ember newb this is very confusing. What is watchman? Why wasn't it installed by npm? So my first try was `npm install -g watchman`, which worked, but installed a completely different watchmen. Now ember-cli complained that watchman was found, but incompatible:
```bash
ember -v
version: 0.2.0-beta.1
Looks like you have a different program called watchman, falling back to NodeWatcher
node: 1.4.3
npm: 2.1.8
```

My thoughts: Hmm, maybe I need an older version of the npm package, but which one?

Only after some more googling I finally found out that I was supposed to install watchman via homebrew.

**So why not put a link to the watchman homepage or installation instructions into the message?**

Example:
```bash
ember -v
version: 0.2.0-beta.1
Could not find watchman, falling back to NodeWatcher for file system events
For better performance install watchman from https://facebook.github.io/watchman/
node: 1.4.3
npm: 2.1.8
```

felixbuenemann:
I'm happy to come up with a PR, if there is consensus about what the message should look like.

rwjblue:
@felixbuenemann - Agreed on both points: 

* only show on platforms that are supported by watchman
* make the message actionable via helpful link and description

stefanpenner:
we should link to something we control

* http://ember-cli.com/watchman

it should likely contain:

* what is watchman
* how to get watchman
* troubleshooting watchman

felixbuenemann:
@stefanpenner Where are the sources for the ember-cli website or who is responsible for updating it?

stefanpenner:
@felixbuenemann https://github.com/ember-cli/ember-cli/tree/gh-pages currently @joostdevries and @chnn have been stewarding it, and working on a refresh.

felixbuenemann:
@stefanpenner For now I've added a small section on watchman to the getting started section (see PR #3543). I think we shouldn't duplicate any info that is just as well covered by the watchman github page.

stefanpenner:
@felixbuenemann thanks, we will keep this open until we link from the error to this.

felixbuenemann:
@stefanpenner See #3544 for those changes.

rwjblue:
Fixed by #3544

PR #3543, opened 2015-03-15
Add a new Getting Started section for watchman

felixbuenemann:
This documents what watchman is and where is can be found.

The related discussion on this topic can be found in issue #3458.


felixbuenemann:
There's also an unrelated tool in npm that is also called watchman. I should probably add a short note about that.

felixbuenemann:
Revised to include version requirement and note on conflicting npm package.

rwjblue:
:+1: - Thank you!

PR #3544, opened 2015-03-15
Add links with watchman info to cli output

felixbuenemann:
This depends on the watchman getting started section from PR #3543 and resolves issue #3458.

I've also added some punctuation, because IMHO it looked weird without.


rwjblue:
:+1:

felixbuenemann:
I'll look into the test failures. It seems some tests are checking the number of lines returned.

stefanpenner:
:+1:

Issue #3424, opened 2015-03-05
[Proposal] Remove `podModulePrefix` from ember-cli


trabus:
As a step towards using [pods as the default structure](https://github.com/ember-cli/ember-cli/issues/2646) for ember-cli, I believe we should get rid of the `podModulePrefix` and just put pods in the app folder. Because the value is customizable, it causes a break in our convention when looking up modules. Most people use the default `pods` anyhow, but the fact that it is customizable means that looking up a pod in your project could be different than how developers [structure pods in their addon](#2768).

rwjblue:
I **hate** podModulePrefix. I'm definitely 👍 for removing it (with a deprecation period).

twokul:
this is nice.

trabus:
I'm going to start work on this likely next week. First I'm going to try and get ember-cli-podify finished, and add support to remove the `podModulePrefix` through a flag `ember podify --noprefix`.

I'm also going to finally get a dev blog setup and will write up a post on how to use ember-cli-podify to update your project so you don't have to do it manually (though I'll include manual instructions as well for those who'd rather do it themselves).

trabus:
I got a bit derailed with replacing marked with markdown-it, but should have some time to pick this back up tomorrow. I'l like to get the deprecation warning in as soon as possible, preferably in the next release.

I'll write up some manual instructions on what needs to be done to move existing files away from using podModulePrefix, since I'm not sure I'll have much time to get the podify addon done before next week.

rwjblue:
@trabus - Deprecation (but still functionally the same) for a version sounds exactly correct.

rwjblue:
The deprecation has landed in ember-resolver (in https://github.com/ember-cli/ember-resolver/pull/88).  

Now ready to update the `generate`/`destroy` commands and bump the included version of the `ember-resolver` to 0.1.13.

@trabus - Thanks for keeping on this..

SaladFork:
I was concerned about this deprecation warning as we have a "services" route and with pods being moved to the `app` folder it seemed we would have to put the JS files alongside Ember services, such as:

```
// Before:

app/pods/services/route.js
app/pods/services/controller.js
app/services/geolocation.js

// After:

app/services/route.js
app/services/controller.js
app/services/geolocation.js
```

Similar cases exist if you had a "templates" or "styles" route, for example.

[...]

PR #3546, opened 2015-03-16
[ENHANCEMENT] Add podModulePrefix deprecation for generate and destroy commands

trabus:
This PR fulfills #3424 by adding deprecation warnings for generate and destroy commands, as well as bumping the ember-resolver version to 0.1.14.

The deprecation warnings were not using the ui, so I added an additional version of deprecate that allows you to pass the ui in. I also changed `cli.run()` so it returns an object containing the ui so we can now check the ui inside of acceptance tests. So now inside any ember-cli acceptance test you need to check the ui output by doing the following:
```js
it('outputs ui text', function() {
  ember(['new','my-app']).then(function(result){
    expect(result.ui.output).to.include('new');
  });
});
```

### DIFF
--- a/bin/ember
+++ b/bin/ember
@@ -28,6 +28,7 @@ resolve('ember-cli', {
     inputStream: process.stdin,
     outputStream: process.stdout
   }).then(function(result) {
-    exit(result && result.exitCode || result);
+    var exitCode = typeof result === 'object' ? result.exitCode : result;
+    exit(exitCode);
   });
 });
--- a/bin/ember
+++ b/bin/ember
@@ -27,5 +27,7 @@ resolve('ember-cli', {
     cliArgs: process.argv.slice(2),
     inputStream: process.stdin,
     outputStream: process.stdout
-  }).then(exit);
+  }).then(function(result) {
+    exit(result && result.exitCode || result);
+  });
 });
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -4,7 +4,7 @@
     "jquery": "^1.11.1",
     "ember": "1.10.0",
     "ember-data": "1.0.0-beta.15",
-    "ember-resolver": "~0.1.13",
+    "ember-resolver": "~0.1.14",
     "loader.js": "ember-cli/loader.js#3.2.0",
     "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.3",
     "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -4,7 +4,7 @@
     "jquery": "^1.11.1",
     "ember": "1.10.0",
     "ember-data": "1.0.0-beta.15",
-    "ember-resolver": "~0.1.12",
+    "ember-resolver": "~0.1.13",
     "loader.js": "ember-cli/loader.js#3.2.0",
     "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.3",
     "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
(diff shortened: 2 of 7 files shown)


blueprints/app/files/bower.json
```diff
@@ -4,7 +4,7 @@
     "jquery": "^1.11.1",
     "ember": "1.10.0",
     "ember-data": "1.0.0-beta.15",
-    "ember-resolver": "~0.1.12",
+    "ember-resolver": "~0.1.13",
```
rwjblue: Had to release 0.1.14 to fix an issue, can you bump to that?

rwjblue:
The message says breaking, but this doesn't seem to actually break anything (but rather it deprecates and keeps everything functional). Am I mis-interpreting something?

trabus:
@rwjblue You're right, I'll remove 'breaking'.

lib/cli/cli.js
```diff
@@ -86,14 +86,21 @@ CLI.prototype.run = function(environment) {
       // Possibly this issue: https://github.com/joyent/node/issues/8329
       // Wait to resolve promise when running on windows.
       // This ensures that stdout is flushed so acceptance tests get full output
+      var result = exitCode;
+      if (this.testing) {
```
trabus: @rwjblue @stefanpenner I just wanted to make sure this was okay before you merge, it doesn't seem to be breaking anything by returning this object instead of just the exitCode.

lib/cli/cli.js
```diff
@@ -86,14 +86,21 @@ CLI.prototype.run = function(environment) {
...
+        result = {
```
trabus: @rwjblue @stefanpenner I just wanted to make sure this was okay before you merge, it doesn't seem to be breaking anything by returning this object instead of just the exitCode.

rwjblue:
LGTM

trabus:
Ugh, random AppVeyor timeout, I don't know why it hates me so.

rwjblue:
restarted build

trabus:
Looks like more random timeouts.

rwjblue:
LGTM

@stefanpenner - r?

bin/ember
```diff
@@ -27,5 +27,7 @@ resolve('ember-cli', {
     cliArgs: process.argv.slice(2),
     inputStream: process.stdin,
     outputStream: process.stdout
-  }).then(exit);
+  }).then(function(result) {
+    exit(result && result.exitCode || result);
```
rwjblue: Actually, I think we need to handle falsey exit codes `0` is an exit code and is falsey.

Something like:

```javascript

exit(exitCode);
```

bin/ember
```diff
@@ -27,5 +27,8 @@ resolve('ember-cli', {
     cliArgs: process.argv.slice(2),
     inputStream: process.stdin,
     outputStream: process.stdout
-  }).then(exit);
+  }).then(function(result) {
+    var exitCode = typeof result === 'object' ? result.exitCode : result;
```
stefanpenner: I'm nervous to make changes in this file.

rwjblue: This file's change should be perfectly backwards compatible (we allow for `result` to either be an object (new thing) or a number (old thing)).  We should add a comment saying why we are doing the ternary (and what version it was added in so we know when we can remove it).

@stefanpenner - In theory we could avoid changing this file, but then we could not test the UI output in acceptance tests...

trabus: I was too, but providing access to the ui for testing is definitely worth it IMO. This change essentially makes returning an object possible while still returning the exitCode.

trabus: I'll gladly add a comment explaining what this does and why, with reference to cli.js

stefanpenner: > I was too, but providing access to the ui for testing is definitely worth it IMO. This change essentially makes returning an object possible while still returning the exitCode.

I believe this is in-fact worth it. I just wanted to be confident that this was thoroughly taken into account. I am happy with this.

trabus: @stefanpenner did you want me to put in the comment, or is this good as-is?

stefanpenner: good as-is.

trabus:
@rwjblue @stefanpenner I think this is ready to go. Let me know if you want anything else done on it.

trabus:
@rwjblue I'll get a PR going to reflect this in the docs to go with the next release. When are we planning to release 0.2.1?

PR #3550, opened 2015-03-16
Add ability to start ember serve on https

drogus:
Sometimes it may be needed to run an app locally with SSL. This PR
adds an --ssl option, which tries to run on https. Default ssl
certificate and ssl key paths are "ssl/server.crt" and "ssl/server.key"
respectively. Custom paths can be added with --ssl-cert and --ssl-key.

Tests don't pass, because I suck at node and npm and I have no idea how to run only
one file (and running all tests is too slow to debug it easily). Additionally I'm not sure
how to [pass ssl options to `subject` in the test](https://github.com/ember-cli/ember-cli/commit/f03d95d103c51f2b1a684781e1f27e75ac5992ef#diff-b60aaf46a4c72af85684ba60f4a134edR89). They're used in [`setupHttpServer`](https://github.com/ember-cli/ember-cli/commit/f03d95d103c51f2b1a684781e1f27e75ac5992ef#diff-efca384df22e0183d0dcd6230de0bd9dR41), but the test error says it's `undefined`. The weird thing is that `ssl` option apparently is passed through correctly, because it reaches [a check for `sslKey`](https://github.com/ember-cli/ember-cli/commit/f03d95d103c51f2b1a684781e1f27e75ac5992ef#diff-efca384df22e0183d0dcd6230de0bd9dR41). Any help is appreciated.


stefanpenner:
I like the idea, i would like to play with it myself to see how it feels.
We would likely need to add some docs

stefanpenner:
@drogus got time to get tests green?

stefanpenner:
@drogus friendly ping

tests/unit/tasks/server/express-server-test.js
```diff
@@ -81,6 +81,21 @@ describe('express-server', function() {
   });
 
   describe('output', function() {
+    it('with ssl', function() {
+      return subject.start({
+        host:  '0.0.0.0',
+        port: '1337',
+        ssl: true
```
stefanpenner: missing comma

tests/unit/tasks/server/express-server-test.js
```diff
@@ -140,6 +155,24 @@ describe('express-server', function() {
...
+      return subject.start({
+        host:  'localhost',
+        port: '1337',
+        ssl: true,
+        sslCert: 'tests/fixtures/ssl/server.crt',
+        sslKey: 'tests/fixtures/ssl/server.key'
+      })
+        .then(function() {
+          request.get('/').expect(200, function(err){
+            console.log(err);
```
stefanpenner: if this fails or is async it will diverge from the async of the test

stefanpenner:
> Tests don't pass, because I suck at node and npm and I have no idea how to run only
> one file (and running all tests is too slow to debug it easily).

select tests:

`describe` -> `describe.only` (now is selectively run)

debugging:

```js
npm run test:debug
```
https://github.com/ember-cli/ember-cli/blob/master/package.json#L13

tests/unit/tasks/server/express-server-test.js
stefanpenner: as this is a self signed certificate this here errors, is that what we expect?

stefanpenner:
for some reason GH wouldn't let me PR to your fork, here is a branch i have rebased and made some fixes too: https://github.com/ember-cli/ember-cli/tree/pr-3550

- [ ] SSL cert is self-signed so the request test fails (as expected) But i wasn't sure your intent
- [ ] someone output seems to be incorrect

drogus:
I'm sorry for a delay, my notifications setup was screwed up. I'll work on it today, thanks for tips, that should get me going!

stefanpenner:
> I'm sorry for a delay, my notifications setup was screwed up. I'll work on it today, thanks for tips, that should get me going!

:)

tests/unit/tasks/server/express-server-test.js
```diff
@@ -144,6 +158,30 @@ describe('express-server', function() {
...
+        host:  'localhost',
+        port: '1337',
+        ssl: true,
+        sslCert: 'tests/fixtures/ssl/server.crt',
+        sslKey: 'tests/fixtures/ssl/server.key',
+        baseURL: '/'
+      })
+        .then(function() {
+          return new Promise(function(resolve, reject) {
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
```
drogus: @stefanpenner the intention of this test was to check if we really start server over https, so I wanted to make a request that succeeds. I found out that the easiest way is to set `process.env.NODE_TLS_REJECT_UNAUTHORIZED` to 0, to not get an error without setting `ca`.

drogus:
It should be green now. Io.js tests fail, but I see them failing also on other branchs/prs, so that's rather not caused by my code.

stefanpenner:
the io.js failure appears different then the rest. The rest are actually the same test, that just happens to be brittle on travis.  I am unable to reproduce it locally, on linux or OS X. :(

drogus:
I see the same issue here: https://travis-ci.org/ember-cli/ember-cli/jobs/61259446#L1843-L1848

Anyways, I can have a look at that as well.

stefanpenner:
ya hmm.. 17 hours ago we got a green build: https://travis-ci.org/ember-cli/ember-cli/builds/61205147 which should be master. It just take a-few runs.

stefanpenner:
@jayphelps looks like you appveyor has hijacked ember-cli's ?

https://ci.appveyor.com/project/jayphelps/ember-cli/build/230

stefanpenner:
some strange interleavings are happening with appveyor. https://ci.appveyor.com/project/embercli/ember-cli/build/2643 seems to be green.

jayphelps:
@stefanpenner I don't even

drogus:
@stefanpenner it seems that tests pass after rebasing to master. Should I squash commits? Does the code look OK?

lib/tasks/server/express-server.js
```diff
@@ -36,7 +38,21 @@ module.exports = Task.extend({
   },
 
   setupHttpServer: function() {
-    this.httpServer = this.http.createServer(this.app);
+    if (this.startOptions.ssl) {
+      if(!fs.existsSync(this.startOptions.sslKey)) {
+        throw new TypeError('Ssl key couldn\'t be found in "' + this.startOptions.sslKey + '", please provide a path to an existing ssl key file with --ssl-key');
```
stefanpenner: `Ssl` -> `SSL`

lib/tasks/server/express-server.js
```diff
@@ -36,7 +38,21 @@ module.exports = Task.extend({
   },
 
   setupHttpServer: function() {
-    this.httpServer = this.http.createServer(this.app);
+    if (this.startOptions.ssl) {
+      if(!fs.existsSync(this.startOptions.sslKey)) {
```
stefanpenner: i think this should be `this.httpServer = this.startSSLHttpServer`...

stefanpenner:
* left some small nitpick comments
* coveralls appears to think we need a bit more test coverage...

drogus:
@stefanpenner I'm not sure what's going on with coveralls, but after my last push (I extracted https server creation and fixed typos) it's green

PR #3562, opened 2015-03-18
dist can be watched, it really is just tmp that matters. This prevents p...

stefanpenner:
...eople from deploying this strange file by mistake.


rwjblue:
:heart:

Issue #3556, opened 2015-03-18
ES6 is not transpiled in addon


aceofspades:
Per https://github.com/babel/ember-cli-babel/issues/13#issuecomment-81176286 it was stated that ES6 in addons should be transpiled the same as with the hosting app.

This example has a simple component definition in both the app and addon, with an ES6 fat arrow function. For the app, it is transpiled, but not in the addon.

To see the issue:

```
git clone git@github.com:aceofspades/es6-in-addon.git
cd es6-in-addon/alpha
ember install
ember build
```

In `alpha/dist/assets/alpha.js` the function is transpiled:
```
double: function (v) {
  return v * 2;
}
```

In `alpha/dist/assets/vendor.js` it is not.
```
double: (v => v * 2)
```

stefanpenner:
Addons the addon needs ember-cli-babel

rwjblue:
https://github.com/aceofspades/es6-in-addon/blob/master/alpha/package.json#L25  needs to be in deps, not dev deps.

jayphelps:
@rwjblue That's the default place it's put when `$ ember addon foo` since it shares the package.json from app blueprint. This just bit me as well, lost 2+ hours trying to figure out why addon tests were failing/hanging was because they weren't getting transpiled. This isn't obvious--should we have this added for addons automatically? Seems like a bug that ember apps have babel by default but addons don't.

rwjblue:
@jayphelps -  This is fundamentally how npm works, if you have a devDep it is installed when working on the project/lib but not when it is used as a dependency of something else. 

I am not opposed to shipping babel in addons by default just like we do with apps.

stefanpenner:
> I am not opposed to shipping babel in addons by default just like we do with apps.

it feels like we actually have 2 types of addons,

1. small addon, that provides 1 simple thing (maybe a command for ember-cli, or jshinting)
2. an addon that provides addon/ and app/

these seem to have different dep requirements, it would be strange for 1 to have babel as a dep, but makes total sense for 2

thoughts?

jschilli:
+1 wonder if having sub blueprints for build-addon and ?? functional-addon (terrible name) would make sense

stefanpenner:
the UX feels crappy around this, needs some exploration.

aceofspades:
Thanks for the help.

I would vote for consistent UX app vs. addon, maybe an option `--lightweight` when creating an addon for those that are concerned about this. 

[...]

PR #3564, opened 2015-03-18
Added babel to addons package.json dependencies by default

jayphelps:
From discussions over at #3556, I don't see any real harm in giving this to addons by default since they don't *have* to use babel syntax and if they truly care, they can simply remove it from them package.json, just like many do for content-security-policy, among others. Good defaults--that's the "ember way"

I think creating options/flags is overcomplicating things. At least as a first pass. If someone wants to add a flag to exclude it, that's cool sugar.

*I would love to write a test for this, but would require some decent refactoring and/or using [mock-fs](https://www.npmjs.com/package/mock-fs) or something. The current addon package.json isn't being tested, I vote get this out then I have cycles to pair with Robert or Stef to refactor so we can test these sort of things, if they'd like.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -19,13 +19,17 @@ module.exports = {
     contents.name = this.project.name();
     contents.description = this.description;
     contents.keywords = contents.keywords || [];
+    contents.dependencies = contents.dependencies || {}
+    // npm doesn't like it when we have something in both deps and devDeps
+    // and dummy app still uses it when in deps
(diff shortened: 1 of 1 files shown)


rwjblue:
I'm happy with this, but leaving ember-cli-babel in both deps and devDeps generally triggers a warning by NPM.

I *think* that if we remove it from devDeps, the dummy app will still work as expected, can you confirm (then fix)?

jayphelps:
@rwjblue confirmed. Dummy still gets babel when in deps-only. Made the change. :+1:

aceofspades:
Thanks @jayphelps

Issue #2957, opened 2015-01-12
baseUrl not set in built index.html if locationType is hash


jyriand:
I have configuration:
```
baseURL: '/ember/'
locationType: 'hash'
```
But running 'ember build' produces index.html file without base tag. If instead of 'hash' I have 'auto' then everything is fine. Similar ticket [#1851](https://github.com/ember-cli/ember-cli/issues/1851)

ember version: 0.1.5
node version: 0.10.35
npm version: 2.1.8

jnhuynh:
I can confirm this issue too on Ember CLI 0.2.0.

knownasilya:
Not sure if this is related, but when I do:

```js
baseURL: '/dashboard/'
locationType: 'hash'
```

Everything works fine at `/dashboard/#/`. But if I have the following:

```js
baseURL: '/dashboard/'
locationType: 'auto'
```

The url is `/dashboard/dashboard/`.

PR #3570, opened 2015-03-19
Fix not setting of base tag when locationType is 'hash'.

jnhuynh:
Should resolve this issue: https://github.com/ember-cli/ember-cli/issues/2957

This is a issue surfaces when you have a Rails backend that serves up multiple Ember apps at different subdirectories. The Ember app needs to be aware of it's `baseURL` so that it can grab the correct set of assets and it needs `locationType = 'hash'` because the Rails router will grab all of the URL changes before Ember's router.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1258,8 +1258,7 @@ EmberApp.prototype.contentFor = function(config, match, type) {
   @return {String}        Base tag or empty string
  */
 function calculateBaseTag(config){
-  var baseURL      = cleanBaseURL(config.baseURL);
-  var locationType = config.locationType;
+  var baseURL = cleanBaseURL(config.baseURL);
 
   if (baseURL) {
     return '<base href="' + baseURL + '" />';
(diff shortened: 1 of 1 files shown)


rwjblue:
Why do you need a base tag at all in this scenario?  With hash location, your assets are loaded relative to the index.html which is how they are done by default.  

Also, you can use history location with rails if you set up your routes to glob all paths under a specific prefix and forward to a controller serving your ember app.

jnhuynh:
@rwjblue You are correct on both points.

The relative assets were served up without the base tag. The initial issues I had with grabbing relative assets was caused by a mistake on my part.

As for using a glob in the Rails route, it sort of works. The glob will match, for example `/ember-app1/*` and redirect to my Ember index file. The problem however is that all sub paths will get gobbled up by my Rails route before it ever gets to my ember app. For example `/ember-app1/some-resource/1` will cause an infinite redirect loop because my Rails router is just redirecting to itself. The only way around this has been to use the hash `locationType`.

The Rails routing issue is beside the point of this pull request anyway. My Rails example was not as illustrative or helpful as I originally thought.

---

I guess my final case for this is PR was caused by my thought "Why does the `locationType` influence the generation of the `base` tag?" It is not obvious to me since they appear to be separate things.

stefanpenner:
this seems to have gone, stale. I think the next time we make a change here, it will be when we remove base href

mkorfmann:
Is there any reason why the base tag is omitted when using hashlocation?

PR #3577, opened 2015-03-20
Added Ember + DS to predef to align with guides

joostdevries:
DO NOT MERGE

In the [guides rewrite](https://github.com/emberjs/guides/issues/9#issuecomment-82526641) it was decided to leave out importing `Ember` and `DS`. In the current implementation that will confront users with jshint errors if they copy code examples from the guides.

If the guides rewrite is going to push a lot of people to try Ember CLI, we should [avoid people running into this](https://www.google.nl/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=%22ember%20is%20not%20defined%22).

Alternate solutions:
- Clearly document the need to import in the guides (for example the "Concepts" section), like on the [ember cli website](https://github.com/ember-cli/ember-cli/pull/1016/files)
- Add imports to the examples/snippets

### DIFF
--- a/blueprints/app/files/.jshintrc
+++ b/blueprints/app/files/.jshintrc
@@ -2,6 +2,8 @@
   "predef": [
     "document",
     "window",
+    "DS",
+    "Ember",
     "-Promise"
   ],
   "browser": true,
--- a/blueprints/app/files/tests/.jshintrc
+++ b/blueprints/app/files/tests/.jshintrc
@@ -18,6 +18,7 @@
     "findWithAssert",
     "wait",
     "DS",
+    "Ember",
     "andThen",
     "currentURL",
     "currentPath",


stefanpenner:
why was this decided?

rwjblue:
I am very opposed to adding ANY predefs. 

We are not importing in the guides for the same reason that we don't always show all of the files: space and noise.

When we have something meaningful to convey (aka have actually useful modules to import from) it is a 100% different scenario, and we will simply update to import what is needed.

In the meantime, adding a small section/page with info on importing seems fine to me, but is ultimately @trek's call.

stefanpenner:
I am sure this was choosen with the best of intent, but I would like be made aware of the choices, as we are about to cut support for non-imported modules for ember-cli.

joostdevries:
A third option to deal with this would be to customize the JSHint error and hint the user to import. That would point people in the right direction straight away and avoid them turning to Google for answers.

rwjblue:
> I am sure this was chosen with the best of intent
 
Agreed. 

> I would like be made aware of the choices

The guide snippets are not and have never been "fully runnable examples", they convey specific concepts that are being taught at that section.  When the imports become something to be taught (e.g. what to import from where), we will teach them as well.

> we are about to cut support for non-imported modules for ember-cli

Good, lets do it!  The wonderful news about versioned guides: we get to update the guides to match the current recommendations at the time of each version of Ember.  When we ship an Ember version with a module system that we need to teach, BOOM we add them to the guides. :smile:

trek:
@joostdevries the lack of imports on the examples in _not_ meant to imply you're relying on `window.Ember`. The code samples are intended to highlight just the specific code to make the necessary point about what the text around the sample. You'll find many of our sample snippets have never been runnable as-is. That's just not the goal for them

joostdevries:
@trek That argument makes sense in a world where everyone understands ES6 modules. Right now, I think the reality is we're dealing with people coming from a globals-based world. I'd prefer to add the extra noise for now and remove it as ES6 modules become the standard.

topaxi:
Almost all ember-cli generators import Ember by default, so it should be, in my opinion, obvious that the imports are needed.

joostdevries:
@topaxi we don't point to generators in the guide at this point.

trek:
> That argument makes sense in a world where everyone understands ES6 modules.  

Happy to address when it proves to be a problem.

stefanpenner:
another to note, as we split up parts of ember into more modules, it will become the responsibility of the guide/docs to help inform the user where to find them.

tomdale:
Agree with the thread above:

1. We should not `predef` these things; in a near future version we will have users import e.g. `Component` directly instead of using `Ember.Component`; let's get people in the mode of remembering to import their dependencies.
2. As such (and I think this is your concern), the guides will need to tell people not only how to use these classes but also where to find them. Having an example in the guides but then having to lose context to look it up in the API docs just to use it sucks.

I think the best course of action for now is:

1. Don't `predef` anything new.
2. Include `import Ember from "ember"` in larger code snippets
3. Don't include `import Ember from "ember"` in shorter snippets where it's obvious the code is an excerpt intended to highlight a particular API, not present a fully runnable app.

Once the ability to consume Ember as discrete modules (e.g. `import Component from "ember-views/component"`) is shipped, let's revisit number 3 and consider always including the module name every time we show a code example that relies on a framework object.

joostdevries:
Agree with the above and yes, my concern is only about # 2.  I'll close this, open a PR in the guides repo where we very clearly point people to the need to `import` and add import statements in snippets that reflect a complete file (IMO it's more about whether it's actually a snippet or a file someone could drop into their app as-is).

tomdale:
@joostdevries Awesome. Thank you so much for your hard work on this. It's gonna be a HUGE improvement for new Ember users. :heart:

joostdevries:
fyi emberjs/guides#89. I'm going over the code blocks later to see check which ones may need imports.

PR #3578, opened 2015-03-20
Teaches updateChecker about dev builds

### LINKED ISSUE #3494: updateChecker does not take into account dev builds with `v0.2.0-sha-checksum`
https://github.com/ember-cli/ember-cli/blob/master/lib/models/update-checker.js#L60

cc @twokul

twokul:
fixes #3494


tests/unit/models/update-checker-test.js
```diff
@@ -57,6 +81,7 @@ describe('Update Checker', function() {
 
         return function() {
           updateChecker.versionConfig = versionConfig;
+          debugger
```
stefanpenner: Forgotten debugger

stefanpenner:
Awesome

lib/models/update-checker.js
```diff
@@ -56,8 +56,16 @@ UpdateChecker.prototype.doCheck = function() {
   }.bind(this)).catch(function() {
     return this.checkNPM();
   }.bind(this)).then(function(version) {
+    // match postfix SHA in dev version
+    var isDevBuild   = !!this.localVersion.match(/\b[0-9a-f]{5,40}\b/);
```
rwjblue: I think I'd rather make [lib/utilities/ember-cli-version.js](https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/ember-cli-version.js) be the only thing that has to be aware of this.  Can we make that script export another function (like `isDevelopment()`) then just consume it here?  

I'm mostly concerned with having the concept of what exactly goes into a dev build version number  be encapsulated in a single place.

twokul: @rwjblue makes sense to me. I'll make a change on the flught

twokul:
@stefanpenner @rwjblue I think this is ready for another look

rwjblue:
Thanks!

Issue #2329, opened 2014-10-16
ember-cli shouldn't watch vim swap files


fivetanley:
noticed this while working on a client project today that typing with vim made the build run multiple times, and may be the source of the testem crash issues i've experienced before. I'm just opening this to come back to it later. Any helpful tips on where to start looking would be great...

FWIW they were using ember-cli-coffeescript but I don't see anything in there that would watch swap files, but I can take a deeper look later.

PR #3585, opened 2015-03-21
[ENHANCEMENT] Add in-repo-addon generate and destroy support

trabus:
This PR adds the ability to generate and destroy blueprints inside of in-repo-addons by adding a `--in-repo-addon` flag to the generate and destroy commands. It also adds support for components as the first blueprint to be able to generate inside an in-repo-addon.

`ember generate component foo-bar --in-repo-addon my-addon`
or using shorthand:
`ember g component foo-bar -in-repo my-addon` 

This needs some follow up with docs, it would be helpful to know that you can generate blueprints inside in-repo-addons.


stefanpenner:
LGTM, my only comment is: it makes me slightly un-easy to add many if statements to add this sort of functionality. It would be great if a later refactoring utilized polymorphism instead.

trabus:
@stefanpenner [This is what structures would provide](https://github.com/trabus/rfcs/blob/e68c778593b8ad924893cf9e8d6f730b250451bf/active/0000-structures.md#rough-concept)

stefanpenner:
> @stefanpenner This is what structures would provide

:+1:

stefanpenner:
kicking appveyor.

noslouch:
I think this caused a regression to #2329. I'm seeing vim swap and undo history files get swept up in the build process.

PR #3587, opened 2015-03-21
Specified which files to publish to npm

jayphelps:
We're currently publishing pretty much everything (except stuff from `.gitignore`, which npm auto-ignores).

I figured we can shed a little weight, especially since `tests/` is large and growing...I know every little bit counts on a project this size.

###### tarball sizes:

```
  25,595,044 bytes (before)
- 25,364,564 bytes (after)
===========================
≈ 230.48 KB savings
```
###### before
image
###### after
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   "bugs": {
     "url": "https://github.com/ember-cli/ember-cli/issues"
   },
+  "files": ["bin/ember", "blueprints", "lib"],
   "bundledDependencies": [
     "bower",
     "bower-config",


rwjblue:
In general, I agree with this pull request, but I am unsure of removing all of tests/ without moving things out into lib/. 

I have been using the test utilities for testing commands, blueprints, etc in addons.

Thoughts?

stefanpenner:
this removes `0.1%` of the over-all file size, it doesn't seem worth it?

jayphelps:
Ohh, yeah, those utils definitely belong in lib then. I Haven't used them, so I have no clue (and am mobile right now) are they trivial to move into lib? How are they referenced in consumer addons? By path directly? (i.e. Breaking change)

rwjblue:
>  those utils definitely belong in lib then

I totally agree, I guess I was just pointing out that we should look at which things in `test/` are actually meant to be used externally (and therefore need to be moved out of `test/`).

jayphelps:
Hehe that's i directly what I hoped you could tell me. 😄 have any canonical example projects that use them?

stefanpenner:
Let's put them in test-helpers

stefanpenner:
closing this:

* doesn't really save us much currently
* will accept it, if the test-helpers where extracted (and deprecated) accordingly.

PR #3596, opened 2015-03-21
Allow addons to have pod based templates.

### LINKED ISSUE #3586: Pod templates inside addons don't get compiled
Currently there seems to be an issue with templates being imported from pods inside in-repo-addons (I haven't yet verified if the same issue affects addons as well). [This repo](https://github.com/trabus/in-repo-addon-pod-template) illustrates it, where trying to load a pod based component's template fails. The non-pod component is able to load with no issue, but the pod component fails because the template is not compiled.

Running the example app above, the following error occurs:
```
 Uncaught Error: Could not find module `foo-bar/components/foo-bar/template` imported from `foo-bar/components/foo-bar/component`
```

The error is not present when the `foo-bar` component is removed from the `application.hbs`.

I believe the issue lies with the way the templates are processed in addons. The [tree passed]( is `addon/templates`, so) into [compileTemplates](https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L443)  is `addon/templates`, so it seems the `addon` tree is missing templates inside pod structure folders.

rwjblue:
Fixes #3586.


gmurphey:
Is there any chance this will be merged before 0.2.4? Pod-based templates are the last major pain-point I have to work around with addon-based apps when working against ember-cli@master.

sglanzer:
Echoing @gmurphey here - looks like a better version of what I was attempting to solve.  What are the chances of this (or @chadhietala's alternative) making the next release?

stefanpenner:
> (or @chadhietala's alternative) making the next release?

this wont happen by the next release.

stefanpenner:
> Is there any chance this will be merged before 0.2.4?

if the tests are green, and @rwjblue feels good about it we can merge.

@chadhietala's solution will ultimately be better, but we can unwind this one when @chadhietala's work lands.

stefanpenner:
@rwjblue do you have time to get this working for 0.2.4? (tomorrow hopefully)

stefanpenner:
@sglanzer maybe if you have time you can rebase + get the test passing on this one.

sglanzer:
I'll see what I can do, it'll take a few days to get back to this.  Thanks for chasing the general status down

mansona:
Has there been any progress on this?

stefanpenner:
Yes. We are reworking the build system to enable this in a performant way.

mansona:
Awesome thanks guys :+1: really appreciate the work that you're doing

stefanpenner:
> Awesome thanks guys  really appreciate the work that you're doing

:) We hope to land that work soon. In the process of testing larger and larger apps and fixing the issues as they arise.

stevehanson:
Wondering if there's been any more progress on this. Really appreciate what you're doing. :)

stefanpenner:
yes, much progress has been made on build tooling performance. Hopefully soon, we can offer this without a hit.

stevehanson:
Awesome! Thanks for the update and the work you're doing, @stefanpenner.

luketheobscure:
This sure would be great to have. 

Could the documentation be updated in the mean time to reflect that you can't use pods in addons?

sglanzer:
We're using pods in all of our addon projects (20+), what problems are you
seeing?

luketheobscure:
@sglanzer I'm working on an in-repo addon. When using the following folder structure, the templates are not working:
```
/my-app
  └ lib
    └ my-addon
      ├─ component.js
      └─ template.hbs
```

I'm not sure if they're not being compiled, or not being resolved (I'm new to ember-cli, but I've been on ember for a long time). The component is resolved just fine.

If it matters, it's actually a `.coffee` and a `.emblem` file, but when using the non-pod style layout it's working fine.

sglanzer:
Ah, interesting - I haven't used an addon in-repo before.  I believe the
broccoli build looks for files in the conventional app and addon
directories, so you may need to override treeForApp in the ember-cli-build
file (I don't know if the addon path is included in app projects).  You'll
also likely need to import your template into the component directly (as
the layout property) because I don't think the resolver will associate the
template to the component otherwise.

Just because I haven't done this before, what is the advantage of an
in-repo addon versus an external addon or a component in the app directory
itself?

sglanzer:
I spent a bit of time trying this out and I couldn't get it to work with
the in-repo addon either.  While the build complains if a handlebars
compiler isn't in the project dependencies, if you look at the dist
afterward the template files from the in-repo addon aren't present.

I didn't take the time to dig into the broccoli tree feeding the template
compiler, but I would imagine there is a missing path for in-repo addons
there.

The good news is that pods and template work in normal addons, I just don't
think in-repo has been given much love.

stefanpenner:
We will revisit this as we finalize pods, @dgeb and @rwjblue are fleshing it out, once that matures this will be tackled (we may just reopen this PR)

PR #3601, opened 2015-03-22
Allow Node 0.13 in platform deprecation check.

rwjblue:


stefanpenner:
Should we allow versions greater then or equal to 11?

PR #3625, opened 2015-03-24
Add data-ember-action pointer to default css

knownasilya:
Seems like a nice default. Thoughts?

### DIFF
--- a/blueprints/app/files/app/styles/app.css
+++ b/blueprints/app/files/app/styles/app.css
@@ -1,0 +1,3 @@
+[data-ember-action] {
+  cursor: pointer;
+}
--- a/blueprints/app/files/app/styles/app.css
+++ b/blueprints/app/files/app/styles/app.css
@@ -1,3 +1,9 @@
+/* 
+ * For enabling pointer cursor on all elements with an action.
+ *
+ * Required if expecting using action on non-buttons,
+ * and expecting to work in iOS.
+ */
 [data-ember-action] {
   cursor: pointer;
 }
--- a/blueprints/app/files/app/styles/app.css
+++ b/blueprints/app/files/app/styles/app.css
@@ -1,7 +1,7 @@
 /* 
  * For enabling pointer cursor on all elements with an action.
  *
- * Required if expecting using action on non-buttons,
+ * Required if expecting to use action on non-buttons,
  * and expecting to work in iOS.
  */
 [data-ember-action] {
--- a/blueprints/app/files/app/styles/app.css
+++ b/blueprints/app/files/app/styles/app.css
@@ -1,8 +1,5 @@
 /* 
- * For enabling pointer cursor on all elements with an action.
- *
- * Required if expecting to use actions on elements other then buttons,
- * and expecting to work in iOS.
+ * This makes touch events work for all elements that have Ember actions.
  */
 [data-ember-action] {
   cursor: pointer;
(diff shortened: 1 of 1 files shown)


rwjblue:
It seems like a reasonable change, but I generally think we shouldn't ship with default styles.  It is unlikely that everyone will want it...

Maybe this could be a nice little addon, that does `app.import` for you...

knownasilya:
@rwjblue I think anyone that know what they are doing, is usually adding a css preprocess so they delete the `app.css` anyway. This is more for those new beginners.

lukemelia:
This particular css rule may be the exception, because without it, actions don't work on iOS with when they are on elements that are not naturally clickable. The only reason you wouldn't want something like this is if you are not supporting iOS, or only using action helpers on divs or buttons. If we keep it, it should be accompanied by short comment explaining why it exists.

stefanpenner:
> If we keep it, it should be accompanied by short comment explaining why it exists.

yes, lets add a comment.

blueprints/app/files/app/styles/app.css
```diff
@@ -0,0 +1,9 @@
+/* 
+ * For enabling pointer cursor on all elements with an action.
+ *
+ * Required if expecting using action on non-buttons,
+ * and expecting to work in iOS.
+ */
```
knownasilya: @stefanpenner @lukemelia how's that?

stefanpenner: LGTM
@lukemelia r?

blueprints/app/files/app/styles/app.css
```diff
@@ -0,0 +1,9 @@
+/* 
+ * For enabling pointer cursor on all elements with an action.
+ *
+ * Required if expecting to use action on non-buttons,
+ * and expecting to work in iOS.
+ */
```
knownasilya: @stefanpenner @lukemelia how's that?

stefanpenner:
so I think actually `for enabling pointer cursor` is no more documenting then the css itself. We should likely just say, something along the lines of: this makes touch events work for all elements with ember actions.

knownasilya:
@stefanpenner updated

stefanpenner:
@knownasilya nice, lets let @lukemelia who is likely better written and spoken then both of us combined give us input :) Then LGTM.

I'll get this out with 0.2.2

rwjblue:
I agree with @lukemelia in concept, but I still believe that this should be an addon that uses `app.import`.  

The following would be upsides to using an addon instead of shipping this as a default `app.css`:

* It makes upgrades massively easier for folks
* Enables this CSS to be added even when using Sass, LESS, etc (because the addon would still be adding to `vendor.css` in those cases).
* Makes adding/removing as simple as modifying your `package.json`.

rwjblue:
> Probably not worthwhile to create an addon, for one small rule

Addons do not have to be "heavy-weight" or a burden, they can be as simple as two files in a repo. We need to embrace our new addon overlords...

stefanpenner:
add-ons can also be versioned, this code-snippet cannot.

knownasilya:
@rwjblue those are all good points, especially the upgrade ones.

knownasilya:
I think I'm going to close this in favor of an addon, although would we add that to the core addons?

stefanpenner:
@knownasilya i think so, but @rwjblue should give us his input.

stefanpenner:
closing in-favor of an add-on.

PR #3629, opened 2015-03-24
Bump ember-qunit

abuiles:
More info https://github.com/rwjblue/ember-qunit/commit/d7016d9e15772efcddc73b75aad1c884d9e6f1b3


rwjblue:
As indicated in the version number change, this is somewhat breaking for folks assuming that `this.render()` returns the jQuery element when using `moduleForComponent`.

Updating here is ABSOLUTELY the right thing to do (thank you @abuiles), but we need to add something to the changelog.md in this pull request to explain the potential hazard when folks update.

abuiles:
@rwjblue correct, I didn't  add it now since I assume we would during release time.

rwjblue:
I just don't want whomever is going to do the release to forget....

rwjblue:
Closed by 97a739d72a95e380e39dfedb6520f7317474a5fd.

PR #3634, opened 2015-03-24
Resolves #3628 postprocessTree for styles with vendor + app

### LINKED ISSUE #3628: move postprocessTree('css') after merge with vendor styles
currently postprocessTree is being called after preprocessing app styles only

to make universally useful it should be called with the merged vendor + app styles tree.

This causes an issue with ember-cli-concat (https://github.com/sir-dunxalot/ember-cli-concat/issues/13) which itself was the result of https://github.com/ember-cli/ember-cli/pull/3581

This would also enable css minification to become an addon iso core

working on PR

jschilli:


stefanpenner:
failure was related to a broken ember-cli-babel release.

rwjblue:
:+1:

PR #3642, opened 2015-03-25
Check if style file with project name exists

btecu:
Currently if a style sheet file named the same as the project exists will cause an exception when building. The error is not very obvious. This check prevents the build and displays an error.


rwjblue:
Seems good to provide a nice warning here. It will only be checked on initial build (which seems OK to me).

@stefanpenner - r?

stefanpenner:
@rwjblue should this be a silent error?

rwjblue:
Yes, definitely (good catch!).  @btecu - Can you update to use `SilentError`?

btecu:
Done.

stefanpenner:
re-kicking app-veyor

PR #3645, opened 2015-03-25
Add ember-disable-prototype-extensions to addons by default.

### LINKED ISSUE #3443: Set prototype extensions to false by default for addons
See: https://github.com/dockyard/ember-validations/blob/master/tests/dummy/config/environment.js#L10

In ember-validations I have disabled prototype extension for the test suite. I believe that all addons should be developed with prototype extensions turned off to ensure the widest compatibility with how Ember is being used.

I don't have the time to add this into ember-cli, if anyone else wants to tackle that would be great. I'm not sure how best to accomplish this but the addon generator should be relying upon the `config/environment.js` from the app generator then modify the `EmberENV` object in the file to disable prototype extensions rather than generating its own file.

rwjblue:
Add here: https://github.com/rwjblue/ember-disable-prototype-extensions

Basically, just adds `window.EmberENV.EXTEND_PROTOTYPES = false;` before Ember is loaded.

Fixes #3443.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -20,6 +20,7 @@ module.exports = {
     contents.description = this.description;
     contents.keywords = contents.keywords || [];
     contents.dependencies = contents.dependencies || {}
+
     // npm doesn't like it when we have something in both deps and devDeps
     // and dummy app still uses it when in deps
     contents.dependencies['ember-cli-babel'] = contents.devDependencies['ember-cli-babel'];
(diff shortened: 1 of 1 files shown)


runspired:
So this snuck past me on the package.json diff and ended up biting me pretty hard tonight.  The errors you get when you aren't looking for this are pretty cryptic, and I only figured this out because I'd read some of the discussion that this was coming, I didn't realize it was landing.

Most of the conversions are simple (I didn't have enough of them to need to use the Watson tool for this guy) http://guides.emberjs.com/v1.10.0/configuring-ember/disabling-prototype-extensions/

But I don't see a clear way to replace `.on`.

rwjblue:
@runspired:

```javascript
import Ember from 'ember';
var on = Ember.on;

  doStuffWhenInserted: on('didInsertElement', function() {
    /* awesome sauce */
  });
})
```

rwjblue:
Submitted a pull request to the guides for this: https://github.com/emberjs/guides/pull/156

runspired:
Awesome! Thanks.

PR #3655, opened 2015-03-26
Uses Ember.keys instead of Object.keys in reexport

danmcclain:
`Object.keys` does not exist in IE8, so we use `Ember.keys` in `reexport`
instead of `Object.keys` to prevent explosions :fire: :boom:


rwjblue:
:+1:

PR #3662, opened 2015-03-26
Pass baseUrl to the router's rootUrl, resolves redirect issues

knownasilya:
Fixes issues with plugins like ember-simple-auth when they want to reload the page.

:warning: Needs tests still.

### DIFF
--- a/blueprints/app/files/app/router.js
+++ b/blueprints/app/files/app/router.js
@@ -2,7 +2,8 @@ import Ember from 'ember';
 import config from './config/environment';
 
 var Router = Ember.Router.extend({
-  location: config.locationType
+  location: config.locationType,
+  rootURL: config.baseURL
 });
 
 Router.map(function() {


rwjblue:
I am under the impression that this is only valid for certain location types.  We need to confirm all combinations of `history` and `hash` location type both with a baseUrl and without before merging...

knownasilya:
@rwjblue are there tests related to this here or in the ember repo?

rwjblue:
We should add tests here for this. 

Ember's tests are completely browser based and testing history location and reloading is not very simple in that context.

stefanpenner:
can you add a more detailed commit message ? Ideally the motivation. A commit message containing exactly what the diff contains, isn't very valuable.

runspired:
Cordova refreshing is pretty frustrating.

Refresh to the same location
```
window.location.href = window.location.toString();
```

Refresh to the base location
```
window.location.href = window.location.toString().substr(0, window.location.toString().indexOf('#'));
```

jayphelps:
Just chiming in to mention [I think I understand his need and agree](https://github.com/emberjs/ember.js/issues/9501#issuecomment-61883135) though I'm not sure this will impact this PR, both https://github.com/emberjs/ember.js/issues/9501 and https://github.com/emberjs/ember.js/issues/5334 are outstanding issues related to how `baseURL` is sometimes used incorrectly in ember.

knownasilya:
@stefanpenner I'm not even sure how to describe this change. Would something like the title of this PR work?

jayphelps:
Remember folks, rootURL !== baseURL in embers internals.

knownasilya:
So if you have `baseURL: '/dashboard/` in the config, this `rootURL` helps with the redirecting, but it adds a duplicate entry to the url, e.g. `/dashboard/dashboard/`

knownasilya:
We need another solution to this issue..

eddflrs:
@knownasilya any other ideas on how to resolve this issue? I'm also seeing the `/dashboard/dashboard/` issue that you mentioned.

knownasilya:
Well for simple-auth, I added the following action in my application route:

```js
  sessionInvalidationSucceeded() {
      if (!Ember.testing) {
        window.location.replace('/dashboard');
      }
    }
```

Without setting that attribute in the router.

knownasilya:
@stefanpenner thoughts on this? Should I just close it, seems like there is uncertainty about what each option does..

stefanpenner:
@knownasilya haven't had time to look into this one yet, at some point i will dig into the baseURL rootUrl base href crap.

stefanpenner:
this seems to have gone stale, i plan to revisit base/url stuff soon

PR #3669, opened 2015-03-27
Update dependencies

btecu:
Updated all dependencies that don't break any test. In general, most libraries have one documentation (not versioned) so it becomes useful to have an updated version.

The following libraries need manual intervention to be updated: `broccoli-clean-css`, `core-object`, `inquirer`, `lodash-node`, `markdown-it`, `markdown-it-terminal`, `nock`.


stefanpenner:
Awesome thanks. Want to open an issue for those outstanding updates?

btecu:
Once I figure what the new updates are breaking (maybe some deprecated api), I'll push..

stefanpenner:
Ah. Sorry I meant. Let's open an issue tracking that those still need to be updated. So we don't forget, and as we update them we can check off that list.

I'll be home soon, I can open it if you want

btecu:
Ah, yes, I can do that.

rwjblue:
We need to revert the update to glob.

stefanpenner:
Because of the symlink thing?

rwjblue:
yes

rwjblue:
We need to audit all of our broccoli based deps, and ensure they are using `{ follow: true }` option before we can unlock.

stefanpenner:
> We need to audit all of our broccoli based deps, and ensure they are using { follow: true } option before we can unlock.

does the new version work with `{follow: true }` ? I have some fixes to glob, I am curious if i need to also backport them...

rwjblue:
The new version should properly follow symlinks with the `{ follow: true }` option, but that option needs to be passed by every one of our dependencies that use `glob`.  We likely just need to audit to get a list of our deps that use glob and update to pass `{ follow: true }`.

stefanpenner:
opened a new issue: https://github.com/ember-cli/ember-cli/issues/4137

PR #3621, opened 2015-03-24
Remove broccoli-static-compiler

btecu:
broccoli-static-compiler has been deprecated, so it has been replaced with broccoli-funnel which has more options and is maintained.

PR #3672, opened 2015-03-27
Use broccoli-funnel instead of broccoli-static-compiler

joliss:
broccoli-funnel has the `allowEmpty` option that we needed CustomStaticCompiler for.

I didn't get any failing tests, so I'm not 100% sure that the code in `lib/models/addon.js` works properly, but it definitely *looks* correct. :)


stefanpenner:
awesome, by sure though that: `npm run test-all` was run.

stefanpenner:
tests failures look race related. I suspect we recently regressed on something related... in all likely-hood one of or deps has a race again...

stefanpenner:
this appears to be a duplicate of: https://github.com/ember-cli/ember-cli/pull/3621

joliss:
Ah, yes, now I'm getting the tests to cover my code. So things are OK then. Travis is green too.

Re dup, I caught a few extra lines I think, so you should merge my PR :)

stefanpenner:
Ya. Done :)

stefanpenner:
Gotta figure out those races. These brittle tests have been getting worse this week

stefanpenner:
ah, I forgot add-ons depend on having access to `this.pickFiles`, we will likely need to add backwards compat

lib/models/addon.js
```diff
@@ -364,7 +363,7 @@ Addon.prototype.treeForPublic = function(tree) {
     return tree;
   }
 
-  return this.pickFiles(tree, {
```
stefanpenner: specifically this case.

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/3675

PR #3673, opened 2015-03-27
Update Broccoli to 0.15.3

joliss:
OK, new plan of attack:

I thought we could update everything in one go and avoid having Broccoli's ugly compatibility code deployed widely. Turns out it's too many changes, and we'll have to do piecemeal plugin updates.

This updates to Broccoli 0.15.1, which supports new-style `.rebuild` trees but does not print noisy deprecation warnings unless you pass BROCCOLI_WARN_READ_API=y in the process environment.

Stef, was there something you had in the ember-cli-broccoli fork that you wanted me to merge? I can probably merge it and push it out in 0.15.2.

(Let's be sure to wait for Travis.)

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -18,7 +18,7 @@ module.exports = Task.extend({
     this.environment = this.environment || 'development';
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
 
-    var broccoli = require('ember-cli-broccoli');
+    var broccoli = require('broccoli');
     this.tree    = broccoli.loadBrocfile();
     this.builder = new broccoli.Builder(this.tree);
   },
--- a/lib/tasks/server/middleware/serve-files/index.js
+++ b/lib/tasks/server/middleware/serve-files/index.js
@@ -13,7 +13,7 @@ ServeFilesAddon.prototype.serverMiddleware = function(options) {
   options = options.options;
   debug('serverMiddleware: %s',options.outputPath);
 
-  var broccoliMiddleware = options.middleware || require('ember-cli-broccoli/lib/middleware');
+  var broccoliMiddleware = options.middleware || require('broccoli/lib/middleware');
   var middleware = broccoliMiddleware(options.watcher);
 
   var baseURL = cleanBaseURL(options.baseURL);
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "bundledDependencies": [
     "bower",
     "bower-config",
-    "ember-cli-broccoli",
+    "broccoli",
     "broccoli-clean-css",
     "broccoli-filter",
     "broccoli-funnel",
@@ -82,7 +82,7 @@
     "abbrev": "^1.0.5",
     "bower": "^1.3.12",
     "bower-config": "0.5.2",
-    "ember-cli-broccoli": "0.13.6",
+    "broccoli": "0.13.6",
     "broccoli-caching-writer": "0.5.5",
     "broccoli-clean-css": "0.2.0",
     "broccoli-es3-safe-recast": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "abbrev": "^1.0.5",
     "bower": "^1.3.12",
     "bower-config": "0.5.2",
-    "broccoli": "0.13.6",
+    "broccoli": "0.15.3",
     "broccoli-caching-writer": "0.5.5",
     "broccoli-clean-css": "0.2.0",
     "broccoli-es3-safe-recast": "^2.0.0",
--- a/tests/unit/models/addon-test.js
+++ b/tests/unit/models/addon-test.js
@@ -216,22 +216,22 @@ describe('models/addon.js', function() {
       describe('trees for it\'s treePaths', function() {
         it('app', function() {
           var tree = addon.treeFor('app');
-          expect(typeof tree.read).to.equal('function');
+          expect(typeof (tree.read || tree.rebuild)).to.equal('function');
         });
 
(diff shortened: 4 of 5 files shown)


stefanpenner:
Primarily the live reload hack. Although If we can make its needs configurable, we can merge and not hack up broccoli.

The two highlights are:

* the ansi stripping hack and {{{ so we emit html
* the live reload js path, so it recovers from an error

Bonus but not required:

* node api, which makes debugging and future tooling easier

stefanpenner:
The upgrade path you suggest makes sense. We can then incrementally submit issues and pr to the most common addons/plugins

joliss:
2 Travis failures: https://travis-ci.org/ember-cli/ember-cli/jobs/56051963#L1658 and https://travis-ci.org/ember-cli/ember-cli/jobs/56051963#L2026
One of them due to GitHub being hit with a DDoS, the other one I'm unsure.

joliss:
I can take a patch for the LiveReload injection if you can split it out from https://github.com/broccolijs/broccoli/pull/239.

I'd prefer not to merge the ANSI stripping hack; Seb says [he'll take the ANSI sequences out](https://github.com/broccolijs/broccoli/pull/239#issuecomment-84057020) in Babel if I understand him correctly, so it should resolve itself anyway.

The Node API has conflicts at the moment (https://github.com/broccolijs/broccoli/pull/219) because the Builder is being changed a lot. Maybe the path of least resistance is to wait for the API change to settle, but if you want it in, I can merge an updated PR right now as well.

stefanpenner:
> I'd prefer not to merge the ANSI stripping hack; Seb says he'll take the ANSI sequences out in Babel if I understand him correctly, so it should resolve itself anyway.

we just need an interim solution

stefanpenner:
> The Node API has conflicts at the moment (broccolijs/broccoli#219) because the Builder is being changed a lot. Maybe the path of least resistance is to wait for the API change to settle, but if you want it in, I can merge an updated PR right now as well.

https://github.com/broccolijs/broccoli/pull/219 is rebased for you

stefanpenner:
> I can take a patch for the LiveReload injection if you can split it out from broccolijs/broccoli#239.

would you be alright with threading an options, which could contain `liveReloadPath: ...` through

https://github.com/broccolijs/broccoli/blob/master/lib/middleware.js#L11 and https://github.com/broccolijs/broccoli/blob/master/lib/middleware.js#L96 ?

something like:

```js
// lib/middleware.js
module.exports = function(watcher, options) {
  return function broccoliMiddleware(request, response, next) {
    watcher.then(function(hash) {
// ...

    }, function(buildError) {
      var context = {
        // ...
        liveReloadPath: options && options.liveReloadPath
      }
```

and

```hbs
// templates/error.html
...
{{if liveReloadPath}}
  <script src="{{liveReloadPath}}"></script>
{{/if}}
...
```

stefanpenner:
reference implementation https://github.com/broccolijs/broccoli/pull/243 for https://github.com/ember-cli/ember-cli/pull/3673#issuecomment-86929617

stefanpenner:
@joliss I think that last thing, is ensuring the content of the error message can be formatted nicely.

joliss:
> on a side note, it would be lovely if we could just inject that live-reload from a different middleware, 

Yeah, this would be cool, or at least having a slightly more generic option to modify the error page. But I think it's ok for now.

> I think that last thing, is ensuring the content of the error message can be formatted nicely.

You mean ANSI colors? I maintain that this doesn't belong in Broccoli. ANSI sequences are a weird terminal-specific half-standard, so in my view they simply don't belong in `err.message`. The solution is for compilers not to do this. Seb [says](https://github.com/babel/babel/issues/837#issuecomment-81060615) there'll be an option to turn it off, so that's good. If we wanted to work around it in the meantime, we could put a hack into broccoli-babel-transpiler to catch errors and strip escape sequences from their `.message` properties.

stefanpenner:
> You mean ANSI colors? I maintain that this doesn't belong in Broccoli. ANSI sequences are a weird terminal-specific half-standard, so in my view they simply don't belong in err.message. The solution is for compilers not to do this. Seb says there'll be an option to turn it off, so that's good. If we wanted to work around it in the meantime, we could put a hack into broccoli-babel-transpiler to catch errors and strip escape sequences from their .message properties.

I currently strip out ansi and re-write as classes, for later styling. I suppose i could actually strip out the classes entirely and rely on the fixed-width formatting?

joliss:
Yes, that's what I'd do for now.

If we wanted to have colorful HTML in the future, it seems we'd want to have a separate `.messageHtml` part of the plugin error API or so. Wouldn't mind adding it, though perhaps it's overkill.

Also, when we add an auto-generated context snippet based on file/line/col, we can trivially run our own colorizer, so that might cover 99% of cases without adding more API surface.

stefanpenner:
> Yes, that's what I'd do for now.

sounds good, someone needs to write this code, but it will work (we aren't using colors anyways currently)

stefanpenner:
kicking appveyor again

joliss:
Updated PR to 0.15.3, to include https://github.com/broccolijs/broccoli/commit/55e908422784514b17f59029cb3887553cd34673, so uncolorized Babel contexts will work.

stefanpenner:
rad, testing this out in some apps right now.

stefanpenner:
app appears to work fine, the output isn't stripped of ansi does it isn't that great yet.

screen shot 2015-03-27 at 2 18 21 pm

stefanpenner:
ok created that locally

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/3682

Issue #3670, opened 2015-03-27
Update breaking dependencies


btecu:
The following dependencies can't be updated to the latest version because they break some tests:
- [ ] broccoli-clean-css
- [ ] core-object
- [ ] inquirer
- [x] lodash-node
- [x] markdown-it
- [x] markdown-it-terminal
- [x] nock

stefanpenner:
> core-object

i suspect this one will be tricky, and a WIP exists: https://github.com/ember-cli/ember-cli/pull/3056

rtm:
What tests do later versions of broccoli-clean-css cause to fail? I'm interested in using it (1.0.0) because it supports the new sourcemap feature in the latest version of clean-css.

stefanpenner:
> What tests do later versions of broccoli-clean-css cause to fail? I'm interested in using it (1.0.0) because it supports the new sourcemap feature in the latest version of clean-css.

unsure, got time to look into it and report back.

stefanpenner:
seems to have gone stale, if someone wants to help upgrade stale / outdated deps PR welcome :)

PR #3674, opened 2015-03-27
Update nock dependency

btecu:
By default `nock` will disable all HTTP requests throwing `NetConnectNotAllowedError`, except for the ones that you specifically create a nock scope for. See #3670.


rwjblue:
Thanks for tracking this down!!

PR #3677, opened 2015-03-27
Export return value from Router.map (closes #3676).

### LINKED ISSUE #3676: blueprint update
`Router.map` now returns an the router factory, where earlier it returned an instance of router.js.

we can update our blueprints to be:
```
import Ember from 'ember';

export default Ember.Router.extend().map(function() {
   // ...
});
```

this aligns us with the new 1.11 ember-cli based guides.

abuiles:


stefanpenner:
@abuiles i think the tests need to be udpated.

abuiles:
Checking.

--
Adolfo Builes

abuiles:
@stefanpenner ok, the issue is that currently we ship with 1.10.0 which returns undefined when calling `.map`, we'll need to put this on hold until 1.11 is released.

stefanpenner:
> @stefanpenner ok, the issue is that currently we ship with 1.10.0 which returns undefined when calling .map, we'll need to put this on hold until 1.11 is released.

ah oko, so we need to wait a bit for this one.

PR #3681, opened 2015-03-27
Deprecate funnel and pickfiles

stefanpenner:


stefanpenner:
@rwjblue r?

lib/models/addon.js
```diff
@@ -112,14 +116,17 @@ Addon.prototype._requireBuildPackages = function() {
 
   this.transpileModules  = this.transpileModules || require('broccoli-es6modules');
```
rwjblue: Needs to be deprecated also

stefanpenner: good catch.

stefanpenner:
@rwjblue updated

lib/models/addon.js
```diff
@@ -103,24 +107,43 @@ Addon.prototype.hintingEnabled = function() {
   @private
   @method _requireBuildPackages
  */
+
 Addon.prototype._requireBuildPackages = function() {
```
rwjblue: Since this now does nothing should we deprecate it also?

stefanpenner: i need it actually to get good error messages.

rwjblue: Ok, I don't see that in this pull request.

stefanpenner: its updated.

stefanpenner: been running this on several apps at work, and working hard to make nice deprecation output (and catching other cases i didn't think of)

I think i got them all now

samselikoff:
I changed this in my addon but consumers are having trouble. I tried adding `broccoli-merge-trees` and `broccoli-funnel` to devDependencies and dependencies, but if they used

```
npm install samselikoff/ember-cli-mirage
```

to install master, they get the error

> Cannot find module 'broccoli-merge-trees'

Any thoughts on how to fix? And, is it okay to use `this.Funnel` and `this.mergeTrees` for now?

samselikoff:
actually, adding to `dependencies` did fix. Is this the suggested path? If so I can add some clarification in the docs

stefanpenner:
Yup

samselikoff:
What about a line [here](http://www.ember-cli.com/#packagejson) for `dependencies` that says 

```js
"dependencies": {
  // add npm packages your addon depends on here
```

? It could be me but I'm still not clear on what goes there vs. devDependencies. Basically if my addon uses an npm package in index.js, it needs to be in dependencies?

This absolutely could have nothing to do with ember-cli and is just my lack of understanding of npm

PR #3690, opened 2015-03-28
[ENHANCEMENT] Addon-import support for built-in blueprints

trabus:
This PR changes the way blueprints are generated inside addons, following the recommendation [here](http://reefpoints.dockyard.com/2015/03/22/tips-for-writing-ember-addons.html) of importing the actual addon code into the app with a wrapper, like so:
```
// addon/components/foo-bar.js
import Ember from 'ember';
  // business logic
});

// app/components/foo-bar.js
import FooBar from 'my-addon/components/foo-bar';
export default FooBar;
```

I copied the `component-addon` blueprint with a general purpose `addon-import` blueprint that will allow it to be used for all blueprints that need it.

I still need to create generate and destroy tests for all the blueprints that had support added, and will do so this evening. I also need to make sure that the blueprints that shouldn't generate the import (anything that doesn't have a `__root__` token and isn't a mixin) aren't doing so, or causing any other side effects. Mostly wanted to put it up for review.


trabus:
Added some tests for generating inside an addon, and added a silent error for generating addons that don't contain a `__root__` token, so we can exclude blueprints that generate inside addons by simply not using the `__root__` token. On of the blueprints that I've removed the token from is routes, since they can't add to route.js when generating so they throw an error, and so they can't support in-addon generation at this time.

trabus:
@rwjblue @stefanpenner this is ready for review.

I wasn't quite sure which blueprints we should generate in this manner, but right now I've got route and resource throwing a warning, since they error anyway trying to modify `route.js`. The `component` and `mixin` blueprints already were setup, so those had no change.

I cleaned up the tests a bit, creating addon and in-repo-addon destroy and generate tests, and I moved the already existing ones out of the generate and destroy tests.

stefanpenner:
LVGTM, but i will want to do a more thorough review and test it out in the AM (getting a bit sleepy)

rwjblue:
PERFECT THANK YOU!!!

jonathanKingston:
@trabus I needed this so much last week, however I am sure I will need it again and again :+1: thanks.

PR #3692, opened 2015-03-29
Replace lodash-node with lodash

btecu:
`lodash-node` has been deprecated in favor of `lodash`. Also grouped and sorted the requires. Five tests were broken after the update, due to a circular reference. For #3670.


rwjblue:
Awesome, thank you!

stefanpenner:
:+1: thanks for all the contributions and help @btecu !

PR #3695, opened 2015-03-29
explain the PhantomJS installation recommendation

runspired:


stefanpenner:
i believe something similar may need to exists here: http://guides.emberjs.com/v1.11.0/ember-cli/

PR #3697, opened 2015-03-29
Update inquirer dependency

btecu:
Also fix tests and typos. For #3670.

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -4,6 +4,7 @@ var PleasantProgress = require('pleasant-progress');
 var Promise          = require('../ext/promise');
 var EOL              = require('os').EOL;
 var writeError       = require('./write-error');
+var Prompt           = require('inquirer').ui.Prompt;
 
 // Note: You should use `ui.outputStream`, `ui.inputStream` and `ui.write()`
 //       instead of `process.stdout` and `console.log`.
@@ -128,7 +129,6 @@ UI.prototype.prompt = function(questions, callback) {
   var promtOutputStream = this.through(null, function() {});
   promtOutputStream.pipe(this.outputStream);
 
-  var Prompt = require('inquirer').ui.Prompt;
   // Note: Cannot move this outside
   //       Need a new readline interface each time, 'cause it gets torn down
   function PromptExt() {
--- a/tests/unit/models/blueprint-test.js
+++ b/tests/unit/models/blueprint-test.js
@@ -286,7 +286,10 @@ describe('Blueprint', function() {
     });
 
     it('re-installing conflicting files', function(done) {
-      blueprint.install(options)
+      return blueprint.install(options)
+        .then(function() {
+          done();
+        })
         .then(function() {
           var output = ui.output.trim().split(EOL);
           ui.output = '';
@@ -327,8 +330,6 @@ describe('Blueprint', function() {
 
           expect(actualFiles).to.deep.equal(basicBlueprintFiles);
         });
-
-      return done();
     });
 
     it('installs path globPattern file', function() {
@@ -380,7 +381,10 @@ describe('Blueprint', function() {
       });
 
       it('ignores files in ignoredUpdateFiles', function(done) {
-        blueprint.install(options)
+        return blueprint.install(options)
+          .then(function() {
+            done();
+          })
           .then(function() {
             var output = ui.output.trim().split(EOL);
             ui.output = '';
@@ -421,8 +425,6 @@ describe('Blueprint', function() {
 
             expect(actualFiles).to.deep.equal(basicBlueprintFiles);
           });
-
-        return done();
       });
     });
 
--- a/tests/unit/models/blueprint-test.js
+++ b/tests/unit/models/blueprint-test.js
@@ -285,8 +285,8 @@ describe('Blueprint', function() {
         });
     });
 
-    it('re-installing conflicting files', function() {
-      return blueprint.install(options)
+    it('re-installing conflicting files', function(done) {
+      blueprint.install(options)
         .then(function() {
           var output = ui.output.trim().split(EOL);
           ui.output = '';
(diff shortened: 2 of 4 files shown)


tests/unit/models/blueprint-test.js
```diff
@@ -285,8 +285,8 @@ describe('Blueprint', function() {
         });
     });
 
-    it('re-installing conflicting files', function() {
-      return blueprint.install(options)
+    it('re-installing conflicting files', function(done) {
+      blueprint.install(options)
```
stefanpenner: why does returning the promise not work?

tests/unit/models/blueprint-test.js
```diff
@@ -377,8 +379,8 @@ describe('Blueprint', function() {
         Blueprint.ignoredUpdateFiles.push('foo.txt');
       });
 
-      it('ignores files in ignoredUpdateFiles', function() {
-        return blueprint.install(options)
+      it('ignores files in ignoredUpdateFiles', function(done) {
+        blueprint.install(options)
```
stefanpenner: ^^

tests/unit/models/blueprint-test.js
btecu: It causes the test to fail:
```
1) Blueprint basic blueprint installation re-installing conflicting files:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (/usr/local/lib/node_modules/mocha/lib/runnable.js:170:19)
      at Timer.listOnTimeout (timers.js:88:15)
```

stefanpenner: We should figure out why and fix it, without this we will have annoyingly brittle tests

btecu: Weird, the tests are passing for me locally.

stefanpenner:
I would prefer to continue to use the promise interface for these tests, rather then the `done` approach.

btecu:
How about calling `done` before returning the promise?

stefanpenner:
The promise technique ensures errors and async are fully integrated with mocha, although sometimes more work, it results in more stability of our tests. If your blocked. Let me know I can investigate

tests/unit/models/blueprint-test.js
```diff
@@ -285,9 +285,12 @@ describe('Blueprint', function() {
         });
     });
 
-    it('re-installing conflicting files', function() {
+    it('re-installing conflicting files', function(done) {
       return blueprint.install(options)
         .then(function() {
+          done();
```
stefanpenner: this is totally in the wrong place

tests/unit/models/blueprint-test.js
```diff
@@ -377,9 +380,12 @@ describe('Blueprint', function() {
         Blueprint.ignoredUpdateFiles.push('foo.txt');
       });
 
-      it('ignores files in ignoredUpdateFiles', function() {
+      it('ignores files in ignoredUpdateFiles', function(done) {
         return blueprint.install(options)
           .then(function() {
+            done();
```
stefanpenner: ^ same as above

tests/unit/models/file-info-test.js
```diff
@@ -96,46 +96,52 @@ describe('Unit - FileInfo', function(){
...
     });
 
-    return fileInfo.confirmOverwrite().then(function(action){
+    fileInfo.confirmOverwrite().then(function(action){
       var output = ui.output.trim().split(EOL);
       expect(output.shift()).to.match(/Overwrite.*\?/);
       expect(action).to.equal('overwrite');
     });
+
+    return done();
```
stefanpenner: same as above, this `completes` before the entire `confirmOverwrite` part of the test.

tests/unit/models/file-info-test.js
```diff
@@ -96,46 +96,52 @@ describe('Unit - FileInfo', function(){
...
     });
 
-    return fileInfo.confirmOverwrite().then(function(action){
+    fileInfo.confirmOverwrite().then(function(action){
       var output = ui.output.trim().split(EOL);
       expect(output.shift()).to.match(/Overwrite.*\?/);
       expect(action).to.equal('skip');
     });
+
+    return done();
```
stefanpenner: ^

tests/unit/models/file-info-test.js
```diff
@@ -96,46 +96,52 @@ describe('Unit - FileInfo', function(){
...
     });
 
-    return fileInfo.confirmOverwrite().then(function(action){
+    fileInfo.confirmOverwrite().then(function(action){
       var output = ui.output.trim().split(EOL);
       expect(output.shift()).to.match(/Overwrite.*\?/);
       expect(action).to.equal('diff');
     });
+
+    return done();
```
stefanpenner: ^

lib/ui/index.js
```diff
@@ -4,6 +4,7 @@ var PleasantProgress = require('pleasant-progress');
 var Promise          = require('../ext/promise');
 var EOL              = require('os').EOL;
 var writeError       = require('./write-error');
+var Prompt           = require('inquirer').ui.Prompt;
```
stefanpenner: this is explicitly required lazily so we don't incure the penalty at cli boot for parts of the CLI that do not require inquirer

stefanpenner:
this upgrade is tricky, something changed that is breaking our sorta gnarly test helpers of these code-paths...

btecu:
From what I understand, inquirer requires now to call `done` before finishing. So I don't know what the best place to call it is. Maybe chain it after the current promise?

stefanpenner:
> From what I understand,

what leads you to believe this?

> inquirer requires now to call done before finishing

i believe this is merely a work-around, we should solve the underlying issue.

btecu:
That's what I thought based on the errors I was getting before calling `done`, but I don't have much experience with tests.
What is the underlying issue?

stefanpenner:
> What is the underlying issue?

this needs to be discovered, I have not yet had a chance to dig in.

stefanpenner:
this seems to have gone stale...

stefanpenner:
Would love this to be actually updated, but it will need additional work then this PR has.

Issue #3671, opened 2015-03-27
Test failure with `git clone`: "Unable to read current working directory:"


joliss:
I'm getting the following test failure on master:

```
    1) ember new with git blueprint uses checks out the blueprint and uses it
    ✓ ember new without skip-git flag creates .git dir (56ms)
    ✓ ember new with --dry-run does not create new directory

  Acceptance: ember destroy pod
    ✓ .ember-cli usePods setting destroys in pod structure without --pod flag (69ms)
    ✓ .ember-cli usePods setting destroys in basic structure with --pod flag (78ms)
Command failed: /bin/sh -c git clone https://github.com/trek/app-blueprint-test.git /tmp/ember-cli115227-56239-1oalbbo
Cloning into '/tmp/ember-cli115227-56239-1oalbbo'...
fatal: Unable to read current working directory: No such file or directory
fatal: remote did not send all necessary objects
Unexpected end of command stream

Error: Command failed: /bin/sh -c git clone https://github.com/trek/app-blueprint-test.git /tmp/ember-cli115227-56239-1oalbbo
Cloning into '/tmp/ember-cli115227-56239-1oalbbo'...
fatal: Unable to read current working directory: No such file or directory
fatal: remote did not send all necessary objects
Unexpected end of command stream

    at ChildProcess.exithandler (child_process.js:716:12)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:169:7)
    at maybeClose (child_process.js:984:16)
    at Socket.<anonymous> (child_process.js:1153:11)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:166:7)
    at Pipe.close (net.js:464:12)
    ✓ .ember-cli usePods setting correctly destroys component (92ms)
    ✓ controller foo --pod (83ms)

...

  829 passing (1m)
  4 pending
  1 failing

  1) Acceptance: ember new ember new with git blueprint uses checks out the blueprint and uses it:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (/home/ubuntu/src/ember-cli/node_modules/mocha/lib/runnable.js:170:19)
      at Timer.listOnTimeout (timers.js:88:15)
```

stefanpenner:
@joliss seems like something with git + your setup is being funky.

joliss:
This is on Linux; error messages like `Unable to read current working directory: No such file or directory` tend to happen when the working directory is deleted.

[...]

Issue #3679, opened 2015-03-27
Docker + ember-cli applications in Development


jakehow:
See previous discussion -> http://discuss.emberjs.com/t/working-with-docker-ember-cli-in-development/7658

I am not sure there is any issue that ember-cli can address per-se or that ember-cli-todos necessarily wants to include docker setup instructions, but you can see my example [here](https://github.com/ember-cli/ember-cli-todos/compare/ember-cli:master...jakehow:docker-development-support), which I can easily PR if desired.

Getting the ember-cli developer environment to run in a container is a challenge, mostly due to a synergy of conflicting issues between docker and node dependencies.  While playing with the ember-cli-todos I figured out an interesting work around to the biggest of them which is volume behavior.

Normally when developing an app locally in a docker container, we would mount the entire application's source tree in the container so that any files we change during development can be reflected in the vm we are testing with immediately.

Because `node_modules` and `bower_components` are within the project's path this is not feasible for ember-cli applications.  If you do your npm install steps before copying your source into the container and mounting the volume, the folders with the dependencies are simply removed because of how docker volume mounting works.

The solution in the code linked above is that we create specific volumes which exclude the root path of the project where `node_modules` and `bower_components` live.  We expect that nearly all change a developer needs to deal with is happening in `app/`, `config/`, `public/` and `test/` and create these as standalone volumes.  The developer can now work on their host machine and changes are reflected in the container.

Performance for builds and rebuilds of ember-cli-todos seem fine to me but I will test this strategy on some other projects to see if any issues arise.

There are might be other ways to resolve this but I could not get any of the other things I tried to work.  Previously I was focused on moving `node_modules` out of the project directory to somewhere like `$HOME`, but ember-cli does not seem to use this in the load path, and I don't know if that is still supposed to work in node.

stefanpenner:
Are you proposing a community maintained docker image?

jakehow:
@stefanpenner either that, or I can put together a guide using ember-cli-todos as an example and easy way to keep the guide up to date, so that developers have something they can reference when they want to put together a docker setup for their apps.

The current big pain here that I am looking to overcome is described more concisely in #3735

WooDzu:
>  We expect that nearly all change a developer needs to deal with is happening in app/, config/, public/ and test/ and create these as standalone volumes.

@jakehow This seems to be reasonable approach to bundle into docker-compose.yml while keeping rest of the ember-cli stuff in docker image. What are your concers about this?

[...]

PR #3698, opened 2015-03-29
adding docker for linux testing/debugging

kellyselden:
Created as a response to comments like https://github.com/ember-cli/ember-cli/issues/3671#issuecomment-86787406 and https://github.com/ember-cli/ember-cli/issues/3671#issuecomment-87162370. It downloads an Ubuntu box, and installs ember-cli ready for `npm test`.

I'm thinking of making docker versions as well to help the multi-OS debugging/slowness effort.

It might make sense to move this to a separate repo at some point, but it would be slightly less accessible/harder to find when you need it.


visualjeff:
Just a comment.

Could a docker image be used for what your trying to accomplish?   I wrote a Dockerfile for our team to test ember-cli versions.  Its at https://registry.hub.docker.com/u/visualjeff/ember-cli/

stefanpenner:
> Could a docker image be used for what your trying to accomplish? I wrote a Dockerfile for our team to test ember-cli versions. Its at https://registry.hub.docker.com/u/visualjeff/ember-cli/

might be interesting to combine the two, (basically make it trivial to use the docker based when testing/deving ember-cli) and also advertising the docker stuff for those who prefer to use it.

stefanpenner:
@visualjeff also worth noting Node v1.6.2, is actual iojs not node.

kellyselden:
I'm planning on working on a /dev/docker solution as well. Us Windows folk tend to be used to Vagrant over Docker so I think it makes since to have them both.

stefanpenner:
ideally i'd prefer one approach, makes it easier to test and work on..

kellyselden:
I added a Linux Docker container as well. But since Windows containers don't exist yet, we are stuck with a Vagrant Windows section.

So as it stands in this branch, there is a
* Vagrant Linux
* Vagrant Windows
* Docker Linux

@stefanpenner @visualjeff Any comments on whether to keep them all/nix some would be appreciated.

Also, @visualjeff I created a Dockerfile using latest node as a base, then cloning ember-cli master. This makes it so you have the latest greatest ember-cli, and you don't have to upload a new container for each new version/commit.

kellyselden:
#3679 might be interested in this.

stefanpenner:
cc @jakehow

jakehow:
See my branch of ember-cli-todos (https://github.com/ember-cli/ember-cli-todos/compare/master...jakehow:docker-development-support) as a POC for a docker setup.  There are also full instructions (from fresh dev instance in https://github.com/ember-cli/ember-cli-todos/blob/6fcbc946311d415d1ddbaf74951e8286ec76ffbc/DOCKER_README.md)

This setup uses the 'anointed' (all currently officially recommended docker tools) path for simplicity's sake.

This would require some tweaking for your testing needs probably but happy to help.

kellyselden:
@jakehow I think the scope of our efforts are different. I think you are making developing a new ember-cli app in a container easier, where I am doing the bare minimum to get `npm test` working on ember-cli/master to help debugging Linux.

jakehow:
@kellyselden I understand both scopes/use cases.  They should have the same (or overlapping) solution ideally so the learnings in one can be applied to the other.

Is there some benefit from having vagrant AND docker support for testing, or is just having docker support sufficient for the goal?

kellyselden:
@jakehow I think Docker would be sufficient if there were Windows containers.

jakehow:
@kellyselden I doubt that is happening anytime soon, and then it will probably only be for future releases of windows since containerization is not currently possible.

The overlap between `dev/docker/linux` and `dev/vagrant/linux` is more what I was asking about.  Is that just to show an example of both and presumably one would get chosen?

kellyselden:
@jakehow I can get behind removing the vagrant linux. I wonder if anyone will miss it. Probably not ha.

kellyselden:
I modified to have only:
* Vagrant Windows
* Docker Linux

Does this sound good to everyone?

PR #3700, opened 2015-03-30
fixes #3613 - added path.normalize

### LINKED ISSUE #3613: Paths printed in console not correct on Windows
paths

swelham:
Updated to fix #3613 by running the display path through path.normalize.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -589,7 +589,7 @@ Blueprint.prototype.buildFileInfo = function(destPath, templateVariables, file) 
   return new FileInfo({
     action: 'write',
     outputPath: destPath(mappedPath),
-    displayPath: mappedPath,
+    displayPath: path.normalize(mappedPath),
     inputPath: this.srcPath(file),
     templateVariables: templateVariables,
     ui: this.ui


rwjblue:
Seems good.

@stefanpenner / @trabus - r?

trabus:
I'm okay with this, though I'm still curious as to why it was happening in the first place. Though it's a minor thing as long as the files are being generated to the right paths still.

swelham:
@trabus it happens because the pods section of the path is being joined through ```path.join``` which internally normalizes the path when it joins them together.

This doesn't actually effect anything other than how it renders in the console, purely a UI thing.

stefanpenner:
window failure appeared to be one of the random ones.

trabus:
@swelham I was more specifically talking about how the filename is being generated with the wrong slash on windows, yet the tokens are being generated properly.

For instance the [model blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/model/files/app/__path__/__name__.js)  which was the example used in the original issue, shouldn't be providing the filename as `app/__path__/__name__.js`. On windows it should be `app\__path__\__name__.js`, correct? 

Anyhow, as long as it's being normalized before generating the actual file, all is well.

PR #3709, opened 2015-03-31
[DOCS] change broccoli-emblem-compiler to ember-cli-emblem-hbs-printer

foxnewsnetwork:
In the asset compilation section on the ember-cli website. This is because the broccoli one doesn't actually work while the ember-cli one does work.


stefanpenner:
Cc @mixonic

rwjblue:
LGTM

mixonic:
lgtm. @jondayton is working on a name change for the CLI addon, but this patch will guide wayward souls in the right direction.

Issue #3718, opened 2015-03-31
The `included` hook and nested addons: different application argument


RSSchermer:
I'm in the process of updating 2 related component addons I maintain, [ember-spin-spinner](https://github.com/RSSchermer/ember-spin-spinner) and [ember-spinner-button](https://github.com/RSSchermer/ember-spinner-button) (ember-spinner-button depends on ember-spin-spinner), to ember-cli 0.2.2 (from 0.0.46, upgrading was long overdue).

I first updated ember-spin-spinner, which was a breeze and when using it as a "top level" addon it still works as expectd. However, when updating ember-spinner-button, which uses ember-spin-spinner as a "second level" (nested) addon, its `included` hook starts throwing errors during `ember s` and `ember test`:

```js
included: function(app) {
  this._super.included(app);

  var spinPath = path.join(app.bowerDirectory, 'spin.js');

  app.import(path.join(spinPath, 'spin.js'));
  app.import(path.join(spinPath, 'jquery.spin.js'));
}
```

The `app` argument is not the same when ember-spin-spinner is used as a "top level" addon as when used as a "second level" (nested) addon. The specific problems in this case are that `app.bowerDirectory` and `app.import` are undefined.

I did some digging through the issues here and came across [this comment](https://github.com/ember-cli/ember-cli/issues/3531#issuecomment-81133458) by @rwjblue, which seems to indicate that this is intended behavior. 

In that comment, a distinction also seems to be made between "top level" and "second level" addons. For ember-spin-spinner the intend was not to make it either a top level or a second level addon, but both: I wanted it to be usable both a standalone addon, as well as a dependency for ember-spinner-button.

rwjblue:
`this._super.included(app);` is incorrect, the first point in the comment you linked explains why.

`this._super.included.apply(this, arguments);` should fix the error.

rwjblue:
Ahhh, I missed that this error was thrown from a second level addon's `included` hook.  My comment is still correct about calling `_super`, it just won't fix the issue.

RSSchermer:
Thanks for the quick reply!

I found it hard to describe the problem. I'll try to summarize:

- ember-spin-spinners `index.js` is fine when using it as a top level addon in an application.
- ember-spin-spinners `index.js` throws errors when using it as a nested addon inside another addon (ember-spinner-button in this case).

[...]

PR #3722, opened 2015-03-31
Updated included hook example

RSSchermer:
Updated included hook example as discussed in issue #3718.

On a bit of a side-note: I've seen both `this.app.import` and `app.import` (using the application argument passed to the hook) used in issue discussions. Both seem to work. Is there a difference/preferred way?


rwjblue:
Thank you!

I think either `this.app` or `app` are fine, but they will only be set on top level addons...

Issue #3650, opened 2015-03-25
Ember generate multiple addon blueprints 


offirgolan:
Running the following command:

```
ember g addon1 addon2
```
Only runs the generator for addon1. Functionality wise, it should run both addon1 and addon2 generator.

stefanpenner:
+1

swelham:
I have had a go at this, really in need of some feedback on it #3724

stefanpenner:
Doesn't seem like on anyones critical path

PR #3724, opened 2015-03-31
[ENHANCEMENT] Adds the ability to run generate on multiple addons at once

swelham:
This is my initial stab at #3650. This isn't quite ready for merge yet (needs tests at least), but I wanted some feedback on this as there are a couple of areas that don't sit well with me.

Firstly I am making the assumption that when running ```ember g addon1``` you cannot specify any options for that command (like you can in ```ember g model foo name:string```).

Secondly I am not to sure on the way that I am detecting if the blueprint is an addon or not. Currently looking in ```Blueprint#defaultLookupPaths()``` and ```Project#localBlueprintLookupPath()``` to work this out.

Overall though, this is working and doesn't break existing tests, so it can't be all that bad :)

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -18,12 +18,14 @@ module.exports = Task.extend({
     if (isBuiltInBlueprint) {
       blueprints.push({
         name: options.args[0],
-        options: options.args
+        args: options.args,
+        options: options
       });
     } else {
       options.args.forEach(function(arg) {
         blueprints.push({
-          name: arg
+          name: arg,
+          options: options
         });
       }, this);
     }
@@ -49,7 +51,8 @@ module.exports = Task.extend({
 
   runBlueprint: function(blueprint) {
     var name = blueprint.name;
-    var options = blueprint.options || [];
+    var args = blueprint.args || [];
+    var options = blueprint.options || {};
 
     var mainBlueprint  = this.lookupBlueprint(name, options.ignoreMissingMain);
     var testBlueprint  = this.lookupBlueprint(name + '-test', true);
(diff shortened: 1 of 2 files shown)


lib/tasks/generate-from-blueprint.js
```diff
@@ -12,8 +12,52 @@ module.exports = Task.extend({
...
+      paths: this.project.blueprintLookupPaths(),
+      ignoreMissing: ignoreMissing
+    });
+  },
+
+  runBlueprints: function(blueprints) {
+    return this.runBlueprint(blueprints.shift())
+      .then(function() {
+        if (blueprints.length > 0) {
+          this.runBlueprints(blueprints);
```
stefanpenner: this needs a return, or we sever the promise chain

lib/tasks/generate-from-blueprint.js
```diff
@@ -12,8 +12,52 @@ module.exports = Task.extend({
...
+
+  lookupBlueprint: function(name, ignoreMissing) {
+    return Blueprint.lookup(name, {
+      paths: this.project.blueprintLookupPaths(),
+      ignoreMissing: ignoreMissing
+    });
+  },
+
+  runBlueprints: function(blueprints) {
+    return this.runBlueprint(blueprints.shift())
```
stefanpenner: this should likely use the sequence helper: https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/sequence.js#L1-L15

swelham: Awesome, I totally missed that helper.

lib/tasks/generate-from-blueprint.js
swelham: Ah yes I see. I will update to switch to using the sequence helper which will kill off this code anyway.

stefanpenner:
@swelham any time coming up to work on

- [ ] using the sequence helper for the promise sequence
- [ ] tests

swelham:
@stefanpenner I am currently away at the moment, I should have time to finish this on my return (about a week or so).

In terms of tests, do you see this just being an acceptance test in [generation-test.js]h(ttps://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/generate-test.js)?

stefanpenner:
this seems to have gone stale. Would consider it, if rebased, tests passing and comments addressed.

Closing for now.

PR #3739, opened 2015-03-31
Use shorthand ES6 re-export for addon components

jayphelps:


stefanpenner:
ah, does esperanto support this already?

Also, i suspect tests will need to be updated.

jayphelps:
@stefanpenner Yep, [completely supported by spec](https://people.mozilla.org/~jorendorff/es6-draft.html#table-42) and esperanto/babel.

I tried to find a test for this, but couldn't. Will take another look if the build fails. #lazyCI

stefanpenner:
> Yep, completely supported by spec and esperanto/babel.

that table doesn't show that specific case, `default` I believe is somewhat special. Lets get confirmation from someone in spec land.

jayphelps:
@stefanpenner Okiedoke, are you pinging someone in particular? seems to be same case of `import { default as another } from "mod";`

jeffmo:
`export {default} from "foo";` means to re-export a named export from the `"foo"` module. This is different from the actual `default` export.

In general, it's impossible to have a local binding named `default` because `default` is a keyword. However, you can do some trickery to alias another valid binding identifier to something like `default` such as `export {blah as default};`. In this case, you'd be able to re-export that *named* binding that happens to go by the name of `default`.

Funky indeed, but it is what it is. There's now a Stage-1 proposal on the table as of last week to have a first-class proxy-export syntax for re-exporting the actual default-export from another module:
https://github.com/leebyron/ecmascript-more-export-from

jayphelps:
@jeffmo Dang. Thank you for clarifying. Advice on how should transpilers like Babel should be handling this case to prevent people adopting this syntax? Babel itself even uses it internally.

jayphelps:
Cc/ @sebmck https://github.com/ember-cli/ember-cli/pull/3739#issuecomment-88307260

stefanpenner:
They should likely error if exporting the default keyword. Because it is mostly just mis leading

sebmck:
I don't understand how this is the case. [Table 42](https://people.mozilla.org/~jorendorff/es6-draft.html#table-42) clearly shows that `export default foo;` uses the `[[ExportName]]` `"default"`.

`default` is a valid `IdentifierName` so the `export { foo as default };` and `export { default } from "foo";` [grammar allows this](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports.).

@jeffmo 

> In general, it's impossible to have a local binding named default because default is a keyword.

This is surely the case for `export { default };` but not for `export {default} from "foo";` as it's not accessing any local bindings.

> In this case, you'd be able to re-export that named binding that happens to go by the name of `default`.

But the `[[ExportName]]` of `export default foo;` is `"default"` and the `[[ExportName]]` of `export { default } from "mod";` is `"default"` since [Static Semantics: ExportEntriesForModule](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-static-semantics-exportentriesformodule) will turn the `default` `IdentifierName` to the `StringValue of IdentifierName` which is `"default"`.

jayphelps:
Yeaaaaaaah

caridy:
@sebmck is correct. Think about `import * as foo from "foo"`, in which case you get the exotic namespace object with a `default` prop if module module happens to have a default export. Same with `loader.import()` (once we finish it jejeje).

@jeffmo the intent of the `*default*` trick in the spec is just for the local bindings, but it does not affects the specifiers in export declarations.

caridy:
to add to that, remember that `import foo from "foo"` is just sugar for `import {default as foo} from "foo";`, where `default` is the specifier, not the weird `*default*`.

caridy:
that's a runtime error, because you're exporting a local identifier called `default` which is invalid.

sebmck:
Ah cool, thanks @caridy!

caridy:
ha, I might be wrong on the runtime error, I will check that in a bit when I get to my computer.

caridy:
ok, after looking at `export {default}`, it is definitely a static error, just like espree and esprima. Just think about this case:

```
window.foo = 1;
window.default = 2;
console.log(foo);
console.log(default);
VM637:2 Uncaught SyntaxError: Unexpected token default
```

So, for the case of `export {default}` it is a static error due to the unexpected token default, but for `export {foo}` it has to be a dynamic error since we don't know if foo is going to be defined, or provided, (as global) or not.

jayphelps:
So what's the official verdict on this original topic? Two conflicting confirmations.

caridy:
`export { default };` :-1: static error
`export { default } from "foo";` :+1: re-export the default export from module `foo`.

jayphelps:
That was my understanding as well. Just looking for spec-defined confirmation and I'm sure transpilers/browsers need it as well.

sebmck:
@jayphelps See my [previous comment](https://github.com/ember-cli/ember-cli/pull/3739#issuecomment-88337404) for spec references.

jayphelps:
Cc/@wycats

This goes back to the destructuring != import syntax we were talking about.

jayphelps:
@sebmck I saw them, but AFAIK we have two differing TC39 members interrupting it differently which to me means the spec isn't clear enough and needs this specifically addressed and clarified.

jayphelps:
The plot thickens.... https://twitter.com/_jayphelps/status/585627159659700224

rwaldron:
@caridy is correct. 


> I saw them, but AFAIK we have two differing TC39 members interrepting it differently which to me means the spec isn't clear enough

No, it just means that one person is wrong. The spec is precisely clear: 

```
ExportDeclaration :
  export * FromClause ;
  export ExportClause FromClause ;
  export ExportClause ;
  export VariableStatement
  export Declaration
  export default HoistableDeclaration[Default]
  export default ClassDeclaration[Default]
  export default [lookahead ∉ {function, class}] AssignmentExpression[In] ;

ExportClause :
  { }
  { ExportsList }
  { ExportsList , }

ExportsList :
  ExportSpecifier
  ExportsList , ExportSpecifier

ExportSpecifier :
  IdentifierName
  IdentifierName as IdentifierName
```

[This is immediately followed by an explicitly defined Early Error](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports-static-semantics-early-errors), that **only** applies to the production that it's specified for: 

```
ExportDeclaration : 
  export ExportClause ;
```

Which is: 

```js
export { default }; // static error
```

Otherwise...

````js
export { default } from "foo"; 
```

Is totally ok, because `default` here is _IdentifierName_, not _Identifier_.

PR #3756, opened 2015-04-02
adding a blueprint uninstall test

kellyselden:
uninstall is untested.


stefanpenner:
> uninstall is untested.

good catch, we should aim to add code-coverage to ember-cli eventually as well.

kellyselden:
@stefanpenner :+1:

PR #3759, opened 2015-04-02
Order bower dependencies alphabetically

pmdarrow:
The blueprint for `package.json` has dependencies ordered alphabetically. This makes diffs easy to read when upgrading Ember CLI. This PR applies alphabetical ordering to the `bower.json` blueprint to make diffs against it similarly easy.


stefanpenner:
thanks

PR #3762, opened 2015-04-02
Improved serializer-test blueprint

bmac:
The existing serializer-test blueprint makes it hard for the user to test the behavior of the serializer. This is because most serializer methods that a user may override to add custom behavior expect the `store` and a `snapshot` to be passed into the method as an argument. Currently these are painful to create in tests because accessing them often requires user's know about private API methods to setup this state. 

This change takes advantage of the existing the `moduleForModel` helper to make it easy to get access to a record instance in a test. Using the record's `serialize` method we can easily pass the `store` and `snapshot` into the serializer without having to worry about creating a `store` or `snapshot` instance. 

I'm very interested in receiving feedback on this pr so don't hesitate to comment if you are using a better technique to test your serializers.

### DIFF
--- a/blueprints/serializer-test/files/tests/unit/__path__/__test__.js
+++ b/blueprints/serializer-test/files/tests/unit/__path__/__test__.js
@@ -1,15 +1,18 @@
 import {
-  moduleFor,
+  moduleForModel,
   test
 } from 'ember-qunit';
 
-moduleFor('serializer:<%= dasherizedModuleName %>', {
+moduleForModel('<%= dasherizedModuleName %>', {
   // Specify the other units that are required for this test.
-  // needs: ['serializer:foo']
+  needs: ['serializer:<%= dasherizedModuleName %>']
 });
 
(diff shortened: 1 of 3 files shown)


theazureshadow:
@bmac ember-data has some interesting [serializer test](https://github.com/emberjs/data/blob/f6e40f1aa7e4d57d0a29357ad9d6f4ddcc381b5c/packages/ember-data/tests/integration/serializers/rest-serializer-test.js) work. It relies on a helper called [`setupStore`](https://github.com/emberjs/data/blob/174fc94488d7e66ebb8b61e3bef142c0d051edca/tests/ember-configuration.js#L49) (that, admittedly, does a lot of mocking that may be fragile).

bmac:
Thanks for reporting that @ballPointPenguin. @stefanpenner is there a way to branch the generator logic for just the application generator?

stefanpenner:
Ya. Look at the component generators

PR #3771, opened 2015-04-03
[ENHANCEMENT] Support custom node_module paths

### LINKED ISSUE #3735: Handling of node_modules outside of project path
This is related to #3679

The documentation for NODE_PATH and the default behavior can be found here -> https://nodejs.org/api/modules.html

I am currently trying to figure out if we can install node dependencies in $HOME/.node_modules which is one of the default search paths according to the node documentation.  It appears based on the stacktrace that code is being loaded from this path but I run into an error.

When I test this with ember-cli-todos by running `ember test` I get the following error:

```shell
➜  ember-cli-todos git:(docker-development-support) ✗ docker-compose run todos ember test
Future versions of Ember CLI will not support v0.10.38. Please update to Node 0.12 or io.js.
version: 0.2.1
0.2.2

Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/#watchman for more info.
Building...
Missing template processor
Error: Missing template processor
    at module.exports.preprocessTemplates (/root/.node_modules/ember-cli/lib/preprocessors.js:164:11)
    at EmberApp._processedTemplatesTree (/root/.node_modules/ember-cli/lib/broccoli/ember-app.js:532:10)
    at EmberApp.appAndDependencies (/root/.node_modules/ember-cli/lib/broccoli/ember-app.js:773:24)
    at EmberApp.javascript (/root/.node_modules/ember-cli/lib/broccoli/ember-app.js:838:34)
    at EmberApp.toArray (/root/.node_modules/ember-cli/lib/broccoli/ember-app.js:1155:10)
    at EmberApp.toTree (/root/.node_modules/ember-cli/lib/broccoli/ember-app.js:1177:30)
    at Object.<anonymous> (/opt/app/Brocfile.js:20:22)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
```

Here are the contents of /root/.node_modules:

```shell
➜  ember-cli-todos git:(docker-development-support) ✗ docker-compose run todos bash -c "ls /root/.node_modules"
broccoli-asset-rev		   ember-cli-inject-live-reload
ember-cli			   ember-cli-qunit
ember-cli-app-version		   ember-cli-uglify
ember-cli-babel			   ember-data
ember-cli-content-security-policy  ember-export-application-global
ember-cli-dependency-checker	   express
ember-cli-htmlbars		   glob
ember-cli-ic-ajax
```

jakehow:
This enables custom node_module paths through an environment variable EMBER_NODE_PATH.  Places that previously constructed node_modules paths manually now use a shared function `nodeModulesPath`.

Because this behavior is implicit right now I am not 100% sure if the tests I wrote capture all of the desired behavior.

This PR is intended to resolve #3735 

cc @stefanpenner

### DIFF
--- a/lib/cli/node-modules-path.js
+++ b/lib/cli/node-modules-path.js
@@ -7,30 +7,52 @@ function isSubdirectoryOf(parentPath, possibleChildPath) {
     possibleChildPath.indexOf(parentPath) === 0;
 }
 
-// A utility function for determining what path an addon may be found at. Addons
-// will only be resolved in the project's local `node_modules/` directory. This
-// is in contrast to a plain `require('some-module')` lookup, which would resolve
-// a library according to the paths in NODE_PATH.
-//
-// A description of node's lookup logic can be found here:
-//
-// * https://nodejs.org/api/modules.html#modules_all_together
-//
-// By requiring with an absolute path this logic is bypassed.
-//
+/**
+  A utility function for determining what path an addon may be found at. Addons
+  will only be resolved in the project's own `node_modules/` directory, they
+  do not follow the standard node `require` logic that a standard
+  `require('mode-module')` lookup would use, which finds the module according
+  to the `NODE_PATH`.
+
+  A description of node's lookup logic can be found here:
+
+  https://nodejs.org/api/modules.html#modules_all_together
+
+  Using this method to discover the correct location of project's `node_modules`
+  directory allows addons to be looked up properly even when that `node_modules`
+  directory is outside of the project's root.
+
+  This method checks the env variable `EMBER_NODE_PATH`. If present, its value
+  is used to determine the `node_modules` path.
+
+  Possible use cases for this include caching
+  the `node_modules/` directory outside of a source code checkout, and
+  ensuring the same source code (shared over a network) can be used with
+  different environments (Linux, OSX) where binary compatibility may not
+  exist.
+
+  For example, if you have a project in /projects/my-app and its `node_modules`
+  directory is at /resource/node_modules, you would:
+
+  ```
+    # Node uses this as its search path for standard `require('module')` calls
+    export NODE_PATH=/resource/node_modules
+
+    # So that ember addon discovery looks here
+    export EMBER_NODE_PATH=/resource/node_modules
+
+    cd /projects/my-app && ember build
+  ```
+
+  @private
+  @method nodeModulesPath
+  @param  {String} context The starting directory to use to find the
+                            node_modules path. This will usually be the
+                            project's root
+  @return {String} absolute path to the node_modules directory
+ */
 module.exports = function nodeModulesPath(context) {
 
-  // Optionally configure a different home for `node_modules/` in a parent
-  // directory of the project. Possible use cases for this include caching
-  // the `node_modules/` directory outside of a source code checkout, and
-  // ensuring the same source code (shared over a network) can be used with
-  // different environments (Linux, OSX) where binary compatibility may not
-  // exist.
-  //
-  // For example, you can specify a different home directory for modules:
-  //
-  //   EMBER_NODE_PATH=/opt/deps/node_modules NODE_PATH=/opt/deps/node_modules ember build
-  //
   var nodePath = process.env.EMBER_NODE_PATH;
   var contextPath = path.resolve(context);
 
@@ -38,14 +60,15 @@ module.exports = function nodeModulesPath(context) {
     var configuredPath = path.resolve(nodePath);
 
     // The contextPath is likely the project root, or possibly a subdirectory in
-    // node_modules/ nested dependencies. If it is more specific (a subdirectory
-    // of) than the configuredPath prefer the more specific contextPath.
+    // node_modules/ nested dependencies. If it is more specific than the
+    // the configuredPath (i.e. it is a subdirectory of the configuredPath)
+    // prefer the more specific contextPath.
     if (isSubdirectoryOf(configuredPath, contextPath)) {
-      return path.resolve(contextPath,'node_modules');
+      return path.resolve(contextPath, 'node_modules');
     } else {
       return path.resolve(nodePath);
     }
   } else {
-    return path.resolve(contextPath,'node_modules');
+    return path.resolve(contextPath, 'node_modules');
   }
 };
--- a/lib/cli/node-modules-path.js
+++ b/lib/cli/node-modules-path.js
@@ -1,0 +1,51 @@
+'use strict';
+
+var path  = require('path');
+
+function isSubdirectoryOf(parentPath, possibleChildPath) {
+  return possibleChildPath.length > parentPath.length &&
+    possibleChildPath.indexOf(parentPath) === 0;
+}
+
+// A utility function for determining what path an addon may be found at. Addons
+// will only be resolved in the project's local `node_modules/` directory. This
+// is in contrast to a plain `require('some-module')` lookup, which would resolve
+// a library according to the paths in NODE_PATH.
+//
+// A description of node's lookup logic can be found here:
+//
+// * https://nodejs.org/api/modules.html#modules_all_together
+//
+// By requiring with an absolute path this logic is bypassed.
+//
+module.exports = function nodeModulesPath(context) {
+
+  // Optionally configure a different home for `node_modules/` in a parent
+  // directory of the project. Possible use cases for this include caching
+  // the `node_modules/` directory outside of a source code checkout, and
+  // ensuring the same source code (shared over a network) can be used with
+  // different environments (Linux, OSX) where binary compatibility may not
+  // exist.
+  //
+  // For example, you can specify a different home directory for modules:
+  //
+  //   EMBER_NODE_PATH=/opt/deps/node_modules NODE_PATH=/opt/deps/node_modules ember build
+  //
+  var nodePath = process.env.EMBER_NODE_PATH;
+  var contextPath = path.resolve(context);
+
+  if (nodePath) {
+    var configuredPath = path.resolve(nodePath);
+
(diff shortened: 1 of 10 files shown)


stefanpenner:
looks like windows is unhappy

jakehow:
@stefanpenner I think my last variation should pass (looks like builds got really backed up and someone cancelled a bunch of them).  If there is a better suggestion as to how to handle this on windows I am open to hear it (I currently do not have a local way to test on windows).

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var path  = require('path');
+
+module.exports = function(context) {
+  var nodePath = process.env.EMBER_NODE_PATH;
+  var contextPath = path.resolve(context);
+
+  if(nodePath) {
```
stefanpenner: whitespace

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,22 @@
...
+    var nodeModulesPath = path.resolve(nodePath);
+
+    if(contextPath.indexOf(nodeModulesPath) === 0) {
```
stefanpenner: ^

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,22 @@
...
+      return path.resolve(contextPath,'node_modules');
+    }
```
stefanpenner: formatting isn't aligned with the rest of the code base.

stefanpenner:
do all the NPM commands `npm install` `npm link` and `ember install:npm` correctly observe these changes?

jakehow:
@stefanpenner fixed whitespace/formatting issues, let me know if anything else is off

`npm` commands will not be effected by this in any way, the environment variable is specific to `ember-cli`

It would be the responsibility of the developer to run `npm install` in the correct directory if they want to set this variable.

The `ember-cli:install` tasks appears to call npm tasks directly so there should be no behavior change here either.  There are no paths passed as options to these tasks or anything like that.

stefanpenner:
I will try to spend some time understanding the docker issue correct. I am hesitant to accept more complexity here, without fully understanding the problem it aims to solve.

jakehow:
@stefanpenner let me know if you need any help with this, I will also bang on this next week with @mixonic (he will be in our office all week).  The heart of the issue is really very simple in that `ember-cli` is hardcoding the `node_module` dependency paths which makes the runtime environment inflexible.  The docker problems are a symptom of this issue.

There is a secondary benefit that comes from changing this which is that the behavior of dependency loading is currently implicit and harder to understand than it should be since there are no explicit requirements or tests.  I actually think the `bower_components` path loading code all works correctly for the docker scenario based on my current understanding, but it would still be valuable to break that out into its own function as well so it can be explicitly tested and its behavior specified.

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var path  = require('path');
+
+module.exports = function(context) {
+  var nodePath = process.env.EMBER_NODE_PATH;
```
stefanpenner: should we just try to observer `NODE_PATH` directly then? I imagine this would align ember-cli with node's own behavior more closely? Or is this going to far?

jakehow: @steffanpenner NODE_PATH brings extra baggage with it. It will try to load globally installed modules and node_modules installed in $HOME which we might want to avoid. My goal here is t be able to say I have a clean controlled node_modules path but it is not necessarily at the project root.

stefanpenner: sounds reasonable, thanks.

mixonic:
> I actually think the bower_components path loading code all works correctly for the docker scenario based on my current understanding, but it would still be valuable to break that out into its own function as well so it can be explicitly tested and its behavior specified.

It sounds like bower will be going away in the future, so if it is working well for now it is likely safe to punt making any changes.

lib/cli/node-modules-path.js
jakehow: Default `NODE_PATH` behavior is described here.  https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders

If this value is unset we could actually do whatever we wanted with it, but I think doing something different than what's described in the node docs would be confusing to developers familiar with the default behavior, so keeping our own envvar seems like the best option.

jakehow:
@mixonic sounds good, only benefit would be it would be in one nice little spot to rip out when this happens, and we would have specs which might make that transition easier.

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,20 @@
...
+  var contextPath = path.resolve(context);
+
+  if (nodePath) {
+    var nodeModulesPath = path.resolve(nodePath);
+
+    if (contextPath.indexOf(nodeModulesPath) === 0) {
+      return path.resolve(contextPath,'node_modules');
```
mixonic: This seems to use `contextPath` when it is more specific than `nodeModulesPath`.

For example given a `contextPath` of `/Projects/cat-time` and a `nodeModulesPath` of `/Projects`, the `/Projects/cat-time` directory would be used (`/Projects` has an index of `0` on `/Projects/cat-time`).

This seems backwards? And perhaps we should wrap this in a small semantic helper called `hasSubdirectoryOf` etc.

jakehow: @mixonic I think this is the correct behavior based on what the code currently does from my reading.  If `node_modules` path is `/Projects/node_modules` and a `contextPath` of `/Projects/node_modules/my-add-on` is passed in, we want to return `/Projects/node_modules/my-add-on/node_modules`

Right now this is reflected in the tests in this way.  Happy to break it out into an intention revealing function though.

mixonic: Right, we discussed. The more specific path is correct, as it may be a specific subdirectory/nested dep in `node_modules/`

bantic:
This has been updated to use a `isSubdirectoryOf` intention-revealing method inside the node-modules-path utility, and had some documentation added. It's also been squashed and rebased against latest master. This should be good to go, now.

mixonic:
Did some Real World testing, and this requires a few ecosystem changes:

* ember-cli-dependency-checker must be improved to use the same `node_modules` as the project itself.
* Master of [rwjblue/ember-cli-ic-ajax](https://github.com/rwjblue/ember-cli-ic-ajax) is required. Fixes a bug where paths were joined and not resolved.
* ember-cli-htmlbars needs to correctly resolve paths instead of join, as submitted in https://github.com/ember-cli/ember-cli-htmlbars/pull/49

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,51 @@
...
+
+function isSubdirectoryOf(parentPath, possibleChildPath) {
+  return possibleChildPath.length > parentPath.length &&
+    possibleChildPath.indexOf(parentPath) === 0;
+}
+
+// A utility function for determining what path an addon may be found at. Addons
+// will only be resolved in the project's local `node_modules/` directory. This
+// is in contrast to a plain `require('some-module')` lookup, which would resolve
+// a library according to the paths in NODE_MODULES_PATH.
```
bantic: should be `NODE_PATH`

bantic:
Here's a test app that works with its node_modules directory outside of the project, along with instructions:

https://github.com/bantic/test-app

It removes the ember-cli-dependency-checker, but we'll update the dep checker to be more tolerant of a different node_modules location.

mixonic:
Sanity checked and good to go. /cc @rwjblue the contains `nodeModulesPath` as a property. It is a fully resolved path (unlike `bowerDirectory`).

lib/cli/node-modules-path.js
```diff
@@ -0,0 +1,51 @@
...
+// will only be resolved in the project's local `node_modules/` directory. This
+// is in contrast to a plain `require('some-module')` lookup, which would resolve
+// a library according to the paths in NODE_PATH.
+//
+// A description of node's lookup logic can be found here:
+//
+// * https://nodejs.org/api/modules.html#modules_all_together
+//
+// By requiring with an absolute path this logic is bypassed.
+//
```
stefanpenner: can this be converted into the YUIdoc form, this will match other documented functions.

we are slowly making progress on improving doc coverage: https://inch-ci.org/github/ember-cli/ember-cli?pending_build=47203

bantic: Yeah I'll update that tonight or tomorrow morning

bantic:
@stefanpenner updated the comments to be yuidoc style.

We also have a PR to ember-cli-dependency-checker pending that will update it to respect these path changes. The [test-app](https://github.com/bantic/test-app) is updated to use it. It now runs fine with a node_modules directory outside of the project root and stock ember-cli addons except that we are using customized-and-PR'd versions for dependency checker, ic-ajax and htmlbars.

rwjblue:
@bantic - ember-cli-ic-ajax@0.1.2 and ember-cli-htmlbars@0.7.5 are released, and should solve the need to run custom versions, please confirm?

bantic:
@rwjblue confirmed! Thanks. Updated the test app to use these versions and it works great.

mixonic:
@stefanpenner I think we're waiting on a +1 and merge from you. Iterating on changes this week is ideal, let us know if we need to change anything.

stefanpenner:
kicking appveyor.

I don't have time to thorough review this this week, so I will take your confidence as sufficient in merging once appveyor is happy. I don't want to block the train, but if we discover issues down the road I will pester the 3 of you ;)

mixonic:
rebased and force pushed.

Issue #3730, opened 2015-03-31
Quest: Improve Code quality


stefanpenner:
I've enabled code climate, it's likely a good idea that we attempt to improve some of the problematic areas. Some things, appear invalid (browser checks, in node code etc), we should report them with code-climate and help them improve.

We are in a relatively good place was(2.2) -> now [![Code Climate](https://codeclimate.com/github/ember-cli/ember-cli/badges/gpa.svg)](https://codeclimate.com/github/ember-cli/ember-cli) but we can strive to improve further.

Anyways, hopefully we can collectively triage some of the problem areas.

- [ ] D	[lib/broccoli/ember-app.js](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js)
- [ ] D	[lib/models/blueprint.js](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js)
- [ ] D	[lib/tasks/npm-install.js](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/npm-install.js)
- [ ] D	[lib/tasks/npm-uninstall.js](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/npm-uninstall.js)
- [ ] D	[lib/models/project.js](https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js)
- [ ] D	[lib/models/addon.js](https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js)
- [ ] C	[lib/models/command.js](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js)
- [ ] B	[blueprints/component-addon/index.js](https://github.com/ember-cli/ember-cli/blob/master/blueprints/component-addon/index.js)
- [ ] B	[blueprints/component/index.js](https://github.com/ember-cli/ember-cli/blob/master/blueprints/component/index.js)
- [ ] B	[lib/commands/install-bower.js](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/install-bower.js)
- [ ] B	[lib/commands/install-npm.js](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/install-npm.js)
- [ ] B	[lib/commands/uninstall-npm.js](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/uninstall-npm.js)

[edit: updated list – Fri May 22 09:45:01 PDT 2015]

[...]

PR #3775, opened 2015-04-03
Code quality blueprint duplicates

kellyselden:
Improved model/blueprint.js from F to D.

Interesting to learn that breaking large functions into more concise smaller functions makes the score decrease because it counts the number of function calls. The solution I guess would be to move functions to new files.

#3730


stefanpenner:
breaking blueprints into their own files and [micro-lib](https://github.com/ember-cli/blueprint) is the end goal

twokul:
:+1:

stefanpenner:
kicking appveyor

PR #3780, opened 2015-04-03
Upgrade to npm 2.7.6

davewasmer:
The release notes for npm@2.7.6 say it all, but recent versions of npm broke installs for private Github repos. Since the CLI uses npm for a variety of uses (including installing addons from blueprint hooks via `addAddonToProject`), this means addons hosted in private Github repos don't work.

2.7.6 fixes this. It's currently pre-release, so we may want to wait to upgrade until it goes stable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "morgan": "^1.5.2",
     "node-uuid": "^1.4.3",
     "nopt": "^3.0.1",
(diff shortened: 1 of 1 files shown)


stefanpenner:
@davewasmer awesome, was just about to do this. Thanks! :)

PR #3784, opened 2015-04-04
Add ember-try for addons.

rwjblue:
@kategengler has been doing an amazing job with [ember-try](https://github.com/kategengler/ember-try), and I think this is ready for default inclusion in the addon blueprint.


rwjblue:
Should be good now, had a few things fixed in ember-try (thanks @kategengler!) to make Windows support better.

MiguelMadero:
:+1:

PR #3786, opened 2015-04-04
Throw if templating a file fails

davewasmer:
Previously, if an error was thrown during the template rendering process, the error was simply swallowed. This was likely to avoid errors when attempting to template a binary file, since the byte stream, when converted to a string, could contain template markers.

Adds the isbinaryfile package to detect if the file is binary. If so, it skips templating entirely. If not, it attempts to template the file, and allows exceptions to reject the render promise.

This happened to expose some tests that should have been failing around the http-proxy generators (the tests were not supplying enough arguments). Fixes those tests too.

### DIFF
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -90,7 +90,12 @@ FileInfo.prototype.render = function() {
         if (isBinaryFile(content, fileStat.size)) {
           return content;
         } else {
-          return processTemplate(content.toString(), context);
+          try {
+            return processTemplate(content.toString(), context);
+          } catch (err) {
+            err.message += ' (Error in blueprint template: ' + path + ')';
+            throw err;
+          }
         }
       });
     });
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -3,9 +3,11 @@
 var fs           = require('fs');
 var Promise      = require('../ext/promise');
 var readFile     = Promise.denodeify(fs.readFile);
+var lstat        = Promise.denodeify(fs.stat);
 var chalk        = require('chalk');
 var EditFileDiff = require('./edit-file-diff');
 var EOL          = require('os').EOL;
+var isBinaryFile = require('isbinaryfile');
 
 function processTemplate(content, context) {
   return require('lodash/string/template')(content)(context);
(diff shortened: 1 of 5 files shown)


QuantumInformation:
I'm going through the upgrade steps for my test-recorder then will link to this branch from my npm and run ember init again.

lib/models/file-info.js
```diff
@@ -84,11 +86,13 @@ FileInfo.prototype.render = function() {
       context = this.templateVariables;
   if (!this.rendered) {
     this.rendered = readFile(path).then(function(content){
-      try {
-        return processTemplate(content.toString(), context);
-      } catch (error) {
-        return content.toString();
-      }
+      return lstat(path).then(function(fileStat) {
+        if (isBinaryFile(content, fileStat.size)) {
```
stefanpenner: we should ad a test for binary files

tests/unit/models/file-info-test.js
```diff
@@ -53,16 +53,19 @@ describe('Unit - FileInfo', function(){
...
     assign(options, validOptions, { inputPath: templateWithUndefinedVariable });
     var fileInfo = new FileInfo(options);
 
-    return fileInfo.render().then(function(output){
-      expect(output.trim()).to.equal('Howdy <%= enemy %>',
-        'expects to fall back to raw text');
+    return fileInfo.render().then(function() {
+      throw new Error('FileInfo.render should reject if templating throws');
+    }).catch(function(e) {
+      if (!e.toString().match(/ReferenceError/)) {
```
stefanpenner: we should ensure this error message is actionable

lib/models/file-info.js
davewasmer: There already is one, see https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/file-info-test.js#L69

stefanpenner: :+1:

stefanpenner:
needs rebase

tests/unit/models/file-info-test.js
davewasmer: Improved the error message a bit. See https://github.com/ember-cli/ember-cli/pull/3786/files#diff-c92e7822b3072a102f36bc32217520e4R96

davewasmer:
Rebased and addressed all the outstanding comments @stefanpenner

stefanpenner:
kicking appveyor

PR #3794, opened 2015-04-04
BUGFIX fixes vars-on-top error in ESLint

jonathanKingston:
Fixes an error reported by ESLint

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -3,9 +3,11 @@ import Resolver from 'ember/resolver';
 import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 
+var App;
+
 Ember.MODEL_FACTORY_INJECTIONS = true;
 
-var App = Ember.Application.extend({
+App = Ember.Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
   Resolver: Resolver


stefanpenner:
doesn't eslint have a setting that disables this?

jonathanKingston:
Well you can turn it off globally however, I like the rule as it prevents obscurities like this:
```
f();
var me;
```

```
function f() {
  me = 12;
}
```

## Turning the rule off locally
```
/*eslint vars-on-top:0 */
```

However this then allows for the following code:
```
/*eslint vars-on-top:0 */

Ember.MODEL_FACTORY_INJECTIONS = true;

var App = Ember.Application.extend({
  modulePrefix: config.modulePrefix,
  podModulePrefix: config.podModulePrefix,
  Resolver: Resolver
});

loadInitializers(App, config.modulePrefix);

var somethingElse = window;
(function () {
 somethingElseBad.write('Hello world!');
}())
var somethingElseBad = document;

export default App;
```

This is why I have requested a more specific concessions file: https://github.com/eslint/eslint/issues/2203#issuecomment-89555519

This could be something that gets put into ember-cli-eslint as a default.

[vars-on-top documentation](http://eslint.org/docs/rules/vars-on-top.html)

rwjblue:
This still annoys me (only slightly), but it seems harmless and if it makes folks using eslint happier I suppose it is fine.

jonathanKingston:
:+1:  sorry for the annoyance and thanks.

rwjblue:
To be clear: it wasn't you that was annoying.

Globegitter:
Even though it is not relevant for this case anymore, @jonathanKingston there is also an option to turn off `eslint-rules` just for a single line via `//eslint-disable-line <ruleName>`, e.g. `//eslint-disable-line no-unused-expressions`

jonathanKingston:
@Globegitter wow I had not seen that no.
I really don't like comment blocks like this as it leads to lots of meta data in comments and bad practice.
Especially as JSHint is the defacto here it would be pretty bad form to do that.

Also I think it is lost here:
````
var App = Ember.Application.extend({
  modulePrefix: config.modulePrefix,
  podModulePrefix: config.podModulePrefix,
  Resolver: Resolver
}); //eslint-disable-line vars-on-top
```

I'll start work on my global concessions for ESLint and see if anyone is interested as managing comments becomes very difficult quickly I think.

Any thoughts on this matter @rwjblue as it likely will raise it's ugly head again?

PR #3795, opened 2015-04-05
Code Climate: adding test coverage

kellyselden:
A task in #3730. Istanbul generates a `./coverage` folder, then Travis sends the info to Code Climate via web service. I believe coverage is only analyzed on the master branch, so you might not see it working until it is merged.

But you can see it in action on my master:
* [test coverage](https://codeclimate.com/github/kellyselden/ember-cli)
* [new travis output at the bottom](https://travis-ci.org/kellyselden/ember-cli/jobs/57241667)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ common-tmp
 docs/build
 .node_modules-tmp
 .bower_components-tmp
+coverage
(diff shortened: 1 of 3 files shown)


rwjblue:
LGTM

PR #3802, opened 2015-04-05
remove pre-mature process.exit when existing `ember test —server`

stefanpenner:
This should prevent the tmp files from leaking in this scenario

### DIFF
--- a/lib/tasks/test-server.js
+++ b/lib/tasks/test-server.js
@@ -9,23 +9,29 @@ module.exports = TestTask.extend({
   init: function() {
     this.testem = this.testem || new (require('testem'))();
   },
+
   invokeTestem: function(options) {
-    this.testem.startDev(this.testemOptions(options), function(code) {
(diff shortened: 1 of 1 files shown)


rwjblue:
The process.exit was added there to prevent an extra 10-12 second delay after quitting testem. Can you confirm that has not resurfaced with this change?

stefanpenner:
> Can you confirm that has not resurfaced with this change?

its still present.

waiting to tests to exist > gigs of tmp dir leaks

rwjblue:
> waiting to tests to exist > gigs of tmp dir leaks

LOL, we could always just tell the builder to cleanup directly (`builder.cleanup()`) and still exit...

Hopefully, we do not reintroduce that delay (many people will report the bug if we do)...

stefanpenner:
> Hopefully, we do not reintroduce that delay (many people will report the bug if we do)...

then we will fix it

PR #3804, opened 2015-04-05
increase some timeouts and prefer mocha’s inheriting timers

stefanpenner:


stefanpenner:
cancelled windows build, rebuilding

PR #3810, opened 2015-04-06
Updates to support ember-giftwrap

ef4:
This PR has two small changes:

1. A small refactor of the app tree broccoli pipeline. I improved it by doing the ES6 module conversion slightly earlier, in a place which is actually more logical, avoiding the need to merge the `_processedEmberCLITree` twice.

2. It makes the builder tasks's broccoli tree overridable.

Both of these changes are intended to support [Giftwrap](https://github.com/ef4/ember-giftwrap), which is a new general-purpose tool for packaging arbitrary sets of Ember addons for use by non-ember-cli apps. Code review of Giftwrap would also be appreciated.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -818,8 +818,8 @@ EmberApp.prototype.appAndDependencies = function() {
   });
 
   var postprocessedApp = this.addonPostprocessTree('js', preprocessedApp);
+
   var sourceTrees = [
-    external,
     postprocessedApp,
     templates,
     config
@@ -827,14 +827,34 @@ EmberApp.prototype.appAndDependencies = function() {
 
   this._addAppTests(sourceTrees);
 
+  var es6app = this._es6ifyAppModules(mergeTrees(sourceTrees, {
+    overwrite: true,
+    description: 'TreeMerger (appModules)'
+  }));
+
   var emberCLITree = this._processedEmberCLITree();
 
-  sourceTrees.push(emberCLITree);
-
-  return mergeTrees(sourceTrees, {
+  return mergeTrees([external, es6app, emberCLITree], {
     overwrite: true,
     description: 'TreeMerger (appAndDependencies)'
   });
+};
+
+EmberApp.prototype._es6ifyAppModules = function(tree) {
+  return new ES6Modules(
+    new Funnel(tree, {
+      include: [new RegExp('^' + escapeRegExp(this.name + '/') + '.*\\.js$')],
+      description: 'Funnel: App JS Files'
+    }),
+    {
+      description: 'ES6: App Tree',
+      esperantoOptions: {
+        absolutePaths: true,
+        strict: true,
+        _evilES3SafeReExports: this.options.es3Safe
+      }
+    }
+  );
 };
 
 /**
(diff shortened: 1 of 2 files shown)


stefanpenner:
tests may be fixed by: https://github.com/ember-cli/ember-cli/pull/3811

stefanpenner:
@rwjblue r?

lib/broccoli/ember-app.js
```diff
@@ -818,8 +818,8 @@ EmberApp.prototype.appAndDependencies = function() {
   });
 
   var postprocessedApp = this.addonPostprocessTree('js', preprocessedApp);
+
   var sourceTrees = [
-    external,
```
ef4: Needs review: I'm assuming here that the external tree isn't going to provide anything that matches the regexp used by `_es6ifyAppModules` below.

lib/broccoli/ember-app.js
```diff
@@ -827,16 +827,36 @@ EmberApp.prototype.appAndDependencies = function() {
...
   this._addAppTests(sourceTrees);
 
-  var emberCLITree = this._processedEmberCLITree();
+  var es6app = this._es6ifyAppModules(mergeTrees(sourceTrees, {
+    overwrite: true,
+    description: 'TreeMerger (appModules)'
+  }));
 
-  sourceTrees.push(emberCLITree);
+  var emberCLITree = this._processedEmberCLITree();
```
ef4: Likewise, I'm assuming that the processedEmberCLITree doesn't need to go through `_es6ifyAppModules`.

ef4:
I don't know why only the iojs tests are reliably failing. It doesn't look related to me.

stefanpenner:
> I don't know why only the iojs tests are reliably failing. It doesn't look related to me.
> Hide all checks

Ya, I believe its just an unrelated race, we need to dig into.

trabus:
@rwjblue you okay with merging this in for 0.2.5?

stefanpenner:
@ef4 is this something you still need for giftwrap?

stefanpenner:
This seems to have gone stale, we can revisit. But an idle PR is just noise on the tracker.

joostdevries:
@stefanpenner @ef4 Can I do anything to get this merged? I'm currently using giftwrap and it'd be better if it wouldn't need to depend on a forked ember-cli.

rwjblue:
@joostdevries - I'm not sure if this is needed any longer, and ember-giftwrap doesn't list a custom/fork of ember-cli as a requirement.

joostdevries:
@rwjblue https://github.com/ef4/ember-giftwrap/blob/master/package.json#L29 am I missing something?

rwjblue:
Well, I feel dumb :stuck_out_tongue_closed_eyes:.  We'd need to track down the reasoning here, and see how to address in master.  Things have changed quite a lot since this PR was submitted....

joostdevries:
Cool, I'll look into it day after tomorrow. Working on an Lambda-based addon giftwrapper (so we can use addons in Ember Twiddle) and since I need to bundle all node modules on every deploy I'm currently uploading 100MB to S3 for every test run :cry:...

rwjblue:
@joostdevries - You should chat with @tomdale about his lamda experiences.  Also, as an aside, I personally think that it would be better to make a generic thing like I've done in https://github.com/emberjs/ember-collection/pull/92 (which is the "core" ember-giftwrap concept), and allow folks to provide these in their own S3 buckets...

joostdevries:
I think the two should probably live side-by-side. Uploading your own app to s3 is good but quickly testing an addon without having to build it would be really awesome too. Got a POC working now (https://github.com/joostdevries/twiddle-backend). Might eventually actually run this on ECS instead of lambda though..

rwjblue:
@joostdevries - I am saying that addon authors publish their addon for twiddle consumption to S3, not that app developers use S3.  See the [demo twiddle](https://ember-twiddle.com/4556826aac0e7b2a60ef?numColumns=1&openFiles=twiddle.json%2C) that I published with that PR.  We could easily add an additional key to the `twiddle.json` format, for `addons: ['ember-collection']`.

joostdevries:
@rwjblue I understand but my aim is to make this something that just works(tm) for the majority of addons.

tomdale:
@joostdevries You may want to check out https://github.com/tomdale/lambda-packager, a tool I used for compiling npm dependencies on AWS. It's a WIP but the idea is to upload your package.json to an EC2 instance where it runs `npm install` and uploads the resulting `node_modules` directory to S3. (This also helps if you have native dependencies like `contextify` in the case of FastBoot.)

I think you could prettily easily modify it so the pipeline would be something like:

```
+----------------------+         +----------------------+        +--------------------+
|                      |         |                      |        |                    |
|  Upload App          |  +----> | Install node_modules | +----> | Lambda             |
|  (no node_modules)   |         | into app directory   |        |                    |
|                      |         |                      |        |                    |
+----------------------+         +----------------------+        +--------------------+
   Dev machine                    Packager server (EC2)
```

TLDR have an intermediary build & deploy server that does the job of building `node_modules`. Because intra-AWS S3 usage is free it should considerably cut down both time and cost.

joostdevries:
@tomdale awesome, will try it out

PR #3818, opened 2015-04-06
Updates

stefanpenner:


blueprints/app/files/bower.json
```diff
@@ -3,9 +3,9 @@
   "dependencies": {
     "ember": "1.11.1",
     "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.3",
-    "ember-cli-test-loader": "ember-cli-test-loader#0.1.4",
+    "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
```
rwjblue: Why are we downgrading? Is there a regression?

stefanpenner: i accidentally upgraded this library, whne i meant to upgrade another. 

0.1.3 is the latest version: https://github.com/ember-cli/ember-cli-test-loader/blob/master/bower.json#L4

rwjblue: Ahh, thanks for the explanation,  sorry for the noise

stefanpenner: nope, thanks for the second set of eyes.

PR #3820, opened 2015-04-06
[ENHANCEMENT] Change blueprint command options to type String to avoid nopt transformations

rodyhaddad:
Before this PR, the `--blueprint` option in the `new` and `init` commands got validated as a `gitUrl` or a `path`.

When used as a `path`, it would get resolved to an absolute path, making it point to a specific blueprint on the filesystem depending on the current working directory.

This PR changes it to a String, so it won't be modified before getting lookup'ed up by `Blueprint.lookup`

This doesn't break it being used as a `gitUrl`, because that check happens again in the [InstallBlueprint task](https://github.com/ember-cli/ember-cli/blob/38256ce796861951c6acadab0eed1e8125fc95da/lib/tasks/install-blueprint.js#L38).

It does break it from being used as a path to a blueprint on the filesystem. Instead now it's used as a blueprint in one of the regular lookup paths.

I don't believe this will affect many people, if any. I imagine people always rely on the default `app` blueprint when using `new` or `init`. 

I considered keeping it backward compatible, but the new behaviour might be what we actually want moving forward.

The only test this breaks is [this one](https://github.com/ember-cli/ember-cli/blob/38256ce796861951c6acadab0eed1e8125fc95da/tests/acceptance/new-test.js#L115-L134), which I'm for removing/modifying, but wanted someone else's opinion.

\cc @stefanpenner

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -1,11 +1,11 @@
 'use strict';
 
-var path      = require('path');
 var Command   = require('../models/command');
 var Promise   = require('../ext/promise');
 
-var SilentError      = require('../errors/silent');
-var validProjectName = require('../utilities/valid-project-name');
+var SilentError        = require('../errors/silent');
+var validProjectName   = require('../utilities/valid-project-name');
+var normalizeBlueprint = require('../utilities/normalize-blueprint-option');
 
 module.exports = Command.extend({
   name: 'init',
@@ -16,7 +16,7 @@ module.exports = Command.extend({
   availableOptions: [
     { name: 'dry-run', type: Boolean, default: false, aliases: ['d'] },
     { name: 'verbose', type: Boolean, default: false, aliases: ['v'] },
-    { name: 'blueprint', type: ['gitUrl', path], aliases: ['b'] },
+    { name: 'blueprint', type: String, aliases: ['b'] },
     { name: 'skip-npm', type: Boolean, default: false, aliases: ['sn'] },
     { name: 'skip-bower', type: Boolean, default: false, aliases: ['sb'] },
     { name: 'name', type: String, default: '', aliases: ['n'] }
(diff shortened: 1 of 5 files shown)


stefanpenner:
> I considered keeping it backward compatible, but the new behaviour might be what we actually want moving forward.

it seems like people being required to provide a relative or absolutely path, should be sufficient we dont need to resolve non-path like things.

rodyhaddad:
Sure, that sounds very reasonable.

How would we technical detect it though? Check for the presence of '/' in the blueprint name and assume it's a path?

stefanpenner:
> How would we technical detect it though? Check for the presence of '/' in the blueprint name and assume it's a path?

ya, relative or absolute.

rodyhaddad:
@stefanpenner I updated the PR and added an extra test for absolute paths.

Let me know if there's anything I should change before merging.

stefanpenner:
looking now

lib/commands/init.js
```diff
@@ -72,6 +72,10 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError('We currently do not support a name of `' + packageName + '`.'));
     }
 
+    if (blueprintOpts.blueprint[0] === '.') {
+      blueprintOpts.blueprint = path.resolve(process.cwd(), blueprintOpts.blueprint);
```
stefanpenner: at first, I thought this was a mistake. But I believe it to be correct, as you want:

```js
cd app/ && ember --blueprint=../blueprints/foo
```

to be relative the cwd, not the project root.

:+1:

lib/commands/new.js
```diff
@@ -55,6 +55,10 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(message));
     }
 
+    if (commandOptions.blueprint[0] === '.') {
+      commandOptions.blueprint = path.resolve(process.cwd(), commandOptions.blueprint);
```
stefanpenner: that being said, can we extract this to a helper function that is shared between new and init?

stefanpenner:
1 last comment, and then i feel good to merge this.

rodyhaddad:
@stefanpenner Tada!

PR #3825, opened 2015-04-07
Exclude dist/ from addon npm publishes by default

jayphelps:
Just like #3539, I've seen `dist/` published on accident in a number of addons. AFAIK, 99.99+% of addons won't want their `dist/` published as it's just their `/tests/dummy` app output.

Maybe I'm missing something?

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -1,6 +1,7 @@
 bower_components/
 tests/
 tmp/
+dist/
 
 .bowerrc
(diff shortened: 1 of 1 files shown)


jayphelps:
I guess there's more stuff that *could* be excluded as well like `config/`...I just thought this one was clearly the second worst offender next to `tmp/`..

rwjblue:
FYI - The addons config hook uses config/ directory by default.

PR #3826, opened 2015-04-07
[fixes #3712] rethrow errors in build task

### LINKED ISSUE #3712: `ember test` in CI mode does not exit on Build failed. 0.2.2
I think due to the new "display errors in browser on autoreload" functionality, `ember test` does not immediately exit when there are compilation errors in CI mode.

Workaround for now is to use `ember build -e test` and then run `ember test` if that succeeds (but involves recompiling it), but I think in CI mode the `ember test` command should fail immediately on build errors as it used to.

marcioj:
The build task was catching the failure and returning 1, which moves the promise to a fulfilled state.

The side effect of this is that in the test command the test task will run even if the build fail, and for some reason when the test task is invoked in this state the command hangs forever.

To fix this we rethrow the error so the test command won't run. And since `CLI.prototype.logError` already exit with 1 when some promise gets rejected, there is no need to return 1 within the task.

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -45,9 +45,8 @@ module.exports = Task.extend({
       })
       .catch(function(err) {
         ui.writeLine(chalk.red('Build failed.'));
-        ui.writeError(err);
 
-        return 1;
+        throw err;
       });
   }
 });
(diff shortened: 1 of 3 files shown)


rwjblue:
Seems like this might log the error to the console twice (once in the builder model via writeError and the second time at the top level error handler). Can you double check that?

stefanpenner:
Looks like he removed it from builder model. This seems good

rwjblue:
:+1: - Thanks!

PR #3830, opened 2015-04-07
Fixes #3829  

### LINKED ISSUE #3829: The 'component-test' blueprint does not support generating inside addons.
Same for goes for all `*-test` generators.

I suspect #3690 is the culprit which introduced this. Particularly the following:

[/lib/models/blueprint.js](https://github.com/ember-cli/ember-cli/blob/a18d9c0ad7741874ca37d8516546dacb6925ccaf/lib/models/blueprint.js#L734)
```javascript
Blueprint.prototype.supportsAddon = function() {
  return this.files().join().match(/__root__/);
};
```

As test generators don't include `__root__` in their path name.

martndemus:
Fixes #3829

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -781,7 +781,7 @@ Blueprint.prototype.mapFile = function(file, locals) {
   @return {Boolean}
 */
 Blueprint.prototype.supportsAddon = function() {
-  return this.files().join().match(/__root__/);
+  return this.files().join().match(/(?:__root__|tests)/);
 };
 
 /**


martndemus:
This was an easy fix for me, don't have the time to add tests right now. Might have some time tomorrow. Maybe @trabus can pick it up?

trabus:
The tests are super easy, just drop this into `tests/acceptance/addon-generate-test.js` at line 175

```
  it('in-addon component-test x-foo', function() {
    return generateInAddon(['component-test', 'x-foo']).then(function() {
      assertFile('tests/unit/components/x-foo-test.js', {
        contains: [
          "import {" + EOL +
          "  moduleForComponent," + EOL +
          "  test" + EOL +
          "} from 'ember-qunit';",
          "moduleForComponent('x-foo'"
        ]
      });
    });
  });
```

And this into `tests/acceptance/in-repo-addon-generate-test.js` at line 185
```
  it('in-repo-addon component-test x-foo', function() {
    return generateInRepoAddon(['component-test', 'x-foo', '--in-repo-addon=my-addon']).then(function() {
      assertFile('tests/unit/components/x-foo-test.js', {
        contains: [
          "import {" + EOL +
          "  moduleForComponent," + EOL +
          "  test" + EOL +
          "} from 'ember-qunit';",
          "moduleForComponent('x-foo'"
        ]
      });
    });
  });
```

trabus:
I actually need to fix this issue for files generated into the 'server' and 'blueprints' folders, so I'm creating a PR for that as well, which will include the fix for this issue. If you don't mind, I'm going to close this so we don't have to worry about merge conflicts. Thank you for the PR though, it's really appreciated!

martndemus:
:+1:

PR #3833, opened 2015-04-07
rm `install:npm`, change `install:addon` -> `install <addon>`

bantic:
As of https://github.com/ember-cli/ember-cli/commit/38cf55199a850a2189a44f4d6d21a68769d3f217 the `ember install:npm` command is gone and `ember install:addon` is now `ember install <addon>`.


stefanpenner:
thanks

PR #3836, opened 2015-04-07
Suggestion: Adding test coverage to pull requests

kellyselden:
Coveralls will add test coverage status to every PR. You can easily see when new code is untested when the coverage decreases.

This can be very helpful, or a nuisance if it is going to be ignored/not enforced.

Here is a PR example: https://github.com/kellyselden/ember-cli/pull/11

capture


stefanpenner:
rad

PR #3844, opened 2015-04-08
Update to ember-router-generator 0.3.3

omni5cience:
It no longer crashes when you have any errant Expression Statements in `Router.map()`

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "debug": "^2.1.3",
     "diff": "^1.3.1",
     "ember-cli-copy-dereference": "^1.0.0",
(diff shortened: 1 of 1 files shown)


abuiles:
@omni5cience thanks! I released a new version yesterday and merged in with other change required https://github.com/ember-cli/ember-cli/pull/3887, thanks!

abuiles:
@omni5cience also since we are using semver we can let that pick up the right version when installing the dependencies.

omni5cience:
Ah, my bad. I always seem to forget the semver range syntax.

Issue #3840, opened 2015-04-08
The 'http-mock' and 'blueprint' blueprints do not support generating inside addons


MiracleBlue:
When running `ember generate http-mock my-mock` inside an ember addon on Ember CLI 0.2.3, I get the following error: `The 'http-mock' blueprint does not support generating inside addons`

And I can't for the life of me find a way around it, after many different attempts.  Any work-arounds would be greatly appreciated, if not a real fix :)

MiracleBlue:
Now it seems even the blueprint blueprint doesn't support generating inside addons.

trabus:
It seems that I overlooked this when trying to prevent some other blueprints from being generated. I'll get on this right now, sorry for the inconvenience.

MiracleBlue:
Thanks heaps :) appreciate it.

trabus:
Related to #3829

trabus:
@MiracleBlue this will be fixed in 0.2.4. If you'd like a quick and dirty fix for this locally in the meantime, you can replace this method in your `node_modules/ember-cli/lib/models/blueprint.js` file.
```js
Blueprint.prototype.supportsAddon = function() {
  return this.files().join().match(/(?:blueprints|__root__|server|tests)/);
 };
```

MiracleBlue:
Thanks so much that will help us out :) I'll try the dirty hack and see how
it goes.

trabus:
I just realized something I did not consider, the Blueprint.supportsAddon() method will fail on custom root directories if the blueprint does not override the supportsAddon method. So if a blueprint had `files/custom-folder/__name__.js`, the same warning would come up because it's not "on the list".  

I also forgot 'public', 'lib', and a couple others that probably shouldn't be blocked. This current fix I have still won't solve the problem completely.This obviously needs some rethinking, and should likely list exclusions rather than inclusions. I'm also wondering if we should stop the blueprint at all, a warning should be sufficient (instead of a silent error). 

The main reason I threw a silent error in the first place was because an error occurs when trying to generate a route inside an addon. Anything else is generally fine as long as it's not trying to make updates to files it's expecting to be in a regular project.

@ember-cli/owners thoughts?

trabus:
I went ahead and removed the silent error, and fixed route generation inside addons, so they no longer attempt to modify the router. The fix is in #3845

trabus:
The fix for this has been merged and will be in 0.2.4, hopefully coming tomorrow.

[...]

PR #3845, opened 2015-04-08
[BUGFIX] Fix ability to generate blueprints (blueprint, http-mock, http-proxy, and tests) inside addons

### LINKED ISSUE #3829: The 'component-test' blueprint does not support generating inside addons.

trabus:
Fixes #3829 and #3840.
Generating certain blueprints (blueprint, http-mock, http-proxy, and tests) inside addons was broken by [this commit](https://github.com/ember-cli/ember-cli/commit/a18d9c0ad7741874ca37d8516546dacb6925ccaf#diff-3b6ee55eb49134b6b858dea4972b3117R735).


trabus:
This still needs some more work, as it has a few more cases it does not handle.

trabus:
I've removed the silent-error, and now allow route generation inside addons.

PR #3846, opened 2015-04-08
[BUGFIX] Prevent addon-import blueprint from generating if entity name is undefined

### LINKED ISSUE #3831: Treat main blueprint as special case / replace with addon hook
TL;DR

The main blueprint is often used by addons simply to [add bower dependencies to `bower.json`](https://github.com/emberjs/data/blob/v1.0.0-beta.16.1/lib/ember-addon/blueprints/ember-data/index.js#L12), without generating any app code. This is at odds with the very concept of 'blueprints' and 'generation', and therefore should be served by a dedicated addon hook. I propose `postInstall`. 

-------------------

After updating to v0.2.3 I started getting this error after installing addons:

```
The `ember generate` command requires an entity name to be specified. For more details, use `ember help`.
```

This is happening because of the recently added `addon-import` blueprint that runs after _every_ addon blueprint. This makes sense for the majority of blueprints, but for the main blueprint it doesn't. The main blueprint usually doesn't generate anything, and the concept of an 'entity' doesn't even make sense for it. For some reason, the first arg (after the addon name) of the `install` command [is used as the entity name](https://github.com/ember-cli/ember-cli/blob/v0.2.3/lib/tasks/generate-from-blueprint.js#L36), which is almost always omitted, resulting in the above error.

This has lead to [some](https://github.com/emberjs/data/blob/v1.0.0-beta.16.1/lib/ember-addon/blueprints/ember-data/index.js#L4-L8) [silliness](https://github.com/lytics/ember-data.model-fragments/blob/master/ember-addon/blueprints/ember-data-model-fragments/index.js#L4-L6) in order suppress the error on addon installs when the main blueprint is run. This workaround no longer helps, since the `addon-import` blueprint inherits options (i.e. entity name which is `undefined`) from the main blueprint options, and now no amount of silliness can prevent it from happening.

IMO, this speaks to the main blueprint having reached the end of it's life – it's gotten us this far, but there really should be a specific pathway/hook for adding bower components and/or just running arbitrary code on addon install without contorting blueprints to suit the purpose.

trabus:
Fixes #3831 by preventing addon-import blueprint from generating if entity name is undefined.

When running `ember install:addon ember-cli-foo` the following silent error was being triggered:
```
The `ember generate` command requires an entity name to be specified. For more details, use `ember help`.
```
This PR fixes the silent error, but #3831 may require further work to solve any further root cause.

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -14,7 +14,7 @@ module.exports = Task.extend({
   run: function(options) {
     var self = this;
     var name = options.args[0];
-    var noAddonBlueprint = ['mixin'];
+    var noAddonBlueprint = ['mixin', 'route'];
 
     var mainBlueprint  = this.lookupBlueprint(name, options.ignoreMissingMain);
     var testBlueprint  = this.lookupBlueprint(name + '-test', true);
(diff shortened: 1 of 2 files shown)


stefanpenner:
i suspect this needs a test of its own.

knownasilya:
What's the status on this?

trabus:
It's ready to go, just needs a quick review. I've been rebasing along with the attempts to get appveyor passing.

PR #3848, opened 2015-04-08
writeError now looks for filename as well as file

wagenet:
UglifyJS sets the filename but not the file. This change now gives
us line and column for UglifyJS errors.

### DIFF
--- a/lib/ui/write-error.js
+++ b/lib/ui/write-error.js
@@ -4,12 +4,13 @@ var chalk = require('chalk');
 module.exports = function writeError(ui, error){
   if (!error) { return; }
 
-  if (error.file) {
-    var file = error.file;
+  // Uglify errors have a filename instead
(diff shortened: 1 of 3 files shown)


lib/ui/write-error.js
```diff
@@ -4,8 +4,9 @@ var chalk = require('chalk');
 module.exports = function writeError(ui, error){
   if (!error) { return; }
 
-  if (error.file) {
-    var file = error.file;
+  // Uglify errors have a filename instead
+  if (error.file || error.filename) {
```
stefanpenner: Can you do `var fileName = error.file || error.fileName;` and then the conditional?

stefanpenner:
This also needs a test

wagenet:
@stefanpenner, got your tests now

stefanpenner:
restarting appveyor

PR #3852, opened 2015-04-08
Cache `bower_components` in CI

seanpdoyle:
Enable directory caching of `bower_components`

### DIFF
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -8,6 +8,7 @@ sudo: false
 cache:
   directories:
     - node_modules
+    - bower_components
 
 before_install:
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1:

stefanpenner:
kicking travis-ci

seanpdoyle:
@rwjblue @stefanpenner this actually might be a bad idea, given the issues we've had on other repos.

rwjblue:
Nearly all of the issues we have seen are related to NPM (unless I am remembering incorrectly).

Also, some large % of CI builds in a normal app or addon, will not be changing NPM/Bower deps and therefore would benefit from this caching...

PR #3598, opened 2015-03-21
[ENHANCEMENT] Replace install:addon with install, remove install:bower and install:npm

DanielOchoa:
This closes https://github.com/ember-cli/ember-cli/issues/3204.

PR #3854, opened 2015-04-08
[ENHANCEMENT] install:addon command will show a deprecation message before running the install command.

DanielOchoa:
Given that a lot of ember-cli addon projects tell you to install them through the `ember install:addon` command, I re-added the command so it can throw a friendly error that points to the now correct command `ember install`. This is related to https://github.com/ember-cli/ember-cli/pull/3598#issuecomment-90922380.

### DIFF
--- a/lib/commands/install-addon.js
+++ b/lib/commands/install-addon.js
@@ -1,0 +1,24 @@
+'use strict';
+
+var InstallCommand = require('./install');
+var chalk          = require('chalk');
+
+module.exports = InstallCommand.extend({
+  name: 'install:addon',
+  description: 'This command has been deprecated. Please use `ember install` instead.',
+  works: 'insideProject',
+
+  anonymousOptions: [
+    '<addon-name>'
+  ],
+
+  init: function() {
+    var warning  = 'This command has been deprecated. Please use `ember ';
+    warning     += 'install <addonName>` instead.';
+    this.ui.writeLine(chalk.red(warning));
+
+    if (this._super.init) {
+      this._super.init.call(this);
+    }
+  }
+});
--- a/lib/commands/install-npm.js
+++ b/lib/commands/install-npm.js
@@ -14,7 +14,7 @@ module.exports = Command.extend({
   ],
 
   run: function() {
-    var err  = 'This command has been deprecated. Please use `npm install ';
+    var err  = 'This command has been removed. Please use `npm install ';
     err     += '<packageName> --save-dev --save-exact` instead.';
     return Promise.reject(new SilentError(err));
   }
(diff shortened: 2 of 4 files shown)


stefanpenner:
:+1:

rwjblue:
Typically, a deprecation will still "work" with a warning. In this case, it seems that we just throw a SilentError and the user needs to run the alternate command.  

I would prefer that this extended the current `install` command, and added the deprecation before calling `_super` in `init`.  That way it doesn't duplicate code to make it work, and isn't annoying to consumers who have to retype their command.

What do y'all think?

DanielOchoa:
You are right, it is not a deprecation but simply a silent error. I like where you are going with this, I'll make the necessary change to extend from the install command.

DanielOchoa:
Interesting. For some reason `init` gets called twice. It's not being instantiated two times, it's only the init function being called twice. 

Do you have any input on this @rwjblue ?

raycohen:
any movement on this?  we're having to put two different sets of install instructions on our repos

stefanpenner:
> any movement on this?

kicking travis

> we're having to put two different sets of install instructions on our repos

Why? Or are you catering to old versions of ember-cli?

lib/commands/install-addon.js
```diff
@@ -0,0 +1,24 @@
...
+  anonymousOptions: [
+    '<addon-name>'
+  ],
+
+  init: function() {
+    var warning  = 'This command has been deprecated. Please use `ember ';
+    warning     += 'install <addonName>` instead.';
+    this.ui.writeLine(chalk.red(warning));
+
+    if (this._super.init) {
```
stefanpenner: when does this not have super?

lib/commands/install-addon.js
```diff
@@ -0,0 +1,24 @@
...
+      this._super.init();
```
stefanpenner: this._super.init.call(this);

stefanpenner:
> Interesting. For some reason init gets called twice. It's not being instantiated two times, it's only the init function being called twice.
> 
> Do you have any input on this @rwjblue ?

this is most likely the result of us still being on a broken version of core-object. https://github.com/ember-cli/ember-cli/pull/3056 sorta stalled, but we should try to re-ignite that effort.

cc @jgwhite

jgwhite:
@stefanpenner yes, yes we should :guilty_face:

DanielOchoa:
I figured as much :smile_cat:

raycohen:
@stefanpenner yes, catering to both...

lib/commands/install-addon.js
DanielOchoa: Init is undefined on the command being extended.

DanielOchoa:
I'll update the pull request to reflect master ..

lib/commands/install-addon.js
stefanpenner: i don't think we need `super` in this simple deprecation stub

PR #3851, opened 2015-04-08
Use shorthand ES6 syntax for addon -> app re-exports

jayphelps:
Duplicate of #3739 but GitHub [won't allow you to reopen](https://github.com/isaacs/github/issues/361) PRs which have been force pushed.

PR #3856, opened 2015-04-08
Use shorthand ES6 re-export for addon-imports as well, which landed in #3690

jayphelps:
Continues work from #3851


stefanpenner:
thanks

jayphelps:
appveyor on Windows seems to be having an aneurysm...

stefanpenner:
ya, we need to rely on more async rimrafing...

jayphelps:
@stefanpenner I've kicked appveyor again and still no dice. Any thoughts?

stefanpenner:
The issue is likely unrelated as this seems to be failing regardless. I'll dig into the windows issue later today

PR #3862, opened 2015-04-09
`ember install:addon` -> `ember install`

folz:
Updating the asset compilation docs to match the new ember-cli install argument.


stefanpenner:
thanks

Issue #3855, opened 2015-04-08
Update docs to change `ember install:addon` to `ember install`


searls:
It sounds like the `ember install:addon` command has been deprecated in favor of just saying `ember install`, this should be updated in the docs.

Mentioned in #3850

visualjeff:
I noticed this too.  In my case I'm attempting to development my addon locally using npm link and I'm running into this issues where addon's are being removed from node_modules when running an ember build.

> npm link <addon name>
> npm install
> ember build <== Here is where addon disappears.  

build dies because it can't find addon index.js file in addon module.  Somehow it has the path.  But the addon has disappeared form node_modules.

I'm using version 0.2.3 of the ember-cli.

visualjeff:
I should open my own issue.

clekstro:
Can this be closed now?  The documentation PR was already merged.

searls:
I don't see why not.

PR #3865, opened 2015-04-09
Updates to reflect addon install change.

willrax:
As picked up in https://github.com/ember-cli/ember-cli/issues/3855. I've updated the docs to reflect the change from 

`ember install:addon <package-name>`

to 

`ember install <package-name>`


willrax:
I also removed references to install:bower and install:npm

stefanpenner:
thanks :)

PR #3868, opened 2015-04-10
Allow blueprints to omit entity names

### LINKED ISSUE #3831: Treat main blueprint as special case / replace with addon hook

davewasmer:
Change blueprints to allow for missing entity names. Entity names are required by default on all blueprints except for an addon's main blueprint (which will never have an entity name).

This setting can be modified in a blueprint's definition via the `entityNameRequired` flag, or at runtime by passing the `entityNameRequired` flag on the options hash to the blueprint's install method. The latter approach is how the main blueprints are exempted automatically.

This should get rid of the need for empty `normalizeEntityName` hooks such as https://github.com/poetic/ember-cli-cordova/blob/master/blueprints/cordova-starter-kit/index.js#L9. This should also fix #3831 and #3846 in a slightly more robust manner.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -272,6 +272,7 @@ Blueprint.prototype.constructor = Blueprint;
 
 Blueprint.prototype.availableOptions = [];
 Blueprint.prototype.anonymousOptions = ['name'];
+Blueprint.prototype.entityNameRequired = true;
 
 /**
   Used to retrieve files for blueprint. The `file` param is an
@@ -309,8 +310,8 @@ Blueprint.prototype.srcPath = function(file) {
   @param {String} entityName
   @return {null}
 */
-Blueprint.prototype.normalizeEntityName = function(entityName) {
-  if (!entityName) {
+Blueprint.prototype.normalizeEntityName = function(entityName, required) {
+  if (!entityName && required) {
     throw new SilentError('The `ember generate` command requires an ' +
                           'entity name to be specified. ' +
                           'For more details, use `ember help`.');
@@ -425,9 +426,9 @@ Blueprint.prototype._checkForPod = function(verbose) {
   @method _normalizeEntityName
   @param {Object} entity
 */
-Blueprint.prototype._normalizeEntityName = function(entity) {
+Blueprint.prototype._normalizeEntityName = function(entity, required) {
   if (entity) {
-    entity.name = this.normalizeEntityName(entity.name);
+    entity.name = this.normalizeEntityName(entity.name, required);
   }
 };
 
@@ -485,7 +486,12 @@ Blueprint.prototype.install = function(options) {
       ' changes will be written.'));
   }
 
-  this._normalizeEntityName(options.entity);
+  var entityNameRequired = this.entityNameRequired;
+  if (typeof options.entityNameRequired !== 'undefined') {
+    entityNameRequired = options.entityNameRequired;
+  }
+  
+  this._normalizeEntityName(options.entity, entityNameRequired);
   this._checkForPod(options.verbose);
   this._checkInRepoAddonExists(options.inRepoAddon);
 
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -16,7 +16,7 @@ module.exports = Task.extend({
     var packageName      = options['package'];
     var extraArgs        = options.extraArgs || [];
     var blueprintOptions = options.blueprintOptions || {};
-    
+
     var npmInstall = new this.NpmInstallTask({
       ui:         this.ui,
       analytics:  this.analytics,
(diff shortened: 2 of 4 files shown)


stefanpenner:
Seems maybe like this should be opt in not opt out. @rwjblue thoughts?

davewasmer:
My two cents: feels like it depends on how popular non-entity generating blueprints are. My gut feel is that, at least so far, most blueprints are entity generators. The notable exception here is the main blueprint for an addon, but that is handled here.

Perhaps an alternative approach would be to allow for required `anonymousOptions`, since thats what the entity name is treated as in the generate command. Although I suppose that doesn't really answer the question of what the default setting should be.

stefanpenner:
Something odd about this needing to be a flag or method detection in the first place. There must be a more reasonable way to just detect the right thing

davewasmer:
Agreed - I get the same feeling, but I can't see an elegant way to make this totally automatic.

The only thing I can think of is: we make entity name optional, and then make sure we throw with a helpful error message if, during the template processing, we encounter a template or filename attempting to use the entity name when it wasn't provided.

The drawbacks: it probably adds some complexity to the CLI, and could leave the project directory in a dirty state (i.e. you hit an error halfway through the template processing). FWIW, my [other PR](https://github.com/ember-cli/ember-cli/pull/3786) that makes template errors throw would leave the directory in a dirty state as well. I suppose we could attempt to do a transactional blueprint install, but that's an entirely different can of worms.

stefanpenner:
The blueprint is more or less transactional today and initially it was entirely transaction. Over time features added made it less transactional. I would love to get this fixed. It should be very hard

davewasmer:
How would you approach those kinds of transactions? Some kind of bulk "backup everything to a tmp dir" at the start (slow), or actually try to track each action by the blueprint and rollback intelligently (complex)? Or perhaps I'm missing a third approach.

That might be something I could spike out a PR for, depending on the approach.

stefanpenner:
> or actually try to track each action

track by each item. `FileInfo` where created to handle this.

stefanpenner:
`FileInfo` already can render without flushing: https://github.com/ember-cli/ember-cli/blob/master/lib/models/file-info.js#L82-L95

stefanpenner:
i suspect if it rendered here: https://github.com/ember-cli/ember-cli/blob/master/lib/models/file-info.js#L45 it would fail fast.

davewasmer:
After playing around a bit with this, I'm not sure rendering early is the best approach, although perhaps I'm missing something.

If the goal here is to provide an easy way for blueprints to omit entity names, rendering early would work. Blueprints that don't need entity names simply wouldn't throw (it's like duck-typing, with exceptions!). 

Meanwhile, blueprints that **do** rely on entity names would throw on render, and the early rendering approach would need to intelligently catch errors and display a helpful error message (i.e. "You forgot an entity name!" rather than "Cannot call split of undefined in leastHelpfulStackTraceEver.js:2934:23")

The problem is that, for blueprints that **do** rely on entity names, they can use entity names all over the place, not just templates. There's no way I can see to determine that the error thrown comes from a missing entity name. It's not just a simple `error.match(/ReferenceError: entityName is undefined/)`.

As long as entity names are supported as first class citizens of blueprints (and not "just another argument"), I don't see a straightforward way to do this without a flag of some kind.

@stefanpenner am I missing something from your suggestion? Any thoughts on this?

stefanpenner:
this seems to have gone stale, we can revisit at a later date.

PR #3869, opened 2015-04-10
[BUGFIX] Use posix path for in-repo-addons in package.json

### LINKED ISSUE #3847: In-repo-addon generated on Windows is not being used on *nix based operating systems.
When you generate an in-repo-addon on Windows by using a command:

```ember g in-repo-addon addon-name```

It adds a section to the package.json:

```
"ember-addon": {
    "paths": [
      "lib\\addon-name"
    ]
  }
```

I have realized that path ```"lib\\addon-name"``` is not valid on *nix operating systems and the postBuild function of my addon wasn't being executed. After I've changed it to ```"lib/addon-name"``` it worked on both Windows and *nix.

### PR DESCRIPTION
**Pull request by trabus**
Fixes #3847 using a posix style path in package.json for in-repo-addons for Windows (and all other environments).

### DIFF
--- a/blueprints/in-repo-addon/index.js
+++ b/blueprints/in-repo-addon/index.js
@@ -22,7 +22,7 @@ module.exports = {
     var packagePath = path.join(this.project.root, 'package.json');
     var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
     var name        = stringUtil.dasherize(options.entity.name);
-    var newPath     = path.join('lib', name);
+    var newPath     = ['lib', name].join('/');
     var paths;
 
     contents['ember-addon'] = contents['ember-addon'] || {};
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: blueprints/in-repo-addon/index.js [hunk 1]
```diff
@@ -22,7 +22,7 @@ module.exports = {
     var packagePath = path.join(this.project.root, 'package.json');
     var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
     var name        = stringUtil.dasherize(options.entity.name);
-    var newPath     = path.join('lib', name);
```
abuiles: @trabus probably we can still use path, but make explicit that we want posix `path.posix.join`.

abuiles:
@trabus your commit will be applied through #4056

PR #3881, opened 2015-04-11
Added CLI banner adjustable to terminal width

bevacqua:
screen shot 2015-04-10 at 22 26 00


ikhattab:
Nice one :+1:

stefanpenner:
this is really cool, but as cool as it is, it does seem somewhat obnoxious. I would love to hear others thoughts

kellyselden:
If it prints every time you run a command, you would lose context to your above work pretty quickly.

lukemelia:
Can this be an addon?

stavarotti:
+1 for an addon or some command intentionally invoked by the user.  I often depend on context when generating or starting (`ember s`) and while nice looking, would quickly get annoyed with the placement.

bevacqua:
How about running it once per version. As in saving a file with the package version and whenever the CLI runs, if the version in file matches the one in the package, it doesn't print out

abuiles:
Let's keep this as an addon or maybe just replace the Tomster when you `ember new`, running always this can be really annoying for some people.

PR #3882, opened 2015-04-11
Removes bower install command.

willrax:
I think this command was deprecated in https://github.com/ember-cli/ember-cli/pull/3598. The npm command was removed but it looks like the bower one was missed.


abuiles:
@willrax thanks!

PR #3883, opened 2015-04-11
more friendly descriptions in test blueprints

eccegordo:
This commit allows for better test descriptions when generating tests.

In this commit I have updated all the built in test blueprints to have more friendly descriptions.

Unit | Components | x-foo
Unit | Routes | foo
Unit | Controllers | foo

etc.

I have defined a new local variable
friendlyDescription (or friendlyPrefix where applicable)

This change serves a few benefits

1.) Make qunit module pull down menu easier to scan with eyes.

2.) Implicity identifies which tests are unit test

3.) Allows tests to be grouped according to their type, routes are together, controllers, components, etc.

4.) Makes it easier to filter tests by type, for example now possible to just select component tests, or just route tests, or just unit tests

5.) Provides a concrete example of how to use the "locals" hash if a user decides to customize the blueprint.

6.) Reinforces where to find the test file when following conventions.

A couple open questions and potential impactful changes:

should adapters use
dasherizedModuleName  instead of classifiedModuleName

should helpers use
dasherizedModuleName  instead of classifiedModuleName

should initializers use
dasherizedModuleName  instead of classifiedModuleName

should mixins use
dasherizedModuleName  instead of classifiedModuleName

should utils always use
camelizedModuleName

I converted to dasherizedModuleName which seems more consistent when printing module(...) which doesn't seem to take a description parameter

e.g.
module('<%= friendlyPrefix %> <%= dasherizedModuleName %>');


eccegordo:
Here is an example of the qunit pull down menu
blueprint_changes_qunit

Pipes seemed like a reasonable separator.

Welcome feedback if this has unintended side affects.

stefanpenner:
restarting travis

stefanpenner:
@eccegordo +1

eccegordo:
yeah that makes sense too, the hierarchical menus

stefanpenner:
@eccegordo tests seem unhappy, can you take a look?

eccegordo:
@stefanpenner OK looking at the tests, the model generators with attributes seem to be broken.
It appears that the `needs` array is not being generated. Not sure why. 

In the blueprint I am using the following pattern below. Note the friendlyDescription local var I am trying to insert.

```
moduleForModel('<%= dasherizedModuleName %>', '<%= friendlyDescription %>', {
  // Specify the other units that are required for this test.
<%= typeof needs !== 'undefined' ? needs : '' %>
});
```
e.g.
https://github.com/eccegordo/ember-cli/blob/master/blueprints/model-test/files/tests/unit/__path__/__test__.js#L6

Any tips?

cc @trabus @jgwhite

eccegordo:
Hmm where does this `needs` variable come from?

Am I clobbering it by defining a `locals` property in the index.js file for the blueprint?

https://github.com/eccegordo/ember-cli/blob/master/blueprints/model-test/index.js#L6

jgwhite:
@eccegordo `needs` comes form the [main `model` blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/model/index.js#L15-L46). When generating the model blueprint, the `generateFromBlueprint` task lookups up the corresponding test blueprint then copies over the `locals` method if—and only if—the test blueprint does not define it:

```js
if (testBlueprint.locals === Blueprint.prototype.locals) {
  testBlueprint.locals = function(options) {
    return mainBlueprint.locals(options);
  };
}
```

https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/generate-from-blueprint.js#L58-L62

I suggested tweaking the model-test blueprint like this:

```js
var ModelBlueprint = require('../model');

module.exports = {
  description: 'Generates a model unit test.',

  locals: function(options) {
    var result = ModelBlueprint.locals.apply(this, arguments);

    result.friendlyDescription = 'Unit | Models | ' + options.entity.name;

    return result;
  }
};
```

This invalidates a couple of tests that expect `needs` not to appear, but I think it‘s reasonable for `needs` to be `[]` in these cases.

eccegordo:
Thanks @jgwhite makes sense.

So now I am seeing another test failure but not sure it is related to my changes.

```
1) Acceptance: express server restart Server restarts successfully on copy1:
```

I merged upstream master and pushed again to my repo.

What's the best process here? Should I fork master again and copy my changes in as a new commit and PR?

When I run tests locally they appeared to all be passing.

jgwhite:
@eccegordo could you rebase and squash down to one commit?

The test failure was just on iojs on Travis. It may be a race condition or intermittent effect. I’ve restarted that build to be sure.

jgwhite:
@eccegordo I like the new test module labels you’re proposing and like pipes as a separator. Do you think we should revise the acceptance test labels too while you’re working on this area?

I was thinking:

- Adopt the pipes
- Humanise the label

```
$ ember g acceptance-test some-cool-feature

module('Acceptance | Some cool feature', function() {
```

eccegordo:
Yeah, that makes sense to me. I will review those blueprints too.

If there any other blueprint files you think of let me know. 

I am going to do some more changes and commit as clean single commit.

Is there a more global way to configure this or should I just continue with the "friendly" local variables.

It might make sense to make this somewhat configurable, perhaps something like 

```
humanizeTestDescription
```

Is there a single place to define this? Or does it have to be defined in several specific blueprint files?

jgwhite:
Adding a `humanize` helper to https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/string.js seems like a useful addition.

```js
humanize('some-cool-string') // => 'Some cool string'
```

I’m nervous of introducing clutter to the codebase but unified test naming functions would also be useful.

Something like this?

```js
var testNames = require('ember-cli/lib/utilities/test-names');

testNames.unit('model', 'foo-bar')        // => 'Unit | Models | FooBar'
testNames.acceptance('some-cool-feature') // => 'Acceptance | Some cool feature'
```

I apologise for the very poor naming in that example. I’m sure you could think of something much better.

@stefanpenner seem reasonable?

eccegordo:
Only thing I can think of is that adding humanize in the string utilities might be confusing if it implies a different type of inflector behavior, ala the rails `humanize`

http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-humanize

But I think we can come up with good name. Was just thinking about how to make this reasonably DRY.

eccegordo:
This PR is superseded by #3945
Closing.

PR #3891, opened 2015-04-13
Fixes problem in initializer tests generated in addons

### LINKED ISSUE #3886: Generated initializer test fails with undefined function 'initialize' in addon in 0.2.3
Example of minimal addon here:
https://github.com/Twinkletoes/ember-cli-browser-update

Initializer test fails with
```
Died on test #1     at http://localhost:7357/assets/dummy.js:276:9
    at mod.state (http://localhost:7357/assets/vendor.js:150:29)
    at tryFinally (http://localhost:7357/assets/vendor.js:30:14)
    at requireModule (http://localhost:7357/assets/vendor.js:148:5)
    at Object.TestLoader.require (http://localhost:7357/assets/test-loader.js:29:9)
    at Object.TestLoader.loadModules (http://localhost:7357/assets/test-loader.js:21:18)
    at Function.TestLoader.load (http://localhost:7357/assets/test-loader.js:40:24): undefined is not a function@ 7 ms
Source: 	
TypeError: undefined is not a function
    at Object.<anonymous> (http://localhost:7357/assets/dummy.js:277:20)
    at Object.Test.run (http://localhost:7357/assets/test-support.js:3443:28)
    at http://localhost:7357/assets/test-support.js:3572:11
    at process (http://localhost:7357/assets/test-support.js:3131:24)
    at begin (http://localhost:7357/assets/test-support.js:3176:2)
    at http://localhost:7357/assets/test-support.js:3192:4
```

Initializer in question is:
```
// in initializers/browser-update.js
export function initialize(/* container, application */) {
  // application.inject('route', 'foo', 'service:foo');
  console.log('initialized');
}

export default {
  name: 'browser-update',
  initialize: initialize
};
```

marcioj:
See [this comment](https://github.com/ember-cli/ember-cli/issues/3886#issuecomment-92135648) for further info
Closes #3886

### DIFF
--- a/blueprints/helper-addon/files/__root__/__path__/__name__.js
+++ b/blueprints/helper-addon/files/__root__/__path__/__name__.js
@@ -1,0 +1,1 @@
+export { default, <%= camelizedModuleName %> } from '<%= modulePath %>';
--- a/blueprints/helper-addon/index.js
+++ b/blueprints/helper-addon/index.js
@@ -1,0 +1,3 @@
+/*jshint node:true*/
+
+module.exports = require('../addon-import');
--- a/blueprints/helper-test/files/tests/unit/helpers/__name__-test.js
+++ b/blueprints/helper-test/files/tests/unit/helpers/__name__-test.js
@@ -1,6 +1,6 @@
 import {
   <%= camelizedModuleName %>
-} from '../../../helpers/<%= dasherizedModuleName %>';
+} from '<%= dependencyDepth %>/helpers/<%= dasherizedModuleName %>';
 import { module, test } from 'qunit';
 
 module('<%= classifiedModuleName %>Helper');
--- a/blueprints/helper-test/index.js
+++ b/blueprints/helper-test/index.js
@@ -1,5 +1,12 @@
 /*jshint node:true*/
 
+var getDependencyDepth = require('../../lib/utilities/get-dependency-depth');
+
 module.exports = {
-  description: 'Generates a helper unit test.'
+  description: 'Generates a helper unit test.',
+  locals: function(options) {
+    return {
+      dependencyDepth: getDependencyDepth(options)
+    }
+  }
 };
--- a/blueprints/initializer-addon/files/__root__/__path__/__name__.js
+++ b/blueprints/initializer-addon/files/__root__/__path__/__name__.js
@@ -1,0 +1,1 @@
+export { default, initialize } from '<%= modulePath %>';
--- a/blueprints/initializer-addon/index.js
+++ b/blueprints/initializer-addon/index.js
@@ -1,0 +1,3 @@
+/*jshint node:true*/
+
+module.exports = require('../addon-import');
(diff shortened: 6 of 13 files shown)


stefanpenner:
Something else seems funny if this is needed. You mentioned the reExport?

marcioj:
@stefanpenner yes, the re-export in addon-import blueprint just works for default exports.
So `import { initialize } from 'path/to/app-initializer'` isn't working.
I believe that `export * from 'path/to/addon-file'` is what we want, but isn't available in esperanto yet based on esperantojs/esperanto#54

stefanpenner:
I'm confused why this won't work with property exports. Let me grab my laptop.

stefanpenner:
are the properties being marked as non-enumerable: https://github.com/ember-cli/ember-cli/blob/18d377b264859548f41aba6c3ea2015b90978068/lib/utilities/reexport-template.js#L4-L6 ?

marcioj:
Interesting, I didn't know about this reexport. But I believe it's not being used in this situation.

The reexport I'm talking about is [this one](https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon-import/files/__root__/__path__/__name__.js)

To reproduce this problem you can create a new addon, generates a new initializer, start the server and open `localhost:4200/tests`. You will notice that the tests for the newly created initializer aren't passing because there is an undefined function. Below the related files:

The initializer declaration. Everything ok.

**addon/initializers/foo.js**
```js
export function initialize(/* container, application */) {
  // application.inject('route', 'foo', 'service:foo');
}

export default {
  name: 'foo',
  initialize: initialize
};
```

Re-exporting the initializer. Notice that just the default export (foo) is re-exported as default,
the `initialize` function is not propagated.

**app/initializers/foo.js**
```js
import foo from 'sample-addon/initializers/foo';

export default foo;
```

Import `initialize` from foo initializer, but the initialize function isn't available and the test will crash.

**tests/unit/initializers/foo-test.js**
```js
import Ember from 'ember';
import { initialize } from '../../../initializers/foo';
import { module, test } from 'qunit';

// ...
```

stefanpenner:
I'm fine with this solution, we may as well remove the non default export from initializers generated entirely.

stefanpenner:
although i suspect helper has the same problem.

https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon-import/index.js could  be made more intelligent, and also do ` export { default, otherName} from 'module';`

so in this case it would have been `export { default, initialize } from '....';`

marcioj:
@stefanpenner updated to export the additional functions in initializer and helper blueprints.

While fixing this issue a noticed another problem, previously running `ember g initializer foo/bar` aren't considering that the dependency was being generated one level deep when importing in the tests. The correct is `import { initialize } from '../../../../initializers/foo/bar';` instead of `import { initialize } from '../../../initializers/foo/bar';`. So I added a variable called `dependencyDepth` in both helper and initializer tests to correctly import the dependencies.

marcioj:
@stefanpenner is there anything I can improve here?

stefanpenner:
looks great thank @marcioj

Twinkletoes:
Thanks everyone :)

marcioj:
:+1:

PR #3895, opened 2015-04-14
[Enhancement] PhantomJS 2.0 running on travis-ci

truenorth:


rwjblue:
We should not need to install it, Travis is installing both phantom 1.9 and 2.0 on their VM's (I'll try to find the issue with details).

rwjblue:
> PhantomJS 2.0.0 is available as /usr/local/phantomjs-2.0.0/bin/phantomjs. 1.9.8 remains the default on $PATH

See https://github.com/travis-ci/travis-cookbooks/commit/e38732de1c9bcc2b345772e38196e9ced1f551a8 (which I see you commented on actually!?!?), and documented in http://docs.travis-ci.com/user/build-environment-updates/2015-04-09/. 

I believe that we just need to add the directory to `$PATH`.

truenorth:
Thanks @rwjblue. Not sure how I commented on that and missed it lol!

stefanpenner:
awesome!

rwjblue:
@stefanpenner - We need to make sure to call this out explicitly in the release notes (both in the detailed list section and the "these changes are required for this version" section.  Folks should be ready to update their local environments to use the same version of phantom (it would suck to use 1.9.8 locally but test in Travis against 2.0).

rwjblue:
HUGE :+1: to this though.

stefanpenner:
:+1:

ef4:
Sadly it looks like Travis rolled back the update that provided Phantom 2. So for now this PATH update doesn't do anything. When they finally release it again, people will suddenly get the new version.

truenorth:
@ef4 Yep they keep rolling back their updates. The workaround for now is to download and install a phantomjs 2.0 binary 

```yaml
before_install:
  - mkdir travis-phantomjs
  - wget https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2
  - tar -xvf $PWD/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -C $PWD/travis-phantomjs
  - export PATH=$PWD/travis-phantomjs:$PATH
  - "npm config set spin false"
  - "npm install -g npm@^2"

```

raycohen:
Seeing no action from Travis on this.  Perhaps it's time to bless the workaround and remove the useless `PATH` line?

stefanpenner:
cc @drogus

runspired:
Travis does seem to be actively encouraging manual installation at the moment.

drogus:
Sorry for the problems with rolled back updates. I'll check what's the status of the image with phantomjs 2.0 and let you know.

drogus:
It seems that the recipe for 2.0 is [on master in cookbooks](https://github.com/travis-ci/travis-cookbooks/blob/master/cookbooks/travis_phantomjs/recipes/2.rb), but it's not executed. I'm not familiar with our current cookbooks system, so I'll clarify with someone who does when they get up.

drogus:
I have an update regarding the situation. Cookbooks that are used for ubuntu precise builds are on `precise-stable` branch and there's indeed no phantomjs 2.0. At the moment we're transitioning from BlueBox to GCE, which takes a lot of effort, so even if we add 2.0 support for ubuntu precise soon, it's uncertain when will it land on VMs.

A workaround would be to use ubuntu trusty or use a manual installation method mentioned above. The downside of using trusty is that it has longer startup time than precise (which uses docker).

There could be also a nicer way of abstracting the setup, ie. I think we could add a simple addon, which would allow you to choose phantomjs version just like so:

```
addons:
  phantomjs: 2.0.0
```

A nice thing about the addon is that we could make it work regardless of the changes on VMs and it could be deployed rather soon. Let me know what do you think about something like that.

stefanpenner:
Can you link docs on how we can create such an add on

drogus:
@stefanpenner it needs to be created in https://github.com/travis-ci/travis-build, unfortunately we don't allow any third party addons at the moment. What I meant by that is that I could create such an addon for you later this week. That said, if anyone wants to give it a shot, you can take a look at firefox addon, I believe that phantomjs addon would be similar (with an exception of checking if a given version is already there, for example there's no need to install 2.0 on trusty): https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/addons/firefox.rb

runspired:
I use Travis with phantom 2 with this

```
---
language: node_js
node_js:
  - "4.2.4"

sudo: required

cache:
  directories:
    - node_modules

env:
  - EMBER_TRY_SCENARIO=default
  - EMBER_TRY_SCENARIO=ember-release
  - EMBER_TRY_SCENARIO=ember-beta
  - EMBER_TRY_SCENARIO=ember-canary

matrix:
  fast_finish: true
  allow_failures:
    - env: EMBER_TRY_SCENARIO=ember-canary

before_install:
  - wget https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2
  - tar -xjf phantomjs-2.0.0-ubuntu-12.04.tar.bz2
  - sudo rm -rf /usr/local/phantomjs/bin/phantomjs
  - sudo mv phantomjs /usr/local/phantomjs/bin/phantomjs
  - npm config set spin false
  - npm install -g npm@^2

install:
  - npm install -g bower
  - npm install
  - bower install
  - phantomjs --version

script:
  - ember try $EMBER_TRY_SCENARIO test
```

PR #3909, opened 2015-04-15
test blueprints now use consistent, less-opinionated import style

jayphelps:
This is subjective..but I personally think this style change is the less opinionated of the two. Also--not all blueprints are using the newline-pattern, so this would make it consistent.

yeah...I know. bikeshedding here...Feel free to hate me.

xiev2tn

### DIFF
--- a/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
@@ -1,8 +1,5 @@
 import Ember from 'ember';
-import {
-  module,
-  test
-} from 'qunit';
+import { module, test } from 'qunit';
 import startApp from '<%= dasherizedPackageName %>/tests/helpers/start-app';
 
 var application;
--- a/blueprints/adapter-test/files/tests/unit/__path__/__test__.js
+++ b/blueprints/adapter-test/files/tests/unit/__path__/__test__.js
@@ -1,7 +1,4 @@
-import {
-  moduleFor,
-  test
-} from 'ember-qunit';
+import { moduleFor, test } from 'ember-qunit';
 
 moduleFor('adapter:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', {
   // Specify the other units that are required for this test.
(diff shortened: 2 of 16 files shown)


stefanpenner:
> yeah...I know. bikeshedding here...Feel free to hate me.

we should have a limit, i dont want 10 import long statements. either 2 or 3 should likely be the limit.

jayphelps:
Note sure why [travis iojs failed](https://travis-ci.org/ember-cli/ember-cli/jobs/58662505#L1884), I didn't change that file. I kicked it to try again..

stefanpenner:
* needs rebase
* can you reply to my comment?

jayphelps:
@stefanpenner 

> we should have a limit, i dont want 10 import long statements. either 2 or 3 should likely be the limit.

I agree. I personally have been limiting it to the historical 80 character ruler, give or take a few characters; because name lengths vary so saying "max of 3" could be really short or really long. But I'm totally cool with a hard import number limit..just devil's advocating here.

Also--I rebased and haven't a clue why iojs is failing [on this seemingly unrelated test](https://travis-ci.org/ember-cli/ember-cli/jobs/59676212#L1922) but I see [similar iojs errors in other merges](https://travis-ci.org/ember-cli/ember-cli/jobs/59377718#L1851) so I'm guessing you're aware of what's up.

gif

abuiles:
I prefer too what we currently have, see the following gist, https://gist.github.com/abuiles/7aa807f0cf9968e54b1f/revisions

multi-lines allow me to follow easily when I'm looking at diffs, If I want to add a new import I just pull a new line and is easier for people to review what exactly changed. Also it helps to keep the 80 columns rule (now I'm bike-shedding)

jayphelps:
@abuiles I totally agree there are diff benefits, this argument can be applied to many things we don't do. Off the top of my head, parameters: `function (first, second, third)`. My _guess_ (from looking at existing libs) is that people generally import on a single line for most imports, which is why I went with that opinion.

That said, I really just want it to be consistent so I don't have to immediately change it _sometimes_ but _sometimes_ not.

jayphelps:
@stefanpenner rebased.

abuiles:
@jayphelps thanks!

PR #3913, opened 2015-04-16
Remove podModulePrefix from app.js

knownasilya:
I think this was left over after removing `podModulePrefix` from the config. If not, just close this PR.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -9,7 +9,6 @@ Ember.MODEL_FACTORY_INJECTIONS = true;
 
 App = Ember.Application.extend({
   modulePrefix: config.modulePrefix,
-  podModulePrefix: config.podModulePrefix,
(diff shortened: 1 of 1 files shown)


abuiles:
@knownasilya thanks!

trabus:
We're going to revert this for the time being, until we can ensure that podModulePrefix is no longer needed. With the current state of pods, we still need to support it.

knownasilya:
Sounds good, I think you'd also have to bring back the podModulePrefix in the config/environment.js file

trabus:
> Sounds good, I think you'd also have to bring back the podModulePrefix in the config/environment.js file

Actually, that part is fine, as it is optional in the config, the resolver handles the case when it's not defined. The problem was that when people did define it, the wiring was no longer there.

knownasilya:
Ah, makes sense.

PR #3915, opened 2015-04-16
Always create addon trees when developing an addon

### LINKED ISSUE #3906: [0.2.3] Unit test blueprint mistaken as actual test
I've only tried this in an addon environment. 

Step 1: `ember generate blueprint foo-test`
Step 2: Visit `/tests`
Step 3: TestLoader Failures: dummy/blueprints/foo-test: could not be loaded

screen shot 2015-04-15 at 12 00 44 pm

***Update*** It also happens in an app environment.

marcioj:
If we create a new addon and start the server `includedModules()` will return an empty object, since there isn't any module, so the addon tree is skipped.
But if later we add a new file, for instance using `ember g initializer foo`, these files aren't included in the pipeline because no tree was created initially and tests will fail with `Could not find module ember-simple-validate/initializers/foo imported from dummy/initializers/foo`

Fixes #3906

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -490,7 +490,7 @@ Addon.prototype.compileTemplates = function(tree) {
 Addon.prototype.compileAddon = function(tree) {
   this._requireBuildPackages();
 
-  if (Object.keys(this.includedModules()).length === 0) {
+  if (!this.isDevelopingAddon() && Object.keys(this.includedModules()).length === 0) {
     return;
   }
 
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "strip-ansi": "^2.0.1",
     "symlink-or-copy": "^1.0.1",
     "temp": "0.8.1",
-    "testem": "^0.8.0-0",
+    "testem": "^0.8.2",
     "through": "^2.3.6",
     "tiny-lr": "0.1.5",
     "walk-sync": "0.1.3",
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -175,4 +175,39 @@ describe('Acceptance: addon-smoke-test', function() {
       handleError(error, 'npm');
     });
   });
+
+  function wait(time) {
+    return new Promise(function(resolve) {
+      setTimeout(function() {
+        resolve();
+      }, time);
+    });
+  }
+
+  it('doesn\'t fail to build new files', function() {
+    var testemOutput = '';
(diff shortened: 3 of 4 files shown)


marcioj:
Travis isn't helping today: `sh: 1: istanbul: not found` :(
Not sure if I'm doing something wrong here

kellyselden:
I added the istanbul coverage, but I don't see anything that would cause this to stop working.

marcioj:
Thanks for the reply @kellyselden.
I believe this is related with the travis ci cache.
Could you please cleanup the cache for this branch and restart the build?

stefanpenner:
@marcioj i purged the cache, it seemed to solve this for other PR :)

marcioj:
Thanks @stefanpenner!
Unfortunately my tests aren't passing in travis, I'll check it out tonight.

stefanpenner:
@marcioj thanks :)

marcioj:
After some debugging I found the issue. Looks like the `process.env.EMBER_ADDON_ENV` wasn't  being properly cleaned in [addon-test.js](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/addon-test.js#L362).
`process.env.EMBER_ADDON_ENV = undefined` will coerce the undefined value to a string, which will return the string 'undefined' when called. This affect the EMBER_ADDON_ENV normalization done in [ember-addon.js](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-addon.js#L24), which will not setup the variable to 'development' because the 'undefined' string is present. And in the end will affect the [Addon#isDevelopingAddon](https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L185) making it always return false and which make my tests fail since it's just intended to be used when the addon is being developed.
I changed addon-test.js to handle this.

stefanpenner:
@marcioj nice find!

marcioj:
I changed the way I was closing the test server to just send a `q`, like we do in the terminal.
Previously using `killCliProcess` was causing a problem of don't killing phantomjs processes, so other tests running phantomjs will hang forever.
Unfortunately sending the `q` in this way will throw other error which I sent a fix to testem in airportyh/testem#544.
With this fix I believe all will work. So we need to wait for this patch be merged and a new testem version be released.

johanneswuerbach:
The fix is included in testem `v0.8.2`, thanks for the PR.

marcioj:
Thanks for the release @johanneswuerbach !

stefanpenner:
@johanneswuerbach <3

marcioj:
@stefanpenner :+1: / :-1: ?

PR #3916, opened 2015-04-17
Disable directory listings on development server

joliss:
@rwjblue, you mentioned that you want this...

My only worry would be that anybody trying to have an `/assets` route in their Ember app *might* have a bad surprise when they deploy to production, depending on how their server or CDN is set up, and now that `autoIndex` is false we're not catching that in development anymore. But I'm unsure.

### DIFF
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -29,7 +29,9 @@ HistorySupportAddon.prototype.serverMiddleware = function(config) {
 
         if (hasHTMLHeader && isForBaseURL && req.method === 'GET') {
           var assetPath = req.path.slice(baseURL.length);
-          if (!fs.existsSync(path.join(results.directory, assetPath))) {
+          var isFile = false;
+          try { isFile = fs.statSync(path.join(results.directory, assetPath)).isFile(); } catch (err) { }
+          if (!isFile) {
             req.serveUrl = baseURL + 'index.html';
           }
         }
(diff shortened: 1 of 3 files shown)


lib/tasks/server/middleware/history-support/index.js
```diff
@@ -29,7 +29,9 @@ HistorySupportAddon.prototype.serverMiddleware = function(config) {
 
         if (hasHTMLHeader && isForBaseURL && req.method === 'GET') {
           var assetPath = req.path.slice(baseURL.length);
-          if (!fs.existsSync(path.join(results.directory, assetPath))) {
+          var isFile = false;
+          try { isFile = fs.statSync(path.join(results.directory, assetPath)).isFile(); } catch (err) { }
+          if (!isFile) {
```
joliss: Broccoli's middleware just calls `next()` instead of serving 404. So if possible, it might be nicer to rearrange the middleware stack so this check becomes unnecessary. I just didn't want to try doing this right now.

rwjblue:
:+1:

lib/tasks/server/middleware/history-support/index.js
joliss: If you need Broccoli to call `next()` on non-GET (or some other requests like that) to make that possible, I'm happy to merge a PR.

abuiles:
@joliss I just tried this and it works great, thanks!

abuiles:
@joliss thanks!

PR #3922, opened 2015-04-17
Export test output dir via ENV

ef4:
This allows custom testem launchers to discover the location of the build output. Browser-based launchers don't need this because they all point at the server URL, but process-based launchers have no other way of finding the build.

For example, I have a project that outputs some node-formatted tests alongside the usual browser-formatted stuff. With this PR I can add a node-based launcher to testem.json:

```js
  "launchers": {
    "NodeTests": {
      "command": "mocha -R tap $EMBER_CLI_TEST_OUTPUT/node-tests",
      "protocol": "tap"
    }
```

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -80,6 +80,7 @@ module.exports = Command.extend({
 
   run: function(commandOptions) {
     var outputPath  = this.tmp();
+    process.env['EMBER_CLI_TEST_OUTPUT'] = outputPath;
     var testOptions = this.assign({}, commandOptions, {
       outputPath: outputPath,
       project: this.project,


rwjblue:
:+1:

stefanpenner:
lets make this `--ouput-path`

rwjblue:
Output path is already configurable, this exposes the specified output path to yhe custom testem runners.

stefanpenner:
> Output path is already configurable, this exposes the specified output path to yhe custom testem run

i get it /me hides

stefanpenner:
does this form of environment variable work in windows?

rwjblue:
Yes, setting an environment variable like this does work on windows. The custom testem.json launcher config may be slightly different than in the description above, but that should be fine for us here (since we aren't actually adding that by default).

stefanpenner:
the npm/node_modules cache was stale

PR #3925, opened 2015-04-18
Don't ban website crawling by default

bcardarella:
With fastboot generated apps we don't want to prevent website crawling by default

I copied the robots.txt from Rails


bcardarella:
Sorry, I shouldn't make PRs at 4:30am

Issue #3935, opened 2015-04-20
Ember-cli creates a resource route in the resource generator


tonkonogov:
As the title said, cli creates a resource route by default. But in view of forthcoming deprecation of the `resource` usage in the router file it should be replaced by a simple `route`.

**Formal things**

**HTR:** run `ember g resource any` in the root dir
**Expected:** when generator is complete `router.js` contains `this.route('any');`
**Actual:** when generator is complete `router.js` contains `this.resource('any', function() {});`
    
Tested with bare ember-cli v0.2.3

apaganobeleno:
is this one closed ?

stefanpenner:
thanks @apaganobeleno

PR #3936, opened 2015-04-21
Replace 'this.resource' with 'this.route' in generators

HeroicEric:
Working on fixing https://github.com/ember-cli/ember-cli/issues/3935

The tests are passing when used with changes to ember-router-generator in https://github.com/ember-cli/ember-router-generator/pull/11

#### Changes

- `this.resource` is never generated. Always `this.route`
- `this.resource('foo', function() {})` is never generated unless you explicitly generate with `ember g route foo/index` or `ember g resource foo/index`.

#### Questions

- Are there some situations where the index route should be implicitly generated? I find it confusing but I saw this https://github.com/ember-cli/ember-cli/pull/1946

### DIFF
--- a/blueprints/resource/index.js
+++ b/blueprints/resource/index.js
@@ -54,7 +54,7 @@ module.exports = {
       }
     });
 
-    var routeOptions = merge({}, options, { type: 'resource' });
+    var routeOptions = merge({}, options);
 
     var self = this;
     return this._processBlueprint(type, 'model', modelOptions)
--- a/blueprints/route/index.js
+++ b/blueprints/route/index.js
@@ -10,16 +10,6 @@ module.exports = {
 
   availableOptions: [
     {
-      name: 'type',
-      type: String,
-      values: ['route', 'resource'],
-      default: 'route',
-      aliases:[
-        {'route': 'route'},
-        {'resource': 'resource'}
-      ]
-    },
-    {
       name: 'path',
       type: String,
       default: ''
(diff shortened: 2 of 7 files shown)


abuiles:
@HeroicEric this might require a change in  https://github.com/ember-cli/ember-router-generator

HeroicEric:
@abuiles alright I'll check that out when I get through the stuff in here

HeroicEric:
@abuiles I now see what you mean :crystal_ball:

HeroicEric:
The tests should pass once https://github.com/ember-cli/ember-router-generator/pull/11 is merged in.

HeroicEric:
I made some changes to ember-router-generator in https://github.com/ember-cli/ember-router-generator/pull/11, which ended up removing the `type: 'resource'` option when generating routes.

I'm wondering if the same should be done here?

The main difference is that previously, generating a route that was type resource would also add the index route for you.

```javascript
// ember g route foos
this.resource('foos');

// ember g route foos --type=resource
this.resource('foos', function() {});
```

This affects:

- `ember g route foo --type=resouce`
- `ember g resouce foo`

Just looking for some guidance before continuing.

abuiles:
@HeroicEric let's kill all the resource support,  if you want the index route you can do `ember g foo/index` and that will create the empty function, we just need to document it :)

abuiles:
@HeroicEric there are some failing tests ;(

HeroicEric:
@abuiles yea. I haven't updated this to match the most recent changes in ember-router-generator yet. I can probably  update this tomorrow.

HeroicEric:
@abuiles Tests are passing. I updated the description up at the top with some info and a question.

grabbou:
I would suggest printing a note `Using resources is likely to be deprecated in upcoming Ember releases. Please use with caution or switch to using routes`.

PR #3945, opened 2015-04-22
[ENHANCEMENT] Friendly test names and descriptions

eccegordo:
Added a test-info utility and refactored test blueprints to provide more friendly names and or descriptions.

The test-info utility contains a few methods

humanize -> convert string to humanized form
name -> return a test name with prefix
description - > return a test description with prefix

The prefix includes the test type (Unit or Acceptance) as well as the blueprint type (model, adapter, component, mixin, etc)

The end result is that tests will have better names and descriptions when displayed inside qunit user interface.

Some examples

Unit | Component | x-foo
Unit | Route | foo
Unit | Controller | foo
Acceptance | my cool feature

### DIFF
--- a/lib/utilities/test-info.js
+++ b/lib/utilities/test-info.js
@@ -1,0 +1,83 @@
+'use strict';
+var stringUtils      = require('../utilities/string');
+/**
+  Utility to provide more friendly test names and descriptions
+  allows for descriptive prefixes such as
+  - Unit | Component | x-foo
+  - Unit | Route | foo
+  - Unit | Controller | foo
+  - Acceptance | my cool feature
+  etc...
+*/
+
+var SEPARATOR = ' | ';
+
+module.exports = {
+  /**
+    Converts a string into friendly human form.
+
+    ```javascript
+    humanize("SomeCoolString") // 'some cool string'
+    ```
+
+    @method humanize
+    @param {String} str The string to humanize.
+    @return {String} the humanize string.
+  */
+  humanize: function(str) {
+    var ret; 
+    ret = stringUtils.dasherize(str).replace(/[-]/g,' ');
+    return ret;
+  },
+
+
+  /**
+    Return a friendly test name with type prefix
+    Unit | Components | x-foo
+
+    ```javascript
+    name("x-foo", "Unit", "Component") // Unit | Component | x foo 
+    ```
+
+    @method name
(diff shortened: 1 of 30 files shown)


stefanpenner:
LGTM – @jgwhite  r?

eccegordo:
Also, I am thinking that the `test-info` utility could, in theory, eventually be expanded if we want to support more sophisticated behaviors inside test blueprints.

Not sure what those might be yet but perhaps there are some interesting use cases. Perhaps stuff like dynamically populating the needs array or passing more complicated test setup boilerplate in a reasonably DRY way.

Anyway that is why I called it `test-info` to make it general purpose.

jgwhite:
@stefanpenner confirm :+1: 

iojs failure seems unrelated.

eccegordo:
Interesting the iojs failures on travis.

Local tests run fine using iojs v1.5.1

stefanpenner:
@eccegordo re-triggering – we have a race condition which is hopefully just in our tests, that becomes more apparent in iojs.

chadhietala:
For some reason I can not start Appveyor but I would like to get this merged.

stefanpenner:
these aren't intermittent windows error, but they are unrelated to this PR.

eccegordo:
Woot! Thanks for merging.

lib/utilities/test-info.js
```diff
@@ -0,0 +1,83 @@
...
+
+  description: function(description, testType, blueprintType) {
+    var ret;
+    if (blueprintType){
+      ret = testType + SEPARATOR + blueprintType + SEPARATOR + this.humanize(description);
+    } else {
+      ret = testType + SEPARATOR + this.humanize(description);
+    }
+    return ret;
+  }
```
Turbo87: @eccegordo @stefanpenner I'm wondering why the implementation of `name()` and `description()` is completely identical. was that intentional? also it currently seems impossible to generate `Unit | Components | x-foo` because the dash in `x-foo` will be removed, so it seems that the documentation does not match the implementation in some cases.

eccegordo: I believe at the time name and description were just two values in the various test files in different places but I was not sure of the distinction myself. They appeared to emit the same content. Might just be some minor inconsistency in the blueprints. Maybe stefan knows better than I what is correct. But yes it intentional but might not be correct.

PR #3946, opened 2015-04-22
[ENHANCEMENT] Add host option to `ember test`.

wangjohn:
One of the problems we've run into while running acceptance tests on ember-cli is the ability to hit different backends. We'd like to hit a real database while performing acceptance tests. Unfortunately, `ember-test` doesn't allow you to change the hostname from where the tests are running. This means that if you have any sort of CORs rules, then you won't be able to actually use your production backends for your acceptance tests.

This PR changes that by allowing you to specify `ember test --host="your_host_name"`, which will pass a `host` option to testem. This will launch ember-cli tests on a host that can be specified by the user, since testem already has support for the `host` option (see the [testem option reference guide](https://github.com/airportyh/testem/blob/master/docs/config_file.md#cli-level-options)).

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -16,6 +16,7 @@ module.exports = Command.extend({
     { name: 'environment', type: String, default: 'test', aliases: ['e'] },
     { name: 'config-file', type: String,  default: './testem.json', aliases: ['c', 'cf'] },
     { name: 'server',      type: Boolean, default: false, aliases: ['s'] },
+    { name: 'host',        type: String,  aliases: ['H'] },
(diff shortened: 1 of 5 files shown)


stefanpenner:
Lgtm

wangjohn:
Looks like all of the continuous integration tests on the repo are failing, which I believe is why the tests for this PR are failing. Other than that, are there any other things for me to do?

stefanpenner:
> Other than that, are there any other things for me to do?

nah the windows test failures are unrelated, they will be green soon :)

PR #3951, opened 2015-04-24
[ENHANCEMENT] Add --directory flag to `ember new`

HeroicEric:
This allows a `--directory` flag to be passed to `ember new foo`. This will let you keep the package name of `foo` while choosing the directory name.

I often find myself wanting to generate an ember-cli app as a sub-directory in an app's root directory.

For example, you might have an ember-cli app living beside a Rails app in the same directory:

```
-- foo_app
  |-- rails
  |-- ember
```

I'll typically want the applications themselves to be named the same thing (such as `foo`) but the directories that they live in would look more like:

```
-- foo_app
  |-- backend
  |-- frontend
```

Currently this requires generating the ember-cli app with `ember new foo` and then renaming the directory.

### DIFF
--- a/lib/commands/new.js
+++ b/lib/commands/new.js
@@ -10,7 +10,7 @@ var normalizeBlueprint = require('../utilities/normalize-blueprint-option');
 
 module.exports = Command.extend({
   name: 'new',
-  description: 'Creates a new folder and runs ' + chalk.green('ember init') + ' in it.',
+  description: 'Creates a new directory and runs ' + chalk.green('ember init') + ' in it.',
   works: 'outsideProject',
 
   availableOptions: [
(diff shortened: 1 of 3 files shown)


twokul:
@HeroicEric looks like tests are failing

HeroicEric:
@twokul Travis is passing now. I'm not sure what the deal is with the appveyor stuff. I don't think it's related to any of these changes... Any idea what those are?

HeroicEric:
It looks like other PRs are failing the AppVeyor build with the same failures.

stefanpenner:
> It looks like other PRs are failing the AppVeyor build with the same failures.
>  Hide all checks

I'm trying to slowly fix those :P.

PR #3954, opened 2015-04-24
[ENHANCEMENT] Generate component pods outside components folder

trabus:
Adds `--path` option to the component blueprint to generate pod components outside 'components' folder. 
By using the `--path` option when generating a component blueprint in a path other than `app/components/`( or `app/pods/components` if you have `podModulePrefix` defined) so you can generate components in associated routes, or just in the root folder if you want. So you can generate as described here: https://github.com/dockyard/styleguides/blob/master/ember.md#use-pods-structure or like here: https://github.com/stefanpenner/ember-jobs/tree/master/app/pods

I'm putting this up early for comments, will be adding tests later today.

### Usage
 `ember g component foo-bar -p -path foo` generates into `app/foo/foo-bar/component.js`

the flag only works for component pods, and the value is `components` by default

so if undefined `ember g component foo-bar -p` generates `app/components/foo-bar/component.js` as it currently does

with `usePods:true` you’d only need to pass the `-path` option obviously

### DIFF
--- a/blueprints/component-addon/index.js
+++ b/blueprints/component-addon/index.js
@@ -12,7 +12,7 @@ module.exports = {
     return {
       __path__: function(options) {
         if (options.pod) {
-          return path.join(options.podPath, 'components', options.dasherizedModuleName);
+          return path.join(options.podPath, options.locals.path, options.dasherizedModuleName);
         }
         return 'components';
       },
@@ -54,7 +54,8 @@ module.exports = {
     }
 
     return {
-      modulePath: pathName
+      modulePath: pathName,
+      path: options.path
     };
   }
 };
--- a/blueprints/component-test/files/tests/unit/__path__/__test__.js
+++ b/blueprints/component-test/files/tests/unit/__path__/__test__.js
@@ -3,7 +3,7 @@ import {
   test
 } from 'ember-qunit';
 
-moduleForComponent('<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', {
+moduleForComponent('<%= componentPathName %>', '<%= friendlyTestDescription %>', {
   // Specify the other units that are required for this test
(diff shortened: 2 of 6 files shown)


stefanpenner:
LGTM

trabus:
Just added the `-no-path` alias to generate into the root folder:
`ember g component foo-bar -p -no-path` generates `app/foo-bar/component.js` or `app/pods/foo-bar/component.js` when podModulePrefix is defined.

abuiles:
@trabus it would be really cool if we could support the pod option out of the box, so instead of specifying path, put the component in its pod if exist (or create one)

trabus:
@abuiles I primarily did it this way so it kept current functionality intact, though I'm open to changing it later. If you generate `ember g component foo/bar/foo-bar -no-path` it will generate into the pod, sans 'components' in the path. We can eventually make this default. 

This also made me wonder if we should make the `--path` option link to the `--pod` option, so if it is a value other than `components`, pods are assumed.

abuiles:
@trabus thanks!

ballPointPenguin:
Is this compatable with ember 1.13 angle-bracket component syntax?
eg. if I have app/pages/navigation-bar/component.js and app/pages/navigation-bar/template.js,
in my pages.hbs can I just use <navigation-bar /> or do I need something like:
 <pages/navigaiton-bar /> or {{pages/navigation-bar}} ?

stefanpenner:
> Is this compatable with ember 1.13 angle-bracket component syntax?

i don't believe so, but i could be mistaken. cc @mmun ?

ballPointPenguin:
I don't think this is an ember-cli issue now that I look into it.
I think my question is basically this https://github.com/emberjs/ember.js/issues/10456

PR #3962, opened 2015-04-26
Disable any file watching done by testem

johanneswuerbach:
Every watching should be done already by ember-cli. This should fix random reloads when using `ember test --serve` and save resources.


rwjblue:
:+1:

rwjblue:
Thank you!

stefanpenner:
Good find. Thanks!

PR #3964, opened 2015-04-26
Support templates in addon pods

sglanzer:
Addresses the following issues:
- Fixes the generator for addon components to reference the correct import path when a podModulePrefix is used.
- Allows addons using pods to store the template for a component as a child in the component directory
- Removes the requirement for consumer applications to use the same podModulePrefix as addons to find the templates in the addon

@rwjblue @trabus I would really appreciate your review on this.

### DIFF
--- a/blueprints/component-addon/index.js
+++ b/blueprints/component-addon/index.js
@@ -11,13 +11,13 @@ module.exports = {
   fileMapTokens: function() {
     return {
       __path__: function(options) {
-        if (options.pod) {
+        if (options.pod && options.inRepoAddon) {
           return path.join(options.podPath, 'components', options.dasherizedModuleName);
         }
         return 'components';
       },
       __name__: function(options) {
-        if (options.pod) {
+        if (options.pod && options.inRepoAddon) {
           return 'component';
         }
         return options.dasherizedModuleName;
@@ -50,7 +50,14 @@ module.exports = {
     var pathName       = [addonName, 'components', fileName].join('/');
 
     if (options.pod) {
-      pathName = [addonName, 'components', fileName,'component'].join('/');
+      var podModulePrefix = options.project.config().podModulePrefix || '';
+      var podPath = podModulePrefix.substr(podModulePrefix.lastIndexOf('/') + 1);
+      if (podPath) {
+        pathName = [addonName, podPath, 'components', fileName,'component'].join('/');
+      }
+      else {
+        pathName = [addonName, 'components', fileName,'component'].join('/');
+      }
     }
 
     return {
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -448,38 +448,68 @@ Addon.prototype.compileStyles = function(tree) {
 
   @private
   @method compileTemplates
-  @param {Tree} tree Templates file tree
+  @param {Tree} addonTree Addon file tree
   @return {Tree} Compiled templates tree
 */
-Addon.prototype.compileTemplates = function(tree) {
+Addon.prototype.compileTemplates = function(addonTree) {
   this._requireBuildPackages();
 
-  if (tree) {
+  var isTemplateInAddon = false;
+
+  var templatePatterns = this.registry.extensionsForType('template').map(function (extension) {
+    return new RegExp('template.' + extension + '$');
+  });
+
+  var standardTemplatesTree = this._treeFor('addon-templates');
+  if (standardTemplatesTree) {
+    isTemplateInAddon = true;
+  }
+  else {
+    var addonTreeFiles = walkSync(addonTree);
+    for (var fileIndex = 0; fileIndex < addonTreeFiles.length; fileIndex++) {
+      for (var patternIndex = 0; patternIndex < templatePatterns.length; patternIndex++) {
+        if (templatePatterns[patternIndex].test(addonTreeFiles[fileIndex])) {
+          isTemplateInAddon = true;
+          break;
+        }
+      }
+
+      if (isTemplateInAddon) {
+        break;
+      }
+    }
+  }
+
+  if (isTemplateInAddon) {
     var plugins = this.registry.load('template');
     if (plugins.length === 0) {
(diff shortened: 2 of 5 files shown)


blueprints/component-addon/index.js
```diff
@@ -11,13 +11,13 @@ module.exports = {
   fileMapTokens: function() {
     return {
       __path__: function(options) {
-        if (options.pod) {
+        if (options.pod && options.inRepoAddon) {
```
sglanzer: The generator handles the path from the app component to the addon component, so we don't need to include the pods path in the app.  This has a side benefit of not requiring the consuming application to use a matching podModulePrefix.

blueprints/component-addon/index.js
```diff
@@ -50,7 +50,14 @@ module.exports = {
     var pathName       = [addonName, 'components', fileName].join('/');
 
     if (options.pod) {
-      pathName = [addonName, 'components', fileName,'component'].join('/');
+      var podModulePrefix = options.project.config().podModulePrefix || '';
+      var podPath = podModulePrefix.substr(podModulePrefix.lastIndexOf('/') + 1);
+      if (podPath) {
+        pathName = [addonName, podPath, 'components', fileName,'component'].join('/');
```
sglanzer: Previously, if a podModulePrefix was used the path from the app component to the addon component was missing the prefix and would break.

lib/models/addon.js
```diff
@@ -501,7 +531,7 @@ Addon.prototype.compileAddon = function(tree) {
   }
 
   var addonJs = this.processedAddonJsFiles(tree);
-  var templatesTree = this.compileTemplates(this._treeFor('addon-templates'));
+  var templatesTree = this.compileTemplates(tree);
```
sglanzer: We need to look for templates across the addons tree to find templates in pods.  The previous scope was too narrow (addon/templates)

lib/models/addon.js
```diff
@@ -448,38 +448,68 @@ Addon.prototype.compileStyles = function(tree) {
...
-  @param {Tree} tree Templates file tree
+  @param {Tree} addonTree Addon file tree
   @return {Tree} Compiled templates tree
 */
-Addon.prototype.compileTemplates = function(tree) {
+Addon.prototype.compileTemplates = function(addonTree) {
   this._requireBuildPackages();
 
-  if (tree) {
+  var isTemplateInAddon = false;
```
sglanzer: I'm sure someone can point out a more efficient way to determine if there is a template in the addons directory.

lib/models/addon.js
```diff
@@ -448,38 +448,68 @@ Addon.prototype.compileStyles = function(tree) {
...
+
+  if (isTemplateInAddon) {
     var plugins = this.registry.load('template');
     if (plugins.length === 0) {
-      throw new SilentError('An `addon/templates` tree was detected, but there ' +
-                            'are no template compilers registered for `' + this.name + '`. ' +
-                            'Please make sure your template precompiler (commonly `ember-cli-htmlbars`) ' +
-                            'is listed in `dependencies` (NOT `devDependencies`) in ' +
-                            '`' + this.name + '`\'s `package.json`.');
+      throw new SilentError('Templates were detected in the addon tree, but there ' +
```
sglanzer: Changing the message since we now look for templates anywhere in the addons tree.

lib/models/addon.js
```diff
@@ -448,38 +448,68 @@ Addon.prototype.compileStyles = function(tree) {
...
-
-    var includePatterns = this.registry.extensionsForType('template').map(function(extension) {
-      return new RegExp('template.' + extension + '$');
-    });
-
-    var podTemplates = new Funnel(tree, {
-      include: includePatterns,
+    var podTemplatesTree = new Funnel(addonTree, {
+      include: templatePatterns,
+      exclude: [new RegExp('^templates/')],
```
sglanzer: Same funnel used for pods in the application, but focused on the addon tree.

lib/models/addon.js
```diff
@@ -448,38 +448,68 @@ Addon.prototype.compileStyles = function(tree) {
...
+    var addonTemplatesTree = podTemplatesTree;
+
+    if (standardTemplatesTree) {
+      addonTemplatesTree = mergeTrees([
+        new Funnel(standardTemplatesTree, {
+          srcDir: '/',
+          destDir: 'modules/' + this.name + '/templates',
+          description: 'Funnel: Addon Standard Templates'
+        }),
+        podTemplatesTree
```
sglanzer: Pick up the standard addon/templates directory if it exists and merge with the pods based templates.

sglanzer:
I couldn't find any tests around how the build works for addons (or more specifically how the build process works for pods in addons).  If there is anything to contribute to, please let me know, but I don't really have the time to start those from scratch.

trabus:
@sglanzer oh man, this is awesome! I'm out for most of the afternoon, but will review as soon as I get home tonight. Thanks for taking this on!

blueprints/component-addon/index.js
sglanzer: I made a mistake here - it's breaking the non-addon component pod generation.  I'll start looking for a way to detect if the generate is being called from an addon project.

sglanzer: Scratch that, this is the blueprint for component addons, not sure what's going on with this test currently.

stefanpenner:
cc @chadhietala i believe your packager code may supersed this? If not, it should be taken into account i think.  Tell me your thoughts.

chadhietala:
@stefanpenner I believe so but I should right some cases for pods.

sglanzer:
@chadhietala Thanks for taking a look, I've got a large addon project (and growing) using pods and a good solution to this would go a very long way.

stefanpenner:
i suspect https://github.com/ember-cli/ember-cli/pull/3596 supersedes this

stefanpenner:
going to look at this again when @chadhietala packager work lands.

Issue #3941, opened 2015-04-21
In-Repo Add-On Component Template Swap/Undo files swept up by Broccoli, causes error


noslouch:
I think #3585 caused a regression to #2329. Here are simple steps to reproduce.

First, output of `ember --version`:

```
$ ember --version
Future versions of Ember CLI will not support v0.10.33. Please update to Node 0.12 or io.js.
version: 0.2.3
Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/#watchman for more info.
node: 0.10.33
npm: 2.7.6
```

To reproduce:

```
$ ember new blank
$ cd blank
$ ember generate in-repo-addon calendar
$ ember generate component foo-bar --in-repo-addon calendar
$ vim lib/calendar/addon/templates/components/foo-bar.hbs
<edit edit save>
$ ember s
```

Output:

```
version: 0.2.3
Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/#watchman for more info.
An `addon/templates` tree was detected, but there are no template compilers registered for `calendar`. Please make sure your template precompiler (commonly `ember-cli-htmlbars`) is listed in `dependencies` (NOT `devDependencies`) in `calendar`'s `package.json`.
```

So...
```
$ cd lib/calendar
$ npm install ember-cli-htmlbars --save
$ cd ../..
$ ember s
```

Output:
```
Future versions of Ember CLI will not support v0.10.33. Please update to Node 0.12 or io.js.
version: 0.2.3
Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/#watchman for more info.
Livereload server on port 35729
Serving on http://localhost:4200/
calendar/templates/components/.foo-bar.hbs.un~
Unexpected character '�' (1:3)
SyntaxError: Unexpected character '�' (1:3)
    at Parser.pp.raise (/home/bwhitton/blank/node_modules/ember-cli/node_modules/broccoli-es6modules/node_modules/esperanto/node_modules/acorn/dist/acorn.js:1009:13)
    at Parser.pp.getTokenFromCode (/home/bwhitton/blank/node_modules/ember-cli/node_modules/broccoli-es6modules/node_modules/esperanto/node_modules/acorn/dist/acorn.js:2682:8)
[...]

PR #3970, opened 2015-04-28
Don't include non javascript files in es6 trees

marcioj:
Looks like the es6 trees passed to broccoli-es6modules might contain non javascript files.
This could be a problem when editors like vim generates tmp swapfiles inside the project, since these files aren't parseable for esperanto which throws error.
Also note this issue just happens in addons/in-repo-addons because the ember app trees already filter the js files in ember-app.js [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L888)

This is related to #3941

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -637,7 +637,7 @@ EmberApp.prototype._addonTree = function _addonTree() {
   });
 
   var addonES6 = new Funnel(addonTrees, {
-    srcDir: 'modules',
+    include: [new RegExp('^modules.*\\.js$')],
     allowEmpty: true,
     description: 'Funnel: Addon JS'
   });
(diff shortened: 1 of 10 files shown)


lib/broccoli/ember-app.js
```diff
@@ -637,7 +637,7 @@ EmberApp.prototype._addonTree = function _addonTree() {
   });
 
   var addonES6 = new Funnel(addonTrees, {
-    srcDir: 'modules',
+    include: [new RegExp('^modules.*\\.js$')],
```
marcioj: I tried:

```js
srcDir: 'modules',
include: [new RegExp('.*\\.js$')],
```

but it gives an error in broccoli-funnel. So I used the modules in the regex. Does it behaves in the same way?

stefanpenner:
> broccoli-es6modules

should likely be made to ignore extensions it does not understand, just as our other filters

marcioj:
@stefanpenner :+1:. I will update broccoli-es6modules to handle this.

PR #3975, opened 2015-04-28
included modules is no longer needed

stefanpenner:
* pre-packager will require an explicit dep-graph.json
* we no longer need module information for the module transpiler
* glob is basically broken in windows on shares – pending: https://github.com/isaacs/node-glob/pull/192

- [ ] this is clearly changes untested behavior, I suspect this iteration is ultimately invalid because now all add-ons are compiled. I would like to explore a better solution anyways.

cc @rwjblue / @chadhietala this is robs original work, but overlaps with some of the pre-packager future. so both your input would be great


chadhietala:
LGTM.

PR #3976, opened 2015-04-28
Relative require

stefanpenner:


rwjblue:
Seems good.

I'd merge, but the many red commit status messages above are scary...

stefanpenner:
strange, all is green for `npm run test-all` locally. Guess its time to dig in :P

stefanpenner:
getting trolled by ci or npm or something... Will revisit later/tomorrow

screen shot 2015-04-28 at 7 27 40 pm

stefanpenner:
one of the travis runs was green, the other had an intermittent failure. But this does fix one of the windows issues.

We will circle back when windows is green and fix those intermittent node 12 and iojs issues.

PR #3978, opened 2015-04-29
Update broccoli-es6modules

marcioj:
This new version will transform just `.js` files. Previously editor temp files like vim swapfiles could be proceed, which cause compilation errors.


stefanpenner:
:+1:

PR #3982, opened 2015-04-29
Changed markdown-color blue to bright-blue to be the same on all platforms

trabus:
Chalk changes [blue to bright blue when on windows](https://github.com/sindresorhus/chalk/blob/master/index.js#L15-L17), causing the utilities/markdown-color unit tests to fail on windows.

This PR sets chalk's blue to bright blue for all platforms, fixing the tests for the time being [until it can be fixed upstream](https://github.com/sindresorhus/chalk/issues/63).


stefanpenner:
this should not have been merged, mucking with a dependencies values isn't a reasonable approach.

rwjblue:
Heh, yet that is what chalk is doing that is causing us grief.

PR #3984, opened 2015-04-30
Allow io.js-next in development in 'valid-platform-version'

laiso:
https://github.com/iojs/io.js/blob/master/src/node_version.h

I thought this change for io.js v2.0.0 because added Node v0.13.x already.


stefanpenner:
thanks

PR #3998, opened 2015-05-03
Ember.ObjectController gives deprecation warning

tpitale:
Trying to use ObjectController gives a deprecation warning. Just a minor fix based on the warning suggestion.


rwjblue:
thanks!!

PR #4000, opened 2015-05-03
use `escape-string-regexp` module

sindresorhus:
DRYing it up.

https://github.com/sindresorhus/escape-string-regexp


stefanpenner:
:+1:

Issue #4009, opened 2015-05-04
Exclude certain assets from app.toTree()


ismay:
[The docs](http://www.ember-cli.com/managing-dependencies/) explain how to *include* extra assets with `broccoli-funnel`, but it's not clear how to *exclude* certain assets from being copied to the `dist/assets` folder.

I might have missed something, but couldn't find it in the docs. I think it would be useful to know how to do this in general. Is this currently possible with ember-cli?

---

Some background info for this issue: I'm building my css with `ember-cli-rework` and `rework-suit`, which concatenates both remote and local css (referenced with `@import`). So `app.css` is the only file that I need in my `dist/assets`, since all other files will be concatenated to `app.css` anyway. But by default all other css is copied as well, which isn't necessary.

stefanpenner:
funnel as inlcude/exclude matchers.

ismay:
You mean that `broccoli-funnel` can exclude files from `app.toTree()` as well? Because I read the example:

```javascript
// Providing additional trees to the `toTree` method will result in those
// trees being merged in the final output.
module.exports = app.toTree(extraAssets);
```

As meaning that specifying additional trees will only result in them being *included* as well. So how do I exclude files from `.toTree()`: specify the entire apptree without the files that I want to exclude, or just pass the files to be excluded?

stefanpenner:
nope, what you did is absolutely correct.  If you have time to update the docs, so future travelers can have an easier time that would be handy.

PR #4012, opened 2015-05-05
Add exclusion example for broccoli-funnel

ismay:
For clarity, as the current example only deals with file inclusion. See #4009 for related issue. A quick sketch, so let me know if there's anything that needs improvement.


stefanpenner:
nice, should likely also include a link to the broccoli-funnel docs

ismay:
Thanks! Did you still want that link to the docs btw? Because below my example there's already a link to the repo:

> Note: broccoli-static-compiler is deprecated. Use broccoli-funnel instead.

Which links to the docs as well, but could be changed to something more explicit like: "docs for broccoli".

stefanpenner:
@ismay only if you think it will improve the doc story. If not, im fine without it.

ismay:
Ok cool. Well I think it's fine, there is a link to the docs at the top and at the bottom and the examples cover both use cases now.

PR #4013, opened 2015-05-05
Just a test of Travis' PR tests

drogus:
For some reason PRs fail on io.js, I'm trying to check if this happens also for a commit that passes as a push


stefanpenner:
i've just move iojs to allowed failures for now. I will revisit when i have patience.

PR #4019, opened 2015-05-05
Update references to ember install:bower and ember install:npm

bdvholmes:


abuiles:
@bdvholmes thanks! If you are trying to install an addon, you can actually do `ember install ` that will run the generators for you too, probably we should do that instead of `npm install --save-dev` ?

bdvholmes:
@abuiles good point :+1:

PR #4027, opened 2015-05-06
Re-order postBuild hook

chadhietala:
Currently the postbuild is not very useful as it is called after everything is already in dist. The `results.directory` folder points to a directory that is going to get cleaned up soon after. This allows you to write to the final tree right before it is copied to dist.

__History__

We need to generate some configs off of the app tree, writing that config (a json file) to the destDir of the preprocessor was not merging into the final output.  Instead it caused the multiple rebuilds to occur.

Example:

1. Change router.js
2. Rebuild happens
3. Write config
4. Rebuild occurs again because we wrote the config

We then tried just writing the config to the root of the project and then merging it in during the postBuild hook. The `results.directory` is not the destDir that ends up in the dist, thus the file was lost again.

/cc @dosco

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -116,8 +116,8 @@ module.exports = Task.extend({
        .then(function() {
          return self.builder.build.apply(self.builder, args);
        })
-      .then(this.processBuildResult.bind(this))
       .then(this.processAddonBuildSteps.bind(this, 'postBuild'))
+      .then(this.processBuildResult.bind(this))
       .catch(function(error) {
 
         this.processAddonBuildSteps('buildError', error);
--- a/tests/unit/models/builder-test.js
+++ b/tests/unit/models/builder-test.js
@@ -187,6 +187,22 @@ describe('models/builder.js', function() {
       });
     });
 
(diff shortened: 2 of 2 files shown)


rwjblue:
Seems good.  Can you add to the hook ordering tests?  Something like https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/builder-test.js#L184-L189 ?

stefanpenner:
LGTM, but +1 to tests

kjhangiani:
This patch broke my in-repo addon that I use to create a symlink in /dist after the final build is done.  Is there another hook or a recommended method to accomplish this now?

I changed my code to use results.directory instead (which points to the tmp dir), but this does not work, as the symlink does not get copied over, the contents of it do.  I need the postBuild dist/ folder to contain an actual symlink

rwjblue:
There is a recent PR to add another hook for this I believe.

pangratz:
> This patch broke my in-repo addon that I use to create a symlink in /dist after the final build is done. Is there another hook or a recommended method to accomplish this now?

For the future reader, it's `outputReady` added in https://github.com/ember-cli/ember-cli/pull/4955

PR #4037, opened 2015-05-08
Add ember-disable-proxy-controllers to app blueprint

cibernox:
This addon reexports `Ember.Controller` as `App.ObjectController` and `App.ArrayController`.

Including this addon in new apps will make generated controllers be always regular `Ember.Controller`, remove lots of deprecation warnings and make migration to ember 2 easier.

Idea from @rwjblue in https://github.com/emberjs/ember.js/pull/11070

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -31,6 +31,7 @@
     "ember-cli-qunit": "0.3.12",
     "ember-cli-uglify": "1.0.1",
     "ember-data": "1.0.0-beta.16.1",
+    "ember-disable-proxy-controllers": "^0.7.0",
     "ember-export-application-global": "^1.0.2"
   }
 }


rwjblue:
:+1:

cibernox:
Is it possible to retrigger the build in app veyor? Failed because of NPM install.

rwjblue:
Restarted

cibernox:
Ok, I give up. I retried appveyor 5 times.

abuiles:
@cibernox seems like the appveyor failures are unrelated cc @raytiley ?

rwjblue:
Confirm.

abuiles:
@rwjblue shall we start the 0.13.x series with this?

rwjblue:
I think this can still be 0.2.x, but we should call it out above the changelog (in the "important things to watch" section). This is only breaking if you add the addon when you were relying on `ArrayController`/`ObjectController`.  We just need to make the messaging clear that you should only add this addon to your `package.json` for existing apps if you are comfortable with dealing with the fallout....

cibernox:
`0.13`? Why that change in numbers from 0.2.3?

As a note, with this addon only people creating their own `App.ObjectController`/`App.ArrayController` will be affected. Controllers that explicitly extend from `Ember.Object/ArrayController` won't break.

cibernox:
Most likely only autogenerated controllers will be affected. 
I think that number of projects with their own overriden implementation of proxy controllers is absolutely minimal.

rwjblue:
> with this addon only people creating their own App.ObjectController/App.ArrayController will be affected

To clarify for others, this only affects auto-generated controllers (those that are *not* created as a module manually).  IMHO, there really shouldn't be breakage...

adamesque:
This broke my app — and worse, the breakage caused my tests (which i ran first after updating) to take 40x-100x longer to run, so it took me a little longer to figure out the cause.

This was the error that showed up in my console:
```
Uncaught Error: Assertion Failed: The value that #each loops over must be an Array. You passed '<DS.RecordArray:ember412>' (wrapped in (generated sites controller))
```

The offending line was an `{{#each item in controller}}` with an autogenerated ArrayController. Just wanted to leave a note for future googling.

rwjblue:
@adamesque - Thanks for the note, hopefully helps others that run into this issue.

Kerrick:
This also broke my application. "The value that #each loops over must be an Array. You passed partner@controller:array:, but it should have been an ArrayController"

One fix, if you can't yet change your application code, is to simply uninstall the addon.

PR #4039, opened 2015-05-08
Update ember-qunit to support glimmer

knownasilya:
Relating to this: https://github.com/emberjs/ember.js/issues/11067


rwjblue:
@HeroicEric updated ember-cli-qunit in https://github.com/ember-cli/ember-cli/pull/4032/, but this file still needed to be updated.  Good catch!! Thank you!

Issue #3529, opened 2015-03-14
enable async/await/generators meta issue


stefanpenner:
- [ ] babel/ember-cli
  - [x] fix options issue to enable experimental (ember-cli-babel -> https://github.com/babel/ember-cli-babel/commit/e5f97bb957bde67aa8628d188ef4b2df2208669b released with [v4.3.0](https://github.com/babel/ember-cli-babel/commit/8d0f971ce612ec3ad204fdafaaed88cc4bac0334))
  - [x] get babel to add a whitelist/optional for async / await as a more granular option
  - [ ] bundle polyfills on demand https://github.com/babel/ember-cli-babel/issues/24
    - [ ] patch for RSVP
- [x] replace JSHint with ESLint (https://github.com/ember-cli/ember-cli/issues/6501)
- [ ] co-routines should use Ember.RSVP.Promise
  - [ ] smash global (to prove it works)
  - [ ] implement our own co-routines + get babel to emit (like it does for bb)

why? because, writing test code like this is much better (Eventually we can drop large chunks of ember-testing)

example app: https://github.com/stefanpenner/ember-jobs

![screen shot 2015-03-13 at 5 45 28 pm](https://cloud.githubusercontent.com/assets/1377/6649500/db567d70-c9a8-11e4-9c60-ca369cd8e130.png)

app code isn't to bad either:

![screen shot 2015-03-13 at 5 46 43 pm](https://cloud.githubusercontent.com/assets/1377/6649506/efa10980-c9a8-11e4-9b90-4ee0ed566953.png)

![screen shot 2015-03-13 at 5 47 24 pm](https://cloud.githubusercontent.com/assets/1377/6649511/0609fd94-c9a9-11e4-9662-d4bd52841256.png)

trabus:
If there is any busywork around this that you'd like to pass off, I'm interested!

stefanpenner:
@trabus we really need jshint to understand this syntax, even if it is a fork of jshint until its ready. There are several PR's that implement it.

Now as for the automatic polyfil inclusion, I think we should come up with an RFC to re-think how ember-cli composes apps and addons so we can better accomplish this. I have some ideas here, just having a hard time coalescing my thoughts.  Tomorrow I'm working with @chadhietala on transitioning to npm (dropping bower) and I think it has some overlapping needs.

[...]

PR #4040, opened 2015-05-08
Switch to eslint for style & code checking allowing ES7 syntax

Globegitter:
Over at [sane](https://github.com/artificialio/sane) and a lot of other projects we switched to [eslint](https://github.com/eslint/eslint).

It allows for style checking as well as static code analysis, but in a much more powerful way than jshint. Every rule is a plugin, so it is easy to add new rules to be checked and things such as [complexity](http://eslint.org/docs/rules/complexity) or [code-smells](https://github.com/elijahmanor/eslint-plugin-smells) can be checked in addition to a lot of the basic checks.

This makes jshint obsolete, since eslint is more or less capable of anything that jshint is (though it does not seem to be too easy to get a 1:1 rule conversion).
Furthermore eslint supports all the syntax that babel supports via [babel-eslint](https://github.com/babel/babel-eslint/) fixing part of https://github.com/ember-cli/ember-cli/issues/3529.

Further to removing jshint I separated the linting out and added it as a `pretest` making it possible to `npm run lint` by itself and if linting fails it will not run any tests, making that faster.

I took the rules we have at SANE as well as in parts at [how-to-sane](https://github.com/mgenev/how-to-sane) and slightly modified it to fit the existing ember-cli code as much as was easily possible.

To finish up this PR the rules should be confirmed so all the necessary errors can be confirmed. All rules can be found here: http://eslint.org/docs/rules/ and in they support `0` for off, `1` for warning and `2` for error. Lastly the `.eslintrc`files in subfolders inherit all the rules from their parent folders.

Let me know what you think and what I should do to finish up this PR.


stefanpenner:
I'm not apposed to this, but would like a comparison chart of what we lose.

topaxi:
+1 I've already switched to ember-cli-eslint and most of my files start with `/* jshint ignore:start */` for using async functions.

I didn't miss anything from jshint yet.

Globegitter:
That is the table I came up with after checking https://github.com/jrajav/jshint2eslint, https://github.com/valorkin/eslint-rules-mapper the jshint and eslint docs:

| jshint        | eslint        |
| ------------- |:-------------:| 
| expr     | no-unused-expressions | 
| proto     | no-proto      | 
| strict | strict / no-global-strict      |
| indent     | indent | 
| camelcase     | camelcase      | 
| boss     | no-cond-assign | 
| curly     | curly      | 
| latedef | no-use-before-define     |
| debug     | no-debugger | 
| devel     | browser      | 
| eqeqeq | eqeqeq      |
| evil     | no-eval | 
| forin     | guard-for-in      | 
| immed | wrap-iife      |
| laxbreak     | **not supported** | 
| newcap     | new-cap      | 
| noarg | no-caller      |
| noempty     | no-empty | 
| quotmark     | quotes      | 
| nonew | no-new      |
| nomen     | no-underscore-dangle | 
| onevar     | one-var      | 
| plusplus | no-plusplus      |
| regexp | **no explanation found in jshint docs, but there is no-invalid-regexp**      |
| undef | no-undef      |
| unused | no-unused-vars     |
| sub | dot-notation      |
| trailing | **no explanation found in jshint docs**     |
| white | **no explanation found in jshint docs**     |
| eqnull | no-eq-null / eqeqeq option      |
| esnext | es6      |

For another big project switching you can check: https://www.drupal.org/node/2264205

Also some of the added rules are a bit stricter now such as `Unexpected trailing comma  comma-dangle` or `["paths"] is better written in dot notation  dot-notation`, so the question is if all of these are wanted. Otherwise I am happy to fix the remaining 370 errors.

IanVS:
As a note, it may be best to wait until after breaking changes in ESLint v1.0.0, which is likely not far off. See the [milestone list] (https://github.com/eslint/eslint/issues?q=is%3Aopen+is%3Aissue+milestone%3Av1.0.0) for what is remaining. One large change is that rules will not be enabled by default, so it will be much easier to configure exactly the rules wanted.

Globegitter:
Another interesting new feature might be http://eslint.org/docs/developer-guide/shareable-configs

trabus:
Might want to look at making an ember-watson task to do an AST transform to convert the jshint syntax to eshint syntax.

jonathanKingston:
I have just seen this thread as a link from Slack; as [ember-cli-eslint](https://www.npmjs.com/package/ember-cli-eslint) has been mentioned and [shareable configs](http://eslint.org/docs/developer-guide/shareable-configs) I thought I would chime in.

I changed ember-cli-eslint to use [eslint-config-ember](https://www.npmjs.com/package/eslint-config-ember) shared config.

My plan for this is to have a config which closely matches the communities defaults, it isn't perfect yet as it has had limited testing however I am hoping that the config can include any plugins to match the JSCS custom tests the community uses (Which seem to be covered by ESLint default rules - https://github.com/eslint/eslint/issues/2206 << I think that was the last one missing which I can enable if people are interested)

Along with my proposal to simplify testGeneration: https://github.com/ember-cli/rfcs/pull/15

My personal motivation is getting people to having the simplest process for this:
```
ember install ember-cli-eslint
ember install <testingframework>
```

With an `.eslintrc` of:
```
{
  "extends": "ember"
}
```

I would be very happy to open up the conversations about using the above setup to match what you guys would like to see as Embers defaults etc in the corresponding GitHub repos.

stefanpenner:
@jonathanKingston nice, i prefer the add-on route.

I think of ember-cli will support eslint, it will be via an add-on. Not directly like this.

I will close this PR, I would accept a PR that extracted jshint similar to how @jonathanKingston has done with eslint. That way users can swap in-and out.

jonathanKingston:
I might be misreading this (it's late here) however I actually think the PR makes sense as well as the separate work to extract away JSHint as a plugin in projects.

Ember(-cli) core could benefit from having linting that everyone is happy with (which the core could choose to have it's own linting custom setup for example) if that is the shared config as the core then even better.

My comment was mostly in the direction of: whatever the outcome of this PR is I would like to unify all Ember apps to have a similar default.

Globegitter:
@stefanpenner Do you mean creating a jshint plugin for generated ember-cli apps? This PR was for ember-cli core code to switch to eslint.

stefanpenner:
> This PR was for ember-cli core code to switch to eslint.

Yes, but that is a non-trivial upgrade for our users. Since all there custom + inherited rules would need to be updated. This seems like a pretty high short-term cost.

What this PR should rather do. Extract jslint & eslint parts into addons, so they can easily be swapped. initially ship with jslint, and if eslint proves to be the more common choice we can swap over.

IanVS:
I would have expected that the ember-cli team would want to specify the linting which is done on the core ember-cli code base (i.e. this repo), rather than allow them to be swapped out.  Then someone might use one tool, and Travis CI use another, and there would be mayhem.

IanVS:
Or are you saying to give everyone a chance to get used to ESLint on their own apps via a plugin, and if most ember-cli contributors like it, then change over the ember-cli core?  That would also make sense.

PR #4042, opened 2015-05-09
allow custom configPath

bryanaka:
Currently, the EmberApp allows passing in a `configPath` to specify where the `config/environment.js` file lives. This works for the build, but falls over if the `project` object is used outside of the build, like in the `serverMiddleware` hook.

This PR allows specifying the configPath in `package.json` - directories.configPath so that addons can properly hook into the config file.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -125,7 +125,7 @@ EmberApp.prototype._initProject = function(options) {
   this.project = options.project || Project.closestSync(process.cwd());
 
   if (options.configPath) {
-    this.project.configPath = function() { return options.configPath; };
+    this.project.setConfigPath(options.configPath);
   }
 };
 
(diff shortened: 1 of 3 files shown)


lib/broccoli/ember-app.js
```diff
@@ -125,7 +125,7 @@ EmberApp.prototype._initProject = function(options) {
   this.project = options.project || Project.closestSync(process.cwd());
 
   if (options.configPath) {
-    this.project.configPath = function() { return options.configPath; };
+    this.project.setConfigPath(options.configPath);
```
stefanpenner: seems like we should just use a property or a ES5 setter?

bryanaka:
closing this. I know some things changed and are changing around this area since ember-app needs cleaning up. I want to rethink how I tackle this, so rather not leave this PR lingering.

PR #4049, opened 2015-05-11
[INTERNAL] Allow internal cli parameters to be configurable by other cli tools

rodyhaddad:
This PR allows for more configuration of the ember-cli, when it's used by another cli tool. Example of configuration [here](https://github.com/rodyhaddad/wrenchjs/blob/1ff0ec8e450522d47308f336b5d109e579bef45a/lib/cli/index.js#L8-L10).

In practice, this adds new properties to the `cli` object:

* name: The name of the cli tool. Although this doesn't change anything now, it will later by outputting the proper cli name for error and help messages
* cliRoot: The root of the cli project. This is used to discover more "in-repo" addons
* npmPackage: This is the npm package name used to see if the project is an "ember-cli" project (as opposed to an ember addon, or a null project)

Also with this PR, the `Project` and `Command` instances get a reference to the `cli` object, so they can make use of the information described above.

----

I tried to do the least changes possible to reach my goal of making these things configurable and expanding `ember-cli`'s flexibility. Don't hesitate to tear up and question any change I did, if there's need for more details about my motives/thought process when making this PR.

/cc @stefanpenner

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -12,9 +12,12 @@ var emberCLIVersion      = versionUtils.emberCLIVersion;
 var InstallationChecker  = require('../models/installation-checker');
 
 function CLI(options) {
+  this.name = options.name;
   this.ui = options.ui;
   this.analytics = options.analytics;
   this.testing = options.testing;
+  this.cliRoot = options.cliRoot;
+  this.npmPackage = options.npmPackage;
 
   debug('testing %o', !!this.testing);
 }
@@ -41,7 +44,8 @@ CLI.prototype.run = function(environment) {
       tasks:     environment.tasks,
       project:   environment.project,
       settings:  environment.settings,
-      testing:   this.testing
+      testing:   this.testing,
+      cli: this
     });
 
     getOptionArgs('--verbose', commandArgs).forEach(function(arg){
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -11,6 +11,7 @@ var CLI           = require('./cli');
 var packageConfig = require('../../package.json');
 var debug         = require('debug')('ember-cli:cli/index');
 var merge         = require('lodash/object/merge');
+var path          = require('path');
 
 var version      = packageConfig.version;
 var name         = packageConfig.name;
@@ -46,10 +47,8 @@ function cli(options) {
     writeLevel:   ~process.argv.indexOf('--silent') ? 'ERROR' : undefined
   });
 
-  var project = Project.projectOrNullProject(ui);
-
   var config = new Yam('ember-cli', {
-    primary: project.root
+    primary: Project.getProjectRoot()
   });
 
   var leekOptions;
@@ -81,6 +80,18 @@ function cli(options) {
   debug('leek: %o', leekOptions);
 
   var leek = new Leek(leekOptions);
+
+  var cliInstance = new CLI({
+    ui:        ui,
+    analytics: leek,
+    testing:   options.testing,
+    name: options.cli ? options.cli.name : 'ember',
+    cliRoot: options.cli ? options.cli.cliRoot : path.resolve(__dirname, '..', '..'),
+    npmPackage: options.cli ? options.cli.npmPackage : 'ember-cli'
+  });
+
+  var project = Project.projectOrNullProject(ui, cliInstance);
+
   var environment = {
     tasks:    tasks,
     cliArgs:  options.cliArgs,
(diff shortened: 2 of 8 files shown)


stefanpenner:
@rodyhaddad sorry I was on vacation, looking now

lib/models/project.js
```diff
@@ -553,18 +558,38 @@ Project.closestSync = function(pathName, _ui) {
...
+    var directory = findup.sync(process.cwd(), 'package.json');
+    var pkg = require(path.join(directory, 'package.json'));
+
+    if (pkg && pkg.name === 'ember-cli') {
+      return process.cwd();
+    }
+
+    return directory;
+  } catch(reason) {
+    return process.cwd();
```
stefanpenner: I would prefer we explicitly only swallow the error if it is an error we match against. Basically I want to be sure we are not swallowing unexpected errors.

This is likely also a good plays to add some `debug` logging.

stefanpenner:
LGTM, left one comment I think should be addressed.

stefanpenner:
@rodyhaddad friendly ping

lib/models/project.js
```diff
@@ -82,21 +83,26 @@ Project.prototype.setupNodeModulesPath = function(){
...
 
-var NULL_PROJECT = new Project(process.cwd(), {});
+var processCwd = process.cwd();
+Project.createNullProject = function (ui, cli) {
+  var NULL_PROJECT = new Project(processCwd, {}, ui, cli);
 
-NULL_PROJECT.isEmberCLIProject = function() {
-  return false;
-};
+  NULL_PROJECT.isEmberCLIProject = function() {
```
stefanpenner: looks like strange indentation here

lib/models/project.js
```diff
@@ -82,21 +83,26 @@ Project.prototype.setupNodeModulesPath = function(){
...
+    return false;
+  };
 
-NULL_PROJECT.isEmberCLIAddon = function() {
-  return false;
-};
+  NULL_PROJECT.isEmberCLIAddon = function() {
```
stefanpenner: ^

lib/models/project.js
```diff
@@ -82,21 +83,26 @@ Project.prototype.setupNodeModulesPath = function(){
...
+    return false;
+  };
 
-NULL_PROJECT.name = function() {
-  return path.basename(process.cwd());
-};
+  NULL_PROJECT.name = function() {
```
stefanpenner: ^

lib/models/project.js
rodyhaddad: No, not really.
The indentation is because null projects are now created by the `Project.createNullProject` function.
Maybe checking out the code without the diff makes it easier to see: https://github.com/rodyhaddad/ember-cli/blob/rodyhaddad/cli-configurable/lib/models/project.js#L87-L105

The reason for the method is because Null Projects need a reference to the `cli` instance of the app.

stefanpenner:
doing a thorough review shortly

lib/cli/cli.js
```diff
@@ -12,9 +12,12 @@ var emberCLIVersion      = versionUtils.emberCLIVersion;
 var InstallationChecker  = require('../models/installation-checker');
 
 function CLI(options) {
+  this.name = options.name;
   this.ui = options.ui;
   this.analytics = options.analytics;
   this.testing = options.testing;
+  this.cliRoot = options.cliRoot;
```
stefanpenner: lets rename this to `root` no need to prefix it.

`cli.cliRoot` vs `cli.root` etc.

(im reviewing now, I may just doit myself quickly, leaving notes here for myself)

lib/cli/index.js
```diff
@@ -81,6 +80,18 @@ function cli(options) {
   debug('leek: %o', leekOptions);
 
   var leek = new Leek(leekOptions);
+
+  var cliInstance = new CLI({
```
stefanpenner: lowercase `cli` is likely sufficient to imply `instance`

rodyhaddad: I did that initially, but jshint complains that we're redefining `cli` in the file, because there's already a [function named `cli`](https://github.com/ember-cli/ember-cli/blob/62e02b515809b8c566bc47c7dddeecf20171bc74/lib/cli/index.js#L36)

lib/cli/cli.js
rodyhaddad: I won't have time to do the changes today, so if you want to do them and merge the PR, go ahead.

stefanpenner: no worries, that was my plan :)

lib/cli/index.js
stefanpenner: ah

stefanpenner: i just made the `cli` function anonymous and the `module.exports`

stefanpenner:
superseded by #4168  (just some cleanup)

PR #4053, opened 2015-05-12
[ENHANCEMENT] `ember generate route foo` within addon should create extendable route

gmurphey:
I’m working on building addon “modules” that can expose ready-made routes to a consuming application. I’d like these routes to be extendable, but, within an addon, the `route` blueprint only creates its files within `app`. I’ve changed the `route` blueprint to generate files within `addon` if it’s within an addon and created a `route-addon` blueprint to create the addon-import wrappers within `app`.


stefanpenner:
LGTM

stefanpenner:
rebased https://github.com/ember-cli/ember-cli/pull/4152

PR #4073, opened 2015-05-14
Changelog for 0.2.4

abuiles:


stefanpenner:
We should likely remove change entries that provide no value like "restarting tests".  In a perfect world all commit messages make sense, but when they don't we should likely massage them before they enter the changelog. This way readers have an easier time.

Also when it says @ember-cli it most likely mean @stefanpenner. This is likely due to me migrating the repo from my user to the org. Ultimately not a big deal.


Thanks for the release guys!

stefanpenner:
It's also worth noting that broccoli-filter has abit of a Perf boost.

PR #4076, opened 2015-05-14
Use caret version for stable dependencies.

abuiles:


rwjblue:
:+1:

vikhyat:
The release notes [here](https://github.com/ember-cli/ember-cli/releases/tag/v0.2.5) has the following line:

> Make sure that ember-dependency-checker is using caret ^1.0.0.

Shouldn't that be:

> Make sure that ember**-cli**-dependency-checker is using caret ^1.0.0.

abuiles:
@vikhyat good catch, thanks!

PR #4080, opened 2015-05-14
Bump ember-cli-dependency-checker to v1.0.1

quaertym:


abuiles:
We don't need to to this, since we have the caret, it will pick the latest version automatically.

PR #4115, opened 2015-05-19
Do not change the git committer during `ember new`

davidcelis:
While I do think that this is a cute feature, I end up amending that
initial commit every time I create a new Ember app using ember-cli so
that it correctly appears under my own name/email address and has the
standard "Initial commit" message.

Signed-off-by: David Celis <me@davidcel.is>


stefanpenner:
I think we will likely keep this around. Feel free to amend if you are not interested in it. In-theory and add-on could be created and combined with a blueprint to disabled this.

davidcelis:
I can see the attachment to the message itself (and I can leave that alone in this PR) but I really don't think that messing with the author of the commit is cool.

stefanpenner:
> but I really don't think that messing with the author of the commit is cool.

One could argue that the commit is actually authentic, as it is the cumulative of the ember community. Subsequent commits are in-fact not, rather the local `git user`.

Issue #4101, opened 2015-05-18
Watchman problem after recent update (0.2.4)


jtsom:
After updating to ember-cli 0.2.4, when I start an app I'm seeing watchman print out a bunch of warnings / errors before the app boots up:

    version: 0.2.5
    Livereload server on port 35729
    Serving on http://localhost:4200/
    Recrawled this watch 1 times, most recently because:
    /Users/john/Devel/ember/borrowers: kFSEventStreamEventFlagUserDropped
    To resolve, please review the information on
    https://facebook.github.io/watchman/docs/troubleshooting.html#recrawl

These lines repeat many times before the app starts. The app seems to work ok, as far as I can tell. This didn't happen before 0.2.4, and is still happening with 0.2.5.

I can get it to stop, at least temporarily, by stopping the watchman process with:
    watchman shutdown-server

Thanks

stefanpenner:
This should also be opened as a watchman issue

Cc @wez thoughts?

cigoe:
I am getting this as well this morning after a 0.2.1 to 0.2.3 update

```
version: 0.2.3

A new version of ember-cli is available (0.2.5). To install it, type ember update.
Livereload server on port 40529
Serving on http://localhost:4200/app/
Recrawled this watch 1 times, most recently because:
/Users/chris/devel/web/test: kFSEventStreamEventFlagUserDropped
To resolve, please review the information on
https://facebook.github.io/watchman/docs/troubleshooting.html#recrawl
```

jtsom:
I had posted it here, because I hadn't seen it until I did an ember-cli update. Watchman hasn't been updated, as far as I know.

quiddle:
I ran into the recrawling issue as well. I ran `watchman shutdown-server` and then ember-cli went back to being awesome.

wez:
That message is telling you that your system is having trouble keeping up with the filesystem notifications.

Can you provide more details on the size of the tree, number of watches and how frequently you are updating them?  Can you also give us an idea of how powerful the system is (CPU, RAM, whether the disk is a spinning laptop disk, SSD or a high rpm spinning disk)

The output from:

```
watchman version
watchman watch-list
```

and making a gist of your watchman log file would also be useful.  (If you installed via homebrew it will be something like `/usr/local/Cellar/watchman/3.1/var/run/watchman/$USER.log`)

[...]

PR #4124, opened 2015-05-20
[ENHANCEMENT] Add watchmanconfig file to blueprints

mikegrassotti:
As discussed in #4101 a .watchmanconfig file is added to the base blueprint. Per @wez it will ignore the tmp dir but not node_modules.

cc @stefanpenner


stefanpenner:
woohoo!

PR #4125, opened 2015-05-20
Add comment about livereloading when developing addons

max:
It seems to be quite common that people are lost on how to get the fantastic live reloading mojo going when `npm link`ing their Ember addon into a project. This surfaces the `isDevelopingAddon` hook.


stefanpenner:
:+1:

PR #4126, opened 2015-05-20
Admin privileges on windows

matheusdavidson:
Opening the terminal with admin privileges is very important on windows!


stefanpenner:
thanks

PR #4134, opened 2015-05-21
Update 2013-04-08-managing-dependencies.md

davidstevens37:
Added documentation for @trabus 's contribution. Explains the behavior of adding `{prepend: true}` to `app.import()`


trabus:
@davidstevens37 thanks! I need to be better about updating docs. :flushed:

PR #4143, opened 2015-05-22
[BUGFIX] Blueprint.load verify blueprint is in a directory

### LINKED ISSUE #4136: Blueprints listing contains random file entries
When listing available blueprints via `ember g --help` I'm seeing random files listed like `.jshintrc` or `.DS_Store`. I've been told the listing should be filtered to folders that actually contain an `index.js`.

pasted_image_at_2015_05_21_11_46_am

First mentioned in Slack [here](https://embercommunity.slack.com/archives/ember-cli/p1432233867000399).

trabus:
Fixes #4136

Verifies that the blueprint path to be loaded is a directory, preventing non-blueprint files, like `.jshintrc` and `.DS_Store` from turning up in the generated blueprint list.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -14,6 +14,7 @@ var path          = require('path');
 var sequence      = require('../utilities/sequence');
 var stat          = Promise.denodeify(fs.stat);
 var stringUtils   = require('../utilities/string');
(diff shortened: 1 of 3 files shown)


jgwhite:
I’m surprised at how long it took this bug to surface/be reported. Fix looks spot on.

PR #4146, opened 2015-05-22
Code Quality: addon.js, project.js D -> C

jkarsrud:
This improves some duplication in `lib/models/addon.js` and `lib/models/project.js`, as listed in #3730.

It adds a new method to `lib/models/addon-discovery.js` to map the discovered packages.

@stefanpenner This can be improved even further, but we should probably discuss if it's worth it in terms of readability.

### DIFF
--- a/lib/models/addon-discovery.js
+++ b/lib/models/addon-discovery.js
@@ -226,5 +226,15 @@ AddonDiscovery.prototype.dependencies = function(pkg, excludeDevDeps) {
   return assign({}, devDependencies, pkg['dependencies']);
 };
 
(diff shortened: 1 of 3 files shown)


stefanpenner:
Awesome, what other improvements did you have in mind?

jkarsrud:
I was just thinking if there was a way to remove all duplication, but that might not even be possible, so this might just be it.

stefanpenner:
> I was just thinking if there was a way to remove all duplication, but that might not even be possible, so this might just be it.

ah, ok. :) Thanks for the contribution!

PR #4147, opened 2015-05-22
Detect ember-cli from deps as well as devDeps

searls:
This is useful to folks who, for whatever reason, need to have access to 
ember-cli with production dependencies (certain shrinkwrap scenarios,
installation into some build server configs, and probably other cases).

Looks like #3015 was closed prematurely when @rwjblue gave a bit of good
implementation advice. Also addresses #2207.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -118,7 +118,7 @@ Project.prototype.name = function() {
   @return {Boolean} Whether this is an Ember CLI project
  */
 Project.prototype.isEmberCLIProject = function() {
-  return this.pkg.devDependencies && 'ember-cli' in this.pkg.devDependencies;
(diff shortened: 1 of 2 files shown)


stefanpenner:
@searls can you add a test?

searls:
@stefanpenner you bet you. Added characterization tests around the method too and fixed an unrelated issue (namely that the method would raise in the event that someone's package.json had no `devDependencies`)

stefanpenner:
> (namely that the method would raise in the event that someone's package.json had no devDependencies)

:+1:

stefanpenner:
one thing to be weary of, is that we don't want addons including ember-cli as a dependency.

PR #4148, opened 2015-05-22
extract common SilentError debug/throw logic

stefanpenner:


trabus:
Thanks for taking care of this.

stefanpenner:
> Thanks for taking care of this.

NP

PR #4150, opened 2015-05-22
Code Quality: npm-install.js, npm-uninstall.js D -> A

jkarsrud:
This removes a pretty big duplication penalty, but I'm not really sure this kind of abstraction is worth it :see_no_evil: It refactors the `npm-install` and `npm-uninstall` tasks to use an `npm-task` with a command and some nice messages. 

Yay or nay?


stefanpenner:
@jkarsrud yay!

PR #4154, opened 2015-05-23
remove duplication from normalize entity name

tyleriguchi:
Let me know if this looks alright, first time contributing. In reference to #3730

### DIFF
--- a/lib/utilities/valid-component-name.js
+++ b/lib/utilities/valid-component-name.js
@@ -1,0 +1,12 @@
+'use strict';
+var SilentError = require('../errors/silent');
+
+module.exports = function(entityName) {
+  if(! /\-/.test(entityName)) {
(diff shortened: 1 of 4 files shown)


lib/utilities/valid-component-name.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+var SilentError = require('../errors/silent');
+
+module.exports = function(entityName) {
```
stefanpenner: can you add an explicit test for this new extracted module?

basically for the two branches:

* test that it throws, with the correct error type and message
* test that it returns entityName

trabus:
>  first time contributing

@tyleriguchi nice work! Let me know if you need any help with the tests.

tyleriguchi:
Added tests to confirm it throws an error and message, and that it returns an entity name

stefanpenner:
Thank you kindly

PR #4158, opened 2015-05-24
Allow addons to have pod based templates

### LINKED ISSUE #3586: Pod templates inside addons don't get compiled

pzuraq:
Updated version of #3596, fixes #3586.


gmurphey:
Thanks for getting the ball rolling on this again, @pzuraq!

pzuraq:
No prob! I realize the [upcoming packager refactor](https://github.com/ember-cli/rfcs/pull/14) may make this approach obsolete, but we're pushing pod structure in our addons so the sooner it gets in the better.

stefanpenner:
@pzuraq yes, i believe an intermediate step is very reasonable.

@chadhietala can you review this, I want to be sure we don't deviate from the packager works.

chadhietala:
@stefanpenner We may not need this in the future, but this doesn't deviate.

PR #4162, opened 2015-05-25
Remove unused variables

quaertym:


quaertym:
It is weird that this decreases coverage.

rwjblue:
Most likely, these lines were considered covered, and now there are less covered lines total (because two were removed)...

rwjblue:
Thank you!

Issue #4161, opened 2015-05-25
docs: more user-friendly sourcemap config section


nruth:
I just saw http://www.ember-cli.com/#source-maps while setting up a new ember-cli 0.2.5 project, and think it needs to give more information. It currently says:

```
Ember CLI supports producing source maps from your CSS and JS files. Source maps are configured by the sourcemaps option and are disabled in production by default.

In dev, the default setting is equal to:

sourcemaps: {
  enabled: true,
  extensions: ['js']
}
```

I'd like to suggest:

- an example with coffeescript
- an example with sass
- or saying which languages are / aren't supported, or how to find out
- explaining whether "extensions" is for the source file (css, js, scss, coffee, etc) or the output file (css, js) 

For reference, I think it was added with https://github.com/ember-cli/ember-cli/pull/2727 but I didn't see mention of these things there either.

stefanpenner:
@nruth sure, would gladly accept a PR that  improves these docs.

nruth:
maybe @ef4 can comment on how/if it works with addons like coffeescript and sass? I'm too lost with it to write something helpful at the moment.

stefanpenner:
the code that @ef4 wrote just works on JS

ef4:
Yes, although I hope that other modules will adopt the same standardized
config, so that you can just list all the file types you want to turn on
source mapping for in one place.

nruth:
Closing in favour of the pr. Please check I've not introduced errors.

PR #4167, opened 2015-05-26
trying to improve source maps docs

nruth:
As mentioned in #4161 I found the new source maps docs a bit confusing on the website, and expected coffeescript and sass to work out of the box. I've tried not to change a lot, but hopefully it makes it a bit clearer that it isn't that simple :)

- Mention that other formats depend on their addons
- Shows how to turn on css
- Remove "configure" from the css bit, as that suggested there are options somewhere to pass to the subsystem.

I'm not familiar with the source maps code so suggest someone who knows it quickly checks what I've changed (in particular re: css).

### DIFF
--- a/_posts/2013-04-09-asset-compilation.md
+++ b/_posts/2013-04-09-asset-compilation.md
@@ -39,8 +39,7 @@ minifyJS: {
 
 ### Source Maps
 
-Ember CLI supports producing source maps from your CSS and JS files.
-Source maps are configured by the `sourcemaps` option and
+Ember CLI supports producing source maps from your CSS and JS source files. For other source formats (sass, coffee, etc) refer to their addons. Source maps are configured by the `sourcemaps` option and
 are disabled in production by default.
 
 In dev, the default setting is equal to:
@@ -52,8 +51,8 @@ sourcemaps: {
 }
 {% endhighlight %}
 
-Pass `{enabled: true}` to your EmberApp constructor to enable source maps for javascript.
-Use the `extensions` option to configure CSS.
+Pass `sourcemaps: {enabled: true}` to your EmberApp constructor to enable source maps for javascript.
+Use the `extensions` option to add CSS source maps: `{extensions: ['js', 'css']}`.
 
 ### Stylesheets
 
--- a/_posts/2013-04-09-asset-compilation.md
+++ b/_posts/2013-04-09-asset-compilation.md
@@ -39,20 +39,23 @@ minifyJS: {
 
 ### Source Maps
 
-Ember CLI supports producing source maps from your CSS and JS source files. For other source formats (sass, coffee, etc) refer to their addons. Source maps are configured by the `sourcemaps` option and
-are disabled in production by default.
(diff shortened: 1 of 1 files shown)


stefanpenner:
@nruth this doc addition looks great. I do hope we can continue to provide better and better support for other source map scenarios, but this should buy us some time.

ef4:
Last time I looked we didn't actually have sourcemap support for CSS. Has that changed?

Otherwise, this seems good. It could be nice to add a complete example of how the default setting would look in the user's brocfile, so they can customize from there:

```js
import EmberApp from 'ember-cli/lib/broccoli/ember-app';
var app = new EmberApp({
  sourcemaps: {
    enabled: EmberApp.env() !== 'production'
  }
});
```

nruth:
I think an example like that would be nice.

Don't know about css. The previous docs seemed to suggested it was supported, so I made it more explicit. If it isn't supported then would be good to make it explicit that only js uglifier is currently supported, but addons can extend this and hook into this config.

I guess the trick is writing something that doesn't confuse framework users  (me) re: what they get out of the box, but gives framework devs a hint that they can plug in there?

stefanpenner:
> Last time I looked we didn't actually have sourcemap support for CSS. Has that changed?

i dont believe so

cibernox:
I've just found this that is barely related to my issue. 
I am having some problems with the sourcemap in some 3rd party libraries (particularly fails for velocity.js imported form liquidfire). Is it possible to only generate sourcemap of application code?

stefanpenner:
> Is it possible to only generate sourcemap of application code?

it should currently make 2 maps, one for vendor and one for app

cibernox:
I've found the broken sourcemap using DBSC (Debugging by Binary Search Commenting). The package with the broken sourcemap is `video.js`.

stefanpenner:
> I've found the broken sourcemap using DBSC (Debugging by Binary Search Commenting). The package with the broken sourcemap is video.js.

but that should have been in vendor.js correct ?

cibernox:
Yes. I debugged by commenting (removing) lines in the package.json until I found the guilty addon.

nruth:
I've changed it to say css isn't supported & re-structured to have the default brocfile example.

nruth:
Is there something else I can do with this?
I imagine it needs Brocfile.js changing to ember-cli-build.js.

I just came back to the docs after a couple of months and got confused all over again re: which asset pipeline stages will get source mapped – right now it still reads to an optimist as if it'll map from chrome inspector to my coffeescript / scss files.

Also this may be a tangent, but is broccoli-uglify-js the right minifier to tell people to install in the paragraph above? It looks like [broccoli-uglify-sourcemap](https://github.com/ef4/broccoli-uglify-sourcemap) would be a better fit.

stefanpenner:
cc @leo

leo:
@nruth Thank you for your help on improving the docs! But could you please re-target this PR to the [new repo][1] of our site? The old branch has been deleted. Thank you very much! :innocent:

[1]: https://github.com/ember-cli/ember-cli.github.io

nruth:
Sure, no problem. I've made a new pr there.

Issue #4094, opened 2015-05-15
Using Pods on 0.2.5


samselikoff:
I'm not sure how to use pods on 0.2.5. I started a fresh app, moved `app/templates/application.hbs` to `app/pods/application/template.hbs`, and updated `/config/environment.js` to try some different things:

```js
var ENV = {
  podModulePrefix: 'my-app-name/pods'
  // podModulePrefix: 'pods'
  // podModulePrefix: 'app/pods'
}
```

but nothing seems to be working. Ideas?

abuiles:
@samselikoff check https://github.com/ember-cli/ember-cli/pull/3954

samselikoff:
Thanks - I read through, but still must be missing something. Can we no longer have route-related classes under `/pods`?

abuiles:
@trabus ^^

trabus:
I'll try and investigate today. This is extremely odd to me, though it may be related to the podModulePrefix being removed from being removed from the app blueprint. https://github.com/ember-cli/ember-cli/pull/3913

trabus:
Actually, that is exactly the problem. It was being linked up to the config value automatically. You now have to manually add it to your `app.js`.

trabus:
@knownasilya it may have been a bit premature to remove podModulePrefix since we had to delay removing it from the response to the deprecation warning we'd put in.

We may want to update the docs to mention that you need to add it manually now.

samselikoff:
woot, that was it!

app.js
```diff
 App = Ember.Application.extend({
   modulePrefix: config.modulePrefix,
+  podModulePrefix: config.podModulePrefix,
   Resolver: Resolver
});
```

zjrosen1:
@samselikoff @trabus confirmed this is working in 0.2.5

@trabus Any progress on the RFC so we don't get to deep into being podheads

trabus:
@zjrosen1 I deferred to @chadhietala's [rfc](https://github.com/emberjs/rfcs/pull/44).

zjrosen1:
@trabus Thanks.

trabus:
@zjrosen1 I'm going to be focusing on work that will allow conversion between different structures, regardless of what finally lands in the end. 

There are a few components that need to be worked on before this is possible. I'm going to be working on [Structures](https://github.com/ember-cli/rfcs/pull/5) soon, and also an `ember mv` command to facilitate moving/renaming.

zjrosen1:
[...]

PR #4189, opened 2015-05-30
Updated podModulePrefix documentation

Matt-Jensen:
Update to the Github pages to reflect the new location, in app.js, to set the pod structure naming convention.

### DIFF
--- a/_posts/2014-04-01-naming-conventions.md
+++ b/_posts/2014-04-01-naming-conventions.md
@@ -265,23 +265,22 @@ In this case, you should name the file as its functionality. Given a resource `U
 - `app/users/route.js`
 - `app/users/template.hbs`
 
-Rather than hold your resource folders on the root of your app you can define a POD path using the attribute `podModulePrefix` within your environment configs. The POD path should use the following format: `{appname}/{poddir}`.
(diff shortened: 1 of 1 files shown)


stefanpenner:
i don't believe this is correct

Matt-Jensen:
Fair enough, was just attempting to reflect the suggestions in #4094. However, I'm happy to close if this is counterproductive.

stefanpenner:
It appears reasonable. I am just unsure what the final fork will be. @trabus do you have thoughts to share?

trabus:
We reverted the change for 0.2.6, as the removal was a bit too early. I'd prefer to keep things as they currently are on master, for which the existing docs are correct.

Eventually, we will remove it, but not until the final state of pods is determined, and naming conflicts with reserved words can be mitigated. Until then, podModulePrefix is a necessary evil for those that need it.

stefanpenner:
> Eventually, we will remove it, but not until the final state of pods is determined, and naming conflicts with reserved words can be mitigated. Until then, podModulePrefix is a necessary evil for those that need it.

:+1: aligned with my thoughts.

@Matt-Jensen i appreciate the contribution, but It may be slightly pre-mature.

PR #4196, opened 2015-05-31
[BUGFIX] Adding fileMapToken __name__ for route-addon blueprint.

gmurphey:
When I submitted #4053, I failed to add tests for pod-based addons. This PR adds that test and generates route addon-imports correctly when using pods.


gmurphey:
Seems like coveralls may've stalled.

stefanpenner:
@trabus can you reivew?

trabus:
LGTM :+1:

PR #4197, opened 2015-05-31
[BUGFIX] add default for path option in component blueprint locals

### LINKED ISSUE #4183: Component blueprint --path option breaks addons with component blueprints that use the built-in blueprint
As reported [here](https://github.com/Vestorly/ember-cli-emblem/issues/20).

Basically, the ember-cli-emblem addon has a component blueprint that does a lookup on the built-in component and calls locals with the command options, like so:
```
locals: function(options) {
    return this.lookupBlueprint('component').locals(options);
}
```

Because the `--path` option default is not properly being set, it comes through as undefined, breaking the `ember g component` command for ember-cli-emblem, and possibly more addons that make use of the same practice.

I've got a PR started for the fix (which is extremely simple), and starting on the test, but I don't see any existing acceptance tests where we're setting up the tests in a way that I need. Basically, I've got to generate an ember app, install a fixture addon that overrides the component blueprint, and then generates a component.

trabus:
Fixes #4183

Added `'components'` as the default value for the `--path` option in the component blueprint. Also added test to verify that blueprints that override built-in blueprints can still call methods of the built-in blueprint they are overriding. For instance:
```
// my-addon/blueprints/component/index.js

locals: function(options) {
    return this.lookupBlueprint('component').locals(options);
}
```

### DIFF
--- a/blueprints/component-addon/index.js
+++ b/blueprints/component-addon/index.js
@@ -3,6 +3,7 @@
 var Blueprint          = require('../../lib/models/blueprint');
 var stringUtil         = require('../../lib/utilities/string');
 var validComponentName = require('../../lib/utilities/valid-component-name');
+var getPathOption      = require('../../lib/utilities/get-component-path-option');
 var path               = require('path');
 
 module.exports = {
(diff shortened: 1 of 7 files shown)


stefanpenner:
needs a rebase

trabus:
I got it. :)

PR #4203, opened 2015-06-01
[BUGFIX] acceptance-test blueprint no longer generates addon re-export in app folder

### LINKED ISSUE #4192: addon-import is generating files when it shouldn't
For example:

A new addon. Run `ember g acceptance-test foo`

The `addon-import` generator will generate a `app/acceptance-tests/foo.js` file.

This is also happening for other test related file:

`ember g model-test foo` => `app/model-tests/foo.js`

trabus:
fixes #4192

Added tests and helper to verify additional files aren't being generated in the `app` folder.

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -14,7 +14,7 @@ module.exports = Task.extend({
   run: function(options) {
     var self = this;
     var name = options.args[0];
-    var noAddonBlueprint = ['mixin'];
+    var noAddonBlueprint = ['mixin', 'acceptance-test'];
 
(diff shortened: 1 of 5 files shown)


stefanpenner:
thanks dude!

trabus:
No problem. After investigating a bit further, I think this is a wider spreading issue than just acceptance tests. I'll be opening an issue once I fully understand what is going on and determine which `addon-import` files are being erroneously generated.

bcardarella:
I think the issue is that the addon-import is an auto opt-in for all other blueprints. I think instead it should be done where addon-import must be specified on the blueprints that it should apply to.

PR #4218, opened 2015-06-03
Add instance-initializer's blueprints family

cibernox:


jgwhite:
Heck yes

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,23 @@
...
+import { initialize } from '<%= dependencyDepth %>/instance-initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+
+var appInstance, application;
+
+module('<%= friendlyTestName %>', {
+  beforeEach: function() {
+    Ember.run(function() {
+      application = Ember.Application.create();
+      appInstance = application.__deprecatedInstance__;
```
cibernox: This doesn't feels right, but it's the only way i've found of getting an application instance

stefanpenner: yes, this isn't something we want to generate.

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,23 @@
...
+      application.deferReadiness();
```
stefanpenner: this also smells poorly

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
cibernox: Any better idea to generate an application instance?

cibernox:
@stefanpenner Can you raise this in the f2f meeting to see if anyone knows a better way of creating an application instance?

stefanpenner:
> @stefanpenner Can you raise this in the f2f meeting to see if anyone knows a better way of creating an application instance?

sure

cibernox:
Ping.

stefanpenner:
Sorry i forgot about this during the meeting. But my gut reaction suggests, maybe this should be handle by the test helpers. Basically a modulForInstanceInitializer that boots in a more realistic way? Thoughts?

trabus:
So is it just the test blueprints that are blocking for this? I like the idea of handling this through a test helper.

stefanpenner:
> So is it just the test blueprints that are blocking for this?

yes



> I like the idea of handling this through a test helper.

ya good short term solution

cibernox:
So the tests part should be addresses with a PR to https://github.com/switchfly/ember-test-helpers.

That helper will still access `application.__deprecatedInstance__` but will be hidden from sight.
I'll try to tackle this over the weekend. I have too many open fronts :(

rwjblue:
>  I have too many open fronts

I know the feeling. :facepunch: :factory:

stefanpenner:
It's all about the macro micro balance

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
acorncom: @cibernox Re: who has this knowledge in his head, do the following links help at all?

https://github.com/emberjs/website/pull/2083
https://github.com/emberjs/website/pull/1951

Looks like @dgeb and @tomdale did most of the work on this ...

tomdale: @cibernox Why do you want to do this?

cibernox: Nothing in particular, I was copying initializer tests and just changed this line 
`appInstance = application.__deprecatedInstance__;`

I already started a PR in ember-test-helpers to create a `moduleForInstanceInitializer` that hides this details, but it's not a trivial thing (all that black magic with the register is pretty obscure). I have to find some time to finish it.

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
tomdale: You should be able to call `app.visit()` which returns a promise that resolves to the app instance.

cibernox:
Taked @tomdale advice and used `app.visit('/)`. 
Works fine but it's feature flagged. Any idea if it will be enabled in the next cycle?

rwjblue:
Unlikely to be enabled in 2.1.

Robdel12:
Is there anything I can do to get this moving again?

cibernox:
Hey, I'm just back from my holidays :disappointed: 

@Robdel12 The generator for the instance initializer would work fine now, but the generated test it's using a feature that is behind a feature flag (`ember-application-visit`) for now. 
The alternative was fiddling with some very internal property in the generated test. This is ugly but could be hidden behind a helper method in ember_test_helpers to hide, but I tried to understand the code and failed miserably.

In case that feature flag is flipped in 2.1 this could be merged (after adding some code to detect that feature) but seems unlikely.

dgeb:
I'm trying to coordinate adding this feature to the guides for 2.1, and hoping that we can land this blueprint along with the explanation that I've written in https://github.com/emberjs/guides/pull/848

I just have one correction to suggest. The commented example given in the blueprint is currently:

```
export function initialize(/* appInstance */) {
  // appInstance.registry.injection('route', 'foo', 'service:foo');
}
```

Now the appInstance supports `RegistryProxy`, this should be rewritten as:

```
export function initialize(/* appInstance */) {
  // appInstance.inject('route', 'foo', 'service:foo');
}
```

Thanks for your work on this @cibernox!

dgeb:
Just read through the full thread here - too bad about the test as a blocker. I don't know the status of that feature flag or what the right move is for the blueprint + test + guides for 2.1, but let me know if I can help move this along.

cibernox:
@dgeb Yeah, this has been hanging around for a long time. Any idea who we need to bribe? :grin:

acorncom:
When this gets merged, can we request a ping on the guides pull request above so we update the docs appropriately?  Thanks!

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,24 @@
...
+import { initialize } from '<%= dependencyDepth %>/instance-initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+
+var appInstance;
+
+module('<%= friendlyTestName %>', {
+  beforeEach: function() {
+    Ember.run(function() {
+      var application = Ember.Application.create();
+      application.visit('/').then(function(instance) {
```
rwjblue: Lets swap this out for `appliction.buildInstance()` (which is available from Ember 1.12.0 forward).  It is labeled as `@private`, but there is no other way to generate an `ApplicationInstance` instance (since `ApplicationInstance` isn't exported on the global).

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
```diff
@@ -0,0 +1,24 @@
+import Ember from 'ember';
+import { initialize } from '<%= dependencyDepth %>/instance-initializers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+
+var appInstance;
+
+module('<%= friendlyTestName %>', {
+  beforeEach: function() {
+    Ember.run(function() {
+      var application = Ember.Application.create();
```
rwjblue: Can you store this as `this.application = Ember.Application.create();`?

blueprints/instance-initializer-test/files/tests/unit/instance-initializers/__name__-test.js
rwjblue: When you update, please store the instance at `this.appInstance`.

rwjblue: Will also need to add a afterEach that cleans things up:

```js
import destroyApp from '../../helpers/destroy-app';

// snip....

afterEach() {
  Ember.run(this.appInstance, 'destroy');
  destroyApp(this.application);
}
```

rwjblue:
@cibernox - With the changes I listed above, I think this would be mergeable.

acorncom:
I've got a pull request pending with the above changes here: https://github.com/cibernox/ember-cli/pull/1

michaelrkn:
I'm not sure why the coveralls check is failing - https://coveralls.io/builds/4011979 shows that none of the files had coverage changes. I'm also not sure what the deal is with the AppVeyor build. Is anybody able to look into that?

rwjblue:
@homu r+

dgeb:
Woot! So glad to have instance initializer blueprints. The guides and generators are now back in sync :shipit:

PR #4227, opened 2015-06-03
Extract process relative require

stefanpenner:


stefanpenner:
test failures are just tiny coverage changes

PR #4234, opened 2015-06-04
Run jekyll via bundler and tell jekyll not to build vendor/

eliotsykes:
- `ruby run.rb` uses bundled jekyll
- jekyll ignores `vendor/` dir when it builds, stops jekyll build error with locally installed bundles in `vendor`


eliotsykes:
Thanks @stefanpenner , much appreciated!

stefanpenner:
thank you for the contribution :)

PR #4235, opened 2015-06-04
Reintroduce comment regarding bodyParser on http-mock

joostdevries:


Samsinite:
FYI, should be:

```md
//app.use('/api/<%= decamelizedModuleName %>', require('body-parser'), <%= camelizedModuleName %>Router);
```

joostdevries:
@Samsinite I don't actually think it's needed. It also reads better if they're separate.

stefanpenner:
my only concern is added extra comments to the generated output that can noisy up diffs. Thoughts?

joostdevries:
@stefanpenner two alternatives:
- Generate additional console output when generating the blueprint
- Generate console output when receiving a post/put call

stefanpenner:
@joostdevries im open to whatever you feel is appropriate.

joostdevries:
how would you feel about basically just trying to import the body-parser, `use` it if present and otherwise log a message on post/put?

stefanpenner:
i think it needs to be done on a case/case basis. So we don't know when or when not to include it even if present.

runspired:
Using body-parser like this causes all requests to the http-mock to hang.

stefanpenner:
Yes. That is why it is not enabled by default.

I have never had time to investigate why this is so fucked up. But something smells fishy

runspired:
It's fixed by making this change:

```
app.use('/api/<%= decamelizedModuleName %>', require('body-parser').json());
```

stefanpenner:
any idea why body-parser doesn't just do this automatically based on the content type?

jalcine:
On Thursday, September 17, 2015 06:57:23 AM Stefan Penner wrote:
> any idea why body-parser doesn't just do this automatically based on the
> content type?
> 
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/ember-cli/ember-cli/pull/4235#issuecomment-141094767

Probably want to allow the user to do checking on that (in the event that the content 
type is one kind of response but the body’s actually another).-- 
Jacky Alciné[1] // Work[2] // Blog[3] 
Software Engineer
---

They can READ + SEE everything in this email. In your texts.
The NSA's been spying on US citizens for too long.
Read more: https://www.eff.org/nsa-spying

--------
[1] https://jacky.wtf/?ref=email
[2] https://jacky.wtf/work/?ref=email
[3] https://jacky.wtf/weblog/?ref=email

PR #4240, opened 2015-06-04
Added ember-cli-release to app/addon devDeps blueprint for simple release cutting

jayphelps:
https://github.com/lytics/ember-cli-release

```bash
$ ember release
$ ember release --minor
$ ember release --major
```

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -29,6 +29,7 @@
     "ember-cli-ic-ajax": "0.1.1",
     "ember-cli-inject-live-reload": "^1.3.0",
     "ember-cli-qunit": "0.3.13",
+    "ember-cli-release": "0.2.3",
     "ember-cli-uglify": "^1.0.1",
     "ember-data": "1.0.0-beta.18",
     "ember-disable-proxy-controllers": "^1.0.0",


stefanpenner:
nice!

does this make sense for apps or just addons?

jayphelps:
@stefanpenner I'm going to be pushing to use it for our apps here at NF. We're moving to ember-cli-deploy as well, so seems natural to me.

That said, many people probably don't version their apps but since it's dev-only and noop if not used, I don't see any harm.

jayphelps:
IMO many people don't version their apps simply because it's not easy enough vs. the benefit for them. This addon makes it so simple that it may reduce that barrier for people and let them start versioning without all the work and worries.

stefanpenner:
> @stefanpenner I'm going to be pushing to use it for our apps here at NF. We're moving to ember-cli-deploy as well, so seems natural to me.
> 
> That said, many people probably don't version their apps but since it's dev-only and noop if not used, I don't see any harm.

sounds reasonable.

Lets give @slindberg some time to respond, just want to get his take.

slindberg:
SGTM :+1: 

We're currently using it to great effect with our app; it's made versioning with continuous deployment a breeze. The addon release story is a bit less fleshed out (see the unchecked bullets [here](https://github.com/lytics/ember-cli-release#workflow)), so I'd love to get more exposure to the needs of different release models.

stefanpenner:
> SGTM 
> 
> We're currently using it to great effect with our app; it's made versioning with continuous deployment a breeze. The addon release story is a bit less fleshed out (see the unchecked bullets here), so I'd love to get more exposure to the needs of different release models.

awesome!

jayphelps:
:+1: / :-1:

stefanpenner:
thanks for the ping

stefanpenner:
@jayphelps if you thoughts on how this should be made visible on the website, feel free to share :)

PR #4249, opened 2015-06-05
Passing options to tiny-lr (live reload) for HTTPS support

dosco:
Ember CLI is currently creating an instance of tiny-lr server without passing an options object which causes the live reload to always start to HTTP mode, with the newly added HTTPS support in express-server we an now pass the key and cert to tiny-lr allowing it to also launch in HTTPS mode.

Currently when ember serve is launched in HTTPS mode live reload will not work due to mixed-content issues in the browser.

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -2,41 +2,53 @@
 
 var Promise     = require('../../ext/promise');
 var path        = require('path');
+var fs          = require('fs');
 var Task        = require('../../models/task');
 var SilentError = require('silent-error');
 
-function createServer() {
+function createServer(options) {
   var instance;
+
   var Server = (require('tiny-lr')).Server;
   Server.prototype.error = function() {
     instance.error.apply(instance, arguments);
   };
-  instance = new Server();
+  instance = new Server(options);
   return instance;
 }
 
 module.exports = Task.extend({
-  liveReloadServer: function() {
+  liveReloadServer: function(options) {
     if (this._liveReloadServer) {
       return this._liveReloadServer;
     }
 
-    this._liveReloadServer = createServer();
+    this._liveReloadServer = createServer(options);
     return this._liveReloadServer;
   },
 
 
-  listen: function(port) {
-    var server = this.liveReloadServer();
+  listen: function(options) {
+    var server = this.liveReloadServer(options);
+
     return new Promise(function(resolve, reject) {
       server.error = reject;
-      server.listen(port, resolve);
+      server.listen(options.liveReloadPort, resolve);
     });
   },
 
   start: function(options) {
+    var tlroptions = {};
+    tlroptions.ssl = options.ssl || false;
+    tlroptions.port = options.liveReloadPort || 35729;
+
     if (options.liveReload !== true) {
       return Promise.resolve('Livereload server manually disabled.');
+    }
+
+    if (options.ssl) {
+      tlroptions.key = fs.readFileSync(options.sslKey);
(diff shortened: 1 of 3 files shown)


stefanpenner:
Nice. Can you add a test?

jrobeson:
Hah, i was going to submit the same PR. Do you think that live reload should have an option for its own key and cert and then default to options.sslKey, options.sslCert  if they aren't set? Or does that not really matter?

dosco:
@jrjohnson this is mostly for dev. environments where the certs are for the most part invalid but on the same domain as the app so having a different cert for LR would make it hard for dev's to accept the exception and to make it work in the browser.

@stefanpenner working on the tests.

jrobeson:
ah, i had setup my own local CA for dev, so all my certs are valid. I'm personally happy with this PR as is. :)

lib/tasks/server/livereload-server.js
```diff
@@ -47,7 +57,7 @@ module.exports = Task.extend({
     this.expressServer.on('restart', this.didRestart.bind(this));
 
     // Start LiveReload server
-    return this.listen(options.liveReloadPort)
+    return this.listen(options)
       .then(this.writeBanner.bind(this, options.liveReloadPort))
       .catch(this.writeErrorBanner.bind(this, options.liveReloadPort));
```
jrobeson: should we show that live reload is using ssl in the banner?

lib/tasks/server/livereload-server.js
```diff
@@ -2,35 +2,45 @@
...
 
-function createServer() {
+function createServer(options) {
   var instance;
+  var tinyLrOptions = {
+    port: options.liveReloadPort
+  };
+
+  if (options.ssl) {
+    tinyLrOptions.key = fs.readFileSync(options.sslKey),
```
jrobeson: line should end with a semicolon.

lib/tasks/server/livereload-server.js
```diff
@@ -2,35 +2,45 @@
...
+    tinyLrOptions.cert = fs.readFileSync(options.sslCert)
```
jrobeson: the above 2 lines should endd with semicolons

chadhietala:
@dosco Can you squash your commits? Otherwise looks good to me.

lib/tasks/server/livereload-server.js
```diff
@@ -2,43 +2,55 @@
...
-    var server = this.liveReloadServer();
+  listen: function(options) {
+    var server = this.liveReloadServer(options);
+
     return new Promise(function(resolve, reject) {
       server.error = reject;
-      server.listen(port, resolve);
-    });
+      server.listen(options.liveReloadPort, resolve);
+    }.bind(this));
```
stefanpenner: i don't see where `this` is used inside this function.

stefanpenner:
left a comment, otherwise LGTM

chadhietala:
I don't think Windows failures are related. Restarting for good measure. Also Coveralls has interesting opinions on math.

stefanpenner:
Needs rebase but Lgtm

chadhietala:
@dosco Can you rebase?

lib/tasks/server/livereload-server.js
stefanpenner: :+1:

lib/tasks/server/livereload-server.js
stefanpenner: also yes

stefanpenner:
two more comments

PR #4250, opened 2015-06-07
[ENHANCEMENT] Add `--local` to new, init, addon commands (closes #3380)

### LINKED ISSUE #3380: Proposal: add option to have git/bower use local caches only
There are a number of cli commands that depend on `npm` and `bower` which currently do not work properly without Internet access e.g. `new`, `init`, `addon`, `install`. 

I think it would be great to have something like an `--offline` option to these commands that would instruct bower and git to use their local caches rather than remote repositories. Currently:

* `bower` commands accept an `--offline` flag which does what we need
* `npm` is [discussing a similar flag](https://github.com/npm/npm/issues/2568) but in the meantime `npm install foo bar --cache-min 99999999` appears to work.

This would make `ember-cli` more usable in situations with no/bad/expensive Internet. I am happy to attempt a PR for this if it something that the core team is interested in exploring further.

eoinkelly:
First pass at this. 

#### tests

I have not added any tests yet because I'm not sure where to add them. I could add some tests that test that the option is passed around correctly. They would rely on mocking `npm-task` and `bower-task` and I'm not sure how much long term value they would provide. That said, I'm happy to add any tests that seem useful.

#### why `ember new foo --local` does not work on 2015-Jun-07

There is a dependency path `ember-cli -> testem -> socket.io -> socket.io-client -> engine.io-client` and `engine.io-client`  has the following in its `package.json`:

```
"dependencies": {
  ...
  "xmlhttprequest": "https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz",
```

`npm` can never satisfy this dependency from cache so it will fail without network access. This means that if you try `ember new foo --local` without network access it will not work (as of 2015-Jun-07). The problem has been fixed in [engine.io-client master](https://github.com/Automattic/engine.io-client) but they have not yet pushed an update to the [engine.io-client npm repo](https://www.npmjs.com/package/engine.io-client)

#### Changes summary

Closes #3380 add option to have git/bower use local caches only

Add `--local` option to commands:

* new
* init
* addon

Add message explaining how to use local cache to the deprecation message for the commands:

* install-npm
* install-bower

`--local` makes ember-cli do its best to execute the command without using the network. It instructs `npm` and `bower` to use their local package caches so is constrained by the "offline" options offered by `npm` and `bower`.

It does **not** guarantee that the command will successfully complete without network access. There are some situations where an `npm install` will fail without network access regardless of the state of its cache (see below)

`bower` and `npm` currently provide some "offline" support:

* Bower has an `--offline` flag which will install packages from its local cache and will fail if it cannot find them there.

* npm does not currently have real offline support. Instead we provide the `--cache-min=999999` option which instructs npm to use any package from the cache newer than the provided number of seconds. `npm` will fallback to trying the network if it cannot find the package in the cache.

Note the difference in cache-miss behavior between bower and npm. For this reason the flag is named `--local` which is abstract enough to apply to both behaviors.

There are some situations where npm cannot satisfy a dependency from its cache e.g. If `package.json` contains a dependency specified as

```js
"dependencies": {
  "xmlhttprequest": "https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz",
}
```

then `npm` will cache it but will never find it in the cache on subsequent installations.

If this occurs in any package that `ember-cli` depends on (directly or indirectly) then `npm` will fail to install without network access and the command will fail.


stefanpenner:
> why ember new foo --local does not work on 2015-Jun-07
> 
> There is a dependency path ember-cli -> testem -> socket.io -> socket.io-client -> engine.io-client and engine.io-client has the following in its package.json:
> 
> "dependencies": {
>   ...
>   "xmlhttprequest": "https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz",
> npm can never satisfy this dependency from cache so it will fail without network access. This means that if you try ember new foo --local without network access it will not work (as of 2015-Jun-07). The problem has been fixed in engine.io-client master but they have not yet pushed an update to the engine.io-client npm repo

ya we are waiting patiently on socketio here, we may just fork ...

stefanpenner:
is the git repo that last piece or..

stefanpenner:
@eoinkelly friendly ping.

eoinkelly:
FYI Just checked and the `engine.io-client` dependency issue (see above) is not resolved yet.

stefanpenner:
> FYI Just checked and the engine.io-client dependency issue (see above) is not resolved yet.
>  Hide all checks

it is

eoinkelly:
Doh, I only checked `engine.io-client` on npm for a fix, not `package.json`. Also yay! :smile:

Am working on the CI failures.

stefanpenner:
> Doh, I only checked engine.io-client on npm for a fix, not package.json. Also yay!

no problem, im not sure why they wont fix it upstream. But, I gave up and hit hot-patched it on our side.

stefanpenner:
> Am working on the CI failures.

any luck?

stefanpenner:
> Am working on the CI failures.
> any luck?

this seems to have gone stale, will gladly revisit if rebased and updated.

PR #4251, opened 2015-06-07
[BUGFIX] Fix generated addon acceptance test

### LINKED ISSUE #4191: Generating an acceptance test for addons produces a buggy test file
https://github.com/ember-cli/ember-cli/commit/07055c25b67fb3d13c071185ff55ac00823473e3 is the culprit

In short the `dasherizedPackageName` being used is going to look for `start-app` in the wrong namespace. For addons the proper namespace for the test support files is `dummy/`.

I recommend changing this back to the relative path name as there is no way to truly discern this value at generation. (the addon generator is overriding the `name` property to `"dummy"`)

trabus:
Fixes #4191 

Generates acceptance tests inside addons with a relative path instead of the addon namespace. Accounts for nested acceptance tests, generating the correct parent path.


stefanpenner:
:+1:

PR #4252, opened 2015-06-07
[WIP][ENHANCEMENT]  Add blueprint for ember-cli-command-test

trabus:
This PR adds a new blueprint to generate node tests for ember-cli commands. It is primarily for use in addons, but could be used in a project with an in-repo-addon. It generates the test, as well as a mocha runner (stolen from [ember-cli-divshot](https://github.com/rwjblue/ember-cli-divshot/blob/master/tests/runner.js)). Additionally, it updates the package.json devDependencies and adds `nodetest` to scripts.

I'm putting this up early (sans tests) for review, as I'm not totally sure what I'm doing here. Most of the code was extracted from ember-cli-divshot, since it was the only example of a test for an addon command I could find. My main motivation is that I'd like it to be easier for addon developers to test their commands by giving them a starting point to do so.

### DIFF
--- a/blueprints/ember-cli-command-test/files/tests/runner.js
+++ b/blueprints/ember-cli-command-test/files/tests/runner.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var glob = require('glob');
+var Mocha = require('mocha');
+
+var mocha = new Mocha({
+  reporter: 'spec'
+});
+
+var arg = process.argv[2];
+var root = 'tests/';
+
+function addFiles(mocha, files) {
+  glob.sync(root + files).forEach(mocha.addFile.bind(mocha));
+}
+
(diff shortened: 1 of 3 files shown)


trabus:
@rwjblue or @stefanpenner if you have any time, I would really appreciate some guidance on this.

stefanpenner:
I like the idea, i am nervous to code-gen that much could. Should we provide some libs to reduce the amount? (i suspect gen related to your blueprint helps may also be good)

Turbo87:
should we extract that into a separate addon instead which is included by default in the addon blueprint?

stefanpenner:
Add-on maybe, my above concern remains.

I'm going to close this, as it seems to have stagnated, we can revisit at our meeting :)

PR #4254, opened 2015-06-08
[BUGFIX] Closes #4253. Add `skipHelp` as an available option to commands.

### LINKED ISSUE #4253: `ember help` shows deprecated commands.
As the title says, certain deprecated or outright removed commands still show when running `ember help`. For example the `install:addon` deprecated command pops up under `ember help` along with the red warning that it is deprecated. 

To me it feels like deprecated and removed commands should not show under `ember-help` but maybe they should show when called by it's specific name. For example, `ember help install:addon` should still show given that it still exists, and removed commands should still show with the description that it no longer works, such as `ember install:npm`, `ember uninstall:npm`, etc.

DanielOchoa:
When setting `skipHelp: true` in a command, it will not show when calling `ember help` but it shows when calling `ember help <command-name>`. This allows us to remove certain deprecated and removed commands from showing under help. Fixes #4253.

### DIFF
--- a/lib/commands/help.js
+++ b/lib/commands/help.js
@@ -100,7 +100,9 @@ module.exports = Command.extend({
   _printHelpForCommand: function(commandName, detailed, options) {
     var command = this._lookupCommand(commandName);
 
-    command.printBasicHelp(options);
+    if (!command.skipHelp || (command.skipHelp && detailed)) {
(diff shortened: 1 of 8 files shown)


stefanpenner:
Lvgtm

trabus:
Nice! :+1:

Issue #4201, opened 2015-06-01
Remove private API's from component-test blueprint.


rwjblue:
Specifically, accessing `component._state` is private API (and may be removed in the future).

https://github.com/ember-cli/ember-cli/blob/master/blueprints/component-test/files/tests/unit/__path__/__test__.js#L14-L18

---

I suggest that the default test assert that the rendered output is empty (which will then convince people to  tweak it when they add content)...

Issue #3815, opened 2015-04-06
switch from string-concat to estemplate for blueprints


stefanpenner:
https://github.com/estools/estemplate

- [ ] estemplate needs to be upgraded to use babel's acorn

Why?

- string concat of blueprints doesn't ensure we emit actual JS
- better composition
- stuff like this: https://github.com/ember-cli/ember-cli/blob/master/blueprints/adapter/index.js#L16
- this may unlock our ability to blueprint code based on the style or .editorconfig of the current project

questions:

- do we need to compose some additional helpers  to make composition somewhat more ergonomic.

cc @RReverser @thejameskyle

stefanpenner:
@trabus / @jgwhite this may interest you.

sebmck:
> estemplate needs to be upgraded to use babel's acorn

Not sure how practical that is. It'd require doing `require("babel-core").acorn;`

stefanpenner:
> Not sure how practical that is. It'd require doing require("babel-core").acorn;

the goal is just to be sure to support the same syntax as babel

RReverser:
If @sebmck doesn't mind, I'd just wish to have features backported to Acorn itself, so that commonly used ES7 stuff could be used by any Acorn consumer. As for JSX and so, I don't think you guys want to have "the same syntax support" here :)

stefanpenner:
this will be a nice to have, but not really a short term bug we will tackle.

If someone has time and is motivated, this would be a great addition.

RReverser:
I kinda love issues which can be closed after some amount of time without any action needed :P

stefanpenner:
> I kinda love issues which can be closed after some amount of time without any action needed :P

i've been thinking of writing a bot, that assigns issues a half-life based on interaction and labels. Eventually they decay to some threshold and get closed, and finally closed for comments.

RReverser:
> and finally closed for comments

Do they bother that much? Why not just leave them for re-opening later?

stefanpenner:
> Do they bother that much? Why not just leave them for re-opening later?

On popular repo's yes, its noise that often hides more important issues. Making it hard for others to triage.

RReverser:
Too bad GitHub doesn't remember filters for repos, that could help.

stefanpenner:
> Too bad GitHub doesn't remember filters for repos, that could help.

[...]

PR #4272, opened 2015-06-10
[ENHANCEMENT] Generate component-tests into `tests/integration` by default

### LINKED ISSUE #4178: Update blueprints for components tests so they go under tests/integrations

trabus:
resolves #4178
By default component-tests now generate into `tests/integration`, with a new `--test-type` option allowing you to generate as a unit test intstead. The `--test-type` option can only be `unit` or `integration` and has the aliases `-unit`, `-u`, `-integration`, and `-i`.
```
ember g component-test foo-bar -unit
```

This PR also adds support for using an array of possible values as a command option type, accepting only the values in the array as valid arguments (essentially an enum type). For example, this is what the option for `test-type` looks like:

```
  availableOptions: [
    {
      name: 'test-type',
      type: ['integration', 'unit'],
      default: 'integration',
      aliases:[
        { 'i': 'integration'},
        { 'u': 'unit'},
        { 'integration': 'integration' },
        { 'unit': 'unit' }
      ]
    }
  ]
```

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -41,6 +41,12 @@ module.exports = {
     var componentPathName = dasherizedModuleName;
     var testTypeDefinition = "integration: true";
     var friendlyTestDescription = testInfo.description(options.entity.name, "Integration", "Component");
+    var testContent = "assert.expect(1);" + EOL + EOL +
+      "  // Set any properties with this.set('myProperty', 'value');" + EOL +
+      "  // Handle any actions with this.on('myAction', function(val) { ... });" + EOL + EOL +
+      "  // Provide a template (string or precompiled) for this.render()" + EOL +
+      "  this.render();" + EOL + EOL +
+      "  assert.equal(this.$().text(), '')";
 
     if (options.pod && options.path !== 'components' && options.path !== '') {
       componentPathName = [options.path, dasherizedModuleName].join('/');
@@ -49,13 +55,22 @@ module.exports = {
     if (options.testType === 'unit') {
       testTypeDefinition = "// Specify the other units that are required for this test" +
         EOL + "  // needs: ['component:foo', 'helper:bar']," + EOL + "  unit: true";
-        
+
+      testContent = "assert.expect(2);" + EOL + EOL +
+        "  // Creates the component instance" + EOL +
+        "  var component = this.subject();" + EOL +
+        "  assert.equal(component._state, 'preRender');" + EOL + EOL +
+        "  // Renders the component to the page" + EOL +
+        "  this.render();" + EOL +
+        "  assert.equal(component._state, 'inDOM');";
+
       friendlyTestDescription = testInfo.description(options.entity.name, "Unit", "Component");
     }
 
     return {
       path: getPathOption(options),
-      testType: options.testType ,
+      testType: options.testType,
+      testContent: testContent,
       componentPathName: componentPathName,
       testTypeDefinition: testTypeDefinition,
       friendlyTestDescription: friendlyTestDescription
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -4,13 +4,30 @@ var path          = require('path');
 var testInfo      = require('../../lib/utilities/test-info');
 var stringUtil    = require('../../lib/utilities/string');
 var getPathOption = require('../../lib/utilities/get-component-path-option');
-
+var EOL           = require('os').EOL;
 
 module.exports = {
-  description: 'Generates a component unit test.',
+  description: 'Generates a component integration or unit test.',
+  
+  availableOptions: [
+    {
+      name: 'test-type',
+      type: ['integration', 'unit'],
+      default: 'integration',
+      aliases:[
+        { 'i': 'integration'},
+        { 'u': 'unit'},
+        { 'integration': 'integration' },
+        { 'unit': 'unit' }
+      ]
+    }
+  ],
 
   fileMapTokens: function() {
     return {
+      __testType__: function(options) {
+        return options.locals.testType || 'integration';
+      },
       __path__: function(options) {
         if (options.pod) {
           return path.join(options.podPath, options.locals.path, options.dasherizedModuleName);
@@ -22,12 +39,21 @@ module.exports = {
   locals: function(options) {
     var dasherizedModuleName = stringUtil.dasherize(options.entity.name);
     var componentPathName = dasherizedModuleName;
-    if(options.pod && options.path !== 'components' && options.path !== '') {
+    var testTypeDefinition = "integration: true";
+    if (options.pod && options.path !== 'components' && options.path !== '') {
       componentPathName = [options.path, dasherizedModuleName].join('/');
     }
+    
+    if (options.testType === 'unit') {
+      testTypeDefinition = "// Specify the other units that are required for this test" + 
+        EOL + "  // needs: ['component:foo', 'helper:bar']," + EOL + "  unit: true";
+    }
+    
     return {
       path: getPathOption(options),
+      testType: options.testType ,
       componentPathName: componentPathName,
+      testTypeDefinition: testTypeDefinition,
       friendlyTestDescription: testInfo.description(options.entity.name, "Unit", "Component")
     };
   }
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -8,7 +8,7 @@ var EOL           = require('os').EOL;
 
 module.exports = {
   description: 'Generates a component integration or unit test.',
-  
+
   availableOptions: [
     {
       name: 'test-type',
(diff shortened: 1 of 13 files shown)


trabus:
@ef4 @stefanpenner ready for review!

kiwiupover:
thanks @trabus for your work here.

blueprints/component-test/index.js
```diff
@@ -4,13 +4,30 @@ var path          = require('path');
...
+        { 'integration': 'integration' },
+        { 'unit': 'unit' }
+      ]
+    }
+  ],
 
   fileMapTokens: function() {
     return {
+      __testType__: function(options) {
+        return options.locals.testType || 'integration';
```
stefanpenner: it seems like we have the "default" value specified in two locations. Should be merge to a single source of truth?

stefanpenner:
left one comment, it may be a valid concern. Otherwise though, LGTM

blueprints/component-test/index.js
```diff
@@ -22,12 +39,21 @@ module.exports = {
...
+      testTypeDefinition = "// Specify the other units that are required for this test" + 
+        EOL + "  // needs: ['component:foo', 'helper:bar']," + EOL + "  unit: true";
+    }
+    
     return {
       path: getPathOption(options),
+      testType: options.testType ,
       componentPathName: componentPathName,
+      testTypeDefinition: testTypeDefinition,
       friendlyTestDescription: testInfo.description(options.entity.name, "Unit", "Component")
```
pangratz: I think we should use different `friendlyTestDescription`s for each case...

ef4:
:+1:

blueprints/component-test/index.js
trabus: I'd like to clean this up so it's not necessary, but there are situations where [addon developers could override locals](https://github.com/ember-cli/ember-cli/issues/4183), making testType unavailable.

blueprints/component-test/index.js
trabus: Good catch. I'll put that in asap.

pangratz:
Wonderful :heart_eyes: 

The sample test case for `integration` needs to be a different case though, since using `this.subject` [will throw an error](http://git.io/vIPao) within an integration test ...

trabus:
@pangratz another good catch. on it now.

trabus:
Actually, this touches on #4201, but I think I'll leave the asserts in and deal with them in a PR for that issue.

trabus:
@pangratz Let me know if this is any better. Added some notes for setting up integration tests to the output.

blueprints/component-test/index.js
```diff
@@ -22,13 +39,41 @@ module.exports = {
...
     var dasherizedModuleName = stringUtil.dasherize(options.entity.name);
     var componentPathName = dasherizedModuleName;
-    if(options.pod && options.path !== 'components' && options.path !== '') {
+    var testTypeDefinition = "integration: true";
+    var friendlyTestDescription = testInfo.description(options.entity.name, "Integration", "Component");
+    var testContent = "assert.expect(1);" + EOL + EOL +
+      "  // Set any properties with this.set('myProperty', 'value');" + EOL +
+      "  // Handle any actions with this.on('myAction', function(val) { ... });" + EOL + EOL +
+      "  // Provide a template (string or precompiled) for this.render()" + EOL +
+      "  this.render();" + EOL + EOL +
```
pangratz: Something needs to be passed to `render`, as otherwise an [error is thrown](http://git.io/vIPAi). The problem is now that the `ember-template-compiler.js` needs to be present ... :cry:

I plan to open a PR for including `htmlbars-inline-precompile` as default, so this line will eventually be changed to:

``` js
this.render(hbs`
  {{<%= dasherizedModuleName %>}}
`);
```

Not sure what to do now? Sadly the current version of the integration test will throw an error, but maybe that's ok and this should be fixed before a new release is cut..

pangratz:
@trabus besides my comment, this looks good!

blueprints/component-test/index.js
trabus: Maybe a failing test to start isn't so bad?

trabus:
@pangratz Thanks for the review(s)!

@stefanpenner The changes I just had to make to have the testContent templated absolutely underscore the justification for implementing #3815.

blueprints/component-test/index.js
pangratz: > Maybe a failing test to start isn't so bad?

In some circumstances this might not be so bad, but the problem is just that fixing the failing test requires adding the `ember-template-compiler.js` to the `Brocfile` – or installing the `htmlbars-inline-precompile`.

Also, this would be the first time a generated test would fail ...

trabus: Can we just get away with adding an empty `'<span></span>'` or something?

pangratz: Unfortunately not since `ember-test-helpers` tries to compile a passed string via [`Ember.Handlebars.precompile()`](http://git.io/vIXtj) which will :boom: ...

pangratz: @stefanpenner https://github.com/babel/ember-cli-babel/pull/42 is currently blocking https://github.com/pangratz/ember-cli-htmlbars-inline-precompile/issues/7. When those are resolved, I will open a PR for using `htmlbars-inline-precompile` inside component integration tests, as suggested in https://github.com/pangratz/ember-cli-htmlbars-inline-precompile/issues/7#issuecomment-103967882.

blimmer:
I just implemented similar functionality to this in [ember-cli-mocha](https://github.com/switchfly/ember-cli-mocha/pull/57). @rwjblue asked me to do a PR to ember-cli to match that, but it's similar to this PR, so I'm hoping to get some feedback before I proceed.

Main Similarities:
* Introduces tests/integration/components/*-test.js

Main Differences:
* Uses [ember-cli-htmlbars-inline-precompile](https://github.com/pangratz/ember-cli-htmlbars-inline-precompile) to precompile the `render` statement.
* Does not change the default behavior when generating a component (a unit test is generated automatically and you have to explicitly generate component integration tests with `ember g component-test -i`)

What I'm wondering is:
1. Do we want to generate an integration test by default instead of a unit test?
2. Is there any issue adding a new dependency on ember-cli-htmlbars-inline-precompile?

rwjblue:
I would like to use integration by default, and use the inline precompiler for those integration tests. 

I haven't had a bunch of time to review this PR, though so am not 100% sure what it does....

trabus:
@blimmer I can bring this PR inline with yours regarding the ember-cli-handlebars-inline-precompile, but I think the default behavior of this PR is preferred (generating integration by default).

blimmer:
Sounds good. I can make a PR against ember-cli-mocha to bring the two back in sync (integration test by default and matching the arguments defined here).

trabus:
> Sounds good. I can make a PR against ember-cli-mocha to bring the two back in sync (integration test by default and matching the arguments defined here).

@blimmer  :+1:
Thanks!

blueprints/component-test/index.js
```diff
@@ -22,13 +39,43 @@ module.exports = {
...
     var dasherizedModuleName = stringUtil.dasherize(options.entity.name);
     var componentPathName = dasherizedModuleName;
-    if(options.pod && options.path !== 'components' && options.path !== '') {
+    var testImports = EOL + "import hbs from 'htmlbars-inline-precompile';" + EOL;
+    var testOptions = "integration: true";
+    var friendlyTestDescription = testInfo.description(options.entity.name, "Integration", "Component");
+    var testContent = "assert.expect(1);" + EOL + EOL +
+      "  // Set any properties with this.set('myProperty', 'value');" + EOL +
+      "  // Handle any actions with this.on('myAction', function(val) { ... });" + EOL + EOL +
+      "  this.render(hbs`{{" + dasherizedModuleName + "}}`);" + EOL + EOL +
```
pangratz: I wonder if this should be changed to show how components with a template block can be tested now? Maybe something like this:

``` js
"this.render(hbs`" + EOL
"  {{#" + dasherizedModuleName + "}}" + EOL
"    template block text" + EOL
"  {{/" + dasherizedModuleName + "}}" + EOL
"`);" + EOL + EOL +
"assert.equals(this.$().text().trim(), 'template block text');"
```

What do you think?

rwjblue:
:+1:

blueprints/component-test/index.js
trabus: Shoot, this didn't make it in, but now that it's in master, we can just do another PR for clarifiying block usage.

rwjblue: Yeah, sorry about that.  Thanks for cleaning it up.

trabus: So would it be preferred to default with a template block, or just have it as an example?

trabus: We could put both plain usage and block usage in, and leave it to the user to delete the part they're not using. I believe having both tests would pass on initial creation.

This is what I currently have:
```
      "  // Set any properties with this.set('myProperty', 'value');" + EOL +
      "  // Handle any actions with this.on('myAction', function(val) { ... });" + EOL + EOL +
      "  this.render(hbs`{{" + dasherizedModuleName + "}}`);" + EOL + EOL +
      "  assert.equal(this.$().text(), '')" + EOL + EOL +
      "  // Template block usage:"
      "  this.render(hbs`" + EOL
      "    {{#" + dasherizedModuleName + "}}" + EOL
      "      template block text" + EOL
      "    {{/" + dasherizedModuleName + "}}" + EOL
      "  `);" + EOL + EOL +
      "  assert.equals(this.$().text().trim(), 'template block text');"
```

rwjblue: Can you also delete the test that uses `component._state` which would resolve https://github.com/ember-cli/ember-cli/issues/4201?

trabus: Will do. Replacing with:
```
      testContent = "assert.expect(1);" + EOL + EOL +
        "  // Creates the component instance" + EOL +
        "  var component = this.subject();" + EOL +
        "  // Renders the component to the page" + EOL +
        "  this.render();" + EOL +
        "  assert.equal(this.$().text(), '');";
```

PR #4280, opened 2015-06-11
[ENHANCEMENT] Add pod support for services blueprint

trabus:
It was requested to generate services in pod structure so a project can group state manager services (ember-state-services) with the route they manage state for.

This PR simply adds the ability to generate services in pod structure.


stefanpenner:
LGTM

Issue #4277, opened 2015-06-10
kill <base href='smth'> or use it properly as in <base href='smth' />


zigotica:
hi,

can we change the code for base tag (closing the element as in <base href='smth' /> would fix many issues at least for IE) or kill it anytime soon-ish with fire?

is there an easy way I can modify that base markup locally while there is no final killing of this?

thank you :D

stefanpenner:
If you search this repo. The code for this is in a single place. Fix it and if the tests stay green it will be merged. That way you fix it for all

stefanpenner:
can you elaborate on which IE issues this causes?

stefanpenner:
killing base is a duplicate of: https://github.com/ember-cli/ember-cli/issues/2633

Im closing this, as I am not aware of non XHTML issues with HTML5 void elements. If someone can demonstrate and issue, please share and I will open up.

PR #4281, opened 2015-06-12
[FEATURE] adding self-closing tags to index.html

williamsbdev:


stefanpenner:
kicking appveyor

rwjblue:
I'm not opposed, but I don't see the point either. meta and link are void tags, and do not require closing.

What am I missing?

stefanpenner:
I am also trying to figure it out, someone reported an issue. But didn't elaborate

stefanpenner:
issue: 
https://github.com/ember-cli/ember-cli/issues/4277

apparently some IE issues?

stefanpenner:
I also believe this is now not valid HTML5.  investigating

stefanpenner:
it appears this is spec'd as non-closing -> http://www.w3.org/TR/html5/document-metadata.html#meta

stefanpenner:
https://github.com/ember-cli/ember-cli/issues/4277

williamsbdev:
I was unaware this is a limitation in the template language we are using at work. I will do more due diligence in the future to make sure that my PRs are according to spec. Thanks for the link to the spec! I appreciate you guidance.

stefanpenner:
@williamsbdev not a problem. I didn't know for sure either, so I had to dig in.

Unfortunately, I don't think we should deviate from the HTML5 spec here. I suspect as a solution you can just adjust your tags post generation.

williamsbdev:
@stefanpenner Absolutely. We are already doing that today and will continue forward with that. Thanks again!

PR #4283, opened 2015-06-12
Allow configuration of the LiveReload hostname

jrobeson:
This should help solve problems like:

 * https://github.com/rwjblue/ember-cli-content-security-policy/issues/18
 * https://github.com/rwjblue/ember-cli-inject-live-reload/issues/13

It can be configured on the cli as `ember server
--live-reload-host=example.org`

or in `.ember-cli`:

```json
{
  "liveReloadHost": "example.org"
}
```

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -33,14 +33,14 @@ module.exports = Task.extend({
 
     return new Promise(function(resolve, reject) {
       server.error = reject;
-      server.listen(options.port, resolve);
+      server.listen(options.port, options.host, resolve);
     });
   },
 
   start: function(options) {
     var tlroptions = {};
     tlroptions.ssl = options.ssl || false;
-    tlroptions.host = '0.0.0.0';
+    tlroptions.host = options.liveReloadHost || '0.0.0.0';
     tlroptions.port = options.liveReloadPort || 35729;
 
     if (options.liveReload !== true) {
--- a/tests/unit/tasks/server/livereload-server-test.js
+++ b/tests/unit/tasks/server/livereload-server-test.js
@@ -75,6 +75,16 @@ describe('livereload-server', function() {
           preexistingServer.close(done);
         });
     });
+
+    it('starts with custom host', function() {
+      return subject.start({
+        liveReloadHost: '127.0.0.1',
+        liveReloadPort: 1337,
+        liveReload: true
+      }).then(function() {
+        expect(ui.output).to.equal('Livereload server on http://127.0.0.1:1337' + EOL);
+      });
+    });
   });
 
   describe('start with https', function() {
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -18,6 +18,7 @@ module.exports = Command.extend({
     { name: 'insecure-proxy', type: Boolean, default: false, description: 'Set false to proxy self-signed SSL certificates', aliases: ['inspr'] },
     { name: 'watcher',  type: String, default: 'events', aliases: ['w'] },
     { name: 'live-reload',  type: Boolean, default: true, aliases: ['lr'] },
+    { name: 'live-reload-host',  type: String, description: 'Defaults to host', aliases: ['lrh'] },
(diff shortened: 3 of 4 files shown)


jrobeson:
I'll have to adjust the server test if https://github.com/ember-cli/ember-cli/pull/4282 lands

jrobeson:
Some folks might also want to be able to have a different configuration for the LiveReload server and the LiveReload script itself.

jrobeson:
rebased post https://github.com/ember-cli/ember-cli/pull/4282

tests/unit/tasks/server/livereload-server-test.js
```diff
@@ -75,6 +75,16 @@ describe('livereload-server', function() {
           preexistingServer.close(done);
         });
     });
+
+    it('starts with custom host', function() {
+      return subject.start({
+        liveReloadHost: '127.0.0.1',
```
jrobeson: if anybody has a suggestion for a better host that will always work, let me know.

stefanpenner: localhost ?

stefanpenner:
@lukemelia this should make your live-reload work in an extremely simple way.

lib/tasks/server/livereload-server.js
```diff
@@ -40,7 +40,7 @@ module.exports = Task.extend({
   start: function(options) {
     var tlroptions = {};
     tlroptions.ssl = options.ssl || false;
-    tlroptions.host = '0.0.0.0';
+    tlroptions.host = options.liveReloadHost || '0.0.0.0';
```
stefanpenner: should these two lines actually be defaults in the commands/server.js file?

stefanpenner:
Left some comments, otherwise this LGTM

tests/unit/tasks/server/livereload-server-test.js
jrobeson: but we already showed localhost, so i wanted there to be some visual change in the text since 0.0.0.0 gets transformed to localhost already.

stefanpenner: ah, good point. 127.0.0.1 should be fine.

If we could assume IPV6 support across all our test runners, `::1` may be an option.

lib/tasks/server/livereload-server.js
jrobeson: commands/serve.js?  there are defaults there already. I just followed the pattern i already saw.

chrislopresto:
Definitely like where this is headed.

lib/tasks/server/livereload-server.js
stefanpenner:
```
   { name: 'live-reload-host',  type: String, description: 'Defaults to host', aliases: ['lrh'] },
     { name: 'live-reload-port', type: Number, description: '(Defaults to port number + 31529)', aliases: ['lrp']},
```

don't appear to have defaults, instead the defaults exist: in the `start` function of livereload-server.js

Seems like they might be better suited as defaults in the command ?

jrobeson: isn't the code in the run method enough? As i can't see how i can default to host here, since host itself isn't decided yet.

jrobeson: by that i mean:  in lib/commands/serve.js line 33 in this PR

jrobeson:
I had missed a key part of this change, because i was working around the problem uncovered by this issue https://github.com/ember-cli/ember-cli/pull/4290

I'll have to rebase once a fix for that issue is integrated.

stefanpenner:
looking into this now.

stefanpenner:
> I had missed a key part of this change, because i was working around the problem uncovered by this issue #4290

is also now merged.

jrobeson:
rebased.

stefanpenner:
@jrobeson this is what i meant re: defaults https://github.com/ember-cli/ember-cli/pull/4292

stefanpenner:
superseded by the combined pr #4292

PR #4284, opened 2015-06-12
Update ember-resolver to 0.1.17.

rwjblue:
The primary change was the addition of `knownForType` method that will allow Ember 1.13.0 to automatically register all helpers.


rwjblue:
Updated to 0.1.17 which fixes a bug that @teddyzeenny found.

Issue #4276, opened 2015-06-10
Add generate support for addon dummy apps


trabus:
We should be able to generate blueprints into the `tests/dummy/app` that gets generated inside an addon. The option would be added to the generate command, and only work inside an addon.

```
ember g component foo-bar --structure dummy
```

This will be a good place to implement work on [Structures](https://github.com/ember-cli/rfcs/pull/5), with `dummy` being one of the structures available to ember-cli.

myartsev:
`ember g component foo-bar --dummy`

Just a note on how this got implemented.

PR #4287, opened 2015-06-12
[ENHANCEMENT] Add ability to generate blueprints into addon `tests/dummy/app`

trabus:
fulfils #4276 - 
This PR adds the ability to generate blueprints into an addon's `tests/dummy/app` folder. It only works from inside an addon project.
`ember g component foo-bar --dummy` with aliases `-du` or `-dum`.

The above command would generate:
```
my-addon/tests/dummy/app/components/foo-bar.js
my-addon/tests/dummy/app/templates/components/foo-bar.hbs
```

Does not work for blueprints without a `__root__` token.

### DIFF
--- a/blueprints/component/index.js
+++ b/blueprints/component/index.js
@@ -55,7 +55,7 @@ module.exports = {
     var importTemplate = '';
     var contents       = '';
     // if we're in an addon, build import statement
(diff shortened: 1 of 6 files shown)


stefanpenner:
LGTM

Issue #4279, opened 2015-06-11
LiveReload server doesn't listen on specified port


jrobeson:
The problem occurs after https://github.com/ember-cli/ember-cli/commit/bb3198aeadfc1d16269c07bcd2d3790bd5eb7742 was introduced.

The simple fix is to change https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/livereload-server.js#L36  to use `options.port` instead of `options.liveReloadPort`.

The tests for that feature pass because we're only testing the UI.  I'm not exactly sure what (or if) anything should be done here test wise without actually checking that the port is bound. Should we check the port property? or do a request?

EDIT: i almost submitted a PR, but didn't know what to do about the tests.

jrobeson:
NOTE:  the real bug is that tinylr doesn't validate the port option passed to it and simply starts up on some random port if one isn't specified.

PR #4290, opened 2015-06-13
Pass the correct port property to LiveReload server

jrobeson:
I filed a ticket (https://github.com/ember-cli/ember-cli/issues/4279) about this issue, but showing is probably better.

Currently the tests pass only incidentally, but LiveReload doesn't actually listen on the specified port.

tinylr doesn't validate its input nor we do truly check that tinylr is listening on that port.

Here (https://github.com/mklabs/tiny-lr/blob/master/lib/server.js#L23) it coerces it to the default port if options.port is undefined, but not here (https://github.com/mklabs/tiny-lr/blob/master/lib/server.js#L150)

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -33,7 +33,7 @@ module.exports = Task.extend({
 
     return new Promise(function(resolve, reject) {
       server.error = reject;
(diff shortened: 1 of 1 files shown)


stefanpenner:
thanks

PR #4292, opened 2015-06-13
Cleanup pr4283

stefanpenner:


lib/commands/serve.js
```diff
@@ -18,7 +18,8 @@ module.exports = Command.extend({
     { name: 'insecure-proxy', type: Boolean, default: false, description: 'Set false to proxy self-signed SSL certificates', aliases: ['inspr'] },
     { name: 'watcher',  type: String, default: 'events', aliases: ['w'] },
     { name: 'live-reload',  type: Boolean, default: true, aliases: ['lr'] },
-    { name: 'live-reload-port', type: Number, description: '(Defaults to port number + 31529)', aliases: ['lrp']},
+    { name: 'live-reload-host', default: '0.0.0.0', type: String, description: 'Defaults to host', aliases: ['lrh'] },
```
jrobeson: Maybe i don't get how the command parsing works, but I wanted it to work like this.

```bash
ember serve --host=example.org
```
and then `liveReloadHost` would have the value of `example.org`. It seems like now, it would have `0.0.0.0`

stefanpenner: so the goal is for the default host to merely inherit, seems reasonable

stefanpenner:
@jrobeson thoughts?

jrobeson:
lookin good to me.

stefanpenner:
@jrobeson awesome, ya i believe both our concerns have been addressed.

jrobeson:
ah i noticed a small thing.

stefanpenner:
> ah i noticed a small thing.

do share

lib/tasks/server/livereload-server.js
```diff
@@ -33,15 +33,16 @@ module.exports = Task.extend({
...
   },
 
   start: function(options) {
     var tlroptions = {};
-    tlroptions.ssl = options.ssl || false;
-    tlroptions.host = '0.0.0.0';
-    tlroptions.port = options.liveReloadPort || 35729;
+
+    tlroptions.ssl = options.ssl;
+    tlroptions.host = options.liveReloadHost || options.host;
```
jrobeson: do we still need this fallback here?

stefanpenner: I believe so, because users should be allowed to explicitly override it.

Issue #4315, opened 2015-06-20
Quest: Replace fs.existsSync with exists-sync in ember-cli and dependencies


trabus:
Ember-cli uses `fs.existsSync` fairly heavily in the codebase, as well as in many of the tests for ember-cli itself. Node.js has slated `fs.exists` and `fs.existsSync` for deprecation, so we need to migrate away from using it. I propose that we replace any usage of `fs.existsSync` in ember-cli with [`exists-sync`](https://github.com/ember-cli/exists-sync).

`exists-sync` provides the same basic functionality as `fs.existsSync`, but it also handles symlinks a little more intelligently. It will recursively resolve the target of a symlink, only returning true if an actual file is found (preventing false positives from symlinks).

It is also used throughout several broccoli plugins, so we should probably look to update any dependencies of ember-cli with it as well.

We should probably follow the pattern used [here](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/create-and-step-into-directory.js#L8):
```
var existsSync  = require('exists-sync');

// later
if (existsSync(filepath)) {
...
```

## instances to replace in ember-cli
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/broccoli-config-replace.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/index.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon-discovery.js
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/installation-checker.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/preprocessors/plugin.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/markdown-color.js 
- [ ] https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/version-utils.js 
[...]

PR #4316, opened 2015-06-20
fs.existsSync deprecated, replace with exists-sync

jasonmit:
https://github.com/ember-cli/ember-cli/issues/4315

~~I took a first pass at this, noticed the tests should likely be using tests/helpers/file-utils where possible for exists. Want me to address this as well @stefanpenner~~ 

I went ahead and updated the tests to use `file-utils.exists`

I will squash once thumbs up


tests/helpers/acceptance.js
```diff
@@ -26,10 +27,10 @@ function downloaded(item) {
   var exists = false;
   switch (item) {
     case 'node_modules':
-      exists = fs.existsSync(path.join(root, '.node_modules-tmp'));
+      exists = exists(path.join(root, '.node_modules-tmp'));
```
trabus: Will need to change these so you're not overwriting exists. :)

tests/helpers/acceptance.js
```diff
@@ -26,10 +27,10 @@ function downloaded(item) {
...
       break;
     case 'bower_components':
-      exists = fs.existsSync(path.join(root, '.bower_components-tmp'));
+      exists = exists(path.join(root, '.bower_components-tmp'));
```
trabus: Same here.

tests/helpers/acceptance.js
jasonmit: :+1:

trabus:
I'm wondering if maybe we should go with `existsSync` for all of it, to prevent any ambiguity about whether it's the async exists or not.

jasonmit:
> I'm wondering if maybe we should go with existsSync for all of it, to prevent any ambiguity about whether it's the async exists or not.

Yup, I agree.  I'll update accordingly.

tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -392,7 +393,7 @@ describe('Acceptance: brocfile-smoke-test', function() {
 
     return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'build', '--silent')
       .then(function() {
-        var exists = fs.existsSync(path.join('.', 'dist', 'assets', appName + '.css'));
+        var exists = exists(path.join('.', 'dist', 'assets', appName + '.css'));
```
trabus: Looks like another one here, but it won't matter when you've changed everything to use `existsSync` instead.

trabus:
LGTM :+1: 

@stefanpenner

jasonmit:
Squashed, thanks @trabus for the review

stefanpenner:
Looks good. That file-utile relative import should likely just point to the actual existsSync no?

jasonmit:
@stefanpenner yes and should already be doing that in this PR

tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -12,6 +12,7 @@ var runCommand          = require('../helpers/run-command');
 var acceptance          = require('../helpers/acceptance');
 var copyFixtureFiles    = require('../helpers/copy-fixture-files');
 var assertDirEmpty      = require('../helpers/assert-dir-empty');
+var existsSync          = require('../helpers/file-utils').existsSync;
```
stefanpenner: @jasonmit what about these?

jasonmit: file-utils uses exists-sync now https://github.com/jasonmit/ember-cli/blob/exist-sync/tests/helpers/file-utils.js#L4

Perhaps I'm misunderstanding

stefanpenner: yes, but why bother with the intermediate file at all?

jasonmit: Okay, will change this shortly

jasonmit:
@stefanpenner updated to remove intermediate file

stefanpenner:
Awesome thanks

trabus:
@jasonmit Thanks for taking care of this!

PR #4322, opened 2015-06-22
Update ADDON_HOOKS.md

jjmiv:
removed references to fs.existsSyncs with existsSync although there are no references to var fs.existsSync as this delivers example code for reference.


rwjblue:
Thank you!

PR #4326, opened 2015-06-23
Drop unused line from app blueprint

ef4:
This line is unnecessary and it causes JSHint failure if you're using `unused: true` on your tests tree.


stefanpenner:
Thanks

PR #4327, opened 2015-06-23
Bring tests jshintrc closer to app jshintrc

ef4:
These are both code-quality settings that have no reason to vary between app and tests, so I'm updating the tests version to match the app version.

### DIFF
--- a/blueprints/app/files/tests/.jshintrc
+++ b/blueprints/app/files/tests/.jshintrc
@@ -26,7 +26,7 @@
   "node": false,
   "browser": false,
   "boss": true,
-  "curly": false,
+  "curly": true,
   "debug": false,
   "devel": false,
   "eqeqeq": true,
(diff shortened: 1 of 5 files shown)


stefanpenner:
I like the approach,tests are unhappy and it needs a rebase

ef4:
Rebased and fixed failures. The remaining test failure is also failing on master, I don't think it is related to this PR.

stefanpenner:
Hmm strange. I'll poke around when I get home.

rwjblue:
Its a race condition with the new `ember-cli-build.js` file tests.  @chadhietala have a chance to review?

chadhietala:
I actually think there may be a race condition in the NPM client. As the failing test on 740274a should have never passed without https://github.com/ember-cli/ember-cli/pull/4348.

rwjblue:
Yes the NPM pack race condition is a constant troll.

PR #4347, opened 2015-06-25
ES3+ and for ES5+ deprecation free keys + forEach

stefanpenner:


stefanpenner:
@rwjblue r?

lib/utilities/reexport-template.js
```diff
@@ -1,7 +1,12 @@
 /* global define */
-define("{{DEST}}", ["{{SRC}}", "ember", "exports"], function(__index__, __Ember__, __exports__) {
-  "use strict";
-  __Ember__["default"].keys(__index__).forEach(function(key){
+define('{{DEST}}', ['{{SRC}}', 'ember', 'exports'], function(__index__, __Ember__, __exports__) {
+  'use strict';
+  var keys = Object.keys || __Ember__['default'].keys;
+  var forEach = Array.prototype.forEach && function(array, cb) {
```
rwjblue: This doesn't seem to be used. Below you are calling 'keys(__index__).forEach' which would use [].forEach (not this method), right?

stefanpenner: oops, even local tests caught it. I should have let them finish before asking for a review. Should be fixed now

PR #4356, opened 2015-06-27
Test is an example of a unit test, not an acceptance test

Gaurav0:


stefanpenner:
Good catch. Thank you kindly sir

PR #4348, opened 2015-06-26
Do not pack ember-cli-build.js

chadhietala:
Blacklist ember-cli-build.js from getting into the tarballed npm packages.

PR #4359, opened 2015-06-29
Update helper blueprint to use `Ember.Helper.helper`

balinterdi:
This is line with the new Helper API as described on
http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_new-ember-js-helper-api .

I assume here that the next version of Ember CLI will install Ember 1.13.x by default. If this is not the case, merging this PR should probably be postponed.

### DIFF
--- a/blueprints/helper/files/__root__/helpers/__name__.js
+++ b/blueprints/helper/files/__root__/helpers/__name__.js
@@ -4,4 +4,4 @@ export function <%= camelizedModuleName %>(params/*, hash*/) {
   return params;
 }
 
-export default Ember.HTMLBars.makeBoundHelper(<%= camelizedModuleName %>);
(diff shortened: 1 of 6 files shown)


rwjblue:
Change looks good (and yes master
of ember-cli is already setup for 1.13.2). Can you fix up the tests?

balinterdi:
I fixed the tests (I think) but the Appveyor build still fails for something that to me seems unrelated to the code changes. I also tried to force a rebuild by pushing an empty commit but that did not help. Does that happen a lot?

stefanpenner:
the windows failure has been happening for the last few days, it looks like am environment issue, but may also be related to https://github.com/ember-cli/ember-cli/pull/4348

cc @chadhietala

PR #4363, opened 2015-06-30
Fix using modules link

gbabiars:
Correctly uses the hash /user-guide#using-modules instead of
/user-guide/using-modules.


stefanpenner:
awesome thanks!

PR #4366, opened 2015-07-01
Remove typo

gregsabo:
I'm guessing we didn't mean to include this "the" before "Esperanto".


joostdevries:
:+1: 

justin-agrees

PR #4376, opened 2015-07-03
Update initializer-test blueprint

quaertym:
This is to avoid deprecation warning:
```
DEPRECATION: register should be called on the registry instead of the container
```


rwjblue:
:+1: to removing `__container__` from our blueprint.  You will likely need to update the blueprint tests for this to pass CI though.

quaertym:
CI seems alright.

rwjblue:
Then I guess we are missing some blueprinted file coverage here :smiley:

quaertym:
Added test and all green :beer:

Issue #4365, opened 2015-06-30
Coverage flag in testem test_page config


jrjohnson:
ember-cli-blanket (and maybe others) trigger test coverage stats by adding the ?coverage query param to the test_page option in testem.json.  I would like to make this a CLI flag so I don't waste time in travis generating coverage a bunch of times when I only need it once.  In taking a look at what that would entail it looks like there are two paths:

1.  Add a test-page option to the ember test command which would override whatever was in testem.json
2.  Add a --coverage option which would append &coverage onto the existing test_page config in buildTestPageQueryString() with module and filter.

Any thoughts on this?  Preferred path?  Wouldn't be interested in merging so I should figure out another way?

cc @sglanzer

sglanzer:
Of the two options described I like option 2 the best from a conceptual standpoint.  I doubt this is something ember-cli is going to want to add directly to the core, but maybe they would be comfortable with exposing a hook to capture cli options that could be passed to addons (such as ember-cli-blanket) which could then do the appropriate work.  Might even exist already and I'm just not aware of it.

cc @jschilli for comments

jschilli:
@jrjohnson @sglanzer agree that option 2 is superior conceptually.

I currently use a 2nd testem.json file with the coverage option enabled and just run `ember test --config testem.cov.json`

which when aliased via my shell is super low friction

jrjohnson:
fixed by #4377

sglanzer:
@jrjohnson Good stuff, would you mind documenting this in the ember-cli-blanket readme?

PR #4377, opened 2015-07-05
Add a test-page option to the test command

jrjohnson:
I took the less controversial path for fixing #4365.  This adds a --test-page option to the test command which allows arbitrary query parameters to be passed  for something like ember-cli-blankets ?coverage option.

eg 
```bash
ember test --test-page=tests/index.html?hidepassedtests&coverage
```

The (admittedly biased) discussion in #4365 was in favor of adding a --coverage option.  I'm worried that adding the option would imply coverage would run for new users even without an addon.  On the other hand its a lot clearer what to do with --coverage as opposed to --test-page.  Its a pretty simple change either way so all feedback welcome.


stefanpenner:
@jrjohnson nice.

Bikeshedding some here, but `--test-page` seems more like `--path`?

jrjohnson:
I didn't want to stray to far from the testem.json value which is test_page, but other than that I have no attachment to it.  Want me to switch it up?

stefanpenner:
:+1:

PR #4384, opened 2015-07-06
[ENHANCEMENT] Add block-template assertion to generated component integration test

### LINKED ISSUE #4201: Remove private API's from component-test blueprint.

trabus:
This PR continues work from #4272 adding a block-template assertion to the generated component integration test.

Also fixes #4201 and removed component._state from generated unit test


rwjblue:
:+1: - LGTM

PR #4385, opened 2015-07-06
Dependency updates

truenorth:


package.json
```diff
@@ -102,19 +102,19 @@
...
     "findup-sync": "^0.2.1",
-    "fs-extra": "0.16.5",
+    "fs-extra": "0.21.0",
     "git-repo-info": "^1.0.4",
     "github": "^0.2.3",
-    "glob": "5.0.3",
+    "glob": "5.0.13",
     "http-proxy": "^1.9.0",
     "inflection": "^1.7.0",
-    "inquirer": "0.5.1",
```
stefanpenner: we have had issues in the past with how inquirer changed its deps, we should be careful and confirm it actually still works.

truenorth: reverted

PR #4391, opened 2015-07-08
Bump ember-data version to 1.13.4

trabus:
:facepunch:


sly7-7:
Maybe better to bump to the last release, ie 1.13.4 ?

trabus:
Will do.

PR #4405, opened 2015-07-09
Fix default generated integration test.

blimmer:
The generated component integration test fails by default. It's because the result of:

```js
assert.equal(this.$().text().trim(), 'template block text');
```

is `' '` not `''`. We need to `.trim()` it.


trabus:
Thanks for catching this!

LGTM :+1:

Issue #4176, opened 2015-05-28
Feature request: Let "ember install" & "ember uninstall" take multiple addons


buschtoens:
It would be great for project initialization, if `ember install <addon>` could take a variable number of addons at once. This makes bootstrapping a new project way more comfortable.

```bash
$ ember install ember-cli-autoprefixer ember-cli-sass ember-cli-emblem ...
```

**EDIT**: As mentioned in #3163 this also applies to `ember uninstall <abc> <def> <ghj> ...`.

stefanpenner:
Ya, been meaning to add this one for some time.

stefanpenner:
* should install the NPM modules
* should aggregate generators that need to be run (for addons)
* should then run generators in the order the NPM module where installed

DanielOchoa:
- [x] `ember install` with multiple arguments
- [ ] `ember uninstall` with multiple arguments

stefanpenner:
> ember uninstall with multiple arguments

this really could just be `npm uninstall --save a b c` , but if you would like to add it is should be as simple

DanielOchoa:
what would be the difference between passing `--save` and not passing it?

DanielOchoa:
Also, with ember install the default generator is run. What would be the equivalent to ember uninstall? I was thinking about it and:

a) removing devDependencies means checking all addons dev dependencies to see if we can safely remove them from bower.json. 
b) for custom generator code, there's really no easy way to undo. 

Thoughts?

stefanpenner:
> a) removing devDependencies means checking all addons dev dependencies to see if we can safely remove them from bower.json.

this is also not easy (I would argue not possible without much additional meta data, that we dont currently store)

DanielOchoa:
On closer inspection I realized you are correct. This feature would need a new approach that would entail a rewrite of how `install` works as well. At least that's the only thing I can think of right now.

stefanpenner:
@DanielOchoa ya its larger problem, both NPM and bower would need to actually address for us. I suspect neither will anytime soon. So `ember uninstall` purely to add symmetry is fine, but not something we cant live without.

DanielOchoa:
sounds reasonable. I'll submit a PR most likely tomorrow :)

stefanpenner:
sounds good. Im going to close this issue, because the "bug like issue" has be resolved.

PR #4406, opened 2015-07-09
[ENHANCEMENT] Let ember install take multiple addons

DanielOchoa:
As the title says, it lets `ember install` take multiple addon names. Fixes half of #4176

### DIFF
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -4,6 +4,7 @@ var Task                = require('../models/task');
 var SilentError         = require('silent-error');
 var merge               = require('lodash/object/merge');
 var getPackageBaseName  = require('../utilities/get-package-base-name');
+var Promise             = require('../ext/promise');
 
 module.exports = Task.extend({
   init: function() {
@@ -14,8 +15,7 @@ module.exports = Task.extend({
   run: function(options) {
     var chalk            = require('chalk');
     var ui               = this.ui;
-    var packageName      = options['package'];
-    var extraArgs        = options.extraArgs || [];
+    var packageNames     = options['packages'];
     var blueprintOptions = options.blueprintOptions || {};
 
     var npmInstall = new this.NpmInstallTask({
@@ -34,14 +34,13 @@ module.exports = Task.extend({
     ui.startProgress(chalk.green('Installing addon package'), chalk.green('.'));
 
     return npmInstall.run({
-      packages: [packageName],
+      packages: packageNames,
       'save-dev': true,
       'save-exact': true
     }).then(function() {
       return this.project.reloadAddons();
     }.bind(this)).then(function() {
-      return this.installBlueprint(blueprintInstall, packageName, extraArgs,
-                                   blueprintOptions);
+      return this.installBlueprint(blueprintInstall, packageNames, blueprintOptions);
     }.bind(this))
     .finally(function() { ui.stopProgress(); })
     .then(function() {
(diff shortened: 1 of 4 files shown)


DanielOchoa:
well, it seems that the windows env does not like one of the tests... I'll dig into this sometime today

lib/tasks/addon-install.js
```diff
@@ -31,33 +32,37 @@ module.exports = Task.extend({
       testing:    this.testing
     });
 
-    ui.startProgress(chalk.green('Installing addon package'), chalk.green('.'));
+    var packageInflection = inflector.inflect('package', packageNames.length);
```
stefanpenner: do we need a full inflection library for this? Seems like this inflection is trivial no?

DanielOchoa: It is.. now that mention it.

stefanpenner:
I really like this, once CI is green we should :shipit:

DanielOchoa:
awesome.

trabus:
I restarted the appveyor build, that timeout happens occasionally.

DanielOchoa:
The timeout is related to my pr, more precisely the slow test to ember
install. Windows does not like (times out) when I try to rm the tmp dir.

DanielOchoa:
So let me know if you have any ideas. :)

stefanpenner:
If this persists it is likely a real bug. To test on windows https://github.com/ember-cli/ember-cli/tree/master/dev/windows may be of help

stefanpenner:
@DanielOchoa any luck?

DanielOchoa:
Not yet. The VM I setup from the link doesn't run any tests (it just hangs up). I'll spend some time this evening digging more through it.

kellyselden:
@DanielOchoa I created that windows VM code, I might be able to help you with any issues.

DanielOchoa:
That would be awesome.

tests/acceptance/install-test-slow.js
```diff
@@ -7,32 +7,29 @@ var assertFile = require('../helpers/assert-file');
...
-  beforeEach(function() {
-    tmpdir = tmp.in(tmproot);
-    process.chdir(tmpdir);
+    process.chdir(root);
   });
 
   afterEach(function() {
-    process.chdir(root);
-    return remove(tmproot);
+    return remove(tmproot, { maxBusyTries: 10 });
```
DanielOchoa: The test fails exactly on this line. (On appveyor.) It returns a busy signal error.

DanielOchoa: Well actually, with the maxBusyTries prop set to 10 it times out, when the prop is not set (defaults to 3) it returns the busy signal error.

DanielOchoa:
In other words, I'm very positive that the change works well on windows, it's just `tests/acceptance/install-test-slow.js` giving us a hard time in windows. As I mentioned before, I'll look at it again later today if @kellyselden hasn't found a fix. :)

stefanpenner:
@DanielOchoa hmm, so we need to figure out how to get CI green :(

DanielOchoa:
we'll get there. I'll take a break and look into it right now.

stefanpenner:
maybe we are accidentally creating a much more massive install then expected. We do some funky business to minimize cost of `npm` installs (lots of symlinking) maybe we are tripping something up.

DanielOchoa:
That seems related yes. The tmp dir in that specific test is created through regular means (I think) so I'll keep that in mind when debugging.

stefanpenner:
its green!

kellyselden:
:+1:

Issue #4409, opened 2015-07-09
No help entry for 'undefined'


tomdale:
Getting some random weird output running on 1.13.1 when I do `ember help`:

![code zsh solarized dark xterm-256color 115x53 2015-07-09 10-12-31](https://cloud.githubusercontent.com/assets/90888/8601766/08c3db64-2623-11e5-8f70-6bcfcb068eaf.jpg)

stefanpenner:
- [ ] we should also remove the `ember update` command entirely (or disable it for now, its a trap)

PR #4411, opened 2015-07-09
Removing unknown command from `ember help`.

gmurphey:
This addresses the original issue in #4409.


stefanpenner:
mind adding a quick test?

PR #4416, opened 2015-07-10
sourcemap-concat version

SSS2557:
@stefanpenner 

I am just leaving it out here as a reminder if that works :)
:pager:


stefanpenner:
Oh thanks!  Once ci is green I'll merge. It will be in the next release

PR #4420, opened 2015-07-10
Update project update link to account for different user guide page

avinmathew:
When I was following the Ember CLI 1.13 changelog, I noticed that the link to the "Project Update" section of the user guide was broken.

Maybe a broken link checker needs to be put in the docs build process?


rwjblue:
Thanks!

PR #4428, opened 2015-07-11
Update #watchman message to point to correct url

supabok:
Information for watchman should point to `http://www.ember-cli.com/user-guide/#watchman`


rwjblue:
Thank you

PR #4429, opened 2015-07-12
[FEATURE] use package.json key/values from outside project directory

noslouch:
In complicated web apps, one can find themselves with a folder structure like so (truncated):
```
complicated-web-app/
|
|__package.json
|
|__node_modules/
|
|__ember-app-one/
|   |__package.json
|   |__node_modules/
|
|__ember-app-two/
    |__package.json
    |__node_modules/
```

When it comes time to deploy, part of the process will require remembering to run `npm install` at the `complicated-web-app/` root as well as inside each `ember-app-one/` and `ember-app-two/`. 

This patch allows one to manage an ember app's dependencies from a location outside of its project directory using a package.json property named `alternateKeys`.

### alternateKeys
In an ember-cli project's package.json, specify an array of Objects, each containing two keys: `key String` and `file String`. 

`key` is the name of the key from the other package.json file you want use in this package. `file` is the path to that file. The path must relative to the project root. 

During cli commands, the instantiated `Project` object will have these values set on its `pkg` property for use in tasks.
```
// complicated-web-app/package.json
{ 
  "name": "complicated-web-app", 
  "version": 1.2.3, 
  "devDependencies": {
    "ember-cli": "1.13.1,
    "grunt": "0.4.5",
    // other complicated-web-app and ember deps
    },
}

// complicated-web-app/ember-app-one/package.json
{
  "name": "ember-app-one",
  "version": 0.1.1,
  "alternateKeys": [
    { "key": "devDependencies", "file": "../package.json" }
  ]
}

// imagine doing the same thing for ember-app-two
```

So this means you can put all your ember-cli deps into one package.json file at the root, tell the ember apps to look to that file for the dependencies and versions it should use, and symlink from the ember-app to the root node_modules.
```
$ cd complicated-web-app
$ npm install
$ rm -rf ember-app-one/node_modules && rm -rf ember-app-two/node_modules
$ cd ember-app-one
$ ln -s ../node_modules node_modules
$ cd ../ember-app-two
$ ln -s ../node_modules node_modules
```

### Why? Why do this?

I might have a bit of an edge case here, but I have myself in a situation as described above, and I don't want to be updating multiple package.json files during development and tracking how many folders need to be descended into to run `npm install`.

I'm hoping this functionality can be merged in so my team and I can continue to track the ember-cli project. Maybe there's a way to put this into an add-on, but dependency version mismatch happens pretty early on in the initialization process.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -46,7 +46,43 @@ function Project(root, pkg, ui, cli) {
   this.setupNodeModulesPath();
   this.addonDiscovery = new AddonDiscovery(this.ui);
   this.addonsFactory = new AddonsFactory(this, this);
+
+  if ( this.pkg.alternateKeys && this.pkg.alternateKeys.length ) {
+      this.setupAlternatePackageKeys();
+  }
 }
+
+/**
+ Use package.json data from an alternate location
+
+ @private
+ @method setupAlternates
+ */
+Project.prototype.setupAlternatePackageKeys = function() {
+    var alternates = this.pkg.alternateKeys;
+
+    debug('list of alternate package.json keys: %j', alternates);
+
+    alternates.forEach(function(a) {
+        var key         = a.key;
+        var packageFile = path.resolve(this.root, a.file);
+
+        if (fs.existsSync(packageFile)) {
+            var packageContents = fs.readFileSync(packageFile);
+            var originalValue = this.pkg[key];
+
+            try {
+                var alternateContents = JSON.parse(packageContents)[key];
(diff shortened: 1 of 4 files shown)


stefanpenner:
proposals should be opened as an issue to: https://github.com/ember-cli/rfcs
or if it is ready and fully formed, and RFC might be appropriate.

noslouch:
oh wow I didn't even know about that repo. I think this should be ready for an RFC, so I will follow the directions there.

PR #4430, opened 2015-07-12
Handle a wide variety of bower endpoints

### LINKED ISSUE #4425: addBowerPackageToProject does not support custom package names
In bower.json, I can place any dependency in any folder, as long as I specify a git branch or something to retrieve the source from

**bower.json**
```js
...
jquery: '^1.11.1',
notember: 'components/emberjs#beta'
...
```

When trying something similar in a blueprint's `addBowerPackageToProject`, it doesn't seem to permit a similar degree of flexibility  

```js
this.addBowerPackageToProject('notember', 'components/emberjs#beta'); // FAILS
```

truenorth:
Fixes #4425 

- [x] Failing test 
- [x] Build solution around [bower/endpoint-parser](https://github.com/bower/endpoint-parser)
- [x] Update documentation to provide examples of valid bower endpoints

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -31,7 +31,7 @@ var SilentError   = require('silent-error');
 var CoreObject    = require('core-object');
 var EOL           = require('os').EOL;
 var deprecateUI   = require('../utilities/deprecate').deprecateUI;
-
+var bowEpParser   = require('bower-endpoint-parser');
 module.exports = Blueprint;
 
 /**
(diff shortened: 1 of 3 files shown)


truenorth:
This is much simpler now. I'm using the same thing that bower uses internally for parsing `bower.json`, so developers will be able to use the same key-value pairs they're used to from `dependencies`, `devDependencies`, etc...

truenorth:
A little explanation as to why old tests needed to be updated, and why I believe it to be safe.


The typical use of `addBowerPackageToProject` is to install versioned (or non-versioned) bower packages in the `afterInstall` hook. These ultimately are transformed into "decomposed" package objects, containing three parts

* source - where to get the package from. This can be a git repo, a named package that has been published to bower, or even a zip file
* target - version constraint. Could be `"*"` or `"~1.11.1"`
* name - local name to use when mounting the package into the local `bower_components` folder

```js
addBowerPackageToProject('jquery', '~1.11.1');
// currently turns into --> {name: 'jquery', target: '~1.11.1'}
addBowerPackageToProject('jquery');
// currently turns into --> {name: 'jquery'}
```

The problem is that we've been using `name` in place of `source`. When using bower's own tooling to transform from "decomposed" to "composed" package information, we end up getting
`jquery=#~1.11.1` instead of `jquery#~1.11`. The general form of these stringified packages is 

```
<name>=<source>#<target>
```

What I have in this PR will maintain the existing `addBowerPackageToProject` and `addBowerPackagesToProject` usages, but it ends up being a little more verbose for the ember-cli internals. 

```diff
-        expect(packages).to.deep.equal([{name: 'foo-bar', target: '1.0.0'}]);
+        expect(packages).to.deep.equal([{name: 'foo-bar', source: 'foo-bar', target: '1.0.0'}]);
```

```diff
-      expect(packages).to.deep.equal(['foo-bar', 'bar-foo']);
+      expect(packages).to.deep.equal(['foo-bar=foo-bar', 'bar-foo=bar-foo']);
```

Where we technically should be getting `jquery#~1.11.1`, we'll get ``jquery#jquery=~1.11.1` -- the good news is that bower will treat these as equivalent, so the end result is the same

truenorth:
imo the coverage decrease should be ignored, since it's due to me reducing size of the of code in tested paths.

stefanpenner:
thanks :)

Issue #4434, opened 2015-07-13
User supplied options don't get merged into defaults with the new ember-cli-build


martndemus:
```js
  var app = new EmberApp(defaults, {
    babel: {
      optional: ['es7.decorators']
    }
  });

  console.log(app.options.babel);
```

```
=>  { plugins: [ [Function] ] } }
```

Currently i workaround by doing
```js
defaults.babel = { optional: ['es7.decorators'] };
var app = new EmberApp(defaults);
```

rwjblue:
My guess is that these are not deeply merged. Definitely a bug, thanks for reporting!

stefanpenner:
cc @chadhietala I think `defaults` doesn't deep merge here. I mentioned it in your earlier (like 3 minutes ago PR)

rwjblue:
Copied over from https://github.com/ember-cli/ember-cli/pull/4440/files#r34471644:

We use a slightly different pattern in ember-app for deeply merged values ([see here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L199-L212)), which is something like:

```javascript
var defaults      = require('lodash/object/defaults');
var merge         = require('lodash/object/merge');

options = merge({}, args, defaults);
```

`merge` takes a "customizer" ([see here](https://lodash.com/docs#merge)) that we use as the standard `defaults`.  This lets us handle defaults at the second levels, but if we want to fully merge at all depths we will need to write our own customizer function.

rwjblue:
We need to decide how deeply we want to merge though, given the following objects?

```javascript
var a = {
  someAddon: {
    configA: 'asdf',
    configB: 'hjkl'
  },
  listOfThings: [ 'a', 'b', 'c'],
  objectsWithArrays: {
    thingA: [ 'foo', 'bar' ]
  }
};

var b = {
  someAddon: {
    configC: 'yuio'
  },
  listOfThings: [ 'd', 'e', 'f' ],
  objectsWithArrays: {
    thingA: [ 'baz', 'qux' ]
  }
};

var app = new EmberApp(a, b);
```

What should the final options be?

martndemus:
Rightside wins, treat an array as primitive.

```js
{
  someAddon: {
    configA: 'asdf',
    configB: 'hjkl',
    configC: 'yuio'
  },
  listOfThings: [ 'd', 'e', 'f' ],
  objectsWithArrays: {
    thingA: [ 'baz', 'qux' ]
  }
}
```

PR #4440, opened 2015-07-13
EmberAddon Should Merge Defaults

### LINKED ISSUE #4435: ember-cli-build.js template for addons doesn't allow passing options
The default template for `ember-cli-build.js` in an addon looks something like:

```js
/* global require, module */
var EmberApp = require('ember-cli/lib/broccoli/ember-addon');

  var app = new EmberApp(defaults, {
    // Add options here
  });

  /*
    This build file specifes the options for the dummy test app of this
    addon, located in `/tests/dummy`
    This build file does *not* influence how the addon or the app using it
    behave. You most likely want to be modifying `./index.js` or app's build file
  */

  return app.toTree();
};
```

However, `ember-cli/lib/broccoli/ember-addon` expects only one argument (the `options`) so options put in `// Add options here` don't get successfully passed into the app.

chadhietala:
This closes #4435.

### DIFF
--- a/lib/broccoli/ember-addon.js
+++ b/lib/broccoli/ember-addon.js
@@ -19,8 +19,20 @@ module.exports = EmberAddon;
   @constructor
   @param options
 */
-function EmberAddon(options) {
-  options = options || {};
+function EmberAddon() {
+  var args = [];
+  var options = {};
+
+  for (var i = 0, l = arguments.length; i < l; i++) {
+    args.push(arguments[i]);
+  }
+
+  if (args.length === 1) {
+    options = args[0];
+  } else if (args.length > 1) {
+    options = defaults.apply(null, args);
+  }
+
   process.env.EMBER_ADDON_ENV = process.env.EMBER_ADDON_ENV || 'development';
 
   this.appConstructor(merge(options, {
--- a/lib/broccoli/ember-addon.js
+++ b/lib/broccoli/ember-addon.js
@@ -30,6 +30,7 @@ function EmberAddon() {
   if (args.length === 1) {
     options = args[0];
   } else if (args.length > 1) {
+    args.reverse();
     options = defaults.apply(null, args);
   }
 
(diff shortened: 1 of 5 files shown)


lib/broccoli/ember-addon.js
```diff
@@ -19,8 +19,20 @@ module.exports = EmberAddon;
...
+  var options = {};
+
+  for (var i = 0, l = arguments.length; i < l; i++) {
+    args.push(arguments[i]);
+  }
+
+  if (args.length === 1) {
+    options = args[0];
+  } else if (args.length > 1) {
+    options = defaults.apply(null, args);
```
stefanpenner: does defaults merge multiple levels? (is that needed?)

stefanpenner:
seems like once this lands we need to cut a new release.

lib/broccoli/ember-addon.js
rwjblue: We use a slightly different pattern in ember-app for deeply merged values ([see here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L199-L212)), which is something like:

```javascript
var defaults      = require('lodash/object/defaults');
var merge         = require('lodash/object/merge');

var defaults      = require('lodash/object/defaults');
var merge         = require('lodash/object/merge');

options = merge({}, args, defaults);
```

chadhietala:
@stefanpenner @rwjblue Added `merge-defaults` which will recursively merge defaults.

stefanpenner:
kicking travis

stefanpenner:
@chadhietala awesome!

rwjblue:
Can this PR also address #4434 (or maybe it does)?
Can we have a test to ensure we don't regress in the future?

chadhietala:
@rwjblue addresses #4434. Added tests.

stefanpenner:
looks correct.

chadhietala:
Added a third level array.

stefanpenner:
> Added a third level array.

thanks, :shipit: when green.

rwjblue:
Thank you!

PR #4443, opened 2015-07-13
Prevent live-reload-port collisions (by default)

stefanpenner:
Unless explicitly overridden with `—live-reload-port`, use `PortFinder` to find a free port between `49152` and `65535`

- [x] confirm this works on windows...
- [x] confirm this works with CSP add-on (a quick skim says it should) cc @rwjblue


stefanpenner:
cc @eccegordo

eccegordo:
This is pretty awesome. Actually solves some real pain points. Thanks!

PR #4444, opened 2015-07-13
Update the overview section because of typos

mikewink:
Fixed some typos, formatting and changed wording for the CSP section as the support got much better in the mean time - according to caniuse.com


stefanpenner:
thanks :)

Issue #3299, opened 2015-02-16
Test-specific EmberENV


ef4:
If you run `ember serve` in dev mode, and visit the `/tests`, you end up with two conflicting EmberENV configs.

The one in `<meta>` correctly contains the test environment, but it gets ignored. The one at the start of `vendor.js` will contain the dev environment, and it gets used.

bjornstar:
Yeah, this is very frustrating.

I get the feeling that `ember test --server` is the "standard" for testing and visiting `/tests` on `ember serve` is not supported. Personally, I prefer running `ember serve` so that I can run my app and my tests at the same time against the same code.

stefanpenner:
Both should be made to works as expected

rwjblue:
Agreed, I'm pretty sure this can be fixed fairly easily. Will look at it in more detail in the morning.

sohara:
:+1: We are using `baseURL` in one of our apps. Testem needs baseURL to be set to '/' but that prevents us from simply hitting '/tests' in our dev environment to run the tests.

Turbo87:
closing this in favor of https://github.com/ember-cli/ember-cli/issues/4535

PR #4448, opened 2015-07-14
Add some warning to anyone using "/tests"

searls:
After attempting to use /tests yesterday, I wish this text had read a bit more like a warning (it would have flagged me off from even trying, because our app has a bunch of test-specific env flags.)

### DIFF
--- a/_posts/2013-04-11-testing.md
+++ b/_posts/2013-04-11-testing.md
@@ -16,7 +16,7 @@ ember test --server   # will run your tests on every file-change
 {% endhighlight %}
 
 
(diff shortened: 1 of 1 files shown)


trek:
In general we prefer affirmative, positive statements to "warnings" and "notes". Can you express this in way that describes the behaviors, not the bugs?

searls:
Apologies, @trek, I pushed this just moments before leaving for vacation and won't have a chance to get to my computer for a while. :(

stefanpenner:
thanks for the contribution. I have adjusted the wording to express the current behavior rather then the bug as @trek recommended

https://github.com/ember-cli/ember-cli/commit/7fc88e890986361254892b4c760227762746d9dd

stefanpenner:
existing issue: https://github.com/ember-cli/ember-cli/issues/3299

searls:
Thanks to you both!

PR #4454, opened 2015-07-14
Always use Brocfile (with deprecation messaging) if it exists

gmurphey:
A number of people have been tripped up by the addition of `ember-cli-build.js`, not realizing that their `Brocfile.js` is ignored. This updates the build logic to always use the `Brocfile.js` if it is present.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -21,10 +21,12 @@ module.exports = Task.extend({
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
 
     var broccoli = require('broccoli');
+    var hasBrocfile = existsSync(path.join('.', 'Brocfile.js'));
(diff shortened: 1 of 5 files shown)


rwjblue:
:+1:

stefanpenner:
:+1:  and thanks for the test

gmurphey:
Investigating this test error now.

gmurphey:
It took a few more fixups than I'd hoped, but this is ready to review now. It turns out that the `Brocfile.js` deprecation was never showing up because the `deprecate` utility requires a boolean `test` argument.

stefanpenner:
@gmurphey your contributions have been awesome, great work!

gmurphey:
@stefanpenner thanks! i'll try to keep them coming.

stefanpenner:
@gmurphey please do

PR #4456, opened 2015-07-14
Updating getPort logic to use liveReloadHost. Fixes #4455.

### LINKED ISSUE #4455: On Windows, `ember serve` does not start with default livereload options
Developing against `master` today, I noticed that `ember serve` will not start on Windows, most likely due to an [issue with portfinder](https://github.com/indexzero/node-portfinder/issues/13). Calling `getPort` with the default options, it always returns 49152, but `wininit.exe` is already using that port on `0.0.0.0`. Should we check both `127.0.0.1` and `0.0.0.0` for available ports before starting up livereload?

gmurphey:


stefanpenner:
Does getPort take host into account?

gmurphey:
https://github.com/indexzero/node-portfinder/blob/master/lib/portfinder.js#L39

It seems so, and in my testing, it returns a different available ports for `127.0.0.1` and `0.0.0.0` (at least on Windows). Investigating this test failure now, and looking refactoring to better test that `getPort` is sent the correct arguments.

stefanpenner:
> It seems so, and in my testing, it returns unavailable ports when it defaults to null (at least on Windows). Investigating this test failure now, and looking refactoring to better test that getPort is sent the correct arguments.
>  Hide all checks

interesting.

gmurphey:
Here's the behavior I'm seeing on my machine, just as a demonstration.

capture

stefanpenner:
strange, but great catch. This would have taken me much time to figure out!

PR #4457, opened 2015-07-15
Removing extraneous newline from the component-test blueprint

gmurphey:
The `component-test` blueprint was adding an unnecessary newline after `testImports`.


rwjblue:
Thanks!

PR #4471, opened 2015-07-17
Make the Examples less confusing

mdragon:
It was really confusing to me what really needed to happen based on the original examples.  I tried what seemed to be the best guess and it worked, but hopefully this will make it clearer for others.


stefanpenner:
i like!

PR #4472, opened 2015-07-17
Fix typo

fishermand46:


lib/commands/test.js
```diff
@@ -10,7 +10,7 @@ var path = require('path');
 module.exports = Command.extend({
   name: 'test',
   aliases: ['test', 't'],
-  description: 'Runs your apps test suite.',
```
stefanpenner: this isn't valid javascript

fishermand46:
I'm an idiot. It's fixed.

PR #4474, opened 2015-07-17
[fixes #4328] patch engion.io-client to use any XMLHTTPRequest, but t…

### LINKED ISSUE #4328: Unable to install when github it not reachable
I work in an environment where we have, basically, a clone of npm. However, we do not have access to github. The issue is that when you try to install ember-cli via npm, it tries to fetch things from github, which completely breaks the install. Is there a way around this? The first thing it fails to get is node-XMLHttpRequest from https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz.

stefanpenner:
…hen bundle the one it wanted in ember-cli’s tarball

pending: https://github.com/socketio/engine.io-client/issues/405

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,13 +18,13 @@ install:
   # Typical npm stuff.
   - md C:\nc
   - npm install -g npm@latest
-  # hide python so node-gyp won't try to build native extentions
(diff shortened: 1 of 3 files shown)


stefanpenner:
quickly merge it before appveyor disagrees :P

PR #4478, opened 2015-07-17
Store acceptance test application in test context.

rwjblue:
This allows custom test assertions to access the application without manually passing in the application.


stefanpenner:
@rwjblue do the contexts get released, so we don't leak the whole app?

rwjblue:
Both QUnit and Mocha create a new context for each test.

PR #4484, opened 2015-07-18
Updating route blueprint to write to router when dummy flag is used

### LINKED ISSUE #4404: Generating routes with `--dummy` does not update `tests/dummy/app/router.js`
Because the dummy folder is contained in an addon, the `router.js` is currently not updated when generating a new route inside a dummy app. We should account for this by checking for the dummy flag [here](https://github.com/ember-cli/ember-cli/blob/master/blueprints/route/index.js#L63) and [here](https://github.com/ember-cli/ember-cli/blob/master/blueprints/route/index.js#L76)

gmurphey:
Fixes #4404. /cc @trabus

### DIFF
--- a/blueprints/route/index.js
+++ b/blueprints/route/index.js
@@ -81,12 +81,17 @@ module.exports = {
 
 function updateRouter(action, options) {
   var entity = options.entity;
+  var actionColorMap = {
+    add: 'green',
+    remove: 'red'
+  };
+  var color = actionColorMap[action] || 'gray';
 
   if (this.shouldTouchRouter(entity.name, options)) {
     writeRoute(action, entity.name, options);
 
     this.ui.writeLine('updating router');
-    this._writeStatusToUI(chalk.red, action + ' route', entity.name);
+    this._writeStatusToUI(chalk[color], action + ' route', entity.name);
   }
 }
 
--- a/blueprints/route/index.js
+++ b/blueprints/route/index.js
@@ -53,7 +53,7 @@ module.exports = {
     };
   },
 
-  shouldTouchRouter: function(name) {
+  shouldEntityTouchRouter: function(name) {
     var isIndex = name === 'index';
     var isBasic = name === 'basic';
     var isApplication = name === 'application';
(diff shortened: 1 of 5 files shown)


gmurphey:
I also did a bit of refactor to the route blueprint to DRY up the `afterInstall` / `afterUninstall` hooks.

gmurphey:
Not sure what's going on with AppVeyor. I'll try to hop on my Windows machine later tonight to figure this out.

tests/acceptance/addon-dummy-generate-test.js
```diff
@@ -272,6 +275,13 @@ describe('Acceptance: ember generate in-addon-dummy', function() {
       assertFile('tests/dummy/app/templates/foo/bar.hbs', {
         contains: '{{outlet}}'
       });
+      assertFile('tests/dummy/app/router.js', {
+        contains: [
+          "this.route('foo', function() {" + EOL,
```
trabus: I don't think asserting the EOL's is necessary, probably what is tripping up Appveyor.

trabus: Hmm, I guess we are asserting them on other tests, nevermind.

stefanpenner:
```
not ok 6 PhantomJS 2.0 - JSHint - .: failing test
 
    ---
        actual: >
            false
        expected: >
            true
        message: >
            test should fail to confirm ember test exit code
        Log: >
    ...
 
ok 7 PhantomJS 2.0 - JSHint - unit: unit/some-test.js should pass jshint
 
 
```

tests/acceptance/addon-dummy-generate-test.js
gmurphey: I'm wondering if using `EOL` in tests works with code generated by ember-cli/ember-router-generator. I'm not seeing `EOL` included in anything that's created with `recast`.

gmurphey:
This should be ready for review now. I've removed `EOL` from the `router.js` tests, and that seems to fix everything. I looked and couldn't find another instance of using `EOL` in tests against `router.js`, so I'm assuming ember-router-generator only uses Unix style line-endings.

blueprints/route/index.js
```diff
@@ -61,49 +61,53 @@ module.exports = {
...
+  var entity = options.entity;
 
-      this.ui.writeLine('updating router');
-      this._writeStatusToUI(chalk.red, 'remove route', entity.name);
-    }
+  if (this.shouldTouchRouter(entity.name, options)) {
+    writeRoute(action, entity.name, options);
+
+    this.ui.writeLine('updating router');
+    this._writeStatusToUI(chalk.red, action + ' route', entity.name);
```
trabus: Is there a specific reason we're using red for adding and removing routes? Might be confusing for users.

trabus:
Aside from the status color issue, the rest looks :+1: to me.

blueprints/route/index.js
gmurphey: Good point -- I'll update this now!

PR #4485, opened 2015-07-18
[INTERNAL] Updating test runner to run jshint first

gmurphey:
I fail hard sometimes and go crazy with parentheses, and my mistakes might not show themselves until two or three minutes into a run of `npm test`. I believe `jshint-test.js` should always run first to help poor, flubby fingers like mine.

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,3 +1,4 @@
 node_modules/*
-blueprints/*
+coverage/*
+blueprints/**/files/**/*.js
 tests/fixtures/*
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -16,18 +16,24 @@ var mocha = new Mocha({
   timeout: 5000,
   reporter: 'spec'
 });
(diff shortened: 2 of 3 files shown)


gmurphey:
After a bit of testing, I don't think that `jshint-helper.js` is really doing anything. Doing a bit of investigation...

gmurphey:
My latest commit updates `mocha-jshint` to the latest version and sets it up in what I believe is the correct way. As you can see by looking at the CI, it seems that setting it up to run on top-level folders wasn't doing what we wanted and not testing against many (if any) files.

I can work on fixes for these linting issues here or, if you'd like, I can open a separate issue for these.

gmurphey:
Okay, I think I have this worked out now. I've updated `mocha-jshint` to use `.jshintignore` for its configuration, and added the correct files there. All the files failing on CI were blueprint files.

stefanpenner:
:+1: thank you kindly sir.

PR #4504, opened 2015-07-22
Improve testing performance - use buildInstance to avoid recreating the application

pete-the-pete:
This depends on having [qunit-once](https://github.com/bahmutov/qunit-once) available ([ember-cli-qunit PR](https://github.com/ember-cli/ember-cli-qunit/pull/70))

Updating startApp to return an instance and the application from separate functions.

Using the qunit-once hooks lets acceptance tests request, and destroy, the
application and instance separately.  The application is created and destroyed once
per module, while the instance is created and destroyed for each test.

This should reduce the time it takes to spin up a test, and offer a step toward better parallelization.

### DIFF
--- a/blueprints/acceptance-test/meta/instance
+++ b/blueprints/acceptance-test/meta/instance
@@ -1,0 +1,15 @@
+  setupOnce: function() {
+    this.application = startApp();
+  },
+
+  setup: function() {
+    this.appInstance = getAppInstance(this.application);
+  },
+
+  teardown: function() {
+    Ember.run(this.appInstance, 'destroy');
+  },
+
+  teardownOnce: function() {
+    Ember.run(this.application, 'destroy');
+  }
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -2,7 +2,8 @@ import Ember from 'ember';
 import Application from '../../app';
 import config from '../../config/environment';
 
-export default function startApp(attrs) {
+
+export function startApp(attrs) {
   var application;
 
   var attributes = Ember.merge({}, config.APP);
@@ -16,3 +17,10 @@ export default function startApp(attrs) {
 
   return application;
 }
+
+export function getAppInstance(application) {
+  return Ember.run(application, 'buildInstance');
+}
+
+// export startApp as default for backwards compatability
+export default startApp;
--- a/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
@@ -1,16 +1,9 @@
 import Ember from 'ember';
 import { module, test } from 'qunit';
-import startApp from '<%= testFolderRoot %>/tests/helpers/start-app';
+import { <%= imports %> } from '<%= testFolderRoot %>/tests/helpers/start-app';
 
 module('<%= friendlyTestName %>', {
-  beforeEach: function() {
-    this.application = startApp();
-  },
-
-  afterEach: function() {
-    Ember.run(this.application, 'destroy');
-  }
-});
+<%= lifeCycleHooks %>});
 
 test('visiting /<%= dasherizedModuleName %>', function(assert) {
   visit('/<%= dasherizedModuleName %>');
--- a/blueprints/acceptance-test/index.js
+++ b/blueprints/acceptance-test/index.js
@@ -3,17 +3,42 @@
 var testInfo    = require('ember-cli-test-info');
 var pathUtil    = require('ember-cli-path-utils');
 var stringUtils = require('ember-cli-string-utils');
+var path        = require('path');
+var fs          = require('fs');
 
 module.exports = {
   description: 'Generates an acceptance test for a feature.',
+
+  availableOptions: [
+    {
+      name: '--instance',
+      type: Boolean,
+      default: false
+    }
+  ],
+
   locals: function(options) {
+    var metaDir = path.join(__dirname, 'meta');
+    var lifeCycleHooksTemplate = 'default';
+    var lifeCycleHooksTemplatePath = '';
+    var startAppImmports = ['startApp'];
     var testFolderRoot = stringUtils.dasherize(options.project.name());
+
+    if (options.instance) {
+      startAppImmports.push('getAppInstance');
+      lifeCycleHooksTemplate = 'instance';
+    }
 
     if (options.project.isEmberCLIAddon()) {
       testFolderRoot = pathUtil.getRelativeParentPath(options.entity.name, -1, false);
     }
+
+    lifeCycleHooksTemplatePath = path.join(metaDir, lifeCycleHooksTemplate);
+
     return {
       testFolderRoot: testFolderRoot,
+      imports: startAppImmports.join(', '),
+      lifeCycleHooks: fs.readFileSync(lifeCycleHooksTemplatePath, { encoding: 'utf8' }),
       friendlyTestName: testInfo.name(options.entity.name, "Acceptance", null)
     };
   }
--- a/blueprints/acceptance-test/meta/default
+++ b/blueprints/acceptance-test/meta/default
@@ -1,0 +1,7 @@
+  beforeEach: function() {
+    this.application = startApp();
+  },
+
+  afterEach: function() {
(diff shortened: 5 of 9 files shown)


blueprints/app/files/tests/helpers/start-app.js
```diff
@@ -16,3 +17,16 @@ export default function startApp(attrs) {
 
   return application;
 }
+
+export function getAppInstance(application) {
+  var instance;
+
+  Ember.run(function() {
```
stefanpenner: this can be re-written as

```
return Ember.run(application, 'buildInstance')
```

stefanpenner:
I like the solution, unfortunately this functionality is only available in newer versions of ember. We plan to move the ember related blueprints into ember itself, which should allow the blueprints to drift along with the capability of the framework version of choice.

stefanpenner:
Also the tests are unhappy

pete-the-pete:
I'll update and fix the tests.  I'd be happy to help move this to wherever it needs to live, in a simple test the results were promising.  I ran the follow old vs new tests (tl;dr new method was faster and less memory intensive):

### Old: Total Time: ~67s
```
javascript
module('Acceptance | old', {
  beforeEach: function() {
    this.application = startApp();
  },

  afterEach: function() {
    Ember.run(this.application, 'destroy');
  }
});

function runInSeries() {
  test('visiting /', function(assert) {
    visit('/');

    andThen(function() {
      assert.equal(currentURL(), '/');
    });
  });
}

var i = 0;
while(i < 10) {
  runInSeries();
  i++;
}
```

results:
old

results with heap (hard to see, but memory usage is consistently >1MB):
old - with heap

### New: Total Time ~42s
```
javascript
module('Acceptance | new', {
  setupOnce: function() {
    this.application = startApp();
  },

  beforeEach: function() {
    this.appInstance = getAppInstance(this.application);
  },

  afterEach: function() {
    Ember.run(this.appInstance, 'destroy');
  },

  teardownOnce: function() {
    Ember.run(this.application, 'destroy');
  }
});

function runInSeries() {
  test('visiting /', function(assert) {
    visit('/');

    andThen(function() {
      assert.equal(currentURL(), '/');
    });
  });
}

var i = 0;
while(i < 1000) {
  runInSeries();
  i++;
}
```

results:
new

results with heap (memory usage isn't visible):
new - with heap

stefanpenner:
I suspect: https://github.com/emberjs/ember.js/issues/11748 will also make some nice improvements.

Unfortunately, the work to extract the blueprint to ember has not yet begun, maybe you are interested in doing this. Until then i don't believe we can provide it as a default (for reason about)

pete-the-pete:
What if I updated the acceptance test generator to take a flag, something like '--instance' so people could opt in to using it?  It could be be opt-in at first, and then depend on the version to become the default?

stefanpenner:
> What if I updated the acceptance test generator to take a flag, something like '--instance' so people could opt in to using it? It could be be opt-in at first, and then depend on the version to become the default?

Ah, I like this.

stefanpenner:
@pete-the-pete just realized, your approach also needs to be validated to now cause similar leaks as: https://github.com/emberjs/ember.js/issues/11748

stefanpenner:
ah seems like your code ignores application.setupForTesting(); and stuff..

pete-the-pete:
working on updating the tests, didn't realize it would update the PR right away...

pete-the-pete:
The setupForTesting call is still in the startApp function, does it need to be somewhere else?

I created the meta directory for the different acceptance test generator templates, opposed to putting them as strings in the index.js.   I also changed the signature of the default import.

I wasn't sure how (or where) to put any sort of saftey checks (version of ember, existence of qunit-once), not what your thoughts are on that.

stefanpenner:
> The setupForTesting call is still in the startApp function, does it need to be somewhere else?

setupForTesting creates helpers that are associated with an app instance, as such it appears we will need to call it once per instance created. This likely needs further investigation.

pete-the-pete:
Ya, i don't think you're trying to poke holes. I have a lot to learn, so guidance is needed.  How can I investigate further?  Do the helpers have state?  Could I call, or set, something between tests to see if it leaks out between tests, or isn't destroyed as it should be?

stefanpenner:
@pete-the-pete i'll attempt to take some time this weekend (early next week) to do my due diligence here.

If you have some time, I would recommend reviewing https://github.com/emberjs/ember.js/issues/11748 and the solutions: https://github.com/emberjs/ember.js/issues/11748#issuecomment-123062148

A few examples in a demo app, should make it quick for either of us to validate ^^ as well.

Now as for addonification of ember, that is a bigger task. (but your suggestion to make it a commandline switch, is a good stepping stone.)

trentmwillis:
I'm picking this up from @pete-the-pete and almost immediately came across an issue with [`injectTestHelpers`](https://github.com/emberjs/ember.js/blob/5574f1947752bc0b1e75793dff9a16ed2e6da02d/packages/ember-testing/lib/test.js#L419). Mainly, it binds the test helpers to the context of the application which creates issues with regards to which container is being used. Specifically noticed this because of the [`visit`](https://github.com/emberjs/ember.js/blob/5574f1947752bc0b1e75793dff9a16ed2e6da02d/packages/ember-testing/lib/helpers.js#L59-L60) helper not properly navigating routes.

Additionally, I don't think not calling [`setupForTesting`](https://github.com/emberjs/ember.js/blob/7a8a46be489871820ca100f536dd03b174788ddb/packages/ember-testing/lib/setup_for_testing.js#L34) would cause any major issues, but to maintain parody with what currently happens, it would probably be best to call it every time we get a new instance.

All that said, because of the `injectTestHelpers` issue, I think some changes will have to occur upstream before this can move forward any more. I plan to open an issue there about allowing different context binding for injected test helpers.

stefanpenner:
@trentmwillis thanks for confirming, I was quite sure these issues where present.

trentmwillis:
Actually, is there any reason to not just use [`App.reset()`](https://github.com/emberjs/ember.js/blob/665c7335d262ccedae8482df9fc5a61da7a9371e/packages/ember-application/lib/system/application.js#L542)? It looks like it would, in theory, do what is needed for updating references to the container/instance.

I tried it and am getting errors about re-registration, but assuming those are straightened out, does this seem reasonable?

stefanpenner:
People are welcome to. But if we change the default, it will be to the app instance model. I would prefer to straighten those issues out

trentmwillis:
Could you elaborate on why this is preferable? Wondering since I just noticed `buildInstance` is currently private and `reset` seems to do things that I anticipate us needing to do for this to work thoroughly in acceptance tests (e.g., restart routing, re-run initializers, etc.)

blueprints/acceptance-test/meta/instance
```diff
@@ -0,0 +1,15 @@
...
+  setup: function() {
+    this.appInstance = getAppInstance(this.application);
+  },
+
+  teardown: function() {
+    Ember.run(this.appInstance, 'destroy');
+  },
+
+  teardownOnce: function() {
+    Ember.run(this.application, 'destroy');
```
stefanpenner: we now have different helper for this

blueprints/acceptance-test/meta/instance
```diff
@@ -0,0 +1,15 @@
+  setupOnce: function() {
+    this.application = startApp();
+  },
+
+  setup: function() {
+    this.appInstance = getAppInstance(this.application);
+  },
+
+  teardown: function() {
+    Ember.run(this.appInstance, 'destroy');
```
stefanpenner: likely should add a helper for this.

stefanpenner:
@trentmwillis any luck on the ember side of these test helpers?

trentmwillis:
@stefanpenner yeah, I believe I have a solution in place, waiting for review and merge https://github.com/emberjs/ember.js/pull/12310

stefanpenner:
@trentmwillis thanks, that looks good. Now once that lands, corresponding ember-test-helpers PR will tidy up the PR, and we can enable this.

Based on that timeline, ember.js will be by then bundling its own blueprints. So I suspect, this PR will be transitioned to that repo.

I hope to land the first ember-addon of ember.js this upcoming week.

stefanpenner:
As the visit API is reaching stability, and blueprints will be moving into ember. The relevant PR wil end up in ember, and will likely use the new visit api

PR #4510, opened 2015-07-22
Adding missing in-addon dummy destroy tests

gmurphey:


gmurphey:
Not sure what's happening with coveralls. I've tried kicking it with updated commits a few times to no avail.

PR #4534, opened 2015-07-28
add os to version output

kellyselden:
I've noticed that some of the first things asked of people reporting bugs is to give as much version info as possible. This adds some OS info to the version output so the second question doesn't have to be "can you also tell us which OS you're using".

Sample output:
```
C:\Users\kelly>ember -v
version: 1.13.1-master-dd6e931f5a
node: 0.12.5
npm: 2.7.4
os: win32 x64
```
My only concern is any side-effects of using `process.platform` and `process.arch`.

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -15,8 +15,9 @@ module.exports = Command.extend({
   run: function(options) {
     var versions = process.versions;
     versions['npm'] = require('npm').version;
+    versions['os'] = process.platform + ' ' + process.arch;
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
good idea!

stefanpenner:
appveyor appears broken unrelated to our code, something went sideways. It appears to be stuck on some much earlier commit. I'll ping there support.

kellyselden:
refreshed commit, now things are green

PR #4536, opened 2015-07-29
Remove unnecessary `"use strict";`s in "app.js"

nathanhammond:
When using ember-cli it is understood that all code in "app.js" is strict by default. If we add `"use strict";` in app-prefix.js as the very first statement then Uglify is clever enough to remove all of the other generated `"use strict";`s throughout app.js. In my app this results in ~33kb on-disk savings.

This should be the default behavior.

### DIFF
--- a/lib/broccoli/app-prefix.js
+++ b/lib/broccoli/app-prefix.js
@@ -1,3 +1,4 @@
+"use strict";
 /* jshint ignore:start */
 {{content-for 'app-prefix'}}
 /* jshint ignore:end */
--- a/tests/acceptance/smoke-test-slow.js
+++ b/tests/acceptance/smoke-test-slow.js
@@ -270,6 +270,22 @@ describe('Acceptance: smoke-test', function() {
     });
   });
 
+  it('ember new foo, build production and verify single "use strict";', function() {
+    return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'build', '--environment=production')
+      .then(function() {
+          var dirPath = path.join('.', 'dist', 'assets');
+          var dir = fs.readdirSync(dirPath);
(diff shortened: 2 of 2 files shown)


stefanpenner:
Pre or post gzip savings?

nathanhammond:
33 kilobytes are saved pre-gzip. Doing this we save 33kb of effort in compression, 355 saved bytes (gzip works), and 2604 statements that the JS parser doesn't have to deal with: `2604 * 13 bytes / 1024 = 33.05 kilobytes`.

stefanpenner:
> and 2604 statements that the JS parser doesn't have to deal with: 2604 * 13 bytes / 1024 = 33.05 kilobytes.

its worth noting the following:

* this saves us no packets (thanks to gzip)
* amounts to likely less then 0.01% of all tokens the parser encounters.
* the JS runtime will intern them to a single shared entry in memory

If this still seems like something worth while I can merge.

Please note, for those interested in this sort of optimization, the following PR (when complicated) will yield many of orders of magnitude more improvements https://github.com/emberjs/ember.js/pull/11576

chadhietala:
This would help us out as we are constrained by pre-gzip size in our CDN. However this PR needs more work as this will just add an additional 13 bytes.  Either esperanto's usestrict option needs to be turned to `false` or we need to pass through and merge the options passed in via the `EmberApp` constructor.

nathanhammond:
@chadhietala Uglify is clever enough to remove the duplicate `"use strict";`s if there is a script-level one included. This PR has been tested locally and confirmed to work. :)

chadhietala:
Gotcha, care to write a test so it doesn't regress?

stefanpenner:
> Gotcha, care to write a test so it doesn't regress?
>  Hide all checks

good call, then we can merge.

note: disregard appveyor, it has gotten itself into a stuck state. (it is actually passing and green, ineed to ping them...)

PR #4548, opened 2015-07-29
Add a breakline for 1.13.0 version in the changelog

juancarlosqr:
Add a breakline for 1.13.0 version in the changelog to properly break in the web


stefanpenner:
thanks :)

PR #4579, opened 2015-08-02
Add .watchmanconfig to .npmignore

jasonmit:
I don't believe it adds any value in npm packages and causes watchman to pick up addons in the watch-list.

Running `watchman watch-list` within your app's root folder will validate this.

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -8,6 +8,7 @@ dist/
 .ember-cli
 .travis.yml
 .npmignore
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1:

jasonmit:
Tests appear to be failing for reasons unrelated to this PR.

stefanpenner:
i think watchman ha fixed this issue, and it will be landing soon.

Updating this blueprint wont really have a quick affect, so we should just sit tight.

Issue #4346, opened 2015-06-25
Update blueprints to use new ES6 features / syntax


buschtoens:
I'm using [ember-suave](https://github.com/dockyard/ember-suave) to check all my js files for ES6 compliance (like using `let` instead of `var`). Currently all default blueprints use `var` although they probably should use `let` instead. 

```
jan@ubuntu:~$ ember new test-app
version: 0.2.7-master-a337a1f253
installing app...
Installed packages for tooling via npm.
Installed browser packages via Bower.
Successfully initialized git.
jan@ubuntu:~$ cd test-app/
jan@ubuntu:~/test-app$ ember g route test-route
installing
  create app/routes/test-route.js
  create app/templates/test-route.hbs
installing
  create tests/unit/routes/test-route-test.js
jan@ubuntu:~/test-app$ ember install ember-suave
Installed packages for tooling via npm.
Installed addon package.
jan@ubuntu:~/test-app$ ember serve
Livereload server on port 35729
Serving on http://localhost:4200/
Variable declarations should use `let` or `const` not `var` at app.js :
     4 |import config from './config/environment';
     5 |
     6 |var App;
--------^
     7 |
     8 |Ember.MODEL_FACTORY_INJECTIONS = true;
Property assignment should use enhanced object literal function. `{ propName: propName}` is not allowed. at app.js :
    11 |  modulePrefix: config.modulePrefix,
    12 |  podModulePrefix: config.podModulePrefix,
    13 |  Resolver: Resolver
----------^
    14 |});
    15 |
Variable declarations should use `let` or `const` not `var` at router.js :
     2 |import config from './config/environment';
     3 |
     4 |var Router = Ember.Router.extend({
--------^
     5 |  location: config.locationType
     6 |});
Variable declarations should use `let` or `const` not `var` at helpers/resolver.js :
     2 |import config from '../../config/environment';
     3 |
     4 |var resolver = Resolver.create();
--------^
     5 |
     6 |resolver.namespace = {
Variable declarations should use `let` or `const` not `var` at helpers/start-app.js :
     5 |
     6 |export default function startApp(attrs) {
     7 |  var application;
----------^
     8 |
     9 |  var attributes = Ember.merge({}, config.APP);
Variable declarations should use `let` or `const` not `var` at helpers/start-app.js :
     7 |  var application;
     8 |
     9 |  var attributes = Ember.merge({}, config.APP);
----------^
[...]

PR #4580, opened 2015-08-03
Update app.js - fix warning

MauriceMohlek:
Fix warning
Property assignment should use enhanced object literal function. `{ propName: propName}` is not allowed.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -10,7 +10,7 @@ Ember.MODEL_FACTORY_INJECTIONS = true;
 App = Ember.Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
-  Resolver: Resolver
+  Resolver
 });
 
 loadInitializers(App, config.modulePrefix);


stefanpenner:
babel is still optional, so we cannot include syntax babel requires by default.

rwjblue:
> babel is still optional

Sure, it is optional but defaulted and included automatically when you create a new project.   If you remove babel from your newly generated project's `package.json` it stands to reason that you would have tweaks to make.  If that is not true, then why include babel by default at all?

stefanpenner:
> If that is not true, then why include babel by default at all?

We will soon enable it by default, after the next pass of ember-cli + broccoli perf work.

rwjblue:
> We will soon enable it by default

It (ES6 syntax transpilation) is enabled by default. Has been since https://github.com/ember-cli/ember-cli/pull/2452 which was first released in 0.2.0 (around 6 months ago).  

> the next pass of ember-cli + broccoli perf work

I'm absolutely in LOVE with the performance work that is being done, but it is unrelated to taking advantage of ES6 features in newly generated apps that are already using babel.  Citing performance reasons as why to not enable Babel would make tons of sense, but saying that using ES6 features when babel is already being used is a performance problem does not make sense to me.

---

We should either use ES6 features in our blueprints, or remove babel from the default app blueprint.

stefanpenner:
I disagree with this binary view

buschtoens:
#4346 for cross-reference.

PR #4582, opened 2015-08-03
blueprints/app/package.json: Sort scripts alphabetically

Turbo87:
Using `npm install --save <something>` will reorder the scripts automatically which leads to an unnecessary diff. (I'm using npm v3)

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -8,8 +8,8 @@
     "test": "tests"
   },
   "scripts": {
-    "start": "ember server",
     "build": "ember build",
+    "start": "ember server",
     "test": "ember test"
   },
   "repository": "",


stefanpenner:
appveyor is actually green, some issues on there end keep making it appear like a failure...

Turbo87:
````
  1) Acceptance: smoke-test ember new foo, build production and verify fingerprint:
     AssertionError: vendor-780f896bd3ef155db48fa8f9702e3d13.js contains the fingerprint (f441758e9661b5e38cbdc8a4fd7c7bff): expected 'vendor-780f896bd3ef155db48fa8f9702e3d13.js' to include 'f441758e9661b5e38cbdc8a4fd7c7bff'
      at tests/acceptance/smoke-test-slow.js:104:31
      at Array.forEach (native)
      at tests/acceptance/smoke-test-slow.js:91:13
````

doesn't look like that's related to my change, right?

Turbo87:
@stefanpenner I've rebased the commit on the current `master` branch. The io.js TravisCI job still seems to fail, but that should be unrelated to this change.

Issue #3329, opened 2015-02-20
Remove Support for Node 0.10


rwjblue:
A warning was added in https://github.com/ember-cli/ember-cli/pull/3326, and will live through the 0.2.x series.  

This issue is meant to keep track of removing support for 0.10 in 0.3.0.

- [x]  ensure common c-extension all work

stefanpenner:
as large swaths of my employer's infrastructure is stuck on 0.10, we will likely not make this change for some time... closing.

ErisDS:
I'm a bit confused as to what is happening with 0.10 support and wanted to come by to check. This issue says it's not being dropped, yet, but the error message says:

> DEPRECATION: Node v0.10.38 is no longer supported by Ember CLI. Please update to a more recent version of Node

This makes it sounds like 0.10 support is gone completely already?

Ghost is following Node LTS, so if Ember-cli or Ember drop support for 0.10 before LTS ends in October, we're going to end up in a really difficult situation where our software cannot be built on a platform that we officially support, or we're going to have to drop behind in terms of our ember/ember-cli versions cc @kevinansfield.

stefanpenner:
We can change the warning to match nodes lts stance on 0.10

ErisDS:
@stefanpenner that'd be great, I was more interested in what the official ember-cli plan is :wink:

rwjblue:
@ErisDS - TBH, I don't think we have an official plan on this just yet.  I agree that we should decide on our position and make it clear on the website/deprecations/etc.

stefanpenner:
So the reality is, we really don't want to support the `npm` that is shipped with node 0.10, as that version of NPM (1.x) is basically foo-bar'd, and using it is basically rolling the dice each and every time. This is likely incompatible with your `LTS` plan, but you can tell node 0.10.x users to upgrade to a reasonable `npm` (maybe something in 2x) and they will have a better time.

ErisDS:
There is a special release of npm with the latest node 0.10, with a banner warning, and the next version (0.10.42) should ship with npm 2. I think it's sensible to get people to upgrade to npm 2, I'm all too aware of how much pot luck there is with in stalls on npm 1 :) 

Node 0.10 usage generally falls into 2 categories now I think - 1) people who have ubuntu LTS and get 0.10.25, and 2) new users who installed the latest version and so at least have something in the high 30's. For security reasons, it's also generally a good idea to encourage users to at least be running the most recent version of 0.10.

[...]

PR #4619, opened 2015-08-07
CI: Remove useless npm upgrade

johanneswuerbach:
npm was upgraded before the node version was switched, making the upgrade effectively useless

As tests are currently passing and node 0.12 ships npm 2.11.3 and iojs ships 2.13.3 I think this can be dropped instead of fixed.

The only old npm version (1.4.28) is shipped with 0.10, which will be removed in 0.3 https://github.com/ember-cli/ember-cli/issues/3329

https://travis-ci.org/ember-cli/ember-cli/jobs/74583522#L112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ env:
   - NODE_VERSION=iojs
 
 before_install:
-  - "npm config set spin false"
-  - "npm install -g npm@^2"
-  - "npm config set spin false"
+  - npm config set spin false
 
 os:
   - osx


johanneswuerbach:
I think this can also be dropped from appveyor (there it is working), but Windows CI should work to be sure.

stefanpenner:
@johanneswuerbach we still support `NODE_VERSION=0.10` so we will need to upgrade in that mode.

stefanpenner:
> but Windows CI should work to be sure.

yup, we can remove it from windows.

johanneswuerbach:
Node 0.10 is actually passing with npm 1.4.28 (the upgrade broke some months ago https://github.com/ember-cli/ember-cli/commit/1e53033dfd21dc7930169ffa4838a6ee4dada9c9) 

Should we bother now to spend the additional CI time to upgrade npm, when the dropping 0.10 is already on the horizon? 

Passing run using the wrong version:
https://travis-ci.org/ember-cli/ember-cli/jobs/74583522

stefanpenner:
> Should we bother now to spend the additional CI time to upgrade npm, when the dropping 0.10 is already on the horizon?

likely not

> Node 0.10 is actually passing with npm 1.4.28 (the upgrade broke some months ago 1e53033) for a while now.

good catch.

johanneswuerbach:
Passed, I'll remove it from appveyor in a separate commit when it is back? Fine with you?

stefanpenner:
> Passed, I'll remove it from appveyor in a separate commit when it is back. Fine with you?
>  Hide all checks

c

PR #4634, opened 2015-08-09
fix typo in configuring output paths for index.html file

jbailey4:
When configuring output paths, the docs currently give instructions to edit the `app.outputPaths.app.index` property. This does not actually change the output path, I believe the correct property is `app.outputPaths.app.html`.


stefanpenner:
good catch, [looks correct](https://github.com/ember-cli/ember-cli/blob/4458852a38d6c54704e5ed6f86eec289af0cc5be/lib/broccoli/ember-app.js#L502)

jbailey4:
Awesome, no problem.

PR #4640, opened 2015-08-10
Try on ci

stefanpenner:
dont merge, this is for debugging.


kellyselden:
I think this will fail instantly with a .only in it

stefanpenner:
@kellyselden so in isolation it works: https://travis-ci.org/ember-cli/ember-cli/jobs/74855440

PR #4643, opened 2015-08-10
Support a (now deprecated) single-argument use of addBowerPackageToProject

mike-north:
A single-argument use of `addBowerPackageToProject` used to work, until #4430 . This is causing some problems ( https://github.com/Gaurav0/ember-cli-jquery-ui/issues/14 )

For example
```js
addBowerPackageToProject('jquery-ui#1.11.1');
```

This PR adds support for this use back in to ember-cli, and deprecates it since we eventually want to get everyone using the two-argument variant
```js
addBowerPackageToProject('jquery-ui', '1.11.1');
```

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -989,7 +989,23 @@ Blueprint.prototype.removePackagesFromProject = function(packages) {
   @return {Promise}
 */
 Blueprint.prototype.addBowerPackageToProject = function(localPackageName, target, installOptions) {
-  var packageObject = bowEpParser.json2decomposed(localPackageName, target);
+  var lpn = localPackageName;
+  var tar = target;
+  if (localPackageName.indexOf('#') >= 0) {
+    if (arguments.length === 1) {
+      var parts = localPackageName.split('#');
+      lpn = parts[0];
+      tar = parts[1];
+      deprecateUI(this.ui)('passing ' + localPackageName +
+        ' directly to `addBowerPackageToProject` will soon be unsupported. \n' + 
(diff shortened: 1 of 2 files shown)


stavarotti:
Any reason why the following now raises a deprecation notice?

```js
this.addBowerPackageToProject('typography', 'https://github.xyz.com/UX/typography.git#1.0.1', {
  saveDev: true
})
```

Deprecation notice:

```bash
DEPRECATION: passing https://github.xyz.com/UX/typography.git#1.0.1 directly to `addBowerPackageToProject` will soon be unsupported.
You may want to replace this with `addBowerPackageToProject('https://github.xyz.com/UX/typography.git', '1.0.1')`
```

We use private repos that are not published to npm and hence the need to specify both the source and target.  If we change to the recommended way, per the deprecation notice, we get an installation error:

```bash
Installing browser packages via Bower...
  not-cached https://github.xyz.com/UX/typography.git=https://github.xyz.com/UX/typography.git#1.0.1
Failed to execute "git ls-remote --tags --heads https://github.xyz.com/UX/typography.git=https://github.xyz.com/UX/typography.git", exit code of #128
fatal: https://github.xyz.com/UX/typography.git=https://github.xyz.com/UX/typography.git/info/refs not valid: is this a git repository?
 
Error: Failed to execute "git ls-remote --tags --heads https://github.xyz.com/UX/typography.git=https://github.xyz.com/UX/typography.git", exit code of #128
fatal: https://github.xyz.com/UX/typography.git=https://github.xyz.com/UX/typography.git/info/refs not valid: is this a git repository?
```

PR #4648, opened 2015-08-10
add 1.13.8 diff to changelog

kellyselden:


rwjblue:
@kellyselden - Thanks for doing this, I use these diffs quite often they are very helpful.

kellyselden:
@rwjblue You're very welcome.

PR #4656, opened 2015-08-11
Replace ember-cli-ic-ajax in favour of ember-ajax

taras:
In this PR, I'm proposing that we replace [`ember-cli-ic-ajax`](https://github.com/rwjblue/ember-cli-ic-ajax) with [`ember-ajax`](https://github.com/embersherpa/ember-ajax) as the default mechanism for making ajax requests. 

**Notable difference**
- ajax code pulled from Ember Data
- doesn't wrap in Ember.run (with Ember 1.13+ in mind)
- uses a service which can be extended and customized
- uses Ember Data like error handing
- allows `ember-cli-mirage` to provide fixture functionality
- provides backwards compatibility for `ic-ajax` but `request` and `raw` are deprecated and will be removed in the future

The addon doesn't automatically provide an `app/services/ajax.js` service because it's designed to be opt-in. I added it to the blueprint, so the new apps opt-in to use it right away. 

I want to start a discussion.
Please, share your thoughts.

/cc @stefanpenner @rwjblue @mixonic @bcardarella @bmac

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -27,7 +27,6 @@
     "ember-cli-dependency-checker": "^1.0.1",
     "ember-cli-htmlbars": "^1.0.0",
     "ember-cli-htmlbars-inline-precompile": "^0.2.0",
-    "ember-cli-ic-ajax": "0.2.1",
     "ember-cli-inject-live-reload": "^1.3.1",
     "ember-cli-qunit": "^1.0.1",
     "ember-cli-release": "0.2.3",
@@ -36,6 +35,8 @@
     "ember-data": "2.0.0",
     "ember-disable-proxy-controllers": "^1.0.0",
     "ember-export-application-global": "^1.0.4",
-    "ember-resolver": "^2.0.2"
+    "ember-resolver": "^2.0.2",
+    "ember-ajax": "0.6.2",
+    "ember-export-application-global": "^1.0.3"
   }
 }


kimroen:
Just wanted to link this in here, as it seems relevant: https://github.com/ember-cli/rfcs/issues/19

jasonmit:
> The addon doesn't automatically provide an app/services/ajax.js service because it's designed to be opt-in. I added it to the blueprint, so the new apps opt-in to use it right away.

Since, currently, there isn't any sort of tree shaking going on it will go unused but still be brought into the dist.  I think it would be worth it to include it in `app/services/` from the start w/o a blueprint.  Less prone to user error.

taras:
I didn't add the it by default because (at least from my experimentation) it doesn't look like overwriting services works as it does with templates. When I added the service to /app/services/ajax.js in the addon, I want able to override that service by creating a file in my app directory. This negated the ability to customize the Ajax service. I would like to be proven wrong, cause it would be better in addon's app directory.

jasonmit:
> When I added the service to /app/services/ajax.js in the addon, I want able to override that service by creating a file in my app directory.

Did you mean you implemented the service solely in `app/services/ajax.js` instead of `addon/services/ajax.js`?  If so, I would keep it like you currently have it but just re-export it from `app/services/ajax`.  This way people for people wanting to customize they would just:

```js
// app/services/ajax
import AjaxService from 'ember-ajax/services/ajax';
export default AjaxService.extend(...);
```

This will noop your addon's `app/services/ajax` since the app will win out on the merge with the addon's app folder.

blueprints/app/files/app/services/ajax.js
```diff
@@ -0,0 +1,3 @@
+import AjaxService from 'ember-ajax/service';
```
stefanpenner: `export { default } from 'ember-ajax/service'`

stefanpenner: actually this whole file should be in the addon.

taras:
I just pushed a commit that does the following:

1. Upgrades to ember-ajax 0.6.1, which includes the following changes:
    1. moved service to `addons/services/ajax.js`
    2. added `app/services/ajax.js` which imports with `import {default} from 'ember-ajax/services/ajax'`
    3. updated documentation to reflect that ajax service is automatically added
2. removed `app/services/ajax.js` from the blueprint

stefanpenner:
@taras I like where this is going, and would like to get it in asap.

I believe we should likely offer some sort of deprecation help in stable-1-13, and make this part of the default ember-cli 2.0 stack.

taras:
@stefanpenner please, let me know what I can do help it move forward.

stefanpenner:
> @stefanpenner please, let me know what I can do help it move forward.
>  Show all checks

I think if you have time  to submit patches to ember-cli-ic-ajax that gives the appropriate deprecation warnings. I can review, we can merge into 1.13.x ember-cli, release. ANd cut over 2.0 to ember-ajax.

taras:
Ok, will do latest tomorrow morning.

taras:
@stefanpenner 
- [ x ] pushed `ember-ajax` to `ember-cli/ember-ajax` and deleted the old repo
- [ x ] updated the `ember-cli-ic-ajax` to point to `ember-cli/ember-ajax`
- [ x ] updated npm module to point to `ember-cli/ember-ajax`

stefanpenner:
@taras very nice!

Just pending @rwjblue on -> https://github.com/rwjblue/ember-cli-ic-ajax/pull/15#issuecomment-131875506

locks:
This needs to be documented.

stefanpenner:
@locks you'll need to provide actionable directions, where does this need to be document etc.

taras:
@locks what do you need me to do to allow this to move forward?

locks:
Sorry, this fell through the cracks:

- [ ] Ember CLI release notes
- [ ] Ember CLI website and User Guide mentions
- [ ] Ember Guides mentions

stefanpenner:
> Ember Guides mentions
>  Show all checks

@locks can you provide context as to where this goes? "ember guides mentions" isn't very actionable for people not familiar

locks:
@stefanpenner I could've sworn there were `import ajax from 'ic-ajax';` or similar in the guides, but they seem to have been PRed away. If there's any similar reference in the CLI website it should be updated.

stefanpenner:
@taras lets get a rebase, and a PR to the website removing or updating ic-ajax references.

Then we can ship this.

taras:
Ok, will do.

taras:
There are no mention of ic-ajax in [emberjs/website](emberjs/website) or [emberjs/guides](emberjs/guides) code. It was mentioned in a few issues but never added to source code.

I ticked off the last checkbox here https://github.com/ember-cli/ember-cli/pull/4656#issuecomment-134703488

taras:
Upgraded to ember-ajax 0.6.2, rebased and squashed.

PR #4663, opened 2015-08-12
[INTERNAL] Refactored component-test blueprint to use template conditional instead of string building

trabus:
Previously we were building strings for content in the component-test blueprint. This PR fixes that by using evaluation tokens to conditionally select content:
```
<% if (testType === 'integration') { %>
integration content
<% } else if (testType === 'unit') { %>
unit content
<% } %>
```

### DIFF
--- a/blueprints/component-test/files/tests/__testType__/__path__/__test__.js
+++ b/blueprints/component-test/files/tests/__testType__/__path__/__test__.js
@@ -1,9 +1,34 @@
-import { moduleForComponent, test } from 'ember-qunit';<%= testImports %>
+import { moduleForComponent, test } from 'ember-qunit';<% if (testType === 'integration') { %>
+import hbs from 'htmlbars-inline-precompile';<% } %>
 
 moduleForComponent('<%= componentPathName %>', '<%= friendlyTestDescription %>', {
-  <%= testOptions %>
+  <% if (testType === 'integration' ) { %>integration: true<% } else if(testType === 'unit') { %>// Specify the other units that are required for this test
+  // needs: ['component:foo', 'helper:bar'],
+  unit: true<% } %>
 });
 
 test('it renders', function(assert) {
-  <%= testContent %>
+  <% if (testType === 'integration' ) { %>assert.expect(2);
+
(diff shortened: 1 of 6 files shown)


trabus:
cc: @blimmer you may want to give ember-cli-mocha the same treatment. I only recently discovered that we're using lodash.template under the hood, which allows for evaluation tokens, thus allowing a better way to handle inserting the content. It's a little ugly due to formatting necessities, but it's easier to deal with that the string building.

blimmer:
Yeah, but it's **ugly** because (at least when I looked) it leaves a ton of whitespace.

trabus:
> Yeah, but it's ugly because (at least when I looked) it leaves a ton of whitespace.

Though the current formatting on this PR matches the string formatting exactly, I was more talking about how the conditional had to be on one line for the formatting to work properly.

trabus:
Not sure why coverage dropped, I actually have a couple assertions I wanted to add, maybe that will fix it.

stefanpenner:
> Not sure why coverage dropped, I actually have a couple assertions I wanted to add, maybe that will fix it.
>  Hide all checks

rounding error likely

trabus:
I wanted to make sure the `{{#<%= dasherizedModuleName %>}}` was properly being interpolated. I missed it on the first pass and it was outputting `{{#" + dasherizedModuleName + "}}` in the generated blueprint.

trabus:
Welp, looks like I found some issues with the `--path` option, glad I added those assertions!

trabus:
I'm utterly confused as to how the coverage has decreased. I've reduced the actual amount of code, and added additional assertions, yet the coverage has decreased even more (from -0.008% to -0.02%).

trabus:
I guess adding to the array of text to assert is what is screwing it up. If there were additional assertions by themselves, it would probably be fine.

kellyselden:
My guess is it's not covering the `|| "integration"`, but the coveralls coverage UI is hard for me to interpret.

blimmer:
Thanks again for the heads up @trabus . The corresponding PR is open on ember-cli-mocha.

PR #4666, opened 2015-08-13
Update Ember to v2.0.0.

rwjblue:


rwjblue:
Ember Data 1.13.8 fails to load under Ember 2.0.0, I think this may need to either update to ED 2.0.0-beta.1 or wait until ED 2.0.0 is out. 

@stefanpenner - Thoughts?

stefanpenner:
I don't actually see a good reason why 1.13.x of Ed doesn't work with ember 2 other then it depending on various array related polyfils. That should be fixable in short time

I'll take a look this morning

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/4675

Issue #4665, opened 2015-08-13
Generating a resource with no entity name throws an error.


kwight:
Generating a resource with no entity name throws an error and stack trace:

```
Arthur:hummingbird kwight$ ember generate resource
version: 1.13.8-master-1ca9464dac
Cannot read property 'toLowerCase' of undefined
TypeError: Cannot read property 'toLowerCase' of undefined
    at Object.inflector._apply_rules (/Users/kwight/Sites/ember-cli/node_modules/inflection/lib/inflection.js:579:52)
    at Object.inflector.singularize (/Users/kwight/Sites/ember-cli/node_modules/inflection/lib/inflection.js:677:24)
    at Class.module.exports._process (/Users/kwight/Sites/ember-cli/blueprints/resource/index.js:55:26)
    at Class.module.exports.install (/Users/kwight/Sites/ember-cli/blueprints/resource/index.js:14:17)
    at Class.module.exports.Task.extend.run (/Users/kwight/Sites/ember-cli/lib/tasks/generate-from-blueprint.js:60:49)
    at Class.module.exports.Command.extend.run (/Users/kwight/Sites/ember-cli/lib/commands/generate.js:80:17)
    at Class.<anonymous> (/Users/kwight/Sites/ember-cli/lib/models/command.js:146:17)
    at lib$rsvp$$internal$$tryCatch (/Users/kwight/Sites/ember-cli/node_modules/rsvp/dist/rsvp.js:493:16)
    at lib$rsvp$$internal$$invokeCallback (/Users/kwight/Sites/ember-cli/node_modules/rsvp/dist/rsvp.js:505:17)
    at lib$rsvp$$internal$$publish (/Users/kwight/Sites/ember-cli/node_modules/rsvp/dist/rsvp.js:476:11)
```

instead of the helpful descriptive error in `lib/models/blueprint.js` for, say, a route:

```
Arthur:hummingbird kwight$ ember generate route
version: 1.13.8-master-1ca9464dac
installing route
The `ember generate` command requires an entity name to be specified. For more details, use `ember help`.
```

kellyselden:
Fixed in #4667

kellyselden:
Thanks for the report. Nice find.

PR #4667, opened 2015-08-13
[BUGFIX] fixes resource generator when invoked without entity name

lazybensch:
addresses #4665 

Currently, when running `ember generate resource` without specifying an entity name, the user does not get the nice `SilentError` message but a stack trace because at some point it tries to inflect `undefined`.

This PR fixes that bug and also adds a regression test

### DIFF
--- a/blueprints/resource/index.js
+++ b/blueprints/resource/index.js
@@ -48,9 +48,11 @@ module.exports = {
   },
 
   _process: function(type, options) {
+    var entityName = options.entity.name;
+
     var modelOptions = merge({}, options, {
       entity: {
-        name: inflection.singularize(options.entity.name)
+        name: entityName ? inflection.singularize(entityName) : ''
       }
     });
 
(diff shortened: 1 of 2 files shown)


lazybensch:
fixing specs..

lazybensch:
although fixing the bug was quite easy i had some trouble writing a failing test for this issue.

the `cli` objects `logError` swallows exceptions ([here](https://github.com/ember-cli/ember-cli/blob/058528977255d0e366b8a1760668875cfadf9067/lib/cli/cli.js#L112)) so first i refactored that part to rethrow every exception after logging but it turned out quite a few tests relied on that swallowing behaviour and failed afterwards. I was unsure wether this is a bug or on purpose so i decided to not touch it for now.

Afterwards I tried to get a hold of the `ui` object that is created by `initApp()` but for some reason that is not the object `writeError` ultimately gets called on. In the end i *kind of* gave up and now i am highjacking the prototype method for that test to validate that the correct error gets displayed.

If you have any clue on how I could improve the test, i would be happy to refactor it.

stefanpenner:
Sounds like a bug. I found can open a related issue that would be great. That way we can fix any of the false positives

lazybensch:
I opened another PR #4678 to fix the issue with the swallowed exceptions

Issue #4631, opened 2015-08-09
Application named app


daxpunk:
Hi,

I've seen that if the application is named 'app' we have an error with the style when we want to 'ember serve'.

Dax

PR #4672, opened 2015-08-13
[BUGFIX] disallow `app` as project name

lazybensch:
addresses #4631

In march, PR #3642 prevented the style sheet name from having the same name as the project (happens to be the case for `ember new app`. So currently you __can__ invoke `ember new app` but get into trouble as soon as you try `ember serve`.

This PR fixes the issue by disallowing `app` as a project name altogether and adds a regression test.

### DIFF
--- a/lib/utilities/valid-project-name.js
+++ b/lib/utilities/valid-project-name.js
@@ -3,7 +3,7 @@
 module.exports = function(name) {
   name = name.toLowerCase();
 
-  if (['test', 'ember', 'ember-cli', 'vendor'].indexOf(name) > -1) { return false; }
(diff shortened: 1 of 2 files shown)


lib/utilities/valid-project-name.js
```diff
@@ -5,6 +5,7 @@ module.exports = function(name) {
 
   if (['test', 'ember', 'ember-cli', 'vendor'].indexOf(name) > -1) { return false; }
   if (name.indexOf('.') > -1) { return false; }
+  if (name === 'app' ) { return false; }
```
stefanpenner: should this maybe be part of line 6 ?

lazybensch: thx, i misread that line. it should. i fixed my commit

stefanpenner:
fantastic , thank you kindly.

PR #4676, opened 2015-08-13
s/which/that

raganwald:
Formally, "which": is used for a parenthetical remark, and "that" is used for the object of the statement e.g.

> Additional components of this project—which are used at runtime in your application—that are included as dependencies in ember-cli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Supports node 0.12.x and npm 2.6.x.
 
 
 ## Project Elements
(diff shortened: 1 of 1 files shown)


stefanpenner:
@raganwald thanks :)

raganwald:
That little pencil octicon in GitHub makes this kind of drive-by contribution so easy :-)

stefanpenner:
@raganwald thats great to hear. The trap was designed as such ;)

PR #4677, opened 2015-08-13
2.0 channel related release changes

stefanpenner:
As we plan to follow similar release plans as ember, we will need to adjust our release.md to incorporate channel related changes.

I believe in practice, we will not publish canary builds to `npm`, users will be better served subscribing to master directly. I believe this because:

* git subscription works fine in NPM, as ember-cli has no `built assets`
* publishing thousands of tarballs to NPM seems wasteful
* publishing betas will give us the needed coverage.

Alternative:
* We could publish nightly canaries – although the value is unknown. We should likely just revisit if we want it later.

#### Schedule:

`ember` and soon `ember/data` release weekly betas, it makes sense for ember-cli to follow this, and publish beta's immediately after `ember` and `ember/data` have.

#### Other important stuff:

- [ ] confirm plan \w others

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -35,17 +35,27 @@ and failures we do not want our users to endure any extra pain.
 
 1. go back to ember-cli directory
 * `git add` the modified `package.json` and `CHANGELOG.md`
-* Commit the changes `git commit -m "Release vx.y.z"` and push `git push`
-* `git tag "vx.y.z"`
-* `git push origin <vx.y.z>`
-* publish to npm
-* `npm publish ./ember-cli-<version>.tgz`
+* if publishing a stable release
+  * Commit the changes `git commit -m "Release vx.y.z"` and push `git push`
+  * `git tag "vx.y.z"`
+  * `git push origin <vx.y.z>`
+  * publish to npm
+  * `npm publish ./ember-cli-<version>.tgz`
+* if publishing a unstable channel release
+  * Commit the changes `git commit -m "Release vx.y.z-channel.n"` and push `git push`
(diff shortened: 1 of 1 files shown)


stefanpenner:
somewhere i should also document what commit messages should look like, maybe contribution.md ?

trabus:
LGTM :+1:

RELEASE.md
```diff
@@ -35,17 +35,27 @@ and failures we do not want our users to endure any extra pain.
...
+  * `git tag "vx.y.z"`
+  * `git push origin <vx.y.z>`
+  * publish to npm
+  * `npm publish ./ember-cli-<version>.tgz`
+* if publishing a unstable channel release
+  * Commit the changes `git commit -m "Release vx.y.z-channel.n"` and push `git push`
+  * `git tag "vx.y.z-channel.n"`
+  * `git push origin <vx.y.z-channel.n>`
+  * publish to npm
+  * `npm publish ./ember-cli-<version>.tgz --tag x.y.z-channel.n`
```
rwjblue: I thought the tag would be '--tag channelName' (not including the version number).

stefanpenner: we may need to do both, as we want to publish

* 1.13.8-beta.1 – so users can lock to a specific beta
* beta – to reference the current beta

stefanpenner: actually, since npm does not have a shrinkwrap file. I am somewhat concerned to release aliases like `beta`, although i think we should, it may be a :footprints: :gun:

stefanpenner:
going to be taking a different approach

PR #4678, opened 2015-08-14
[BUGFIX] rethrow exceptions after logging

lazybensch:
Currently the `logError` message of the internal `Cli` object is swallowing all exceptions. While I have not discovered any user facing bugs due to this yet, it was a pain to write tests, checking for `SilentError`s and I also found a few tests that where falsely relying on that method not throwing.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -163,6 +163,7 @@ CLI.prototype.logError = function(error) {
     if (error.stack) {
       console.error(error.stack);
     }
+    throw error;
   }
   this.ui.errorLog.push(error);
   this.ui.writeError(error);
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -134,7 +134,7 @@ Command.prototype.validateAndRun = function(args) {
 
   if (this.works === 'outsideProject' && this.isWithinProject) {
     return Promise.reject(new SilentError(
-      'You cannot use the ' +  chalk.green(this.name) + ' command inside an ember-cli project.'
+      'You cannot use the ' + chalk.green(this.name) + ' command inside an ember-cli project.'
     ));
   }
 
@@ -232,7 +232,7 @@ Command.prototype.assignOption = function(option, parsedOptions, commandOptions)
     }
   } else {
     this.ui.writeLine('The specified command ' + chalk.green(this.name) +
-                     ' requires the option ' + chalk.green(option.name) + '.');
+                      ' requires the option ' + chalk.green(option.name) + '.');
   }
   return isValid;
 };
--- a/tests/acceptance/generate-test.js
+++ b/tests/acceptance/generate-test.js
@@ -520,11 +520,12 @@ describe('Acceptance: ember generate', function() {
     });
   });
 
-  it('resource without entity name does not throw exception', function() {
+  it('resource without entity name does throw a warning', function() {
     return generate(['resource']).then(function() {
       expect(false).to.be.ok;
     }, function(err) {
-      expect(err.errorLog[0].message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
     });
   });
 
@@ -755,7 +756,8 @@ describe('Acceptance: ember generate', function() {
     return generate(['adapter', 'application', '--base-class=application']).then(function() {
       expect(false).to.be.ok;
     }, function(err) {
-      expect(err.errorLog[0]).to.match(/Adapters cannot extend from themself/);
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.match(/Adapters cannot extend from themself/);
     });
   });
 
@@ -763,7 +765,8 @@ describe('Acceptance: ember generate', function() {
     return generate(['adapter', 'foo', '--base-class=foo']).then(function() {
       expect(false).to.be.ok;
     }, function(err) {
-      expect(err.errorLog[0]).to.match(/Adapters cannot extend from themself/);
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.match(/Adapters cannot extend from themself/);
     });
   });
 
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -14,6 +14,7 @@ var util       = require('util');
 var conf       = require('ember-cli-internal-test-helpers/lib/helpers/conf');
 var EOL        = require('os').EOL;
 var assertFile = require('ember-cli-internal-test-helpers/lib/helpers/assert-file');
+var chalk      = require('chalk');
 
 describe('Acceptance: ember new', function() {
   this.timeout(10000);
@@ -71,8 +72,9 @@ describe('Acceptance: ember new', function() {
       ''
     ]).then(function() {
       throw new Error('this promise should be rejected');
-    }, function(result) {
-      expect(result.errorLog[0].message).to.contain('The `ember new` command requires a name to be specified.');
+    }, function(err) {
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.contain('The `ember new` command requires a name to be specified.');
     });
   });
 
@@ -81,8 +83,9 @@ describe('Acceptance: ember new', function() {
       'new'
     ]).then(function() {
       throw new Error('this promise should be rejected');
-    }, function(result) {
-      expect(result.errorLog[0].message).to.contain('The `ember new` command requires a name to be specified.');
+    }, function(err) {
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.contain('The `ember new` command requires a name to be specified.');
     });
   });
 
@@ -117,9 +120,10 @@ describe('Acceptance: ember new', function() {
         '--skip-git'
       ]).then(function() {
         throw new Error('this promise should be rejected');
-      }, function(result) {
-        expect(result.errorLog[0].message).to.match(/You cannot use the .*new.* command inside an ember-cli project./);
+      }).catch(function(error) {
         expect(existsSync('foo')).to.be.false;
+        expect(error.name).to.equal('SilentError');
+        expect(error.message).to.equal('You cannot use the ' + chalk.green('new') + ' command inside an ember-cli project.');
       });
     }).then(confirmBlueprinted);
   });
@@ -167,7 +171,7 @@ describe('Acceptance: ember new', function() {
   });
 
 
-  it('ember new with git blueprint uses checks out the blueprint and uses it', function() {
+  it('ember new with git blueprint checks out the blueprint and uses it', function() {
     this.timeout(20000); // relies on GH network stuff
 
     return ember([
--- a/tests/unit/cli/cli-test.js
+++ b/tests/unit/cli/cli-test.js
@@ -554,10 +554,11 @@ describe('Unit: CLI', function() {
     var help = stubValidateAndRun('help');
 
     return ember(['unknownCommand']).then(function() {
-      var errors = ui.errors.trim().split(EOL);
-      var helpfulMessage = /The specified command .*unknownCommand.* is invalid\. For available options/;
-      expect(errors[0]).to.match(helpfulMessage, 'expected an invalid command message');
+      expect(false).to.be.ok;
+    }).catch(function(error) {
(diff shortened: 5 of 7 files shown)


stefanpenner:
hmm windows appears unhappy :(

lazybensch:
what a pain to debug those smoke tests :D but tests are green now @stefanpenner .
The culprit was that prior to my changes the promise returned by `cli.run` always resolved and the return value was parsed to generate an exit code. Now that `cli.run` properly throws, the promise never resolves successfully and tests that depend on an erroneous status code of a spawned `ember` process fails.

actually i wander how travis could be green before :confused:

bin/ember
```diff
@@ -30,5 +30,7 @@ resolve('ember-cli', {
   }).then(function(result) {
     var exitCode = typeof result === 'object' ? result.exitCode : result;
     exit(exitCode);
+  }).catch(function() {
```
stefanpenner: the goal is for ember internals, to always resolve but with a status code.  This should is an invariant that is easy to maintain.

lib/cli/cli.js
```diff
@@ -153,5 +153,5 @@ CLI.prototype.logError = function(error) {
     console.error(error.stack);
   }
   this.ui.writeError(error);
-  return 1;
+  throw error;
```
stefanpenner: this was correctly resolving with the status code.

tests/acceptance/destroy-test.js
```diff
@@ -16,6 +16,7 @@ var root       = process.cwd();
 var tmp        = require('tmp-sync');
 var tmproot    = path.join(root, 'tmp');
 var EOL        = require('os').EOL;
+var chalk       = require('chalk');
```
stefanpenner: whitespace

lazybensch:
rebased the branch, is there anything still missing for a merge now?

stefanpenner:
yes the above two comments.

lazybensch:
But isn't the behaviour outside the scope of that `cli()` code still the way you described it? the process still exits in every scenario with the correct error code. The problem i fixed was that **inside the `cli()` call** you basically never knew whether an exception was thrown or not.

Now, every peace of code inside the call to `cli()` will be aware of exceptions *(so that you can properly write tests that expect them [like here](https://github.com/ember-cli/ember-cli/pull/4678/files#diff-1653bd216fa57a58dca0fdd226d398b6R497))*. For every function that is invoking the `cli()` call however, the behaviour stays the same as it was before my PR.

Or do you mean the `cli()` call should always resolve ? If so, can you explain a bit why? In this case i could easily change [this line](https://github.com/lazybensch/ember-cli/blob/log_error_should_not_swallow_exceptions/lib/cli/index.js#L101) to `return cli.run(environment).catch(function() { return 1; });`. In this case `cli()` would always resolve, always with the correct status code and tests would still be able to detect exceptions.

Hope I am not being stupid right now...

stefanpenner:
I'll look closely after breakfast. It's possible I skimmed to fast

lazybensch:
Yeah no problem, I really didn't describe it that good in the first place. So maybe I can show you why this fix is important:

It all started when I saw issue #4665. Due to an unrelated bug the generate method was throwing instead of printing an error message. When I fixed that, I wanted to write a regression test for that case, so in my test i wanted to catch the SilentError and check it's content __but__ the error never arrived at my test. [here is my passing test](https://github.com/ember-cli/ember-cli/pull/4667/files#diff-1653bd216fa57a58dca0fdd226d398b6R489) you can see how the `generate` call resolves and i had to highjack the `writeError` call to see if the correct message got delivered to the user.

That PR got merged but it left me wondering how others have dealt with this problem in the past and while looking through the code i figured they didn't. That is the reason ticket #4665 was opened in the first place, cause there was no proper test coverage for a correctly displayed error message. Even worse i found multiple tests that where relying on methods to resolve that really should not resolve.

That is because all the tests are calling `cli()` at some point and every exception inside `cli()` gets swallowed by its `logError` function. I fixed that by making it re-throw the error. But to not mess with embers return code (that is what you described, that it should always return with a proper status code) I added [this line](https://github.com/ember-cli/ember-cli/pull/4678/files#diff-b07d0362e8bdfba18a97ee0eca7014e1R34).

bin/ember
stefanpenner: @lazybensch not this ^^

lazybensch:
ok @stefanpenner. :)
Unfortunately I see no other way to address this issue then. Without moving the `return 1` up into the script I see no way to test error handling of the cli properly. It is still *somehow* possible though. Just keep in mind that without my PR instead of testing an error like this:

```js
  it('resource without entity name does not throw hard exception', function() {
    return generate(['resource']).catch(function(error) {
      expect(error.name).to.equal('SilentError');
      expect(error.message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
    });
  });
```

you need to do it like this:

```js

  it('resource without entity name does not throw exception', function() {

    var restoreWriteError = MockUI.prototype.writeError;
    MockUI.prototype.writeError = function(error) {
      expect(error.message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
    };

    return generate(['resource']).then(function() {
      MockUI.prototype.writeError = restoreWriteError;
    });

  });
```


If you have an alternative idea on how to tackle this problem, I would try to change my implementation. If not, you can probably close this PR

stefanpenner:
@lazybensch we could insert a throw here: https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L151 within the testing flag.

Although it would be different then prod, a quick unit test should be able to test the various permutations of that function.

This would allow us to have tests as you describe (they are good), but without causing grief for the executable.

lazybensch:
@stefanpenner good point. I've put the `throw` behind the testing flag and wrote a unit test for `logError`s exception handling.

CI looks good, the only test i see failing is
```
  1) Acceptance: addon-smoke-test npm pack does not include unnecessary files:
     AssertionError: expected packaged addon to not contain file or folder '.ember-cli': expected 'package/README.md\npackage/ember-cli-build.js\npackage/index.js\npackage/config/ember-try.js\npackage/config/environment.js\npackage/.watchmanconfig\npackage/LICENSE.md\npackage/.jshintrc\n' not to match /.ember-cli/
```
which seems to be broken on master.

thx for your patience :D

stefanpenner:
> CI looks good, the only test i see failing is

Ya seems some annoying bugs on master :P. Been slowly squashing.

lazybensch:
@homu rebased the branch. CI appears green.

lazybensch:
@Turbo87 fixed your comments and squashed

tests/acceptance/generate-test.js
```diff
@@ -524,7 +524,15 @@ describe('Acceptance: ember generate', function() {
...
-      expect(err.errorLog[0].message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
+    });
+  });
+
+  it('resource without entity name does not throw hard exception', function() {
+    return generate(['resource']).catch(function(err) {
+      expect(err.name).to.equal('SilentError');
+      expect(err.message).to.equal('The `ember generate <entity-name>` command requires an entity name to be specified. For more details, use `ember help`.');
```
Turbo87: are you aware that this test will also pass if `generate(...)` resolves without an error? this is essentially the reason for the weird `expect(false).to.be.ok;` above, but if you add this, the new test will look pretty much the same as the one above.

lib/models/command.js
```diff
@@ -127,14 +127,14 @@ Command.prototype.validateAndRun = function(args) {
 
   if (this.works === 'insideProject' && !this.isWithinProject) {
     return Promise.reject(new SilentError(
-      'You have to be inside an ember-cli project in order to use ' +
-      'the ' + chalk.green(this.name) + ' command.'
+      chalk.yellow('You have to be inside an ember-cli project in order to use the ') +
+      chalk.green(this.name) + chalk.yellow(' command.')
```
Turbo87: why did you modify the colors here?

tests/acceptance/pods-destroy-test.js
```diff
@@ -14,6 +14,8 @@ var remove      = Promise.denodeify(fs.remove);
 var root        = process.cwd();
 var tmproot     = path.join(root, 'tmp');
 var mkTmpDirIn  = require('../../lib/utilities/mk-tmp-dir-in');
+var EOL         = require('os').EOL;
```
Turbo87: unused?

lib/models/command.js
lazybensch: for consistency
`$ git grep -nC 1  chalk.green lib/models/command.js`
```javascript
lib/models/command.js-129-    return Promise.reject(new SilentError(
lib/models/command.js:130:      chalk.yellow('You have to be inside an ember-cli project in order to use the ') +
lib/models/command.js:131:      chalk.green(this.name) + chalk.yellow(' command.')
lib/models/command.js-132-    ));
--
lib/models/command.js-136-    return Promise.reject(new SilentError(
lib/models/command.js:137:      chalk.yellow('You cannot use the ') + chalk.green(this.name) + chalk.yellow(' command inside an ember-cli project.')
lib/models/command.js-138-    ));
--
lib/models/command.js-234-    this.ui.writeLine(
lib/models/command.js:235:      chalk.yellow('The specified command ') + chalk.green(this.name) +
lib/models/command.js:236:      chalk.yellow(' requires the option ') + chalk.green(option.name) + chalk.yellow('.')
lib/models/command.js-237-    );
```

tests/acceptance/generate-test.js
lazybensch: i was not. thx for pointing this out!

lib/models/command.js
Turbo87: IMHO that should've been a separate commit, but I don't want to make this more complicated than it already is :wink:

lazybensch: well, actually i just saw those other occurrences were also introduced by my PR (i honestly can't remember why i did this, maybe it made sense before i rebased for some reason). i will revert those changes.

lazybensch:
comments addressed, squashed, CI "green"

tests/acceptance/new-test.js
```diff
@@ -81,8 +83,19 @@ describe('Acceptance: ember new', function() {
...
+    });
+  });
+
+  it('ember new without app name doesn\'t throw hard exception', function() {
+    return ember([
+      'new',
+      ''
+    ]).catch(function(error) {
+      expect(error.name).to.equal('SilentError');
+      expect(error.message).to.equal(chalk.yellow('The `ember new` command requires a name to be specified. For more details, use `ember help`.'));
```
Turbo87: same as "ember new with empty app name fails with a warning"?

tests/acceptance/new-test.js
```diff
@@ -179,6 +193,9 @@ describe('Acceptance: ember new', function() {
       '--blueprint=https://github.com/ember-cli/app-blueprint-test.git'
     ]).then(function() {
       expect(existsSync('.ember-cli')).to.be.true;
+    }).catch(function(error) {
+      expect(error.name).to.equal('Error');
+      expect(error.message).to.equal(chalk.yellow('You cannot use the new command inside an ember-cli project.'));
```
Turbo87: is this supposed to fail?

lazybensch:
@Turbo87 another round? :relaxed: I think i now resolved everything that got messed up during my rebase =S

tests/unit/cli/cli-test.js
```diff
@@ -554,10 +554,10 @@ describe('Unit: CLI', function() {
     var help = stubValidateAndRun('help');
 
     return ember(['unknownCommand']).then(function() {
-      var errors = ui.errors.trim().split(EOL);
-      var helpfulMessage = /The specified command .*unknownCommand.* is invalid\. For available options/;
-      expect(errors[0]).to.match(helpfulMessage, 'expected an invalid command message');
-      expect(help.called, 'help command was executed').to.not.be.ok;
```
Turbo87: where did those assertions go?

lazybensch: the (old) lines 557-559 are equivalent to the (new) line 560
the (old) line 560 got lost -.- will add it to the `catch` block

stefanpenner:
:+1:

stefanpenner:
@homu r+

kellyselden:
Shoot I think I screwed up homu by manually merging a comment change.

kellyselden:
@homu retry

Issue #4499, opened 2015-07-21
Remove warning when generating dashless helpers


miguelcobain:
Currently ember-cli still displays the following message when generating a dashless helper:
```
[WARN]: helpers without a dash will not automatically be resolved
```

Since we're currently generating the new helpers which can be dashless, we probably should remove this message.

PR #4679, opened 2015-08-14
[BUGFIX] remove non-dash helper warning

lazybensch:
addresses #4499

This PR is removing the warning message for when users generate dash-less template helpers since they __do__ get resolved nowadays.

I felt like this should not get a regression test right?


rwjblue:
Thanks!

PR #4680, opened 2015-08-14
[BUGFIX release] fixes #4330 adding `--classic` flag to destroy and generate commands

### LINKED ISSUE #4330: `--pod` doing the opposite of `usePods` setting is confusing
I find the [following behavior of Ember CLI](http://www.ember-cli.com/#pods) unexpected and confusing:
> If you would like to use the pods structure as the default for your project, you can set `usePods` in your `.ember-cli` config file to `true` ... When `usePods` is `true`, the `--pod` flag is essentially inverted.

The issue is that one is unable to tell from the command line whether a generator will use pods or not, despite a flag being explicitly added. That is, typing
```bash
$ ember generate route foo --pod    # may or may not use pods
```
May create the Foo Route in pod structure, or it may not, depending on the value of `usePods`. 

I feel it would be more clear if instead `--pod` guaranteed the pod structure, and another flag (*e.g.*, `--no-pod`) would guarantee traditional structure. The absence of either flag would use the setting as defined in `.ember-cli`.
```bash
$ ember generate route foo          # uses setting in .ember-cli
$ ember generate route foo --pod    # uses pods
$ ember generate route foo --no-pod # does not use pods
```

trabus:
The behavior of the `--pod` flag when the `usePods` setting is true has proven confusing to a lot of users. This PR adds a `--classic` flag (alias `-c`) that will generate a blueprint in classic structure. It is only effective when `usePods` is true, and does not have precedence over the `--pod` flag.

When a user generates or destroys with the `--pod` flag with `usePods` true, the blueprint will now be generated or destroyed in pod structure as would be expected.

so for example with `usePods` true in your project:
```
ember g component foo-bar --classic
// generates
app/components/foo-bar.js
app/templates/components/foo-bar.hbs
```

### DIFF
--- a/lib/commands/destroy.js
+++ b/lib/commands/destroy.js
@@ -17,7 +17,10 @@ module.exports = Command.extend({
     { name: 'dry-run', type: Boolean, default: false, aliases: ['d'] },
     { name: 'verbose', type: Boolean, default: false, aliases: ['v'] },
     { name: 'pod', type: Boolean, default: false, aliases: ['p'] },
-    { name: 'in-repo-addon', type: String, default: null, aliases: ['in-repo', 'ir']}
+    { name: 'classic', type: Boolean, default: false, aliases: ['c'] },
+    { name: 'dummy', type: Boolean, default: false, aliases: ['dum','id'] },
+    { name: 'in-repo-addon', type: String, default: null, aliases: ['in-repo', 'ir'] }
+
   ],
 
   anonymousOptions: [
@@ -64,7 +67,7 @@ module.exports = Command.extend({
       args: rawArgs
     };
 
-    if(this.settings && this.settings.usePods) {
+    if(this.settings && this.settings.usePods && !commandOptions.classic) {
       commandOptions.pod = !commandOptions.pod;
     }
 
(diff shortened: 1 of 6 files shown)


stefanpenner:
does this need a docs?

trabus:
> does this need a docs?

Yes, will get those updated shortly.

stefanpenner:
> does this need a docs?
> Yes, will get those updated shortly.

thanks :)

stefanpenner:
LGTM:

commit should be `[BUGFIX release]`

trabus:
> commit should likely be [BUGFIX release] (i'll handle the cherry-picks )
@stefanpenner I'm sorry, I'm not sure I understand.
Do you mean the git commit message should literally have `[BUGFIX release]`, or should `release` be replaced with the release version, or are you talking about something else entirely?

stefanpenner:
> @stefanpenner I'm sorry, I'm not sure I understand.

no problem, I still need to flesh out the docs for this.

the commit should literally contain `[BUGFIX release]` as to go out with the next 1.13.x release, but be included in master.

trabus:
So would the same go with `[ENHANCEMENT release]` and `[INTERNAL release]`?

stefanpenner:
> So would the same go with [ENHANCEMENT release] and [INTERNAL release]?
>  Show all checks

yes

PR #4689, opened 2015-08-15
[ENHANCEMENT] On Windows, check for elevation, warn if missing

felixrieseberg:
- Added a small utility that checks if we’re running on Windows, and if
so, checks if ember cli is currently running with elevated rights
(think sudo, but windows)
- If no elevation is detected, we log a small warning, informing about
the performance impact
- Elevation check uses `NET SESSION`, which works on Windows versions
20+ years old
- Test included

### DIFF
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -2,6 +2,7 @@
 
 var path    = require('path');
 var Command = require('../models/command');
+var win     = require('../utilities/windows-admin');
 
 module.exports = Command.extend({
   name: 'build',
@@ -22,7 +23,10 @@ module.exports = Command.extend({
       analytics: this.analytics,
       project: this.project
     });
-    return buildTask.run(commandOptions);
+
+    return win.checkWindowsElevation(this.ui).then(function() {
+      return buildTask.run(commandOptions);
+    });
   },
 
   taskFor: function(options) {
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -5,7 +5,8 @@ var path        = require('path');
 var Command     = require('../models/command');
 var Promise     = require('../ext/promise');
 var SilentError = require('silent-error');
-var PortFinder = require('portfinder');
+var PortFinder  = require('portfinder');
+var win         = require('../utilities/windows-admin');
 
 PortFinder.basePort = 49152;
 
@@ -59,7 +60,9 @@ module.exports = Command.extend({
         project: this.project
       });
 
-      return serve.run(commandOptions);
+      return win.checkWindowsElevation(this.ui).then(function() {
+        return serve.run(commandOptions);
+      });
     }.bind(this));
   },
 
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -6,6 +6,7 @@ var Builder   = require('../models/builder');
 
 var fs = require('fs');
 var path = require('path');
+var win = require('../utilities/windows-admin');
 
 var defaultPortNumber = 7357;
 
@@ -126,7 +127,8 @@ module.exports = Command.extend({
       var test  = new TestTask(options);
       var build = new BuildTask(options);
 
-      return build.run({
+      return win.checkWindowsElevation(this.ui).then(function() {
+        return build.run({
           environment: commandOptions.environment,
           outputPath: outputPath
         })
@@ -134,6 +136,7 @@ module.exports = Command.extend({
           return test.run(testOptions);
         })
         .finally(this.rmTmp.bind(this));
+      }.bind(this));
     }
   }
 });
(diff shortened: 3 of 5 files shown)


felixrieseberg:
As discussed with @stefanpenner, but obviously also @kellyselden

stefanpenner:
Lvgtm

tests/runner.js
```diff
@@ -14,7 +14,7 @@ rimraf.sync('.node_modules-tmp');
 rimraf.sync('.bower_components-tmp');
 
 var root = 'tests/{unit,acceptance}';
-var _checkOnlyInTests = RSVP.denodeify(mochaOnlyDetector.checkFolder.bind(null, root + '/**/*{-test,-slow}.js'));
+var _checkOnlyInTests = RSVP.denodeify(mochaOnlyDetector.checkFolder.bind(null, root + '/**/smoke{-test,-slow}.js'));
```
kellyselden: Was this intentional? or are you still testing?

kellyselden:
@felixrieseberg My only comment is that the Windows non-admin case where it prints to console has no test. Other than that, looks great.

felixrieseberg:
Happy to write a test for that if you need one, just not sure that it'd be all to valuable - because I'd have to mock `exec` and `ui.this.writeLine`, at which point I'm not sure that we're still testing anything ;-) If you have a good idea, I'm happy to write it!

kellyselden:
@felixrieseberg Maybe I'm just jaded by writing so many `ember help` acceptance tests... ha

felixrieseberg:
FYI, the AppVeyor test passed - no idea why GitHub thinks it's still running.

felixrieseberg:
@stefanpenner @kellyselden Let me know if you guys need anything else from me here.

stefanpenner:
:)

felixrieseberg:
@broerse: Hi, I'm the Microsoft engineer who wrote this - happy to explain what's going on here. Simple symlink rights are actually not sufficient, because UAC will strip that right from all shells not run with elevated permissions (for Administrator accounts). It's a slightly different story for user accounts without administrative rights and systems with UAC disabled, but we know that those users are in a very, very small minority.

The issues you're having stem from running as an entirely different user. That's not what we're asking here. We want you to make full use of the rights granted to your account.

Simple test: Run PowerShell, enter `$env:username`. Run as Administrator, enter the same thing. If both return the same name, you're good - and you should run Ember Cli with elevated rights.

broerse:
Thanks @felixrieseberg The test in PowerShell returns the same user. I did not disable UAC. I am a local admin. What do I do to run ember-cli on my computer and still have access to F: in the domain and run ember-cli with elevated rights? I just use the command prompt now, not PowerShell.

felixrieseberg:
If you are the same user in both modes, things should just work (do use powershell though, since it will cover some edge cases you night run into when using CMD). If Git or npm aren't working well, I'd assume that something is broken there - all that should be different when running with elevated rights is that the shell is allowed to do more - you should still be you, Git shouldn't be confused who you are ;-)

That being said, nothing keeps you from running the same way you did in the past - you'll loose a few seconds here and there, but we actually didn't change anything about how Ember Cli works, we're just trying to educate people that they can save a few seconds during build.

broerse:
I can't access F: on the domain in "Run as Administrator" in both PowerShell and CMD. I am not an Administrator on the domain.

So I want to:
 1. Run ember-cli as me.
 2. Access the domain for `git pull` / `git push`

Perhaps I still don't understand how to run this.
 
Or we can write a new mklink.exe that sets elevated permissions and stays in memory. That way we get a warning for UAC once a day.

felixrieseberg:
I don't know how your domain is configured or how it's even possible to have less access if you are running as the same user with more rights, but *do* know that this warning doesn't change anything about how Ember Cli runs - it just warns you about how things have been for years :/

stefanpenner:
Its worth noting, leaving ember-cli running, but opening a second window to do all your git stuff should work just fine then, it has the added benefit of not always requiring a full restart of the build server, rather incremental builds.

broerse:
@stefanpenner If I find the time I will write a better mklink.exe For now just go ahead with this. Still like to disable the warning though. 

@felixrieseberg I see the environment is not the same and I don't have drive-letters as Admin on the domain but I can access UNC Paths. Thanks for your help but that is out of scope for ember-cli.

stefanpenner:
> @stefanpenner If I find the time I will write a better mklink.exe For now just go ahead with this. Still like to disable the warning though.

this sounds interesting.

felixbuenemann:
@stefanpenner You probably meant to ping @felixrieseberg.

stefanpenner:
@felixbuenemann man, i keep doing that My apologies.

PR #4693, opened 2015-08-16
[ENHANCEMENT] Update Readme with correct npm support information

felixrieseberg:
This might be the tiniest PR, but correct npm support information is tremendously important for all of us that work with npm and Ember Cli on extremely low-powered systems or Windows, where npm@3 is literally the best thing ever.

- npm@2 version support taken from `package.json`
- npm@3 version support as discussed with @stefanpenner

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 The Ember.js command line utility.
 
-Supports node 0.12.x and npm 2.6.x.
+Supports node 0.12.x and npm 2.7.x and 3.x.
 
 ## Community
 * irc: #ember-cli on freenode


stefanpenner:
@felixbuenemann ah good catch! 

I have tried to stay on top with the recent NPM work, reporting issues as I discover them:

* https://github.com/npm/npm/issues/7816
* https://github.com/npm/npm/issues/8138
* https://github.com/npm/npm/issues/8643

We are still pending one more (although it shouldn't affect consumers)
* https://github.com/npm/npm/issues/9239

I suspect some add-ons that rely on a specific node_modules topology will need to be fixed like https://github.com/rwjblue/ember-cli-ic-ajax/pull/9  but those are easyish to fix, and soon cli will hide the complexity of npm debts from both users and addon authors.

PR #4685, opened 2015-08-14
[BUGFIX release] addonLintTree copies falsey logic from addonTreesFor

kellyselden:
Solution to #4340. Used the filter(Boolean) suggested by @rwjblue.

Added tests for `lintTree`.

Also used `this._mergeTrees` instead of `mergeTrees` in `ember-app.js` for test stubbing.

PR #4696, opened 2015-08-16
[INTERNAL release] proxy require for unit tests

kellyselden:
I wasn't too happy with how I stubbed out mergeTrees in https://github.com/ember-cli/ember-cli/pull/4685/files#diff-392bb207fea8bcbeb4a084bb445f7cc7R121 and then change everyone from `mergeTrees` to `this._mergeTrees` just for unit testing.

So I experimented and found a `require` proxying library. It allows for me to revert all that noise, for the tradeoff of a little bit more complicated test setup when you want to use it.

Thoughts? Better way? @stefanpenner @rwjblue

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -27,11 +27,10 @@ var preprocessMinifyCss = p.preprocessMinifyCss;
 var ES6Modules  = require('broccoli-es6modules');
 var concatFilesWithSourcemaps = require('broccoli-sourcemap-concat');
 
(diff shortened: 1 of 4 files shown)


stefanpenner:
> So I experimented and found a require proxying library. It allows for me to revert all that noise, for the tradeoff of a little bit more complicated test setup when you want to use it.

I'm fine with either approach, this one does prevent accidentally leaking mergeTrees to scenarios where we dont want it exposed.

kellyselden:
backported to stable in  #4697

PR #4700, opened 2015-08-17
[FEATURE] Allow addon commands to override core command if marked explicitly

rodyhaddad:
Sometimes it's useful for addons to be able to override a core command, at least I know I'm using it.
This permits it via a flag on a command called `overrideCore` that can be set to true.

I wasn't sure where's the best place to add docs to this, since it's a "static" property of the command.

We can move it to be a flag on the command instances, and it would align better with other options, but then we couldn't know if an addon command is intended to overwrite a core command unless we instantiate it which doesn't feel ideal.

If anybody has any thoughts on that last issue, please let me know :)

### DIFF
--- a/lib/cli/lookup-command.js
+++ b/lib/cli/lookup-command.js
@@ -38,6 +38,13 @@ module.exports = function(commands, commandName, commandArgs, optionHash) {
   }
 
   if (command && addonCommand) {
+    if (addonCommand.overrideCore) {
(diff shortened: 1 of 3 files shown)


rwjblue:
Seems related to https://github.com/ember-cli/rfcs/pull/6.

stefanpenner:
@rwjblue ya, this seems like a step in the right direction.

PR #4701, opened 2015-08-17
prepare-release improvements

stefanpenner:
* logging to know what is going on
* correctly path both WS packages


stefanpenner:
going to be taking a different approach

PR #4576, opened 2015-08-01
[WIP] adding --json to help

kellyselden:
ember-cli/rfcs#12

PR #4712, opened 2015-08-20
[INTERNAL release] add acceptance test for addon commands help

kellyselden:
Found another area that needs a regression test for json help #4576


stefanpenner:
:)

PR #4729, opened 2015-08-21
EmberApp -> EmberAddon in ember-cli-build.js

mike-north:


stefanpenner:
this should be under test

rwjblue:
@stefanpenner - It is under test, but there wasn't an actual bug (just very annoying).  It was that the local variable was incorrectly named (but consistently wrong so the tests passed fine).

PR #4730, opened 2015-08-21
Replaced "npm cache clear" on "npm cache clean"

Restuta:
As per NPM docs there is no "clear" option and "clean" should be used. https://docs.npmjs.com/cli/cache


stefanpenner:
Both work. But let's go with the one that has docs

PR #4731, opened 2015-08-21
Remove wasted work in app.import.

rwjblue:
In the early days of ember-cli the watcher was very expensive and we did not want to watch all of `bower_components`, but we have always intended to allow `app.import`ing a file from a given directory to allow that directory to be watched.

The way that we selectively watched these subdirectories in `bower_components` was by creating a tree for each file `app.import`'ed and then merging those together but not using any of their output!! This basically meant that for every `app.import` we created an extra ultimatley unused `Funnel` instance that we merged with the rest of the system resulting in merging some subdirectories of `bower_components` with themselves AND the rest of `bower_components` multiple times over again in each build.

This change removes all that crazy shenanigans and if `watchman` is present, just watches all of `bower_components`.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -96,7 +96,6 @@ function EmberApp() {
   this.legacyFilesToAppend     = [];
   this.vendorStaticStyles      = [];
   this.otherAssetPaths         = [];
-  this._importTrees            = [];
   this.legacyTestFilesToAppend = [];
   this.vendorTestStaticStyles  = [];
 
@@ -197,7 +196,7 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
     templates: existsSync('app/templates') ? unwatchedTree('app/templates') : null,
 
     // do not watch vendor/ or bower's default directory by default
-    bower: unwatchedTree(this.bowerDirectory),
+    bower: this.project._usingWatchman ? this.bowerDirectory : unwatchedTree(this.bowerDirectory),
     vendor: existsSync('vendor') ? unwatchedTree('vendor') : null,
 
     public: existsSync('public') ? 'public' : null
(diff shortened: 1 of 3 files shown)


stefanpenner:
the question is, should we keep the old code around for non watchman users? My vote is no, but there is a window until watchman for windows is ready.

@kellyselden / @felixrieseberg our windows contributors, what are your thoughts?

rwjblue:
@stefanpenner - I can certainly support both scenarios, but restarting a running `ember serve` after changing `bower_components` is likely a pretty small price to pay for every rebuild speedups (by doing much less work with `Funnel` and merges)...

stefanpenner:
> I can certainly support both scenarios, but restarting a running ember serve after changing bower_components is likely a pretty small price to pay for every rebuild speedups (by doing much less work with Funnel and merges)...

you wouldn't need to restart, merely `touch app/app.js` and the rebuild would detect changes.

felixrieseberg:
Correct me if I'm wrong, but people without watchman just have to restart `ember serve` after changing `bower_components`, while everybody wins a bit of performance? If so, that sounds great to me.

Pluuuuus: Watchman 1.0.9 passes all tests on AppVeyor/Windows, so it's probably just a matter of weeks before we get watchman Windows support.

rwjblue:
@felixrieseberg - You are correct (so is Stef), the worst case is restarting 'ember s', and the normal rebuild perf benefits are across the board (possibly even more perfs on windows when symlinking is disabled).

stefanpenner:
@felixbuenemann we have another plan that has the potentially to make those without symlinks and those with slow drives not suffer nearly as bad. It will also improve normal users experiences.

kellyselden:
@stefanpenner I'm on board with this.

stefanpenner:
Will confirm this evening all works as expected, then plan to merge.

PR #4734, opened 2015-08-22
[ENHANCEMENT] allow custom transforms for model generation

lazybensch:
When I have a custom transform, like 'utc' and then generate a model, I should be able to go 
```
  ember g model foo name:string startDate:utc data
```
and it should generate a file containing:
```
  name: DS.attr('string'),
  startDate: DS.attr('utc'),
  data: DS.attr()
```

currently however it is doing
```
  name: DS.attr('string'),
  startDate: DS.attr(),  // <-- narf
  data: DS.attr()
```


stefanpenner:
good catch!

PR #4735, opened 2015-08-22
Allow addon trees to be watched automatically with watchman ^3.7.0.

rwjblue:
`watchman` has added the ability to nest roots properly in version 3.7.0, this changes the `treeGenerator` to automatically allow trees to be watched if using `watchman` `^3.7.0`.


stefanpenner:
jshint failure

Stef Penner [7:42 AM]
then +1

–––

note this doesn't fix symlinked addons yet, we can likely implement a fix on our side. But watchman will be fixing this soon.

rwjblue:
The coveralls failure is because I added 3 extra lines in the "when watchman present" branch, which isn't tested via Travis (because we do not install `watchman` there).

@stefanpenner - Should be ready for another review...

PR #4737, opened 2015-08-23
Bump ember-export-application-global to v1.0.4

quaertym:


rwjblue:
Fixes deprecation with initializers in 2.1.0.

Issue #4562, opened 2015-07-30
Upgrading plugins to broccoli-plugin


joliss:
This issue tracks which plugins have been updated to derive from the broccoli-plugin base class. To see the plugins that need to be updated, run: `BROCCOLI_WARN_READ_API=y ember build`

Motivation: We want to rewrite the Broccoli builder to take advantage of the new broccoli-plugin base class. In particular, this will allow us to centralize tmp dir handling. But we need to update all plugins we use to derive from broccoli-plugin.

- [x] broccoli-filter [0.2.0 / 1.0.0](https://github.com/broccolijs/broccoli-filter/blob/master/CHANGELOG.md#100)
    - [x] broccoli-asset-rev 2.1.1: https://github.com/rickharrison/broccoli-asset-rev/pull/68
    - [x] broccoli-jshint 1.0.0: https://github.com/rwjblue/broccoli-jshint/pull/26
    - [x] broccoli-asset-rewrite 1.0.9
    - [x] ember-cli-htmlbars 0.7.10: https://github.com/ember-cli/ember-cli-htmlbars/pull/59
        - [x] ember-cli-app-version 1.0.0: https://github.com/embersherpa/ember-cli-app-version/pull/23
        - [x] ember-cli-htmlbars-inline-precompile 0.3.1: https://github.com/pangratz/ember-cli-htmlbars-inline-precompile/pull/23
- [x] cauliflower-filter (merged into broccoli-filter): https://github.com/broccolijs/broccoli-filter/pull/26
    - [x] broccoli-babel-transpiler 5.4.3: https://github.com/babel/broccoli-babel-transpiler/pull/48
- [x] broccoli-caching-writer 2.0.0: https://github.com/ember-cli/broccoli-caching-writer/pull/54
    - [x] ~~broccoli-es6modules: https://github.com/ember-cli/broccoli-es6modules/pull/26~~  we will drop this shortly in-favor of `ember-cli-babel`
    - [x] broccoli-sourcemap-concat 1.1.1 (variously shows as "Concat", "SourcemapConcat", or "Class"): https://github.com/ef4/broccoli-sourcemap-concat/pull/15
    - [x] [ConfigLoader](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/broccoli-config-loader.js): https://github.com/ember-cli/ember-cli/pull/4827
- [x] broccoli-merge-trees 0.2.2
[...]

PR #4740, opened 2015-08-23
Replace broccoli-unwatched-tree with broccoli-source

joliss:
Also start using `WatchedDir` in a few places. There are surely many more
plain-string trees (nodes) left. We'll want to replace them as we go.

broccoli-source supports the [new API](https://github.com/ember-cli/ember-cli/issues/4562).

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "broccoli-merge-trees": "0.2.3",
     "broccoli-sane-watcher": "^1.1.1",
     "broccoli-sourcemap-concat": "^1.0.0",
-    "broccoli-unwatched-tree": "0.1.1",
+    "broccoli-source": "^1.1.0",
     "broccoli-writer": "0.1.1",
     "chalk": "1.1.0",
     "clean-base-url": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "broccoli-merge-trees": "0.2.3",
     "broccoli-sane-watcher": "^1.1.1",
     "broccoli-sourcemap-concat": "^1.0.0",
-    "broccoli-unwatched-tree": "0.1.1",
+    "broccoli-source": "^1.1.0",
```
stefanpenner: quick google reveals this is quite healthy:

> It is a very good source of dietary fiber, pantothenic acid, vitamin B6, vitamin E, manganese, phosphorus, choline, vitamin B1, vitamin A (in the form of carotenoids), potassium, and copper. Broccoli is also a good source of vitamin B1, magnesium, omega-3 fatty acids, protein, zinc, calcium, iron, niacin, and selenium.

stefanpenner:
> Also start using WatchedDir in a few places. There are surely many more
> plain-string trees (nodes) left. We'll want to replace them as we go.

looks good.

Curious, can you point me to where `watched` property is used ? https://github.com/broccolijs/broccoli-source/blob/master/index.js#L31

joliss:
Very healthy! :)

At the moment, `watched` is only used by the `.read`-compat code: https://github.com/broccolijs/broccoli-source/blob/72f1ca198f659ef69a0ebc56f2374fd4b65f2f5b/index.js#L44
In the future, the Broccoli builder will use it as well, when it queries `__broccoliGetInfo__` directly.

stefanpenner:
@joliss thanks for the commit.

PR #4744, opened 2015-08-24
Derive CustomReplace plugin from Plugin instead of Writer

joliss:
Not sure if this will pass since I'm low on battery - need to wait for Travis. ;-)

### DIFF
--- a/lib/broccoli/broccoli-config-replace.js
+++ b/lib/broccoli/broccoli-config-replace.js
@@ -3,52 +3,27 @@
 var fs         = require('fs-extra');
 var existsSync = require('exists-sync');
 var path       = require('path');
-var Writer     = require('broccoli-writer');
+var Plugin     = require('broccoli-plugin');
 
-function CustomReplace (inputTree, configTree, options) {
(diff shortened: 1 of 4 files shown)


stefanpenner:
LGTM lets see what travis/appveyor say

joliss:
Travis keeps having weird [failures](https://travis-ci.org/ember-cli/ember-cli/jobs/77054959#L2755), and #4747 doesn't seem to have fixed it.

The Travis failures happen every time (they don't flicker), but when I run `npm run test-all:cover` locally, the test suite passes (1212 passing, 8 pending), so now I'm at a bit of a loss.

Is there some way we can improve the error reporting to figure out what's going on?

joliss:
Okay, found the issue (with some help from #4756) and fixed.

If you check the Travis build, there's [one weird failure](https://travis-ci.org/ember-cli/ember-cli/builds/77145887) - but I think it's either a network problem or a race condition. PR #4752, which is built on top of this PR, is [green](https://travis-ci.org/ember-cli/ember-cli/builds/77145880) as well.

So I think this is good to merge now.

PR #4747, opened 2015-08-24
Fix race condition in test

joliss:


stefanpenner:
@joliss ooh good catch!

rwjblue:
Thanks!

PR #4748, opened 2015-08-24
BROCCOLI_VIZ=true ember <s|b|test> will now produce broccoli instrume…

stefanpenner:
…ntation for external tools (like dot, and soon a js based viz tool) to display

After each rebuild, the output will be replaced.

The goal isn't just to make pretty things, but to help improve visibility into what is going on. Also, it will help pinpoint perf issues as we continue to add more instrumentation.

A quick output of a small app:
(300ms) or so after 100ms live-reload debounce.

graph


chadhietala:
:+1: This is awesome and we can build dashboards around this.

stefanpenner:
I believe @lightblade is playing with some ideas

stefanpenner:
I'll be adding more metrics as i go to these nodes.

PR #4756, opened 2015-08-25
Discard runCommand stdout/stderr unless we have a test failure

joliss:
This enables us to drop a lot of `--silent` flags in a second commit without cluttering the test output if everything passes.

(Being able to drop `--silent` flags then allowed me to debug an issue I had, changing from [silent failure](https://travis-ci.org/ember-cli/ember-cli/jobs/77054959#L2755) to [a useful error message](https://travis-ci.org/ember-cli/ember-cli/jobs/77137215#L2382).)

### DIFF
--- a/tests/acceptance/blueprint-test-slow.js
+++ b/tests/acceptance/blueprint-test-slow.js
@@ -39,8 +39,7 @@ describe('Acceptance: blueprint smoke tests', function() {
     return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'generate',
                       'http-proxy',
                       'api',
-                      'http://localhost/api',
-                      '--silent')
+                      'http://localhost/api')
       .then(function() {
         var packageJsonPath = path.join(__dirname, '..', '..', 'tmp', appName, 'package.json');
         var packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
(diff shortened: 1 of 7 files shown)


stefanpenner:
Awesome. Been wanting to do this for some time, thank you :)

stefanpenner:
We may also want it to log on timeout. But we can do that in some future cycle

joliss:
Hm, the logging on timeout works for me, it gets treated like any other test failure:

```js
diff --git a/tests/acceptance/preprocessor-smoke-test-slow.js b/tests/acceptance/preprocessor-smoke-test-slow.js
index 3abc9c8..ee43fa9 100644
--- a/tests/acceptance/preprocessor-smoke-test-slow.js
+++ b/tests/acceptance/preprocessor-smoke-test-slow.js
@@ -38,8 +38,8 @@ describe('Acceptance: preprocessor-smoke-test', function() {
     });
   });

-  it('addons with standard preprocessors compile correctly', function() {
-    this.timeout(100000);
+  it.only('addons with standard preprocessors compile correctly', function() {
+    this.timeout(1000);

     return copyFixtureFiles('preprocessor-tests/app-with-addon-with-preprocessors')
       .then(function() {
@@ -51,6 +51,9 @@ describe('Acceptance: preprocessor-smoke-test', function() {
         return fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson));
       })
       .then(function() {
+        return runCommand('echo', 'foo');
+      })
+      .then(function() {
         return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'build');
       })
       .then(function() {
```

Unless some other `afterEach` hook fails, then the log printer doesn't get to run. But normally I think it should print. Can you show me an example where it doesn't log on timeout?

stefanpenner:
Awesome. If it does that already via the existing timeouts then it should be good :)

PR #4759, opened 2015-08-25
Move MODEL_FACTORY_INJECTIONS flag to environment.js

xtian:
emberjs/ember.js#3271 made it possible to set this flag on `EmberENV` instead of `Ember` itself.

In my opinion, it would be nice to have it out of app.js and kept with the rest of the config flags.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -3,11 +3,7 @@ import Resolver from 'ember/resolver';
 import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 
-var App;
-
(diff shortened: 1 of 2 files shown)


pixelhandler:
@xtian yes much better in my opinion to have with the rest of the config :+1:

stefanpenner:
I fear this change is more of an upgrade hazard.

xtian:
Do you think it would be appropriate as part of the 2.0 upgrade?

stefanpenner:
It doesn't really seem like something one can turn off or on based on environment. It is a relic from the past, that is stuck with us. I believe leaving it as is, until eventually removed altogether, is most appropriate.

xtian:
Alright

PR #4764, opened 2015-08-26
Replace Esperanto With Babel

chadhietala:
- [x] Remove broccoli-es6modules
- [x] Skip funnel and transpilation if babel is installed in user land
- [x] Hookup to an app to verify.

This makes babel required. The next step after this lands is to figure out what `Class` is referring to come up with a better concat strategy as it appears to be an issue.

__Final__

__Before__

__After__


stefanpenner:
:dances:

rwjblue:
Just to make sure I understand properly, `amd-name-resolver` is used to ensure no relative paths end up in the transpiled output?

chadhietala:
@rwjblue yes

rwjblue:
I believe that `ember-cli-babel` disables `es6.modules` via a blacklist [here](https://github.com/babel/ember-cli-babel/blob/master/index.js#L71) unless specifically present in the config.

lib/broccoli/ember-app.js
```diff
@@ -139,6 +139,22 @@ EmberApp.prototype._initProject = function(options) {
...
+
+  if (this._addonInstalled('ember-cli-babel')) {
+    var amdNameResolver = require('amd-name-resolver');
+    babelOptions = {
+      babel: {
+        compileModules: true,
+        modules: 'amdStrict',
+        moduleIds: true,
+        exportModuleMetadata: true,
+        sourceMaps: !isProduction,
```
rwjblue: This should be based on `options.sourcmaps.enabled` (which is defaulted to `!isProduction` below but can be enabled and disabled independently of a production build).

rwjblue:
I am very excited about this! Thanks for spearheading @chadhietala!

chadhietala:
From what i gathered you have to set `compileModules` to true to get around this in `ember-cli-babel.`

chadhietala:
No idea how any tests are passing on node 0.10.0 due to the fact broccoli-sri-hash uses a [native promise](https://github.com/jonathanKingston/broccoli-sri-hash/blob/master/index.js#L160).

chadhietala:
Update:

Need to do further investigation as the trees created in `_addAppTests` need to be babelified.

joostdevries:
This completely removes esperanto right? So module transpilation is done using Babel? Maybe update the PR title to reflect that?

rwjblue:
@joostdevries - It only removes the usage of Esperanto *if* the project has `ember-cli-babel` installed.  Due to various reasons (@stefanpenner can elaborate) we do not *require* apps to use `ember-cli-babel`.

@stefanpenner / @chadhietala - I believe that master is ultimately going to become ember-cli 2.0.0, I think we should change to requiring `ember-cli-babel` in 2.0.0.  This would be a breaking change (hence my desire to include it in 2.0.0, but largely not affect the majority of our users since `ember-cli-babel` is included in all blueprinted apps.  Thoughts?

stefanpenner:
@rwjblue yes agreed

chadhietala:
This PR actually drops esperanto altogether and uses babel for the transpilation if they do not have `ember-cli-babel` installed. So if you do or do not have ember-cli-babel it should be transparent. We can go 1 step further and whitelist `es6.modules` in the event the app does not have `ember-cli-babel`.

rwjblue:
> This PR actually drops esperanto all together and uses babel for the transpilation if they do not have ember-cli-babel installed

Nice! Must have been added after I reviewed the diff initially...

If we intend to support apps without ember-cli-babel, then we need some slow acceptance tests to confirm that works. I personally think that we should just require that ember-cli@2.0.0 apps have ember-cli-babel.  If that is what we intend to do, we should add a deprecation on top of 1.13.8 and release as 1.13.9 to make it clear that this will be removed in 2.0.0.

chadhietala:
I'm up for making it required and ember-cli will just have some sane defaults for babel. I'm currently dealing with the fact that some addons were relying on the fact that in the post process hook they had un-processed files. In our app this is ember-browserify. Will probably need to PR and [fix this](https://github.com/ef4/ember-browserify/blob/master/lib/index.js#L38) as well as modules are completely transpiled once they enter this hook.

chadhietala:
So I verified that we will save  ~5sec in a pathological case.

__Before__


__After__


I would have expected more but could be our app ¯\_(ツ)_/¯

@rwjblue and @stefanpenner not sure what is going on with CI. Feel free to pull this locally and test.

chadhietala:
After the latest commit I identified where some double work was occurring. Here is the new rebuild time so the savings looks to be close to ~8sec which is the approximate time `ES6` and `App JS Files` were taking.

stefanpenner:
20% - 25% win in our big app – and its test suite passes.

stefanpenner:
rebased and stuff here: https://github.com/ember-cli/ember-cli/pull/4792

jcope2013:
so is the assumption of this PR that ember-cli master is now required to use ember-cli-babel? It seems taking out ember-cli-babel will make an app no longer work properly due to

``` 
Uncaught ReferenceError: exports is not defined
````

screen shot 2015-11-01 at 12 41 50 pm

```
ember new foo
cd foo
npm link ember-cli
remove ember-cli-babel from package.json
rm -rf node_modules/ember-cli-babel/
ember server
```

stefanpenner:
> so is the assumption of this PR that ember-cli master is now required to use ember-cli-babel?

It bundles its own.

chadhietala:
@stefanpenner I can take a look, but I had been using the master branch locally for a while with no issues.

PR #4771, opened 2015-08-30
Make default generated tests pass.

blimmer:
The changes to the .jshintrc defaults and as part of #4663  cause the default tests to fail.

### DIFF
--- a/blueprints/component-test/files/tests/__testType__/__path__/__test__.js
+++ b/blueprints/component-test/files/tests/__testType__/__path__/__test__.js
@@ -15,7 +15,7 @@ test('it renders', function(assert) {
 
   this.render(hbs`{{<%= componentPathName %>}}`);
 
-  assert.equal(this.$().text(), '');
+  assert.equal(this.$().text().trim(), '');
 
   // Template block usage:" + EOL +
   this.render(hbs`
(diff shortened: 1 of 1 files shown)


blueprints/component-test/files/tests/__testType__/__path__/__test__.js
```diff
@@ -27,8 +27,8 @@ test('it renders', function(assert) {
   assert.equal(this.$().text().trim(), 'template block text');<% } else if(testType === 'unit') { %>assert.expect(1);
 
   // Creates the component instance
-  var component = this.subject();
+  /*var component =*/ this.subject();
```
blimmer: I've seen this style used in other place, e.g. in ember-cli-mirage. If we don't like this style. We could assert that `component` is truthy so the variable is used and JSHint doesn't complain. I do that in the [ember-cli-mocha](https://github.com/switchfly/ember-cli-mocha/blob/v0.9.2/blueprints/component-test/index.js#L33) flavor of the default test.

stefanpenner:
strange, the default tests themselves are under test. Any idea why they are not reporting a failure?

blimmer:
This issue only occurs when generating a component test. I did some spelunking in the code, and I couldn't see a place where a component was generated and that test was subsequently run. Did I miss something?

stefanpenner:
seems reasonable for now :) thanks

PR #4772, opened 2015-08-30
Implement a destroyApp helper.

blimmer:
This is similar to startApp, and is very simple for the time being. However, this allows addon authors to hook into when the app is destroyed in a testing environment.

### DIFF
--- a/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
+++ b/blueprints/acceptance-test/files/tests/acceptance/__name__-test.js
@@ -1,6 +1,6 @@
-import Ember from 'ember';
 import { module, test } from 'qunit';
 import startApp from '<%= testFolderRoot %>/tests/helpers/start-app';
+import destroyApp from '<%= testFolderRoot %>/tests/helpers/destroy-app';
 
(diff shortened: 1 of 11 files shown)


blimmer:
Specifically, this came up with ember-cli-mirage. We need to shutdown the pretender server via  `server.shutdown()` after each acceptance test (described [here](https://github.com/samselikoff/ember-cli-mirage/issues/226)). However, because of differences in testing frameworks, there's not a clean place for us to implement this behavior.

With this change, we'd be able to edit the `destroy-app.js` file on `ember install` of the addon to shutdown the server.

blimmer:
If this gets accepted, I'll go ahead and update ember-cli-mocha to utilize it, too.

trabus:
LGTM :+1:

cibernox:
Simple and clever :+1:

blimmer:
I have an [open PR](https://github.com/abuiles/ember-watson/pull/84) on ember-watson to assist folks in migrating their old acceptance tests to use the new helper.

stefanpenner:
@blimmer :+1:

PR #4781, opened 2015-08-31
[BUGFIX] Set `ember serve --host` default to `undefined`.

### LINKED ISSUE #4775: ember serve --host's default disables IPv6
When `ember serve` is run without the `--host` option, it uses the default host, which is [defined as `"0.0.0.0"`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/commands/serve.js#L21).

This works fine for the old IPv4 stack. However, by setting the host to `"0.0.0.0"` and not passing a `null` / `undefined` value, Node.js (implicitly?) only binds to all IPv4 interfaces.

To make ember-cli listen to the IPv6 devices instead, I currently have to pass `--host ::`. However, this simultaneously disables the IPv4 stack.

I see multiple solutions to this problem and would make a PR, if we can agree on one.

1. Set the default value of `host` to `null` in [`commands/serve.js`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/commands/serve.js#L21) and teach [`express-server.js#displayHost`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/tasks/server/express-server.js#L37-L39) and [`livereload-server.js#displayHost`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/tasks/server/livereload-server.js#L70-L72) about it.

2. Implicitly discard `host` if it is `"0.0.0.0"` and pass `null` instead at [`express-server.js#startHttpServer`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/tasks/server/express-server.js#L204) and [`livereload-server.js#start`](https://github.com/ember-cli/ember-cli/blob/v1.13.8/lib/tasks/server/livereload-server.js#L44).

3. Set the default value of `host` to a magic keyword like `INADDR_ANY` or `all` and implement the changes accordingly.

Personally, I'd prefer solution 1. Solution 3 just feels very clunky and I don't like how solution 2 hides what's really happening; if I read that `"0.0.0.0"` is the default, I'd expect that this really is the value that is passed to the `.listen()` call.

Solution 1 is the cleanest imho. It might only look a bit weird in `ember help serve`, as I don't know how `nopt` handles `null` values.

buschtoens:
Fixes #4775.

I adapted the test cases where appropriate and removed obsolete tests. The following test cases were removed, because they are no longer applicable:

- server command has correct options
  - [a livereload port is found using the default host](https://github.com/ember-cli/ember-cli/compare/master...buschtoens:fix-4775-default-host?expand=1#diff-6b87c9728965f1ddffe252463882d03cL51) (removed)
  - [has correct liveReload host](https://github.com/ember-cli/ember-cli/compare/master...buschtoens:fix-4775-default-host?expand=1#diff-6b87c9728965f1ddffe252463882d03cL57) (removed)

All other relevant tests have been updated accordingly.

```
$ ember help serve
version: 1.13.8
Requested ember-cli commands:

ember serve <options...>
  Builds and serves your app, rebuilding on file changes.
  aliases: server, s
  --port (Number) (Default: 4200)
    aliases: -p <value>
  --host (String) Listens on all interfaces by default
    aliases: -H <value>
  ...
```

### DIFF
--- a/tests/unit/tasks/server/express-server-test.js
+++ b/tests/unit/tasks/server/express-server-test.js
@@ -48,12 +48,12 @@ describe('express-server', function() {
   });
 
   describe('displayHost', function() {
-    it('should use the specified host if not 0.0.0.0', function() {
+    it('should use the specified host if specified', function() {
       expect(subject.displayHost('1.2.3.4')).to.equal('1.2.3.4');
     });
 
-    it('should use the use localhost if specified host is 0.0.0.0', function() {
-      expect(subject.displayHost('0.0.0.0')).to.equal('localhost');
+    it('should use the use localhost if host is not specified', function() {
+      expect(subject.displayHost(undefined)).to.equal('localhost');
     });
   });
 
@@ -86,7 +86,7 @@ describe('express-server', function() {
 
     it('with ssl', function() {
       return subject.start({
-        host:  '0.0.0.0',
+        host: undefined,
         port: '1337',
         ssl: true,
         sslCert: 'tests/fixtures/ssl/server.crt',
@@ -101,7 +101,7 @@ describe('express-server', function() {
     it('with proxy', function() {
       return subject.start({
         proxy: 'http://localhost:3001/',
-        host:  '0.0.0.0',
+        host: undefined,
         port: '1337',
         baseURL: '/'
       }).then(function() {
@@ -114,7 +114,7 @@ describe('express-server', function() {
 
     it('without proxy', function() {
       return subject.start({
-        host:  '0.0.0.0',
+        host: undefined,
         port: '1337',
         baseURL: '/'
       }).then(function() {
@@ -126,7 +126,7 @@ describe('express-server', function() {
 
     it('with baseURL', function() {
       return subject.start({
-        host:  '0.0.0.0',
+        host: undefined,
         port: '1337',
         baseURL: '/foo'
       }).then(function() {
@@ -141,7 +141,7 @@ describe('express-server', function() {
       preexistingServer.listen(1337);
 
       return subject.start({
-        host:  '0.0.0.0',
+        host: undefined,
         port: '1337'
       })
         .then(function() {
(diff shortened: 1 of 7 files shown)


tests/unit/tasks/server/express-server-test.js
```diff
@@ -692,7 +692,7 @@ describe('express-server', function() {
 
       it('calls serverMiddleware on the addons on start', function() {
         return subject.start({
-          host:  '0.0.0.0',
+          host: undefined,
```
stefanpenner: Question, should we just remove this property entirely? Or do you feel specifying `undefined` explicitly provides value.

buschtoens: I left it in, because it's more explicit. Might make it easier to quickly scan the tests.

But I can remove it and squash the commits, if you want. :)

stefanpenner: @buschtoens nope, thats fine with me.

stefanpenner:
LGTM

r? @rwjblue

rwjblue:
The changes look good.  The only thing that seemed odd was that the permissions changed on all the files from 0644 to 0755 (making them all executable).

stefanpenner:
> The only thing that seemed odd was that the permissions changed on all the files from 0644 to 0755 (making them all executable).

good catch, lets get those restored.

buschtoens:
Oh yeah. That's weird. I'll fix that. Sorry.

buschtoens:
File permissions should be fixed now.

stefanpenner:
> File permissions should be fixed now.

:+1:

buschtoens:
Is there anything else holding this back from being merged? :)

rwjblue:
Nope, looks good, thank you!

buschtoens:
:beers:

stefanpenner:
hmm, so this seems to cause some issues as well. Im unclear what the best solution is.

It appears their is a order of precedents i wasn't expecting, let me try to illustrate the issue I ran into.

* X services binds to localhost:4XXXX ipv4
* our portfinder now using host undefined, checks *:4XXXXX ipv6, does not notice the ipv4 variant is used
* curl localhost:4XXXXX hits our services
* in /etc/hosts and entry 127.0.0.0.1 foo.com  exists
* curl foo.com:4XXXXX hits our IPV4 service.

sadness and confusion prevail.

Issue #4527, opened 2015-07-25
Improve test suite times


stefanpenner:
We have several types of tests, currently we are leaning to heavily on our end 2 end acceptance tests. These provide great coverage, but unfortunately slow down active development.

We likely need to take several approaches
- improve the throughput of the existing acceptance tests.
- rely more heavily on fast unit tests when enumerating permutations.
- transition some of the acceptance tests to be fast unit tests.

Some places where E2E tests are slow:
- CLI startup time has been increasing again, a quick look yielded: https://github.com/ember-cli/ember-cli/pull/4526 (this compounds for E2E tests)

martypenner:
I'll be of little help on this until my node debugging ability improves. Once I uncover something useful, I'll submit a PR.

stefanpenner:
@martypenner I can likely make some time later this week or early next week to pair for like 30 min to an hour. Maybe I can share some tips

martypenner:
That would be amazing. I'll figure out my schedule soon and see what you've got available.

stefanpenner:
tuesday maybe?

martypenner:
Sure, that works. What time are you thinking? I can likely make any time work; just have to OK it with the team.

stefanpenner:
what timezone are you in?

martypenner:
EDT

martypenner:
You're on PDT, right?

kellyselden:
I will look into this.

yaymukund:
I second this wish! I'll see what I can dig up when I have time maybe this weekend.

webPapaya:
What about running all tests in nodejs with jsdom instead of the browser or phantomjs? This would probably improve testing times.

stefanpenner:
> What about running all tests in nodejs with jsdom instead of the browser or phantomjs? This would probably improve testing times.

these tests refer to ember-cli's own test suite.

webPapaya:
@stefanpenner sorry I completely missed the point of this ticket. I'm currently looking for a solution to run  unit tests of an ember application in nodejs instead of phantomjs and came across this ticket. I'll file a new ticket.

stefanpenner:
One thing i noticed, we often have `generate x` test and also a `destroy x` test, the later generates and then destroys. I suspect all of those should be merged into the same test. No need to duplicate all that work.

Turbo87:
[...]

PR #4785, opened 2015-09-01
some safe runCommand removals

kellyselden:
replacing with ember test helper

cc #4527

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -11,6 +11,7 @@ var chalk      = require('chalk');
 var expect     = require('chai').expect;
 
 var runCommand          = require('../helpers/run-command');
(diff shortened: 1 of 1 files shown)


stefanpenner:
nice, although CI doesn't agree yet.

kellyselden:
I tested each individually using `.only`. Of course when run together they start failing... I will look into this tomorrow.

stefanpenner:
if the run `npm` install internally, they are not safe.

kellyselden:
I had to remove all the offenders, leaving only one removal. Still, shaves off time.

@stefanpenner the commands being run were `test` and `build`, which I don't believe involve npm.

stefanpenner:
> @stefanpenner the commands being run were test and build, which I don't believe involve npm.
>  Hide all checks

crap :P

PR #4786, opened 2015-09-01
[BUGFIX] Updates base adapter class for Ember Data 2.0

locks:
As of Ember Data 2.0 the default adapter should be JSONAPIAdapter. This changes the blueprint accordingly.

Aside, should this generator be moved to the ember-data addon?

### DIFF
--- a/blueprints/adapter/index.js
+++ b/blueprints/adapter/index.js
@@ -13,7 +13,7 @@ module.exports = {
 
   locals: function(options) {
     var adapterName     = options.entity.name;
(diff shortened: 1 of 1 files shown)


trabus:
> Aside, should this generator be moved to the ember-data addon?

I believe so, as well as the serializer.

stefanpenner:
ember-cli will keep these frozen for now, and soon will be removing them alltogether.

Ember-data should provide its own blueprints, that way the blueprints generated are that of the version of ED the user has installed.

PR #4796, opened 2015-09-03
[FIXES #4794] tested needs null integrity value (since it always chan…

### LINKED ISSUE #4794: "Waiting for runners" when using Chrome
Hey there,

My project is using:
* Ember: 1.13.8
* EmberCLI: 1.13.8
* Ember Data: 1.13.8 
* Chrome Stable 45
* Node: v0.12.0
* NPM: 2.7.3
* OS: OSX 10.10.5

We use Chrome to test our app locally and on CircleCI. Yesterday out of nowhere testem stopped functioning normally. Chrome would open and our tests would run, but testem would sit there saying "Waiting for runners". Even if Chrome completes the test suite, testem never seems to be "connected." PhantomJS works fine. 

**What works:**
I reverted to a version of our project that uses ember 1.11 and destroyed and rebuilt my bower, node, and tmp directories and ran our suite and it ran fine. 

Then I setup an empty directory with a single JS specfile and the testem.json file that EmberCLI generates. I removed the "test_page" property and ran `testem`. It opened Chrome and ran the single spec without issue.

**What doesn't work:**
I then setup a brand new EmberCLI 1.13.8 project and ran the test suite and encountered the same "Waiting for runners" error. The testem.json that I used for this is here: https://gist.github.com/jakemauer/1eb4c05d3ea08a6defd6

Additionally if I open the provided testem URL directly in Chrome, it will run the tests just fine, but never signal back to testem that they're running. Firefox however WILL communicate and testem's CLI interface will show the Firefox tab. 

Sometimes when I quit `ember test --server` I get an error dump visible here: https://gist.github.com/jakemauer/1eb4c05d3ea08a6defd6

I'm not sure where to go from here. I was convinced it was something in our code but because it's acting weird with a brand new 1.13.8 project I'm at a loss. Thanks for your eyeballs and help.

stefanpenner:
…ges)

cc @jonathanKingston

### DIFF
--- a/blueprints/app/files/tests/index.html
+++ b/blueprints/app/files/tests/index.html
@@ -24,7 +24,7 @@
     <script src="assets/vendor.js"></script>
     <script src="assets/test-support.js"></script>
     <script src="assets/<%= name %>.js"></script>
-    <script src="testem.js"></script>
+    <script src="testem.js" integrity=""></script>
     <script src="assets/test-loader.js"></script>
 
     {{content-for 'body-footer'}}


jonathanKingston:
Yup that was the commit that I have not submited... I was looking for the blueprint tester before submitting.

jonathanKingston:
+1 sorry for the terse response, I had just woken up. I appreciate this going in a lot!

stefanpenner:
We should back port a test. I was in a hurry, something that just validates that tag exists with the important attribute in the final output

jonathanKingston:
@stefanpenner as mentioned some blueprint helpers that build to a temp file and allow AST parsing / regex there would be the most robust I think.

stefanpenner:
> @stefanpenner as mentioned some blueprint helpers that build to a temp file and allow AST parsing / regex there would be the most robust I think.

:+1:

PR #4809, opened 2015-09-08
Update bower dependencies and use tilde.

abuiles:
After starting a new project working off `ember-cli#master`, I had to update the packages in the PR to get things working (also I'm using Ember and Ember-Data beta) 

`ember-cli-shims` and `ember-load-initializers` were not registered in bower, I just did that so this should work. I know we want to get off-bower but while we do we should make the process as easy as possible for newcomers.

### DIFF
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -2,14 +2,14 @@
   "name": "<%= name %>",
   "dependencies": {
     "ember": "2.0.0",
-    "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.5",
-    "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
+    "ember-cli-shims": "~0.0.5",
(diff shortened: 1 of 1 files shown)


rwjblue:
We should not use permissive versioning with < 1.0.0 projects (that has been our general rule).  I am not opposed to bumping the libraries that we control to 1.0.0 to allow this though...

abuiles:
@rwjblue sounds good, all the bumped libs are under ember-cli org. 

Also since we  are specifying patch # the tilde will keep only bubbling up in patch, which I believe is fine.

stefanpenner:
needs rebase

abuiles:
@stefanpenner putting on hold while we bump the other packages to 1.0.0

abuiles:
@rwjblue I'm having seconds thoughts about it, shouldn't we better wait and aim to port this to addons?

PR #4816, opened 2015-09-08
[ENHANCEMENT] Allow OS to choose ephemeral port if --test-port=0

williamsbdev:
We ran into this problem on our CI server as we have many ember project build running at the same time and colliding on the default port. We started getting tricky by trying to bind to a port prior to running the tests but then we still ended up with weird race conditions. If we are allowed to pass 0, the OS will choose an open port at the time of test execution. We were unable to set the `--test-port=0` as the check in the `TestCommand` was a simple truthy check that would fail if it was set to `0`. With the modification to check if it is an int, we will be able to pass 0 through.

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -34,12 +34,15 @@ module.exports = Command.extend({
   ],
 
   run: function(commandOptions) {
+    var port = commandOptions.port ? Promise.resolve(commandOptions.port) : this._getPort({ host: commandOptions.host });
     var liveReloadHost = commandOptions.liveReloadHost || commandOptions.host;
-    var liveReloadPort = commandOptions.liveReloadPort ? Promise.resolve(commandOptions.liveReloadPort) :
-                                                         this._getPort({ host: liveReloadHost });
+    var liveReloadPort = commandOptions.liveReloadPort ? Promise.resolve(commandOptions.liveReloadPort) : this._getPort({ host: liveReloadHost });
 
-    return liveReloadPort.then(function(liveReloadPort) {
+    return Promise.all([liveReloadPort, port]).then(function(values) {
+      var liveReloadPort = values[0];
+      var port = values[1];
       commandOptions = assign({}, commandOptions, {
+        port: port,
         liveReloadPort: liveReloadPort,
         liveReloadHost: liveReloadHost,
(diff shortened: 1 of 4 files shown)


stefanpenner:
some things:

* rebase the commits, some of those commits do not appear related to the diff
* 0 as the key seems funky, should we maybe do `--test-port=auto`
* we should also allow this for --port and --live-reload-port (which already does this by default)

williamsbdev:
> rebase the commits, some of those commits do not appear related to the diff

done

> 0 as the key seems funky, should we maybe do `--test-port=auto`

This does seem a little funky, however it is unix convention that port 0 means to dynamically generate an unused port. See this [stackexchange question/answer]( http://unix.stackexchange.com/questions/180492/is-it-possible-to-connect-to-tcp-port-0). Additionally, on this [wikipedia article](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) in the *Well-known ports* section, you will see that the port 0 description is "Programming technique for specifying system-allocated (dynamic) ports".

> we should also allow this for --port and --live-reload-port (which already does this by default)

I have updated `--port` to mimic the `--live-reload-port` behavior.

williamsbdev:
@stefanpenner are these real failures or just timeout errors during the build?

kellyselden:
@williamsbdev They appear network related. Kicking the builds.

williamsbdev:
@kellyselden thank you!

williamsbdev:
@kellyselden and @stefanpenner it appears that appveyor keeps timing out. Is this common?

stefanpenner:
@williamsbdev ya windows CI is flakey, giving it a kick now.

stefanpenner:
> This does seem a little funky, however it is unix convention that port 0 means to dynamically generate an unused port. See this stackexchange question/answer. Additionally, on this wikipedia article in the Well-known ports section, you will see that the port 0 description is "Programming technique for specifying system-allocated (dynamic) ports".

then i believe that is totally reasonable, thanks for pointing that out.


Test failure it totally unrelated. Upstream dep is borked, I will be looking into it shortly

williamsbdev:
@stefanpenner thank you for merging this and looking into the reason for the broken build!

stefanpenner:
:)

PR #4819, opened 2015-09-10
[FEATURE] Add clean script to package.json in app blueprint

elwayman02:
Adds a clean script the default app blueprint, allowing users to run the following command:

`npm run clean`

which will clean the npm/bower caches and remove node_modules/bower_components/dist/tmp

As outlined [here](http://jhawk.co/upgrading-ember-i).

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "build": "ember build",
+    "clean": "npm cache clean && bower cache clean && rimraf node_modules bower_components dist tmp",
     "start": "ember server",
     "test": "ember test"
   },


rwjblue:
:+1: - I like it. :smiling_imp:

quaertym:
:+1:

tsing80:
this may not work in windows

kellyselden:
Yes, `rm` is a windows powershell command that works differently.

rwjblue:
It may also have issues with path length, perhaps we could use the `rimraf` binary (it should already be available) instead?

elwayman02:
Can someone confirm whether this command works on windows?

`npm cache clean && ./node_modules/bower/bin/bower cache clean && ./node_modules/rimraf/bin.js node_modules bower_components dist tmp`

kellyselden:
@elwayman02 
```
'.' is not recognized as an internal or external command,
operable program or batch file.
```

elwayman02:
Yup knew it was something like that, thanks @kellyselden. Any suggestions on how to access this cross-platform?

kellyselden:
This command works in both Linux and Windows:
```
npm cache clean && bower cache clean && node node_modules/ember-cli/node_modules/rimraf/bin node_modules bower_components dist tmp
```
But, it doesn't rerun safely. The second run will hard error because rimraf is gone.

elwayman02:
We could use bash to check if rimraf exists, but I'm not sure how to do that on both platforms

something like this:
```npm cache clean && bower cache clean && [ -f node_modules/ember-cli/node_modules/rimraf/bin ] && node_modules/ember-cli/node_modules/rimraf/bin node_modules bower_components dist tmp```

Though if we are worried about rimraf not being a global, we should probably ask the same thing about bower and check that as well...I'm wondering if it makes more sense to just assume that npm, bower, and rimraf are all globals, and people can install them as such or modify the default script after executing this blueprint. At the end of the day, this is a command that will only be run explicitly by the user, so I'm not sure how in-depth this needs to be (though we want to catch the broadest reasonable use-cases).

Thoughts @rwjblue?

stefanpenner:
can we make `ember clean` instead? And defer to a node script, rather then some very specific shell syntax.

elwayman02:
:+1: @stefanpenner that sounds like a great idea

stefanpenner:
@elwayman02 no need, i think you script for `clean` is exactly what we need, we should just implement in a way that is portable (some shells don't support that syntax, although most do)

elwayman02:
Ok, well I'm open to suggestions on how to get started...working on two addons this weekend so I don't have a ton of time, but I wouldn't mind giving it a shot if you have an idea of what we could use in node to get this done.  Alternatively, if someone has time and wants to take this up, they're welcome to do so.

stefanpenner:
this seems to have gone stale, closing.

PR #4821, opened 2015-09-10
[ENHANCEMENT] Open default browser to express app on serve

jayphelps:
While this would be helpful for me, I'm not sure if this is ideal for a majority of people's workflow or just annoying.

### DIFF
--- a/lib/tasks/serve.js
+++ b/lib/tasks/serve.js
@@ -2,6 +2,7 @@
 
 var existsSync       = require('exists-sync');
 var path             = require('path');
+var opn              = require('opn');
 var LiveReloadServer = require('./server/livereload-server');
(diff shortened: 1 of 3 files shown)


jherdman:
It may be a tad annoying for me. The majority of my development is done on a headless VM. I connect my browsers to it from outside of the VM.

kellyselden:
Ditto. I develop and run ember serve on a Linux VM, and connect to it from Windows host browser.

kellyselden:
I could see it as an opt-in feature though.

stefanpenner:
> I could see it as an opt-in feature though.

:+1: maybe `--open` ?

then for those who always want it, they can insert it into there `.emberrc` file

PR #4829, opened 2015-09-12
broccoli-plugin{description -> annotation}

stefanpenner:


stefanpenner:
failure is upstream, and unrelated. Will be looking into it myself.

PR #4836, opened 2015-09-12
Make config replace cache

chadhietala:


lib/broccoli/broccoli-config-replace.js
```diff
@@ -22,16 +24,54 @@ CustomReplace.prototype.build = function () {
...
 
-    this.processFile(config, filePath, destPath);
+    if (this._cache[key.hash]) {
+      this.writeFile(this._cache[key]);
+    } else {
+      filePath = path.join(inputPath, file);
+      destPath = path.join(this.outputPath, file);
+
+      this.processFile(config, filePath, destPath, key);
+    }
```
stefanpenner: we may want to evict cache entries

chadhietala: Yea it's doable. Would grow unbounded to the times you change the config.

stefanpenner:
Although it seems like a very slow growing cache. If possible we should do our do diligence here.

stefanpenner:
failing tests are unrelated, will look into them later this weekend

stefanpenner:
we should extract this and write tests for it, and this PR is actually broken on rebuilds.

stefanpenner:
fixed: https://github.com/ember-cli/ember-cli/commit/8ddf7b74e5eef247a9558caecf68ace2a1330ec0

PR #4837, opened 2015-09-12
Add `Blueprint.prototype.filesPath`.

rwjblue:
By default this returns `path.join(this.path, 'files')`, but it can easily be overridden by blueprints.

One common example is to provide blueprints for both `mocha` and `qunit`. Before these changes you would have to override both `files` and `srcPath` methods (because `path.join(this.path, 'files')` was hard coded in both).

Now, you could add the following to support `mocha` and `qunit` style tests in a test blueprint:

```javascript
var path = require('path');

module.exports = function() {
  description: 'Generate a test for XXXX',
  filesPath() {
    if (this._filesPath) { return this._filesPath; }

    var type;
    if ('ember-cli-mocha' in this.project.addonPackages) {
      type = 'mocha';
    } else if ('ember-cli-qunit' in this.project.addonPackages) {
      type = 'qunit';
    }

    return path.join(this.path, type + '-files');
  }
}
```

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -282,6 +282,18 @@ Blueprint.prototype.availableOptions = [];
 Blueprint.prototype.anonymousOptions = ['name'];
 
 /**
+  Used to determine the path to where files will be stored. By default
+  this is `path.join(this.path, 'files)`.
+
+  @method filesPath
+  @return {String} Path to the blueprints files directory.
+*/
+
+Blueprint.prototype.filesPath = function() {
+  return path.join(this.path, 'files');
+};
+
+/**
   Used to retrieve files for blueprint. The `file` param is an
   optional string that is turned into a glob.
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
@rwjblue mind adding a entry in the addon hooks?

stefanpenner:
@rwjblue 1 legit failure:

```
  1) jshint should pass for working directory:
     Identifier 'Blueprint_filesPath' is not in camel case. (W106)
 at (/home/travis/build/ember-cli/ember-cli/lib/models/blueprint.js:292:0)
```

the other is upstream ...

rwjblue:
Fixed JSHint failure, rebased, and pushed.  I believe it should pass since you fixed the upstream failures.

rwjblue:
Coveralls job seems stalled, but actually finished successfully: https://coveralls.io/builds/3560411

PR #4838, opened 2015-09-13
[DOCS] Add `npm test` to Addon README

elwayman02:
The default blueprint aliases `npm test` to `ember try:testall` rather than `ember test`, yet the README only mentions running `ember test` and `ember test --server` to test the addon.


stefanpenner:
thanks, good catch!

PR #4839, opened 2015-09-13
[FEATURE] addAddonsToProject for blueprints

elwayman02:
This adds the `addAddonsToProject` method to mirror the same aggregate functionality provided by `addPackagesToProject` and `addBowerPackagesToProject`
Additionally, refactors `addAddonToProject` to leverage the new method
Also adds `_writeStatusToUI` to `addBowerPackagesToProject` to bring it to parity with the npm/addon versions

Here's an example blueprint you can use to test `addAddonsToProject`:

```
module.exports = {
  description: 'Set up additional packages for Ember apps',
  normalizeEntityName: function() {}, // no-op since we're just adding dependencies

  afterInstall: function(options) {
    return this.addAddonsToProject({
      packages: [
        'ember-cli-blanket',
        'ember-cli-sass-pods',
        'ember-cpm',
        'ember-feature-flags',
        'ember-metrics',
        'ember-moment'
      ]
    });
  }
};
```

I chose several common addons that have various blueprint actions in their afterInstall so that it's clear each blueprint is being executed as expected. Additionally, you may pass objects in the `packages` array to indicate version compatibility or include it in the string itself.

```
module.exports = {
  description: 'Set up additional packages for Ember apps',
  normalizeEntityName: function() {}, // no-op since we're just adding dependencies

  afterInstall: function(options) {
    return this.addAddonsToProject({
      packages: [
        'ember-cli-blanket',
        'ember-cli-sass-pods',
        'ember-cpm',
        { name: 'ember-feature-flags', target: '1.1.0' },
        'ember-metrics',
        'ember-moment@3.4.0'
      ]
    });
  }
};
```

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -855,7 +855,7 @@ Blueprint.prototype._locals = function(options) {
 };
 
 /**
-  Used to add a package to the projects `package.json`.
+  Used to add a package to the project's `package.json`.
 
   Generally, this would be done from the `afterInstall` hook, to
   ensure that a package that is required by a given blueprint is
@@ -877,7 +877,7 @@ Blueprint.prototype.addPackageToProject = function(packageName, target) {
 };
 
 /**
-  Used to add multiple packages to the projects `package.json`.
+  Used to add multiple packages to the project's `package.json`.
 
   Generally, this would be done from the `afterInstall` hook, to
   ensure that a package that is required by a given blueprint is
@@ -891,8 +891,8 @@ Blueprint.prototype.addPackageToProject = function(packageName, target) {
   @return {Promise}
 */
 Blueprint.prototype.addPackagesToProject = function(packages) {
-  var task         = this.taskFor('npm-install');
-  var install      = (packages.length > 1) ? 'install packages' : 'install package';
+  var task = this.taskFor('npm-install');
+  var installText = (packages.length > 1) ? 'install packages' : 'install package';
   var packageNames = [];
   var packageArray = [];
 
@@ -908,7 +908,7 @@ Blueprint.prototype.addPackagesToProject = function(packages) {
     packageArray.push(packageNameAndVersion);
   }
 
-  this._writeStatusToUI(chalk.green, install, packageNames.join(', '));
+  this._writeStatusToUI(chalk.green, installText, packageNames.join(', '));
 
   return task.run({
     'save-dev': true,
@@ -918,7 +918,7 @@ Blueprint.prototype.addPackagesToProject = function(packages) {
 };
 
 /**
-  Used to remove a package from the projects `package.json`.
+  Used to remove a package from the project's `package.json`.
 
   Generally, this would be done from the `afterInstall` hook, to
   ensure that any package conflicts can be resolved before the
@@ -935,7 +935,7 @@ Blueprint.prototype.removePackageFromProject = function(packageName) {
 };
 
 /**
-  Used to remove multiple packages from the projects `package.json`.
+  Used to remove multiple packages from the project's `package.json`.
 
   Generally, this would be done from the `afterInstall` hook, to
   ensure that any package conflicts can be resolved before the
@@ -948,8 +948,8 @@ Blueprint.prototype.removePackageFromProject = function(packageName) {
   @return {Promise}
 */
 Blueprint.prototype.removePackagesFromProject = function(packages) {
-  var task         = this.taskFor('npm-uninstall');
-  var install      = (packages.length > 1) ? 'uninstall packages' : 'uninstall package';
+  var task = this.taskFor('npm-uninstall');
+  var installText = (packages.length > 1) ? 'uninstall packages' : 'uninstall package';
   var packageNames = [];
   var packageArray = [];
 
@@ -961,7 +961,7 @@ Blueprint.prototype.removePackagesFromProject = function(packages) {
     packageArray.push(packageNameAndVersion);
   }
 
-  this._writeStatusToUI(chalk.green, install, packageNames.join(', '));
+  this._writeStatusToUI(chalk.green, installText, packageNames.join(', '));
 
   return task.run({
     'save-dev': true,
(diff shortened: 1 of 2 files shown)


elwayman02:
1__bash

stefanpenner:
SG

@mike-north got cycles to review?

elwayman02:
Failing tests are from master afaict

stefanpenner:
> Failing tests are from master afaict

ya, i think we are getting trolled by something upstream https://github.com/npm/fstream/issues/46

mike-north:
this needs at the very least a test case that touches on the difference between `source` and `name`. We ran into some annoying problems with an incomplete `addBowerPackageToProject` implementation, and it has been quite a headache to fix and support

elwayman02:
@mike-north what is `source`? As far as I can tell, that's not even a parameter currently supported by the existing `addAddonToProject`. I'm not sure that's in scope for this feature, but I'm happy to address it to prevent future headaches if you can explain to me what I'm looking for.

elwayman02:
`addPackagesToProject` also does not support a `source` param, for the record.

mike-north:
Ok, upon further review, looks like npm doesn't support source to customize the name of a library in the node_modules folder [like bower allows](https://github.com/bower/endpoint-parser).

lib/models/blueprint.js
```diff
@@ -1046,29 +1049,57 @@ Blueprint.prototype.addBowerPackagesToProject = function(packages, installOption
...
+      }
 
-  return task.run(taskOptions);
+      return pkg.name;
+    });
+  } else {
+    throw new SilentError('You must provide package to addAddonsToProject');
+  }
+
+  var install = (packages.length > 1) ? 'install addons' : 'install addon';
```
mike-north: confusing variable name

tests/unit/models/blueprint-test.js
```diff
@@ -1046,6 +1046,38 @@ describe('Blueprint', function() {
...
+      blueprint.addAddonToProject('foo-bar');
+    });
+
+    it('passes a packages array with target for addAddonsToProject', function() {
+      blueprint.addAddonsToProject = function(options) {
+        expect(options.packages).to.deep.equal([{name: 'foo-bar', target: '^123.1.12'}]);
+      };
+
+      blueprint.addAddonToProject({name: 'foo-bar', target: '^123.1.12'});
+    });
```
mike-north: Please add test coverage for a git tag or wildcard target

lib/models/blueprint.js
elwayman02: Agreed, this is the existing pattern from addPackagesToProject. I can refactor all usages to something else.

tests/unit/models/blueprint-test.js
elwayman02: @mike-north I'm not sure what that would accomplish.  I could literally pass *anything* in target and the result would be the same.

For example: 
```
it('passes a packages array with target for addAddonsToProject', function() {
      blueprint.addAddonsToProject = function(options) {
        expect(options.packages).to.deep.equal([{name: 'foo-bar', target: '^123.1.12'}]);
      };

      blueprint.addAddonToProject({name: 'foo-bar', target: 'banana pudding pops'});
    });
```

addAddonToProject/addAddonsToProject aren't responsible for sanitizing the inputs, any more than addPackage(s)ToProject. This is something that is (or should be) handled in a more centralized place downstream, probably in `addon-install` or add`npm-task`.

elwayman02:
@stefanpenner @mike-north Please let me know if there are any remaining unaddressed issues for this PR or if we can get this merged, thanks! I just rebased with master again, so it should be up-to-date now.

elwayman02:
I've created an issue for Mike's comment regarding test coverage of targets, since that shouldn't be handled as part of this PR.

lib/models/blueprint.js
```diff
@@ -892,7 +892,7 @@ Blueprint.prototype.addPackageToProject = function(packageName, target) {
 */
 Blueprint.prototype.addPackagesToProject = function(packages) {
   var task         = this.taskFor('npm-install');
-  var install      = (packages.length > 1) ? 'install packages' : 'install package';
+  var installText      = (packages.length > 1) ? 'install packages' : 'install package';
```
rwjblue: Can you fix the indentation here?  I personally would prefer a single space, but if we are going to keep the alignment the `=` should be lined up...

lib/models/blueprint.js
```diff
@@ -949,7 +949,7 @@ Blueprint.prototype.removePackageFromProject = function(packageName) {
 */
 Blueprint.prototype.removePackagesFromProject = function(packages) {
   var task         = this.taskFor('npm-uninstall');
-  var install      = (packages.length > 1) ? 'uninstall packages' : 'uninstall package';
+  var installText      = (packages.length > 1) ? 'uninstall packages' : 'uninstall package';
```
rwjblue: indentation tweaks

lib/models/blueprint.js
```diff
@@ -1058,29 +1061,57 @@ Blueprint.prototype.addBowerPackagesToProject = function(packages, installOption
...
+  @return {Promise}
+*/
+Blueprint.prototype.addAddonsToProject = function (options) {
+  var taskOptions = {
+    'packages': [],
+    extraArgs: options.extraArgs || [],
+    blueprintOptions: options.blueprintOptions || {}
+  };
+
+  var packages = options['packages'];
```
rwjblue: Can you swap this to `options.packages` instead?

lib/models/blueprint.js
```diff
@@ -1058,29 +1061,57 @@ Blueprint.prototype.addBowerPackagesToProject = function(packages, installOption
...
-    taskOptions.extraArgs        = options.extraArgs || [];
-    taskOptions.blueprintOptions = options.blueprintOptions || {};
-  }
+  @method addAddonsToProject
+  @param {Object} options
+  @return {Promise}
+*/
+Blueprint.prototype.addAddonsToProject = function (options) {
+  var taskOptions = {
+    'packages': [],
```
rwjblue: Why does `packages` need to be quoted here (might have been that way before, but seems slightly odd)?

rwjblue:
I left a few nit-picky comments, but this looks good to me.  I believe that the test failures will be addressed by a rebase...

lib/models/blueprint.js
elwayman02: Yea, was following an existing pattern. I don't understand why it was done this way, my initial guess was that someone got confused between `package` and `packages` as to which was a reserved keyword in JS?

lib/models/blueprint.js
elwayman02: I'd also prefer a single space...will just fix to that.

elwayman02:
@rwjblue addressed comments and rebased

elwayman02:
Running the tests on my local box times out on a seemingly unrelated `ember new` test. All other tests seem to be passing except that one, but if I run on master they all succeed. Trying again on my branch to see if I can reproduce. Any idea what might have caused that?

elwayman02:
Hmm tests complete for me locally on second attempt. Might have to re-trigger the CI builds though, if they timeout.

elwayman02:
:tada:

raido:
Super!

PR #4840, opened 2015-09-13
Move test helper into addon

### LINKED ISSUE #4200: Move boilerplate test code into an addon.
These files should be brought in via an addon, and not generated in each app:

* `tests/helpers/start-app.js`
* `tests/helpers/resolver.js`
* `tests/test-helper.js`

lazybensch:
resolves #4200
upstream [ember-cli-test-helper](https://github.com/lazybensch/ember-cli-test-helper)


stefanpenner:
not having startApp at a specific module path is going to break many apps.

At the very least, we should add a module at that location, that logs a warning but then re-exports startApp from the addon.

lazybensch:
ember-cli-test-helper is merging those files back into the app tree. so the module path does not change for users. only those files, *that sit there and no one ever touches anyway,* are now brought in by the addon.

so you still import them with:
```js
import startApp from 'your-app/tests/helpers/start-app';
```

I hope i interpreted the intention of the original issue correctly (@rwjblue). Although it would be nice to serve them as utility functions from an addon, most of them depend on `Application` and the apps `config` to be available.

As a nice-to-have those methods are still importable directly from the addon:

```js
import { startApp, destroyApp } from 'ember-cli-test-helper';
```
only that in this scenario the signature of startApp changes to take `Application` as first, `config` as second and `attrs` as third parameter. [as seen here](https://github.com/lazybensch/ember-cli-test-helper/blob/master/app/tests/helpers/start-app.js).

blueprints/app/files/package.json
```diff
@@ -31,12 +32,12 @@
     "ember-cli-qunit": "^1.0.1",
     "ember-cli-release": "0.2.3",
     "ember-cli-sri": "^1.0.3",
+    "ember-cli-test-helper": "0.0.3",
     "ember-cli-uglify": "^1.2.0",
     "ember-data": "2.0.0",
     "ember-disable-proxy-controllers": "^1.0.0",
+    "ember-export-application-global": "^1.0.3",
     "ember-export-application-global": "^1.0.4",
```
quaertym: Double entry?

lazybensch: good catch! addressed by #4863

rwjblue:
Can you rebase and squash?

rwjblue:
@mmun - We need to account for this in our plans with https://github.com/switchfly/ember-test-helpers/pull/129.

lazybensch:
i plan on revisiting this beginning on next week, sorry for abandoning the PR like that :)

Turbo87:
@lazybensch any news on this?

Turbo87:
`ember-cli-test-helpers` might not be a good name since it's already taken on NPM by https://github.com/toranb/ember-cli-test-helpers

lazybensch:
@Turbo87 rebased the branch and started fixing some convergence issues with master. CI is failing although tests are all green locally for me. atm i'm trying to reproduce the issue travis has.

lazybensch:
the addon i moved the code to is `ember-cli-test-helper` (singular) to avoid naming collision.

CI fails because when executing `import { startApp } from 'ember-cli-test-helper';` ([here](https://github.com/lazybensch/ember-cli-test-helper/blob/master/app/tests/helpers/start-app.js#L1)) it does not seem to get the correct function back. 

That causes a `TypeError: 'undefined' is not a function (evaluating '_emberCliTestHelper.startApp.bind(this, _someCoolAppApp['default'], _someCoolAppConfigEnvironment['default'])')` later on.

I can not, for the life of me, reproduce this issue. When i run the tests they all pass. If anyone has an idea i'd be happy about a hint.

lazybensch:
ok the branch is up to date now and test suite is green (except for the appveyor tests that fail on master as well)

If this PR is fine with you i will give you github and npm ownership of the `ember-cli-test-helper` addon so that we can merge this in then.

Turbo87:
how does this PR play with ember-cli-mocha?

lazybensch:
the only clash between those is in [test-helper.js:2](https://github.com/switchfly/ember-cli-mocha/blob/master/blueprints/ember-cli-mocha/files/tests/test-helper.js#L2) but this file will be generated (overridden) by the ember-cli-mocha blueprint anyway, right?

And the hole [module-for-acceptance.js](https://github.com/lazybensch/ember-cli-test-helper/blob/b2abc838eebc1d8998715470387d3e63395e5f21/app/tests/helpers/module-for-acceptance.js#L1) file depends on qunit but if you use ember-cli-mocha you'll just never use its exported method.

besides those points it should play with ember-cli-mocha as good as a vanilla ember-cli build project would.

nathanhammond:
@lazybensch ping

Turbo87:
this will likely be replaced by https://github.com/emberjs/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-acceptance.js soon

PR #4842, opened 2015-09-13
[Safe for 2.1] Update initializer blueprint.

nathanhammond:
Fixes the test blueprint to invoke the initializer in a way such that `this` is set. See: https://github.com/emberjs/ember.js/pull/10179


stefanpenner:
these changes, will need to be part of the ember addonification effort. And as such will need to be part of the final extracted blueprints, that end up living in the ember repo itself.

This will ensure the blueprints are correctly entangled to the version of ember installed.

rwjblue:
@stefanpenner - We should still update them here, so that we have an up to date version that gets extracted.

rwjblue:
@homu r+

rwjblue:
@homu retry

nathanhammond:
This commit is still good and should still land.

rwjblue:
@homu retry

PR #4843, opened 2015-09-13
adds travis build matrix

makepanic:
Changes `travis.yml` to use the [node](http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Choosing-Node-versions-to-test-against) [build matrix](http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix).
This removes the manual nvm, npm and node installation. In addition the manual version logging is removed because travis does that automatically in a `node_js` language config ([example](https://travis-ci.org/lovell/sharp/jobs/79881738#L41)).

The change also includes node 4, iojs-v2 and iojs-v3 as node test version.

The updated `.travis.yml` will still fail to build because there exiting test errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,18 @@
+language: node_js
 sudo: false
-env:
-  - NODE_VERSION=0.10
-  - NODE_VERSION=0.12
-  - NODE_VERSION=iojs
+node_js:
+  - "0.10"
+  - "0.12"
+  - "iojs-v2"
+  - "iojs-v3"
+  - "4"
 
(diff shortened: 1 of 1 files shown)


stefanpenner:
the matrix was changed, to support OSX builds. We disabled OSX CI builds, because they ended up being brittle. I would like to bring them back.

I am unsure if travis has improved support on the OSX.

@drogus  does the language feature work on OSX ?

stefanpenner:
closing, will reconsider when OSX supports this matrix

PR #4846, opened 2015-09-14
Update ember-cli-htmlbars-inline-precompile dependency

joliss:


stefanpenner:
could not replicate windows failure locally

PR #4849, opened 2015-09-15
cleanup whitespace in the commands

kellyselden:
started with a few eyesores, ended syncing up all the options' layouts.


stefanpenner:
LVGTM

stefanpenner:
canceling tests, since whitespace should be fine, and several of them already passed. This will prevent this run from clogging up CI

PR #4851, opened 2015-09-15
proxy _getPort as it was only used for stubbing

kellyselden:
use proxy to avoid exposing _getPort


stefanpenner:
failing test should be dealt with: https://github.com/ember-cli/ember-cli/commit/7bef5d9ad00059fb0468b294f4802d99c23e68d4

PR #4857, opened 2015-09-15
Only exclude node_modules at root

joliss:
We have some checked-in node_modules (e.g. at
tests/fixtures/preprocessor-tests/app-with-addon-with-preprocessors/node_modules)
that we don't want to ignore.

Also adding slashes to `docs/build` and `coverage` for good measure. (I believe these only ever exist at the root as well, yes?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
-node_modules
+/node_modules
 npm-debug.log
 tmp*
 _posts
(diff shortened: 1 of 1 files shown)


stefanpenner:
Ug failure is a node 4 thing, unrelated to this. I'll take a look this morning

Issue #4872, opened 2015-09-18
Option to run tests on previous build


trentmwillis:
Was wondering why there is no option to run tests on an existing build of the application (something like `ember test --no-build`). Specifically, here are several cases I can think of where this would be desirable:

1. You've been developing in browser and then switch to make sure tests are passing in Phantom,
2. Your tooling pipeline validates a build and then runs tests as two separate tasks, and
3. Attempting to parallelize or distribute tests based off a single build

It seems this shouldn't be too difficult to implement, but maybe I'm missing something. I'd be glad to take a stab at it if this sounds useful.

stefanpenner:
something like:

```sh
ember test --path path/to/build
```

would be nice

trentmwillis:
I feel the `query` option is strictly tied to this, but I think that could be a useful new addition to make. Should we have a command to alias `path` to the default build location?

stefanpenner:
> Should we have a command to alias path to the default build location?

I don't think thats needed.

Please note, I suspect you believe the default build location to be `dist` but in-fact, dist is just a dereferenced copy of the actual build output. It exists, to make it easy for users to see what is going on.

By default tests are run against something deep in `tmp`

trentmwillis:
> dist is just a dereferenced copy of the actual build output.

Even in the case where `ember build` was run? That's the scenario I was envisioning for the last two use cases I mentioned above.

stefanpenner:
> Even in the case where ember build was run?

yup, dist is 100% of the time just a copy of the internal. This is because we wanted to allow for concurrent builds, without interfering.

`--path` or w/e we name it, will basically override that.

related code: https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L153

trentmwillis:
Okay, good to know. I'll probably submit `--path` and `--query` as two separate PRs. Is it going to be possible to make this compatible with the `--server` option? I feel like the answer is no.

trentmwillis:
Closed via #4874 and #4906

PR #4874, opened 2015-09-18
Add ability to specify a build path for running tests

trentmwillis:
Addresses #4872. Adds the `--path` option to allow you to run the tests from a previous build of the application.

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -1,11 +1,13 @@
 'use strict';
 
-var Command = require('../models/command');
-var Watcher = require('../models/watcher');
-var Builder = require('../models/builder');
-var fs      = require('fs');
-var path    = require('path');
-var win     = require('../utilities/windows-admin');
+var Command     = require('../models/command');
+var Watcher     = require('../models/watcher');
+var Builder     = require('../models/builder');
+var SilentError = require('silent-error');
+var fs          = require('fs');
+var path        = require('path');
+var win         = require('../utilities/windows-admin');
+var existsSync  = require('exists-sync');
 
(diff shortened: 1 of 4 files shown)


stefanpenner:
I like where this is going. Instead of using run command, can you use the ember helper instead? It does its thing much faster. Let me know if you have questions

kellyselden:
The ember test helper currently has a problem generating the artifacts after build, @trentmwillis by all means try using it on both your calls, but you might only get it working on the second instance, needing to leave runCommand on the first.

stefanpenner:
Ah. We should figure out what and fix it

trentmwillis:
@kellyselden seems to be working for me. Only issue is that the output string (`result.ui.output`) doesn't seem to contain any information from the test run so the assertions fail. I can see the tests ran in the console output though, thoughts?

kellyselden:
@trentmwillis I don't have any ideas without a deeper dive.

trentmwillis:
Okay, I'll keep looking into it. It seems that the testem/phantom output isn't written to the output stream that is fed into the UI.

trentmwillis:
Updated to use the `ember` helper. Apparently the `MockUI` used under the hood for the `ember` command doesn't capture all writes to `stdout` which is why it wasn't working before. Left a todo note to fix it later.

trentmwillis:
@stefanpenner @kellyselden I believe I got all the tests working for this but AppVeyor seems to be having a hard time. May need to be restarted.

kellyselden:
restarted appveyor

stefanpenner:
@chadhietala please use homu

PR #4880, opened 2015-09-19
consolidate test setup to not use fixtures, instead use mocking

kellyselden:


kellyselden:
Another side effect is this makes it a faster test.

PR #4888, opened 2015-09-22
Introduce the `vendor-shim` blueprint

seanpdoyle:
There isn't much documentation around how to properly wrap an otherwise
global `window` dependency in an ES6 compliant module.

The generated blueprint will nudge consumers in the correct direction.

Insprired by [this stack overflow response][shim].

[shim]: http://stackoverflow.com/questions/30443351/import-dependencies-in-ember-cli-e-g-import-math-js/30584671#30584671


seanpdoyle:
I wasn't sure how to cover this with tests, so I'd love some guidance.

seanpdoyle:
@stefanpenner @jakecraige is something like this even a good idea?

seanpdoyle:
@homu rebased.

seanpdoyle:
@stefanpenner @abuiles could either of you give some guidance on how to add test coverage for a blueprint, so that coveralls doesn't fail CI?

seanpdoyle:
@stefanpenner this is green on Travis. I'd love some feedback on what its missing.

blueprints/vendor-shim/files/__name__.js
```diff
@@ -0,0 +1,9 @@
+(function() {
+  function vendorModule() {
+    'use strict';
+
+    return { 'default': window['<%= name %>'] };
```
stefanpenner: `self` not `window`

stefanpenner:
> @stefanpenner @jakecraige is something like this even a good idea?

I'm unsure, we are actively working to remove the need for shims entirely. Although there will always be legacy code that likely needs it. So I am torn.

tests/acceptance/help-test.js
```diff
@@ -752,6 +752,13 @@ ember version \u001b[36m<options...>\u001b[39m' + EOL + '\
                   overridden: false
                 },
                 {
+                  name: 'vendor-shim',
+                  description: 'Generates an ES6 module shim.',
```
stefanpenner: for a global

stefanpenner:
I think  my concerns will be removed, if there was a nice little snippet in the docs/website that described this process with an example.

blueprints/vendor-shim/files/vendor/__name__-shim.js
```diff
@@ -0,0 +1,9 @@
+(function() {
```
seanpdoyle: @stefanpenner is this file named properly for the following to work?

```bash
# we want to generate `vendor/moment-shim.js`
$ ember generate vendor-shim moment
```

Additionally, is there an API to add `app.import('vendor/moment-shim.js')` to the `ember-cli-build.js`?

rwjblue: Unsure if related, but we should generally not generate individual files in `vendor/` (they should be in subdirectories).

https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1372

seanpdoyle: @rwjblue that's fine.

Would `vendor/shims/moment.js` be a good auto-generated name?

rwjblue: Seems fine to me. The `vendor/some-file.js` just triggered flash backs for me :stuck_out_tongue_closed_eyes:

seanpdoyle: :ok_hand: @rwjblue.

Renamed and documented in https://github.com/ember-cli/ember-cli/pull/4935

seanpdoyle:
@stefanpenner what do you think about https://github.com/ember-cli/ember-cli/pull/4935?

stefanpenner:
@homu r+

PR #4890, opened 2015-09-23
Allow content to be mutated in contentFor.

nathanhammond:
This allows me to register my addon as `after` another addon and to modify the output generated by any previous addon. Results in coupling, but keeps it pretty loose. Current use-case is modifying the output of `ember-cli-qunit` in my addon to not include any of the `test-body` content.

```javascript
{
  contentFor: function(type, config, content) {
    if (type !== 'test-body') { return; }

    for (var i = 0; i < content.length; i++) {
      if (~content[i].toString().indexOf('ember-testing-container')) {
        content = content.splice(i,1);
      }
    }
  }
}
```

(You're probably noticing a theme in my PRs...)

### DIFF
--- a/ADDON_HOOKS.md
+++ b/ADDON_HOOKS.md
@@ -373,6 +373,7 @@ Allow addons to implement contentFor method to add string output into the associ
 
 - type
 - config
+- content
 
 **Source:** [lib/broccoli/ember-app.js:1167](https://github.com/ember-cli/ember-cli/blob/v0.1.15/lib/broccoli/ember-app.js#L1167)
 
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1438,7 +1438,7 @@ EmberApp.prototype.contentFor = function(config, match, type) {
   }
 
   content = this.project.addons.reduce(function(content, addon) {
-    var addonContent = addon.contentFor ? addon.contentFor(type, config) : null;
+    var addonContent = addon.contentFor ? addon.contentFor(type, config, content) : null;
     if (addonContent) {
       return content.concat(addonContent);
     }
--- a/tests/unit/broccoli/ember-app-test.js
+++ b/tests/unit/broccoli/ember-app-test.js
@@ -167,6 +167,35 @@ describe('broccoli/ember-app', function() {
 
         expect(actual).to.equal('blammo\nblahzorz');
       });
+
+      it('allows later addons to inspect previous content', function() {
+        var calledContent;
+
+        project.addons.push({
+          contentFor: function() {
+            return 'zero';
+          }
+        });
+
+        project.addons.push({
+          contentFor: function() {
+            return 'one';
+          }
+        });
+
+        project.addons.push({
+          contentFor: function(type, config, content) {
+            calledContent = content.slice();
+            content.pop();
+            return 'two';
+          }
+        });
+
+        var actual = emberApp.contentFor(config, defaultMatch, 'foo');
+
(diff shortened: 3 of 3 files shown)


stefanpenner:
unsure about this, but it def needs tests.

nathanhammond:
The problem is basically that once an addon supplies content via `contentFor` it's guaranteed to end up in the output. There's nothing any other addon (even registering `after` it) can do to change that. This makes my addon impossible to build as it is dependent on `ember-cli-qunit`, but needs to have the ability to run tests without modifying the `<body>` markup.

Writing tests now. :)

stefanpenner:
@nathanhammond is it possible that ember-cli-qunit needs to be more configurable

nathanhammond:
@stefanpenner We can't guarantee that all addon maintainers are going to agree on what should and shouldn't be configurable, how certain pieces of code should work together, or even like each other. This is the escape hatch. I think of it as a bit like rack middlewares.

I also don't believe it's sustainable for every single addon to request configuration options from all of the other addons it has to integrate with. Permutations, especially for core libraries, will grow exponentially as our addon library grows.

nathanhammond:
New push has tests. :)

rwjblue:
> it possible that ember-cli-qunit needs to be more configurable

Agreed.  We already have some configuration (to disable container styles from being appended), would just need another option for this.

> I also don't believe it's sustainable for every single addon to request configuration options from all of the other addons it has to integrate with. 

I agree in principle, but this isn't "every addon". This is a single addon that is added by default by Ember CLI.  Addons that you as a user choose are somewhat different than those that are part of the core out of the box experience.

That being said, I am definitely :+1: on exposing a way to override a prior `contentFor` result (I just think that it should be the last resort instead of the first).

stefanpenner:
> I also don't believe it's sustainable for every single addon to request configuration

The counter argument is, is it sustainable to monkey patch output from another add-on? The implicit protocol to enable monkey patching enables short-term wins (maybe buying some time), but in the long-term these things do fester and can easily result high risk.

nathanhammond:
My opinion remains that this PR should land. Tests pass. :) You can have your addon specify a peer dependency that guarantees compatibility/has NPM yell at you which will make it less fragile if you depend on checking output.

I don't disagree that this should be last resort. I'm also in the process of modifying https://github.com/ember-cli/ember-cli-qunit/blob/master/index.js as well to provide a few new configuration options and have all test-related assets inserted into `tests/index.html` using `contentFor` (read: making their presence configurable).

nathanhammond:
> is it possible that ember-cli-qunit needs to be more configurable

In trying to make `ember-cli-qunit` more configurable there's one pretty big problem: only `included` is called with `(app, parentAddon)`. All of the rest are simply invoked. This means that we don't currently have the ability to read options except inside of the `included` hook (without knowing method invocation order and mutating `this`).

```javascript
{
  included: function included(app, parentAddon) {
    var target = (parentAddon || app);
    this._super.included(target);

    this.options = target.options['ember-cli-qunit'];
    /* SNIP */
  },
  contentFor: function(type, config, content) {
    // this.options is available here, but I only know that because I traced it.
  }
}
```

Thoughts? (In short: our configuration story probably isn't good enough yet.)

stefanpenner:
> All of the rest are simply invoked.

Yes, this a larger issue leaking in.

nathanhammond:
Can we restart appveyor and possibly land this?

nathanhammond:
AppVeyor is finally happy. :)

stefanpenner:
@homu r+

PR #4895, opened 2015-09-23
move help output tests from acceptance to unit

kellyselden:
This moves almost all help acceptance tests into the unit test category. This is only the first pass. I plan to get more fine grained to find more testing gaps and speed gains.

This should shave some time off of CI.


stefanpenner:
Lvgtm

kellyselden:
The coverage decrease should be fixed in my coming PRs. I vote for merging now so the next diff isn't as extreme, @stefanpenner ?

stefanpenner:
> I vote for merging now so the next diff isn't as extreme, @stefanpenner ?

:+1:

kellyselden:
:smile:

PR #4899, opened 2015-09-26
Make the ExpressServer object re-emit listening.

nathanhammond:
It's presently impossible to attach events to the express-server that is generated, though you can get a reference to our wrapper of it. Have our wrapper re-emit the `listening` event from the server.


stefanpenner:
@homu r+

PR #4900, opened 2015-09-27
help tests cleanup

kellyselden:
* increasing coverage
* removing redundant tests
* fixed bug when printing a single command using json
* fixed bug when printing an options type using json


stefanpenner:
Lvgtm

PR #4906, opened 2015-09-29
Add `--query` option to test command to allow passing arbitrary query params

trentmwillis:
As discussed in #4872, this adds an option to the test command to allow user's to pass in arbitrary query parameters to the test page URL.

Example:

```bash
ember test --query='nojshint'
```

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -27,7 +27,8 @@ module.exports = Command.extend({
     { name: 'launch',      type: String,  default: false,                            description: 'A comma separated list of browsers to launch for tests.' },
     { name: 'reporter',    type: String,                            aliases: ['r'],  description: 'Test reporter to use [tap|dot|xunit]' },
     { name: 'test-page',   type: String,                                             description: 'Test page to invoke' },
-    { name: 'path',        type: String,                                             description: 'Path to a build to run tests against.' }
+    { name: 'path',        type: String,                                             description: 'Path to a build to run tests against.' },
+    { name: 'query',       type: String,                                             description: 'A query string to append to the test page URL.' }
   ],
 
   init: function() {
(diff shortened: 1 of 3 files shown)


rwjblue:
Looks good.  

@trentmwillis - Can you update [ember-cli-mocha here](https://github.com/switchfly/ember-cli-mocha/blob/master/index.js#L15-L27) with the same tweak to `buildTestPageQueryString`?

trentmwillis:
@rwjblue sure thing

stefanpenner:
I'm assuming adding multiple queries also works? I don't see that in the tests 

Otherwise looks good

trentmwillis:
@stefanpenner I guess that would depend on what you want it to look like. Assuming you add multiple queries like `--query='nojshint&hidepassed&something=else'`, then yes it works. Is there a better syntax?

stefanpenner:
> @stefanpenner I guess that would depend on what you want it to look like. Assuming you add multiple queries like --query='nojshint&hidepassed&something=else', then yes it works. Is there a better syntax?

that looks great to me.

trentmwillis:
Okay cool. I'll add a test for that and update.

trentmwillis:
@stefanpenner added the test for multiple queries. Looks like AppVeyor needs restarting again.

chadhietala:
@trentmwillis restarted.

trentmwillis:
Well, we've restarted a couple times but AppVeyor is still red. Not sure what to do about it as I'm fairly certain these changes aren't the root cause

stefanpenner:
@homu r?

trentmwillis:
@stefanpenner looks like homu didn't properly run

stefanpenner:
@homu r+

chadhietala:
@homu r+

Issue #4908, opened 2015-09-29
No support for testem.js


asakusuma:
While trying to define [testem.js instead of testem.json](https://github.com/airportyh/testem/blob/master/docs/config_file.md), I ran into some errors that seem to indicate that the testem.js wasn't getting picked up. I did some digging and it appears that ember-cli [assumes that testem.json is used](https://github.com/airportyh/testem/blob/master/docs/config_file.md
https://github.com/ember-cli/ember-cli/blob/master/lib/commands/test.js#L59). We probably need to pass the `test_page` and `launch` parameters directly into the testem api instead of re-writing testem.json.

I'm planning on putting together a PR.

I need to define testem.js so I can define a [custom reporter](https://github.com/airportyh/testem/blob/master/docs/custom_reporter.md).

stefanpenner:
> I'm planning on putting together a PR.

:+1:

asakusuma:
@stefanpenner @chadhietala ran into a bit of a pickle and want ya'lls opinions.

When ember-cli invokes testem, it [modifies cwd](https://github.com/ember-cli/ember-cli/blob/058528977255d0e366b8a1760668875cfadf9067/lib/tasks/test.js#L42). This doesn't affect loading testem.json, because testem uses `fs` to read json configs. However, testem relies on `cwd` to to resolve testem.js config files, since it has to use `require`. ember-cli modifies `cwd` to be the build directory, so testem ends up trying to load the dummy [testem.js](https://github.com/ember-cli/ember-cli/blob/058528977255d0e366b8a1760668875cfadf9067/lib/broccoli/testem.js) file which happens to be in the output directory.

Right now I see 4 options:

1. Figure out a way to make ember-cli work without modifying the testem cwd
2. Have ember manually resolve a testem config file, and if it happens to be a javascript file, provide an absolute path
3. Move the existing testem.js (for running testem in the browser) to a different filepath and copy /testem.js to `outputPath`/testem.js
4. Open a PR in testem to ignore the passed in cwd with regards to loading configs

From my limited vantage point, I think option 1 would be ideal, but I'm going to guess that option 2 is the most realistic. Thoughts?

stefanpenner:
@johanneswuerbach thoughts?

asakusuma:
If we go the testem PR route, I think that would be as simple as creating a version of `resolvePath()` that uses process.cwd instead of the passed in cwd, and then use this function for resolving config paths:

[...]

PR #4912, opened 2015-10-01
Full testem config file support

asakusuma:
Adds support for any kind of testem config, includin `testem.js`

Before, ember-cli assumed `testem.json`. Now, we're letting testem do the config file resolution.

Fixes https://github.com/ember-cli/ember-cli/issues/4908

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -4,7 +4,6 @@ var Command     = require('../models/command');
 var Watcher     = require('../models/watcher');
 var Builder     = require('../models/builder');
 var SilentError = require('silent-error');
-var fs          = require('fs');
 var path        = require('path');
 var win         = require('../utilities/windows-admin');
 var existsSync  = require('exists-sync');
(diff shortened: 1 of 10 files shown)


asakusuma:
@stefanpenner the AppVeyor build timed out when it had almost finished. I noticed some of the [passing builds](https://ci.appveyor.com/project/embercli/ember-cli/build/5000) were less than a minute away from timing out, and I did add a smoke acceptance test. All the tests are passing in Travis CI.

Perhaps the AppVeyor timeout limit should be increased?

stefanpenner:
@kellyselden thoughts re: tests?

kellyselden:
I guess your smoke test is the one that broke the camels back in Windows. Try bumping up the timeout from 400000 to 500000, it might help Windows until we mitigate the time to run an acceptance test helper.

stefanpenner:
@kellyselden any reason why this can't use the `ember` test helper?

kellyselden:
@stefanpenner I don't think the `ember` test helper returns the same `result.code`, so the test would have to change. But with some tweaking, the test might be able to assert something similar using `ember`. @asakusuma I would suggest trying it, because if it works you get a faster test.

asakusuma:
@stefanpenner I changed the test to use the ember test helper

@kellyselden Also bumped the timeout to 500000

stefanpenner:
@homu r+

stefanpenner:
thanks asa!

kellyselden:
nice work!

PR #4915, opened 2015-10-01
Update help string for route blueprint/generator

jhr007:
Updated the 'route' description in the blueprint to reflect it also generates a template file.


kellyselden:
The matching tests must be updated, but then :+1:

jhr007:
Done. The tests should be restarted.

jhr007:
Never mind, the tests already kicked off. <spam-apology>

kellyselden:
@homu r+

PR #4916, opened 2015-10-01
Update 2013-04-04-deployments.md

Ducain:
Wording change for clarity. This is subjective of course.


kellyselden:
Sounds better to me. Thanks!

PR #4920, opened 2015-10-02
Update initializer blueprints for Ember 2.1.

dgeb:
Expecting `registry` as an initial argument is now deprecated.


rwjblue:
:+1:

stefanpenner:
@homu r+

nathanhammond:
Initializers are now objects and invoked with a context. We need to import the entire object in the tests, not just the `initialize` function.

See: https://github.com/emberjs/ember.js/pull/10179

nathanhammond:
Updated this PR which we need to land in order to address initializers being called with context: https://github.com/ember-cli/ember-cli/pull/4842

PR #4923, opened 2015-10-02
Merging of hash properties in EmberApp#initOptions fails.

nathanhammond:
This is a failing unit test for merging properties inside of EmberApp#initOptions. We somehow end up duplicating the entire current object as a property.

Haven't traced for the bug yet.

### DIFF
--- a/lib/broccoli/ember-addon.js
+++ b/lib/broccoli/ember-addon.js
@@ -5,7 +5,6 @@
 @module ember-cli
 */
 var defaults = require('merge-defaults');
-var merge    = require('lodash/object/merge');
 var Funnel   = require('broccoli-funnel');
 var EmberApp = require('./ember-app');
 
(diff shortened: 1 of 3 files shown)


nathanhammond:
New push has a fix, but has one tradeoff: properties aren't merged after their first occurrence in the series of objects. Can you explain the behavior you were trying to accomplish with `merge({...}, {...}, defaults);` so I can emulate it?

nathanhammond:
This is good to merge, it fixes the recursive merge problem that comes from mutating the first object. Old behavior and new behavior compared in the jsfiddle.

chadhietala:
@homu r+

stefanpenner:
I think the deep merge stuff was actually intentional. @rwjblue do you recall

nathanhammond:
@stefanpenner The behavior didn't change in this PR, and it still does a deep merge. I had assumed that my first-pass fix would change behavior, but it didn't as shown by the jsfiddle.

PR #4928, opened 2015-10-02
Remove unnecessary configuration for ember-cli-qunit.

nathanhammond:
As a consequence of ember-cli/ember-cli-qunit#82 we can remove configuration from the default `EmberApp` that is relevant only to `ember-cli-qunit`. We now expect you to explicitly declare a setting as `false` or `true` in order to opt in or out of certain behaviors. `undefined` will be treated as the default behavior for that setting.

This is blocked by releasing a new version of ember-cli-qunit.


stefanpenner:
@nathanhammond needs a rebase sir

Turbo87:
is there any reason why this is not merged yet?

PR #4930, opened 2015-10-03
[WIP][FEATURE] Clean command

beatle:
Idea originated from here #4819 

Currently it's an a raw  implementation but it looks like most of work is done.
Currently clean command handles `--skip-npm` and `--skip-bower` flags. 
You can also specify `--output-path` to  customize dist/ folder path. 

Currently I'm covering this feature with tests. Seems like I need advise on how to make it right way.
node_modules/ and tmp/ are not configurable paths. So it's not obvious for me whether it should be tested via acceptance tests or not. 
If yes I'm happy to hear any directions on how to properly isolate project for acceptance test and play with directories inside(node_modules, bower directory, output-path and tmp).


beatle:
Sorry, was inaccurate. a lot of trash in changed files.

PR #4931, opened 2015-10-03
[Feature] Clean command

beatle:
Hi all,

Idea originated from here #4819 

Yeah, it took more time than I could expect.
Nevertheless it looks like now we can use `npm run clean` to clean up our ember apps.
The `ember clean` command handles `--skip-npm` and `--skip-bower` flags.
You can also specify `--output-path` to customize dist/ folder path.

Currently we don't have any acceptance tests covering for npm/bower cache clean functionality.
I've tried to set custom cache directory for `npm` to test `npm cache clean` feature. 
Unfortunately seems like nor `npm.load()` neither `lib/utilities/npm` takes in account `cache` setting in the project's `.npmrc`. under investigation


lib/commands/clean.js
```diff
@@ -0,0 +1,92 @@
...
+module.exports = Command.extend({
+  name: 'clean',
+  description: 'Cleans up the project from downloads(npm/bower), flushes caches and removes build results directories(tmp/ and dist/ by default).',
+  works: 'insideProject',
+  aliases: ['c'],
+
+  availableOptions: [
+    { name: 'dry-run',     type: Boolean, default: false,   aliases: ['d'] },
+    { name: 'verbose',     type: Boolean, default: false,   aliases: ['v'] },
+    { name: 'skip-npm',    type: Boolean, default: true,   aliases: ['sn'] },
```
kellyselden: alignment issue

lib/commands/clean.js
```diff
@@ -0,0 +1,92 @@
...
+      .then(function() {
+        if (!commandOptions.skipNpm) {
+          return npmCacheClean.run({
+            verbose: commandOptions.verbose
+          });
+        }
+      })
+      .then(function() {
+        dirsToRemove.forEach(function(dirToRemove) {
+          rimraf.sync(dirToRemove);
```
beatle: Any reason to not use `rimraf` in command directly? Feel like something wrong with it, but not sure how to deal with it.
Possibly it should be extracted in something like RemoveDirTask?

beatle:
@homu thank you for the pointer. I'll take a look on it. 

It's fun but true, currently the main blocker for me is an obscurity on how to make [help test](https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/help-test.js#L44 ) work.
It tests a raw (nodejs?) output of ember-cli's help command. 

I just can't figure out how to effectively get this output to extract `ember clean` output and insert it to the appropriate test.
 
`ember help > help`
doesn't provide me an output with special characters. 

Smells like it's a newbie question and the answer should be trivial, but currently I've decided to make this test pass first before I can move forward.

@kellyselden @homu any right solution for this?

kellyselden:
Those character codes are generated from the `chalk` package. I would copy the entire segment from another command and fill in your command's values. To get it right, you could temporarily remove the output above and below your command, and then change the `.to.equal(` if there is one to a `.to.include(`, and you should be able to perfect your command output without too much tedium.

beatle:
> To get it right, you could temporarily remove the output above and below your command
Sorry, I'm not sure what do you exactly mean here. Possibly my English is a barrier here :/ 
Currently I just can't figure out how to get a chalk generated output with all the special characters in a plain text presentation after running `ember help`.

Anyway I've eventually changed my priorities. The next step is covering of `clean` command and related tasks with unit tests.

elwayman02:
@beatle thanks for taking up this work, I've been swamped with other projects lately! I'll take a look at this shortly. In the meantime, would you mind squashing your commits down? We probably don't want to merge 25 commits into master for this one feature. Thanks!

lib/commands/clean.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+var path    = require('path');
+var Command = require('../models/command');
+var Promise = require('../ext/promise');
+
+module.exports = Command.extend({
+  name: 'clean',
+  description: 'Cleans up installations, flushes caches and removes build results(tmp/ and dist/ by default).',
```
elwayman02: I would reword this somewhat:

```
Removes installed packages, flushes npm/bower caches and removes build files (tmp/ and dist/).
```

lib/commands/clean.js
```diff
@@ -0,0 +1,92 @@
...
+  works: 'insideProject',
+  aliases: ['c'],
+
+  availableOptions: [
+    { name: 'dry-run',     type: Boolean, default: false,   aliases: ['d'] },
+    { name: 'verbose',     type: Boolean, default: false,   aliases: ['v'] },
+    { name: 'skip-npm',    type: Boolean, default: true,    aliases: ['sn'] },
```
elwayman02: I don't think `skip-npm` should be true by default. Please set this to false.

lib/commands/clean.js
```diff
@@ -0,0 +1,92 @@
...
+      .then(function() {
+        if (!commandOptions.skipBower) {
+          return bowerCacheClean.run({
+            verbose: commandOptions.verbose,
+            dryRun: commandOptions.dryRun
+          });
+        }
+      })
+      .then(function() {
+        if (!commandOptions.skipNpm) {
```
elwayman02: I have a slight preference to run the npm cache clean before the bower cache clean (as that seems to be pretty standard), though in practice it should not particularly matter.

lib/tasks/bower-install.js
```diff
@@ -1,57 +1,19 @@
 'use strict';
 
-// Runs `bower install` in cwd
+// Runs `bower` commands(install, uninstall, cache clean...) in cwd
```
elwayman02: This task only runs install, doesn't it?

lib/tasks/npm-cache-clean.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+// Runs `npm cache clean` in cwd
+
+var NpmTask = require('./npm-task');
+
+module.exports = NpmTask.extend({
+    command: 'cache clean',
+    startProgressMessage: 'Cleaning Npm cache',
+    completionMessage: 'Npm cache cleaned successfully',
```
elwayman02: Prefer `npm` over `Npm`, as that's how npmjs.org references it (in all lower case)

lib/tasks/bower-install.js
elwayman02: Looks like you meant to add this comment to `bower-task`

lib/tasks/npm-task.js
```diff
@@ -37,7 +42,11 @@ module.exports = Task.extend({
     // misplaced console.log
     this.disableLogger();
 
-    return npm(this.command, packages, npmOptions, this.npm).
+    var command = (options.dryRun) ?
```
elwayman02: This code will never execute if `options.dryRun` is true...line 26 above will exit the method. You either want to revert this change (lines 45-49) or revert the above (lines 25-28).

lib/tasks/remove-files.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+// Creates a directory with the name directoryName in cwd and then sets cwd to
+// this directory.
+
+var Promise     = require('../ext/promise');
+var Task        = require('../models/task');
+var rimraf  = require('rimraf');
+var _ = require('lodash');
+var chalk = require('chalk');
```
elwayman02: We should either line up all the `=` or none of them. I lean towards none, but I don't care as long as it's consistent.

lib/tasks/remove-files.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    return Promise.resolve().then(function() {
+        paths.forEach(function(dirToRemove) {
+          if (!options.dryRun) {
+            rimraf.sync(dirToRemove);
+          }
+
+          var withTrailingSlash = ensureWithTrailingSlash(dirToRemove);
+          ui.writeLine(chalk.green('Directory \'' + withTrailingSlash + '\' removed'));
```
elwayman02: Just use double quotes here instead of escaping the single quotes.

lib/utilities/npm.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 var Promise = require('../ext/promise');
+var findCommand = require('./package-manager-command');
 
-// 
+//
```
elwayman02: This line can probably be removed

lib/tasks/remove-files.js
kellyselden: That will probably fail the linter, and therefore probably not worth it.

tests/acceptance/help-test.js
```diff
@@ -314,6 +314,51 @@ ember version \u001b[36m<options...>\u001b[39m' + EOL + '\
           anonymousOptions: []
         },
         {
+          name: 'clean',
+          description: 'Cleans up the project from downloads(npm/bower), flushes caches and removes build results directories(tmp/ and dist/ by default).',
```
elwayman02: Would prefer that this description matches the one used in the actual clean task definition above

tests/acceptance/help-test.js
```diff
@@ -314,6 +314,51 @@ ember version \u001b[36m<options...>\u001b[39m' + EOL + '\
...
+            {
+              name: 'verbose',
+            default: false,
+              aliases: ['v'],
+              key: 'verbose',
+              required: false
+            },
+            {
+              name: 'skip-npm',
+            default: true,
```
elwayman02: See above re: default

elwayman02:
Needs test coverage for new tasks, such as `clean` and `bower-task`. Most likely the existing tests for `bower-install` should be moved to `bower-task` and then tests added for `buildOptions` and `buildArgs`.

lib/commands/clean.js
beatle: As for beginning of phrase - yes.
I use "by default" because at least dist/ path is configurable

lib/tasks/bower-install.js
beatle: Yes, thanks

lib/commands/clean.js
beatle: Have no such preference. Will change it as soon as continue work on it. Hopefully before this weekend

lib/tasks/npm-task.js
beatle: Good catch. This logic extracted from npm task. I've already fixed it locally

lib/tasks/remove-files.js
beatle: Yes. Jshint complains about mixing quotes

elwayman02:
@beatle any updates?

beatle:
Currently finishing appropriate acceptance test. Had some questions which took some time to learn ember-cli testing internals.

lib/tasks/bower-install.js
```diff
@@ -2,56 +2,28 @@
...
-    var ui             = this.ui;
-    var packages       = options.packages || [];
-    var installOptions = options.installOptions || { save: true };
+module.exports = BowerTask.extend({
+  command: 'install',
+  startProgressMessage: 'Installing browser packages via Bower',
+  completionMessage: 'Installed browser packages via Bower.',
 
-    ui.startProgress(chalk.green('Installing browser packages via Bower'), chalk.green('.'));
+  buildOptions: function(options) {
```
beatle: @stefanpenner @kellyselden looks difficult. Do you have any ideas how to make it more simple or reasonable?

beatle:
Tried to test almost everything(possibly not in the most efficient way) but coveralls says that coverage decreased by -0.009%.

I've implemented few new tests in this pr. @kellyselden @homu can you point me please how can I add new tests to code coverage?

lib/tasks/bower-install.js
```diff
@@ -2,56 +2,28 @@
...
-          ui.writeLine('    ' + chalk.green((index + 1) + ')') + ' ' +
-                       message.data.name + ' ' + pick.endpoint.target);
-        });
-      } else if (message.level === 'info' && options.verbose) {
-        // e.g.
-        //   cached git://example.com/some-package.git#1.0.0
-        ui.writeLine('  ' + chalk.green(message.id) + ' ' + message.message);
-      }
-    }
+    return _.compact([].concat(superArgs, packages));
```
beatle: `buildOptions`/`buildArgs` implementations looks difficult to me. @stefanpenner @kellyselden do you have any ideas on how to make it more simple or reasonable?

tests/acceptance/clean-test-slow.js
```diff
@@ -0,0 +1,157 @@
...
+        return tmp.setup(path.join('tmp', appName, 'tmp'));
+      }).then(function() {
+        return tmp.setup(path.join('tmp', appName, 'dist'));
+      }).then(function() {
+        return tmp.setup(path.join('tmp', appName, customBowerDir));
+      }).then(function() {
+        return tmp.setup(path.join('tmp', appName, 'bower_components'));
+      }).then(function() {
+        return tmp.setup(path.join('tmp', appName, 'node_modules'));
+      });
```
beatle: Is it better to use `acceptance.createTestTargets` here? I'm not 100% sure how to make it. @kellyselden any guidance please?

tests/acceptance/clean-test-slow.js
```diff
@@ -0,0 +1,157 @@
...
+      assertFileNotExists('dist');
+      assertFileNotExists('bower_components');
+      assertFileExists('node_modules/');
+    });
+  });
+
+  // @todo: Fix cache settings .npmrc
+  // Currently looks like lib/utilities/npm doesn't take it in account
+  // which causes global cache clean
+  xit('skip bower', function() {
```
beatle: Currently this test cleans global cache from `~/.npm`. Looks like npm task should pass npm config to `utilities/npm`.

tests/acceptance/clean-test-slow.js
```diff
@@ -0,0 +1,157 @@
...
+    process.chdir('./tmp/' + appName);
+    return ember([
+      'clean',
+      '--skip-bower'
+    ]).then(function() {
+      assertFileNotExists('tmp');
+      assertFileNotExists('dist');
+      assertFileNotExists('node_modules');
```
beatle: Currently this test cleans global cache from ~/.npm. Looks like npm task should pass npm config to utilities/npm.

lib/tasks/remove-files.js
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error('No file paths specified to remove');
+    }
+
+    return Promise.resolve().then(function() {
+        paths.forEach(function(dirToRemove) {
+          if (!options.dryRun) {
+            rimrafSync(dirToRemove);
+          }
+
+          var withTrailingSlash = ensureWithTrailingSlash(dirToRemove);
```
beatle: Currently `remove-files` is not the best name for this task. 
It assumes that all the files are directories and adds trailing slash for each file name in output.

beatle:
appveyor build fails. but looks like it's not pr related failures(EBUSY).

elwayman02:
@beatle is the PR otherwise ready to go, or?

rwjblue:
AppVeyor should be passing as of https://github.com/ember-cli/ember-cli/pull/4973, can you rebase and squash commits?

beatle:
@elwayman02 yes, I believe it is.

Actually there are still some raw places like:
 - `npm cache clean` and `bower cache clean` are not covered with acceptance tests. Just figured out yesterday how to make it but need some time to sit and make it.
 - arguable `utilities/package-manager-command` implementation and necessity. I feel it may be implemented in a different way but currently not sure which one (to be discussed and refactored).
 - refactor the `acceptance/clean-slow-test` to use an `acceptance.createTestTargets`

But all the points mentioned above are underthe hood stuff. 
As for the functionality everything is done.

elwayman02:
Let's get acceptance tests for the `cache clean` commands and refactor the `clean-slow-test`, then go back and work on `package-manager-command` in a separate PR. In addition to rebasing and squashing commits, once those two are done, we can merge this. Thanks for your hard work, @beatle!

beatle:
@elwayman02 I would rather say, joyful work! :)

@rwjblue yes, thank you. Will be back later with improved tests and squashed commits.

beatle:
Closing it after a short discussion in the `ember-cli` slack chat.

The reason is it seems like this feature quickly becomes redundant according to the current plans:
- get rid of [bower](https://github.com/ember-cli/ember-cli/issues/5028#issue-114602083). So we don't need clean bower_components and bower cache anymore
- bundle npm3 #5082. npm3 should handle cache and deps better than ever. So the same as for the bower.
- `tmp/` folder moves to the system tmp, so it will be automatically cleaned up each time ember server is terminated.

Anyway I want to say Thank You to ember-cli team for such a great product. Found a lot of interesting stuff working on this pr.

PR #4932, opened 2015-10-03
help tests faster and code cleanup

kellyselden:
* reusing more commandOptions in tests
* using a faster way to acceptance test help commands


kellyselden:
added `chai-string` to dev deps to get `endsWith`, very convenient when testing console output.

kellyselden:
@homu r+

PR #4934, opened 2015-10-04
Generate blueprint unit tests

kellyselden:
* additional unit tests for 100% coverage
* moved blueprint printing logic from generate command to blueprint model
* code cleanup/refactor Blueprint.printBasicHelp

This could use a second pair of eyes, to make sure someone agrees that blueprint option printing logic should be in the blueprint.


lib/utilities/printable-properties.js
```diff
@@ -0,0 +1,36 @@
...
+];
+var blueprintProperties = [
+  'name',
+  'description',
+  'availableOptions',
+  'anonymousOptions',
+  'overridden'
+];
+
+function forEachProperty(properties, forEach, context) {
```
stefanpenner: is this actually forEachWithProperty ?

stefanpenner:
obviously, this is lots of (existing crappy) code so i may have missed something. But this feels like the right direction.

stefanpenner:
Did another pass +1 from me (with the exception of my 1 comment).

lib/utilities/printable-properties.js
kellyselden: I don't recognize that keyword. Google didn't help. Can you explain?

stefanpenner: I think it is a better name for the function you have already defined.

It appears, we will forEach only members with a specific property, hence the naming recommendation of `forEachWithProperty`

lib/utilities/printable-properties.js
```diff
@@ -0,0 +1,36 @@
...
+  return properties.filter(function(key) {
+    return this[key] !== undefined;
```
stefanpenner: should this actually be `key` in this` ?

kellyselden:
made the suggested change of `forEachWithProperty`

kellyselden:
@homu r+

PR #4938, opened 2015-10-06
Add ember-truth-helpers to project blueprint

elwayman02:
ember-truth-helpers is now a "highly recommended" addon according to the Ember Core team, as referenced here: http://emberjs.com/blog/2015/10/04/ember-2-1-released.html#toc_code-get-code-helper

As such, it makes sense to include it in the default blueprint provided by ember-cli

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -24,6 +24,9 @@ module.exports = {
     contents.keywords = contents.keywords || [];
     contents.dependencies = contents.dependencies || {};
 
+    // only include `ember-truth-helpers` for project blueprints, not addons
+    delete contents.devDependencies['ember-truth-helpers'];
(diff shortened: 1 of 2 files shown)


stefanpenner:
This will need core team +1, as adding it here essentially makes it part of the long term public API.

I am inclined to be -1 on this, as I believe it is totally fine to have a quick-to-use really good optional addon.

elwayman02:
It's already included in [ember-cli-opinionated](https://github.com/elwayman02/ember-cli-opinionated#ember-truth-helpers), so I can go either way on this one, but I thought it merited discussion since it was specifically called out in the blog post for the latest release.

stefanpenner:
I think they work well as part of `opinionated`.

PR #4943, opened 2015-10-09
Fixes for ember-cli-blueprint-test-helper support

trabus:
Added a fixture `package.json` for the `tests/ember.js` helper. By using this fixture, it prevents the ember helper instance of `cli/index.js` from detecting any other `package.json` files, forcing it to act as if it were the global ember-cli (which is important for generating new projects for tests).

This is necessary to allow the ember helper be used by external test helper libs, otherwise it detects the consuming project's `package.json`, which causes a number of issues.

This PR is a requirement for [`ember-cli-internal-test-helpers`](https://github.com/ember-cli/ember-cli-internal-test-helpers) and [`ember-cli-blueprint-test-helpers`](https://github.com/ember-cli/ember-cli-blueprint-test-helpers).


trabus:
Looks like it just timed out. Will try re-running in a bit.

trabus:
I also needed to make `cli.logError` return an object containing the error and ui so we can assert `SilentErrors` (otherwise the only way is to read stdout). I also had to add some test coverage.

lib/cli/cli.js
```diff
@@ -149,9 +149,15 @@ CLI.prototype.callHelp = function(options) {
...
-    console.error(error.message);
-    console.error(error.stack);
+    debug(error.message);
+    if (error.stack) {
+      debug(error.stack);
+    }
   }
   this.ui.writeError(error);
-  return 1;
+  return {
```
stefanpenner: doesn't this mess up the bin/ember eecutable

trabus: How about I only return the object when `this.testing` is true then?

trabus:
Only returning the object when `cli.testing` is true.

lib/cli/cli.js
```diff
@@ -148,10 +148,18 @@ CLI.prototype.callHelp = function(options) {
 };
 
 CLI.prototype.logError = function(error) {
+  var result = 1;
   if (this.testing && error){
-    console.error(error.message);
-    console.error(error.stack);
+    debug(error.message);
```
trabus: @stefanpenner I put this behind a debug statement because it was noisy and confusing to see the intentional errors from testing failure cases in the test logs, but I talked to @lukemelia about it (who originally added it) and his reasoning was that it was previously obscuring errors during testing. I'm not quite sure how to best hide the errors when we are intentionally tripping them, but surfacing them when we need to, and was wondering if you had any input on how to handle this.

trabus:
Don't merge this yet, I've got a few more changes to go with this. I'll be adding them in an hour or so.

trabus:
Closing for now, will make a new PR with a proper way to pass in the UI soon.

PR #4692, opened 2015-08-16
Introduce `moduleForAcceptance`

seanpdoyle:
Closes [switchfly/ember-test-helpers#94][94] and
[rwjblue/ember-qunit#71][71].

[94]: https://github.com/switchfly/ember-test-helpers/issues/94
[71]: https://github.com/rwjblue/ember-qunit/pull/71

PR #4946, opened 2015-10-12
Fix beforeEach/afterEach callbacks with moduleForAcceptance.

rwjblue:
These were previously calling the provided `beforeEach`/`afterEach` blocks with a single argument (that is the original arguments object) instead of the correct arguments.

Updated to use `.apply` instead of `.call`.


rwjblue:
This was added in https://github.com/ember-cli/ember-cli/pull/4692, but I don't see any tests in the repo for `moduleForAcceptance` (which would likely have caught that providing `beforeEach`/`afterEach` would not have worked).

I added https://github.com/ember-cli/ember-cli/issues/4947 to track getting tests added.

PR #4955, opened 2015-10-15
Added a new outputReady builder hook

### LINKED ISSUE #4247: postBuild fails -- dist is empty
Using 0.2.7.

I have an addon that copies files from dist on postBuild. This has been working fine. Suddenly, (as in, I merely deleted a console.log() line) i'm getting the following error:
```
Build failed.
ENOENT, no such file or directory '/var/www/vhosts/employeronestats/client/dist/index.html'
Error: ENOENT, no such file or directory '/var/www/vhosts/employeronestats/client/dist/index.html'
    at Error (native)
    at Object.fs.statSync (evalmachine.<anonymous>:801:18)
    at Object.copySync (/var/www/vhosts/employeronestats/client/node_modules/fs-extra/lib/copy.js:74:58)
    at Class.module.exports.postBuild (/var/www/vhosts/employeronestats/client/lib/asset-delivery/index.js:11:6)
    at /var/www/vhosts/employeronestats/client/node_modules/ember-cli/lib/models/builder.js:109:34
    at Array.map (native)
    at Class.module.exports.Task.extend.processAddonBuildSteps (/var/www/vhosts/employeronestats/client/node_modules/ember-cli/lib/models/builder.js:107:43)
    at lib$rsvp$$internal$$tryCatch (/var/www/vhosts/employeronestats/client/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:489:16)
    at lib$rsvp$$internal$$invokeCallback (/var/www/vhosts/employeronestats/client/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:501:17)
    at lib$rsvp$$internal$$publish (/var/www/vhosts/employeronestats/client/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:472:11)
```

Looking through lib/models/builder.js, in build() i see that processAddonBuildSteps, which calls the addOn, is called before processBuildResult, which copies files to dist. Has this changed recently? I can see how it could potentially work out ok, being promises. But i've noticed that, since upgrading, sometimes a build does not include changes just made, and i have to restart the server to generate a new build. Now that i'm getting this error it does seem that something is different about building in this version.

UPDATE: The order was changed May 6. I don't know how to properly link to this; apologies.

 https://github.com/ember-cli/ember-cli/commit/1ef86173e00cd755886f55d33eb473a32af310be

I can't say that i agree with the reasoning. If it's really worthwhile having a hook available after the build but before copying, then perhaps a new hook should be created. The docs for postBuild do say that it can be used as a hook for copying files. This is no longer the case.

patocallaghan:
In ember-cli/ember-cli#4027 the `postBuild` hook was re-ordered to appear before the `processBuildResult` step. This allowed you to do some work on the final tree before it was output to `dist`. This also meant there was no longer a hook which fired to do work after everything was copied to `dist`, which is something we're trying to do.

This PR re-adds the old hook, fired after `processBuildResult`, but with a new name `postResult` (name was suggested in #4247). I'm open to changing the hook name if there are any better ideas.

This should also resolve #4247

### DIFF
--- a/ADDON_HOOKS.md
+++ b/ADDON_HOOKS.md
@@ -6,6 +6,7 @@ Table of Contents:
 - [serverMiddleware](#servermiddleware)
 - [postBuild](#postbuild)
 - [preBuild](#prebuild)
+- [outputReady](#outputready)
 - [buildError](#builderror)
 - [included](#included)
 - [setupPreprocessorRegistry](#setuppreprocessorregistry)
@@ -171,7 +172,6 @@ Gives access to the result of the tree, and the location of the output.
 
 - Slow tree listing
 - May be used to manipulate your project after build has happened
-- Opportunity to symlink or copy files elsewhere.
 
 **Examples:**
 
@@ -191,17 +191,27 @@ Hook called before build takes place.
 
 **Uses:**
 
+<a name='outputready'></a>
+## outputReady
+
+Hook called after the build has been processed and the files have been copied to the output directory
+
+**Received arguments:**
+
+- Result object from broccoli build
+  - `result.directory` - final output path
+
+**Default implementation:** None
+
 **Examples:**
 
-- [ember-cli-rails-addon](https://github.com/rondale-sc/ember-cli-rails-addon/blob/v0.0.11/index.js#L41)
-  - In this case we are using this in tandem with a rails middleware to create a lock file.
-  *[See postBuild]*
+- Opportunity to symlink or copy files elsewhere.
 
 <a name='builderror'></a>
 ## buildError
 
 buildError hook will be called on when an error occurs during the
-preBuild or postBuild hooks for addons, or when builder#build
+preBuild, postBuild or outputReady hooks for addons, or when builder#build
 fails
 
 **Received arguments:**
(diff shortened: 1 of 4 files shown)


stefanpenner:
LGTM, can you add a short blurb about this in https://github.com/ember-cli/ember-cli/blob/master/ADDON_HOOKS.md

rwjblue:
LGTM also.

patocallaghan:
@stefanpenner I've added some blurb to the `ADDON_HOOKS` file as well as added some comments to the `models/addon.js` file.

I just want to also make sure that `postResult` is the correct hook name? Other suggestions might be `postBuildResult` or `postOutput`?

stefanpenner:
> or postOutput?

i like postOutput

stefanpenner:
another is `outputReady`, the `output` step really is just a copy.

patocallaghan:
`outputReady` seems most descriptive to me. I think I'll run with that one.

stefanpenner:
> outputReady seems most descriptive to me. I think I'll run with that one.

:)

patocallaghan:
@stefanpenner should be good to go now. I've renamed the hook to `outputReady`

stefanpenner:
<3

stefanpenner:
@homu r+

kjhangiani:
Any idea when this will be released?  I'm in the process of going from ember-cli 0.2.2 to 0.2.7, and ember 1.11.3 to ember 1.13, but the re-ordering of postBuild in ember-cli 0.2.4 has broken our app.  We depend on postBuild to create symlinks in dist... have got everything updated, but unfortunately cannot actually deploy anything to production without this fix.

Is there a workaround for creating symlinks in ember-cli 0.2.7 in the dist folder?

PR #4891, opened 2015-09-23
[ENHANCEMENT] Upgrade dependencies for app/addon blueprints

elwayman02:
1) Bumped Ember & Ember-Data to the latest stable versions (2.1.0/2.1.0, respectively)
2) Removed github tags for `ember-cli-shims`, `ember-cli-test-loader`, `ember-load-initializers`, and `loader.js`, since these are all registered with bower and redirect to the correct repositories
3) Upgraded `ember-cli-release`, `ember-qunit` & `ember-cli-shims` to latest patch versions
4) Upgraded versions for several npm dependencies that were using `^`, which should make no difference in installation other than having the current version listed now.

PR #4958, opened 2015-10-15
Package upgrades

sethbrasile:
ember-cli-release 0.2.4 through 0.2.7 breaks ember-cli tests

https://travis-ci.org/ember-cli/ember-cli/jobs/83942510

### DIFF
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -1,18 +1,18 @@
 {
   "name": "<%= name %>",
   "dependencies": {
-    "ember": "2.0.0",
-    "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.5",
-    "ember-cli-test-loader": "ember-cli-test-loader#0.2.1",
-    "ember-data": "2.0.0",
-    "ember-load-initializers": "ember-cli/ember-load-initializers#0.1.7",
-    "ember-qunit": "0.4.10",
+    "ember": "2.1.0",
+    "ember-cli-shims": "0.0.6",
+    "ember-cli-test-loader": "0.2.1",
+    "ember-data": "2.1.0",
+    "ember-load-initializers": "0.1.7",
+    "ember-qunit": "0.4.13",
     "ember-qunit-notifications": "0.1.0",
     "jquery": "^1.11.3",
-    "loader.js": "ember-cli/loader.js#3.3.0",
+    "loader.js": "3.3.0",
     "qunit": "~1.19.0"
   },
   "resolutions": {
-    "ember": "2.0.0"
+    "ember": "2.1.0"
   }
 }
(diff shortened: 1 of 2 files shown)


elwayman02:
@stefanpenner @rwjblue can we get @homu to try this out, now that travis tests are passing?

slindberg:
Just starting to look into this, but I'm unsure why you think that ember-cli-release is breaking appveyor, since it's still failing after downgrading to v0.2.4?

slindberg:
So, ember-cli-release does some mild hackery with the `Command` class to initialize the `availableOptions` array, and something goes wrong when it [gets required](https://github.com/lytics/ember-cli-release/blob/v0.2.7/lib/commands/release.js#L12). The test suite is doing some fancy things to mimic node modules, one of them being [symlinking the `ember-cli` package to the project root](https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/acceptance.js#L166-L174). Near as I can tell, this is what's causing it to blow up, but I don't have a solution at the moment.

Perhaps the easiest way would be to avoid using the `Command` class altogether by adding the ability for commands to specify `availableOptions` as a function (ember-cli feature). @rwjblue mentioned that the the `release` command could just short circuit if the class isn't found, but I'm not a huge fan of that solution. I'll look into better options this weekend.

rwjblue:
making availableOptions a function seems like a good option

stefanpenner:
kicking appveyor

slindberg:
@rwjblue I opened a PR (#4966) for supporting `availableOptions` as a function. I think this is the best long-term solution, but I realized it doesn't really help for this situation since I can't target an ember-cli release with the fix, even if the above PR is merged. In the meantime I've removed the command model require in ember-cli-release with a questionable hack (lytics/ember-cli-release#32), which I've verified fixes this issue. Would love to get your input on those PRs.

rwjblue:
@slindberg - Thank you for digging into this.  It is definitely annoying that everything actually works in reality, but the test infrastructure here in ember-cli causes an issue with addons extending the command object.  I left a few comments on the various PR's you linked.

slindberg:
Released version [0.2.8](https://github.com/lytics/ember-cli-release/tree/v0.2.8) of ember-cli-release, which should fix this issue.

sethbrasile:
thanks @slindberg, updated. We'll see if it passes in a couple hours :D

slindberg:
Hard to believe that this diff reduces test coverage... but either way, looks like `ember-cli-release` isn't causing test failures anymore :tada:

kellyselden:
Can we get a squash please?

elwayman02:
@sethbrasile @slindberg @kellyselden I updated & squashed the original PR #4891

sethbrasile:
:+1:

PR #4962, opened 2015-10-16
Declares node.js version as documented by Travis

xcambar:
As per http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Choosing-Node-versions-to-test-against

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,15 @@
 sudo: false
-env:
-  - NODE_VERSION=0.10
-  - NODE_VERSION=0.12
-  - NODE_VERSION=iojs
+language: node_js
+node_js:
+  - "0.10"
+  - "0.12"
+  - "iojs"
 
 os:
   - linux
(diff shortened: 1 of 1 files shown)


xcambar:
From appveyor: 

```
PhantomJS not installed. An error occurred during installation:
 The remote server returned an error: (500) Internal Server Error.
```

stefanpenner:
Travis ci for OS X does not support this. If we wish to restore that we would need to once again change this

xcambar:
What are the odds/conditions for it to be restored? Did the Travis team say something about it?

stefanpenner:
> What are the odds/conditions for it to be restored

Hopefully high, we disabled it because our tests where so slow. @kellyselden has been doing a great job improving them.

Im fine with merging this, and we can restore once we restore OSX ci. maybe by then travis will support the full travis.yml/lang: for OSX as well.

stefanpenner:
@homu r+

PR #4954, opened 2015-10-14
Add smoke test for `moduleForAcceptance`

seanpdoyle:
Closes [#4947].

[#4947]: https://github.com/ember-cli/ember-cli/issues/4947

PR #4963, opened 2015-10-16
Use `ember` test helper in Smoke Tests

seanpdoyle:
The `ember` command is a faster way to run ember-cli in tests.

https://github.com/ember-cli/ember-cli/pull/4954#discussion_r42235864

### DIFF
--- a/tests/acceptance/smoke-test-slow.js
+++ b/tests/acceptance/smoke-test-slow.js
@@ -8,7 +8,6 @@ var walkSync = require('walk-sync');
 var appName  = 'some-cool-app';
 var EOL      = require('os').EOL;
 
-var runCommand          = require('../helpers/run-command');
 var acceptance          = require('../helpers/acceptance');
 var copyFixtureFiles    = require('../helpers/copy-fixture-files');
 var killCliProcess      = require('../helpers/kill-cli-process');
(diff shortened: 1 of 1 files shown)


seanpdoyle:
@kellyselden do we have any idea why this would break the test sutie?

kellyselden:
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/smoke-test-slow.js#L149

These are throughout the code. The `ember` helper has an obscure bug in it. Unfortunately, this means we can't use it everywhere yet.

stefanpenner:
seems to have gone stale. I would like to recommend closing PR if they are no longer being pursued.

seanpdoyle:
@stefanpenner looking at the diff, is there any reason this shouldn't pass?

stefanpenner:
@homu r+

seanpdoyle:
@stefanpenner do I just keep re-running until we're green? What do you think is the root of the problem here?

stefanpenner:
unsure, let me try locally

stefanpenner:
found whats causing grief: https://github.com/joliss/node-quick-temp/issues/8

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/5140

PR #4981, opened 2015-10-22
alphabetize the keys in dependencies and devDependencies. (for addon creation)

kellyselden:
JSON.stringify in this case is resulting in out of order dependencies.
The next `npm install --save-dev` for a user will screw up their diffs:

capture

This will keep the files ordered.

Also puts back the newline that gets removed in the stringify.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -15,14 +15,15 @@ module.exports = {
   description: 'The default blueprint for ember-cli addons.',
 
   generatePackageJson: function() {
-    var packagePath = path.join(this._appBlueprint.path, 'files', 'package.json');
-    var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
+    var contents = readContentsFromFile.call(this, 'package.json');
 
     delete contents.private;
     contents.name = this.project.name();
     contents.description = this.description;
+    contents.scripts = contents.scripts || {};
     contents.keywords = contents.keywords || [];
     contents.dependencies = contents.dependencies || {};
+    contents.devDependencies = contents.devDependencies || {};
 
     // npm doesn't like it when we have something in both deps and devDeps
     // and dummy app still uses it when in deps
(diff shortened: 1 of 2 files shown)


kellyselden:
This could probably use a test...

kellyselden:
I've added tests, and this is ready for review. I've tested locally on Windows, so any CI issues I "think" are pre-existing?

blueprints/addon/index.js
```diff
@@ -52,7 +55,16 @@ module.exports = {
...
 
-    fs.writeFileSync(path.join(this.path, 'files', 'bower.json'), JSON.stringify(contents, null, 2));
+    this._writeContentsToFile(contents, 'bower.json');
+  },
+
+  _writeContentsToFile: function(contents, fileName) {
+    fs.writeFileSync(path.join(this.path, 'files', fileName), JSON.stringify(contents, null, 2) + '\n');
+  },
+
+  _alphabetizeDependencies: function(contents) {
```
stefanpenner: this method doesnt use the classes internal state, it should not be on the prototype.

blueprints/addon/index.js
```diff
@@ -52,7 +55,16 @@ module.exports = {
 
     contents.name = this.project.name();
 
-    fs.writeFileSync(path.join(this.path, 'files', 'bower.json'), JSON.stringify(contents, null, 2));
+    this._writeContentsToFile(contents, 'bower.json');
+  },
+
+  _writeContentsToFile: function(contents, fileName) {
+    fs.writeFileSync(path.join(this.path, 'files', fileName), JSON.stringify(contents, null, 2) + '\n');
```
stefanpenner: i believe instead https://github.com/substack/json-stable-stringify could be used.

blueprints/addon/index.js
kellyselden: agreed

blueprints/addon/index.js
kellyselden: I thought there was a reason I didn't/couldn't use that, but it's escaping me. I will look into it again.

stefanpenner: Ya its possible I missed something.

kellyselden:
comments addressed. `json-stable-stringify` worked like a charm!

blueprints/addon/index.js
```diff
@@ -149,3 +154,11 @@ module.exports = {
     return entityName;
   }
 };
+
+function alphabetizeObjectKeys(unordered) {
```
stefanpenner: This I think is no longer needed

kellyselden:
I remember why I didn't use `json-stable-stringify` now. It alphabetizes the whole file, putting `dependencies` up top above `scripts`. I think I have to revert.

stefanpenner:
> I think I have to revert.

ok

kellyselden:
This should be good now. I had to add lots of unit tests because there were none previously. The failures all have to do with EADDRINUSE, which I don't believe I cause. Tests work locally.

rwjblue:
Looks like some JSHint failures are all that is left.

kellyselden:
ahh forgot the `/*jshint multistr: true */`

kellyselden:
looking green now.

kellyselden:
@homu r+

PR #4985, opened 2015-10-24
Remove prototype extensions from addon docs

Robdel12:
Since prototype extensions are disabled in addons we shouldn't shown
them used in the examples.


rwjblue:
Thank you!

Robdel12:
:heart:

PR #4909, opened 2015-09-30
remove blueprints that will be provided by:

stefanpenner:
- [x] extract ember: https://github.com/emberjs/ember.js/pull/12412
- [x] ember-legacy-blueprint-addon: https://github.com/ember-cli/ember-cli-legacy-blueprints
- [x] depend on addonified ember
- [x] depend on addonified ember-data

- [ ] tests
  clearly without the blueprints we will have many test failures, once we depend on addonified ember, those may stop. We should still extract nice blueprint test helpers for the various extracted addons to use.

PR #4986, opened 2015-10-24
fix model relationship name generation

gabrielgrant:
It is not currently possible to generate a model with a relationship named differently from the model's name ([despite the help indicating it should be](https://github.com/ember-cli/ember-cli/blob/master/blueprints/model/HELP.md))

This patch brings the code's behavior in line with the docs.

The blueprint part of this patch will be obsoleted if #4909 lands, so I've also PRed the changes to the blueprint in ED: emberjs/data#3885 

(the arg-parsing-fix part of this patch is still needed to pass a distinct name, though)

### DIFF
--- a/blueprints/model/index.js
+++ b/blueprints/model/index.js
@@ -19,25 +19,40 @@ module.exports = {
 
     for (var name in entityOptions) {
       var type = entityOptions[name] || '';
+      var foreignModel = name;
+      if (type.indexOf(':') > -1) {
+        foreignModel = type.split(':')[1];
+        type = type.split(':')[0];
+      }
       var dasherizedName = stringUtils.dasherize(name);
-      var dasherizedNameSingular = inflection.singularize(dasherizedName);
       var camelizedName = stringUtils.camelize(name);
       var dasherizedType = stringUtils.dasherize(type);
+      var dasherizedForeignModel = stringUtils.dasherize(foreignModel);
+      var dasherizedForeignModelSingular = inflection.singularize(dasherizedForeignModel);
 
+      var attr;
       if (/has-many/.test(dasherizedType)) {
         var camelizedNamePlural = inflection.pluralize(camelizedName);
-        attrs.push(camelizedNamePlural + ': ' + dsAttr(dasherizedName, dasherizedType));
+        attr = dsAttr(dasherizedForeignModelSingular, dasherizedType);
+        attrs.push(camelizedNamePlural + ': ' + attr);
+      } else if (/belongs-to/.test(dasherizedType)) {
+        attr = dsAttr(dasherizedForeignModel, dasherizedType);
+        attrs.push(camelizedName + ': ' + attr);
       } else {
-        attrs.push(camelizedName + ': ' + dsAttr(dasherizedName, dasherizedType));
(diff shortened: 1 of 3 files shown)


gabrielgrant:
To clarify ([requested on the ember-data pull request](https://github.com/emberjs/data/pull/3885#issuecomment-151051084)): 

[The docs for the model blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/model/HELP.md) state that `ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc` should result in the following model:
 
```
import DS from 'ember-data';
export default DS.Model.extend({
  filling: DS.belongsTo('protein'),
  toppings: DS.hasMany('topping'),
  name: DS.attr('string'),
  price: DS.attr('number'),
  misc: DS.attr()
});
```

But instead, this is what that command actually generates:

```
import DS from 'ember-data';

export default DS.Model.extend({
  filling: DS.belongsTo('filling'),  // <------ should be 'protein'
  toppings: DS.hasMany('topping'),
  name: DS.attr('string'),
  price: DS.attr('number'),
  misc: DS.attr()
});
```

That is, the name of the model gets ignored, and instead just pulled from the name of the field. This patch fixes the issue in the blueprint and the argument parsing code ([it currently just throws out everything after the second `:`](https://github.com/ember-cli/ember-cli/pull/4986/files#diff-8690f2d0f74746d76ea9fc51ecfc0ab6L8))

Sorry for lack of tests; it doesn't look like there's testing infrastructure for blueprint generation at the moment, and putting that in place should probably be a separate PR.

stefanpenner:
At first glance this looks good, but we should add a unit test covering this `locals` method. It seems to have gotten quite complicated, and we should be sure not to regress on your current fix.

stefanpenner:
also kicking travis

gabrielgrant:
Thanks for taking a look. Not sure how I came to the conclusion there aren't tests for any blueprints, I obviously wasn't looking hard enough. I'll add one now. Make sense to base it on the [component test?](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/blueprints/component-test.js)

stefanpenner:
> Make sense to base it on the component test?

yup

gabrielgrant:
Ok, added some tests (for this PR's code and the module in general). The only behavior I'm not totally sure of is [in the final test](https://github.com/ember-cli/ember-cli/commit/846ed72a531c323939f36a6d8304fed10629f677#diff-fd60f88b64655e5de4b4a88240ab4bf2R96): the property name for hasMany relationships [gets pluralized](https://github.com/ember-cli/ember-cli/commit/846ed72a531c323939f36a6d8304fed10629f677#diff-fd60f88b64655e5de4b4a88240ab4bf2R99)

To be clear, this is just something I noticed, [I haven't changed the behavior here at all](https://github.com/ember-cli/ember-cli/pull/4986/files#diff-e942de8dc860c139ab1d1390be588dd5L29). But in the case that you're setting the property name and model name explicitly (now possible) it is a little surprising (probably makes more sense if you're just doing something like `topping:has-many`)

stefanpenner:
LGTM

stefanpenner:
@homu r+

PR #4996, opened 2015-10-27
Use shorthand methods for ED transform blueprint

knownasilya:
Promotes users to use ES6 shorthand methods by making them the default.

PR #4999, opened 2015-10-27
[INTERNAL] Fix tests that were breaking due to #4996

yaymukund:
These tests need to be updated to match the ES6 syntax in the blueprint.


kellyselden:
Thanks! (I forgot to Homu...)

PR #5000, opened 2015-10-27
[WIP] Add DOM guard unless `preventDOMGuard` specified on `import`

rondale-sc:
The idea here is to make it relatively painless to be fastboot compliant by automatically wrapping imported files in:

```
return "if (document !== 'undefined') {" + content + "}";
```

This can be opted out of by specifying  `preventDOMGuard`.

For example, `app.import(this.bowerDirectory + '/foo/foo.js', { preventDOMGuard: true });` would produce no guard.

I still have some questions about this, so I'm going to mark it as WIP for now. 

❤️

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -38,6 +38,7 @@ var merge         = require('lodash/object/merge');
 var omit          = require('lodash/object/omit');
 var ES3SafeFilter = require('broccoli-es3-safe-recast');
 var Funnel        = require('broccoli-funnel');
+var stewMap       = require('broccoli-stew').map;
 
 module.exports = EmberApp;
 
@@ -98,6 +99,7 @@ function EmberApp() {
   this.otherAssetPaths         = [];
   this.legacyTestFilesToAppend = [];
   this.vendorTestStaticStyles  = [];
+  this.preventDOMGuard         = {};
 
   this.trees = this.options.trees;
 
@@ -1045,6 +1047,7 @@ EmberApp.prototype.javascript = function() {
   var legacyFilesToAppend = this.legacyFilesToAppend;
   var appOutputPath       = this.options.outputPaths.app.js;
   var appJs = applicationJs;
+  var preventDOMGuard     = this.preventDOMGuard;
 
   // Note: If ember-cli-babel is installed we have already performed the transpilation at this point
   if (!this._addonInstalled('ember-cli-babel')) {
@@ -1083,7 +1086,17 @@ EmberApp.prototype.javascript = function() {
     .concat('vendor/addons.js')
     .concat('vendor/ember-cli/vendor-suffix.js');
 
-  var vendor = this.concatFiles(applicationJs, {
+
+  var vendor = stewMap(applicationJs, function(content, relativePath){
+    if (legacyFilesToAppend.indexOf(relativePath) > -1) {
+      if (!preventDOMGuard[relativePath]) {
+        return "if (document !== 'undefined') {" + content + "}";
+      }
+    }
+    return content;
+  });
+
+  vendor = this.concatFiles(vendor, {
     inputFiles: inputFiles,
     outputFile: this.options.outputPaths.vendor.js,
     separator: EOL + ';',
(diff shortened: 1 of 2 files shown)


rwjblue:
I like this PR, thanks for working on it! 

I think the biggest question is what the default value should be (aka should we wrap in a guard if no options are provided). Arguments can be made for either side...

/cc @tomdale @chancancode

rwjblue:
@rondale-sc - We will definitely want to have this fairly well tested, so we'll need to add some here once you are fairly confident with the implementation. Happy to help brain storm the best test strategy...

rondale-sc:
@rwjblue Absolutely tests, I just wasn't sure about how to do so without a dummy app.  Going to give it some thought and try to tackle it tomorrow.

lib/broccoli/ember-app.js
```diff
@@ -1176,8 +1191,19 @@ EmberApp.prototype.testFiles = function(coreTestTree) {
 
   var emberCLITree = this._processedEmberCLITree();
 
-  var testJs = this.concatFiles(mergeTrees([external, coreTestTree]), {
-    inputFiles: this.legacyTestFilesToAppend,
+  var coreExt = mergeTrees([external, coreTestTree])
+
+  coreExt = stewMap(coreExt, function(content, relativePath){
+    if (legacyTestFilesToAppend.indexOf(relativePath) > -1) {
```
stefanpenner: this is going to pwn sourcemaps

rondale-sc: Perf-wise?

chadhietala:
I actually think this is a bad pattern and what really needs to happen is that the dom-helper needs to be exposed as a public API. We have a lot of branching like this in our app and I it is less than desirable; I wouldn't want this to come off as a nod as to being ok.

lib/broccoli/ember-app.js
stefanpenner: just the potential risk of changing the output without doing so in-terms of sourceMaps

rondale-sc: Ah, I see.  Should the wrapping happen before source maps then?

stefanpenner: i haven't taken much time to think about this or this PR, but I'm inclined to think it is something that should happen during sourceMapConcat

rwjblue: Ya, that would work.  Instead of prepending each file, we would just concat a `before-document-guard.js` file, the file contents, then an `after-document-guard.js` file.  That would properly preserve sourcemaps.

stefanpenner: i would recommend instead using `header` and `footer` slots

```js
var node = concat(node, {
  outputFile: '/output.js',
  header: ";(function() {",
  inputFiles: ['**/*']
  footer: "}());"
});
```

that being said, I believe this is would require a per file header/footer.

maybe something like `map` should be added to sourceMap concat, if map returns an `array` each chunk should be source-map aware concated?

```js
var node = concat(node, {
  outputFile: '/output.js',
  inputFiles: ['**/*'],

  map(content) {
    return [
      'perf file header',
      content,
      'per file footer'
    ];
  }
});
```

tomdale:
This is awesome work but I don't think it's exactly the right approach. After discussing it with @rwjblue and others, I think the best thing to do is to wrap each addon in a `try/catch` and report any failures as an error. That way users can start filing issues/submitting PRs to addons to fix their FastBoot compatibility issues.

rwjblue:
@tomdale - Agreed, that sounds like a great plan. I'll circle around with @rondale-sc and see if we can update this PR to do that...

chadhietala:
In some cases you need/want the dom-helper. An example of this is that you can return a DOM node from a helper, while this is in the "power user" arena, would be nice if you could build up a DOM manually that did the right thing instead of forking and having 2 implementations.

rwjblue:
@chadhietala - I agree, and I would need to dig into the updated versions of the DOM helper stuff in glimmer2 to understand better how that would work. In general though, that is a different concern than ensuring that the addon code itself does not throw an error at eval time under fastboot.

stefanpenner:
This seems to have gone stale, closing for now.

tomdale:
@chadhietala I think the plan is to expose the DOM helper as public once G2 lands.

I think closing this is appropriate, there's a PR for this to FastBoot that wraps addons in a try/catch.

chadhietala:
@tomdale Sounds good.

PR #5004, opened 2015-10-28
Test blueprints to use const instead of var

sivakumar-kailasam:
App and source blueprints use `const` / `let` but the test blueprints don't. This change fixes that across test blueprints. 

This is the first thing I change in any newly generated test file so felt it should be in the cli blueprint.

### DIFF
--- a/blueprints/adapter-test/files/tests/unit/__path__/__test__.js
+++ b/blueprints/adapter-test/files/tests/unit/__path__/__test__.js
@@ -7,6 +7,6 @@ moduleFor('adapter:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>
 
 // Replace this with your real tests.
 test('it exists', function(assert) {
-  var adapter = this.subject();
+  let adapter = this.subject();
   assert.ok(adapter);
 });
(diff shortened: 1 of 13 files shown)


stefanpenner:
Lets keep const for top level variables that should remain constant in all circumstances, and use let for local variables that may or may not remain constant

sivakumar-kailasam:
Sure

sivakumar-kailasam:
@stefanpenner made the changes and builds are successful (appveyor is done by taking time to update) :smile:

stefanpenner:
LGTM, thanks

@homu r+

sivakumar-kailasam:
The build failed with `The command "nvm install iojs" failed and exited with 105 during .`

stefanpenner:
@homu retry

sivakumar-kailasam:
:confounded: Failed because of the same reason as last time

stefanpenner:
Maybe an issue upstream, will try again later.

stefanpenner:
@homu retry

sivakumar-kailasam:
Awesome :smile:

PR #5012, opened 2015-10-29
Add method and property shorthand for blueprints

jimenglish81:
Make blueprints consistent for es6 syntax.


rwjblue:
@homu r+

jimenglish81:
thanks :+1:

PR #5017, opened 2015-10-30
[BUGFIX] close #5014 move git init to before npm and bower install

### LINKED ISSUE #5014: do git init before npm and bower on new projects
I started a new addon, and while I was waiting for npm, I was filling in the package.json. I saved it before npm finished. When everything was finished, I noticed the "Initial Commit from Ember CLI ..." included my change, which I didn't expect.

I figure this is an easy change, and if others agree, could be low hanging fruit for new contributors.

acburdine:
Moves the git setup to the 'init' command process

Includes a check for the skipGit flag to be explicitly false, so that if it is 'undefined' it doesn't run.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -46,6 +46,15 @@ module.exports = Command.extend({
       project: this.project
     });
 
+    // needs an explicit check in case it's just 'undefined'
+    // due to passing of options from 'new' and 'addon'
+    if (commandOptions.skipGit === false) {
+      var gitInit = new this.tasks.GitInit({
+        ui: this.ui,
+        project: this.project
+      });
+    }
+
     if (!commandOptions.skipNpm) {
       var npmInstall = new this.tasks.NpmInstall({
         ui: this.ui,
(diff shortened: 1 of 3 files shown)


lib/commands/init.js
```diff
@@ -90,6 +99,11 @@ module.exports = Command.extend({
     debug('before:installblueprint');
     return installBlueprint.run(blueprintOpts)
       .then(function() {
+        if (commandOptions.skipGit === false) {
+          return gitInit.run.bind(gitInit, commandOptions, rawArgs)();
+        }
+      })
+      .then(function() {
         debug('after:installblueprint');
```
kellyselden: This should move to before the git init, or to its own `then`.

lib/commands/init.js
```diff
@@ -90,6 +99,11 @@ module.exports = Command.extend({
     debug('before:installblueprint');
     return installBlueprint.run(blueprintOpts)
       .then(function() {
+        if (commandOptions.skipGit === false) {
+          return gitInit.run.bind(gitInit, commandOptions, rawArgs)();
```
kellyselden: I don't think you have to bind when running it like this.

kellyselden:
Looks good!

@homu r+

PR #5023, opened 2015-10-31
Add assertion to make sure tests file is included when running tests

trentmwillis:


rwjblue:
This looks good and should help folks quite a bit while upgrading. Can you tweak it to actually load the tests also?

trentmwillis:
@rwjblue sure, via `$.getScript`? Do you think it would be worthwhile concatenating this stuff into `test-loader.js` instead of via `content-for`?

stefanpenner:
@rwjblue I am inclined (5/10) to just fail with a good error, if you feel strongly i'll gladly swing in in-favor of  your suggestion

rwjblue:
I don't care a ton, some messaging (as is added here) is massively better than none.  This particular gotcha has tripped me up while upgrading a few apps to master.  An error like the one added here would have let me fix it instantly...

stefanpenner:
> An error like the one added here would have let me fix it instantly...

:+1: 

im good with this then, @rwjblue r?

rwjblue:
@homu r+

Issue #5027, opened 2015-11-02
"Future Versions" Node error out-of-date


patocallaghan:
Running Node 5.0 gives the Node support error

```
Future versions of Ember CLI will not support 5.0.0 Please update to Node 0.12 or io.js.'
```

https://github.com/ember-cli/ember-cli/blob/3c615aa55dfe6be1adf14ef67e74028b09560138/lib/cli/cli.js#L58

I'll can cut a PR for this but what should it be saying here instead? It seems outdated to be referencing 0.12 and io.js

stefanpenner:
It should likely just warn that the version is untested and may have issues. (Or something along those lines)

PR #4815, opened 2015-09-08
[BUGFIX #4814] Add Node.js 4.0 as valid platform version

szines:
Connected Issue: https://github.com/ember-cli/ember-cli/issues/4814

image

PR #5031, opened 2015-11-03
Updated unsupported node version error

patocallaghan:
Fixes ember-cli/ember-cli#5027

Removed `validPlatformVersion` utility function in favour of a new `PlatformChecker` module.

With the `PlatformChecker` we can now determine whether the Node platform is:

* `isValid` -  sits within the accepted version range
* `isDeprecated` - sits below the accepted version range
* `isUntested` - sits above the accepted version range

**Valid**
image

**Untested**
image

**Deprecated**
image


rwjblue:
This message looks good for Node v5.x, but we should still say that v0.10 is deprecated.

Can you tweak to ensure that the deprecation is printed for v0.10, and this newer (much nicer) warning is printed for other unknown versions?

patocallaghan:
@rwjblue so what does ember-cli officially support? I've noticed that we actually just [accept any node version that is greater than](https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/valid-platform-version.js#L7) `>=0.12.0`. Should this be more granular? `>=0.12.0 <=5.0.0` perhaps?
If not, perhaps the only part of the deprecation error I need to change is the text about *Please update to Node 0.12 or io.js.*

rwjblue:
Sorry @patocallaghan, I got confused by how long it has been since we have released :crying_cat_face:. 

Looks like @stefanpenner and I are generally in agreement here though based on https://github.com/ember-cli/ember-cli/pull/4815#issuecomment-138739871.  I would like a deprecation for < 0.12 and a warning for > 4.0.0 (until we can test 5.x).

patocallaghan:
@rwjblue I've updated the PR and description above. I've introduced a new module for checking the platform and removed the previous lib that was used. Let me know if that is okay.

Looking at the `travis.yml` it looks like we run the [CI tests against v5](https://github.com/ember-cli/ember-cli/blob/master/.travis.yml#L8). Does this not imply that we already support Node v5?

Also should the warnings be yellow in both cases or red? Yellow implies a warning but should a deprecation message be red?

lib/utilities/platform-checker.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+var semver = require('semver');
+
+var LOWER_RANGE = '0.12.0';
+var UPPER_RANGE = '4.2.1';
```
kellyselden: This should be <5.0.0, because we want point releases like 4.2.2 to pass without a code change.

patocallaghan: @kellyselden sure, makes sense. I've just updated it.

rwjblue:
> Looking at the travis.yml it looks like we run the CI tests against v5. Does this not imply that we already support Node v5?

Agreed. Let's make the upper bounds use 6.0.0.

patocallaghan:
@rwjblue I've updated to allow v5 in the valid range

rwjblue:
@homu r+

Issue #5038, opened 2015-11-05
Errors are not written to `STDERR`


seanpdoyle:
A missing dependency detected by [quaertym/ember-cli-dependency-checker](https://github.com/quaertym/ember-cli-dependency-checker/blob/master/lib/reporter.js#L65) raises an Error.

EmberCLI logs the error to `STDOUT`, but not the `STDERR`:

```bash
2.2.3 in frontend/ on master 
› rm -rf node_modules/ember-data

2.2.3 in frontend/ on master 
› ember build 2> tmp/stderr.txt 

Missing npm packages: 
Package: ember-data
  * Specified: 2.1.0
  * Installed: (not installed)

Run `npm install` to install missing dependencies.


2.2.3 in frontend/ on master 
› cat tmp/stderr.txt 

2.2.3 in frontend/ on master 
› 
```

seanpdoyle:
stefanpenner [8:40 PM] 
that in theory should go through ui.writeError

stefanpenner [8:41 PM]
i would add a debugger/logger to confirm

rwjblue [8:43 PM] 
@seanpdoyle: it should be hitting https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L156-L163 (edited)

rwjblue [8:44 PM]
Which uses https://github.com/ember-cli/ember-cli/blob/master/lib/ui/write-error.js (edited)

seanpdoyle [8:44 PM] 
@rwjblue: right

rwjblue [8:45 PM] 
ugh

rwjblue [8:45 PM]
we don't write anything to stderr

rwjblue [8:45 PM]
ever

seanpdoyle [8:45 PM] 
I’ve added a console and it goes through there

rwjblue [8:46 PM] 
`UI` only manages a single `outputStream` (edited)

rwjblue [8:46 PM]
and we pass in `process.stdout`

rwjblue [8:46 PM]
here: https://github.com/ember-cli/ember-cli/blob/3c615aa55dfe6be1adf14ef67e74028b09560138/bin/ember#L29 (edited)

seanpdoyle [8:47 PM] 
hmm

seanpdoyle [8:47 PM]
but we don’t want to write EVERYTHING to stderr, just errors and warns, right?

rwjblue [8:47 PM] 
just error should go to stderr

seanpdoyle [8:47 PM] 
does that mean we introduce another argument to CLI?

rwjblue [8:48 PM] 
but we don't thread stderr through anywhere

rwjblue [8:49 PM]
[...]

PR #5039, opened 2015-11-05
Add `UI.errorStream`

seanpdoyle:
If `ui.writeLine(error, 'ERROR')` is invoked, write to the
`ui.errorStream`, which defaults to `process.stderr`.

Closes [#5038].

[#5038]: https://github.com/ember-cli/ember-cli/issues/5038

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -44,8 +44,8 @@ function UI(options) {
   this.outputStream.setMaxListeners(0);
   this.outputStream.pipe(this.actualOuputStream);
 
-  // Input stream
   this.inputStream = options.inputStream;
+  this.errorStream = options.errorStream;
 
   this.writeLevel = options.writeLevel || 'INFO';
   this.ci = !!options.ci;
@@ -61,7 +61,9 @@ function UI(options) {
   @param {Number} writeLevel
 */
 UI.prototype.write = function(data, writeLevel) {
-  if (this.writeLevelVisible(writeLevel)) {
+  if (writeLevel === 'ERROR') {
+    this.errorStream.write(data);
+  } else if (this.writeLevelVisible(writeLevel)) {
     this.outputStream.write(data);
   }
 };
@@ -76,9 +78,7 @@ UI.prototype.write = function(data, writeLevel) {
   @param {Number} writeLevel
 */
 UI.prototype.writeLine = function(data, writeLevel) {
-  if (this.writeLevelVisible(writeLevel)) {
-    this.write(data + EOL);
-  }
+  this.write(data + EOL, writeLevel);
 };
 
 /**
--- a/bin/ember
+++ b/bin/ember
@@ -26,7 +26,8 @@ resolve('ember-cli', {
   cli({
     cliArgs: process.argv.slice(2),
     inputStream: process.stdin,
-    outputStream: process.stdout
+    outputStream: process.stdout,
+    errorStream: process.stderr
   }).then(function(result) {
     var exitCode = typeof result === 'object' ? result.exitCode : result;
     exit(exitCode);
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -41,6 +41,7 @@ module.exports = function(options) {
   var ui = new UI({
     inputStream:  options.inputStream,
     outputStream: options.outputStream,
+    errorStream:  options.errorStream || process.stderr,
     ci:           process.env.CI || /^(dumb|emacs)$/.test(process.env.TERM),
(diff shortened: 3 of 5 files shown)


lib/ui/index.js
```diff
@@ -79,6 +79,10 @@ UI.prototype.writeLine = function(data, writeLevel) {
   if (this.writeLevelVisible(writeLevel)) {
     this.write(data + EOL);
   }
+
+  if (/ERROR/i.match(writeLevel)) {
```
rwjblue: I think this section needs to be tweaked a bit.  Won't this write the output to both `this.outputStream` (from the line directly above) and to `this.errorStream`?

seanpdoyle: Yes. That's what we'd want, right? 

`STDOUT` to display everything (including errors), `STDERR` to display errors.

rwjblue: Honestly, I'm not sure.  My gut instinct would be that we would only want to log to one or the other, not both. We should just research what is "normal" in *nix land and do that.

lib/ui/index.js
```diff
@@ -64,6 +64,10 @@ UI.prototype.write = function(data, writeLevel) {
   if (this.writeLevelVisible(writeLevel)) {
     this.outputStream.write(data);
   }
+
+  if (/ERROR/i.match(writeLevel)) {
```
stefanpenner: wont this cause double output, since the above line (65) may also run.

stefanpenner:
good catch!

lib/ui/index.js
seanpdoyle: Please see https://github.com/ember-cli/ember-cli/pull/5039#discussion_r43968897

lib/ui/index.js
```diff
@@ -177,3 +183,17 @@ UI.prototype.writeLevelVisible = function(writeLevel) {
...
+
+  @method writeToErrorStream
+  @private
+  @param {String} writeLevel
+  @return {Boolean}
+*/
+UI.prototype.writeToErrorStream = function(writeLevel) {
+  writeLevel = writeLevel || 'INFO';
+
+  return writeLevel.match(/ERROR/i);
```
rwjblue: Is there a specific reason to use a regexp here? Would 

    return writeLevel === 'ERROR';

Work?

seanpdoyle: The idea was to guard against user error, but if you're ok with a `===` I am too.

seanpdoyle: @rwjblue if you're in favor of `===`, this method could be removed and the check can happen inline in the `write` method.

rwjblue: Ya, === is good for me.

rwjblue:
We still need to decide about https://github.com/ember-cli/ember-cli/pull/5039#discussion_r43968897. I feel that the output should only go to one stream (not both). I'm totally willing to be wrong if there is a general understanding in the *nix world that output should go to both...

lib/ui/index.js
seanpdoyle: > IMO, error output should only go to stderr, not to stdout.
> In a normal interative terminal session, you'd see everything twice if it went to both streams.

> - @georgebrock

rwjblue:
Changes look good to me. 

I'd like to get a test that checks the mock objects errors string. Doesn't need to be super complicated, just something that vaguely touches this area....

seanpdoyle:
This suite passed for all but `node@0.12.x`:

https://travis-ci.org/ember-cli/ember-cli/builds/89431587

Retrying.

Also, we were somehow green this morning without the test changes that were made in https://github.com/seanpdoyle/ember-cli/commit/894ed98aa120667cf2bc0f1a04d464be158f7a4b

How is that possible? I would have expected `tests/unit/ui/write-error-test.js` to fail since we were no longer writing to `outputStream`. Any ideas? Are these tests being skipped?

seanpdoyle:
The Appveyor Failures seem to be unrelated:

https://ci.appveyor.com/project/embercli/ember-cli/build/5385/job/3kjqa74p5ocw94jb#L2087

rwjblue:
@seanpdoyle - Can you confirm that the tests are being executed?  Seems like the changes made were required (but only after you tweaked the `write` method to avoid duplication) for the tests to pass...

seanpdoyle:
@rwjblue:

```diff
   it('no error', function() {
+    throw new Error('fail just because');
     writeError(ui);
 
     expect(ui.output).to.equal('');
```

```
  16) writeError no error:
     Error: fail just because
      at Context.<anonymous> (tests/unit/ui/write-error-test.js:18:11)
```

seanpdoyle:
@homu r+

PR #5052, opened 2015-11-10
Update minimum versions of ember-qunit / ember-cli-qunit.

rwjblue:


stefanpenner:
Should this be part of the pending 1.13.9 release?

rwjblue:
Quite a number of bugs were fixed between ember-qunit@0.4.10 and ember-qunit@0.4.16 (as well as support for tests against the future Ember 2.3). 

Depending on how hard it is to do, I would say yes (the changes are backwards compatible to all ember-cli 1.x and Ember versions).

stefanpenner:
@rwjblue thanks, will backport

PR #5061, opened 2015-11-11
Testem 0.9.11

johanneswuerbach:
Fixed Microsoft Edge detection. I would also propose to unlock testem again as patching its deps is no longer required, but thats your call.


rwjblue:
> I would also propose to unlock testem again as patching its deps is no longer required, but thats your call.

We generally only unlock deps that are > 1.0, as they have "declared stability".

johanneswuerbach:
k, re-added the strict lock.

PR #5065, opened 2015-11-11
Programmatic access to the CLI

eibrahim:
Added documentation on how to programmatically access the cli from a node application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,23 @@ Use `npm run docs` to build HTML and JSON documentation with YUIDoc and place it
 * [Change history of new Ember-CLI apps](https://github.com/kellyselden/ember-cli-output)
 * [Change history of new Ember-CLI addons](https://github.com/kellyselden/ember-addon-output)
 
+## Programmatic Access
+
+You can programmatically access the CLI from your node application.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +107,23 @@ Use `npm run docs` to build HTML and JSON documentation with YUIDoc and place it
...
+* Install the cli `npm install ember-cli -save`
+* Create index.js 
+* Use the cli
+```
+var cli = require("ember-cli/lib/cli");
+cli({
+    inputStream: process.stdin,
+    outputStream: process.stdout,
+    cliArgs: ['help']
+});
```
stefanpenner: we should indicate this returns a promise, maybe just by putting a `then`. We should also indicate that the fulfillment  value and rejection values are the exit cost

eibrahim: @stefanpenner thanks for the comment.  I made a change on a different pull request but I was trying to do it quickly in github.com and somehow ended up with 3 commits and a different PR... In any case, the PR is https://github.com/ember-cli/ember-cli/pull/5069 and hopefully you can collapse the 3 commits... if it's too much headache, I can reset and start from scratch.  tx. emad.

PR #5080, opened 2015-11-12
Support passing arbitrary options to moduleForAcceptance

trentmwillis:
We have some testing tools that require us to pass options into `module()` other than `beforeEach` and `afterEach` during our acceptance tests. Currently, `moduleForAcceptance` won't pass those through as it doesn't pass the entire object forward; this fixes that.

### DIFF
--- a/blueprints/app/files/tests/helpers/module-for-acceptance.js
+++ b/blueprints/app/files/tests/helpers/module-for-acceptance.js
@@ -3,21 +3,23 @@ import startApp from '../helpers/start-app';
 import destroyApp from '../helpers/destroy-app';
 
 export default function(name, options = {}) {
-  module(name, {
-    beforeEach() {
-      this.application = startApp();
+  const beforeEach = options.beforeEach;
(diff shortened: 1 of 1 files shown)


stefanpenner:
I am unconvinced this is a good default, it is quite abit more verbose.

some thoughts
* maybe, create your own addon internally that overrides this blueprint
* maybe come up with an alternative API that doesn't require this extra code gen.

trentmwillis:
@stefanpenner how would an addon work since this file is only generated at app creation?

My goal here isn't necessarily to introduce new behavior, but rather preserve existing behavior from QUnit's `module` which currently allows you to pass in additional information to it's test environment.

stefanpenner:
> but rather preserve existing behavior from QUnit's module

maybe this shouldn't be generated at all but a default implementation provided via lib, that can be overriden?

stefanpenner:
This seems to have gone stale.

Relevant note: generated code that has to change, or is visually complex should likely become library code

PR #5089, opened 2015-11-12
[BUGFIX] Forward `--config-file` test option to testem

cspanring:
Fixes an issue where the `--config-file` test option would not be forwarded to testem and the test runner would use the default config file (`./testem.json`) instead.

### DIFF
--- a/lib/tasks/test.js
+++ b/lib/tasks/test.js
@@ -42,6 +42,7 @@ module.exports = Task.extend({
       reporter: options.reporter,
       middleware: this.addonMiddlewares(),
       launch: options.launch,
+      file: options.configFile,
       /* jshint ignore:start */
       config_dir: process.cwd(),
       test_page: options.testPage,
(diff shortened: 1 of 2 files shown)


stefanpenner:
@cspanring a unit test for the task/test.js covering this would be nice

cspanring:
@stefanpenner done. 

Note: the existing assertions were never reached. I ended up refactoring the test a little to at least see if the options are transformed correctly.

jaswilli:
Would be great if this could get merged and released. I've tested against a project that relies on `--config-file` and it works as expected.

rwjblue:
@homu r+

rwjblue:
Failure is fixed on master (Node 5 failure due to npm@3.3.11 issue), merging...

rwjblue:
Thank you @cspanring (and @jaswilli for reminding us)!

PR #5095, opened 2015-11-13
Visual Studio fix

moander:


kellyselden:
Can we have a little more info on what problem this solves?

stefanpenner:
We don't produce commonjs, something smells fishy

moander:
Sorry for not explaining.

The jshint dependency enables the IDE to highlight errors and warning using the jshintrc and jshint bin instead of using the editor defaults.

jsconfig.json instructs the IDE to support ES6 IntelliSense/AutoComplete. I'm not sure what purpose the module setting have. The json schema is similar to the tsconfig.json file.

See this link for more information:
https://code.visualstudio.com/docs/languages/javascript

This is my first day of Ember so please bear with me :)

kellyselden:
@moander Welcome :smiley:

stefanpenner:
> This is my first day of Ember so please bear with me :)

First day of ember and already trying to help us make it better. I wish we have more like you!

@felixrieseberg is our resident windows/VS expert, I would love his thoughts.

blueprints/app/files/package.json
```diff
@@ -36,6 +36,7 @@
     "ember-data": "2.1.0",
     "ember-disable-proxy-controllers": "^1.0.1",
     "ember-export-application-global": "^1.0.4",
-    "ember-resolver": "^2.0.3"
+    "ember-resolver": "^2.0.3",
+    "jshint": "^2.8.0"
```
rwjblue: JSHint dep was moved down into broccoli-jshint so that it can properly use the JSHint API's without worrying about getting an invalid version.  I'd prefer to not add this back at the project level again.

felixrieseberg:
Thanks for the effort! In my humble opinion, let's not add JSHint - that's something we should just get better at without adding modules to Visual Studio. 

I do see the value in adding a jsconfig, but I *also* believe that we have stuff in store that will make this addition mostly unnecessary. Would you mind having this sit here for about 10 days so that we can see whether or not my proposed (and then public) fix will do it for y'all :egg:?

That being said, we should probably document this somewhere. I don't think we should have it *in* Ember Cli, but having a few lines somewhere telling you how to set Ember Cli up with VS/JetBrains/etc would probably be nice.

rwjblue:
@felixrieseberg - Once you have confirmed that your changes do or do not work, whatever config is left could sit nicely in an editor vendor addon (ala `ember-cli-visual-studio`)...

moander:
Oh.. Sorry (again). The problems solved are related to Visual Studio Code. 

I had to make some additional changes to a third file (.vscode/settings.json) for VS Code to behave as expected, so after discovering the blueprint/add-on feature I would argue that these patches should be placed in it's own module.

If you agree I will close this pr and keep the branch until I figure out how these add-ons work :)

rwjblue:
I think it makes sense to keep the PR around while @felixrieseberg figures out the best plan forward.

stefanpenner:
Sounds good. Let's let it stand for now

stefanpenner:
Going to close this. Adding editor specific config isn't something we should add by default

An addon could provide it via a generator + blueprint

PR #5098, opened 2015-11-13
[ENHANCEMENT] Install htmlbars precompiler when generating component integration tests

drspaniel:
A project originally generated with an older version of ember-cli might not have the inline precompiler installed. This means a generated component integration test won't work out of the box.

@rwjblue

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -3,6 +3,7 @@
 var path          = require('path');
 var testInfo      = require('ember-cli-test-info');
 var stringUtil    = require('ember-cli-string-utils');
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1:

stefanpenner:
I like the idea :). Gotta figure out what's up with appveyor

drspaniel:
It looks like build just timed out because the test suite is so long

rwjblue:
Yeah, the appveyor node v0.12 build passed but the v4.2 one timed out at the maximum build run time of 40 min.  I restarted it, but believe that it is fine.

@homu r+

PR #5105, opened 2015-11-15
Make npm/bower install more copy/paste friendly

knownasilya:


rwjblue:
I'm fairly sure && doesn't work on all supported platforms.

knownasilya:
Does `rm -rf` work one windows? I assumed that those instructions were not for all OSs. But `&&` does work on windows http://stackoverflow.com/a/8055430/483616

jonnii:
@knownasilya `&&` only works in `cmd.exe`, doesn't work if the user is using powershell.

knownasilya:
What about `rm -rf`, which is already in the instructions.

jonnii:
@knownasilya I _think_ so? I can test it tomorrow when I'm at work. I know `rm` exists as an alias, but there are two flags to `remove-item` that start with `f` (filter and force) so not sure if you need to disambiguate by doing something like `rm -force -recurse`.

stefanpenner:
also doesn't work in fish shell

PR #5106, opened 2015-11-15
Configure Hound to lint with JSCS

seanpdoyle:
[Hound] is a [continuous integration service][site] designed to comment
during Code Review on Pull Requests with [style guide
infractions][why-does-style-matter] in their diffs.

We're currently linting our JavaScript with [CodeClimate] (which seems
to use JSHint), but unfortunately those infractions are separate from
the typical Code Review workflow that occurs in GitHub, and are instead
buried deep within the CodeClimate website.

Alternatively, we could introduce JSCS linting into the build process,
the way `emberjs/ember.js` and `emberjs/data` do.

This approach would add additional tests to an already lengthy test
suite. Local style infractions would slow down the development feedback
loop, and could introduce terminal noise, which would distract the
developer from the task at hand.

Dealing with style guide infractions are the Code Review / Pull Request
delays the checks until the task at hand is complete and ready for CI.

Maintaining a consistent style would help [make collaboration
easier][collaboration], as well as make the codebase more approachable
for new contributors.

[Hound]: https://robots.thoughtbot.com/introducing-hound
[site]: https://houndci.com
[why-does-style-matter]: https://robots.thoughtbot.com/why-does-style-matter
[collaboration]: https://robots.thoughtbot.com/collaboration-is-other-people
[CodeClimate]: https://codeclimate.com/github/ember-cli/ember-cli/issues/categories/style

# NOTE

The `.jscsrc` file is intentionally left blank, since I don't personally have a sense of the project's currently de facto styleguide / rules.

I encourage someone in-the-know to help me generate a good baseline configuration.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,3 @@
+jscs:
+  enabled: true
+  config_file: .jscsrc
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,77 @@
+{
+  "esnext": true,
+  "verbose": true,
+
+  "disallowEmptyBlocks": true,
+  "disallowKeywordsOnNewLine": ["else"],
+  "disallowMultipleLineBreaks": true,
+  "disallowMultipleVarDecl": "exceptUndefined",
+  "disallowNewlineBeforeBlockStatements": true,
+  "disallowOperatorBeforeLineBreak": ["."],
+  "disallowSpaceAfterObjectKeys": true,
+  "disallowSpaceAfterPrefixUnaryOperators": true,
+  "disallowSpaceBeforePostfixUnaryOperators": true,
+  "disallowSpaceBeforeSemicolon": true,
+  "disallowSpacesInFunction": {
+    "beforeOpeningRoundBrace": true
+  },
+  "disallowSpacesInsideParentheses": true,
+  "disallowSpacesInCallExpression": true,
+  "disallowTrailingComma": true,
+  "disallowTrailingWhitespace": true,
+  "requireBlocksOnNewline": true,
+  "requireCamelCaseOrUpperCaseIdentifiers": true,
+  "requireCapitalizedConstructors": true,
+  "requireCommaBeforeLineBreak": true,
+  "requireCurlyBraces": [
+    "if",
+    "else",
+    "for",
+    "while",
+    "do",
+    "try",
+    "catch",
+    "switch"
+  ],
+  "requireDotNotation": true,
+  "requireLineBreakAfterVariableAssignment": true,
+  "requireParenthesesAroundArrowParam": true,
+  "requireSemicolons": true,
+  "requireSpaceAfterBinaryOperators": true,
+  "requireSpaceAfterKeywords": [
+    "do",
+    "for",
+    "if",
+    "else",
+    "switch",
+    "case",
+    "try",
+    "void",
+    "while",
+    "with",
+    "return",
+    "typeof"
+  ],
+  "requireSpaceAfterLineComment": true,
+  "requireSpaceBeforeBinaryOperators": true,
+  "requireSpaceBeforeBlockStatements": true,
+  "requireSpaceBeforeKeywords": [
+    "else",
+    "while",
+    "catch"
+  ],
+  "requireSpaceBeforeObjectValues": true,
+  "requireSpaceBetweenArguments": true,
+  "requireSpacesInFunction": {
+    "beforeOpeningCurlyBrace": true
+  },
+  "requireSpacesInConditionalExpression": true,
+  "requireSpacesInForStatement": true,
+  "requireSpacesInsideObjectBrackets": "all",
+  "validateIndentation": 2,
+  "validateParameterSeparator": ", ",
+  "validateQuoteMarks": {
+    "mark": "'",
+    "escape": true
+  }
+}


stefanpenner:
> This approach would add additional tests to an already lengthy test
> suite.

im curious how long

> Local style infractions would slow down the development feedback
loop, and could introduce terminal noise, which would distract the
developer from the task at hand.

but it may fix the problem ASAP, rather then post push

stefanpenner:
> Maintaining a consistent style would help make collaboration
> easier, as well as make the codebase more approachable
> for new contributors.

yes very much +1

seanpdoyle:
@stefanpenner 

> but it may fix the problem ASAP, rather then post push

This is usually a good idea for tools like JSHint that help ensure code correctness and help avoid JS anti patterns and bad practices.

However, JSCS is more concerned with *style* (like whitespace, commas, braces, etc).

These types of infractions are less pressing, and would serve as distractions, demanding developer brain / computer CPU cycles that could detract from the task at hand.

seanpdoyle:
@stefanpenner I'm certainly open to the idea of adding this to the build process the way ember and ember-data do (although I plan on opening similar PRs against those repos as well).

Either way, I think JSCS linting is a good idea, no matter where / when it happens.

My only issue is I'm not sure what to put into the JSCS config file. That's where I'm hoping someone involved with the project with strong opinions could help.

seanpdoyle:
> Local style infractions would slow down the development feedback
loop, and could introduce terminal noise, which would distract the
developer from the task at hand.

> -- seanpdoyle

> but it may fix the problem ASAP, rather then post push

> -- stefanpenner

It's also worth noting that Hound will **only comment on lines in the diff**. This means we could be more relaxed about existing infractions, correcting them over time as the code they live in changes.

If we added JSCS linting to the build process, we'd have to take a boil-the-ocean approach to fixing existing style guide infractions.

raido:
> My only issue is I'm not sure what to put into the JSCS config file. That's where I'm hoping someone involved with the project with strong opinions could help.

-- seanpdoyle 

Could it based on ember-suave?

seanpdoyle:
@raido yes, with the caveat that Hound couldn't support the [custom rules](https://github.com/dockyard/ember-suave/blob/master/lib/jscsrc.json#L5-L7) just yet.

I've updated the PR to include the contents of `ember-suave`'s [JSCS config](https://github.com/dockyard/ember-suave/blob/master/lib/jscsrc.json#L5-L7), with the custom rules removed.

jakecraige:
In the long term, I think a combination of local and PR based violations would be ideal. 

Being able to see it locally before a PR is opened and on the PR if you miss something. In my experience even with local violations(assuming it doesn't completely fail a build) they sometimes still make it into a PR and it's nice to know that Hound has got your back when that happens.

> If we added JSCS linting to the build process, we'd have to take a boil-the-ocean approach to fixing existing style guide infractions.

This seems like a big reason that adding Hound would be a great first step towards a consistently enforced style and iteratively updating the code to follow it where it currently doesn't without too much disruption.

rwjblue:
> I think a combination of local and PR based violations would be ideal.

I agree. To me having a CI run something that I am incapable of running local (to fix/confirm) would be massively frustrating.

stefanpenner:
> I agree. To me having a CI run something that I am incapable of running local (to fix/confirm) would be massively frustrating.

yup it would be.



> If we added JSCS linting to the build process, we'd have to take a boil-the-ocean approach to fixing existing style guide infractions.

or we can gradually tighten the rules, I think this approach will be the most successful.
It really wouldn't be much work for us to slowly tighten up the rules.

rwjblue:
@lukemelia @chrislopresto and I had a call a little while back to come up with best plan to update a code base to use ember-suave (or any opinionated JSCS preset). 

The conclusion was that we should chose the rules we want and add:

    // jscs ignore:file

Comments to each file in a single PR. Then work through one file at a time until the comment can be removed. This gives us a few things:

* Makes adding new files conform to the full rule set. 
* Reduces massive PR's that try to touch all files. 
* Provides a super easy way to track which files still need to be cleaned up.
* Allows files to be fixed individually (much easier to review and land quickly to avoid rebase hell). 

@stefanpenner - Any objection to be above plan of attack?

stefanpenner:
> The conclusion was that we should chose the rules we want and add:
> 
> // jscs ignore:file

:+1: 

cc @kellyselden

kellyselden:
+1 for Hound, for those of us that commit without running the full test suite locally :smile: 

There was a debate in Slack whether JSCS or ESLint is the way to go. ESLint could probably replace both JSHint and JSCS, but I believe we have more say in JSHint and JSCS development for dogfeeding purposes.

@stefanpenner @rwjblue Do we want to continue the debate here, or go ahead with this JSCS PR?

kellyselden:
or this could be upgraded to an RFC?

Turbo87:
> The conclusion was that we should chose the rules we want

are you following the same rules as in https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md? in that case some of the rules are pretty much already decided.

kellyselden:
So I added JSCS tests in #5435. PRs are now style guide tested, although the rules aren't fully populated yet. One of the concerns with hound was yet another github PR hook. I think the .jscsrc file can be removed from this PR, and we can let this stay open for a while while we assess the mocha tests only route for style checking.

Turbo87:
> The conclusion was that we should chose the rules we want and add:
> 
> // jscs ignore:file

@kellyselden what happend to the plan above? adding those ignore comments now would essentially revert your PR because nothing would be checked anymore.

what would be the path forward now? adding useful rules to the JSCS config and fixing the resulting issues directly in each commit?

/cc @rwjblue @stefanpenner

kellyselden:
@Turbo87 We discussed it in one of our team meetings. Basically we didn't want another github PR hook if there was another way. So I just did minimum viable by getting JSCS running in our tests. I didn't want to solve the problem of what our rules are just yet. We can still go the route of jscsignoring all files and filling in rules.

I'll bring it up again at our meeting today. You're welcome to join :smile:

Turbo87:
I guess this PR is obsolete now, right?

kellyselden:
Hound is shelved for now. We decided to go with an ESLint only solution instead of any JSHint/JSCS combo. Although Hound supports ESLint, we have our rules running on Travis and feel that's alright... for now....

PR #5107, opened 2015-11-15
[FEATURE] Implement addon black- and whitelist

dschmidt:
... per RFC https://github.com/ember-cli/rfcs/blob/master/active/0000-addon-black-and-whitelist-for-apps.md

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -170,7 +170,8 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
     jshintrc: {},
     'ember-cli-qunit': {
       disableContainerStyles: false
-    }
+    },
+    addons: {}
   };
 
   this.options = defaults(options, config, babelOptions);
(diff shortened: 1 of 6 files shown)


lib/broccoli/ember-app.js
```diff
@@ -341,7 +341,15 @@ EmberApp.prototype._notifyAddonIncluded = function() {
   this.project.addons = this.project.addons.filter(function(addon) {
     addon.app = this;
 
-    if (!addon.isEnabled || addon.isEnabled()) {
+    var addonEnabled = !addon.isEnabled || addon.isEnabled();
+
+    if (addonEnabled && this.options.addonBlacklist && this.options.addonBlacklist.indexOf(addon.name) !== -1) {
```
stefanpenner: lots of complexity in these lines, let me recommend this be extract to its own function, which then itself can be nicely unit tested.

stefanpenner:
question, thoughts on:

```js
addons: {
  blacklist: [...],
  whiltelist: [...]
};
```
?

lib/broccoli/ember-app.js
dschmidt: Okies, will do.

dschmidt:
Dunno, I don't care tbh :trollface:

stefanpenner:
i prefer `addons: { ...}` @rwjblue r?

dschmidt:
The naming `disabledByBlacklisted` and `disabledByWhitelist` might seem weird at first, but even while writing the code `blacklisted` and `whitelisted` seemed very ambiguous.
Blacklisted would be rather clear, but what is whitelisted? Is an addon only whitelisted if there is a whitelist? If so, ifs get rather complicated and unreadable again.

This is why I went for these rather bulky var names.

stefanpenner:
Should we also error if an addon is mentioned, that is not actually an addon? Might be a great way to catch bugs.

dschmidt:
I'm afraid, you are right. I thought about it, but was too lazy ... will do :P

dschmidt:
Updated

stefanpenner:
SGTM

dschmidt:
Cool, will squash commits once tests are complete

rwjblue:
LGTM

stefanpenner:
@homu r+

PR #5108, opened 2015-11-16
Fix instance-intitalizer-test test-info import to use `ember-cli-test-info`

trabus:
Currently the instance-initializer-test is importing test-info from a local path, which no longer exists.


rwjblue:
@homu r+

PR #5109, opened 2015-11-16
Fix body-parser suggested use due to deprecations

knownasilya:
BodyParser has changed how it expects to be called, so this
follows that change to use `.json()` directly.


stefanpenner:
@homu r+

PR #5111, opened 2015-11-16
[FIXES #5110] set null integrity value for tests.js

### LINKED ISSUE #5110: reloading tests fails on chrome when using ember test --server
I have a repo demonstrating the failure, including steps to reproduce, and workaround/fix: https://github.com/kenglxn/ember-cli-reload-bug

Basically when making changes to a test, or adding a new test, tests blow up in chrome due to integrity check with the following message:

`Uncaught Error: Assertion Failed: The tests file was not loaded. Make sure your tests index.html includes "assets/tests.js"`

browser

console

kenglxn:


rwjblue:
While this is a good temporary solution for this issue in your application, I do not believe this is the right fix.  We need to figure out how to get the hash updated when the asset fingerprint changes.

/cc @jonathanKingston

kenglxn:
I forgot to mention that it only blows up in chrome, in phantomJS the failure is shown as expected.

rwjblue:
Chrome checks the integrity and phantom does not (this is why it is only affecting chrome).

jonathanKingston:
Argh, sorry if it wasn't for @rwjblue on slack I would have missed this completely thanks to Inbox folding my messages away.

I'm looking into this thanks.

rwjblue:
Thank you @jonathanKingston!

stefanpenner:
kicking av

rwjblue:
Closing PR for now, will update issue with details.

PR #5118, opened 2015-11-17
[ENHANCEMENT] Changes to support `ember-cli-blueprint-test-helpers`

trabus:
This PR contains changes to support `ember-cli-blueprint-test-helpers`. The primary changes are for the `tests/helpers/ember` test helper, adding the following:

* the ability to capture and return UI output, including input, output, errors, and the statusCode, so errors can be asserted.
* add the ability to disable the `ember-cli-dependency-checker` through the `disableDependencyChecker` option. Will require [this PR](https://github.com/quaertym/ember-cli-dependency-checker/pull/49) to be merged, and a new version to be cut.
* allow the `package.json` path to be passed in, allowing the ember-cli commands to be run in different project structures (`app` vs `addon`). Some modifications to the `project` model were necessary for it to allow the passed in `package.json` to be discovered.

It was also necessary to modify the `addon-import` blueprint for some of the changes done in extracting the blueprints to `ember-cli-legacy-blueprints`.

This is not ready for merge yet, it still has a couple tweaks needed, as well as some general cleanup. It should hopefully be ready later tonight.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -569,20 +569,35 @@ Project.closest = function(pathName, _ui, _cli) {
  */
 Project.closestSync = function(pathName, _ui, _cli) {
   var ui = ensureUI(_ui);
-
-  try {
-    var directory = findup.sync(pathName, 'package.json');
-    var pkg = require(path.join(directory, 'package.json'));
-
-    if (pkg && pkg.name === 'ember-cli') {
-      return Project.nullProject(_ui, _cli);
+  var directory, pkg;
+  
+  if (_cli && _cli.testing) {
+    directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
+    
+    if (directory) {
+      pkg = require(path.join(directory, 'package.json'));
+    } else {
+      pkg = {name: 'ember-cli'};
     }
-
-    debug('closestSync %s -> %s', pathName, directory);
-    return new Project(directory, pkg, ui, _cli);
-  } catch(reason) {
-    handleFindupError(pathName, reason);
+ 
+  } else {
+    try {
+      directory = findup.sync(pathName, 'package.json');
+      pkg = require(path.join(directory, 'package.json'));
+       
+    } catch(reason) {
+      handleFindupError(pathName, reason);
+    }
   }
+  
+  debug('dir' + directory);
+  debug('pkg: %s', pkg);
+  if (pkg && pkg.name === 'ember-cli') {
+    return Project.nullProject(_ui, _cli);
+  }
+
+  debug('closestSync %s -> %s', pathName, directory);
+  return new Project(directory, pkg, ui, _cli);
 };
 
 /**
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -571,15 +571,25 @@ Project.closestSync = function(pathName, _ui, _cli) {
   var ui = ensureUI(_ui);
   var directory, pkg;
   
+  // special lookup rules only when testing
   if (_cli && _cli.testing) {
-    directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
-    
-    if (directory) {
-      pkg = require(path.join(directory, 'package.json'));
-    } else {
-      pkg = {name: 'ember-cli'};
-    }
- 
+      directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
+      
+      if (directory) {
+        pkg = require(path.join(directory, 'package.json'));
+      // if we're anywhere inside the app, use findup
+      } else if (pathName.indexOf('/app') > -1){
+        try {
+          directory = findup.sync(pathName, 'package.json');
+          pkg = require(path.join(directory, 'package.json'));
+           
+        } catch(reason) {
+          handleFindupError(pathName, reason);
+        }
+      } else {
+        pkg = {name: 'ember-cli'};
+      }
+      
   } else {
     try {
       directory = findup.sync(pathName, 'package.json');
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -571,35 +571,22 @@ Project.closestSync = function(pathName, _ui, _cli) {
   var ui = ensureUI(_ui);
   var directory, pkg;
   
-  // special lookup rules only when testing
   if (_cli && _cli.testing) {
-      directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
-      
-      if (directory) {
-        pkg = JSON.parse(fs.readFileSync(path.join(directory, 'package.json')));
-      // if we're anywhere inside the app, use findup
-      } else if (pathName.indexOf('/app') > -1){
-        try {
-          directory = findup.sync(pathName, 'package.json');
-          pkg = JSON.parse(fs.readFileSync(path.join(directory, 'package.json')));
-           
-        } catch(reason) {
-          handleFindupError(pathName, reason);
-        }
+    directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
+    if (!directory) {
+      if (pathName.indexOf('/app') > -1){
+        directory = findupPath(pathName);
       } else {
         pkg = {name: 'ember-cli'};
       }
-      
-  } else {
-    try {
-      directory = findup.sync(pathName, 'package.json');
-      pkg = require(path.join(directory, 'package.json'));
-       
-    } catch(reason) {
-      handleFindupError(pathName, reason);
     }
+  } else {
+    directory = findupPath(pathName);
   }
-  
+  if (!pkg) {
+    pkg = JSON.parse(fs.readFileSync(path.join(directory, 'package.json')));
+  }
+
   debug('dir' + directory);
   debug('pkg: %s', pkg);
   if (pkg && pkg.name === 'ember-cli') {
@@ -696,6 +683,14 @@ function closestPackageJSON(pathName) {
         pkg: require(path.join(directory, 'package.json'))
       });
     });
+}
+
+function findupPath(pathName) {
+  try {
+    return findup.sync(pathName, 'package.json');       
+  } catch(reason) {
+    handleFindupError(pathName, reason);
+  }
 }
 
 function isFindupError(reason) {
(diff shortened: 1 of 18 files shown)


trabus:
I added some additional fixes for the `addon-import` blueprint, specifically to prevent it from generating an `addon-import` for the `blueprint-test` blueprint, and also to prevent it from being generated directly. 

I also added some tests to fix the coverage, as some of my changes exposed some un-covered code.

stefanpenner:
> Will require this PR to be merged, and a new version to be cut.

merged, but waiting for a release (i dont have rights for that)

lib/commands/new.js
```diff
@@ -86,7 +86,7 @@ module.exports = Command.extend({
         return initCommand
           .run(commandOptions, rawArgs)
           .catch(function(err){
-            console.log(arguments)
+            console.log(arguments);
```
stefanpenner: likely not required

tests/helpers/mock-ui.js
```diff
@@ -25,6 +32,7 @@ MockUI.prototype.constructor = MockUI;
 MockUI.prototype.clear = function(){
   this.output = '';
   this.errors = '';
+  this.errorLog = [];
```
stefanpenner: should this also clear outputStram/errorStream?

tests/helpers/mock-ui.js
```diff
@@ -5,16 +5,23 @@ var through = require('through');
 var Promise = require('../../lib/ext/promise');
 
 module.exports = MockUI;
-function MockUI() {
+function MockUI(options) {
   this.output = '';
   this.errors = '';
+  this.errorLog = options && options.errorLog || [];
```
stefanpenner: is this in addition to errorStream?

tests/helpers/mock-ui.js
trabus: I think I did have it storing and clearing those at one point, but it doesn't store the value defined in options now, so there's no way to reference it. I should probably rename the options to something less confusing. It's expecting an array for that value so it can push the messages to it.

lib/models/project.js
```diff
@@ -569,20 +569,45 @@ Project.closest = function(pathName, _ui, _cli) {
...
+  if (_cli && _cli.testing) {
+      directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
+      
+      if (directory) {
+        pkg = require(path.join(directory, 'package.json'));
+      // if we're anywhere inside the app, use findup
+      } else if (pathName.indexOf('/app') > -1){
+        try {
+          directory = findup.sync(pathName, 'package.json');
+          pkg = require(path.join(directory, 'package.json'));
```
stefanpenner: lets fs.readFileSync + JSON.parse so we don't cache this file.

trabus: Okay, will do.

trabus:
@stefanpenner @rwjblue I'm not quite sure why my changes are causing this, but node 4.2 is passing all the tests but not finishing for some reason: https://travis-ci.org/ember-cli/ember-cli/jobs/97388659#L2234

This PR is a prerequisite to getting the blueprint helper working properly, and should probably precede any in-the-wild usage of `ember-cli-blueprint-test-helpers`. Without these changes, it is unable to test generating blueprints into an addon properly, and cannot assert expected errors.

stefanpenner:
@trabus i can take a look later today.

trabus:
@stefanpenner thanks!

trabus:
I just noticed this seems to be affecting other recent PRs as well, so it must be something in master. Let me know if there is anything I can do from my end.

trabus:
@stefanpenner I spent several hours this weekend trying to debug why node 4.x isn't exiting and came up empty. I'm going to keep at it this week, pulling apart my changes to figure out where this is going wrong. I'm thinking it may be one of the tests I added to fix the coverage failures.

We will definitely need these changes in place before we get the Ember addonification completed though, as `ember-cli-blueprint-test-helpers` doesn't function properly without them.

trabus:
@stefanpenner I finally tracked down the test that is getting stuck: https://github.com/ember-cli/ember-cli/blob/master/tests/unit/settings-file/leek-options-test.js

I'm not exactly sure why it's not exiting, though I assume it has something to do with my changes to MockUI. Anyhow, I'll keep digging into it to try and figure it out tomorrow.

stefanpenner:
@trabus so removing that test, and all works?

trabus:
@stefanpenner I believe so, yes. All the tests were already passing, it's just that they weren't exiting, so the runner would hang indefinitely.

trabus:
@stefanpenner and actually, that test passes as well, it just hangs the runner.

tests/helpers/mock-ui.js
```diff
@@ -5,16 +5,23 @@ var through = require('through');
...
     inputStream: through(),
     outputStream: through(function(data) {
+      if (options && options.outputStream) {
+        options.outputStream.push(data);
+      }
       this.output += data;
     }.bind(this)),
     errorStream: through(function(data) {
+      if (options && options.errorStream) {
+        options.errorStream.push(data);
```
trabus: @stefanpenner it's this line that's holding things up. I'm going to look through and make sure that I actually need it, and if not, I'll just remove it.

stefanpenner:
@trabus awesome, that narrows it down some. I should have some time tonight or tomorrow morning to debug further if no resolution is reached.

trabus:
@stefanpenner ITS GOOD!

So the AppVeyor failures are due to something that came from this PR: https://github.com/ember-cli/ember-cli/pull/5345

It seems that changing the EOL changed what the blueprint tests were expecting.

stefanpenner:
I will review, we should likely look into addressing the Windows stuff soon

kellyselden:
@stefanpenner feel free to bug me about that

lib/models/project.js
```diff
@@ -616,20 +616,45 @@ Project.closest = function(pathName, _ui, _cli) {
...
+          handleFindupError(pathName, reason);
+        }
+      } else {
+        pkg = {name: 'ember-cli'};
+      }
+      
+  } else {
+    try {
+      directory = findup.sync(pathName, 'package.json');
+      pkg = require(path.join(directory, 'package.json'));
```
stefanpenner: we should readFileSync here as well, instead of require

lib/models/project.js
```diff
@@ -616,20 +616,45 @@ Project.closest = function(pathName, _ui, _cli) {
...
-    var directory = findup.sync(pathName, 'package.json');
-    var pkg = require(path.join(directory, 'package.json'));
-
-    if (pkg && pkg.name === 'ember-cli') {
-      return Project.nullProject(_ui, _cli);
+  var directory, pkg;
+  
+  // special lookup rules only when testing
+  if (_cli && _cli.testing) {
+      directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
```
stefanpenner: so the idea is, if we are testing, and their is a pathName/package.json, we should instead prefer process.cwd(); ?

stefanpenner: this function seems to be more complex then needed, we have 3 different sub-routines that read + parse the package.json.

lib/models/project.js
trabus: Will do.

lib/models/project.js
trabus: > so the idea is, if we are testing, and their is a pathName/package.json, we should instead prefer process.cwd(); ?

I don't recall specifically why I had to do this, but I agree that it is funky. Will get rid of it with the refactor of the function.

I think my intent was to not change or disturb the existing functionality, but I agree that it should be simpler. I will refactor this later tonight.

stefanpenner: > I think my intent was to not change or disturb the existing functionality, but I agree that it should be simpler. I will refactor this later tonight.

Ya thats a good reason, but I feel unifying will reduce the chance of test vs prod trolling

trabus:
@stefanpenner I refactored `project.closestSync` last night, please let me know if [this](https://github.com/trabus/ember-cli/commit/4bdb2a46a137d7d3f4cccd8a9be87a673009768d) is any better.

lib/models/project.js
```diff
@@ -616,20 +616,40 @@ Project.closest = function(pathName, _ui, _cli) {
...
-
-  try {
-    var directory = findup.sync(pathName, 'package.json');
-    var pkg = require(path.join(directory, 'package.json'));
-
-    if (pkg && pkg.name === 'ember-cli') {
-      return Project.nullProject(_ui, _cli);
+  var directory, pkg;
+  
+  function findupPath() {
```
stefanpenner: can we define this as a standalone method, that takes pathName as an argument?

stefanpenner:
one last comment then :+1:

lib/models/project.js
trabus: Will do.

stefanpenner:
this seems good now. I have fixed the failing test [here](https://github.com/ember-cli/ember-cli/commit/e50b1915675b85c9f4994eb3124727a5bee5fa89) a rebase and this should be green on posix.

trabus:
@stefanpenner thanks, I've rebased my branch.

stefanpenner:
@homu r+

trabus:
Next steps will be to get `ember-cli-blueprint-test-helpers` cut with the next ember-cli beta version as a dependency. I will also need to clean up the blueprint installed dependencies, and get `ember-cli-legacy-blueprints` completed and cut (doing a final pass to bring in any blueprint changes in ember-cli) so we can bring it back into ember-cli and start removing blueprints.

pangratz:
Thanks for the hard work on this @trabus!

lib/models/project.js
```diff
@@ -569,20 +569,32 @@ Project.closest = function(pathName, _ui, _cli) {
...
-    var pkg = require(path.join(directory, 'package.json'));
-
-    if (pkg && pkg.name === 'ember-cli') {
-      return Project.nullProject(_ui, _cli);
+  var directory, pkg;
+  
+  if (_cli && _cli.testing) {
+    directory = existsSync(path.join(pathName, 'package.json')) && process.cwd();
+    if (!directory) {
+      if (pathName.indexOf('/app') > -1){
```
Turbo87: `/app` doesn't work on Windows and seems to cause two failing tests. I've fixed it by using `path.sep + 'app'` instead and will push that solution in a moment.

trabus: Sorry about that. Thanks for catching and fixing.

Turbo87: sure, no problem. for reference: 9aa6b09f17895d02efac9a91d52ed2c4fdf0a28d

PR #5120, opened 2015-11-17
Update loader.js to prevent deprecation.

rwjblue:
/cc @twokul this needs to end up in 1.13.13


twokul:
@rwjblue updated

rwjblue:
Node 5 issue is from https://github.com/ember-cli/ember-cli/issues/5126.

PR #5130, opened 2015-11-19
Update bower.json

rwjblue:
Fix invalid resolution value in `bower.json`


rwjblue:
@homu r+

stefanpenner:
@rwjblue whats requiring the resolution in the first place, why is there a conflict.

Also, rebase and the tests will likely be green.

rwjblue:
@stefanpenner - TBH, I'm not sure why the resolution is needed, I just noticed that we updated ember in dependencies but not resolutions.

PR #5137, opened 2015-11-19
Bump ember-cli package.json version number to 1.13.13

patocallaghan:
I'm presuming this should be in-step with the release number?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-cli",
-  "version": "1.13.8",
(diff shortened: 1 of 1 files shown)


rwjblue:
Ya, though I wish we could make it `2.2.0-alpha` or something.

stefanpenner:
* master and 1.13.x have diverged, master will be 2.x
* bumping to 2.x version before it is released will break tests related to NPM.

PR #5139, opened 2015-11-20
Use ES6 shorthand property for layout in component blueprint

martndemus:
Because my linter (ember-suave) doesn't like non-shorthand properties that should have been shorthanded.


rwjblue:
@homu r+

PR #5140, opened 2015-11-20
Pr 4963

stefanpenner:


stefanpenner:
The original issue is fixed, but there are more scenarios where this hits. I will try a quick-fix, but maybe a patch/fork of node-quick-temp is required.

stefanpenner:
node-quick-temp is released

PR #5144, opened 2015-11-21
[ENHANCEMENT] Cleans up generated application directory if `ember new` errors

trek:
If `ember new` errors, removes the partially generated project instead of leaving behind in a partially generated state.  Useful if generating from a blueprint and the blueprint is broken.


rwjblue:
@homu r+

stefanpenner:
What happens if npm or bower fail?

rwjblue:
Pft, we all know NPM could never fail!

PR #5146, opened 2015-11-22
[ENHANCEMENT] Skip bower/npm install on blueprint install if manifests are missing

trek:
I couldn't think of a more useful test :-/

### DIFF
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -1,9 +1,9 @@
 'use strict';
 
 // Runs `bower install` in cwd
-
-var Promise = require('../ext/promise');
-var Task    = require('../models/task');
+var existsSync     = require('exists-sync');
+var Promise        = require('../ext/promise');
+var Task           = require('../models/task');
 
 module.exports = Task.extend({
   init: function() {
@@ -14,10 +14,16 @@ module.exports = Task.extend({
   run: function(options) {
     var chalk          = require('chalk');
     var bower          = this.bower;
+    var bowerJson      = this.project.root + '/bower.json';
     var bowerConfig    = this.bowerConfig;
     var ui             = this.ui;
     var packages       = options.packages || [];
     var installOptions = options.installOptions || { save: true };
+
+    if (!existsSync(bowerJson)) {
+      ui.writeWarnLine('Skipping bower install: bower.json not found');
+      return;
+    }
 
     ui.startProgress(chalk.green('Installing browser packages via Bower'), chalk.green('.'));
 
--- a/lib/tasks/npm-install.js
+++ b/lib/tasks/npm-install.js
@@ -1,11 +1,23 @@
 'use strict';
 
 // Runs `npm install` in cwd
-
-var NpmTask = require('./npm-task');
+var path         = require('path');
+var existsSync   = require('exists-sync');
+var NpmTask      = require('./npm-task');
 
 module.exports = NpmTask.extend({
   command: 'install',
   startProgressMessage: 'Installing packages for tooling via npm',
-  completionMessage: 'Installed packages for tooling via npm.'
+  completionMessage: 'Installed packages for tooling via npm.',
+  run: function(){
+    var ui = this.ui;
+    var packageJson = path.join(this.project.root, 'package.json');
+    if (!existsSync(packageJson)) {
+
+      ui.writeWarnLine('Skipping npm install: package.json not found');
+      return;
+    } else {
+      return this._super.run.apply(this, arguments);
+    }
+  }
 });
--- a/tests/unit/tasks/bower-install-test.js
+++ b/tests/unit/tasks/bower-install-test.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var BowerInstallTask = require('../../../lib/tasks/bower-install');
+var MockUI           = require('../../helpers/mock-ui');
+var MockProject      = require('../../helpers/mock-project');
+var expect           = require('chai').expect;
+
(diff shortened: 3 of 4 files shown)


lib/tasks/npm-install.js
```diff
@@ -1,11 +1,23 @@
...
 
 module.exports = NpmTask.extend({
   command: 'install',
   startProgressMessage: 'Installing packages for tooling via npm',
-  completionMessage: 'Installed packages for tooling via npm.'
+  completionMessage: 'Installed packages for tooling via npm.',
+  run: function(){
+    var ui = this.ui;
+
+    if (!existsSync('./package.json')) {
```
rwjblue: think this belongs in the npm task we invoke below?

rwjblue:
Seems like these tests would pass without these changes since they are using --skip-npm and --skip-bower. Am I missing something here?

lib/tasks/npm-install.js
trek: I didn't want to expand the scope beyond install failures on `package.json`-less blueprints. I can update to have it warn on all npm commands, if you want.

trek:
> Seems like these tests would pass without these changes since they are using --skip-npm and --skip-bower. Am I missing something here?

The test for npm only includes `--skip-bower`, the test for bower only includes `--skip-npm`.

lib/tasks/npm-install.js
rwjblue: Seems ok, I just wanted to make sure it was intentionally limited to install.

rwjblue:
> The test for npm only includes --skip-bower, the test for bower only includes --skip-npm.

D'oh! I missed that :crying_cat_face:

trek:
AppVeyor build failure is a timeout.

lib/tasks/npm-install.js
stefanpenner: people can invoke tasks no at the root of the project, project.root + '/package.json' will likely need to be used

lib/tasks/npm-install.js
```diff
@@ -1,11 +1,23 @@
...
+      ui.writeLine(chalk.yellow('Skipping npm install: package.json not found'));
```
stefanpenner: `ui.writeWarnLine`

stefanpenner:
@trek travis also appears to be failing, but green on master. Can you fix the issue?

stefanpenner:
Also can someone explain what the motivating example is for this?

trek:
@stefanpenner for https://github.com/ember-cli/ember-cli/issues/5163, use as a general purpose build tool.

lib/tasks/bower-install.js
```diff
@@ -12,13 +12,20 @@ module.exports = Task.extend({
...
     var chalk          = require('chalk');
     var bower          = this.bower;
+    var bowerJson      = path.join(this.project.root, 'bower.json');
     var bowerConfig    = this.bowerConfig;
     var ui             = this.ui;
     var packages       = options.packages || [];
     var installOptions = options.installOptions || { save: true };
 
+    if (!existsSync(bowerJson)) {
+      ui.writeWarnLine('Skipping bower install: bower.json not found');
```
stefanpenner: I believe this should just throw, as the command failed and we likely want the status code to reflect this.

lib/tasks/bower-install.js
```diff
@@ -12,13 +12,20 @@ module.exports = Task.extend({
...
+      return Promise.resolve();
```
stefanpenner: a naked return should be sufficient

lib/tasks/npm-install.js
```diff
@@ -1,11 +1,25 @@
...
-  completionMessage: 'Installed packages for tooling via npm.'
+  completionMessage: 'Installed packages for tooling via npm.',
+  run: function(){
+    var ui = this.ui;
+
+    var packageJson = path.join(this.project.root, 'package.json');
+    if (!existsSync(packageJson)) {
+
+      ui.writeWarnLine('Skipping npm install: package.json not found');
+      return Promise.resolve();
```
stefanpenner: naked return should be sufficient.

lib/tasks/bower-install.js
```diff
@@ -12,13 +12,20 @@ module.exports = Task.extend({
   },
   // Options: Boolean verbose
   run: function(options) {
+    var existsSync     = require('exists-sync');
     var chalk          = require('chalk');
     var bower          = this.bower;
+    var bowerJson      = path.join(this.project.root, 'bower.json');
```
stefanpenner: `this.project.root + '/bower.json'` is sufficient, path isn't needed as all node `fs` operations already normalize.

lib/tasks/bower-install.js
```diff
@@ -12,13 +12,20 @@ module.exports = Task.extend({
   },
   // Options: Boolean verbose
   run: function(options) {
+    var existsSync     = require('exists-sync');
```
stefanpenner: this can be loaded at the top of the module, as it is very fast to require and most likely already loaded.

`chalk` is lazy required, because (at-least at some point) it was extremely slow to require. If that is no-longer the case it can also be promoted to the top of the file.

stefanpenner:
I believe these tests should umit test the tasks. I domt think we need full e2e for this branching

stefanpenner:
attempting a rebase : https://github.com/ember-cli/ember-cli/pull/5395

stefanpenner:
superseded https://github.com/ember-cli/ember-cli/pull/5395

PR #5149, opened 2015-11-22
cherry-picking changelog from stable

kellyselden:


stefanpenner:
so, I dont think we should continue cherry-picking the change-logs. The 2 branches will diverge, there may end up being overlap but that will continue to change.

rwjblue:
@stefanpenner - When people come to look at what has changed they look at the repo, by default that means they land on the master branch. Having master include all changelogs for cut versions is pretty common (we do this for Ember as well). IMO, all stable versions should be included in the CHANGELOG.md on master.

stefanpenner:
@rwjblue non-linear changelogs do not make sense, at the point of divergence please put a link to the appropriate changelog.

stefanpenner:
in this case, the link would be: https://github.com/ember-cli/ember-cli/blob/stable-1-13/CHANGELOG.md

something like: 'forked off stable-1-13, to click [here](https://github.com/ember-cli/ember-cli/blob/stable-1-13/CHANGELOG.md) its change-log'

rwjblue:
From users perspective the changes are linear, and the majority of changes listed for 1.13.13 are also in master.

If we don't want to maintain the file in the repo, that seems fine (we can use https://github.com/ember-cli/ember-cli/releases/ links), but to have the file and it be wrong and incomplete makes no sense.

stefanpenner:
> and the majority of changes

^

PR #5150, opened 2015-11-22
[INTERNAL] Set correct version of ember-cli

jayeff:
I suspect that package.json version should be 1.13.13

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-cli",
-  "version": "1.13.8",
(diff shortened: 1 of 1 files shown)


jayeff:
I'm closing this. I believe I didn't look close enough. There is a stable-1-13 branch which probably I should have used.

rwjblue:
Master should at least be the same version. I'd prefer it to be 2.2.0-alpha, but this is better than just an old version ;)

PR #5154, opened 2015-11-24
reduce number of funnels created to support other asset app.imports

stefanpenner:
before:

screen shot 2015-11-23 at 5 42 26 pm

after:

screen shot 2015-11-23 at 5 41 58 pm

Not really a major cost, but could likely spiral out of control, pretty quickly

@rwjblue r?


rwjblue:
:+1:

stefanpenner:
@homu r+

PR #5159, opened 2015-11-25
Update appveyor.yml

stefanpenner:
We should have no more native dependencies


raytiley:
Hours of my debugging... just washed away like it never happened :)

stefanpenner:
@raytiley lest we forget those dark hours.

PR #5162, opened 2015-11-26
[WIP][ENHANCEMENT] Hot reload styles

gcollazo:
This PR adds support for "hot reloading" styles. That means that instead of doing a full page reload when you save a `css`, `sass`, `scss`, `less` or `styl` file the live reload server will just inject the compiled `project-name.css` into the browser.

This is my first PR to this repo. I would love feedback and suggestions on how to clean it up and getting it merge.

When I initially suggested this enhancement in the ember-cli channel on Slack @rwjblue referred me to the great [ember-cli-styles-reloader](https://github.com/xomaczar/ember-cli-styles-reloader) addon. After reading the source of the addon I ended up taking a lot it and including it in this PR. Thanks to @xomaczar.

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -5,6 +5,10 @@ var path        = require('path');
 var Task        = require('../../models/task');
 var SilentError = require('../../errors/silent');
 
+// default reload css extensions
+var styleExtensions = ['css', 'scss', 'sass', 'less', 'styl'];
+var reloadCssPattern = new RegExp('\.(' + styleExtensions.join('|') + ')$');
+
 function createServer() {
   var instance;
   var Server = (require('tiny-lr')).Server;
@@ -62,6 +66,15 @@ module.exports = Task.extend({
 
   didChange: function(results) {
     var filePath = path.relative(this.project.root, results.filePath || '');
+    var appStyleResource = this.project.pkg.name + '.css';
+
+    var reloadPaths = function() {
+      if (reloadCssPattern.test(filePath)) {
+        return [appStyleResource];
+      } else {
+        return ['LiveReload files'];
+      }
+    };
 
     var canTrigger = this.project.liveReloadFilterPatterns.reduce(function(bool, pattern) {
       bool = bool && !filePath.match(pattern);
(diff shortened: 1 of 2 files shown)


lib/tasks/server/livereload-server.js
```diff
@@ -5,6 +5,10 @@ var path        = require('path');
 var Task        = require('../../models/task');
 var SilentError = require('../../errors/silent');
 
+// default reload css extensions
+var styleExtensions = ['css', 'scss', 'sass', 'less', 'styl'];
```
rwjblue: We can get the known extensions for styles from the registry (something similar is done in the addon model with template extensions that you can probably tweak).

rwjblue:
This looks pretty good (not sure why you closed it).  I left one comment that needs to be addressed, then I can pull it down and play with it a bit locally.

gcollazo:
Thanks for the feedback @rwjblue I will make the suggested fixes and create a new PR. After opening this PR I tried to fix the merge conflicts with master and was about to force push to this branch and decided to close it first, fix the merge issues with then re-open.

Do you prefer a new PR or re-opening this one?

rwjblue:
> Do you prefer a new PR or re-opening this one?

Generally, I'd prefer to keep the same one (makes it easier to track suggestions/tweaks/etc), but if that is too difficult we can do a new one.

stefanpenner:
> I'd prefer to keep the same one (makes it easier to track suggestions/tweaks/etc)

:+1:

lib/tasks/server/livereload-server.js
gcollazo: I just checked the source for the most popular css precompilers and they are not consistent in terms of registering their file extensions.

https://github.com/aexmachina/ember-cli-sass/blob/master/index.js#L21
https://github.com/gdub22/ember-cli-less/blob/master/index.js#L12
https://github.com/drewcovi/ember-cli-stylus/blob/master/index.js#L6

I tried using the same technique used in the Addon model but couldn't get to the extensions. Also search through the registry instance and did not find the extensions or any other info I could use to extract the file extensions.

Any suggestions on how to proceed?

gcollazo:
TIL: if I delete the repo I will not be able to re-open the PR :disappointed:. Trying to OSS while suffering from food coma.

Issue #2371, opened 2014-10-23
Add ability to reload CSS without reloading the whole page


aexmachina:
I've submitted [a PR to broccoli-sane-watcher](krisselden/broccoli-sane-watcher#13) that will provide the set of changed files in the `change` event. We can inspect this and tell LiveReload to only reload the CSS if only CSS files have changed.

I'm happy to do this, but the question is how to handle CSS preprocessors. For example if I use SASS then I need to check for changes to `\.(scss|sass)$`, but if I'm using LESS then I need to check for `\.less$`. Obviously we can just provide a list of file extensions that covers the current CSS preprocessors, but it would be better to be able to get this information from the Registry.

I can get this information from the Registry (I'd need to capture it in the `StylePlugin` but that's fine), but what I can't see is how I can get ahold of the registry from the livereload-server.js.

Any suggestions?

pixelhandler:
This would be great :+1:

rwwagner90:
What's the status on this? I can try to help out, but not sure how much help I would be.

stefanpenner:
> What's the status on this? I can try to help out, but not sure how much help I would be.

we are re-working the internal build system to better provide selective builds. Once that lands we will likely dig in. If someone wants to make this work before, they are welcome to submit. If the PR is reasonable we will accept it.

aexmachina:
I have been using ember-cli-styles-reloader as an interim solution and it appears to work well.

oligriffiths:
@stefanpenner What's the progress on this?

rwjblue:
@oligriffiths - This is being replaced by #5165.

PR #5165, opened 2015-11-26
[ENHANCEMENT] Initial implementation of css hot realod

gcollazo:
[Link to original PR](https://github.com/ember-cli/ember-cli/pull/5162)

Replaces https://github.com/ember-cli/ember-cli/issues/2371.

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -5,6 +5,7 @@ var fs          = require('fs');
 var Task        = require('../../models/task');
 var SilentError = require('silent-error');
 var walkSync    = require('walk-sync');
+var path        = require('path');
 var FSTree      = require('fs-tree-diff');
 var debug       = require('debug')('ember-cli:live-reload:');
 
@@ -108,8 +109,17 @@ module.exports = Task.extend({
   },
 
   didChange: function(results) {
-    var previousTree = this.tree;
+    var canTrigger = true;
 
+    if (this.project.liveReloadFilterPatterns.length > 0) {
+      var filePath = path.relative(this.project.root, results.filePath || '');
+
+      canTrigger = this.project.liveReloadFilterPatterns.every(function(pattern) {
+        return pattern.test(filePath) === false;
+      });
+    }
+
+    var previousTree = this.tree;
     this.tree = new FSTree.fromEntries(this.getDirectoryEntries(results.directory));
 
     var files = previousTree.calculatePatch(this.tree)
@@ -119,16 +129,18 @@ module.exports = Task.extend({
 
     debug('files %a', files);
 
-    this.liveReloadServer().changed({
-      body: {
-        files: files
-      }
-    });
+    if (canTrigger) {
+      this.liveReloadServer().changed({
+        body: {
+          files: files
+        }
+      });
 
-    this.analytics.track({
-      name:    'broccoli watcher',
-      message: 'live-reload'
-    });
+      this.analytics.track({
+        name:    'broccoli watcher',
+        message: 'live-reload'
+      });
+    }
   },
 
   didRestart: function() {
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var Promise     = require('../../ext/promise');
-var path        = require('path');
 var fs          = require('fs');
 var Task        = require('../../models/task');
 var SilentError = require('silent-error');
@@ -18,6 +17,10 @@ function createServer(options) {
   };
   instance = new Server(options);
   return instance;
+}
+
+function isNotDirectory(entry) {
+  return !entry[2].isDirectory();
 }
 
 function relativePath(patch) {
@@ -100,12 +103,17 @@ module.exports = Task.extend({
                           '.  It is either in use or you do not have permission.');
   },
 
+  getDirectoryEntries: function(directory) {
+    return walkSync.entries(directory);
+  },
+
   didChange: function(results) {
     var previousTree = this.tree;
 
-    this.tree = new FSTree.fromEntries(walkSync.entries(results.directory));
+    this.tree = new FSTree.fromEntries(this.getDirectoryEntries(results.directory));
 
     var files = previousTree.calculatePatch(this.tree)
+      .filter(isNotDirectory)
       .map(relativePath)
       .filter(isNotSourceMapFile);
 
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -121,10 +121,10 @@ module.exports = Task.extend({
       result = this.project.liveReloadFilterPatterns.every(function(pattern) {
         return pattern.test(filePath) === false;
       });
-    }
 
-    if (result === false) {
-      this.writeSkipBanner(filePath);
+      if (result === false) {
+        this.writeSkipBanner(filePath);
+      }
     }
 
     return result;
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -104,21 +104,33 @@ module.exports = Task.extend({
                           '.  It is either in use or you do not have permission.');
   },
 
+  writeSkipBanner: function(filePath) {
+    this.ui.writeLine('Skipping livereload for: ' + filePath);
+  },
+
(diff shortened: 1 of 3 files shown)


lib/tasks/server/livereload-server.js
```diff
@@ -81,6 +85,15 @@ module.exports = Task.extend({
 
   didChange: function(results) {
     var filePath = path.relative(this.project.root, results.filePath || '');
+    var appStyleResource = this.project.pkg.name + '.css';
+
+    var reloadPaths = function() {
```
stefanpenner: i believe the function can be removed, and rather reloadPaths calculated here directly.

lib/tasks/server/livereload-server.js
```diff
@@ -81,6 +85,15 @@ module.exports = Task.extend({
...
+      if (reloadCssPattern.test(filePath)) {
+        return [appStyleResource];
```
stefanpenner: what if there are other css files that changes?

stefanpenner:
This seems to be taking the FS events as a cue to figure out what actually changed. That may work sometimes, but not always.

stefanpenner:
I have submitted a PR to your branch (which is this PR) with my recommendations https://github.com/gcollazo/ember-cli/pull/1

stefanpenner:
It appears that hot-reloading with css that has sourcemaps (like sass, the sourcemaps dont appear to reload correctly).... I tried both embedded and not embedded but without luck. Thats not a blocker for getting this in, but worth some investigation.

gcollazo:
@stefanpenner Just pushed a few changes. Most noticeable things are:

- Filter out *.map files to prevent a full browser reload when a precompiled style (scss, styl, less) file changes
- Filter out directories since live reload can't do anything useful with them. This also helps with tests.
- Removed all references to `Project.liveRealoadFilterPatterns` and it' related tests since I couldn't find any references to that on the project.

Let me know how can I improve it.

rwjblue:
liveRealoadFilterPatterns Is used by addons

stefanpenner:
@rwjblue for what?

rwjblue:
Suppressing live reload for certain files. 

Robert Jackson

-- twitter: rwjblue
-- github: rjackson

stefanpenner:
@rwjblue seems backwards the watchers filters should be used to prevent builds. Or is there a use-case where a built should happen, but a live reload shouldn't. That seems dubious.

I suspect anyone doing this, is likely doing so to prevent spurious live reloads, due to spurious builds, likely due to various temp files from various editors being written. These scenarios should actually be guarded at the watcher.

I am unaware of a reasonable use-case, where a build should be triggered, but a life reload should be prevented. If a use-case exists, please share.

rwjblue:
@stefanpenner - I am not really saying that it should be kept, just responding to the comment that "X isn't used internally so we should remove it".

The original use case for `Project.liveReloadFilterPatterns`'s was for the ember-cli-proxy-fixtures addon to use it as a filter to enable "VCR" like proxy fixture recording during a test suite run. On first run, the addon writes AJAX request fixtures to the `tests/fixtures` directory, which would cause a livereload (and restart the test suite).  The filter pattern was created/used to prevent that build from triggering a livereload.  This seems like a reasonable use case for wanting a build (so that subsequent refreshes include the fixtures properly) but prevent livereload.  However, if we want to remove the functionality that seems fine as long as it is intended for a major version and the breakage in API is documented when we release.

stefanpenner:
That sounds like a legit use-case, @gcollazo mind adding back that support? (I realized I removed it, so if you don't have time I can likely doit later)

gcollazo:
I'll do it.

stefanpenner:
@gcollazo be sure to use the `files` as the source of truth for what changed.

gcollazo:
@stefanpenner didn't read that last message before making these changes. What I did there should be backwards compatible. I'll give it another shot later tonight.

lib/tasks/server/livereload-server.js
```diff
@@ -76,21 +100,39 @@ module.exports = Task.extend({
...
+  getDirectoryEntries: function(directory) {
+    return walkSync.entries(directory);
   },
 
   didChange: function(results) {
-    var filePath = path.relative(this.project.root, results.filePath || '');
+    var canTrigger = true;
+
+    if (this.project.liveReloadFilterPatterns.length > 0) {
+      var filePath = path.relative(this.project.root, results.filePath || '');
```
stefanpenner: can we extract this block into a separate function, it will make it easer to read/test etc.

stefanpenner: I believe `canTrigger` may be flawed, as the filePath isn't going to always be right. It is very easy for many files to change, and get batched together...

gcollazo:
@stefanpenner @rwjblue 

If you look at [ember-cli-proxy-fixtures](https://github.com/dockyard/ember-cli-proxy-fixtures/blob/master/index.js#L61) they are filtering based on the path of the input file (tests/fixtures/proxy/file-a.js) which makes sense from the perspective of the user.

Right now `previousTree.calculatePatch(this.tree)` gives us a diff between trees but it seems there's only information about the destination of files (assets/tests.js) and not about their source.

In orden to really maintain backwards compatibility we need to know the original path of the files. The only way that I know of to achieve this is using `filePath` which only reports one file causing us to potentially miss files. Suggestions?

I searched all addons on npm and ember-cli-proxy-fixtures and ember-cli-styles-reloader are the only addons making reference to `liveReloadFilterPatterns`. Maybe we should just ask if the current implementation works for them and keep that. Thoughts?

stefanpenner:
> In orden to really maintain backwards compatibility we need to know the original path of the files. The only way that I know of to achieve this is using filePath which only reports one file causing us to potentially miss files. Suggestions?

seems like we just keep that old behavior for now, rethink again soon.

gcollazo:
@stefanpenner moved the skip reload logic to `shouldTriggerReload` and added a few more tests.

gcollazo:
@stefanpenner earlier today @rwjblue confirmed that the test failures are not related to this PR. I think this is ready for review whenever you have the time.

gcollazo:
*Adding this info here just to keep track of it:*

I did some testing based on [this tweet](https://twitter.com/addyosmani/status/671378820021264384) to try to determine why css source maps where not updating after live reload and it seems like an issue with chrome. I made some videos and demo apps with ember-cli and webpack.

Tested on Chrome:

- 47.0.2526.73 (64-bit): Has the problem
- 47.0.2526.73 beta (64-bit): Has the problem
- 49.0.2579.0 canary (64-bit): Works sometimes

Using ember-cli on Beta:
 - Video [https://youtu.be/P2_Te5LdXYE](https://youtu.be/P2_Te5LdXYE)
 - [Demo app source](https://github.com/gcollazo/ember-cli-livereload-source-maps-demo)

Using webpack on Beta:
- Video [https://youtu.be/X0fLP3tyNNQ](https://youtu.be/X0fLP3tyNNQ)
- [Demo app source](https://github.com/gcollazo/-webpack-livereload-source-maps-demo)

Using webpack on Canary:
- Video [https://youtu.be/8UmcadkpkFA](https://youtu.be/8UmcadkpkFA)

stefanpenner:
@gcollazo awesome, thanks for the investigation. I suspect, this should be reported [here](https://code.google.com/p/chromium/issues/list?can=2&q=Cr%3DPlatform-DevTools)

stefanpenner:
@gcollazo this looks really good, thanks for iterating with me on this one. I'm going to take at the (very likely to be unrelated) test failures shortly and try and get them green so we can merge

rwjblue:
@homu r+

rwjblue:
@gcollazo - Thanks for all your hard work on this!

gcollazo:
Thanks @rwjblue and @stefanpenner for the help.

cibernox:
This. Is. Big. :tada:

raido:
Awesome 🎉

matthooks:
Just wanted to leave props for doing this work! :+1:

lukemelia:
Great job!

gfvcastro:
Awesome :tada: :+1:

ynnoj:
:clap:

he9qi:
:+1: to everyone involved. Great job!!!

sandstrom:
Awesome! Thanks! :balloon:

bradcrawford:
👍

PR #5168, opened 2015-11-28
fix swallowed failures

stefanpenner:
* on failure, don’t break stdout.write ensure it is restored. cc @trentmwillis 
* fix failures it was hiding.
- [ ] additional failurea


stefanpenner:
@homu r+

PR #5174, opened 2015-12-01
Added jshint dependency

EmilTemirov:
Does not work "ember generate model" without this one


kellyselden:
appears to be duplicate of #5095. specifically the comment https://github.com/ember-cli/ember-cli/pull/5095#discussion_r44803341.

kellyselden:
Will reopen if it turns out to be a bigger issue.

PR #5185, opened 2015-12-04
Ensure rimraf is available to consumers.

### LINKED ISSUE #5173: Cannot find module 'rimraf'
I see a number of issues related to the same problem but this appears to be different.

I installed ember-cli for a project i am working on from the master branch on github as there was a fix i needed. I used 7cdcd310558580dd6b94ad5e0999c525294db678 as the sha for the dependency, changed the package.json dependency to
```json
"ember-cli": "https://github.com/ember-cli/ember-cli.git#7cdcd31",
```
After installing this i get `Cannot find module 'rimraf'` from `node_modules/ember-cli/lib/commands/new.js:8:26`

Looking at new.js i can see the require here https://github.com/ember-cli/ember-cli/blob/7cdcd310558580dd6b94ad5e0999c525294db678/lib/commands/new.js#L8

And in package.json https://github.com/ember-cli/ember-cli/blob/7cdcd310558580dd6b94ad5e0999c525294db678/package.json#L133
it is specified in the `devDependencies` section. Should this be in the `dependencies` section?

This issue was reproducible on CI as well as my local machine.

Running `ember -v` i get
```bash
ember -v
version: 1.13.13
node: 4.0.0
npm: 2.14.10
os: darwin x64
```

I have added `rimraf` to my dependencies and it has resolved the issue locally and on CI

rwjblue:
Fixes #5173


rwjblue:
@homu r+

stefanpenner:
It's really annoying that this is extremely annoying to test. Although I can imaging some glorious symlink hacks. It sounds like way to much work.

rwjblue:
@stefanpenner - Yes, I thought a bit about how we could test it.  The only think I could think of, would be to actually link an app from git or local reference (so that `npm install` in the app would only grab the deps and not the devDeps), but TBH I'm not sure that would even work. :crying_cat_face:

PR #5190, opened 2015-12-07
Only use `temp` dependency, not `tmp-sync`

akatov:
Don't use `tmp-sync` dependency which eventually depends on a deprecated `lodash` version.
Instead, use promise-wrapped functions from `temp`

### DIFF
--- a/lib/utilities/mk-tmp-dir-in.js
+++ b/lib/utilities/mk-tmp-dir-in.js
@@ -1,0 +1,15 @@
+'use strict';
+
+var Promise   = require('../ext/promise');
+var fs        = require('fs-extra');
+var temp      = require('temp');
+var mkdir     = Promise.denodeify(fs.mkdir);
+var mkdirTemp = Promise.denodeify(temp.mkdir);
+
+function mkTmpDirIn(dir) {
+  return mkdir(dir).then(function() {
+    return mkdirTemp({ dir: dir });
+  });
+}
+
+module.exports = mkTmpDirIn;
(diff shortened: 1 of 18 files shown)


lib/utilities/mk-tmp-dir-in.js
```diff
@@ -0,0 +1,25 @@
...
+
+var Promise   = require('../ext/promise');
+var fs        = require('fs-extra');
+var temp      = require('temp');
+var mkdir     = Promise.denodeify(fs.mkdir);
+var mkdirTemp = Promise.denodeify(temp.mkdir);
+
+function exists(dir) {
+  return new Promise(function(resolve) {
+    fs.exists(dir, resolve);
```
stefanpenner: Can't this also be denodeify?

lib/utilities/mk-tmp-dir-in.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+}
+
+function mkTmpDirIn(dir) {
+  return exists(dir).then(function(doesExist) {
+    if (!doesExist) {
+      mkdir(dir);
```
stefanpenner: Missing return

lib/utilities/mk-tmp-dir-in.js
akatov: `fs.exists` has a different signature - the callback receives only one argument - the boolean value.

stefanpenner:
@akatov good point

lib/utilities/mk-tmp-dir-in.js
akatov: hmm, we aren't using the return value of the promise, nor does the promise actually return a value..
however, if mkdir rejects we want the containing promise to reject also. So I'll update this.

akatov:
I was tempted to include normalize/mz but didn't want to add an extra dependency...

stefanpenner:
@homu r+

stefanpenner:
@homu retry

PR #5192, opened 2015-12-09
Use consistent double quotes in index.html blueprints

Dhaulagiri:
There seems to be a consensus around using double quotes in template/html files, so this PR makes ember-ci's blueprints use double quotes consistently in its `content-for` hooks to match the rest of the file.  

If the use of single quotes was intentional for some reason you can feel free to disregard.


kellyselden:
:+1:

rwjblue:
@homu r+

PR #5197, opened 2015-12-11
update leek

akatov:
the latest version of leek does not depend on deprecated lodash any more

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "inquirer": "0.5.1",
     "is-git-url": "^0.2.0",
     "isbinaryfile": "^2.0.3",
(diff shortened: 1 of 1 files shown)


stefanpenner:
@homu +r

akatov:
seems like a problem with leek's bundledDependencies. Let's see whether we can fix this with @twokul

stefanpenner:
@akatov i think we should removed bundledDependencies from leek.

akatov:
@stefanpenner just forced push branch with a version of leek without bundledDependencies.

stefanpenner:
thanks

stefanpenner:
@homu r+

PR #5205, opened 2015-12-13
Add addon-test-support tree.

rwjblue:
This commit adds a new tree type to be used during the build processing.  It adds a way for addons to provide testing helpers (in `<addon-name>/test-support` to the consuming app) in the `test-support` file (essentially "vendor" for "tests").

---

- [x] Sanity check
- [x] Add tests
- [x] Perf tests before / after

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1241,23 +1241,32 @@ EmberApp.prototype.testFiles = function(coreTestTree) {
   testSupportPath = testSupportPath.testSupport || testSupportPath;
 
   var external = this._processedExternalTree();
-
   var emberCLITree = this._processedEmberCLITree();
 
-  var testJs = this.concatFiles(mergeTrees([external, coreTestTree]), {
-    inputFiles: this.legacyTestFilesToAppend,
-    outputFile: testSupportPath,
-    annotation: 'Concat: Test Support JS'
+  var addonTestSupportTree = mergeTrees(this.addonTreesFor('addon-test-support'), {
+    overwrite: true,
+    annotation: 'TreeMerger (addon-test-support)'
+  });
+  var transpiledAddonTestSupportTree = new Babel(addonTestSupportTree, this._prunedBabelOptions());
+
+  var finalAddonTestSupportTree = new Funnel(transpiledAddonTestSupportTree, {
+    allowEmpty: true,
+    destDir: 'addon-test-support',
+    annotation: 'Funnel: Addon Test Support'
   });
 
-  testJs = this.concatFiles(mergeTrees([testJs, emberCLITree]), {
-    inputFiles: [
-      'vendor/ember-cli/test-support-prefix.js',
-      testSupportPath.slice(1),
-      'vendor/ember-cli/test-support-suffix.js'
-    ],
+  var inputFiles = [].concat(
+    this.legacyTestFilesToAppend,
+    'addon-test-support/**/*.js'
+  );
+
+  var baseMergedTree = mergeTrees([emberCLITree, external, coreTestTree, finalAddonTestSupportTree]);
+  var testJs = this.concatFiles(baseMergedTree, {
+    headerFiles: [ 'vendor/ember-cli/test-support-prefix.js' ],
+    inputFiles: inputFiles,
+    footerFiles: [ 'vendor/ember-cli/test-support-suffix.js' ],
     outputFile: testSupportPath,
-    annotation: 'Concat: Test Support Suffix'
+    annotation: 'Concat: Test Support JS'
   });
 
   var testemPath = path.join(__dirname, 'testem');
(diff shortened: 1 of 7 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1241,23 +1241,32 @@ EmberApp.prototype.testFiles = function(coreTestTree) {
...
-      'vendor/ember-cli/test-support-suffix.js'
-    ],
+  var finalAddonTestSupportTree = new Funnel(transpiledAddonTestSupportTree, {
+    allowEmpty: true,
+    destDir: 'addon-test-support',
+    annotation: 'Funnel: Addon Test Support'
+  });
+
+  var inputFiles = [].concat(
+    this.legacyFilesToAppend,
```
stefanpenner: i think this wants to be: `legacyTestFilesToAppend`

rwjblue: WHOOPS! yep, good catch

stefanpenner:
@rwjblue if you can, it would be great to just double check this doesn't regress the performance in a real app. Nothing is jumping out, but I would like to have PR's author that change the build itself to do some manual tests comparing before/after build times.

https://github.com/ember-cli/stress-app may be a good candidate, but your own apps are also good.

rwjblue:
Currently, it should actually result in a bit faster build when no addons are present that utilize the new tree (because I replaced two back to back concats with one smarter one).

But I definitely agree with you!  I added perf tests to the checklist above.

stefanpenner:
> Currently, it should actually result in a bit faster build when no addons are present that utilize the new tree (because I replaced two back to back concats with one smarter one).

Ya i noticed, that is nice. I also suspect it will be better. (It would be great to see confirmation of this)

rwjblue:
> It would be great to see confirmation of this

Ya, definitely agreed.  I spiked it out to support the changes in https://github.com/ember-cli/ember-cli-qunit/pull/92, and wanted a quick sanity check before continuing down the rabbit hole :rabbit: :rabbit2:

stefanpenner:
Ya it seems good

rwjblue:
Times using ember-cli/stress-app:

| type | time before | time after |
--------|-------|---------|
|  Initial Cold Build | 164441ms | 173884ms |
| Initial Warm Build | 12040ms | 10264ms
| Rebuild (JS file change) | 4879ms | 5057ms |
| Rebuild (SCSS file change) | 5713ms | 4510ms |

stefanpenner:
> Rebuild (JS file change)	4879ms	5057ms

the baseline numbers don't seem correct (they should be around 2s, unless you are on battery or something), can you include the slow tree?

rwjblue:
> can you include the slow tree?

Yep, I kept them in [this document](https://usecanvas.com/rwjblue/addon-test-support-timings/3NAMy5kG2T5zibZ5APpL71).

stefanpenner:
@rwjblue those merge trees look out of place, what is the result of `npm ls broccoli-merge-trees` ?

rwjblue:
> what is the result of npm ls broccoli-merge-trees ?

https://gist.github.com/rwjblue/1adb26dae4f53451968a

stefanpenner:
@rwjblue ah ya, that explains it you 1.0.0 should be 1.1.0 for ember-cli (ideally the rest aswell) it seems like you npm install trolled you. As ember-cli actually depends o ^1.0.0 which should have drifted to https://github.com/ember-cli/ember-cli/blob/master/package.json#L51

rwjblue:
Ugh.  OK, will `rm -rf node_modules && npm install` in all the repos and try again :sob:

stefanpenner:
ember-intl can be bumped to ^2.0.0 (which should promote all the old merge-trees  i see aswell)

rwjblue:
> ember-intl can be bumped to ^2.0.0 

Ya, I spotted that also and PR'ed in https://github.com/ember-cli/stress-app/pull/12.

stefanpenner:
last remaining one will be addressed by: https://github.com/jasonmit/ember-intl-messageformat-parser/pull/1

stefanpenner:
oh and by: https://github.com/aexmachina/ember-cli-sass/commit/73bdf7963335252b2e7e788ebfd7d75a92604060

stefanpenner:
with all that, i see:

```
file changed app.js

Build successful - 2081ms.

Slowest Trees                                 | Total
----------------------------------------------+---------------------
SassCompiler                                  | 1403ms
ConcatWithMaps: Concat: App                   | 189ms

Slowest Trees (cumulative)                    | Total (avg)
----------------------------------------------+---------------------
SassCompiler (1)                              | 1403ms
ConcatWithMaps: Concat: App (1)               | 189ms
Babel (17)                                    | 113ms (6 ms)
```

rwjblue:
OK, updated times (sorry for the confusion earlier):

| type | time before | time after |
--------|-------|---------|
| Initial Warm Build | 7951ms | 8894ms
| Rebuild (JS file change) | 3564ms | 2494ms |
| Rebuild (SCSS file change) | 3061ms | 2479ms |

Looks like rebuild speeds improved a tiny bit...

rwjblue:
@homu r=stefanpenner

stefanpenner:
@homu r+

stefanpenner:
tested without sass (satting the masssive 86,000 files in bower) and the rebuild times are like 600ms or so)  so this is good 2 go

PR #5207, opened 2015-12-13
Make ember test helper clearer.

rwjblue:
Took me a little bit to figure out that this test helper uses `lib/cli/index.js` (and not `lib/cli/cli.js`). I removed the `new` since `lib/cli/index.js` returns a normal function (and not a constructor).

Return value object is found here: https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L103-L106

### DIFF
--- a/tests/helpers/ember.js
+++ b/tests/helpers/ember.js
@@ -2,14 +2,47 @@
 
 var MockUI        = require('./mock-ui');
 var MockAnalytics = require('./mock-analytics');
-var Cli           = require('../../lib/cli');
+var cli           = require('../../lib/cli');
 
+/*
+  Accepts a single array argument, that contains the
+  `process.argv` style arguments.
+
+  Example:
+
+  ```
+  ember test --no-build --test-port=4567
+  ```
+
+  In this example, `process.argv` would be something similar to:
+
+  ```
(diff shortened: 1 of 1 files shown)


tests/helpers/ember.js
```diff
@@ -2,14 +2,38 @@
 
 var MockUI        = require('./mock-ui');
 var MockAnalytics = require('./mock-analytics');
-var Cli           = require('../../lib/cli');
+var cli           = require('../../lib/cli');
 
+/*
+  Accepts a single array argument, that contains the same
+  type of array as is available from a command line.
```
stefanpenner: might be worth mentioning that it is basically part of `process.argv`

process.argv is basically `[pathToNode, pathToFile, ...argsEmber-CliCares about]`

So

```
ember new foo
```

becomes `process.argv === ["path/to/node", "path/to/bin/ember", "new", "foo"]`

rwjblue: Updated to talk a bit about `process.argv`.

stefanpenner:
LGTM

rwjblue:
@homu r+

rwjblue:
@homu r+

PR #5213, opened 2015-12-13
adding safe restore test helper, because safer and cuts on code dupli…

kellyselden:
…cation


kellyselden:
tests should be good.

kellyselden:
@homu r+

PR #5230, opened 2015-12-16
Helper Unit test should pass a params array as first argument

taras:
Helper function accepts params as first argument, but helper unit tests currently pass `42`. It should pass `[42]` instead. 

Changed to pass `[42]`.


rwjblue:
@homu r+

Issue #5240, opened 2015-12-21
Remove `app-prefix` and `app-suffix`?


nathanhammond:
`app-prefix` and `app-suffix` are a misfeature for a future modules-only version of Ember. Fortunately from what I can tell "nobody" is using these hooks:

https://github.com/search?l=JavaScript&q=app-prefix&type=Code
https://github.com/search?l=JavaScript&q=app-suffix&type=Code

(I've reviewed every single result page.)

I'd like to propose that we deprecate these prior to releasing 2.2 so that we don't have to support it until 3.0. If this is something we wish to do I'll pull together a PR today.

stefanpenner:
I would also like this.

@rwjblue would love your thoughts and maybe an enumeration of usages we may be missing.

nathanhammond:
I found one use case: https://github.com/jkarsrud/ember-cli-defeatureify/blob/master/index.js#L20

Neither of `app-prefix` or `app-suffix` is documented anywhere, so it isn't surprising that there is approximately no usage. I'd call them internal APIs.

I don't have a simple backwards-compatible way to accomplish ember-cli-defeatureify without `app-prefix` since any existing code will rely on globals set up in `app-prefix`. To fix this the ember-cli-defeatureify addon will need to generate a module which can be `require`d in anywhere that needs defeatureifying.

The migration path would be mechanical, adding `import 'ember-cli-defeatureify/config' as myNamespace;` to any module which uses it.

Still haven't found any usage of `app-suffix`. I'll submit a PR to ember-cli-defeatureify/broccoli-defeatureify/defeatureify as well if it means we can get this to land. :)

rwjblue:
I'm :+1: on deprecating both in a 1.13 point release and in master, but I'm not convinced we should remove on master immediately after deprecating.

I want to avoid the self inflicted issues that we caused in Ember itself from 1.13 to 2.0, where we continued to backport deprecations for quite a while in the 2.0.0 cycle.  I would prefer to deprecate in both stable and master and target removal for a future version (prior to 3.0 though).  This is roughly similar to the Ember LTS concept (in my head at least :smile_cat:).

---

In the end, I believe deprecating has roughly the same effect as removal (deprecation noise is extremely annoying, and is mostly in the control of the app developer to remove) without the immediate breaking change before the deprecation can sink in and be dealt with in the ecosystem.

(moved comment from #5245)

stefanpenner:
[...]

PR #5179, opened 2015-12-03
Updating deprecations that use the deprecate utility

gmurphey:
Updating deprecations that use the deprecate utility to use the `deprecateUI` method instead. Fixes #5176.

This will make deprecation messages obey `--silent` flags in commands.

PR #5245, opened 2015-12-22
Deprecate app-prefix and app-suffix - 1.X

nathanhammond:
Adds deprecation notices into 1.13 stable. Related to #5240.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -38,6 +38,8 @@ var merge         = require('lodash/object/merge');
 var omit          = require('lodash/object/omit');
 var Funnel        = require('broccoli-funnel');
 
+var deprecate     = require('../utilities/deprecate');
+
 module.exports = EmberApp;
 
 /**
@@ -1438,6 +1440,8 @@ EmberApp.prototype.toTree = function(additionalTrees) {
  */
 EmberApp.prototype.contentFor = function(config, match, type) {
   var content = [];
+  var deprecatedHooks = ['app-prefix', 'app-suffix'];
+  var deprecateUI = deprecate.deprecateUI(this.project.ui);
 
   switch (type) {
     case 'head':             this._contentForHead(content, config);           break;
@@ -1449,6 +1453,7 @@ EmberApp.prototype.contentFor = function(config, match, type) {
   content = this.project.addons.reduce(function(content, addon) {
     var addonContent = addon.contentFor ? addon.contentFor(type, config, content) : null;
     if (addonContent) {
+      deprecateUI('The `' + type + '` hook used in ' + addon.name + ' is deprecated. The addon should generate a module and have consumers `require` it.', ~deprecatedHooks.indexOf(type));
       return content.concat(addonContent);
     }
 


stefanpenner:
test failure is unrelated, basically some version of `npm` just can't install stuff... Feel free to move that version to allowed failures so that we can move this forward.

nathanhammond:
Since this is somewhat breaking, I feel like I'm obligated to bring this to resolution before this can land: https://github.com/jkarsrud/ember-cli-defeatureify/issues/15

Will keep this thread updated.

lib/broccoli/ember-app.js
```diff
@@ -38,6 +38,8 @@ var merge         = require('lodash/object/merge');
 var omit          = require('lodash/object/omit');
 var Funnel        = require('broccoli-funnel');
 
+var deprecate     = require('../utilities/deprecate');
```
rwjblue: I'd prefer to use `deprecateUI` so all writes go through the project's `ui` object (to honor things like `--silent` flag and whatnot).

https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/deprecate.js#L11

rwjblue:
I'm :+1: on deprecating both in a 1.13 point release and in master, but I'm not convinced we should remove on master immediately after deprecating.

I want to avoid the self inflicted issues that we caused in Ember itself from 1.13 to 2.0, where we continued to backport deprecations for quite a while in the 2.0.0 cycle.  I would prefer to deprecate in both stable and master and target removal for a future version (prior to 3.0 though).  This is roughly similar to the Ember LTS concept (in my head at least :smile_cat:).

---

In the end, I believe deprecating has roughly the same effect as removal (deprecation noise is extremely annoying, and is mostly in the control of the app developer to remove) without the immediate breaking change before the deprecation can sink in and be dealt with in the ecosystem.

lib/broccoli/ember-app.js
gmurphey: `deprecateUI` itself is deprecated. :stuck_out_tongue_winking_eye: This should probably go through `ui.writeDeprecateLine`.

https://github.com/ember-cli/ember-cli/blob/master/lib/ui/index.js#L129

rwjblue: hehe, thank you for pointing that out @gmurphey. We definitely want to have things go through the `ui` object if at all possible...

gmurphey: Oh, actually... those `ui` helpers don't exist on `stable-1-13`. This will have to go through `deprecateUI`, or those helpers will have to be pulled in, too. (#5179)

nathanhammond: @rwjblue So far as I can tell there are currently no deprecations inside of lib/broccoli, and the `ui` object from the task isn't available.

Would you like me to pass it through here? https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L55

nathanhammond: NINJA UPDATE: Disregard my above lies. Didn't know it existed on `this.project`.

nathanhammond:
I've updated this to use `deprecateUI`. I'll use the `ui` version in the master branch.

By my estimation this is ready to merge.

lib/broccoli/ember-app.js
rwjblue: ninja update

nathanhammond:
Failures appear unrelated. Should be good to kick over and rerun.

rwjblue:
Failure is due to Node 5 having a broken NPM version. On master Node 5 is in allowed failures due to this.

stefanpenner:
ember-cli:stable-1-13 now also has this set to allowed failures, future PR will be ok.

rwjblue:
Thank you sir!

Issue #5242, opened 2015-12-21
Refactor `app-boot`?


nathanhammond:
I'd like to propose rethinking the `app-boot` hook. Treating it as an appended section of imperative code inside of `app.js` makes it difficult to reason about in a modules-only world. We only have a few occurrences of this in our extended codebases (I audited every occurrence on GitHub), which imply that we can reduce the API surface area and accomplish the same underlying goal of triggering the initial `require` that unravels everything.

## Use cases:
* [ember-cli](https://github.com/ember-cli/ember-cli/blob/a2a9a886e146ce7bdfbdd5338542538ea7b38d14/lib/broccoli/ember-app.js#L1600) - Generates the initial require.
* [ember-cli-delay-app-boot](https://github.com/rwjblue/ember-cli-delay-app-boot/blob/master/index.js#L31) - Throws a timeout around the initial require and includes some copypasta from above.
* [ember-cli-fastboot](https://github.com/tildeio/ember-cli-fastboot/blob/2058dd1ebf08be818851d1e65651db49cdd0e69f/index.js#L73) - This adds a new specially-named module and doesn't call `require`. Instead, it itself becomes the module which must be required to boot the app.
- In the future (read: I'm already encountering this) a Service Worker may wish to tell the application when it should boot or bind that functionality onto a promise chain. My current approach is to delay responding to the app-boot request and maintain that promise chain inside of the service worker.

## Analysis & Proposal
It seems to me that the primary use case for `app-boot` is to tweak and modify how the initial `require` is called. None of these above use cases are dramatically different. I'd propose that we adopt the behavior of `fastboot` generically, creating a magic module that you can `require` and use to unravel the application. It covers all existing use cases, reduces copypasta, can move us toward reduced API surface area, and makes the `app-boot` hook much simpler: `require('ember-app')()`.

rwjblue:
In general, we should use ember-cli/rfcs for this kind of proposal in the future.

---

Using a single module that can be required sounds nice generally speaking, but how does that work when you have two apps running at once?  Wouldn't they would both try to create the `ember-app` module and one would be unbootable?

nathanhammond:
I didn't think of RFCs for this because I was aiming for a nearly transparent change. Will do next time. :)

***

[...]

PR #5246, opened 2015-12-22
Split config module from app-boot.

nathanhammond:
`EmberApp` creates a config module which can be included at any point in the output code and is not order dependent. However, since `app-boot` *is* order dependent, and included as part of the same config file it forces the entire chunk of code to be order dependent.

This decouples the two, reducing the amount of specific ordering required in the code. No functional changes.

Related to #5242.


stefanpenner:
:+1:

stefanpenner:
test failure seems unrelated

rwjblue:
@homu r+

PR #5247, opened 2015-12-22
Unfortunately remove CSP add-on

stefanpenner:
* UX in browsers is so poor, this confuses most
* it is a painful barrier for new users
* it can still be pulled in for those security savvy
* Pain has been brought up with vendors, but no movement yet

—

I would love to restore this once the UX is reasonable,
right now it does more harm then good.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -24,7 +24,6 @@
     "ember-cli": "<%= emberCLIVersion %>",
     "ember-cli-app-version": "^1.0.0",
     "ember-cli-babel": "^5.1.5",
-    "ember-cli-content-security-policy": "0.4.0",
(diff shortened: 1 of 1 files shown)


stefanpenner:
@homu +r

stefanpenner:
@homu r+

kellyselden:
I think this warrants a special step in the changelog upgrade steps. "This plugin is now optional. If you liked it, keep it in your package.json". Some people might zealously remove it because our official blueprint removed it, when they may want to keep it. Thoughts?

rwjblue:
@kellyselden - Yeah, I agree. I'll try to come up with some good wording, will ping you on release note PR so you can review...

kellyselden:
@rwjblue Many thanks. I will add an issue to track.

PR #5252, opened 2015-12-22
Deprecate app-prefix and app-suffix - master

nathanhammond:
Related to #5240 and #5245. Deprecates `app-prefix` and `app-suffix` in the master branch.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1534,6 +1534,8 @@ EmberApp.prototype.toTree = function(additionalTrees) {
  */
 EmberApp.prototype.contentFor = function(config, match, type) {
   var content = [];
+  var deprecatedHooks = ['app-prefix', 'app-suffix'];
+  var deprecate = this.project.ui.writeDeprecateLine.bind(this.project.ui);
 
   switch (type) {
     case 'head':             this._contentForHead(content, config);           break;
@@ -1545,6 +1547,7 @@ EmberApp.prototype.contentFor = function(config, match, type) {
   content = this.project.addons.reduce(function(content, addon) {
     var addonContent = addon.contentFor ? addon.contentFor(type, config, content) : null;
     if (addonContent) {
+      deprecate('The `' + type + '` hook used in ' + addon.name + ' is deprecated. The addon should generate a module and have consumers `require` it.', !~deprecatedHooks.indexOf(type));
       return content.concat(addonContent);
     }
 


lib/broccoli/ember-app.js
```diff
@@ -1534,6 +1534,8 @@ EmberApp.prototype.toTree = function(additionalTrees) {
  */
 EmberApp.prototype.contentFor = function(config, match, type) {
   var content = [];
+  var deprecatedHooks = ['app-prefix', 'app-suffix'];
+  var deprecate = this.project.ui.writeDeprecateLine.bind(this.project.ui);
```
nathanhammond: @gmurphey @rwjblue Fun story, the API for `writeDeprecateLine` is reversed from the API for `deprecate`/`deprecateUI`. Threw me for a loop.

gmurphey: That could probably be better documented. I built it so it to mirror the `Ember.warn` (a falsy test causes the message to be logged).

stefanpenner: #programming

nathanhammond:
Failures appear unrelated. Should be good to kick over.

chnn:
I just saw this deprecation notice after upgrading Ember CLI. I was using `app-prefix` to add a copyright notice to the built app code. The addon looks something like:

```
var fs = require('fs');
var copyright = fs.readFileSync('public/copyright.txt', 'utf8');

module.exports = {
  name: 'prepend-copyright',

  contentFor: function(type, config, content) {
    if (type === 'app-prefix') {
      return "/* @copyright\n" + copyright + "*/";
    }
  }
};
```

Is there an alternate way to achieve this?

rwjblue:
Great question. I think that is something that we should definitely support (and can't think of something better than app-prefix today), but I'm unsure what the best go forward API would be. 

Can you open an issue in ember-cli/rfcs so we can iterate on an API?

chnn:
Will do! Thanks.

nathanhammond:
@chnn This was deprecated to prevent arbitrary code insertion locations, so your goal doesn't conceptually conflict with this PR. We at least don't have a more fundamental disagreement. :stuck_out_tongue: I wanted this removed to help us move toward a modules-only world, one in which app.js can be split into hundreds of pieces for loading and caching. I'm not sure how much sense a global copyright makes with "weird" concatenation patterns in some future state.

Separately from the goal of prefixing with non-code-content, it seems a bit legally grey to add a global copyright notice to `app.js`. `app.js` can have numerous files (from addons) and portions of code (inserted via Babel) that aren't actually your own. To make copyright insertion legitimate you would need to make it aware of all of the portions of code you actually wrote and that gets very legally hairy.

I don't know the proper way to handle something like this, but I suspect a complete solution requires a _lot_ more work than just inserting some text notice. Or, maybe all of that is handled inside of your `copyright.txt` and wordsmithed such that you disclaim all of the things which aren't yours. I dunno.

chnn:
> I wanted this removed to help us move toward a modules-only world, one in which app.js can be split into hundreds of pieces for loading and caching.

Understood, I look forward to such a world :)

I discussed the legal issues with someone on my team—those are great points you bring up. We agree that such comments are problematic, and may alter or remove them altogether once the situation is exacerbated by different module loading strategies or the `app-prefix` API being removed. If we can think of a clear use case and solution in the future, I'll submit an RFC.

stefanpenner:
license comments should be inserted during the concat phase, as the first headerFile. This will ensure source-maps work correctly.

rwjblue:
@stefanpenner - Yes, agreed. My point above was that this is not exposed/possible to users IIRC.

PR #5258, opened 2015-12-23
[WIP] Updating ember-addon to lint addon entry file.

### LINKED ISSUE #5253: [Addons][Bug] Addon entry file (index.js) is not run through JSHint
Steps to reproduce:

1. Create an addon
2. Add `var path = require('path');` to addon entry point
3. Run `ember test`
4. Witness tests passing

I would think this would be caught by the `unused` flag in `.jshintrc`, but I don't see `index.js` being run through JSHint.

gmurphey:
This is my first go with anything Broccoli related in ember-cli (yay!), so I'm hoping to get some feedback on this before I finalize it for review.

### DIFF
--- a/lib/broccoli/ember-addon.js
+++ b/lib/broccoli/ember-addon.js
@@ -5,6 +5,7 @@
 @module ember-cli
 */
 var defaults = require('merge-defaults');
+var Babel  = require('broccoli-babel-transpiler');
 var Funnel   = require('broccoli-funnel');
 var EmberApp = require('./ember-app');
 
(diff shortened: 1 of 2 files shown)


rwjblue:
I do not believe that we can create a watched broccoli tree for a file in the project root (because it will trigger an endless loop of changes when ran with ember serve or ember test --serve)). It may be possible to use an unmatched tree though.

stefanpenner:
Unless im misreading, this index file is part of node side of an add-on, and should be validated by node side tests, rather then client side tests.

gmurphey:
Both good points, and things I hadn't really thought about. I'm going to close this PR and do some more digging into this issue.

PR #5262, opened 2015-12-24
consolidating command/blueprint print logic

kellyselden:
Part of my quest to clean up code.
Command and blueprint printing were remarkably similar, so I extracted into a print command utility. There were only minor inconsistencies that resulted in "breaking" changes. By breaking I mean the help text changed a little bit, for instance whether "(Required)" came before or after "(Default: value)".

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1224,6 +1224,8 @@ Blueprint.prototype.insertIntoFile = function(pathRelativeToProjectRoot, content
   }
 };
 
+Blueprint.prototype._printCommand = printCommand;
+
 Blueprint.prototype.printBasicHelp = function(verbose) {
   var initialMargin = '      ';
   var output = initialMargin;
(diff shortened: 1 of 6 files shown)


lib/models/blueprint.js
```diff
@@ -1231,76 +1232,19 @@ Blueprint.prototype.printBasicHelp = function(verbose) {
   } else {
     output += this.name;
 
-    var options = this.anonymousOptions;
+    output += printCommand.call(this, initialMargin, true);
```
stefanpenner: lets install the shared command on the respective prototypes, so they can be invoked off `this` rather then via call.

kellyselden: Can do. Is there a performance reason I don't know about, or just style issue?

stefanpenner: both (although the performance in this case doesn't matter)

kellyselden:
@stefanpenner comment addressed. Installed it as a private member function. Also changed the test stubbing as a result.

stefanpenner:
:+1:

kellyselden:
@homu r+

PR #5263, opened 2015-12-24
cleanup test timeouts

stefanpenner:
* share similar timeout for a given describe
* bump some timeouts we keep exceeding.


kellyselden:
:+1:

kellyselden:
@homu r+

PR #5269, opened 2015-12-26
Printable properties tests

kellyselden:
unit test cleanup.
Extracting out the login regarding what help properties to print in json, to it's own section.


kellyselden:
@homu r+

PR #5274, opened 2015-12-26
Fix the order of  test files to append

mkempe:
Hey there,

first of all, many thanks for Ember CLI, happy holidays and a great new year!

Second, this is only a draft for a fix and probably not the final / right solution, @stefanpenner certainly knows a better solution.

The problem: I want to use ember-cli-mocha, but it is currently not working with Ember CLI 2.2, see [ember-cli-mocha #93](https://github.com/switchfly/ember-cli-mocha/issues/93) or [ember-cli-mocha #94](https://github.com/switchfly/ember-cli-mocha/issues/94).

The problem is `ember-mocha` and `ember-mocha-adapter` are included in the compiled `test-support.js` file before Mocha. Ember CLI 2.2 uses a new version of `broccoli-concat` where all files in `inputFiles` gets un-ordered concated and so you have no control over the order.

This fix moves all `legacyTestFilesToAppend` into `headerFiles`, to preserve the order. Additional the `prepend` option can now be used for test files.

Could someone please look over and give me feedback.

Thanks

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1403,7 +1403,7 @@ EmberApp.prototype._import = function(assetPath, options, directory, subdirector
   var basename = path.basename(assetPath);
 
   if (isType(assetPath, 'js', {registry: this.registry})) {
-    if(options.type === 'vendor') {
+    if (options.type === 'vendor') {
       if (options.prepend) {
         this.legacyFilesToAppend.unshift(assetPath);
       } else {
@@ -1419,7 +1419,7 @@ EmberApp.prototype._import = function(assetPath, options, directory, subdirector
       throw new Error('You must pass either `vendor` or `test` for options.type in your call to `app.import` for file: ' + basename);
     }
   } else if (extension === '.css') {
-    if(options.type === 'vendor') {
+    if (options.type === 'vendor') {
       this.vendorStaticStyles.push(assetPath);
     } else {
       this.vendorTestStaticStyles.push(assetPath);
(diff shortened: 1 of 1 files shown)


devinus:
@mkempe Wow, nice work tracking this down.

devinus:
@mkempe What happens to `addon-test-support` (https://github.com/ember-cli/ember-cli/pull/5274/files#diff-392bb207fea8bcbeb4a084bb445f7cc7L1262)?

devinus:
@mkempe Shouldn't `addon-test-support` technically be in the headerFiles too?

devinus:
I can't imagine a scenario when a user would be cool with it loading randomly with their test files...

mkempe:
@devinus Thanks for looking over. I agree with you, but i also have to admit that this is my first commit to Ember CLI and i probably don't see the full picture, respectively I suppose there are reasons why this has been implemented in that way.

devinus:
I'm almost positive `addon-test-support/**/*.js` should go right before `legacyTestFilesToAppend` in the `headerFiles`.

mkempe:
Currently, `addon-test-support` can't be in `headerFiles`, because `headerFiles` and `FooterFiles` cannot contain a glob, see https://github.com/ember-cli/broccoli-concat/blob/master/concat.js#L41.

devinus:
Oh wow, well this will get interesting then.

rwjblue:
:+1: - This looks good to me, thank you!

@homu r+

rwjblue:
@devinus - `addon-test-support` is intended (like `test-support`, `addon`, and `app` in an addon) to contain only modules and therefore order should not matter (because the loader will determine order based on imports and whatnot).  The change here ensures that the truly order dependent things (the legacy test files) are properly ordered.

devinus:
Nice!

mkempe:
😃

@rwjblue Thanks for looking over and merging in.

Issue #5277, opened 2015-12-27
Cannot find module 'broccoli-sourcemap-concat'


ballPointPenguin:
after upgrading from ember-cli 1.13.x to 2.2.0-beta.2
```
version: 2.2.0-beta.2-v2.2.0-beta.2-c78d55f84b
Cannot find module 'broccoli-sourcemap-concat'
```

rwjblue:
What is the rest of the stack trace?

ballPointPenguin:
```
Cannot find module 'broccoli--concat'
Error: Cannot find module 'broccoli--concat'
    at Function.Module._resolveFilename (module.js:327:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:355:17)
    at require (internal/module.js:13:17)
    at Class.Addon.concatFiles (/home/brosas/forks/ember-cli/lib/models/addon.js:184:10)
    at Class.module.exports.postprocessTree (/home/brosas/codex/fire-drake/node_modules/ember-cli-blanket/index.js:82:28)
    at /home/brosas/forks/ember-cli/lib/broccoli/ember-app.js:493:27
    at Array.forEach (native)
    at EmberApp.addonPostprocessTree (/home/brosas/forks/ember-cli/lib/broccoli/ember-app.js:491:23)
    at EmberApp.toTree (/home/brosas/forks/ember-cli/lib/broccoli/ember-app.js:1508:15)
```

ballPointPenguin:
I think I have a working PR.

rwjblue:
Looks like https://github.com/ember-cli/ember-cli/pull/5221 forgot to fix https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L173-L185. My guess is that some addon you are using is utilizing `this.concatFiles`.

A short term work around would be `npm install --save-dev broccoli-sourcemap-concat`.

rwjblue:
@ballPointPenguin - PR sounds great.  Ultimately, updating those references from `broccoli-sourcemap-concat` to `broccoli-concat` should fix things for you, but we also need to ensure that this function gets under test...

stefanpenner:
This.concatFiles should be deprecated and removed

ballPointPenguin:
closing in favor of #5280

PR #5278, opened 2015-12-27
replace broccoli-sourcemap-concat usage with broccoli-concat

ballPointPenguin:
closes https://github.com/ember-cli/ember-cli/issues/5277

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -171,17 +171,17 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
 }
 
 /**
-  Shorthand method for [broccoli-sourcemap-concat](https://github.com/ef4/broccoli-sourcemap-concat)
+  Shorthand method for [broccoli-concat](https://github.com/ember-cli/broccoli-concat)
 
   @private
   @method concatFiles
   @param {tree} tree Tree of files
-  @param {Object} options Options for broccoli-sourcemap-concat
+  @param {Object} options Options for broccoli-concat
   @return {tree} Modified tree
 */
 Addon.prototype.concatFiles = function(tree, options) {
   options.sourceMapConfig = this.app.options.sourcemaps;
-  return require('broccoli-sourcemap-concat')(tree, options);
+  return require('broccoli-concat')(tree, options);
 };
 
 /**


lib/models/addon.js
```diff
@@ -171,17 +171,17 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
 }
 
 /**
-  Shorthand method for [broccoli-sourcemap-concat](https://github.com/ef4/broccoli-sourcemap-concat)
+  Shorthand method for [broccoli--concat](https://github.com/ember-cli/broccoli--concat)
```
rwjblue: Looks like an extra dash here.

lib/models/addon.js
```diff
@@ -171,17 +171,17 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
...
 
   @private
   @method concatFiles
   @param {tree} tree Tree of files
-  @param {Object} options Options for broccoli-sourcemap-concat
+  @param {Object} options Options for broccoli--concat
```
rwjblue: Looks like an extra dash here

lib/models/addon.js
```diff
@@ -171,17 +171,17 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
...
   @return {tree} Modified tree
 */
 Addon.prototype.concatFiles = function(tree, options) {
   options.sourceMapConfig = this.app.options.sourcemaps;
-  return require('broccoli-sourcemap-concat')(tree, options);
+  return require('broccoli--concat')(tree, options);
```
rwjblue: Looks like an extra dash here

rwjblue:
Definitely looks good.  I'd like to get this under test also, to prevent future regressions. Adding usage of `concatFiles` to one of the dummy addons and adding a section for `concatFiles` to `tests/unit/models/addon-test.js` that has similar setup/teardown to the `treeForStyles` describe block should work well....

ballPointPenguin:
@rwjblue good catch. Too much vim in my vim :smiley_cat: 
I'll put in testing.

ballPointPenguin:
So... I don't actually know what I'm doing with this unit test.
I'm not very familiar with the inner workings of ember-cli addons, or broccoli. I took a stab at it but could use some help.

tests/fixtures/addon/with-concat-files/index.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = {
+  name: 'concat-some-files',
+
+  included: function(app) {
+    this._super.included(app);
+
+    var vendor = this.treeGenerator(path.join(__dirname, 'vendor'));
+
+    return this.concatFiles(vendor, {
```
stefanpenner: This is not to be public api. People should depend on their own dep of bsmc

rwjblue: I'm not sure why concatFiles was introduced honestly. I thought we learned our lesson after exposing this.pickFiles and such :smiley_cat:

rwjblue: Either way though, the method was completely untested (it would have errored loudly upon any invocation). I'd be in favor of deprecating this publicly in beta.3...

ballPointPenguin:
The only addon I am aware of using concatFiles is ember-cli-blanket:
https://github.com/sglanzer/ember-cli-blanket/blob/2fc593cd0e2aa4bad6c45f6ba14f163501be8c35/index.js#L82

ballPointPenguin:
I pulled my (broken) test out. It seems like this won't be merged in favor of deprecating this.concatFiles. But at least it'll merge cleanly without breaking the test suite, if it is merged.

rwjblue:
Tracking the larger issue of deprecating `this.concatFiles` in https://github.com/ember-cli/ember-cli/issues/5280.

stefanpenner:
:+1:

PR #5279, opened 2015-12-28
Use a repo that is in the org for testing Github blueprints

trek:
This will keep Trek from accidentally deleting important things. Already forked that repo into the org: https://github.com/ember-cli/app-blueprint-test


stefanpenner:
@homu r+

PR #5286, opened 2015-12-30
Update ember-cli-sri minimum version to 2.0.0

jonathanKingston:
To enable SRI when unicode is present in the file as Chrome is now fixed meaning the paranoia check can be removed.


rwjblue:
@homu r+

PR #5291, opened 2015-12-30
getCallerFile now comes from its own node_module

stefanpenner:


rwjblue:
Looks like one usage was missed here: https://github.com/ember-cli/ember-cli/blob/3c615aa55dfe6be1adf14ef67e74028b09560138/lib/utilities/require-as-hash.js#L19.  This is the source of the Travis failures.

stefanpenner:
@rwjblue danke danke, will fix momentarily

stefanpenner:
@homu r+

Issue #4948, opened 2015-10-12
outputPaths always send the app css


dukex:
Hi everyone

Given the current [ember-app.js#L183-L185](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L183-L185):

```
    app: {
...
      css: {
        'app': '/assets/' + this.name + '.css'
      }
``` 

This 'requires' that all projects using ember-cli has the app.css file, and cause a bug in projects like [ember-cli-sass](https://github.com/aexmachina/ember-cli-sass) and [ember-cli-less](https://github.com/gdub22/ember-cli-less) when using the follow config:

````
var app = new EmberApp({
  outputPaths: {
    app: {
      css: {
        'theme-orange': '/assets/theme-orange.css',
        'theme-purple': '/assets/theme-purple.css'
      }
    }
  }
});
```

I believe the problem is in ember-cli because the outputPaths always has the app css even when the app override the config

```
> options.outputPaths
{ 'theme-orange': '/assets/theme-orange.css',
  'theme-purple': '/assets/theme-purple.css',
  app: '/assets/myapp.css' }
```


Related: 
https://github.com/aexmachina/ember-cli-sass/issues/81
https://github.com/gdub22/ember-cli-less/issues/30

aexmachina:
Any movement on this issue? Is the fix to remove any keys that are not defined in the incoming array?

zzarcon:
I'm having the same issue in my app, some news?

dukex:
@zzarcon I send a pull request

zzarcon:
@dukex :heart_eyes:

Turbo87:
@dukex have you tried setting it like this?

```js
var app = new EmberApp({
  outputPaths: {
    app: {
      css: {
        'app': null,
        'theme-orange': '/assets/theme-orange.css',
        'theme-purple': '/assets/theme-purple.css'
      }
    }
  }
});
```

nathanhammond:
Closing with the `app: null` workaround and lack of activity.

PR #5305, opened 2016-01-05
[WIP][BUGFIX] Not overwrite the custom css configuration 

dukex:
Today the `defaults` merge is overwrite the css `outputPath` configured always adding the app config.
This pull request fix https://github.com/ember-cli/ember-cli/issues/4948 and is related with https://github.com/aexmachina/ember-cli-sass/issues/81 

See the problem, given the follow config
```
var app = new EmberApp({
  outputPaths: {
    app: {
      css: {
        'theme-orange': '/assets/theme-orange.css',
        'theme-purple': '/assets/theme-purple.css'
      }
    }
  }
});
```
With this config today it's expected a app.css but the user explicitly overwrote the css config. In the other hand the guides teach the user put the app file when he is overwrite the css outputPath

```
var app = new EmberApp({
  outputPaths: {
    app: {
      css: {
        'app': '/assets/application-name.css',
        'themes/alpha': '/assets/themes/alpha.css'
      }
    }
  }
});
```

A alternative and easy fix is tell to users always have a app.css because the ember will always send to addon the `outputPaths` with `app`

ps: Maybe I need fix some acceptance tests

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -286,7 +286,9 @@ EmberApp.prototype._initVendorFiles = function() {
         }
       }
     ]
-  }, this.options.vendorFiles), isNull);
+  }, this.options.vendorFiles), function(value) {
+    return value === null;
+  });
 
   if (!!this.registry.availablePlugins['ember-resolver']) {
     // if the project is using `ember-resolver` as an addon
--- a/tests/unit/broccoli/ember-app-test.js
+++ b/tests/unit/broccoli/ember-app-test.js
@@ -97,6 +97,31 @@ describe('broccoli/ember-app', function() {
       });
     });
 
+    it('should not override custom css config if it was configured', function() {
+      var app = new EmberApp({
+        project: project
+      });
+
+      // testing the current behavior
+      expect(app.options.outputPaths.app.css).to.deep.equal({app: '/assets/test-project.css'});
+
+      var customCssConfig = {
+        'theme-one': '/assets/theme-one.css',
+        'theme-two': '/assets/theme-two.css',
+      };
+
+      app = new EmberApp({
+        outputPaths: {
+          app: {
+            css: customCssConfig
+          }
+        }
+      });
+
+      expect(app.options.outputPaths.app.css).to.not.have.property('app');
+      expect(app.options.outputPaths.app.css).to.deep.equal(customCssConfig);
+    });
+
     it('should do the right thing when merging default object options', function() {
       var app = new EmberApp({
         project: project,


stefanpenner:
cc @chadhietala i believe your change to fix the options merging may be related, can you :eyes:

lib/broccoli/ember-app.js
```diff
@@ -202,6 +199,13 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
     }
   });
 
+  // Do not override custom css config if it was configured
+  this.options.outputPaths.app = Object.assign({
```
chadhietala: If we are supporting Node 0.12.0 this will not work due to Object.assign not being a thing.

stefanpenner: we also support node 0.10.x

stefanpenner:
this needs a rebase:

@chadhietala can you take another look, `assign` now comes from `lodash`
@rwjblue r?

chadhietala:
Seems good.

lib/broccoli/ember-app.js
dukex: Thanks @chadhietala, fixed

dukex:
I need to check the tests failing.

lib/broccoli/ember-app.js
```diff
@@ -256,7 +256,7 @@ EmberApp.prototype._initVendorFiles = function() {
     handlebarsVendorFiles = null;
   }
 
-  this.vendorFiles = omitBy(merge({
+  this.vendorFiles = omit(merge({
```
rwjblue: I do not believe that we have `omit` being required any more, I believe it should be `omitBy` and this sholdn't need to be changed...

stefanpenner:
@dukex can you update based on the above feedback + rebase ?

dukex:
@stefanpenner sure, I'll do it today

lib/broccoli/ember-app.js
```diff
@@ -286,7 +286,9 @@ EmberApp.prototype._initVendorFiles = function() {
         }
       }
     ]
-  }, this.options.vendorFiles), isNull);
+  }, this.options.vendorFiles), function(value) {
```
rwjblue: I am not seeing how this is any different that what we already have since `_.isNull` is actually [the same code](https://github.com/lodash/lodash/blob/4.3.0/lodash.js#L10063):

```js
    function isNull(value) {
      return value === null;
    }
```

I'm also confused how the merging of `vendorFiles` is related to `outputPaths`.

stefanpenner:
@dukex ping

Turbo87:
closing due to inactivity

PR #5321, opened 2016-01-10
EmberAddon access to `env`

jasonmit:
This trolled me for a bit when `EmberApp.env` wasn't a function until I realized I was dealing with `ember-cli/lib/broccoli/ember-addon` and not `ember-cli/lib/broccoli/ember-app`

If this fine, I'll write some tests to cover it.

### DIFF
--- a/lib/broccoli/ember-addon.js
+++ b/lib/broccoli/ember-addon.js
@@ -54,6 +54,8 @@ function EmberAddon() {
   }));
 }
 
+EmberAddon.__proto__ = EmberApp;
+
 EmberAddon.prototype = Object.create(EmberApp.prototype);
 EmberAddon.prototype.constructor = EmberAddon;
 EmberAddon.prototype.appConstructor = EmberApp.prototype.constructor;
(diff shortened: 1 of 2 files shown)


rwjblue:
This looks good to me. Tests would be great...

lib/broccoli/ember-addon.js
```diff
@@ -54,6 +54,12 @@ function EmberAddon() {
   }));
 }
 
+for (var key in EmberApp) {
```
stefanpenner: this seems like it is trying to emulate `__proto__`

jasonmit: :+1: wasn't able to dig up whether it was supported in older versions of node but I tested it all the way back to 0.10.0 and seems to work.

stefanpenner: ya its a thing in anything ES5+

jasonmit:
@rwjblue ready when you are :+1:

stefanpenner:
sg

stefanpenner:
test failure is unrelated, and fix upstream. Should be good to go - thanks!

PR #5325, opened 2016-01-11
bump testem to latest to compensate for breakage in xmldom

et:
Our CI which uses the xunit reporter was reporting the following:

```js
/home/bamboo/node_modules/ember-cli/node_modules/testem/node_modules/xmldom/dom.js:906
        node.serializeToString(attributeSorter);
```

The fix for this was addressed in
https://github.com/testem/testem/pull/730
and a new release was deployed as v1.0.0-rc.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "silent-error": "^1.0.0",
     "symlink-or-copy": "^1.0.1",
     "temp": "0.8.3",
-    "testem": "^1.0.0-rc.4",
+    "testem": "^1.0.0-rc.7",
     "through": "^2.3.6",
     "tiny-lr": "0.2.1",
     "walk-sync": "^0.2.6",


rwjblue:
@homu r+

jayeff:
I'm using ember-cli 1.13.14. How can I update testem to v1.0.0-rc.7 to fix the problem with XUnit Reporter? Because testem is a dependency of ember-cli I seem to be stuck.

Is it necessary to backport this fix to the 1.13 branch and release a new version 1.13.15?

uitgewis:
Would also like this fix backported to 1.13.x please.

rwjblue:
This fix isn't technically needed (but it was nice to include to ensure we get the right minimum version).  All you should need to do in your project is clear `node_modules` and reinstall (which should grab the latest version).

`rm -rf node_modules && npm cache clear && npm install`

jayeff:
@rwjblue: Unfortunately does not work work for 1.13.13/1.13.14 as testem is pinned to exactly 0.9.11. 

Maybe an alternativ to updating the 1.13.x branch to testem to v1.0.0-rc.7 could be to backport the fix from testem v1.0.0-rc. to the 0.9.x branch, release testem 0.9.12 and then update the ember-cli with the newer testem version.

rwjblue:
@jayeff - Ahh, thank you for pointing that out RE: the 1.13 branch. Yes, I believe that a release in the 0.9.x series to fix makes sense.

@johanneswuerbach - Any objections?

rwjblue:
In theory, you can lock xmldom to an unaffected version at the app level also (basically installing the dep at the app level forces that version to be used throughout).  But I'm not sure what the working version of xmldom is, so you'll have to research that...

johanneswuerbach:
@rwjblue backporting should definitely be possible as no breaking changes were done api wise. I'll also look into backporting that qunit fix into testem `0.9` later today, but that dependency still needs to be unpinned.

et:
Thanks for looking into the backport @johanneswuerbach 

For those that are on the 1.13.x branch and having broken builds, an option to is shrinkwrap your dependencies.
`npm shrinkwrap`
`vim npm-shrinkwrap.json`

Find the reference to `xmldom` and change it to the following:
```json
    "xmldom": {
      "version": "0.1.19",
      "from": "xmldom@>=0.1.19 <0.2.0",
      "resolved": "https://registry.npmjs.org/xmldom/-/xmldom-0.1.19.tgz"
    },
```

stefanpenner:
@johanneswuerbach are we near a 1x release? Should we just bundle the release beta?

johanneswuerbach:
I haven't received any further issue for a while so I guess v1 is good to go.

johanneswuerbach:
Bumped and https://github.com/ember-cli/ember-cli/pull/5332

jayeff:
> In theory, you can lock xmldom to an unaffected version at the app level also (basically installing the dep at the app level forces that version to be used throughout).

@rwjblue How would this work? If I add `"xmldom": "0.1.19"` to my package.json it will install 0.1.19 of xmldom on app level. But testem will still install and use it's dependencies with the broken version 0.1.20 of xmldom. Maybe I'm doing something wrong?

@johanneswuerbach Congrats on 1.0.0 :tada:

stefanpenner:
@johanneswuerbach :tada:

PR #5326, opened 2016-01-11
[BUGFIX release] pin jQuery version.

stefanpenner:
* hard coded check in ember exists, causing grief with the recent jQuery release
* a fix is in ember, but it cannot be released quite yet.
* this allows new apps to work in the interim, and users upgrading will also benefit


kellyselden:
:+1:

stefanpenner:
@homu r+

PR #5329, opened 2016-01-11
[ENHANCEMENT] Support custom blueprint options in new and init commands

bendemboski:
The new and init commands can be given a custom blueprint to use. This modifies those commands to be aware of blueprint-specific options and pass them through, like the generate command.

This is handy in the near-term for custom app/addon blueprints that accept options, and could pave the way to add options to the built-in blueprints if that ever seems valuable. A few examples I can think of immediately are the server port and the root URL.

I didn't make any changes to the help output because the default blueprints don't support any custom options and ```ember help new appName --blueprint=./my-blueprint``` seems like a weird use case.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -1,5 +1,8 @@
 'use strict';
 
+var copy               = require('lodash/lang/clone');
+var merge              = require('lodash/object/merge');
+var Blueprint          = require('../models/blueprint');
 var Command            = require('../models/command');
 var Promise            = require('../ext/promise');
 var SilentError        = require('silent-error');
@@ -32,6 +35,27 @@ module.exports = Command.extend({
     } else {
       return 'app';
     }
+  },
+
+  beforeRun: function(rawArgs) {
+    if (!rawArgs.length) {
+      return;
+    }
+    // merge in blueprint availableOptions
+    var blueprint;
+    try{
+      blueprint = this.lookupBlueprint(rawArgs[0]);
+      this.registerOptions(blueprint);
+    }
+    catch(e) {
+      SilentError.debugOrThrow('ember-cli/commands/init', e);
+    }
+  },
+
+  lookupBlueprint: function(name) {
+    return Blueprint.lookup(name, {
+      paths: this.project.blueprintLookupPaths()
+    });
   },
 
   run: function(commandOptions, rawArgs) {
@@ -82,19 +106,18 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(message));
     }
 
-    var blueprintOpts = {
-      dryRun: commandOptions.dryRun,
-      blueprint: commandOptions.blueprint || this._defaultBlueprint(),
+    var blueprintOpts = copy(commandOptions);
+    merge(blueprintOpts, {
       rawName: packageName,
       targetFiles: rawArgs || '',
       rawArgs: rawArgs.toString()
-    };
+    });
 
     if (!validProjectName(packageName)) {
       return Promise.reject(new SilentError('We currently do not support a name of `' + packageName + '`.'));
     }
 
-    blueprintOpts.blueprint = normalizeBlueprint(blueprintOpts.blueprint);
+    blueprintOpts.blueprint = normalizeBlueprint(blueprintOpts.blueprint || this._defaultBlueprint());
 
     debug('before:installblueprint');
     return installBlueprint.run(blueprintOpts)
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -1,14 +1,14 @@
 'use strict';
 
-var copy               = require('lodash/lang/clone');
-var merge              = require('lodash/object/merge');
-var Blueprint          = require('../models/blueprint');
-var Command            = require('../models/command');
-var Promise            = require('../ext/promise');
-var SilentError        = require('silent-error');
-var validProjectName   = require('../utilities/valid-project-name');
-var normalizeBlueprint = require('../utilities/normalize-blueprint-option');
-var debug              = require('debug')('ember-cli:command:init');
+var copy                  = require('lodash/lang/clone');
+var merge                 = require('lodash/object/merge');
+var Command               = require('../models/command');
+var Promise               = require('../ext/promise');
+var SilentError           = require('silent-error');
+var validProjectName      = require('../utilities/valid-project-name');
+var normalizeBlueprint    = require('../utilities/normalize-blueprint-option');
+var mergeBlueprintOptions = require('../utilities/merge-blueprint-options');
+var debug                 = require('debug')('ember-cli:command:init');
 
 module.exports = Command.extend({
   name: 'init',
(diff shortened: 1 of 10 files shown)


stefanpenner:
looks great. @trek would love your :eyes: since the remote/external blueprint was your creation, want to make sure this is aligned with your goals for it.

tests/unit/commands/new-test.js
```diff
@@ -90,4 +102,53 @@ describe('new command', function() {
...
+          { name: 'custom-blueprint-option', type: String }
+        ]
+      };
+    }, true);
+
+    try {
+      command.beforeRun(['app']);
+      expect(pluck(command.availableOptions, 'name')).to.contain('custom-blueprint-option');
+    } finally {
+      safeRestore(Blueprint, 'lookup');
```
kellyselden: our other tests have the `safeRestore` in the `afterEach` so it's only needed once for all tests.

bendemboski: Okay. I didn't want to stub out ```lookup``` in every test, but you're suggesting that I put the ```safeRestore``` in ```afterEach```, but keep the ```stub``` in the test body?

stefanpenner: :+1:

kellyselden: Yep, we added the safe version with a string arg so it only restores if it was stubbed.

stefanpenner:
kicking windows

trek:
@bendemboski can you give some usage examples in the description?

lib/commands/init.js
```diff
@@ -34,6 +37,27 @@ module.exports = Command.extend({
...
+  beforeRun: function(rawArgs) {
+    if (!rawArgs.length) {
+      return;
+    }
+    // merge in blueprint availableOptions
+    var blueprint;
+    try{
+      blueprint = this.lookupBlueprint(rawArgs[0]);
+      this.registerOptions(blueprint);
+    }
```
stefanpenner: style nitpic: `} catch(e) {`

lib/commands/init.js
```diff
@@ -34,6 +37,27 @@ module.exports = Command.extend({
     }
   },
 
+  beforeRun: function(rawArgs) {
+    if (!rawArgs.length) {
+      return;
+    }
+    // merge in blueprint availableOptions
+    var blueprint;
+    try{
```
stefanpenner: style nitpic: `try {`

lib/commands/init.js
```diff
@@ -34,6 +37,27 @@ module.exports = Command.extend({
     }
   },
 
+  beforeRun: function(rawArgs) {
+    if (!rawArgs.length) {
```
stefanpenner: `rawArgs.length === 0`

lib/commands/init.js
```diff
@@ -82,19 +106,18 @@ module.exports = Command.extend({
...
       rawArgs: rawArgs.toString()
-    };
+    });
 
     if (!validProjectName(packageName)) {
       return Promise.reject(new SilentError('We currently do not support a name of `' + packageName + '`.'));
     }
 
-    blueprintOpts.blueprint = normalizeBlueprint(blueprintOpts.blueprint);
+    blueprintOpts.blueprint = normalizeBlueprint(blueprintOpts.blueprint || this._defaultBlueprint());
```
stefanpenner: why not do this during the marge on 110?

lib/commands/new.js
```diff
@@ -29,6 +30,27 @@ module.exports = Command.extend({
     '<app-name>'
   ],
 
+  beforeRun: function(rawArgs) {
+    if (!rawArgs.length) {
```
stefanpenner: `rawArgs.length === 0`

lib/commands/new.js
```diff
@@ -29,6 +30,27 @@ module.exports = Command.extend({
...
+      return;
+    }
+    // merge in blueprint availableOptions
+    var blueprint;
+    try{
+      blueprint = this.lookupBlueprint(rawArgs[0]);
+      this.registerOptions(blueprint);
+    }
+    catch(e) {
+      SilentError.debugOrThrow('ember-cli/commands/init', e);
```
stefanpenner: should likely be `ember-cli/commands/new`
both this and the `init` variant look very similar, maybe we should create a helper method that does this?

tests/unit/commands/init-test.js
```diff
@@ -200,4 +210,37 @@ describe('init command', function() {
...
+        return Promise.reject('Called run');
+      }
+    });
+
+    buildCommand();
+
+    return command.validateAndRun(['--custom-option=customValue'])
+      .catch(function(reason) {
+        expect(reason).to.equal('Called run');
+      });
```
stefanpenner: this test may pass, if the promise does not reject. This can easily result in false positives, we should likely assert the promise returned from `validateAndRun` always rejects

lib/commands/init.js
bendemboski: These are all my stylistic preferences as well, but I basically copied this method from [here](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/generate.js#L32). Should I prefer style over consistency here?

lib/commands/new.js
bendemboski: Same question here about style vs. consistency.

lib/commands/new.js
bendemboski: Yes, agreed. I'm new to ember-cli development, and am still figuring out the patterns. Creating a base class with the method seems like overkill. A new module in utilities perhaps?

```javascript
// utilities/merge-blueprint-options.js
function mergeBlueprintOptions(command, rawArgs) {
  ...
}
```

and

```javascript
import mergeBlueprintOptions from '../utilities/merge-blueprint-options';

module.exports = Command.extend({
  ...
  beforeRun: function(rawArgs) {
    mergeBlueprintOptions(this, rawArgs);
  }
  ...
});
```

lib/commands/init.js
bendemboski: Sure

tests/unit/commands/init-test.js
bendemboski: :+1:

lib/commands/init.js
stefanpenner: I would prefer the above mentioned style guide.

I would love to enforce style with JSCS or similar sometime soon, that way we can keep this consistent.

lib/commands/new.js
stefanpenner: this is my preference, as it (without coercion) indicates our exact intentions

lib/commands/new.js
stefanpenner: @bendemboski yes, a single module in a method sounds great.

Another approach could be:

```js
Command.extend({
  mergeBlueprintOptions: mergeBlueprintOptions,
  beforeRun: function(rawArgs) {
    this.mergeBlueprintOptions(rawArgs);
  }
})
```
or even more concisely 

```js
Command.extend({
  beforeRun: mergeBlueprintOptions
})
```

stefanpenner:
single failure is unrelated and fixed upstream. Thanks for the contribute :)

PR #5340, opened 2016-01-13
adding `git clean -fdx` to `prepare-release`

kellyselden:
An additional step for safety
cc @stefanpenner

### DIFF
--- a/bin/prepare-release
+++ b/bin/prepare-release
@@ -1,7 +1,6 @@
 #!/bin/sh
 
-rm -rf node_modules && \
-rm -rf ./ember-cli-*.tgz && \
+git clean -fdx && \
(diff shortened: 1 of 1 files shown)


kellyselden:
friendly ping @stefanpenner

bin/prepare-release
```diff
@@ -3,6 +3,7 @@
 rm -rf node_modules && \
 rm -rf ./ember-cli-*.tgz && \
```
stefanpenner: we can drop these first two lines, as the `git clean -fdx` should do the trick

kellyselden: updated

stefanpenner:
:+1:

PR #5345, opened 2016-01-13
use eol=lf for windows

kellyselden:
from the file:
```
# force unix line endings on all files, needed for Windows npm publishing
# https://github.com/ember-cli/ember-cli/issues/5342
# this may be a temporary solution, as it may affect the dev flow for Windows
# contributors that aren't npm publishers
# another solution may be https://www.npmjs.com/package/npm-text-auto
```
cc @stefanpenner @rwjblue

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# force unix line endings on all files, needed for Windows npm publishing
+# https://github.com/ember-cli/ember-cli/issues/5342
+# this may be a temporary solution, as it may affect the dev flow for Windows
+# contributors that aren't npm publishers
+# another solution may be https://www.npmjs.com/package/npm-text-auto
+* text eol=lf
+*.png binary


stefanpenner:
> +# this may be a temporary solution, as it may affect the dev flow for Windows

i've always been confused by this, is it reasonable to expect this to not be an issue for windows contributors ? Or is it reasonable to expect it.

kellyselden:
In my experience, it's nothing more than a minor inconvenience. Any programming editor or IDE will render the `\n`. The only time you will notice is if you open it up in `notepad.exe`, all lines are munged onto one line. Then you know just to open it in another editor.

Would love to hear from some other Windows devs.

stefanpenner:
@kellyselden I'm fine with requiring unix new lines, even for windows developers. As long as we quickly document it in the readme, then I'm on-board.

rwjblue:
Seems reasonable to try. We can always revisit if it causes Windows devs of ember-cli pain.

PR #5353, opened 2016-01-14
Copy symlinks without dereference after broccoli build when env is development

jpmarques:
After the broccoli build we are always copying the files and dereference them.

This leads to high build times in cases where we have big files (my project, for example, includes 90MB of assets which take forever to be copied to the outputPath).

I'm testing directly the env which isn't clearly the best approach here. I believe we should allow the desired behavior (either dereference or not) to be configured somewhere (config\environment.js?) for each environment type.

I was thinking something like the following (I’m just not sure how to test the variable inside the builder).
if (environment === 'development') {
  ENV.dereferenceOnBuild = false;
}

Any thoughts on this?

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -13,6 +13,7 @@ var attemptNeverIndex = require('../utilities/attempt-never-index');
 var findBuildFile = require('../utilities/find-build-file');
 var viz = require('broccoli-viz');
 var FSMonitor = require('fs-monitor-stack');
+var symlinkOrCopySync = require('symlink-or-copy').sync;
 
 var signalsTrapped = false;
 var buildCount = 0;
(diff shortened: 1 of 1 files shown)


stefanpenner:
> configured somewhere (config\environment.js?)

this is for client-side configuration, I suspect you want serverside.

stefanpenner:
If we are to accept this, I would prefer to see a seperate tested module to replace CPD here, inlining this code will just lead to extra tech debt.

rwjblue:
I'm generally :+1: on the direction. I'd like to see the extraction as Stef mentioned above and some tests to be able to land this...

jpmarques:
@stefanpenner Thanks for pointing out ember-cli-build.js, all makes sense now.

I will have a look into creating a new module.

stefanpenner:
should be superseded by https://github.com/ember-cli/ember-cli/pull/5370

PR #5354, opened 2016-01-14
[Fixes #5352] fix issue were we accidentally were checking unexpected…

### LINKED ISSUE #5352: live-reload host stuff

stefanpenner:
… interface for available ports

* configure baseport as an option, and not as something on the contructor

*note: this may need more investigation*


stefanpenner:
this will be addressed by fixing portfinder to support ipv4 and ipv6

PR #5356, opened 2016-01-14
add .npmignore to limit unnecessary files in package

kellyselden:


.npmignore
```diff
@@ -0,0 +1,9 @@
+assets
+blueprints/.jshintrc
+dev
+docs
+tests
```
stefanpenner: please depend on some of the test helpers

kellyselden: fixed

stefanpenner:
I think this is good, @rwjblue can you :eyes: you are likely to remember some of the usages.

kellyselden:
I need to make some changes, hold off on merge please

kellyselden:
made changes, gotta keep the beginning slash to only check root folders. also wildcard exclude all files at root not in folders because the package doesn't need any

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

here's the result of `npm pack`

capture

@stefanpenner @rwjblue ready for review again.

kellyselden:
friendly ping @stefanpenner @rwjblue

.npmignore
```diff
@@ -0,0 +1,7 @@
+/assets
+/blueprints/.jshintrc
+/dev
+/docs
+/tests/!(helpers)
+/tmp
+/*.*
```
stefanpenner: what does this match?

kellyselden: all root level files, aside from the automatically excluded readme, changlog, license, and package.json

kellyselden: @stefanpenner did you see the image here? https://github.com/ember-cli/ember-cli/pull/5356#issuecomment-172391663

stefanpenner: Ah so this doesn't include package.json right? That one is essentially white-listed.

rwjblue: Confirm. There are a list of files that you cannot ignore (per npm).

PR #5363, opened 2016-01-15
syncing the bin/changelog from master

kellyselden:
needed for homu support


rwjblue:
:+1:

stefanpenner:
@kellyselden it would be really cool to extract this as a module (it would be handy to use in other projects)

kellyselden:
@stefanpenner I agree, I'll file an issue.

kellyselden:
#5366

PR #5369, opened 2016-01-17
Implement application level resolver in blueprint.

### LINKED ISSUE #5368: resolver in app + tests should come from same location
1. add `app/resolver.js` which by default `export { default } from 'ember-resolver';`
2. in `app.js` -> `import resolver from './resolver';`
3. in [tests/helpers/resolver.js](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/tests/helpers/resolver.js) we should import `<app-name>/resolver`

----

obviously needs deprecation phase etc.

rwjblue:
Closes #5368.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import Resolver from 'ember-resolver';
+import Resolver from './resolver';
 import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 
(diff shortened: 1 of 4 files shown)


rwjblue:
#5368 discusses a deprecation phase, but I'm not sure that is needed here. New apps would be generated with the correct content, any properly upgraded apps will be updated to the new content, and any apps that do not update content will continue to work.

stefanpenner:
Ya. I can go either way, i figured the deprecation would be nice to quickly get everyone on-board, but meh.

stefanpenner:
@homu r+

rwjblue:
Wat, says it passed and merged but no merge :(

PR #5377, opened 2016-01-19
Update Bower resolution to match Ember version

kpfefferle:
It looks like the 2.3.0-beta.1 release updated the Ember version, but not the matching resolution in `bower.json`


kpfefferle:
cc: @rwjblue who bumped the Ember version

rwjblue:
@homu r+

PR #5380, opened 2016-01-21
[ENHANCEMENT] print double dash for long aliases in availableOptions

ghedamat:
Presently if one alias for `availableOptions` is a long word the help function for that command will still print the option with a single dash although the `--` version works as well.

i.e.
https://github.com/ember-cli/ember-cli/blob/282641ba662292afb40cb88cdf31557a3e4cc6b7/lib%2Fcommands%2Fbuild.js#L13
will be printed as `-dev` even if also `--dev` is supported.

before this patch:
```
aliases: -a <value>, -long-a <value>, -b (--test-option=c), -long-b (--test-option=c)
```

after:
```
aliases: -a <value>, --long-a <value>, -b (--test-option=c), --long-b (--test-option=c)
```

### DIFF
--- a/lib/utilities/print-command.js
+++ b/lib/utilities/print-command.js
@@ -73,10 +73,10 @@ module.exports = function(initialMargin, shouldDescriptionBeGrey) {
     if (option.aliases && option.aliases.length) {
       output += EOL + initialMargin + '    ' + chalk.grey('aliases: ' + option.aliases.map(function(a) {
         if (typeof a === 'string') {
-          return '-' + a + (option.type === Boolean ? '' : ' <value>');
+          return (a.length > 4 ? '--': '-') + a + (option.type === Boolean ? '' : ' <value>');
         } else {
           var key = Object.keys(a)[0];
-          return '-' + key + ' (--' + option.name + '=' + a[key] + ')';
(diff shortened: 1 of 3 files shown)


ghedamat:
tests will fail as some existing options will be changed by this (in the help) 

i.e.
```diff
   --live-reload-port (Number) (Defaults to port number within [49152...65535])
-    aliases: --lrp <value>
+    aliases: -lrp <value>
```

before going in and changing the tests I'd like to know if the ember-cli team is ok with this change or if the current is intended behaviour

thanks!

stefanpenner:
This lgtm, @trabus / @rwjblue r+

ghedamat:
thanks @stefanpenner , if we set the limit to 4 chars instead of one the only help that would change is for the `-in-repo` and `-inspr` aliases

rwjblue:
Ya, I agree that this is a good idea.

ghedamat:
@stefanpenner @rwjblue appveyor is still running but I think we're good

I settled on a 4 char limit, let me know if that sounds good

thanks!

trabus:
lgtm :+1:

stefanpenner:
@homu r+

stefanpenner:
@homu retry

PR #5387, opened 2016-01-22
Ensure prepended app.imports are properly ordered.

rwjblue:
Prior to this change given the following:

```js
app.import('foo/bar/one.js', { prepend: true });
app.import('foo/bar/two.js', { prepend: true });
```

The resulting output would include `two.js` *BEFORE* `one.js` (because the last prepend was being `unshift`'ed onto the same array).

After this change, the list of prepended assets is managed separately and in order of `app.import`.  This means that given the snippet above, `one.js` is before `two.js` (and sanity is restored to the world).

---

This is supporting the changes discussed in https://github.com/ember-cli/ember-cli/pull/5379#issuecomment-173765481.

/cc @nathanhammond

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -95,9 +95,11 @@ function EmberApp() {
   this._initVendorFiles();
 
   this.legacyFilesToAppend     = [];
+  this.filesToPrepend          = [];
   this.vendorStaticStyles      = [];
   this.otherAssetPaths         = [];
   this.legacyTestFilesToAppend = [];
+  this.testFilesToPrepend      = [];
   this.vendorTestStaticStyles  = [];
 
   this.trees = this.options.trees;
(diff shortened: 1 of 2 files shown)


stefanpenner:
this may likely need to be called out as a 2x potential breaking change, some may rely on this.

stefanpenner:
> The resulting output would include two.js BEFORE one.js (because the last prepend was being unshift'ed onto the same array).

i mean that actually sounds reasonable no?

rwjblue:
> i mean that actually sounds reasonable no?

It was pretty odd to me on first blush.  Consider the scenario when separate addons using `before` or `after` to force being prepended in the right place. You would have to specify `after: 'addonA'` to be able to put your content before it.  Seems weird to me....

---

Ultimately, if you prefer the current behavior, I do understand.  It seemed like a bug to me so I fixed it :smiling_imp:...

rwjblue:
I do get what you mean though.  I guess the issue is what the mental model of people using `prepend` is.  To me it means "put this stuff before the normal vendor stuff", which this PR still supports.  I would prefer to use `before` / `after` and the DAG for addon ordering.

stefanpenner:
> I would prefer to use before / after and the DAG for addon ordering.

I don't believe that is granular enough, its totally reasonable to have 1 add-on that must be both before and after another for different things. Anyways that is somewhat orthogonal.

My vote is to leave this as is because:

1. is isn't unreasonable
2. changing it would potentially break apps for people who depend on the current ordering.

rwjblue:
@stefanpenner - Sounds good.  I'll close this for now, we can always make changes if we need to in the future...

PR #5390, opened 2016-01-22
only hide folders at root in .gitignore

kellyselden:
I noticed this while making the .npmignore. `tmp*` had the side-effect of hiding `/tests/helpers/tmp.js`. I remember @tomdale was burned by this in the past, since npm used .gitignore to exclude files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /node_modules
 npm-debug.log
-tmp*
-common-tmp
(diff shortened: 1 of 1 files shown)


tomdale:
@kellyselden Wow, good memory. :) That was very painful to debug. Thanks for doing this!

stefanpenner:
@homu r+

PR #5391, opened 2016-01-23
Fix host interface + port binding issues

### LINKED ISSUE #5352: live-reload host stuff

eriktrom:
Fixes #5352 

also fixes the same issue that the express server had as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "nopt": "^3.0.1",
     "npm": "2.14.21",
     "ora": "^0.2.0",
-    "portfinder": "^1.0.3",
+    "ember-cli-portfinder": "^1.0.3",
     "promise-map-series": "^0.2.1",
     "quick-temp": "0.1.5",
     "readline2": "0.1.1",
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "nopt": "^3.0.1",
     "npm": "2.14.21",
     "ora": "^0.2.0",
-    "portfinder": "^1.0.2",
+    "portfinder": "^1.0.3",
     "promise-map-series": "^0.2.1",
     "quick-temp": "0.1.5",
     "readline2": "0.1.1",
--- a/tests/unit/commands/serve-test.js
+++ b/tests/unit/commands/serve-test.js
@@ -6,7 +6,7 @@ var stub           = require('../../helpers/stub');
 var commandOptions = require('../../factories/command-options');
 var Task           = require('../../../lib/models/task');
 var Promise        = require('../../../lib/ext/promise');
-var PortFinder     = require('portfinder');
+var PortFinder     = require('ember-cli-portfinder');
 
 PortFinder.basePort = 32768;
 
--- a/tests/unit/commands/serve-test.js
+++ b/tests/unit/commands/serve-test.js
@@ -2,23 +2,18 @@
 
 var expect         = require('chai').expect;
 var EOL            = require('os').EOL;
-var proxyquire     = require('proxyquire');
 var stub           = require('../../helpers/stub');
 var commandOptions = require('../../factories/command-options');
 var Task           = require('../../../lib/models/task');
+var Promise        = require('../../../lib/ext/promise');
+var PortFinder     = require('portfinder');
+
+PortFinder.basePort = 32768;
 
 var safeRestore = stub.safeRestore;
 stub = stub.stub;
 
-var getPortStub;
-
-var ServeCommand = proxyquire('../../../lib/commands/serve', {
-  'portfinder': {
-    getPort: function() {
-      return getPortStub.apply(this, arguments);
-    }
-  }
-});
+var ServeCommand = require('../../../lib/commands/serve');
 
 describe('serve command', function() {
   var tasks, options, command;
@@ -33,9 +28,6 @@ describe('serve command', function() {
     });
 
     stub(tasks.Serve.prototype, 'run');
-    getPortStub = function(options, callback) {
-      callback(null, 49152);
-    };
 
     command = new ServeCommand(options);
   });
@@ -44,7 +36,19 @@ describe('serve command', function() {
     safeRestore(tasks.Serve.prototype, 'run');
   });
 
-  it('has correct options', function() {
+  it('has correct default options', function() {
+    return command.validateAndRun([
+    ]).then(function() {
+      var serveRun = tasks.Serve.prototype.run;
+      var runOps = serveRun.calledWith[0][0];
+
+      expect(serveRun.called).to.equal(1, 'expected run to be called once');
+      expect(runOps.port).to.be.gte(4000, 'has correct port');
+      expect(runOps.liveReloadPort).to.be.within(32768, 65535, 'has correct liveReload port');
+    });
+  });
+
+  it('setting --port without --live-reload-port', function() {
     return command.validateAndRun([
       '--port', '4000'
     ]).then(function() {
@@ -53,8 +57,56 @@ describe('serve command', function() {
 
       expect(serveRun.called).to.equal(1, 'expected run to be called once');
 
-      expect(runOps.port).to.equal(4000, 'has correct port');
-      expect(runOps.liveReloadPort).to.be.within(49152, 65535, 'has correct liveReload port');
+      expect(runOps.port).to.be.gte(4000, 'has port >= than chosen port');
+      expect(runOps.liveReloadPort).to.be.within(32768, 65535, 'has live reload port > port');
+    });
+  });
+
+  it('setting both --port and --live-reload-port', function() {
+    return command.validateAndRun([
+      '--port', '4000',
+      '--live-reload-port', '8005'
+    ]).then(function() {
+      var serveRun = tasks.Serve.prototype.run;
+      var runOps = serveRun.calledWith[0][0];
+
+      expect(serveRun.called).to.equal(1, 'expected run to be called once');
+
+      expect(runOps.port).to.be.gte(4000, 'has correct port');
+      expect(runOps.liveReloadPort).to.be.within(8005, 65535, 'has live reload port > port');
+    });
+  });
+
+  it('should throw error when -p PORT is taken', function() {
+    var originalEnvPort = process.env.PORT;
+
+    function testServer(opts, test) {
+      var server = require('http').createServer(function() {});
+      return new Promise(function(resolve) {
+        process.env.PORT = 65525;
+        server.listen(opts.port, opts.host, function() {
+          resolve(test(opts, server));
+        });
+      }).finally(function() {
+        return new Promise(function(resolve) {
+          server.close(function() { resolve(); });
+          process.env.PORT = originalEnvPort;
+        });
+      });
+    }
+
+    return testServer({ port: '65525' }, function() {
+      return command.validateAndRun([
+        '--port', '65525'
+      ])
+      .then(function() {
+        expect(true).to.equal(false, 'assertion should never run');
+      })
+      .catch(function(err) {
+        var serveRun = tasks.Serve.prototype.run;
+        expect(serveRun.called).to.equal(0, 'expected run to never be called');
+        expect(err.message).to.contain('is already in use.');
+      });
     });
   });
 
@@ -66,8 +118,8 @@ describe('serve command', function() {
       var runOps = serveRun.calledWith[0][0];
 
       expect(serveRun.called).to.equal(1, 'expected run to be called once');
-
-      expect(runOps.port).to.be.within(49152, 65535, 'has correct port');
+      expect(runOps.port).to.be.within(32768, 65535, 'has correct liveReloadPort');
+      expect(runOps.liveReloadPort).to.be.gt(runOps.port, 'has a liveReload port greater than port');
     });
   });
 
@@ -79,18 +131,11 @@ describe('serve command', function() {
       var ops = serveRun.calledWith[0][0];
 
       expect(serveRun.called).to.equal(1, 'expected run to be called once');
-
-      expect(ops.liveReloadPort).to.equal(4001, 'has correct liveReload port');
+      expect(ops.liveReloadPort).to.be.gte(4001, 'has correct liveReload port');
     });
   });
 
-  it('has correct liveLoadHost', function() {
-    var getPortOpts;
-    getPortStub = function(options, callback) {
-      getPortOpts = options;
-      callback(null, 49152);
-    };
-
+  it('has correct liveReloadLoadHost', function() {
     return command.validateAndRun([
       '--live-reload-host', '127.0.0.1'
     ]).then(function() {
@@ -98,8 +143,6 @@ describe('serve command', function() {
       var runOps = serveRun.calledWith[0][0];
 
       expect(serveRun.called).to.equal(1, 'expected run to be called once');
-
-      expect(getPortOpts.host).to.equal('127.0.0.1', 'gets a port based on the liveReload host');
       expect(runOps.liveReloadHost).to.equal('127.0.0.1', 'has correct liveReload host');
     });
   });
@@ -125,7 +168,6 @@ describe('serve command', function() {
       var ops = serveRun.calledWith[0][0];
 
       expect(serveRun.called).to.equal(1, 'expected run to be called once');
-
       expect(ops.proxy).to.equal('http://localhost:3000/', 'has correct port');
     });
   });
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -101,7 +101,7 @@
     "nopt": "^3.0.1",
     "npm": "2.14.10",
     "pleasant-progress": "^1.0.2",
-    "portfinder": "^0.4.0",
+    "portfinder": "eriktrom/node-portfinder",
```
eriktrom: change when port finder pr merges

tests/unit/commands/serve-test.js
```diff
@@ -54,7 +68,159 @@ describe('serve command', function() {
...
+        // TODO: this test should throw. It does in real life. We've never tested
+        // the following error though yet:
+        //
+        // Example: say nginx is running on port 8000, this is what we
+        // want to test for, our expectation:
+        //
+        // ember serve -lrp 8000 -p 8000 -lrh ::1 -h ::1
+        // version: 2.3.0-beta.1-dev-984fefccda
+        // Could not serve on http://::1:8000. It is either in use or you do not have permission.
+      });
```
eriktrom: i will cleanup these tests - truth is they're not really very accurate anyway, thoughts welcome

to actually test npm link to your project, start a server on a known port(or a couple one for express, another for live reload) then play with ember server -p PORT -lrp LIVERELOADPORT -h HOST -lrh LIVERELOADHOST

manually trying all the variations, including port 0 works. these tests though, they do lie and dont catch all edge cases, fyi

tests/unit/commands/serve-test.js
```diff
@@ -54,7 +68,159 @@ describe('serve command', function() {
...
+
+      return Promise.all(variations.map(function(variation) {
+        return testVariation(variation);
+      })).then(function() {
+        expect(tasks.Serve.prototype.run.called).to.equal(variations.length, 'expected run to be called '+variations.length);
+      });
+    });
+  }
+
+  // will eventually fail if you set the number of iterations high enough
```
eriktrom: no longer true, at least i cant make it fail anymore, which is good. i think async.series may have fixed it (instead of Promise.all)

stefanpenner:
Ya that rag is expected. Short of adding a spinlock or something we are out of luck. And that is likely overkill

stefanpenner:
I'm on my phone. I will review it more closely this weekend

lib/tasks/serve.js
```diff
@@ -52,13 +52,16 @@ module.exports = Task.extend({
       expressServer: expressServer
     });
 
-    return Promise.all([
-        liveReloadServer.start(options),
-        expressServer.start(options)
-      ]).then(function() {
+    // note Promise.all will hang under certain cirumstances, aka
```
stefanpenner: why would that hang? I could imagine the all rejects, unless their is an upstream bug not fulfilling or rejecting.

tests/runner.js
```diff
@@ -50,8 +50,11 @@ function checkOnlyInTests() {
 }
 
 function runMocha() {
+  var start = Date.now();
   mocha.run(function(failures) {
     process.on('exit', function() {
+      var end = Date.now();
+      console.log('Running Time of Tests (in minutes)', ((end - start)/1000)/60);
```
stefanpenner: :+1: maybe we should add this to another PR, so we can sneak it in quickly.

stefanpenner: console.time + console.timeEnd may also work

lib/tasks/serve.js
eriktrom: when we launch the express server, we also use the getPort check, just before calling createServer(or whatever) - if this check fails, I throw. Promise.allSettled may fix this(bluebird has this sort of function but don't recall rsvp)? - but Promise.all does not catch the exception I throw when the port is taken

this is different the live reload case where we just bump up to the next port. in the express server we want to shutdown b/c the port they are expecting to open should not as users would not notice that we instead bumped it(in the console output)

related to this is the test port - this does not port scan - this is something i think we should add to this pr

tests/runner.js
eriktrom: made a pr for that: https://github.com/ember-cli/ember-cli/pull/5398

stefanpenner:
@eriktrom :+1:

eriktrom:
passes on travis :)

once portfinder pr merges i'll bump the version in package.json and we can :ship:

(crosses fingers with appveyor, its in a 'queue' atm)

pr is much more minimal now as well, fyi

eriktrom:
one question - when starting express on a port that is not available, when specifically passing the `-p` option should we bail and tell the user the port is in use(for clarity) or increment to an open port? what about when running ember serve with no options? (for example if you ember serve an addon and then your app)

we should probably throw when they provide a port that is not the defaultPort of 4200 and the port they give is taken. If they don't provide a port and 4200 is taken, increment express to the next open port

let me know, will update

rwjblue:
> when starting express on a port that is not available, when specifically passing the -p option should we bail and tell the user the port is in use(for clarity)

If the specified port is in use, we should bail and inform the user (roughly what we do today).

> what about when running ember serve with no options?

I think it would be awesome if we increment the port until one is found.  I personally get somewhat frustrated by the current workflow: 1) `ember s`; 2) Error port in use; 3) Try to guess an alternate port; 4) `ember s --port=<WAG>`; 5) DAMMIT, wrong guess, start over...

stefanpenner:
I believe we should maintain the current behavior, if people want automatic port choosing --port=0 can be used, and if they want that by default the system emberrc file can contain that.

eriktrom:
was wondering - k - should be ready now - just need upstream merge of portfinder to ship now i believe

stefanpenner:
Yay

I'm actually not totally aposed to what @rwjblue suggest, but I feel such a change is it-of-scope for a bugfix like this. If we want to make the change we should follow the rfc cycle.

rwjblue:
> I feel such a change is it-of-scope for a bugfix like this

Agreed.  I was mostly talking about ergonomics, not a bug.

> If we want to make the change we should follow the rfc cycle.

Makes total sense.

rwjblue:
@homu r+

eriktrom:
@homu - will fix that as well as bump upstream portfinder again very soon, thanks for the reminder

eriktrom:
Update here - there is [one remaining upstream pr](https://github.com/indexzero/node-portfinder/pull/25) that @winfinit is doing an awesome job working through. Shout out to him for the hard work he's put in finding the bug and then tackling it. 

I'm sure we'll be done soon, sorry for the delay, I left a pretty big bug in the initial release of port finder, which this pr depends on. Thanks again @winfinit.

stefanpenner:
@eriktrom thanks for pushing this forward :)

eriktrom:
I'll push an empty commit when portfinder is pushed to npm to make these tests pass, fyi

stefanpenner:
or just ping me here, i can restart the build then.

eriktrom:
something i can do from my phone any hour of the day, i'm all about that :)  thanks

stefanpenner:
> something i can do from my phone any hour of the day, i'm all about that :) thanks
>  Hide all checks

:+1:

rwjblue:
Looks like the portfinder update is done (in #5557), but we haven't landed the remaining changes here.  

Is there anything blocking a rebase + merge?

eriktrom:
@rwjblue - thanks for the ping
@stefanpenner thanks for noticing that portfinder got pushed to npm

i just rebased and pushed - when it turns green this is good to go, Sorry for the long delay(s) btw.

eriktrom:
sorry folks, looks like we got an error - the first rebase created a merge conflict - maybe thats it - at any rate i'll be opening the ember side of my project tonight/tomorrow(been a many weeks, had to build something else) - when I do I'll fix the test, npm link while devving and in general make sure this is bulletproof, fyi

rwjblue:
Sounds great, thanks for keeping on it...

eriktrom:
@anyone - there is a (which means 2) upstream bugs i was aware of but didn't have much real proof of upstream - https://github.com/indexzero/node-portfinder/issues/26 and https://github.com/indexzero/node-portfinder/issues/28 - we already merged porfinder 1.0.2 - which means its possible we may get a bug - if anyone has suggestions, https://github.com/indexzero/node-portfinder/issues/26 would be glad to have them, i don't really know the true answer, except to handle them as they come in, or the answer we have sort of come to in that PR, which is filter the big ones and handle someday the edge cases if they arise. (aka, on my machine i was originally only able to make up so many scenarios, anyone with a good operating system background (sysadmin) - we could use some <3)

eriktrom:
@homu - will do - we're waiting on another portfinder merge, https://github.com/indexzero/node-portfinder/pull/29, will push both together as a single commit, unless someone believes I should not wait. Should not be much longer, current version that was merged has a flaw. (which i was actually aware of)

(for context - IIRC - all this code does is make sure that both the server port and live reload port are open before starting, and ensures that when staring with port 0 they do not choose the same portfinder base port to start on) - actually finding a port for a host(which includes interfaces ipv4 and ipv6) is all upstream, so not a lot we're missing here (although will rebase and push w/ correct portfinder version and merge conflict changes when upstream is all clear, which has taken time and a few good men to get through, sorry 'bout that!)

eriktrom:
Merge me :)

Only issue is portfinder has not been pushed to npm, but we're using `^1.0.2` so when 1.0.3 is released the changes will come in on their own

(I merged the upstream commit for 1.0.3, but dont have npm push access - only reason not to merge would be that this is the only bridge to that future occurrence)

Coverage:
image

stefanpenner:
the window test failure seems related

cc @kellyselden

eriktrom:
Hmm - odd it fails on windows - upstream this test (implicitly) tests that it moves to the next port - https://github.com/indexzero/node-portfinder/blob/master/test/port-finder-test.js#L25-L42 - and passes on windows - maybe the -p is not working properly - I'll look into it but have a full day today - @kellyselden - if you try this manually and get the same results let me know - thanks man

eriktrom:
to make it easier to discuss, this is the test that fails:

```
  it('should throw error when -p PORT is taken', function() {
    function testServer(opts, test) {
      var server = require('http').createServer(function() {});
      return new Promise(function(resolve) {
        server.listen(opts.port, opts.host, function() {
          resolve(test(opts, server));
        });
      }).finally(function() {
        return new Promise(function(resolve) {
          server.close(function() { resolve(); });
        });
      });
    }

    return testServer({ port: '32773' }, function() {
      return command.validateAndRun([
        '--port', '32773'
      ])
      .then(function() {
        expect(true).to.equal(false, 'assertion should never run');
      })
      .catch(function(err) {
        var serveRun = tasks.Serve.prototype.run;
        expect(serveRun.called).to.equal(0, 'expected run to never be called');
        expect(err.message).to.contain('is already in use.');
      });
    });
  });
```

also i just pushed a commit that starts that server on port 32773 - and to clarify - the idea with that test is to start a normal server on port 32773, then run ember serve at 32773 and expect that b/c we passed a port explicitly, and it was taken, the express server should not try to start and give a nice error message. (https://github.com/eriktrom/ember-cli/blob/dev/lib/commands/serve.js#L74-L77)

the odd thing is the catch block is called, b/c in the error message `run was called once`, so the error was thrown, otherwise the .then block would have been called. If anyone has any ideas, drop them here.

also @kellyselden - if you run this manually, i should mention that when you pull run 
`npm i --save node-portfinder@github:indexzero/node-portfinder#master`  (just to clarify)

eriktrom:
I'm trying to see if this will pass CI with the upstream changes - i put this in package.json

`"portfinder": "https://github.com/indexzero/node-portfinder#2e22f95d831428b5cf6ce63e8681533a02daa034",`

and i get this only 0.10

`npm ERR! not a package /tmp/npm-3005--yYpL20v/github.com/indexzero/node-portfinder`

thoughts anyone?

stefanpenner:
kicking: https://travis-ci.org/ember-cli/ember-cli/jobs/116927892 also bumping some timeouts to mitigate intermittent failures.

stefanpenner:
also ensuring npm v3 is atleast installed https://github.com/ember-cli/ember-cli/pull/5646

stefanpenner:
and finally just going to disable node 0.12 on appveyor, testing it and test harness code intermittently failing is really not worth the cost to us: https://github.com/ember-cli/ember-cli/pull/5647

eriktrom:
@stefanpenner - do want to release with a sha or wait for npm - if we release with a sha i need to bump it one more commit(this one had error mis-spelled word, although not ever likely to get hit path) - anyway - around this afternoon - let me know if i should push one more time w/ the latest sha or push `^1.0.3` to this branch. It may be time before the next npm release, hard to know

eriktrom:
i'll push with 1.0.3 let me know if we should push latest sha instead

eriktrom:
pushed w/ non released ^1.0.3 - so travis will fail. If we dont get an update to npm soon, ping me and i'll freeze it at current master sha

(sorry, didn't need 3 comments for that)

stefanpenner:
We can't release with a sha, unfortunately many firewall users will be unable to checkout a sha :(

eriktrom:
gotcha - let me know if fork is needed (probably before next release b/c non ipv6 enabled computers have broken live reload port finding)

stefanpenner:
@eriktrom ya i think we a slowly running out of time, our users have been quite patient, but apparently all circle CI users are having issues. If you have time for a friendly interim fork, you should doit. So we can ship in the next day or so.

eriktrom:
yeah i saw some of the circle ci issues with something called mockgoose - anyway - yeah sure we can do that

just so i understand, b/c i have not forked then pushed to npm but on quick read the bottom of it the publish page in the docs says:

>Fails if the package name and version combination already exists in the specified registry.

Once a package is published with a given name and version, that specific name and version combination can never be used again, even if it is removed with npm-unpublish.

which i assume means i can push with version 1.0.3 from my own fork, is that correct? 

give me a pointer or two if mis-understanding

stefanpenner:
> which i assume means i can push with version 1.0.3 from my own fork, is that correct?

You will need to give it a new name, as only owners of modules can publish updates to one.

ember-cli-portfinder has been what I have called the friendly forks, once mainline catches up i typically deprecate the forks and sync back up to mainline.

eriktrom:
Awesome that's exactly what I needed

stefanpenner:
> Awesome that's exactly what I needed

Thanks again for all your help here, i know its been a slog.

eriktrom:
Not to worry, its actually been a good learning experience

eriktrom:
ready to :boat:  (after CI)

eriktrom:
clearly we have event emitter leak - maybe upstream library is using fs.createWriteStream and process.nextTick without waiting for finish event - anyway, node 5.8 fails b/c of test timeout likely due to  too much i/o - otherwise, we pass(or i guess we'll know when app veyor is done again)

eriktrom:
maybe cpr - https://github.com/davglass/cpr/blob/master/lib/index.js#L133-L193

(which in the test that fails, has many skipped tests nearby due to maybe a similar issue with comments like 'we dont know why this fails on travis')

stefanpenner:
does CPR have a throttle?

stefanpenner:
@eriktrom do you think its actually a bug in cpr, or merely needs some throttling?

eriktrom:
so i got the same bug recently on my node server - im not *totally* sure of course - but it was from opening write streams and not closing them(added Bunyan logger to my db, which i was already using in lots of beforeEach hooks, got same error about leaking emitters the day i added it - so i looked it up a bit

im sure you know these docs, but they helped me understand, so pasting for argument sake

>Event: 'finish'# 
When the stream.end() method has been called, and all data has been flushed to the underlying system, this event is emitted.

cpr never listens for `finish`, only `end` on the read side plus an added bonus

(from process.nextTick)
>Note: the nextTick queue is completely drained on each pass of the event loop before additional I/O is processed. As a result, recursively setting nextTick callbacks will block any I/O from happening, just like a while(true); loop.

cpr does it recursively using nextTick :)


so long story short - we use the `copyFixtureFile` method 58 times (w/ cpr) - it would not surprise me

also looks like `can-symlink` leaks as well, we should check what they're doing(dep for broccoli merge trees it seems)

cant say for sure though, ever in programming, till its been fixed :)

eriktrom:
im a bit weary about getting into another side project upstream ATM though - haha

stefanpenner:
> also looks like can-symlink leaks as well, we should check what they're doing(dep for broccoli merge trees it seems)

this leaks really?

> im a bit weary about getting into another side project upstream ATM though - haha
>  Hide all checks

a understood, share only what you have time for or interest in.

stefanpenner:
@homu r+

eriktrom:
@stefanpenner - i mean im no expert on event emitter leaks :)  I just read the stack trace in travis and made some guesses - `can-symlink` was in the stack trace - `copyFixtureFile` was an assumption

yeah probably wont jump into that right now, have some fun things I'm working on

PR #5395, opened 2016-01-26
Skip bower/npm install on blueprint install if manifests are missing

stefanpenner:
cc @trek


stefanpenner:
@homu r+

trek:
This is failing even worse than before :(

stefanpenner:
@trek its the same problem I believe, this also seems to just break `ember new foo`

nathanhammond:
@homu r+

nathanhammond:
@homu retry

PR #5410, opened 2016-02-01
Ensure that Testem can use `testem.js` by default.

rwjblue:
Support was added to allow folks to use `testem.js` instead of `testem.json` (in #4912), but that PR accidentally forgot to pass through the `--config-file` option to the created Testem instance.

Shortly after that #5089 landed and passed the `--config-file` option through to Testem.  Unfortunately, #5089 broke the ability for Testem to automatically choose its own config because the `--config-file` option was always defaulted to `./testem.json`.

The tests added in #4912 happened to continue to pass because the options supplied to Testem are actually enough to ensure that it runs the tests (but it was not actually invoking `testem.js` at all).

I tweaked the fixture and test to ensure that the file is actually ran, and removed the default (Testem internally will still default things properly).

/cc @cspanring @asakusuma

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -17,7 +17,7 @@ module.exports = Command.extend({
 
   availableOptions: [
     { name: 'environment', type: String,  default: 'test',          aliases: ['e'] },
-    { name: 'config-file', type: String,  default: './testem.json', aliases: ['c', 'cf'] },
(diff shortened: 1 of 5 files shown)


stefanpenner:
:+1:

acorncom:
Thank you, thank you, thank you! ;-)

PR #5421, opened 2016-02-03
Make options available in all Blueprint hooks as this.options

trentmwillis:
Currently, not all the available Blueprint hooks receive the `options` hash (e.g., `filesPath` and `normalizeEntityName`). This is a simple change to ensure that all hooks can access the options via `this.options`.

cc @rwjblue

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -157,6 +157,10 @@ module.exports = Blueprint;
       }
     },
 
+    filesPath: function(options) {
+      return path.join(this.path, 'files');
+    },
+
     beforeInstall: function(options) {},
     afterInstall: function(options) {},
     beforeUninstall: function(options) {},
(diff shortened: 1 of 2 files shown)


rwjblue:
I'm :+1: on passing in `options`.  Can you add a test that confirms `this.options` is present in `files` hook (or another hook if there are easier ones)?

trentmwillis:
Will add a test. Would you want me to pass `this.options` into `filesPath` as an argument (e.g., `this.filesPath(this.options)`)? Or should just having `this.options` be fine?

trentmwillis:
Added tests. I noticed that most of the blueprint hooks didn't really have any tests, so I just added some simple ones to make sure they get called with the proper arguments during install/uninstall.

Also documented `filesPath` as one of the available hooks and set it up so that it got `options` as an argument (similar to the other hooks).

rwjblue:
Awesome, thank you!

PR #5423, opened 2016-02-04
Initializer blueprint should reflex accurate params

xdumaine:
The commented parameters are not accurate. For developer clarity, show that the first param is registry, and the application is second.


stefanpenner:
@dgeb r?

rwjblue:
This was correct through 1.x and 2.0.0, but changed as of 2.1.0. An arity of 2 as of Ember 2.1.0 will trigger a deprecation.

Relevant code is [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/engine.js#L99-L108).

See http://emberjs.com/deprecations/v2.x/#toc_initializer-arity for deprecation details.

xdumaine:
Good to know. Thanks.

PR #5426, opened 2016-02-05
Update ember-try in addons to 0.1.2.

rwjblue:


blueprints/addon/index.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
     contents.devDependencies['ember-disable-prototype-extensions'] = '^1.1.0';
 
     // add `ember-try` to addons by default
-    contents.devDependencies['ember-try'] = '~0.0.8';
+    contents.devDependencies['ember-try'] = '^0.1.2';
```
kellyselden: Are you sure we want a `^`, since the jump from 0.0 to 0.1 looks like it was a breaking change? 0.1 to 0.2 could also be a breaking change.

rwjblue: ^ will not change the first non-zero number. 

^0.1.2 will allow everything below 0.2.0, ^1.0.0 will allow everything below 2.0.0

---

Also, 0.1.0 wasn't a breaking change, the older syntax for config/ember-try.js still works with 0.1.0 but using that syntax doesn't allow controlling npm deps.

rwjblue: See https://docs.npmjs.com/misc/semver#caret-ranges-1-2-3-0-2-5-0-0-4 for details on caret ranges.

kellyselden: > ^ will not change the first non-zero number. 

Ah I didn't know that zero bit. Thanks.

kellyselden:
@homu r+

PR #5430, opened 2016-02-06
move dev-only scripts to dev folder, so they aren't sent with the release

kellyselden:
the recent .npmignore change makes this easy to do. /bin is not ignored and /dev is.


rwjblue:
LGTM

kellyselden:
@homu r+

Issue #4918, opened 2015-10-02
In-repo-addon within an addon fails


aceofspades:
It may be useful to group a number of shared in-repo-addons within another wrapper addon. This seems to fail when building the app:

Sample: https://github.com/aceofspades/ember-cli-addon-with-ira

```
version: 1.13.8
Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/user-guide/#watchman for more info.
Cannot read property 'addonLintTree' of undefined
TypeError: Cannot read property 'addonLintTree' of undefined
    at Class.Addon.jshintAddonTree (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/models/addon.js:543:31)
    at Class.treeFor (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/models/addon.js:305:21)
    at /private/tmp/tmp/myapp/node_modules/ember-cli/lib/models/addon.js:253:32
    at Array.map (native)
    at Class.eachAddonInvoke (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/models/addon.js:251:22)
    at Class.treeFor (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/models/addon.js:298:20)
    at EmberApp.<anonymous> (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/broccoli/ember-app.js:358:20)
    at Array.map (native)
    at EmberApp.addonTreesFor (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/broccoli/ember-app.js:356:30)
    at EmberApp._processedAppTree (/private/tmp/tmp/myapp/node_modules/ember-cli/lib/broccoli/ember-app.js:616:25)
```

stefanpenner:
this is not currently supported.

runspired:
This error also occurs when including an addon in another addons npm `dependencies`, breaking addons that necessitate other addons

trentmwillis:
I've been running into this frequently while trying to utilize Engines. [A workaround exists by turning off `hintingEnabled`](https://github.com/dgeb/ember-engines/blob/master/tests/dummy/lib/ember-blog/index.js#L11-L13), but this is obviously less than ideal.

I'll take a look into it and see if I can find an actual solution.

ro0gr:
It's great to see it fixed!   
Just curious about is it supposed to be merged to 1.13?

Turbo87:
apparently this got fixed in #5432

PR #5432, opened 2016-02-06
Fix linting error in nested developing addons

trentmwillis:
Addresses #4918, the root cause is that `app` wasn't being set for nested addons, so then when it later tried to call `addonLintTree` it would fail. The solution proposed here is to pass `app` to all addons in the dependency tree after initializing the addons.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -637,9 +637,13 @@ Addon.prototype.jshintAddonTree = function() {
   }
 
   var addonJs = this.addonJsFiles(addonPath);
-  var jshintedAddon = this.app.addonLintTree('addon', addonJs);
+  var lintTrees = this.eachAddonInvoke('lintTree', [addonJs]);
+  var lintedAddon = mergeTrees(lintTrees, {
+    overwrite: true,
+    annotation: 'TreeMerger (addon-lint)'
+  });
 
-  return new Funnel(jshintedAddon, {
+  return new Funnel(lintedAddon, {
     srcDir: '/',
     destDir: this.name + '/tests/'
   });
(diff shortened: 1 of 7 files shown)


stefanpenner:
LGTM

@rwjblue double check?

rwjblue:
Discussed with @trentmwillis in slack, I believe we should use `this.eachAddonInvoke('lintTree', addonJs)` instead of `this.app.addonLintTree`.  The addon itself shouldn't require `this.app` to be present...

trentmwillis:
@rwjblue updated, should I do anything about the coveralls numbers?

kellyselden:
@trentmwillis nope, its close-to-zero approximation is messed up.

lib/models/addon.js
```diff
@@ -637,11 +637,11 @@ Addon.prototype.jshintAddonTree = function() {
   }
 
   var addonJs = this.addonJsFiles(addonPath);
-  var jshintedAddon = this.app.addonLintTree('addon', addonJs);
+  var lintTrees = this.eachAddonInvoke('lintTree', addonJs);
 
-  return new Funnel(jshintedAddon, {
```
rwjblue: Do we need this Funnel still?  Unless `lintTree` returns the files in `<current-addon-name>/tests` it seems that we do need it...

trentmwillis: Yup, we do need it. From the addon hooks docs:
> Return value is merged into the tests tree.

trentmwillis:
Okay, @rwjblue I think this should be good to go now.

rwjblue:
Thanks @trentmwillis!

PR #5433, opened 2016-02-06
Fix two spelling errors in ui.js

lukemelia:
- ember-cli-deploy intends to use ui.actualOutputStream (See https://github.com/ember-cli-deploy/ember-cli-deploy-plugin/blob/55b872daa0df949fa8d22adcf1b75e6b22968b87/index.js#L54-L66)

/cc @ghedamat


stefanpenner:
@homu r+

PR #5434, opened 2016-02-06
consistent newlines across all broccoli files

kellyselden:
just something I noticed. not all files were consistent.


stefanpenner:
@homu r+

PR #5438, opened 2016-02-07
add ember 1.13 to ember try

kellyselden:
I've seen when addons bump up to cli beta, they add a try for ember v1. This would make it default if we want to nudge addons to support 1.x.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -11,6 +11,7 @@ cache:
 
 env:
   - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=ember-1-13
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -8,6 +8,17 @@ module.exports = {
       }
     },
     {
+      name: 'ember-1-13',
+      bower: {
+        dependencies: {
+          'ember': '~1.13.0'
+        },
+        resolutions: {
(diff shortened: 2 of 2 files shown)


rwjblue:
@kellyselden - I've been thinking about this a bit (surprise), I think we should curate a common set of supported versions in a separate package, and allow folks to consume that in their `config/ember-try.js`.

rwjblue:
I'd like to make a small helper package that folks can use in their addons so that when I release new Ember's, I can update that package and they will automatically start testing against the additional Ember release...

 ```js
var buildTryConfig = require('ember-try-config-generator');

module.exports = buildConfig('1.13', '^2.0.0');
```

Turbo87:
@rwjblue I think it would be a good idea to merge this until `ember-try-config-generator` is actually a thing.

would it make sense to restrict the additional environment to ember-1.13?

@kellyselden I think this environment should probably also be added to the TravisCI manifest

kellyselden:
@Turbo87 Good catch. Updated the travis file. `^1.0.0` will resolve to the latest so it isn't really necessary to constrain it.

Turbo87:
@rwjblue @stefanpenner I'd like to merge this. Any objections?

rwjblue:
I'd prefer `~1.13.0` because it is clearer what you mean to support (saying that addons "support all of 1.x" seems scary ;P ).

Turbo87:
> I'd prefer ~1.13.0 because it is clearer what you mean to support

:+1:

Turbo87:
@kellyselden could you update the PR to use `~1.13.0` instead and change the scenario name to `ember-1.13`?

kellyselden:
@Turbo87 updated.

Turbo87:
awesome. @rwjblue are you okay with merging this now?

stefanpenner:
@rwjblue are you still good with supporting ~1.13 by default? I do believe (for now) it is a good citizen thing to do.

Issue #5436, opened 2016-02-07
Add path option to in-repo-addon blueprint


trentmwillis:
Currently, all in-repo-addons are generated into a `lib` directory. If you want to change the location you can do so manually, but with the coming of engines and specifically in-repo-engines, I wonder if it would be good to build in a `path` option. That way you can organize code in a more semantic way.

For example, maybe your app is broken up into different `areas`, so you want to generate engines at `areas/<name>`. Or maybe you have some build addons you want to live at `tools/<name>`.

As always, if this sounds good I'll take on implementing it.

rwjblue:
Sounds good to me.

Turbo87:
closing as related https://github.com/ember-cli/ember-cli/pull/5448 got closed too

PR #5448, opened 2016-02-10
Add support for in-repo-addon custom paths

trentmwillis:
Addressing #5436.

This works for initial generation but doesn't handle the use case for generating other blueprints in the in-repo-addon (there's a failing test for this feature). Not sure of the best way to handle that; it seems unwieldy to have to pass `--path` every time. Maybe we could infer it from the path in `package.json`? Would love your input @rwjblue @stefanpenner.


stefanpenner:
> Maybe we could infer it from the path in package.json? Would love your input @rwjblue @stefanpenner.

`.emberrc` might be appropriate place to provide a different default, maybe `in-repo-path` or something.

trentmwillis:
Would `in-repo-path` be used for all in-repo-addons?

My initial thought is to store the path on a per-addon basis:

```
in-repo-addons: [
  { name: 'my-addon', path: 'lib' },
  { name: 'my-engine', path: 'areas' }
]
```

stefanpenner:
that is a tricky question, I would hate to have to configure a trillion things here.

is the issue you want engines to be a special form of in-repo add-ons? (i would agree)

If that is the path we go down, I would be inclined to recommend toplevel `engines`...

trentmwillis:
> is the issue you want engines to be a special form of in-repo add-ons?

Sort of. I think forcing them into a specific path (e.g., `lib`) is unnecessarily limiting, but as you point out, also want to make sure we're not overdoing it on configuration/setup.

rwjblue:
I'm happy with allowing `--path` to be specified here, but I assume that `in-repo-engine` would customize and make a specific "stand" on a default value.

@trentmwillis - There are a few failing tests, unsure if related to these changes or just needs a rebase...

trentmwillis:
@rwjblue:

> This works for initial generation but doesn't handle the use case for generating other blueprints in the in-repo-addon (there's a failing test for this feature).

If we feel that `in-repo-engines` will also primarily stick to a default path, then I think I may drop this feature as the blueprint generation seems non-trivial to solve correctly.

stefanpenner:
> If we feel that in-repo-engines will also primarily stick to a default path, then I think I may drop this feature as the blueprint generation seems non-trivial to solve correctly.

we can reopen, if this does want to be addressed. Next time this may be good as an RFC (or one in addiition), so we can dig in etc.

PR #5450, opened 2016-02-10
skips version banner in --json mode

makepanic:
Currently ember-cli will always print the version banner (i.e. `version: 2.2.0-beta.6`). 
In combination with the `--json` flag that results in invalid json stdout.

I'm not sure if it's currently possible to add a test for this change because https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/help-test.js#L219 isn't including the previous ui stdout.

### DIFF
--- a/lib/commands/help.js
+++ b/lib/commands/help.js
@@ -6,6 +6,9 @@ var lookupCommand   = require('../cli/lookup-command');
 var string          = require('../utilities/string');
 var assign          = require('lodash/assign');
 var GenerateCommand = require('./generate');
+var versionUtils    = require('../utilities/version-utils');
+
+var emberCLIVersion      = versionUtils.emberCLIVersion;
 
 var RootCommand = Command.extend({
   isRoot: true,
@@ -43,6 +46,7 @@ module.exports = Command.extend({
     });
     if (commandOptions.json) {
       json = rootCommand.getJson(commandOptions);
+      json.version = emberCLIVersion();
       json.commands = [];
       json.addons = [];
     }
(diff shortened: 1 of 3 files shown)


stefanpenner:
needs a test

kellyselden:
Does the version get in the json via another method? If not, I would like it to move to the json instead of removed completely.

stefanpenner:
> I would like it to move to the json instead of remove completely.

:+1:

makepanic:
I added the ember-cli version to the root commands generated json and extended the `help-test` expected json output.

lib/commands/help.js
```diff
@@ -6,6 +6,9 @@ var lookupCommand   = require('../cli/lookup-command');
 var string          = require('../utilities/string');
 var assign          = require('lodash/assign');
 var GenerateCommand = require('./generate');
+var versionUtils   = require('../utilities/version-utils');
```
kellyselden: minor, equal alignment

tests/acceptance/help-test.js
```diff
@@ -13,6 +13,9 @@ var HelpCommand       = require('../../lib/commands/help');
 var requireAsHash     = require('../../lib/utilities/require-as-hash');
 var Command           = require('../../lib/models/command');
 var commands          = requireAsHash('../../lib/commands/*.js', Command);
+var versionUtils   = require('../../lib/utilities/version-utils');
```
kellyselden: equal alignment

kellyselden:
After the style fixes, I :+1:

makepanic:
realigned equals sign

kellyselden:
@homu r+

PR #5455, opened 2016-02-12
Add ESLint to lint tests

Turbo87:
In https://github.com/emberjs/core-notes/blob/master/ember-cli/2016-01/january-28.md#jshinteslintjscs it was mentioned that the project is planning to move from JSHint to ESLint. I've taken the liberty to convert most of the old JSHint to ESLint configuration files and I've added `mocha-eslint` to generate the necessary tests.

In the process of activating the individual ESLint rules I've also fixed a few resulting issues in separate commits. If this PR should not get accepted at least these commits should be cherrypicked to the master branch.


rwjblue:
These rules are all replacing rules in the .jshintrc, right? The .jshintrc should be removed as part of this PR. 

This seems great, thanks for taking the time to work on it!

Turbo87:
> These rules are all replacing rules in the .jshintrc, right?

yes, although I had to disable some of them for now because ESLint found much more issues than JSHint. I've annotated those options with a `disabled due to warnings` comment.

I'm not sure how these rules should be handled and would need to get some feedback on whether we should actually enforce those rules and fix the problems or keep them disabled.

rwjblue:
Can you create a new issue with a checklist of the options that are failing (which you disabled)? That way we don't loose track of deleting or fixing them.

Turbo87:
@rwjblue done: #5456

rwjblue:
Node 0.10 build should be fixed by recent #5461. I restarted the erroring build, but it should pass at this point.

I'm happy with this, and would like to land it. If anyone objects, speak now or suck it up :kissing_smiling_eyes: ...

rwjblue:
thanks @Turbo87!

Turbo87:
credits go to https://www.npmjs.com/package/polyjuice for helping me with the conversion btw :wink:

kellyselden:
@Turbo87 nice, gonna have to bookmark that.

PR #5458, opened 2016-02-12
app blueprint: Update jQuery version

xtian:
This is what Ember’s jQuery version requirements were updated to in emberjs/ember.js#12958 although I can update to just point at ^2.2.0 if so desired.


rwjblue:
I'd rather remove it altogether.  If you depend on `ember` in your `bower.json`, and it depends on`jquery` at whatever your `ember` version requires.

PR #5459, opened 2016-02-13
[BUGFIX] Remove npm version from `ember --version` output.

### LINKED ISSUE #5447: Fix `ember version`'s NPM version output.
Currently, ember-cli installs its own local copy of `npm` (so that we can use it from `ember install foo`).  Unfortunately, that makes the NPM output from `ember version` completely useless:

```
% ember version
version: 2.2.0-beta.3
node: 4.2.4
npm: 2.14.10
os: darwin x64
```

IMO, we should either "fix" the listed `npm` version number (so that it shows the system's default `npm version`), or remove it from the `ember version` command output.

chriskrycho:
Fixes #5447.

  - Add a utility to check the global npm version by spawning
    `npm --version` and retrieving its result.
  - Use the new utility to set a global npm version.
  - Print both local and global npm versions in the Version command.

### DIFF
--- a/lib/commands/version.js
+++ b/lib/commands/version.js
@@ -14,10 +14,9 @@ module.exports = Command.extend({
 
   run: function(options) {
     var versions = process.versions;
-    versions['npm'] = require('npm').version;
     versions['os'] = process.platform + ' ' + process.arch;
 
-    var alwaysPrint = ['node', 'npm', 'os'];
+    var alwaysPrint = ['node', 'os'];
 
     for (var module in versions) {
       if (options.verbose || alwaysPrint.indexOf(module) > -1) {
--- a/tests/unit/commands/version-test.js
+++ b/tests/unit/commands/version-test.js
@@ -24,7 +24,6 @@ describe('version command', function() {
     return command.validateAndRun().then(function() {
       var lines = options.ui.output.split(EOL);
       expect(someLineStartsWith(lines, 'node:'), 'contains the version of node');
-      expect(someLineStartsWith(lines, 'npm:'), 'contains the version of npm');
       expect(someLineStartsWith(lines, 'os:'), 'contains the version of os');
     });
   });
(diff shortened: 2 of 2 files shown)


chriskrycho:
I do have a couple questions on this:

1. I couldn't come up with a better way to get the global npm version than simply shelling out for it, and this does work, but if there's a better way, I'm all ears.
2. I couldn't come up with a good way to unit test the new utility, off the top of my head. I'm quite happy to add a test for it, given a suggestion of what that test should look like!

lib/utilities/global-npm-version.js
```diff
@@ -0,0 +1,9 @@
+// Get the global npm version. You cannot require global npm modules, including
+// global npm itself, but you can still *run* global npm.
+const spawnSync = require('child_process').spawnSync;
```
rwjblue: We can't use `const` or `spawnSync` on platforms older than 0.12 (we support 0.10 still).

stefanpenner: is there a point to telling us what the installed version of npm is? It has nothing to do with ember-cli. i would just update the output to say bundled-npm, if people want to now their version of npm they can run npm version

rwjblue: @stefanpenner - I am happy with that path also.  The original reason we added `npm` to the output of `ember version` was so that bug reports can include most of what we needed to know with simply sharing `ember version`.  This was of course before we started bundling our own, so the resulting value was a more accurate reflection of the system state.

stefanpenner: removing `npm` entirely from `ember version` actually seems like the best coarse of action, as we can tell which version of npm it bundles by looking at github.

rwjblue: :+1:

chriskrycho: Noted on all fronts (including `const` and `spawnSync`—I didn't realize Ember CLI supported Node 0.10!); I'll push up an alternate commit with the no-npm-version-at-all strategy and rename the PR accordingly in a little bit.

chriskrycho:
I've updated the pull request to simply remove the npm version entirely.

chriskrycho:
Curious: Travis failed during `npm install` on 0.10. Shouldn't be this change, given what it does.

Turbo87:
@chriskrycho the 0.10 build is currently failing due to the 0.10.41 -> 0.10.42 update. see #5461

chriskrycho:
@Turbo87 ah, gotcha. :+1:

rwjblue:
Thank you @chriskrycho!

chriskrycho:
My pleasure, @rwjblue. Hopefully a lot more (and more substantial) contributions ahead; [Olo] has started using Ember internally and we're actively interested in contributing back.

[Olo]: http://www.olo.com

rwjblue:
@chriskrycho - Awesome!

PR #5460, opened 2016-02-13
Increase the test timeout for proxy tests.

rwjblue:
These tests seem to consistently fail on node 0.10, this increases the timeout from 5s to 10s to give older node more time...

### DIFF
--- a/tests/unit/tasks/server/express-server-test.js
+++ b/tests/unit/tasks/server/express-server-test.js
@@ -248,6 +248,8 @@ describe('express-server', function() {
         });
     });
     describe('with proxy', function() {
+      this.timeout(10000);
+
       beforeEach(function() {
         return subject.start({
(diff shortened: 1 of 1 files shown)


rwjblue:
This does not fix the issues.  My guess is that this is related to supertest@1.2.0 ([diff 1.1.0...1.2.0](https://github.com/visionmedia/supertest/compare/v1.1.0...v1.2.0)) which was released 2 days ago (roughly when our v0.10 failures started)...

stefanpenner:
may as well just set that file wide

rwjblue:
Ugh, I can't seem to replicate the failure locally :(

rwjblue:
@stefanpenner - The higher timeout doesn't seem to help in Travis (though it does pass locally in v0.10.38 + npm 1.4.28).

PR #5461, opened 2016-02-13
Fix proxy tests

Turbo87:
This PR should fix the broken proxy test on node 0.10.42 and also fixes two other issues with that test suite.

see https://github.com/nodejitsu/node-http-proxy/issues/960 and https://github.com/visionmedia/superagent/issues/236 for details.


Turbo87:
looks like it actually works :smiley: 

(https://travis-ci.org/ember-cli/ember-cli/jobs/109072497#L2110-L2111)

PR #5470, opened 2016-02-15
Replace \n with native EOL in processed templates

Turbo87:
This PR fixes most of the line ending issues introduced by the `.gitattributes` file. Once the blueprint template is processed it will (if necessary) replace any `\n` characters with native EOLs.

This should fix some of the AppVeyor issues. Some issues remain, because the fixture files that are used for `assertFileEquals()` obviously use `\n` but the generated files on Windows will now have `\r\n` again. `assertFileEquals()` will most likely need to be modified to support a line ending option for the fixture file.

**Update:** Fixture files are read with native EOLs now too

### DIFF
--- a/tests/helpers/assert-file-equals.js
+++ b/tests/helpers/assert-file-equals.js
@@ -2,6 +2,7 @@
 
 var expect = require('chai').expect;
 var fs     = require('fs');
+var EOL    = require('os').EOL;
 
 /*
   Assert that a given file matches another.
@@ -13,6 +14,10 @@ var fs     = require('fs');
 module.exports = function assertFileEquals(pathToActual, pathToExpected) {
   var actual = fs.readFileSync(pathToActual, { encoding: 'utf-8' });
   var expected = fs.readFileSync(pathToExpected, { encoding: 'utf-8' });
+
+  if (EOL !== '\n') {
+    expected = expected.replace(/\n/g, '\r\n');
+  }
 
   expect(actual).to.equal(expected);
 };
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -97,7 +97,11 @@ FileInfo.prototype.render = function() {
           return content;
         } else {
           try {
-            return processTemplate(content.toString(), context);
+            var result = processTemplate(content.toString(), context);
+            if (EOL !== '\n') {
+              result = result.replace(/\n/g, EOL);
+            }
+            return result;
           } catch (err) {
             err.message += ' (Error in blueprint template: ' + path + ')';
             throw err;


rwjblue:
Seems like a good path to me. 

We should ensure that the blueprints in the repo use the correct line endings (enforced by tests), in case one is PR'ed from a Windows machine...

rwjblue:
@kellyselden / @raytiley - What do y'all think?

Turbo87:
looks like there are only two issues left:

- the "brocfile-smoke-test" tests (looks like those are related to line ending issues too)
- the "correctly identifies the root of the project" tests (unrelated?)

stefanpenner:
I'm unsure basically many windows developers develop with posix line endings. If anything it could attempt to follow editor config like settings.

Silly line endings

Turbo87:
@stefanpenner correct me if I'm wrong, but this PR is essentially just keeping the behavior the same as it was before the `.gitattributes` change.

raytiley:
This is brining up old wounds :)

The approach seems fine to fix the problem brought on by enforcing `\n` via the `.gitattributes` changes, but it is annoying how this issue leaks through out the test suite. There are multiple places this needs to be changed (2 in the current PR) and obviously some more needed since the tests are still failing.

@stefanpenner brings up a good point that there is still inconsistencies. Default ember projects include and `.editorconfig` that sets line endings `\n`. So in one place we are saying the project should use `\n` but then in the generators using `\r\n` on windows with no way to change.

Let me know if I can help... I'll dust off my `ember-cli-windows-dev` vm so I can run the tests locally.

tests/helpers/assert-file-equals.js
```diff
@@ -14,5 +15,9 @@ module.exports = function assertFileEquals(pathToActual, pathToExpected) {
   var actual = fs.readFileSync(pathToActual, { encoding: 'utf-8' });
   var expected = fs.readFileSync(pathToExpected, { encoding: 'utf-8' });
 
+  if (EOL !== '\n') {
+    expected = expected.replace(/\n/g, '\r\n');
```
raytiley: this should probably be `.replace(/\n/g, EOL);` for consistency.

Turbo87: right, it should...

Turbo87:
@raytiley the editorconfig file is actually a good point. I'm totally okay with going `\n` everywhere, I just didn't know what the desired behavior was. If we use `\n` everywhere we basically only have to replace the native EOLs in the testsuite with `\n` where appropriate.

Turbo87:
I've opened #5477 instead now, that changes the tests to expect `\n` everywhere.

PR #5472, opened 2016-02-15
ESLint: Use recommended rules preset

Turbo87:
This PR adds the `eslint:recommended` rules preset to `.eslintrc.js` and fixes the few issues that came up.

- `no-console` is disabled because we obviously do need to use the console in a CLI.
- `comma-dangle` is disabled because the code in the project is inconsistent about this and I wasn't sure what the desired style is and whether we want to enforce it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  extends: 'eslint:recommended',
   env: {
     browser: false,
     node: true,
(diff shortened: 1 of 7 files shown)


Turbo87:
@rwjblue @stefanpenner @kellyselden any thoughts on this?

the "recommended" rules are the ones with a checkmark at http://eslint.org/docs/rules/

rwjblue:
@homu r+

Turbo87:
hmm, I don't think the test failure is related to my changes. the original Travis PR test seems to have passed.

rwjblue:
@homu retry

Turbo87:
oh well... looks like we have a few more flaky tests to fix...

PR #5476, opened 2016-02-16
TravisCI: Adjust command order in manifest

Turbo87:
This is functionally identical but easier to read because the manifest order matches the execution order now.


rwjblue:
@homu r+

PR #5477, opened 2016-02-16
Fix test suite on Windows

Turbo87:
This PR fixes most of the currently failing tests on Windows/AppVeyor. It sort of obsoletes #5470.


Turbo87:
```
 Downloading PhantomJS 32 bit
   from 'https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-windows.zip'
 Exception calling "GetResponse" with "0" argument(s): "The remote server return
 ed an error: (429) TOO MANY REQUESTS."
```

:bomb:

Turbo87:
well, at least the node 5 build looks a lot better now :smiley:

rwjblue:
Should this also delete the `.gitattributes` file?

rwjblue:
Will need @kellyselden and/or @raytiley to review I think.  I'm definitely not a qualified Windows developer any more....

trek:
Qualified© Windows® Developer

kellyselden:
@rwjblue The motivation for the `.gitattributes` change was doing releases. Making sure the files on disk on a Windows machine were \n before sending to npm, no matter what the global git settings were on the machine. So even though adding the file got us into this mess, I think removing it would bring us back to square one.

rwjblue:
Gotcha, makes sense

Turbo87:
I agree with what @kellyselden said. I originally misunderstood the situation before the `.gitattributes` file was introduced and thought Windows users had always gotten `\r\n` files. But since the files published to NPM were always `\n` our test suite actually didn't test the reality. Even worse, if any Windows devs would clone the project without using auto CRLF conversion the test suite would have failed for them even before `.gitattributes` was introduced.

tl;dr having `.gitattributes` force consistent line endings on all platforms is a good thing and our tests should reflect that.

feel free to `r+` this PR. I'm not sure about the other failing tests, but I suspect that raising the timeouts on them might work. Since AppVeyor is actually really slow we could also skip the slow tests there and run them only on TravisCI. thoughts on that?

rwjblue:
Does the .gitattributes file ensure that all newly added files have \n line endings? Do we need to do anything to ensure that?

rwjblue:
@homu r+

Turbo87:
@rwjblue I thought it did, but when I read about it yesterday it seemed to only concern `git checkout` but not `git commit`. I'm not sure if we need active protection against `\r\n` files, but if we do ESLint got us covered (http://eslint.org/docs/rules/linebreak-style.html)

Turbo87:
@homu test failure looks like a flaky test again... :-/

rwjblue:
Sounds good. Can you enable that rule, and rebase? Then we'll get this landed...

Turbo87:
@rwjblue it's actually enabled already (https://github.com/ember-cli/ember-cli/blob/master/.eslintrc.js#L56-L57) :joy:

rwjblue:
HA! Awesome...

rwjblue:
ZOMG!  AppVeyor is green!

kellyselden:
nice work @Turbo87 !

Issue #5487, opened 2016-02-19
Test Coverage Mega Thread


kellyselden:
There are some areas that are missing test coverage. This will track any we find. This is a perfect place for new contributors to start understanding the code. Link pull requests next to its check mark.
- [ ] We could use an "acceptance/addon-test.js" file, similar to ["acceptance/new-test.js"](https://github.com/ember-cli/ember-cli/blob/master/tests%2Facceptance%2Fnew-test.js)
- [x] #5488 We need a test for 'doesn\'t allow to create an addon named `Ember`' (borrow from [here](https://github.com/ember-cli/ember-cli/blob/master/tests%2Funit%2Fcommands%2Fnew-test.js#L56))
- [x] #5488 We need a test for 'doesn\'t allow to create an addon named `ember-cli`' (borrow from [here](https://github.com/ember-cli/ember-cli/blob/master/tests%2Funit%2Fcommands%2Fnew-test.js#L65))
- [x] #5490 We need a test for 'shows a suggestion messages when the addon name is a period' (borrow from [here](https://github.com/ember-cli/ember-cli/blob/master/tests%2Funit%2Fcommands%2Fnew-test.js#L101)) (also, message needs to change to "addon")
- [x] #5873 We need a test for 'registers blueprint options in beforeRun' for addons (borrow from [here](https://github.com/ember-cli/ember-cli/blob/master/tests%2Funit%2Fcommands%2Fnew-test.js#L110))
- [ ] We need a test for 'passes command options through to init command' for addons (borrow from [here](https://github.com/ember-cli/ember-cli/blob/master/tests%2Funit%2Fcommands%2Fnew-test.js#L124))
- More to come... If you find any untested features, feel free to add to this list!

Turbo87:
make sure to check #5483 before working on any of those issues and just copy pasting other text cases!

courajs:
I don't really understand either of
> - [ ] We need a test for 'registers blueprint options in beforeRun' (borrow from here)
- [ ] We need a test for 'passes command options through to init command' (borrow from here)

It looks like you're suggesting to add tests that already exist?

robbiepitts:
Added a PR (#5873) for:

> - [ ] We need a test for 'registers blueprint options in beforeRun' (borrow from here)

kellyselden:
@courajs I updated the text. I forgot the addon references.

Turbo87:
@kellyselden is this still up-to-date?

locks:
[...]

PR #5483, opened 2016-02-17
ESLint: Use "chai-expect" plugin

Turbo87:
This PR adds the "chai-expect" plugin to ESLint that checks for `expect(...)` usage without any assertions. All resulting lint issues have be resolved, but let's see if Travis finds any tests that are broken now...

PR #5488, opened 2016-02-19
Add a test to ensure addons can't be named Ember or ember-cli

johnotander:
Related #5487

### DIFF
--- a/tests/unit/commands/addon-test.js
+++ b/tests/unit/commands/addon-test.js
@@ -37,6 +37,15 @@ describe('addon command', function() {
     });
   });
 
+  it('doesn\'t allow to create an addon named `Ember`', function() {
+    return command.validateAndRun(['Ember']).then(function() {
+      expect(false, 'should have rejected with an addon name of test');
+    })
+    .catch(function(error) {
+      expect(error.message).to.equal('We currently do not support a name of `Ember`.');
+    });
+  });
+
   it('doesn\'t allow to create an addon named `vendor`', function() {
     return command.validateAndRun(['vendor']).then(function() {
       expect(false, 'should have rejected with an addon name of `vendor`');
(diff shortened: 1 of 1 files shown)


tests/unit/commands/addon-test.js
```diff
@@ -30,13 +30,31 @@ describe('addon command', function() {
...
+      expect(false, 'should have rejected with an addon name of ember');
     })
     .catch(function(error) {
       expect(error.message).to.equal('We currently do not support a name of `ember`.');
     });
   });
 
+  it('doesn\'t allow to create an addon named `Ember`', function() {
+    return command.validateAndRun(['Ember']).then(function() {
+      expect(false, 'should have rejected with an addon name of Ember');
```
Turbo87: `expect(false)` is not actually asserting anything, you need to use something like `.to.be.true` at the end. see #5483

johnotander: Ah, thanks, that makes sense. I've gone ahead and fixed those.

After a quick grep I noticed that this `expect(false, 'some string');` pattern is repeated throughout the tests. So, I can go ahead and update those, too. I'm assuming I should make that a separate PR?

Turbo87: @johnotander I've updated those in #5483 already, but thanks for the offer!

johnotander: Whoops, missed that part :blush: . Awesome!

Turbo87:
r+ from my side. @kellyselden r?

kellyselden:
@Turbo87 other than you needing a rebase on those other test changes, LGTM.

kellyselden:
@homu r+

kellyselden:
Thank you @johnotander

PR #5491, opened 2016-02-19
Change "ember i" alias from "ember init" to "ember install"

### LINKED ISSUE #5489: `ember i` should alias `ember install`
Currently `ember i` aliases `ember init` for reasons of timeline and/or alphabet. We need to make it such that `ember i` aliases the intended (and much more frequently used) `ember install`.

Turbo87:
This PR resolves #5489 as discussed in the team meeting yesterday.


rwjblue:
@homu r+

Issue #5348, opened 2016-01-13
Apply options before destroyApp in Module-for-acceptance.js afterEach


kgish:
It seems to make more sense to me that the `destroyApp()` function be called AFTER the calls to `options.afterEach.apply(...)` as these may depend on the existence of `this.application` in order to work properly.

I understand that the `Module-for-acceptance.js` file is more of a template that users are free to modify, but shouldn't the above be the default?

See: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/tests/helpers/module-for-acceptance.js#L16

rwjblue:
Yes, teardown/afterEach should be the reverse of setup/beforeEach.  So the supplied `options.afterEach` should be called before `destroyApp` is called.

PR #5497, opened 2016-02-20
Call destroyApp after handling afterEach options

johnotander:
As discussed in #5348, `destroyApp(...)` should be
called after `options.afterEach.apply(...)`.


rwjblue:
Thank you!!

Issue #5308, opened 2016-01-06
implement rfc#28


stefanpenner:
https://github.com/ember-cli/rfcs/pull/28 is ready to be implemented.

ro0gr:
I've started playing with output-file option https://github.com/ro0gr/ember-cli/commits/import-output-file.

PR #5499, opened 2016-02-20
output-file import

ro0gr:
Resolves https://github.com/ember-cli/ember-cli/issues/5308

Here we have an ability to import file to a custom output file:
```javascript
app.import('vendor/important-file.js', {outputFile: '/assets/very-important.js'})
```

It also works for css files:
```javascript
app.import('vendor/important-file.css', {outputFile: '/assets/very-important.css'})
```

The side effect of the current implementation is that `prepend` option is now applies for css files(not sure if it's ok).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1097,6 +1097,10 @@ EmberApp.prototype.javascript = function() {
   var appOutputPath       = this.options.outputPaths.app.js;
   var appJs = applicationJs;
 
+  this.import('vendor/ember-cli/vendor-prefix.js', {prepend: true});
+  this.import('vendor/addons.js');
+  this.import('vendor/ember-cli/vendor-suffix.js');
+
   // Note: If ember-cli-babel is installed we have already performed the transpilation at this point
   if (!this._addonInstalled('ember-cli-babel')) {
     appJs = new Babel(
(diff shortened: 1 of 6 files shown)


ro0gr:
There is a strange coverage results. 
Code coverage increased for broccoli/ember-app but the total score decreased by 0.1% somehow. 
I'm not sure what should be done in the scope of this pr to fix it.

rwjblue:
Coveralls is just a guide for us to generally know if a PR significantly changes coverage. 

I'll try to review this in more detail this morning...

lib/broccoli/ember-app.js
```diff
@@ -93,8 +93,9 @@ function EmberApp() {
   this._initOptions(options, isProduction);
   this._initVendorFiles();
 
-  this.legacyFilesToAppend     = [];
-  this.vendorStaticStyles      = [];
+  this.legacyFilesToAppend = {};
```
rwjblue: Odd spacing here...

rwjblue:
RE: prepend for CSS files, I think that is OK in general but I'm not sure how used it would be...

rwjblue:
These changes generally look good, but we need to add some smoke tests to confirm that using the outputFile option actually results in the files being emitted in the build.

ro0gr:
> but we need to add some smoke tests
Agreed. Looks like https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/brocfile-smoke-test-slow.js is the best place to add it. I'll dig in this direction.

Turbo87:
@ro0gr we are hoping to reduce the amount of end-to-end tests in the future because they take a long time to run. it would be awesome if there was a faster way to test this.

ro0gr:
@rwjblue added end-to-end test for output file

ro0gr:
@Turbo87 yeah, it definitely would be great to have fast feedback from the test suite.
Working on this pr I've found useful to comment out tearing down of test targets in the test file I'm working on and disabling https://github.com/ember-cli/ember-cli/blob/master/tests/runner.js#L13-L14
Yes, it's dirty and unreliable but it sped up my test runs significantly :p

rwjblue:
I don't see the new fixture files in the diff (maybe I missed them?)...

ro0gr:
Ouch.. Sorry, few mins

ro0gr:
A little bit unrelated but looks like there are bunch of forever-green-tests in the brocfile-smoke-test-slow like:
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/brocfile-smoke-test-slow.js#L316
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/brocfile-smoke-test-slow.js#L334
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/brocfile-smoke-test-slow.js#L371
and others..
I've just copy-pasted skeleton from one of them and mentioned that my test passes with a wrong condition. I may miss something(not very familar with chai/mocha) but looks like these tests are wrong.

Turbo87:
@ro0gr good catch. this is actually being addressed already in #5483. unfortunately fixing those broken assertions revealed a few broken tests that need to be fixed before that PR can be merged.

ro0gr:
Hope they will be fixed soon. Evergreen slow tests is kind of a very specific pleasure :laughing:

Turbo87:
@ro0gr there is only one failing test left to fix, unfortunately it is exactly one of those brocfile-smoke-test-slow tests... :wink:

ro0gr:
Have made a rebase.

rwjblue:
Awesome, thank you for working on this!!!

Issue #4977, opened 2015-10-22
Come up with a nice "Welcome to Ember" design as part of initial setup


acorncom:
I've seen various references to a redesign coming for the main Ember site (done by thoughtbots), do we have a similar design for the "Welcome to Ember.js" page that is shown after initial setup?

Just a simple touch like that could make the initial startup with Ember feel much more polished.

See https://github.com/emberjs/website/pull/2393#issuecomment-150071985 for a glimpse of the upcoming design ...

michaelrkn:
I think it would be helpful to link to the Guides, something like:

```html
<h2>Welcome to Ember.js!</h2>
<p>If you're not sure where to start, take a look at the <a href="http://guides.emberjs.com/" target="_blank">Ember Guides</a>.</p>
```

joseperales:
@acorncom  @michaelrkn Cool !! so this issue should be in https://github.com/emberjs/ember.js

michaelrkn:
@joseperales why do you say that?

joseperales:
Because the index file is in emberjs not in ember-cli

acorncom:
@joseperales The index file I was thinking we could update is this one here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/index.html

H3Chief:
I think it would actually be cool if the default page displayed to total output of `ember --help`

acorncom:
Recommendation from @wifelette 

> How about… same background color as the redesign, `#FFFBF5`, and we just stick this dead center:
![Construction tomster](http://emberjs.com/images/tomsters/construction.png)

PR #5501, opened 2016-02-22
Add welcome script to improve 'getting started' experience

acorncom:
This addresses #4977 and needs to be merged once https://github.com/emberjs/website/pull/2469 has been deployed to production.

Background:
The 'getting started' experience for newcomers to Ember has left something to be desired when first firing up Ember-CLI.  This aims to help address that while also aiming to make it as easy as possible to continue doing upgrades without interference for existing users.  We're hoping to use an external script to allow editing / improving things without affecting blueprints within Ember-CLI.

More background information available at: https://github.com/emberjs/website/pull/2469

### DIFF
--- a/blueprints/app/files/app/templates/application.hbs
+++ b/blueprints/app/files/app/templates/application.hbs
@@ -1,3 +1,5 @@
 <h2 id="title">Welcome to Ember</h2>
 
+<script src="http://emberjs.com/javascripts/ember-cli-welcome.js></script>
+
 {{outlet}}


stefanpenner:
do we want to encourage users put script tags in hbs files?

mmun:
I agree with improving the splash screen, but we can't use `<script>` tags.

I like Rail's approach. Rails ships with a [welcome page](https://github.com/rails/rails/blob/master/railties/lib/rails/templates/rails/welcome/index.html.erb) as part of the framework and it gets used until you override it. This would let us remove the application.hbs from the blueprint as well, which would help make `ember init` less weird when it tries to reset your application.hbs.

The message at the bottom could be replaced with something like "use `ember generate route index` to get started".

acorncom:
> I agree with improving the splash screen, but we can't use `<script>` tags.

Regarding the above, is it philosophically can't or can't, it won't work?  Several other folks on the `#documentation` channel were saying script tags wouldn't work as well, but I've got a prototype here that shows it working (https://github.com/acorncom/ember-welcome-prototype) albeit with a slightly older version of Ember.

We'd mainly been trying to use a script tag to make this easier to update without churn on the Ember-CLI side and also to keep the `ember init` diffs to a minimum when folks upgrade.  Not emotionally invested in this way of implementing it though, so let me know how you want to do it and I can adjust the PR

mmun:
It throws an error on more recent versions of Ember. It was explicitly stated by HTMLBar's parser that [script tags are not supported](https://github.com/tildeio/simple-html-tokenizer#simple-html-tokenizer-). In theory, we can add script tag support but it will require going through the RFC process.

Regardless, putting script tags in your template is not something I want to promote. I would prefer the implementation I described, or in the meantime, just put the content directly in application.hbs in ember-cli.

acorncom:
> Regardless, putting script tags in your template is not something I want to promote

Sounds good.  I was actually surprised it worked ;-)  

Re: implementation details, I'm unclear on how this welcome setup would lead to the removal of `application.hbs`.  Would we have some logic checking to see if no routes had been setup and displaying this as a fallback?

Re: `<iframe>`, the design looks best (with background color) if we can style the `<body`> tag so I'd prefer to skip that if possible.  Happy to put content directly in the `application.hbs` file if that's not a deal-breaker for you guys?

rwjblue:
I'd like to see a different concept implemented and would love to help that come to fruition if you have time.  Basically, I'd like to see a ember-getting-started addon (added by default) that adds a separate Ember app mounted at `/ember-getting-started` (or some other non-conflicting path) and takes over `/` if no routes are present in `Router.map` or on disk. This would:

1. enable usage of a nice interactive getting started app when no routes/templates are present
2. enable usage of interactive guide when navigating to a specific URL (which could print to the console on initial `ember serve` or something)
3. remove extra boilerplate that is generated into ember-cli apps (the current presence of `app/templates/application.hbs` is very annoying to me every time I run `ember init`)

@acorncom - Lets chat about this in Slack or the next Docs call, I'd love to help guide someone on this...

Turbo87:
closing this because the `<script>` tag approach was rejected. 

I like the `ember-getting-started` approach. I guess it would already be enough to just include an `application.hbs` in that addon, right?

PR #5507, opened 2016-02-23
[BUGFIX beta] Replace testem.json with testem.js

### LINKED ISSUE #5503: Migrate from testem.json to testem.js?
Hi. 
I've found myself annoyed, as most people, by the fact that json files cannot have comments, trailing commas and other things that make them less usable that js files, and also you can't access environment variables in json files.

I recently discovered that you can config testem by exporting the same json object. Would you be opened to replace `testem.json` by `testem.js`?
The only difference would be a `module.exports = ` at the beginning of the file (and a jshint comment to mark the file as for node).

Yay comments.

cibernox:
Replace testem.json with testem.js.

JSON syntax is quite limited (no comments, no trailing commas, no single quotes).
Javascript files don't have this limitations and besides give the users access to
more features, like environment variables.

No idea if bugfix beta is the right tag.

Closes #5503

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -1,4 +1,5 @@
-{
+/*jshint node:true*/
+module.exports = {
   "framework": "qunit",
   "test_page": "tests/index.html?hidepassed",
   "disable_watching": true,
@@ -9,4 +10,4 @@
     "PhantomJS",
     "Chrome"
   ]
-}
+};
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -213,7 +213,7 @@ describe('Acceptance: addon-smoke-test', function() {
           '.gitkeep',
           '.travis.yml',
           '.editorconfig',
-          'testem.json',
+          'testem.js',
           '.ember-cli',
           'bower.json',
           '.bowerrc'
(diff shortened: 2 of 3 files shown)


blueprints/app/files/testem.js
```diff
@@ -1,4 +1,6 @@
-{
+/*jshint node:true*/
+/* global module */
```
rwjblue: `node:true` should prevent the `global module` from being needed, right?

rwjblue:
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/smoke-test-slow.js#L106 is causing failing tests.

Turbo87:
:+1: from my side. I prefer being able to use trailing commas and single quotes although I've probably never touched that particular file so far.

cibernox:
Ready. I'm pretty sure the node 0.12 failure in windows is entirely random.

Turbo87:
@homu r+

Turbo87:
looks like @homu doesn't like me yet...

stefanpenner:
Im assuming we still support both, just blueprint the js one by default?

cibernox:
yes, just the blueprint changed

Turbo87:
@cibernox could you possibly add a test to verify that json files are still supported as well?

cibernox:
Not sure how to verify that, because I'd have to run `ember test` to check that testem picks it

stefanpenner:
@cibernox actually ember-cli loads the file, and passes it to testem.

rwjblue:
@stefanpenner - Confirm, this was relatively recently changed (and fixed up a bit more)...

PR #5508, opened 2016-02-23
Expose a project.generateTestFile() method to be overwritten by test framework addons

Turbo87:
This PR adds a `generateTestFile()` method to the `Project` model that is supposed to be overwritten by test framework addons like `ember-cli-qunit` and `ember-cli-mocha`. This function can then be used by linter addons like `ember-cli-eslint` or `broccoli-jscs` to generate tests files without knowledge of the actual test framework being used.

/cc @rwjblue

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -530,6 +530,30 @@ Project.prototype.findAddonByName = function(name) {
 };
 
 /**
+  Generate test file contents.
+
+  This method is supposed to be overwritten by test framework addons
+  like `ember-cli-qunit` and `ember-cli-mocha`.
+
+  @public
+  @method generateTestFile
+  @param {String} moduleName Name of the test module (e.g. `JSHint`)
+  @param {Object[]} tests Array of tests with `name`, `passed` and `errorMessage` properties
+  @return {String} The test file content
+ */
+Project.prototype.generateTestFile = function(/* moduleName, tests */) {
+  var message = 'Please install an Ember.js test framework addon or update your dependencies.';
+
+  if (this.ui) {
+    this.ui.writeDeprecateLine(message)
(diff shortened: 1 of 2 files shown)


kellyselden:
How would broccoli-jscs generate tests without framework knowledge? Would it be supplying just a fragment that returns only true or false?

Turbo87:
@kellyselden like this: https://github.com/kellyselden/broccoli-jscs/compare/master...Turbo87:cli-test-generator

kellyselden:
@Turbo87 That's pretty cool!

lib/models/project.js
```diff
@@ -530,6 +530,32 @@ Project.prototype.findAddonByName = function(name) {
 };
 
 /**
+  Generate test file contents.
+
+  This method is supposed to be overwritten by test framework addons
+  like `ember-cli-qunit` and `ember-cli-mocha`.
+
+  @protected
```
Turbo87: just noticed that `protected` is obviously bullshit here because other addons would need to use this so it should be a `public` API

stefanpenner: :+1:

stefanpenner: ya, its basically public, and intended to be optionally overiden by a subclass, so public makes sense.

Turbo87: rebased and fixed

stefanpenner:
:+1:

Turbo87:
I guess this PR is somewhat related to the https://github.com/ember-cli/rfcs/pull/15 RFC

jonathanKingston:
:+1: thanks, feel free to close the RFC when this lands as it solves the same issue

dfreeman:
:heart:

PR #5512, opened 2016-02-24
Anonymous AMD Support

ef4:
This adds an option named `amdModule` to `app.import`. It causes any anonymous AMD module in the imported file to be registered under the given name.

    app.import('bower_components/some-dep/some-dep.js', { amdModule: 'some-dep' });

This should work correctly with a large number of libraries that:

 - check for `typeof define === 'function' && define.amd`
 - call `define` with just a factory function, or a dependency list and factory function.

Question: is this use of broccoli-stew's `map` sufficiently performant?

### DIFF
--- a/lib/broccoli/amd-shim.js
+++ b/lib/broccoli/amd-shim.js
@@ -1,0 +1,19 @@
+'use strict';
+var stew = require('broccoli-stew');
+
+module.exports = function shimAmd(tree, nameMapping) {
+  return stew.map(tree, function(content, relativePath) {
+    var name = nameMapping[relativePath];
+    if (name) {
+      return [
+        '(function(define){',
+        content,
+        '})((function(){ function newDefine(){ var args = Array.prototype.slice.call(arguments); args.unshift("',
+        name,
+        '"); return define.apply(null, args); }; newDefine.amd = true; return newDefine; })());'
+      ].join('');
+    } else {
+      return content;
+    }
+  });
+};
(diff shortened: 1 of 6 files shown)


stefanpenner:
look like a good quick solution.

> Question: is this use of broccoli-stew's map sufficiently performant?

Map, is more or less the speed of a persistent-filter so it should be sufficiently performant. One thing worth noting, is that it isn't sourcemap aware but I'm honestly fine with supporting this sans source-map.

stefanpenner:
@ef4 as this does not work with source-maps, should it automatically strip source-map urls and warn or debug log?

ef4:
I suspect this will actually just disable any source map that was present in the file you're shimming, since the comment will no longer be at the end of the file. So you won't end up with a broken source map, you'll just get no sourcemapping inside this particular file.

But I will investigate -- it's probably not too hard to just preserve the sourcemap.

stefanpenner:
> But I will investigate

Ya, if you could double check that would be great (if its good, we can merge this)

> it's probably not too hard to just preserve the source-map.

Ya, I don't believe it would be (but its fine to do this early)

ef4:
I confirmed that this does not break sourcemaps, it just causes any preexisting sourcemap in the AMD module to be ignored, such that your resolution will stop at the level of the imported file and not go any deeper. 

So this is ready to merge.

stefanpenner:
@homu r+

stefanpenner:
thanks

stefanpenner:
@homu retry

stefanpenner:
@homu retry

PR #5513, opened 2016-02-24
[BUGFIX beta] Replace testem.json by testem.js in npmignore too

cibernox:
Following with #5507, I forgot to update the `.npmignore` blueprint


Turbo87:
@homu r+

PR #5520, opened 2016-02-26
[WIP] Propagates testem errors

chancancode:
Couldn't get the test to pass locally for some reason ¯\_(ツ)_/¯

TODOS:
- [ ] Make the test work
- [ ] Properly handle the testem timeout

```
Error: Timed out after 540s
    at null._onTimeout (/Users/godfrey/Code/ember.js/node_modules/testem/lib/app.js:309:21)
    at Timer.listOnTimeout (timers.js:89:15)
```

### DIFF
--- a/lib/tasks/test-server.js
+++ b/lib/tasks/test-server.js
@@ -12,12 +12,12 @@ module.exports = TestTask.extend({
 
   invokeTestem: function(options) {
     var task = this;
-
-    return new Promise(function(resolve) {
-      task.testem.startDev(task.testemOptions(options), function(code) {
+    return new Promise(function(resolve, reject) {
+      task.testem.startDev(task.testemOptions(options), function(code, error) {
         remove(options.outputPath)
         .finally(function() {
-          resolve(code);
+          if (error) { reject(error); }
+          else       { resolve(code); }
         });
       });
     });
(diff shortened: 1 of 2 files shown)


stefanpenner:
I'll continue the ember-cli side of things.

@johanneswuerbach we should likely talk about how we should handle these timeouts (or prevent them from happening) Basically, it seems that during `ember test --server` we end up de-syncing some session that times out. Maybe @chancancode can share some more details, i'm a-bit hazy on the exact details, but it feels like something fishy is going on. Maybe it is how we continuously reinvoke testem, maybe leaking some intermediate result?

stefanpenner:
So a quick look, suggests maybe testems  `stopCurrentRun` which is invoked by `triggerRun` which happens on restart of testem, should also cancel the same timeout cancelled by `exit` [here](https://github.com/testem/testem/blob/master/lib/app.js#L167-L170)

johanneswuerbach:
Looks right, if you can implement that :+1: otherwise I should have time end of the weekend to fix it myself.

stefanpenner:
@johanneswuerbach I will at the very least submit a PR today.

stefanpenner:
superseded by: https://github.com/ember-cli/ember-cli/pull/5526

PR #5524, opened 2016-02-26
Enable X-Forwarded-* headers in http-proxy

jbacklund:
Reasoning: when using e.g. a Rails backend running on a separate port, proxying it using the ember-cli proxy feature, the generated _links_ sometimes tends to be incorrect. They are based on settings in the X-Forwarded-* headers. Hence, xfwd must be enabled.

If you like, we can make this a command-line option.

/cc @perlun, @jesjos

### DIFF
--- a/lib/tasks/server/middleware/proxy-server/index.js
+++ b/lib/tasks/server/middleware/proxy-server/index.js
@@ -14,7 +14,8 @@ ProxyServerAddon.prototype.serverMiddleware = function(options) {
       target: options.proxy,
       ws: true,
       secure: !options.insecureProxy,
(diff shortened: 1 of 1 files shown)


stefanpenner:
I wonder if we can leave this on by default... Im not familiar enough with this to know forsure.

perlun:
For reference: #5986 made this configurable.

alexander-alvarez:
For others who may be google searching why their django backends all of sudden started returning `302`s and `ERR_CONNECTION_REFUSED` when upgrading, I needed to set `transparent-proxy` to false.

Thanks for this

PR #5526, opened 2016-02-26
Propagates testem errors

stefanpenner:
* when testem crashes, reject with its error (not just status code)
* remove extra remove within the test task, the caller creates the dir, the caller must clean it up (which it already does)
* when wrapping the watcher, the entire event emitter callback needs to be try/catch wrapped to propogate the error
* move ui.StartProcess to within the promise executor (so the functions signature is accurate, always returning a promise)
* improve test coverage

### DIFF
--- a/lib/tasks/test-server.js
+++ b/lib/tasks/test-server.js
@@ -2,7 +2,6 @@
 
 var TestTask = require('./test');
 var Promise  = require('../ext/promise');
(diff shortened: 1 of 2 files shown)


stefanpenner:
@chadhietala gives +1 in chat

stefanpenner:
@homu r+

PR #5527, opened 2016-02-26
Fix many test and test server issues

stefanpenner:
clean test and test —server tasks
* share more
* always cleanup build
* use existing inheritance (don’t override methods that are basically the same)
* correctly propogate testem failures without —server being set


stefanpenner:
@homu r+

stefanpenner:
@homu r+

kellyselden:
@homu retry

PR #5528, opened 2016-02-26
preserve NPM/bower caches

stefanpenner:
lets give this a try:

* don't cache bower_components or node_modules directly
* cache system level caches of both

this should be better then previous attempts, as it aims to merely reduce network IO, not preserving local node_modules etc which resulted in many false positives. In addition npm cache has improved alot since the old days. so lets give it a go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 sudo: false
+cache:
+  directories:
+    $HOME/.npm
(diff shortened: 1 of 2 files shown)


stefanpenner:
@homu r+

rwjblue:
:+1:

PR #5531, opened 2016-02-27
load ember from ember-core if available

stefanpenner:
cc @rwjblue 

- [x] choose ember module name
- [ ] ship ember-core module

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -244,7 +244,7 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
   @method _initVendorFiles
 */
 EmberApp.prototype._initVendorFiles = function() {
-  var ember = this.project.findAddonByName('ember-source');
+  var ember = this.project.findAddonByName('ember-core');
   var developmentEmber;
   var productionEmber;
   var emberShims;
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -260,7 +260,7 @@ EmberApp.prototype._initVendorFiles = function() {
     productionEmber = this.bowerDirectory + '/ember/ember.prod.js';
     developmentEmber = this.bowerDirectory + '/ember/ember.debug.js';
     if (!existsSync(path.join(this.project.root, developmentEmber))) {
-      developmentEmber = this.bowerDirectory + 'vendor/ember/ember.js';
+      developmentEmber = this.bowerDirectory + '/ember/ember.js';
     }
   }
 
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -254,7 +254,7 @@ EmberApp.prototype._initVendorFiles = function() {
     productionEmber = ember.paths.prod;
     emberShims      = ember.paths.shims;
   } else {
-    emberShims = this.bowerDirectory + '/ember-cli-shims/app-shims.js',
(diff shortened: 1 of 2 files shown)


rwjblue:
We should check how far back findAddonByName goes, this project still supports early 1.13.x versions of Ember CLI. So we would either need to bump majors or detect and fallback (either is fine with me).

@stefanpenner - What do you think?

rwjblue:
Ugh, that comment was for the ember-cli-htmlbars PR not this one, sorry about the noise :(

rwjblue:
This looks good to me, not sure why it's failing though (haven't dug into the Travis errors).

stefanpenner:
@rwjblue its been around since ember-cli  0.2.0 https://github.com/ember-cli/ember-cli/commit/45f93f1d5b4f7d109003ba52e312cd1f70616875 I think we are good

lib/broccoli/ember-app.js
```diff
@@ -242,11 +242,24 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
   @method _initVendorFiles
 */
 EmberApp.prototype._initVendorFiles = function() {
-  // in Ember 1.10 and higher `ember.js` is deprecated in favor of
-  // the more aptly named `ember.debug.js`.
-  var defaultDevelopmentEmber = this.bowerDirectory + '/ember/ember.debug.js';
-  if (!existsSync(path.join(this.project.root, defaultDevelopmentEmber))) {
-    defaultDevelopmentEmber = this.bowerDirectory + '/ember/ember.js';
+  var ember = this.project.findAddonByName('ember-source');
```
Turbo87: the discussion in https://github.com/emberjs/ember.js/pull/13022 seemed to have converged on using `ember-core` instead of `ember-source`

lib/broccoli/ember-app.js
```diff
@@ -263,20 +276,14 @@ EmberApp.prototype._initVendorFiles = function() {
...
       { type: 'test' }
     ],
-    'app-shims.js': [
-      this.bowerDirectory + '/ember-cli-shims/app-shims.js', {
-        exports: {
-          ember: ['default']
-        }
-      }
-    ],
+    'app-shims.js': emberShims,
```
Turbo87: is the `exports` key not needed anymore?

rwjblue: Exports has been unused for quite some time (at least since ember-cli@1.13.0 but likely 0.2.0).

Turbo87: why is it still mentioned in http://ember-cli.com/user-guide/#standard-amd-asset then? or am I confusing two totally unrelated things here?

lib/broccoli/ember-app.js
stefanpenner: update to ember-core

lib/broccoli/ember-app.js
stefanpenner: > why is it still mentioned in http://ember-cli.com/user-guide/#standard-amd-asset then? or am I confusing two totally unrelated things here?

mistake :P

stefanpenner: leaving it for more, future work can clean this up.

Turbo87:
@stefanpenner LGTM r+

I think we can merge this already even though `ember-core` is not published yet.

stefanpenner:
@Turbo87 you are totally correct

@homu r+

PR #5532, opened 2016-02-28
use fork for better error messages

stefanpenner:
switch to broccoli fork, for now. We improved error messages, by making it obvious:

1. which broccoli plugin was invoked via `The Broccoli Plugin: [Name + description] failed With:`
2. where it was instantiated `The Broccoli plugin was instantiated at: ... stack ...`
3. error thrown, this was what existed before.

Latest broccoli (which needs more work) has something similar.

- [x] test if browser is correctly informed when a build fails during rebuild

screen shot 2016-02-27 at 4 53 48 pm

some future work should make the actual pages we reload to look pretty.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -44,7 +44,7 @@ module.exports = Task.extend({
     this.environment = this.environment || 'development';
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
 
-    var broccoli = require('broccoli');
+    var broccoli = require('ember-cli-broccoli');
     var hasBrocfile = existsSync(path.join('.', 'Brocfile.js'));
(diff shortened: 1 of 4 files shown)


chadhietala:
:+1: Such (:fork_and_knife: - :hocho: )

Turbo87:
@stefanpenner looks like there are some failing tests. missing dependency maybe?

stefanpenner:
@Turbo87 that was exactly the issue thank you (submitted https://github.com/ember-cli/ember-cli/commit/ed15446043ead1aca39a8798927abc3a13f81c5e to address) lets see what CI says .

stefanpenner:
@homu r+

PR #5533, opened 2016-02-28
reporter test option should tell us what its default is

ibroadfo:


kellyselden:
@homu r+

kellyselden:
Thank you.

rwjblue:
Hmm, I'm not sure about this. Does `testem` have fallbacks setup for reporter?  If it does, I'm afraid we have created another instance of the problem that I fixed in https://github.com/ember-cli/ember-cli/pull/5410.  

@ibroadfo - Can you double check how `testem` selects the reporter?  If there is a fallback system (like there was for config file) we should move the default information into the description instead...

kellyselden:
@rwjblue My mistake...

rwjblue:
Nah, I agree that it would be way better to list the default. But the listing is actually associated with passing options down into testem so I wanted to make sure...

ibroadfo:
happy to investigate; i had assumed this was effectively a documentation fix

rwjblue:
Ya, totally understandable (and you might be completely right anyways), I'm just scarred from tracking down the other issue I linked so wanted to double check...

ibroadfo:
so I tried setting via command line and config file to see how testem behaves; weirdly, it looks like setting the reporter in a config file doesn't actually work?

i'd certainly rather be conservative and just document the default rather than cause more problems, so i'll go ahead and push another commit to reflect that.

Issue #5538, opened 2016-03-01
Getting corrupted font assets in ember-cli 2.4.0


holandes22:
After upgrading to 2.4.0 I started getting font errors in chrome and FF console:

```
Failed to decode downloaded font: http://localhost:4200/assets/themes/default/assets/fonts/icons.woff2

OTS parsing error: incorrect entrySelector for table directory
```
which turn into broken icons. I use semantic-ui, so those come from that library (via semantic-ui-ember addon)

A difference I spotted between 2.3 and 2.4 is in the console output of the ember server command:

2.3 prints
```
...
SourceMapConcat: Concat: Vendor               | 394ms
...
```

whereas 2.4 prints
```
...
SourceMapConcat: Concat: Vendor /assets/ve... | 289ms
...
```

I checked the file sizes of the fonts and there is difference between the source and the destination:

```
$ ls -lah bower_components/semantic-ui/dist/themes/default/assets/fonts/
total 716K
drwxr-xr-x 2 pablo users 4.0K Jan  8 06:45 .
drwxr-xr-x 4 pablo users 4.0K Mar  1 10:13 ..
-rw-r--r-- 1 pablo users  60K Jan  8 06:45 icons.eot
-rw-r--r-- 1 pablo users  92K Jan  8 06:45 icons.otf
-rw-r--r-- 1 pablo users 307K Jan  8 06:45 icons.svg
-rw-r--r-- 1 pablo users 120K Jan  8 06:45 icons.ttf
-rw-r--r-- 1 pablo users  70K Jan  8 06:45 icons.woff
-rw-r--r-- 1 pablo users  56K Jan  8 06:45 icons.woff2

$ ls -lah dist/assets/themes/default/assets/fonts/
total 996K
drwxr-xr-x 2 pablo users 4.0K Mar  1 10:15 .
drwxr-xr-x 4 pablo users 4.0K Mar  1 10:15 ..
-rw-r--r-- 1 pablo users 109K Mar  1 10:24 icons.eot
-rw-r--r-- 1 pablo users 175K Mar  1 10:24 icons.otf
-rw-r--r-- 1 pablo users 307K Mar  1 10:24 icons.svg
-rw-r--r-- 1 pablo users 154K Mar  1 10:24 icons.ttf
-rw-r--r-- 1 pablo users 130K Mar  1 10:24 icons.woff
-rw-r--r-- 1 pablo users 103K Mar  1 10:24 icons.woff2
```

[...]

PR #5540, opened 2016-03-01
Revert amd module support

stefanpenner:
- breaks https://github.com/ember-cli/ember-cli/issues/5538 although https://github.com/stefanpenner/broccoli-persistent-filter/pull/48/files fixes
- walks + processes over all inputDependency paths (basically everything in bower_components) causing a performance regression in proportion to the number of files in bower_components

I originally thought the performance hit would be ok, but I overlooked the fact that this forces substantially more work in its current form. I would like to bring this back once the performance regression is addressed.

I suspect addressing the performance regression will also address #5538, as we should only be processing files we explicitly need to.

Currently we rely on funnel + mergeTrees not traversing to the leaves of these directories, and then finally concat [carefully only reading the files it truly cares about](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1140-L1151) this prevents us from incurring a large penalty if bower_components happens to contain many files, most of which are not used.

I suspect this tool could do something similar, pluck those files out first, or maybe become part of the concat story.

cc @ef4


rwjblue:
:+1: - LGTM

rwjblue:
I'm happy to help @ef4 work through the identified issues, but I think it is important for us to get a 2.4.1 out soonish fixing #5538.

ef4:
I think we can do something much more focused here, because the user is
declaring explicitly which files are subject to this feature. We can
construct a completely separate broccoli tree for them.

stefanpenner:
@ef4 ya totally, I just don't have cycles today or I would try and sort it out. We can restore once its updated.

as @kellyselden mentioned in chat, we should likely consider feature flags or similar.

stefanpenner:
@homu r+

PR #5547, opened 2016-03-02
Update core-object

btecu:
Current version of `core-object` is pre 1.0.0 (from 2014) and it causes installation to print warnings due to deprecated dependencies.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -107,7 +107,7 @@ module.exports = {
   },
 
   mapFile: function(file, locals) {
-    var result = this._super.mapFile.call(this, file, locals);
+    var result = this._super(file, locals);
     return this.fileMapper(result);
   },
 
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -33,6 +33,8 @@ module.exports = Command.extend({
   ],
 
   init: function() {
+    this._super.apply(this, arguments);
+
     this.assign    = require('lodash/assign');
     this.quickTemp = require('quick-temp');
 
--- a/lib/models/addon-discovery.js
+++ b/lib/models/addon-discovery.js
@@ -20,241 +20,238 @@ var findup     = require('findup');
   @extends CoreObject
   @constructor
 */
+var AddonDiscovery = CoreObject.extend({
+  init: function(ui) {
+    this.ui = ui;
+  },
 
-function AddonDiscovery(ui) {
-  this.ui = ui;
-}
+  /**
+    This is one of the primary APIs for this class and is called by the project.
+    It returns a tree of plain objects that each contain information about a
+    discovered addon. Each node has `name`, `path`, `pkg` and
+    `childAddons` properties. The latter is an array containing any addons
+    discovered from applying the discovery process to that addon.
 
-AddonDiscovery.__proto__ = CoreObject;
-AddonDiscovery.prototype.constructor = AddonDiscovery;
+    @private
+    @method discoverProjectAddons
+   */
+  discoverProjectAddons: function(project) {
+    var projectAsAddon = this.discoverFromProjectItself(project);
+    var internalAddons = this.discoverFromInternalProjectAddons(project);
+    var cliAddons = this.discoverFromCli(project.cli);
+    var dependencyAddons;
 
-/**
-  This is one of the primary APIs for this class and is called by the project.
-  It returns a tree of plain objects that each contain information about a
-  discovered addon. Each node has `name`, `path`, `pkg` and
-  `childAddons` properties. The latter is an array containing any addons
-  discovered from applying the discovery process to that addon.
-
-  @private
-  @method discoverProjectAddons
- */
-AddonDiscovery.prototype.discoverProjectAddons = function(project) {
-  var projectAsAddon = this.discoverFromProjectItself(project);
-  var internalAddons = this.discoverFromInternalProjectAddons(project);
-  var cliAddons = this.discoverFromCli(project.cli);
-  var dependencyAddons;
-
-  if (project.hasDependencies()) {
-    dependencyAddons = this.discoverFromDependencies(project.root, project.nodeModulesPath, project.pkg, false);
-  }  else {
-    dependencyAddons = [];
-  }
-
-  var inRepoAddons = this.discoverInRepoAddons(project.root, project.pkg);
-  var addons = projectAsAddon.concat(cliAddons, internalAddons, dependencyAddons, inRepoAddons);
-
-  return addons;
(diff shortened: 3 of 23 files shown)


stefanpenner:
awesome, will review in the AM.

We should be careful re: add ons expecting old behavior.

btecu:
There are two tests that I can't figure why they are failing.

Turbo87:
> We should be careful (but its totally worth the effort required) re: add ons expecting old behavior (old crappy super + stuff)

any idea how to resolve that? I'm not sure how many addons use those super calls, but if some of them do then this would likely be a breaking change for `ember-cli`.

stefanpenner:
I believe with some work the old and new way can be both supported

rwjblue:
Yes, agreed.  I did have a spike before introducing the original (and obviously broken) current implementation that did this.

rwjblue:
I am not really sure why we went with `this._super()` here though, since the ES6 `super` keyword is only valid in constructors and elsewhere in classes you would use `super.someMethodName()` (which is closer to what we already have today without the stupid binding issue that I created).

stefanpenner:
@rwjblue I'm fine with someone making it `super.method` as long as it doesn't have the binding issue you  mentioned

btecu:
@stefanpenner, those changes have to happen on `core-object` first, right?

stefanpenner:
if we go that route, ya

btecu:
Rebased and updated to use `core-object 2.0.0`.

Any suggestions about the last two failing tests?

stefanpenner:
i'll take a quick look right now, see if anything jumps out.

stefanpenner:
ya, those two failing tests look to be what i would consider invalid. They are calling `super` on methods that where not part of an extend block.

I may be able to make it work, but Im unsure if we want to support that or not.

stefanpenner:
Ya, it seems the following patch "makes it work".

```js
Object.defineProperty(CoreObject.prototype, '_super', {
  get: function() {
    return this.____thing || this.__proto__;
  },

  set: function(value) {
    this.____thing = value;
  }
});
```

That being said, Im inclined to say we don't support that (I suspect it has more holes then it fixes), and suggest the test should be update to `extend` blueprint to add `_super` using functions.

@rwjblue i dont believe blueprints to that sort of thing in the wild, but maybe you are aware of some?

rwjblue:
@btecu - Can you link to the source of the two failing tests?

stefanpenner:
@rwjblue 

```
  1) Blueprint basic blueprint installation calls appropriate hooks with correct arguments:
     TypeError: Cannot read property 'apply' of undefined
      at CoreObject.blueprint.filesPath (tests/unit/models/blueprint-test.js:776:37)
      at CoreObject.extend.files (lib/models/blueprint.js:312:26)
      at CoreObject.extend.install (lib/models/blueprint.js:488:43)
      at Context.<anonymous> (tests/unit/models/blueprint-test.js:808:24)

  2) Blueprint basic blueprint uninstallation calls appropriate hooks with correct arguments:
     TypeError: Cannot read property 'apply' of undefined
      at CoreObject.blueprint.normalizeEntityName (tests/unit/models/blueprint-test.js:890:47)
      at CoreObject.extend._normalizeEntityName (lib/models/blueprint.js:441:26)
      at CoreObject.extend.uninstall (lib/models/blueprint.js:536:10)
      at Context.<anonymous> (tests/unit/models/blueprint-test.js:934:24)
```

stefanpenner:
specifically these sorts of overrides: https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/blueprint-test.js#L752-L803

btecu:
@rwjblue any thoughts?
I'd like to complete this as soon as possible since rebases are a lot of work.

stefanpenner:
I'm inclined to say we change the test.

rwjblue:
Agreed

Robert Jackson

-- twitter: rwjblue
-- github: rjackson

btecu:
@stefanpenner those two tests are using `fixture/blueprints/basic/index.js` which already extends `Blueprint`. Calling `_super` there doesn't seem to fix those tests (and it seems unnecessary):
```
init: function() {
}
```

rwjblue:
@btecu - The issue is that by [this point](https://github.com/ember-cli/ember-cli/blob/caa24640bd986c8879291d6fd0b99c9d61fc1a86/tests/unit/models/blueprint-test.js#L752-L803) the core-object super wrapping has already happened (around [here](https://github.com/ember-cli/ember-cli/blob/caa24640bd986c8879291d6fd0b99c9d61fc1a86/tests/unit/models/blueprint-test.js#L476)). So overriding the functions after `.extend` will not have a functioning `this._super` (and I don't think we want to try to make that work...).

The fix here, would be to call `.extend` properly with these methods defined at `.extend` time...

stefanpenner:
@btecu ya, that test should instead subclass via `extend`

stefanpenner:
@rwjblue lol... good timing.

btecu:
So in `beforeEach` instead of `BasicBlueprintClass`:
```
...
blueprint: Blueprint.extend({
  filesPath: function(opts) {
    filesPathCalled = true;
  },
 ...
});
...
```

stefanpenner:
remaining failing tests:

```
  1) Blueprint basic blueprint installation calls appropriate hooks with correct arguments:
     TypeError: Cannot read property 'filesPath' of undefined
      at CoreObject.blueprint.filesPath (tests/unit/models/blueprint-test.js:775:27)
      at CoreObject.extend.files (lib/models/blueprint.js:9:3410)
      at CoreObject.extend.install (lib/models/blueprint.js:9:9068)
      at Context.<anonymous> (tests/unit/models/blueprint-test.js:807:24)
  2) Blueprint basic blueprint uninstallation calls appropriate hooks with correct arguments:
     TypeError: Cannot read property 'normalizeEntityName' of undefined
      at CoreObject.blueprint.normalizeEntityName (tests/unit/models/blueprint-test.js:889:27)
      at CoreObject.extend._normalizeEntityName (lib/models/blueprint.js:9:7448)
      at CoreObject.extend.uninstall (lib/models/blueprint.js:9:11186)
      at Context.<anonymous> (tests/unit/models/blueprint-test.js:933:24)
```

stefanpenner:
working on the rebase now, planning to take a look at the failures once thats resolved

stefanpenner:
got it mostly done, working on the failures next (time for lunch)

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/5652

nathanhammond:
Closing and moving to #5737.

PR #5550, opened 2016-03-02
Use ES6 shorthand for object literals

Ky6uk:


rwjblue:
Nice, thank you!

rwjblue:
@homu r+

PR #5551, opened 2016-03-02
Use spread operator in module-for-acceptance blueprint

Dhaulagiri:
It's more...suave this way

### DIFF
--- a/blueprints/app/files/tests/helpers/module-for-acceptance.js
+++ b/blueprints/app/files/tests/helpers/module-for-acceptance.js
@@ -8,13 +8,13 @@ export default function(name, options = {}) {
       this.application = startApp();
 
       if (options.beforeEach) {
-        options.beforeEach.apply(this, arguments);
+        options.beforeEach(...arguments);
       }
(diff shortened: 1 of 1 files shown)


rwjblue:
Sadly, this is actually incorrect ([see here](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&playground=false&code=let%20options%20%3D%20%7B%20beforeEach()%20%7B%20%7D%20%7D%3B%0A%0Aoptions.beforeEach(...arguments)%3B)), as it binds the `beforeEach` to the `options` object instead of `this`.  

I believe that the rule in JSCS is incorrect, and should be fixed (this is a valid use case of `.apply`).  See https://github.com/DockYard/ember-suave/issues/109 for tracking in ember-suave.

Dhaulagiri:
TIL, thanks @rwjblue

rwjblue:
NP, sorry for the troubles.  I need to find some time to try and fix the JSCS rule. :disappointed:

PR #5560, opened 2016-03-03
Fix issue with amd-name-resolver update.

rwjblue:


stefanpenner:
Oops. Thanks good catch

stefanpenner:
@homu r+

rwjblue:
Homu's build had a random failure, Travis PR and branch build both passed around the same time. 

Merging....

PR #5561, opened 2016-03-03
Add ability to lint templates.

rwjblue:
Allows tooling like ember-cli-template-lint to work.

### DIFF
--- a/ADDON_HOOKS.md
+++ b/ADDON_HOOKS.md
@@ -387,8 +387,8 @@ linter output as test results.
 
 **Received arguments:**
 
-- tree type ('app', 'tests', or 'addon')
-- tree of Javascript files
+- tree type ('app', 'tests', 'templates', or 'addon')
+- tree of files (JavaScript files for `app`, `tests`, and `addon` types)
 
 **Source:** [lib/broccoli/ember-app.js:335](https://github.com/ember-cli/ember-cli/blob/v0.1.15/lib/broccoli/ember-app.js#L335)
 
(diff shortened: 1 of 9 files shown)


rwjblue:
Updated to add tests and remove `WIP` label.  This should be ready to go.

rwjblue:
Node 0.12 and 5 passed in AppVeyor, the failures there seem random (I've restarted 3 times and a different node version fails each time :disappointed:).

rwjblue:
@stefanpenner - any objections?

rwjblue:
Chatted in slack about this with Stef, we need to do some benchmarks to make sure we don't introduce a perf regression...

rwjblue:
Tests of master vs these changes in the ember-cli/stress-app (after updating it in https://github.com/ember-cli/stress-app/pull/13) are [here](https://usecanvas.com/rwjblue/stress-app-tests-for-ember-cli5561/6aK0wqjOWr2pgUFmHeukqE).

Basic gist:

Description | master (b88de69ba3) | #5561 |
-----|------|-----|
 Initial Boot - Cold | 171216ms | 172612ms |
 Initial Boot - Warm | 9329ms | 7931ms |
 Rebuild - js file | 2027ms | 1660ms |
 Rebuild - sass file | 1826ms | 1872ms |

stefanpenner:
@rwjblue  :+1:  looks like nothing surprising snuck in re: perf, :ship:

rwjblue:
Ya, but definitely better to be safe than sorry...

rwjblue:
I'm doing a few more small checks locally with my addon (ember-cli-template-lint) to make sure all is well.  Will land it as soon as everything there checks out...

rwjblue:
@homu r+

PR #5233, opened 2015-12-18
multi meta module support

offirgolan:
Support the ability to load modules from meta tags in the index.html while being SRI compliant.

```html
<meta name="my-app/config/user" data-module=true content="<escaped json>" />
```

Any meta tag with `data-module=true` (including config/environment.js) will be defined as a module via the `name` attribute. 

This will allow us to do the following:

```js
import UserConfig from 'my-app/config/user'
// do some stuff here
```

PR #5566, opened 2016-03-04
Remove require cache caused shared mutation in slow tests.

rwjblue:
Prior to this change, the require cache was causing us to mutate a single shared `package.json` (one for `some-cool-app` and one for `some-cool-addon`), but each of these tests assume they are modifying the default blueprints `package.json`. Instead, they were all mutating a shared copy (cached on the first test).

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -81,7 +81,7 @@ describe('Acceptance: addon-smoke-test', function() {
     return copyFixtureFiles('addon/component-with-template')
       .then(function() {
         var packageJsonPath = path.join(__dirname, '..', '..', 'tmp', addonName, 'package.json');
-        var packageJson = require(packageJsonPath);
+        var packageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf8' }));
         packageJson.dependencies = packageJson.dependencies || {};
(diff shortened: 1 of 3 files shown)


Turbo87:
Seems like AppVeyor disagrees with this change, but I'm not sure why. Doesn't look like random test failures though.

nathanhammond:
Knocked over AppVeyor again. Looking forward to trying this in combination with #5233.

rwjblue:
Looks like node 4 and 5 pass in AppVeyor but not 0.12, but all of the node versions passed properly in Travis. 

I'm not sure what about readFileSync/writeFileSync that would cause issues on Windows only in node 0.12...

Turbo87:
do we have to manually call https://nodejs.org/api/fs.html#fs_fs_closesync_fd or something like that? I would assume that `readFileSync()` calls this itself but who knows...

rwjblue:
I think its just random failures, see https://github.com/ember-cli/ember-cli/pull/5561#issuecomment-192312602. That PR has the same failures as this one has one only for Node 4 (and Node 0.12 passed there)...

stefanpenner:
looks like appveyor is now crashing because phantom failed to be installed (Seeing it elsewhere as well) kicking it again.

I can try this out on my windows machine tonight if we haven't sorted it out

Turbo87:
seems to work fine now

@homu r+

PR #5575, opened 2016-03-04
blueprints/route-addon: Fix __path__ token

### LINKED ISSUE #5574: `ember generate route-addon foo` generating route file in wrong folder
running `ember generate route-addon foo` in an addon should generate two wrapper files:
- `app/routes/foo.js`
- `app/templates/foo.hbs`

but instead it generates
- `app/route-addons/foo.js`
- `app/templates/foo.hbs`

Turbo87:
when generated directly this caused the route to be generated in the `route-addons` folder instead of `routes`

resolves #5574 

/cc @rwjblue

### DIFF
--- a/blueprints/route-addon/index.js
+++ b/blueprints/route-addon/index.js
@@ -29,13 +29,11 @@ module.exports = {
         return options.dasherizedModuleName;
       },
       __path__: function(options) {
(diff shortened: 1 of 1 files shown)


rwjblue:
LGTM

Turbo87:
@homu r+

PR #5584, opened 2016-03-06
Help command cleanup

Turbo87:
This PR cleans up the `help` command and the corresponding acceptance test


Turbo87:
@kellyselden could you take a quick look at  70cc7b3? looks like a bug to me, but maybe I'm wrong and this is actually intended.

kellyselden:
@Turbo87 Looks intended to me. I think `this` is scoped to the addon in that `eachAddonCommand` block, so it should be printing addon json.

Turbo87:
@kellyselden well, the unit test kinda shows that it doesn't

kellyselden:
@Turbo87 on second thought, it looks like https://github.com/ember-cli/ember-cli/blob/53e5174467a18c2bd0b5afff6b64d267a9a5b69d/lib/commands/help.js#L78 says `this` isn't addon scope, so your change looks good.

Turbo87:
@kellyselden r?

kellyselden:
@homu r+

Turbo87:
@homu retry

PR #5592, opened 2016-03-07
Fix issue with addon directory not being linted

### LINKED ISSUE #5498: Linters not running on addon files
I'm one of the maintainers of [`ember-ajax`](https://github.com/ember-cli/ember-ajax) and am working on upgrading the addon to use Ember CLI`2.3.0`.  I noticed that some of the file were being linted before are no longer being linted after upgrading from `1.13.15`.  Specifically, any of the files that are inside the `addon` directory are now being ignored, both by JSHint and JSCS (through `ember-suave`) although all of the other files are still being handled correctly.

I took the outputs of `ember test` for both versions, cleaned up the output to make them comparable, and the diff can be found [here](https://www.diffchecker.com/zvdrblyf).  There are two new lines in the `2.3.0` file, which are expected, but the missing lines are unexpected.

## Todo

- [x] Add method as mentioned in https://github.com/ember-cli/ember-cli/issues/5498#issuecomment-191798300
- [x] Unit test `_eachProjectAddonInvoke` method
- [x] Add an addon smoke test confirming that ember test fails when an addon's addon/ tree.

alexlafroscia:
Closes #5498

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -8,6 +8,7 @@ var addonName  = 'some-cool-addon';
 var spawn      = require('child_process').spawn;
 var chalk      = require('chalk');
 var expect     = require('chai').expect;
+var EOL        = require('os').EOL;
 
 var symlinkOrCopySync   = require('symlink-or-copy').sync;
 var runCommand          = require('../helpers/run-command');
@@ -161,6 +162,19 @@ describe('Acceptance: addon-smoke-test', function() {
       })
       .then(function(result) {
         expect(result.exitCode).to.eql(0);
+      });
+  });
+
+  it('ember addon with linting errors', function() {
+    return copyFixtureFiles('addon/with-linting-errors')
+      .then(function() {
+        return ember(['test']);
+      })
+      .then(function() {
+        expect(false, 'should have rejected with a failed linter').to.be.ok;
+      })
+      .catch(function(result) {
+        expect(result.exitCode).to.not.eql(0);
       });
   });
 
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -587,6 +587,26 @@ Addon.prototype._addonTemplateFiles = function addonTemplateFiles(tree) {
 };
 
 /**
+  Invoke the specified method for each of the project's addons.
+
+  @private
+  @method _eachProjectAddonInvoke
+  @param {String} methodName the method to invoke on each addon
+  @param {Array} args the arguments to pass to the invoked method
+*/
+Addon.prototype._eachProjectAddonInvoke = function eachProjectAddonInvoke(methodName, args) {
+  this.initializeAddons();
+
+  var invokeArguments = args || [];
+
+  return this.project.addons.map(function(addon) {
+    if (addon[methodName]) {
+      return addon[methodName].apply(addon, invokeArguments);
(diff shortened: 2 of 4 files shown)


miguelcobain:
:tada:

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -164,6 +164,16 @@ describe('Acceptance: addon-smoke-test', function() {
       });
   });
 
+  it('ember addon with linting errors', function() {
+    return copyFixtureFiles('addon/with-linting-errors')
+      .then(function() {
+        return ember(['test']);
+      })
+      .then(function(result) {
+        expect(result.exitCode).to.not.eql(0);
```
rwjblue: Can we also check the `results.output` to confirm that the right test failed? Something like (stolen from another test I recently added):

```
var output = result.output.join(EOL);
expect(output).to.match(/TemplateLint:/, 'ran template linter');
```

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -164,6 +164,16 @@ describe('Acceptance: addon-smoke-test', function() {
       });
   });
 
+  it('ember addon with linting errors', function() {
+    return copyFixtureFiles('addon/with-linting-errors')
+      .then(function() {
+        return ember(['test']);
+      })
+      .then(function(result) {
```
rwjblue: The `ember` helper should reject with non-zero exit codes so this `then` is likely not being called (probably needs to change to `.catch`).  This kind of thing makes be wish for `assert.expect(1)` from QUnit...

https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/ember.js#L85

rwjblue:
@alexlafroscia - Looks really good, thanks for tackling.  I left two comments (one is a bug and the other is more of a nit-pick) to address, then we can land this...

tests/acceptance/addon-smoke-test-slow.js
alexlafroscia: Huh, interesting.  After working with this a little, it seems like the linting is actually passing, even with that `with()` statement in the file, which I would have thought should make the linter unhappy.  Need to investigate this a bit more.

alexlafroscia:
Alright, so I pushed an update to the previous commit that updates the acceptance-style test to use `.catch` to verify that the linting works.

One thing that I was unable to do was verify the output text the way you mentioned.  Inside the `catch` handler, `result.outputStream` didn't have any information about the actual test that failed, and there didn't seem to be any way to get the output from the `ember test` command.  The information on the test that failed gets output to the screen instead.

tests/acceptance/addon-smoke-test-slow.js
alexlafroscia: I tried to do this; `results.output` was undefined, and `results.outputStream` didn't have anything in it that could be used to verify which test exactly failed.

tests/acceptance/addon-smoke-test-slow.js
alexlafroscia: Got this part verified :+1:

rwjblue:
Thank you @alexlafroscia!

alexlafroscia:
@rwjblue thanks for your help and guidance on this! I'm excited to finally be contributing to the CLI itself :smile:

PR #5593, opened 2016-03-07
Fix: Prepend legacy files

### LINKED ISSUE #5589: app.legacyFilesToAppend.push deprecation
Regarding Issue #5537 which should cause a deprecation warning since 2.4.1 #5545 does not work.
issue
It says "Build successful" but it is not successful because the page is empty and does not load.

I use ember-material-lite in my package.json and it uses app.legacyFilesToAppend.push function to import vendor javascript. (https://github.com/mike-north/ember-material-lite/blob/master/index.js)

ro0gr:
Fixes #5589

I believe  some day we will have a backward compatible implementation of `app.import(.., {outputFile: ..})`  :)

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1174,7 +1174,7 @@ EmberApp.prototype.javascript = function() {
   if (this.legacyFilesToAppend.length > 0) {
     deprecate('Usage of EmberApp.legacyFilesToAppend is deprecated. Please use EmberApp.import instead for the following files: \'' + this.legacyFilesToAppend.join('\', \'') + '\'');
     this.legacyFilesToAppend.forEach(function(legacyFile) {
-      this.import(legacyFile);
+      this.import(legacyFile, {prepend: true});
     }.bind(this));
   }
 


rwjblue:
I don't think this is better than what we have now. The issue is that people seem to expect to be able to maintain ordering in that array while interspersing app.imports and array pushes. 

AFAIK, only ember-material-lite is affected (and it is doing some pretty bizarre things in its included hook). If @mike-north could release an update that uses app.import this would be resolved...

ro0gr:
It looks like we've broken `legacyFilesToAppend` original behaviour(`ember-material-lite` is only one we are aware of) so I think it might have a sense to keep backward compatibility. 

From another side I agree that the problem is in the `ember-material-lite` package. I still don't understand what exactly it does [here](https://github.com/mike-north/ember-material-lite/blob/master/index.js#L10) :p

Hirtenknogger:
Some time ago i created an issue and PR (https://github.com/mike-north/ember-material-lite/issues/100) but it would be great to keep backward compatibility untill @mike-north releases an update.

Hirtenknogger:
what happens next? 
can the PR be merged?

ro0gr:
After some time I'd rather agree with @rwjblue. Trying to fix each broken addon in the wild from the `ember-cli` side we can:
 - break dozens of "good" addons
 - over-complicate `ember-cli` itself

which is undesirable.
Also this pr looks quite strange even for me cause I still don't understand how exactly it fixes this `ember-material-lite` issue :flushed: 

@Hirtenknogger I'm going to close this pr for now. Hope @mike-north will accept your pr as soon as possible and cut a new release for `ember-material-lite`.

PR #5597, opened 2016-03-08
Use internal test helpers

Turbo87:
This PR removes a few test helpers in favor of https://github.com/ember-cli/ember-cli-internal-test-helpers


stefanpenner:
:+1:

stefanpenner:
@homu r+

broerse:
What is the upgrade path for assert in add-on's? See:

https://github.com/ember-cli-deploy/ember-cli-deploy-build/blob/master/tests/unit/index-nodetest.js

Turbo87:
you can use either the built-in assert module or chai or any other js assertion library

broerse:
Thanks. Perhaps I can also switch to https://github.com/ember-cli/ember-cli-internal-test-helpers ? Or are the only for ember-cli internal?

broerse:
What is the the built-in assert module called?

broerse:
It seems I have to write this:

```javascript
var chai           = require('chai');
var chaiAsPromised = require('chai-as-promised');

chai.use(chaiAsPromised);

module.exports = chai.assert;
```
But ember-cli had that as a helper. Is there a new helper from the build-in assert module and where can I find it?

broerse:
Could not find the new helper so switched to internal test helpers. They works fine. Thanks! See:

https://github.com/martinic/ember-cli-deploy-build-plus/blob/master/tests/unit/index-nodetest.js

PR #5599, opened 2016-03-09
Revert Enable X-Forwarded-* headers in http-proxy

jrjohnson:
#5524 is a breaking change as these headers cause proxying to break when using a symfony backend in development environment mode.

I'm not in any way opposed to including these headers - in fact they maybe should have always been included - just probably needs to be behind an option flag now or else indicated as a breaking.

@jbacklund is putting this as an option doable?

### DIFF
--- a/lib/tasks/server/middleware/proxy-server/index.js
+++ b/lib/tasks/server/middleware/proxy-server/index.js
@@ -14,8 +14,7 @@ ProxyServerAddon.prototype.serverMiddleware = function(options) {
       target: options.proxy,
       ws: true,
       secure: !options.insecureProxy,
-      changeOrigin: true,
-      xfwd: true
+      changeOrigin: true
     });
 
     proxy.on('error', function (e) {


perlun:
I think we shouldn't drop it straight away, but conditionalize it may make sense (i.e. make it an option like `proxyXfwd`). ember-cli is broken for us without it.

nathanhammond:
We discussed this at our last team meeting. :) Our goal here is to make it configurable as it impacts two separate groups in different ways. Unless we can determine that one is "more-correct" we'd like to fix forward and make this configurable. Would any of you like to get involved in making this happen?

We'll help review PRs and push this in quickly. :)

perlun:
@jrjohnson - do you know more about the breakage? We had to do a change in https://github.com/nodejitsu/node-http-proxy/pull/967 to get it working, and then upgrade the version used in `ember-cli` to use a custom-built version of the `http-proxy` package including that change. Only then could we get the xfwd stuff working correctly (when proxying data from an ember-cli app to a rails app).

(Note: I'm not saying here that we _shouldn't_ make it configurable. Just saying that that specific problem - the symfony backend - could potentially be resolved by using a more recent node-http-proxy build.)

PR #5606, opened 2016-03-13
Don't bother pointless npm/bower restore in addon CI

mike-north:
After tests pass (or fail), ember-try will (by default) restore `bower_components` and `node_modules` to their original states, as defined by the committed `bower.json` and `package.json`, respectively.

This is kind of pointless for most addon-ci-testing scenarios, since you're really just after the pass or fail -- no point in going back to the original dependency state.

CC: @kategengler @kellyselden

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -32,4 +32,4 @@ install:
   - bower install
 
 script:
-  - ember try $EMBER_TRY_SCENARIO test
+  - ember try $EMBER_TRY_SCENARIO test --skip-cleanup
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -32,4 +32,6 @@ install:
   - bower install
 
 script:
+  # Usually, it's ok to finish the test scenario without reverting
+  #  to the addon's original dependency state, skipping "cleanup".
   - ember try $EMBER_TRY_SCENARIO test --skip-cleanup


kategengler:
The gotcha with --skip-cleanup is that if CI is also deploying, it could deploy with the last scenario run. Probably doesn't apply to many addons, though I know some deploy demo sites, but worth noting.

kellyselden:
@kategengler Good point. I think I'm pro skipping being the default, because I think it will benefit the majority. The rest can just remove it from the config. Thoughts?

kategengler:
That seems fine, I just wanted to document that caution here.

mike-north:
@kategengler - awesome point. Maybe there's something we can enhance ember-cli-deploy to warn if a `bower.json.ember-try` is found or something? People usually [will have to implement something custom](https://github.com/mike-north/ember-cli-materialize/blob/master/lib/try-generate-pages.sh) if they're deploying from a scenario, to avoid deploying from **all** scenarios.

blueprints/addon/files/.travis.yml
```diff
@@ -32,4 +32,4 @@ install:
   - bower install
 
 script:
-  - ember try $EMBER_TRY_SCENARIO test
+  - ember try $EMBER_TRY_SCENARIO test --skip-cleanup
```
Turbo87: how about adding a small comment here that explains why the flag is used and in which cases it should be removed?

Turbo87:
@homu r+

kellyselden:
@homu r+

PR #5607, opened 2016-03-13
Removing view blueprint

jasonmit:
Since ember-cli now targets 2.x, which doesn't expose `Ember.View`, we should likely remove it and instead move the blueprint to the `ember-legacy-views` addon.

Related: http://discuss.emberjs.com/t/can-not-generate-a-view-for-the-index-route/10131


rwjblue:
Agreed.

stefanpenner:
@homu r+

jasonmit:
@stefanpenner tests should now be passing

kellyselden:
@homu r+

mike-north:
In the future  we may wish to at least leave the de-generator intact for this kind of thing. `ember d view-test my-view` continues to be of value as people are upgrading to ember 2.5.

jasonmit:
The blueprints were moved into the legacy view addon but it looks like that was never versioned.

/cc @rwjblue @stefanpenner

nathanhammond:
If we do a new release of ember-legacy-views then a new `npm install` would give you the ability to `destroy`. It's not too late. Can we bump that version to 0.2.1?

toranb:
@stefanpenner @jasonmit ...so is ember-cli 2.4.x not ember 1.13 friendly anymore ? What is the last version ember 1.13 devs can use w/out conflict ?

jasonmit:
In lock step with Ember, but I see no reason why it wouldn't work with 1.13.  If you need the blueprints you can continue to pull them in through the legacy view addon (assuming that was versioned).

rwjblue:
Or just make the files by hand...

toranb:
@rwjblue @jasonmit awesome! just wanted to make sure this was (overall) ember 1.13/2.x friendly :)

PR #5609, opened 2016-03-14
[ENHANCEMENT] Add in v1 of welcome page addon

acorncom:
Opening this to continue the discussion about the Ember welcome page.  History on this is available here:

- https://github.com/locks/ember-welcome-page/issues/4
- https://github.com/emberjs/website/pull/2469
- and https://github.com/ember-cli/ember-cli/pull/5501

Per private discussions with @rwjblue we're aiming for a simple `/ember-getting-started` route that won't interfere with other urls in consuming (new) apps and will iterate on this (and add acceptance tests) as discussed in the `ember-welcome-page` repo once we've got a base version merged.

Thoughts on this?  Text and use of the Tomster image has been reviewed / signed-off on by @wifelette 

If we're ok with the way this looks, I can start working on getting an NPM package pushed and getting the repo for this moved to the `ember-cli` organization ...

Screenshot:
screen shot 2016-03-14 at 2 12 18 pm

To do:

- [x] finish work on taking over `/` (having some issues with that)
- [x] write some basic tests
- [x] change to using a published NPM package
- [ ] move the ember-welcome-page add-on to Ember-CLI org


acorncom:
Hmm, I seem to be blowing up some smoke tests due to this line (suggested by rwjblue):

https://github.com/locks/ember-welcome-page/blob/master/index.js#L19

Any suggestions on how to handle this?

rwjblue:
Ya, it should be `this.ui.writeLogLine` most likely (just to funnel through our normal UI object).  Though that may not be why they are failing.

acorncom:
The smoke tests had my `console.log` note in them, might be as well ;-)  Just changed that on `master` and re-kicked the tests ...

acorncom:
Screenshot of CLI output:
screen shot 2016-03-14 at 4 30 01 pm

acorncom:
Earlier build issues were due to using a template string which node 0.10 and 0.12 didn't like in our smoke tests ...

Turbo87:
how do you feel about introducing a `--skip-welcome` parameter to `ember new`?

Turbo87:
also what is the reason for providing that page at `/ember-getting-started` instead of `/`?

stravid:
@Turbo87 I assume the reason behind this URL path is:

> [...] won't interfere with other urls in consuming (new) apps [...]

rwjblue:
@Turbo87 - The idea is that we will always have it at `/ember-getting-started` and we will take over `/` when we detect that:

* No custom routes have been added
* No custom templates have been added

I mocked up a [naive detection for this here](https://github.com/locks/ember-welcome-page/issues/5#issuecomment-196521365), this would only run on boot.

Turbo87:
> we will take over `/` when we detect ...

:+1: sounds good

blueprints/app/files/package.json
```diff
@@ -37,6 +37,7 @@
     "ember-export-application-global": "^1.0.4",
     "ember-load-initializers": "^0.5.0",
     "ember-resolver": "^2.0.3",
+    "ember-welcome-page": "locks/ember-welcome-page#master",
```
stefanpenner: we will need this to be from npm, before we can merge. Firewall users, often have issues with direct github deps

stefanpenner:
looks good

blueprints/app/files/package.json
acorncom: :+1:

acorncom:
> I mocked up a naive detection for this here, this would only run on boot.

@rwjblue Do we want the naive detection to notice when the filesystem has been updated?  If we've taken over `/`, seems like we'd want that to disappear if someone runs `ember g template application` in another window ...  Thoughts?

acorncom:
> how do you feel about introducing a --skip-welcome parameter to ember new?

@Turbo87 I'm a :+1: on the overall idea, but don't have a lot of spare cycles on my end.  Would we want that to be part of a v1 implementation of this or could we implement that post merge of the first version here?  I'd prefer to see it done post initial implementation myself ...

rwjblue:
@acorncom we can have the detection run for every '/' and remove the middleware when things change. In other words we do not need to proactively watch the trees, we can just detect when we hit a '/' request.

rwjblue:
I'm not sure we should have a flag to disable this. We don't have flags to disable any of the other default addons (ember-ajax, ember-data, etc). It is pretty simple to remove from package.json after initial install...

stefanpenner:
> I'm not sure we should have a flag to disable this.

Ya, i don't think we need a flag for this.

acorncom:
I think this is ready at this point! :smile:

blueprints/app/files/app/templates/application.hbs
```diff
@@ -1,3 +0,0 @@
-<h2 id="title">Welcome to Ember</h2>
```
rwjblue: This makes me so happy!

raido:
Links in welcome.html point to v2.3.0, should these point to upcoming LTS?

I would be nice to fill in repository information in package.json.

rwjblue:
Good catch @raido! We definitely need to fix the repo info, but I think it should point to whatever ember version is "current stable". For that to work well, we need to permissively version this (which likely means 1.0.0), so we can allow bumps along with ember-cli.

rwjblue:
When we get more sophisticated in the addon, we could detect the running ember version and use that directly. But Id prefer to ship this as an MVP soon and continue iterating.

acorncom:
> Links in welcome.html point to v2.3.0, should these point to upcoming LTS?

Thanks for catching that.  I've updated to `v2.4.0` for now.  On the guides side of things, we're hoping to setup a `/current` once we work out the SEO issues that come with that.  But that might take a bit

> It would be nice to fill in repository information in package.json.

I was hoping to do that once the repo had been moved under the `ember-cli` org (if we're still doing that @rwjblue ?)  If we aren't, I'll update that.

> For that to work well, we need to permissively version this (which likely means 1.0.0), so we can allow bumps along with ember-cli.

@rwjblue I've been bumping the numbers up slowly as we iterate on this, but once I get a thumbs up, I'll change version to `1.0.0`.  Would we want to change to using `^1.0.0` instead of `1.0.0` in the `package.json` file?

blueprints/app/files/app/templates/application.hbs
acorncom: Me too ;-)

acorncom:
@rwjblue Ready to merge here when ready

rwjblue:
Thank you!

@homu r+

rwjblue:
Test failure seems legitimate. There is some test (something about wrapInEval, which isn't even a thing we support anyways) that is checking for the string 'Welcome to Ember'.

acorncom:
@rwjblue Confirmed, I'm seeing the test failing as well locally.  How should we proceed?

I can remove all `wrapInEval` items if desired (here's what I would remove), is that what we want to do?

- https://github.com/ember-cli/ember-cli/blob/6f276023a6e96da35d4c8ffdd7a0bde12ef45aee/tests/acceptance/brocfile-smoke-test-slow.js#L66-L71
- https://github.com/ember-cli/ember-cli/blob/2f4622f47e772afda83f5db32a13481fce0e41ef/tests/fixtures/brocfile-tests/wrap-in-eval/ember-cli-build.js
- https://github.com/ember-cli/ember-cli/blob/2f4622f47e772afda83f5db32a13481fce0e41ef/tests/fixtures/brocfile-tests/wrap-in-eval/tests/integration/wrap-in-eval-test.js

If we don't do that, I'm puzzled about how to make this test pass ... ;-)

rwjblue:
Ya, kill those tests (we don't do anything special for that flag any more).

acorncom:
@rwjblue Any preferences on how we fix the broken test now?  We're doing a check to see if "Welcome to Ember" shows up with a default app here:

- https://github.com/ember-cli/ember-cli/blob/6f276023a6e96da35d4c8ffdd7a0bde12ef45aee/tests/acceptance/brocfile-smoke-test-slow.js#L166-L171
- https://github.com/ember-cli/ember-cli/blob/6f276023a6e96da35d4c8ffdd7a0bde12ef45aee/tests/fixtures/brocfile-tests/default-development/tests/integration/app-boots-test.js#L24

My inclination is to yank that out as well, but not sure that's a good idea.  Do we support installing other add ons as part of a test?

acorncom:
@rwjblue Looks like I've finally got the tests passing here ;-)

acorncom:
Hmm, looks like the tests that are failing are unrelated to what I changed ...

rwjblue:
Awesome, great job!!

rwjblue:
@homu r+

jeanleonino:
@acorncom, @rwjblue +1 for `/current` or relevant aliases for docs.

rwjblue:
We don't need `/current` since we actually know the version being used by the project we can link to the exact version of the guides for the project.

jeanleonino:
True indeed and would be more useful than getting current...

acorncom:
@jeanleonino If you're interested in helping work on this, I've got more details on what we need over here: https://github.com/ember-cli/ember-welcome-page/issues/27 😄 Otherwise, will aim to get that knocked off prior to Ember-CLI 2.6 release ...

PR #5614, opened 2016-03-15
Resolve directories relative to EmberApp project root

trentmwillis:
Currently, a lot of the directories used within `EmberApp` are implicitly relative to the process' current working directory. This causes issues if you want to invoke a build from a process outside the cwd.

For example, without going into too much detail, we're working on multiple applications that we would like to build in parallel. Since the invoking command will be run outside the individual application directories, the lookups can't be relative to the process cwd.

cc @chadhietala

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -218,25 +218,35 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
   this.options.babel.sourceMaps = false;
 
   this.options.trees = defaults(this.options.trees, {
-    app:       new WatchedDir('app'),
-    tests:     new WatchedDir('tests'),
+    app:       new WatchedDir(this._resolveLocal('app')),
+    tests:     new WatchedDir(this._resolveLocal('tests')),
 
     // these are contained within app/ no need to watch again
     // (we should probably have the builder or the watcher dedup though)
-    styles:    new UnwatchedDir('app/styles'),
-    templates: existsSync('app/templates') ? new UnwatchedDir('app/templates') : null,
+    styles:    new UnwatchedDir(this._resolveLocal('app/styles')),
+    templates: existsSync(this._resolveLocal('app/templates')) ? new UnwatchedDir(this._resolveLocal('app/templates')) : null,
 
     // do not watch vendor/ or bower's default directory by default
     bower: this.project._watchmanInfo.enabled ? this.bowerDirectory : new UnwatchedDir(this.bowerDirectory),
-    vendor: existsSync('vendor') ? new UnwatchedDir('vendor') : null,
+    vendor: existsSync(this._resolveLocal('vendor')) ? new UnwatchedDir(this._resolveLocal('vendor')) : null,
 
-    public: existsSync('public') ? new WatchedDir('public') : null
+    public: existsSync(this._resolveLocal('public')) ? new WatchedDir(this._resolveLocal('public')) : null
   });
 
   this.options.jshintrc = defaults(this.options.jshintrc, {
     app: this.project.root,
-    tests: path.join(this.project.root, 'tests'),
+    tests: this._resolveLocal('tests'),
   });
+};
+
+/**
+  Resolves a path relative to the project's root
+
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -218,28 +218,38 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
...
 };
 
 /**
+  Resloves a path relative to the project's root
+
+  @private
+  @method _resolveLocal
+*/
+EmberApp.prototype._resolveLocal = function(to) {
+  return path.resolve(this.project.root, to) || to;
```
rwjblue: It probably doesn't matter, but why use resolve instead of join here?

rwjblue:
Also, I think at least https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L76 needs to be tweaked (possibly more methods too).

stefanpenner:
I am a fan of this, we should be careful these path / resolve changes often uncover other issues. We should be sure to double check on windows, and other strange configurations.

cc @kellyselden maybe you got some time to give this a while on windows.

lib/broccoli/ember-app.js
trentmwillis: No real reason, just the first thing my fingers typed out. I can change to `path.join`.

trentmwillis:
@rwjblue looks like it. I imagine there are more places to update.

trentmwillis:
@rwjblue I looked into changing the `bowerDirectory` implementation, but there are some usages that I don't think should be resolved, [such as this one](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1347). So I'm not 100% sure how to proceed.

stefanpenner:
@trentmwillis that one should already be correctly relative.

lib/broccoli/ember-app.js
```diff
@@ -218,28 +218,38 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
...
 
   this.options.jshintrc = defaults(this.options.jshintrc, {
     app: this.project.root,
-    tests: path.join(this.project.root, 'tests'),
+    tests: this._resolveLocal('tests'),
   });
 };
 
 /**
+  Resloves a path relative to the project's root
```
Turbo87: typo :wink:

kellyselden:
> cc @kellyselden maybe you got some time to give this a while on windows.

@stefanpenner I'll check it out today.

trentmwillis:
@stefanpenner that's my point. It seems like `bowerDirectory` is intended to be relative in some use cases, but we could also resolve it in others. So I'm not sure how to proceed, should we have a property to represent the Bower directory and one to represent the path to the directory?

kellyselden:
> windows, and other strange configurations.

@stefanpenner lol strange Windows. Anyways, tried it out in Windows and all seems good.

lib/broccoli/ember-app.js
```diff
@@ -218,28 +218,38 @@ EmberApp.prototype._initOptions = function(options, isProduction) {
...
   });
 };
 
 /**
+  Resolves a path relative to the project's root
+
+  @private
+  @method _resolveLocal
+*/
+EmberApp.prototype._resolveLocal = function(to) {
```
stefanpenner: this function should at-least be unit-tested

trentmwillis: Added a unit test for this. Let me know if you think I should cover any other cases.

stefanpenner: the current implementation has a branch, `join(x,to) || to`  although i', unclear when falsy case will be taken. I suspect if there is such a scenario, it should also be added to the test.

trentmwillis: Doh. That was a remnant of prior work. I'll clean it up.

stefanpenner:
@trentmwillis i think a rebase, and all should be green.

trentmwillis:
@stefanpenner rebased.

rwjblue:
@homu r+

PR #4709, opened 2015-08-19
[FEATURE] Autocompletion

lazybensch:
check out the [RFC](https://github.com/lazybensch/rfcs/blob/add_shell_completion/active/0000-command-line-completion.md) to learn more about this feature.

things left to be done:
- [x] make CI green
- [x] apply code review comments
- [ ] add `fish` support
- [x] decide on cache invalidation technique
- [ ] re-source shell config file

to entertain you until then:

PR #5615, opened 2016-03-16
Extract JSON help generator from HelpCommand code

Turbo87:
This PR extracts the code responsible for `ember help --json` into a `json-generator` utility, that can be reused in other parts of the code.

/cc @kellyselden @lazybensch

### DIFF
--- a/lib/commands/help.js
+++ b/lib/commands/help.js
@@ -34,8 +34,6 @@ module.exports = Command.extend({
   ],
 
   run: function(commandOptions, rawArgs) {
-    var multipleCommands = [GenerateCommand.prototype.name].concat(GenerateCommand.prototype.aliases);
-    var command;
     var json;
     var rootCommand = new RootCommand({
       ui: this.ui,
@@ -92,8 +90,9 @@ module.exports = Command.extend({
         }.bind(this));
       }
 
+      var multipleCommands = [GenerateCommand.prototype.name].concat(GenerateCommand.prototype.aliases);
       if (multipleCommands.indexOf(rawArgs[0]) > -1) {
-        command = rawArgs.shift();
+        var command = rawArgs.shift();
         if (rawArgs.length > 0) {
           commandOptions.rawArgs = rawArgs;
         }
--- a/lib/commands/help.js
+++ b/lib/commands/help.js
@@ -7,8 +7,7 @@ var string          = require('../utilities/string');
 var assign          = require('lodash/assign');
 var GenerateCommand = require('./generate');
 var RootCommand     = require('../utilities/root-command');
-var versionUtils    = require('../utilities/version-utils');
-var emberCLIVersion = versionUtils.emberCLIVersion;
+var JsonGenerator   = require('../utilities/json-generator');
 
 module.exports = Command.extend({
   name: 'help',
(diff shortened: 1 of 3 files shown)


lazybensch:
if we wanna go this way i should probably pause https://github.com/ember-cli/ember-cli/pull/4709 until this one is in master. If you like i can help out here as i am already familiar with help JSON generation. Just let me know here or in slack if i can contribute to this PR

Turbo87:
@lazybensch feel free to work on caching the output based on this PR/branch and your previous work in #4709, but should go in a separate PR then.

lib/commands/help.js
```diff
@@ -103,36 +87,30 @@ module.exports = Command.extend({
...
-  },
+    var json = generator.generate(commandOptions, rawArgs);
 
-  _addCommandHelpToJson: function(commandName, single, options, json) {
-    var command = this._lookupCommand(commandName);
-    if (!command.skipHelp || single) {
-      json.commands.push(command.getJson(options));
-    }
+    var outputJsonString = JSON.stringify(json, function(key, value) {
+      // build command has a recursive property
```
lazybensch: I think we should fix that instead of introducing a mapping function.
the [build command](https://github.com/Turbo87/ember-cli/blob/json/lib/commands/build.js#L14) should specify `type: String` as it is done [here](https://github.com/Turbo87/ember-cli/blob/json/lib/commands/test.js#L31).
If you do so, you can get rid of [this line](https://github.com/Turbo87/ember-cli/blob/json/lib/commands/build.js#L3) aswell. same would need to be done [here](https://github.com/Turbo87/ember-cli/blob/json/lib/commands/serve.js#L33).

Then you could rewrite this hole thing into
```javascript
var outputJsonString = JSON.stringify(json, null, 2);
```

lib/commands/help.js
```diff
@@ -92,8 +75,9 @@ module.exports = Command.extend({
         }.bind(this));
       }
 
+      var multipleCommands = [GenerateCommand.prototype.name].concat(GenerateCommand.prototype.aliases);
       if (multipleCommands.indexOf(rawArgs[0]) > -1) {
-        command = rawArgs.shift();
+        var command = rawArgs.shift();
         if (rawArgs.length > 0) {
```
lazybensch: this `if` is redundant inside [this `if`](https://github.com/Turbo87/ember-cli/blob/json/lib/commands/help.js#L43).

lib/commands/help.js
Turbo87: I agree that this should be cleaned up, but I just moved what was already existing, so IMHO this can be cleaned up in a later commit/PR.

lib/commands/help.js
Turbo87: not quite correct because `rawArgs.shift()` modifies `rawArgs`

lazybensch: true that

lib/commands/help.js
lazybensch: ja ok cool, rest looks good to me then

lazybensch:
I already prepared two branches based off this work. one two clean up the circular dependency and another one to write the generated json into a cache and to refresh that cache after every `npm i` or `ember install`

I will publish the PRs as soon as this one here gets merged.

lazybensch:
@nathanhammond since you've added this to the core meeting agenda:
the work in this PR (and what i wrote in my last comment) is a subset of what was done in #4709. It will, however reduce the complexity of the autocompletion PR a bit by merging these things in stages.

kellyselden:
This looks good to me. Nice work @Turbo87!

kellyselden:
@homu r+

nathanhammond:
@lazybensch I'm happy for this to be split into pieces, it makes it nice and easily digestible.

PR #5617, opened 2016-03-16
Move "coveralls" and "codeclimate" dependencies into TravisCI manifest

Turbo87:
This PR moves the "coveralls" and "codeclimate" dependencies into the TravisCI manifest. They are usually only needed for CI and not on developer machines.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,6 @@ install:
 script: npm run-script test-all:cover
 
 after_script:
+  - npm install coveralls codeclimate-test-reporter
   - cat coverage/lcov.info | codeclimate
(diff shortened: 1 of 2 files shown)


stefanpenner:
I believe we may want to run coveralls locally as well.

Turbo87:
@stefanpenner are you sure? the coverage information is still shown locally when `npm run test-all:cover` is used because it's provided by `istanbul`. `coveralls` only feeds that information into https://coveralls.io/ which I don't think should be done locally.

stefanpenner:
Oops. You are correct. :ship:

Turbo87:
@homu r+

PR #5618, opened 2016-03-16
Cleanup unused utilities

Turbo87:
This PR removes unused utilities that were previously extracted into separate NPM packages.


Turbo87:
@stefanpenner @kellyselden @trabus r?

stefanpenner:
@homu r+

kellyselden:
Only thing to watch out for is addons reaching in and using these utils. Not much of a concern for me though.

Turbo87:
we could deprecate it but I'm not sure it's worth it... should be mentioned in the release notes though

PR #5621, opened 2016-03-17
[BUGFIX] Adds in fix for #5620 to see if the actualOutputStream isTTY by default.

### LINKED ISSUE #5620: Prompts are messed up (arrow keys not working, highlighting working, checkbox busted).
Prompts are busted, choices are supposed to be highlighted when you enter a number, arrow keys aren't being read properly, and multiple choice types aren't working.

Arrow Key being pressed.
image

Nothing being highlighted.
image

Scrolling being busted.
image

Multiple Choice being busted.
image

Ember 2.4.2
Node: 5.9.0
OS: OS X 10.11.3

iheanyi:
So I filed the bug in #5620 and after digging deeper into the code, I found the problem.

https://github.com/ember-cli/ember-cli/blob/master/lib/ui/index.js#L40-L55

`actualOutputStream` is equal to `process.stdout`. So `process.stdout.isTTY` exists. Making a call to `outputStream` using `through` gets rid of the `isTTY` function from the actualOutputStream because it's no longer returning a `tty.WriteStream` anymore. 

Looking at createInterface options [here](https://nodejs.org/api/readline.html#readline_readline_createinterface_options), you can see that they say that the `terminal` option defaults to 

> terminal - pass true if the input and output streams should be treated like a TTY, and have ANSI/VT100 escape codes written to it. Defaults to checking isTTY on the output stream upon instantiation.

Since the outputStream in `createInterface` is not the actual output stream, the `isTTY` function therefore does not exist. Therefore, we have to make a call to `this.actualOutputStream.isTTY`in the `terminal`option to get expected behavior, thus fixing errors with the Inquirer package and multiple selection in prompts. See screencaps.

HIghlighting on selection.
image

Scrolling fixed.
image

Checkbox selection.
image

It was a one-line fix and wasn't sure how to test it...so let me know if there's any way I can improve this pull request. =) 

resolves #5620

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -205,7 +205,8 @@ UI.prototype.prompt = function(questions, callback) {
   PromptExt.prototype.constructor = PromptExt;
   PromptExt.prototype.rl = this.readline.createInterface({
     input: this.inputStream,
-    output: promptOutputStream
+    output: promptOutputStream,
+    terminal: this.actualOutputStream.isTTY,
   });
 
   // If no callback was provided, automatically return a promise


stefanpenner:
@homu r+

Turbo87:
unfortunately this seems to have broken the prompt in `models/file-info.js`.

this can be quickly tested by inserting the following code into `cli/index.js`:

```js
  var FileInfo = require('../models/file-info');
  var fi = new FileInfo({ ui: ui });
  return fi.confirmOverwrite('test.js');
```

```
# prompt works in both cases
[?] Overwrite test.js? (Yndh) 

# the entered text should appear after the prompt. the whole prompt disappears with this fix.
[?] Overwrite test.js? (Yndh) y

# doesn't appear at all with this fix
[?] Overwrite test.js? Yes, overwrite
```

PR #5622, opened 2016-03-17
[CLEANUP] command types

lazybensch:
the `output-path` option of `ember build` and `ember serve` should be described as `type: String` to be instead of as the path module object. Also this way we do not need to special treat circular dependencies in help json output.

### DIFF
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -2,6 +2,7 @@
 
 var Command = require('../models/command');
 var win     = require('../utilities/windows-admin');
+var optionType = require('../utilities/option-type');
 
 module.exports = Command.extend({
   name: 'build',
@@ -10,7 +11,7 @@ module.exports = Command.extend({
 
   availableOptions: [
     { name: 'environment', type: String,             default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
-    { name: 'output-path', type: function Path() {}, default: 'dist/',       aliases: ['o'] },
+    { name: 'output-path', type: optionType('Path'), default: 'dist/',       aliases: ['o'] },
     { name: 'watch',       type: Boolean,            default: false,         aliases: ['w'] },
     { name: 'watcher',     type: String }
   ],
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var path    = require('path');
 var Command = require('../models/command');
 var win     = require('../utilities/windows-admin');
 
@@ -11,7 +10,7 @@ module.exports = Command.extend({
 
   availableOptions: [
     { name: 'environment', type: String,  default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
-    { name: 'output-path', type: path,    default: 'dist/',       aliases: ['o'] },
+    { name: 'output-path', type: String,  default: 'dist/',       aliases: ['o'] },
     { name: 'watch',       type: Boolean, default: false,         aliases: ['w'] },
     { name: 'watcher',     type: String }
   ],
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -9,9 +9,9 @@ module.exports = Command.extend({
   aliases: ['b'],
 
   availableOptions: [
-    { name: 'environment', type: String,  default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
-    { name: 'output-path', type: 'Path',  default: 'dist/',       aliases: ['o'] },
-    { name: 'watch',       type: Boolean, default: false,         aliases: ['w'] },
+    { name: 'environment', type: String,           default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
+    { name: 'output-path', type: ['Path', String], default: 'dist/',       aliases: ['o'] },
+    { name: 'watch',       type: Boolean,          default: false,         aliases: ['w'] },
     { name: 'watcher',     type: String }
   ],
 
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -9,9 +9,9 @@ module.exports = Command.extend({
   aliases: ['b'],
 
   availableOptions: [
-    { name: 'environment', type: String,           default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
-    { name: 'output-path', type: ['Path', String], default: 'dist/',       aliases: ['o'] },
-    { name: 'watch',       type: Boolean,          default: false,         aliases: ['w'] },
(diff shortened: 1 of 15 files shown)


Turbo87:
did `path` have any hidden meaning when validating the command line option?

lazybensch:
it was never checked or used except for when generating the help output. There the option was checked to be the path object and then mapped into the 'Path' String. So i expect no side effects at all.

It's by the way correct for the test command that always had `  --path (String) Reuse an existing build at given path.` as output

Turbo87:
can't we add a `Path` utility that is essentially a string but will be converted to "path" for the help output?

lazybensch:
no need. if you want the help output to display `type: Path` it is enough to change the type to `"Path"` (being the string with value "Path") and [print-command](https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/print-command.js#L57) will sort it out.

I have no opinion about whether it should display "Path" or "String" as type, so your call, i will change the PR accordingly.

Turbo87:
if it's that easy then we should probably change it to "Path". also you will probably have to adjust the `ember help` tests due to these changes.

lazybensch:
On it, one remark though that i just thought of:

IDEs that are using ember help --json might wanna check the option type (String, Boolean, Number) to render an appropriate input field, a type of "Path" might confuse here. I am still fine with having it "String" though, just wanted to mention it.

kellyselden:
@lazybensch Yes, we are going to need to consider what are acceptable changes in our json output, now that others are consuming it. At what point to we consider semver in our json output.

Turbo87:
> a type of "Path" might confuse here

- it has been "path" before due to the special JSON rendering code
- it actually allows IDEs to display a file or path chooser instead of a string input field

kellyselden:
@nathanhammond I'll bring up the issue of semver and what it means for our json interface at the meeting today.

Turbo87:
@kellyselden related: https://github.com/ember-cli/ember-cli/issues/5626

lib/commands/serve.js
```diff
@@ -20,20 +19,20 @@ module.exports = Command.extend({
...
-    { name: 'ssl-key',          type: String,  default: 'ssl/server.key' },
-    { name: 'ssl-cert',         type: String,  default: 'ssl/server.crt' }
+    { name: 'port',                 type: Number,  default: defaultPort,   aliases: ['p'] },
+    { name: 'host',                 type: String,                          aliases: ['H'],     description: 'Listens on all interfaces by default' },
+    { name: 'proxy',                type: String,                          aliases: ['pr', 'pxy'] },
+    { name: 'insecure-proxy',       type: Boolean, default: false,         aliases: ['inspr'], description: 'Set false to proxy self-signed SSL certificates' },
+    { name: 'watcher',              type: String,  default: 'events',      aliases: ['w'] },
+    { name: 'live-reload',          type: Boolean, default: true,          aliases: ['lr'] },
+    { name: 'live-reload-host',     type: String,                          aliases: ['lrh'],   description: 'Defaults to host' },
+    { name: 'live-reload-base-url', type: String,                          aliases: ['lrbu'], description: 'Defaults to baseURL' },
```
kellyselden: description alignment is still off by one

Turbo87:
@lazybensch TravisCI still has some broken tests

lazybensch:
ya will take care of those tomorrow

lib/commands/build.js
```diff
@@ -10,9 +9,9 @@ module.exports = Command.extend({
   aliases: ['b'],
 
   availableOptions: [
-    { name: 'environment', type: String,  default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
-    { name: 'output-path', type: path,    default: 'dist/',       aliases: ['o'] },
-    { name: 'watch',       type: Boolean, default: false,         aliases: ['w'] },
+    { name: 'environment', type: String,             default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }] },
+    { name: 'output-path', type: function Path() {}, default: 'dist/',       aliases: ['o'] },
```
Turbo87: why can't we just use `type: 'Path'` here?

tests/unit/models/builder-test.js
```diff
@@ -86,6 +86,7 @@ describe('models/builder.js', function() {
     it('allow outputPath to contain the root path as a substring, as long as it is not a parent', function() {
       var outputPathArg = '--output-path=.';
       var outputPath = command.parseArgs([outputPathArg]).options.outputPath;
+      outputPath = path.resolve(outputPath);
```
Turbo87: why do we need this?

lazybensch: shoot! left-over

lib/commands/build.js
lazybensch: while fixing tests it turned out that quite a few places expect the type to be a constructor and rely on `type.name` to exist or check `typeof type === 'function'` (not only the output formatter but also validations and the logic that decides to use the default value or not and so on).

So the easiest way is to let it be a function with a proper name and be consistent with all the other types available.

Turbo87: that what I meant with:

> did path have any hidden meaning when validating the command line option?

I'm not quite convinced yet that this is the right way to go. If we move forward in this direction we should at least extract the `Path` thing into a utility module/file.

lazybensch: yeah i remember, back then i thought that was simply not the case.
how do you imagine such a module?

`var optionType = require('utils/option-type')`
`type: optionType('Path')`

did i understand your idea correctly here?

Turbo87: nope, just:

```js
var Path = require('utils/path-option-type');

{ ..., type: Path, ... }
```

lazybensch: ok i will do it like this.
but just to show you what i prepared with the other approach:

utilities/option-type.js:
```javascript
'use strict';

module.exports = function(name) {
  return new Function('return function ' + name + '() {};')();
};
```

this way we could not only do: `type: optionType('Path')` but also `type: optionType('Url')` or  `type: optionType('FooBar')` if we want to in the future.

I don't know whether this type of meta programming is a bad practice though.

any thoughts? otherwise i will implement it the way you proposed as `utilities/path-option-type.js`

lazybensch: i will push "my version" for now so that you can have a look at it but again, if you object it, i will change it to a pathOptionType.

kellyselden:
I don't know the best issue to post this on because we have multiple JSON interface fronts right now, but here it is. We discussed the JSON interface and how it applies to SemVer on the team meeting yesterday. We determined it is an experimental feature, because we didn't know how people would want to consume it when I wrote it. That being said, feel free to break it all you want to make it best for all. After it stabilizes, we will "1.0" it and be more careful about changes.

lazybensch:
Ci seems happy now

stefanpenner:
@homu +R

stefanpenner:
@homu r+

Issue #5630, opened 2016-03-17
Cache `ember help --json` output


Turbo87:
For consuming applications (ember-hearth, vsc-ember-cli, intellij-ember) of `ember help --json` it can make sense to cache the output to make it faster and to simplify reading it because the application would only need to read a JSON file instead of shelling out.

Something similar was already proposed in #4709 and the result of this issue should be reused in the PR.

This cached file obviously would need to updated at the appropriate times:

- installing addons via `ember install` and `npm install`
- removing addons

it might be possible to use the `postinstall` of NPM for this and we should probably provide an ember-cli command to manually update the cache.

/cc @lazybensch @stefanpenner @kellyselden

lazybensch:
i have prepared a branch that implements exactly this, will publish it tomorrow

stefanpenner:
this isn't really an issue, future proposals like this should be RFC's.

Issue #5629, opened 2016-03-17
inconsistent `ember help <something> --json` output


lazybensch:
when i do `ember help --json` the output is:

```
{
  "name": "ember",
  "description": null,
  "aliases": [],
  "works": "insideProject",
  "availableOptions": [],
  "anonymousOptions": [
    "<command (Default: help)>"
  ],
  "version": "2.4.2-master-ada0309dcf",
  "commands": [
    ...
  ],
  "addons": [
    {
      "name": "ember-cli-release",
      "commands": [
        {
          "name": "release",
          ....
        },
        ...
      ]
    },
    ...
  ]
```

i abbreviated most of the output but note how the release command is nested in its `ember-cli-release` addon in the `addons` list. However if i do `ember help release --json`

```
{
  "name": "ember",
  "description": null,
  "aliases": [],
  "works": "insideProject",
  "availableOptions": [],
  "anonymousOptions": [
    "<command (Default: help)>"
  ],
  "version": "2.4.2-master-ada0309dcf",
  "commands": [
    {
      "name": "release",
      ...
    },
    ...
  ],
  "addons": []
}
```

it is now listed under the `commands` list and the `addon` list is empty.

third party consumers might get confused when using both ways to fetch data so we should agree on one API here.

(be aware of #5626 though)

Turbo87:
that is exactly one of the reasons why I opened #5626

the output of the JSON help is/was too closely bundled to the original help output which caused such inconsistencies

lazybensch:
so in case we deprecate/remove the special case, should we still fix this issue for the current version or do we ignore this since not many seem to use this feature atm anyway?

Turbo87:
IMHO we should leave it as it is

kellyselden:
https://github.com/ember-cli/ember-cli/pull/5622#issuecomment-198360045 Just here to confirm the API can change as we see fit. No need to worry about backwards compatibility.

Issue #5626, opened 2016-03-17
Deprecate `ember help <something> --json`


Turbo87:
As mentioned in https://github.com/EmberTown/ember-hearth/issues/97 `ember help --json` already contains all the necessary information needed to derive the other ones which makes it pointless to use any specialized version of it. This is different from `ember help` vs `ember help generate` because `ember help --json` *does* contain information about the available blueprints and their options already.

I would like to propose to deprecate `ember help <something> --json` in ember-cli@2.5 and remove it in ember-cli@2.6.

AFAIK ember-hearth is the only user of `ember help --json` so far and they are only using the generic variant which should remain the same.

kellyselden:
@Turbo87 https://github.com/felixrieseberg/vsc-ember-cli also uses it to scrape blueprints

Turbo87:
indeed... (https://github.com/felixrieseberg/vsc-ember-cli/blob/b4903fa0fe34116ae83e459f5ac617aa3553b21b/src/ember-ops.ts#L154)

@felixrieseberg can you comment on why you're using the "specialized" form?

felixrieseberg:
There's no need for the specialized form on my end, I can switch.

kellyselden:
@Turbo87 see https://github.com/ember-cli/ember-cli/pull/5622#issuecomment-198360045. No need for the extra work of deprecation, Feel free to just gut it out. We can do our best to notify the consumers, but other than that, it should be considered experimental while we stabilize the interface.

Turbo87:
@kellyselden still I wouldn't want to break vs-ember-cli so I think deprecating it until @felixrieseberg has adjusted the code makes sense

felixrieseberg:
Quick update: I just got around to fixing it. VSCode is now cool with deprecating the command.

Turbo87:
@felixrieseberg awesome, thanks for the feedback!

PR #5635, opened 2016-03-18
[FEATURE][WIP] Cache help json

lazybensch:
resolves https://github.com/ember-cli/ember-cli/issues/5630

i still need to refresh the cache on addon removal and write tests but you can already have a look at it.

i created a command `generate-json` that stores the help json under `your/project/path/node_modules/ember-cli/commands-json.js` (feel free to suggest better naming for both)

I added that (undocumented) command as a post-install hook in the app blueprint (and will probably need to add it to addon blueprint aswell)

the `ember help --json` command will now, read the cache instead of building the json itself. Except when you provide any arguments to it, then it will calculate it (which is slower but as mentioned in https://github.com/ember-cli/ember-cli/issues/5629 the output of `ember help some-argument --json` is currently inconsistent so i won't bother reimplementing broken logic for that case. Also https://github.com/ember-cli/ember-cli/issues/5626 suggests deprecating this feature all together


lib/commands/generate-json.js
```diff
@@ -0,0 +1,76 @@
...
+
+  _getJson: function(commandOptions, rawArgs) {
+    var generator = new JsonGenerator({
+      ui: this.ui,
+      project: this.project,
+      commands: this.commands,
+      tasks: this.tasks
+    });
+
+    return generator.generate(commandOptions, rawArgs);
```
Turbo87: I don't think we should pass any options here because the output should always be the same independent from any user-provided options

lib/commands/generate-json.js
```diff
@@ -0,0 +1,76 @@
...
+    var cache = JSON.stringify(json, function(key, value) {
+      var path = require('path');
+      // build command has a recursive property
+      if (value === path) {
+        return 'path';
+      }
+      return value;
+    }, 2);
+
+    if (rawArgs.length === 0) {
```
Turbo87: why do you check `rawArgs` here?

lib/commands/generate-json.js
```diff
@@ -0,0 +1,76 @@
...
+  ],
+
+  anonymousOptions: [],
+
+  cacheDirectory: '/node_modules/ember-cli',
+  cacheFileName: 'commands-help.json',
+
+  run: function(commandOptions, rawArgs) {
+
+    commandOptions.json = true;
```
Turbo87: is that really necessary? is the `JsonGenerator` still checking this anywhere?

lib/commands/generate-json.js
```diff
@@ -0,0 +1,76 @@
...
+  skipHelp: true,
+  works: 'everywhere',
+
+  availableOptions: [
+    { name: 'silent', type: Boolean, default: true, aliases: ['s'] },
+  ],
+
+  anonymousOptions: [],
+
+  cacheDirectory: '/node_modules/ember-cli',
```
Turbo87: unused?

lib/commands/generate-json.js
```diff
@@ -0,0 +1,76 @@
...
+      commands: this.commands,
+      tasks: this.tasks
+    });
+
+    return generator.generate(commandOptions, rawArgs);
+  },
+
+  _cacheJson: function(json) {
+
+    var directoryPath = path.join(__dirname, '../..')
```
Turbo87: it might make more sense to use something like `path.join(this.project.path, 'node_modules', 'ember-cli')` here for cases where `npm link` was used to use the development version of `ember-cli`

although I agree that this is probably a pretty uncommon use-case

lib/commands/help.js
```diff
@@ -93,24 +94,26 @@ module.exports = Command.extend({
...
-    }, 2);
+    var hasCache = fs.existsSync(cachePath);
+
+    if (rawArgs.length === 0 && hasCache) {
+
+      jsonString = fs.readFileSync(cachePath, 'utf8');
+
+    } else {
+
+      jsonString = generateJsonCommand.run(commandOptions, rawArgs);
```
Turbo87: I don't think we should reuse the generate-json command here for calls like `ember help <something> --json`. The generate-json command should only generate the main JSON file structure and nothing else. Handling those specific cases should be the reponsibility of the help command and the JSON generator for now.

lib/commands/init.js
```diff
@@ -107,17 +107,17 @@ module.exports = Command.extend({
         }
       })
       .then(function() {
-        if (!commandOptions.skipNpm) {
-          return npmInstall.run({
-            verbose: commandOptions.verbose,
-            optional: false
+        if (!commandOptions.skipBower) {
+          return bowerInstall.run({
+            verbose: commandOptions.verbose
```
Turbo87: why are those changes needed?

Turbo87:
I think I'd prefer something like `ember update-package-cache` or something like this for the command name and maybe `.structure.json` for the filename

lib/commands/init.js
lazybensch: having the generation command as a postinstall hook means it gets executed when you create a new ember-cli project and npmInstall.run() executes. Which is great but any ember command will fail if bowerInstall did not run before. So here i make sure that first all bower deps are installed, then all node modules and only **then** `ember update-package-cache` gets called

lib/commands/help.js
lazybensch: sounds reasonable, this will also address most of the comments you had inside the generate-json command (it was effectively handling args and commands for this very case here.

lib/commands/init.js
Turbo87: oh okay, I missed that bower and npm switched places. I just saw that the `optional` option was removed, but it looks like it just moved further down.

Turbo87: I'm wondering though whether it would be possible to disable the "installed packages" check for the new command...

lib/commands/generate-json.js
lazybensch: i actually changed it to `__dirname` for this exact reason. Here is the different behaviour:
```
$ /foobar -> npm link ember-cli
$ /foobar -> ember generate-json

this.project.root: /Users/lazybensch/src/opensource/foobar
__dirname: /Users/lazybensch/src/opensource/ember-cli/lib/commands
```

`__dirname` will **always** point to the correct location (since its that file that is executed it has to be in the ember-cli directory that is in charge right now)

lib/commands/init.js
lazybensch: what do you mean with "installed packages" check? you mean that the postinstall runs immediately after you create a new project? If so: isn't it a good thing to create the cache right away?

lib/commands/generate-json.js
Turbo87: true, but we actually want to use the project folder since the each project might have different addons installed which will lead to different JSON files.

also IDEs will expect the file to be in `/path/to/project/node_modules/ember-cli/something.json` and figuring this `npm link` stuff out would be much harder in those cases

lazybensch: it will work in this scenario too. If i am already inside an ember-cli project and i link ember-cli the following happens:
```
$ /my-project -> npm link ember-cli
$ /my-project -> ember generate-json

this.project.root: /Users/lazybensch/src/opensource/ember-cli/lib/commands
__dirname: /Users/lazybensch/src/opensource/ember-cli/lib/commands
```

if i am a regular user and do not link the ember-cli module this is the behaviour:

```
$ /my-project -> ember generate-json

this.project.root: /Users/lazybensch/src/opensource/my-project/node_modules/ember-cli/lib/commands
__dirname: /Users/lazybensch/src/opensource/my-project/node_modules/ember-cli/lib/commands
```

You see, inside projects __dirname works as expected and outside of projects __dirname is **required** for this to work.

Keep in mind that __dirname is the dirname of that very file that is executed right now and also that this directory has to be inside the correct ember-cli directory (be it global installation or inside a project, cause, well, thats how npm works)

hope i'm not being stupid right now :)

lib/commands/init.js
Turbo87: > any ember command will fail if bowerInstall did not run before

this is what I meant. I assume that you meant that it will fail because the bower deps are not installed, right? but the cache file doesn't actually depend on any bower dependencies afaik so we could just keep the npm first, bower second order if we manage to get around that "bower deps are not installed" check for this particular command.

the reason why I think this would be beneficial is that the `ember-cli` update instructions mention `npm install && bower install`, which would also fail if we keep the current solution.

lib/commands/generate-json.js
Turbo87: are you sure about `this.project.root: /Users/lazybensch/src/opensource/ember-cli/lib/commands` in the first case?

lazybensch: no i miss-pasted there: it would be
`this.project.root: /Users/lazybensch/src/opensource/ember-cli-filter-by-query` (which would still be correct. when i join it with 'node_modules/ember-cli' but since this directory is just a symlink to `/Users/lazybensch/src/opensource/ember-cli` it has the same effect as path.join(__dirname, '../..')

so still, both versions work inside a linked project
both versions work inside a not-linked project
only __dirname works when outside a project

lib/commands/init.js
lazybensch: super good point. will look into this

lib/commands/generate-json.js
Turbo87: symlink... right...

this will still lead to issues if `ember-cli` is linked into two or more projects, but I don't think that use case is common enough to have special handling code for it. 

/cc @rwjblue @stefanpenner

stefanpenner:
It's not just on addon removal, rather an addon change. This cache key is abit tricky to get right

stefanpenner:
Also before we cache, can we do an analysis of what is slow that we can make sure we aren't just covering up bugs with our caching

lazybensch:
when an addon changes you would still need to do `npm i` for it to take effect right? and that will trigger regenerate the cache.

when you don't cache that json then you will need to "boot up" ember, you need to traverse all your addon directories and nested addon directories to generate that hash. this takes probably +1sec. so not a big deal for the plain old `ember help` but `ember help --json` consuming apps might not want to wait that long + this is mainly in preparation for autocompletion feature.

Actually we moved it out of the autocompletion ticket to make the merge more managable and when you want to use the json for shell completion you cannot wait +1sec anymore imho

stefanpenner:
> npm i for it to take effect right? and that will trigger regenerate the cache.

that is true, unless one is developing an addon. ala npm link

stefanpenner:
> when you don't cache that json then you will need to "boot up" ember

so if we had a persistent process this would also not be a problem?

lazybensch:
> that is true, unless one is developing an addon. ala npm link
true, for that scenario i have no good idea except communicating the `ember update-package-cache` command to addon authors, so they can use it when they make changes

stefanpenner:
Seems reasonable. We should just make sure to message that in the developing addons doc

lib/commands/init.js
lazybensch: seems like i can disable this check as you assumed but i'll get:
```
Using a version of ember-cli-shims prior to 0.1.0 will cause errors while loading Ember Data 2.3.0-beta.3+. Please update ember-cli-shims from undefined to 0.1.0.
```

as `ember-cli-shims` is brought in via bower.json

any thoughts?

lazybensch:
I am currently implementing a way to skip dependency validation for this command (as discussed in the code-line-comment above) and turns out it's not that complicated **but**:

while running `jsonGenerator.generate()` and it is traversing through the addons to build the json it will at some point throw `Using a version of ember-cli-shims prior to 0.1.0 will cause errors while loading Ember Data 2.3.0-beta.3+. Please update ember-cli-shims from undefined to 0.1.0.`

this seems to be since `ember-cli-shims` is brought in via bower.json but the hole exercise here is to be able to run this command without bower dependencies. :(

any thoughts on that issue?

Turbo87:
@lazybensch please retry with the current ember-data master branch. I think this should be fixed in the next release.

lazybensch:
yeah that was it, thx!

Turbo87:
@lazybensch see https://github.com/emberjs/data/pull/4230 for context

lazybensch:
ran into another small issue that i need to pick your brains about. I am fixing all existing tests currently and there is one last problem i'm stuck with.

my PR is introducing this new command `ember update-package-cache` and obviously the current ember release does not know this command. However when smoke tests run and do `ember new` there is a postinstall hook that will try to run `ember update-package-cache` once all npm modules are installed.  However this newly created project specifies `"ember-cli": "<%= emberCLIVersion %>", which will crash since it is trying to execute a command it does not know.

```
"The specified command update-package-cache is invalid. For available options, see `ember help`.\n"
"some-cool-addon@0.0.0 postinstall: `ember update-package-cache`\nExit status 1\n"
```

Turbo87:
closing this for now since cache invalidation still seems to be an unsolved issue

PR #5642, opened 2016-03-20
update ember-try

kellyselden:
The only thing different is the way they call it I believe: `ember try scenario` to `ember try:one scenario`

@kategengler feel free to add.


stefanpenner:
:+1:

kellyselden:
@homu r+

PR #5643, opened 2016-03-20
[ENHANCEMENT] Exclude dist and tmp dirs from search in vscode

Tyriar:
Improve find in files in Visual Studio Code by excluding generated files in
a workspace config file.

---

Not sure if you accept editor-specific changes like this but I found ember-cli far more convenient with vscode when this was included after a little experimentation. (FYI I work on vscode)

image

I couldn't find any similar tests but let me know if you want one.

### DIFF
--- a/blueprints/app/files/.vscode/settings.json
+++ b/blueprints/app/files/.vscode/settings.json
@@ -1,0 +1,9 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/bower_components": true,
+    "dist": true,
+    "tmp": true
+  }
+}


stefanpenner:
We don't currently provide editor specific configuration, rather such configuration should be provided be an addon. Otherwise we will end up generating a dozen set of such configurations speculatively. 

Although I recognize the value, the cost of lots of unused files also quite high.

Turbo87:
I tend to agree. Creating such files should probably be the responsibility of vsc-ember-cli instead.

Tyriar:
No worries, thanks for considering the change.

stefanpenner:
@Tyriar if you have an idea to allow us to bundle these helpful files, without eagerly polluting the users file trees, im all ears.

Turbo87:
something like an `ember-cli-ide-files` addon with blueprints for the major IDEs maybe?

Tyriar:
The best thing may be to have vscode use the contents of `.gitignore` as part of `search.exclude`, that way this behavior would be the default.

stefanpenner:
@Tyriar yes totally

PR #5645, opened 2016-03-20
bump some common test timeouts, reduce intermittent failures

stefanpenner:


chadhietala:
Such timeouts.

stefanpenner:
ya, a better solution would be to adjust the timeouts if the process if still obviously working. But that is more work, maybe something we can look at in the future...

stefanpenner:
@homu r+

PR #5646, opened 2016-03-20
ensure at-least npm v3

stefanpenner:


stefanpenner:
cc @Turbo87

Turbo87:
@stefanpenner what's the reason for that?

stefanpenner:
@Turbo87 node 0.10.x installs npm v1 by default, which is riddle with bugs. Which caused this failure: https://github.com/ember-cli/ember-cli/pull/5391

kellyselden:
A couple questions from the uninformed:
Noticeable test time increase?
Included in the node_modules cache after first run?

Turbo87:
@stefanpenner could you add a comment to that line explaining what it does? I'm sure I've forgotten about it two weeks from now :wink:

stefanpenner:
> Included in the node_modules cache after first run?

yes

> Noticeable test time increase?

Yes, but I'm unsure if we have an option, as without this we are subject to npm 1.x issues in our tests. We are still going to support node 0.10.x (people like ghost, and my employer still need it).

kellyselden:
@stefanpenner Looks like the bases are covered, so I'm a :+1: after a comment.

stefanpenner:
the hit for older npm -> npm v3 appears to be about 1m. But that was without cache prime, futures runs should demonstrate this better.

Turbo87:
@homu r+

PR #5649, opened 2016-03-20
commands: Use plain strings instead of "option-type" utility

Turbo87:
Using the "option-type" utility in #5622 broke `nopt` option validation for paths (see https://github.com/npm/nopt/blob/master/lib/nopt.js#L20) and removed the `type: 'Path'` from the JSON help output.

/cc @lazybensch @stefanpenner

### DIFF
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -2,7 +2,6 @@
 
 var Command = require('../models/command');
 var win     = require('../utilities/windows-admin');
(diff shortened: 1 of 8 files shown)


stefanpenner:
I actually just noticed that, when i was rebasing  https://github.com/ember-cli/ember-cli/pull/5644

stefanpenner:
@homu r+

PR #5650, opened 2016-03-20
reset aliases for deprecated install commands

kellyselden:
Since the old install-addon command was inheriting the aliases of its parent, this happened:
```
>ember i ember-cli-mirage
DEPRECATION: This command has been deprecated. Please use `ember install <addonName>` instead.
```


stefanpenner:
Good catch

kellyselden:
@homu r+

Turbo87:
ouch... :+1:

PR #5653, opened 2016-03-21
Upgrade packages for app/addon blueprints

elwayman02:
The only material change here is `ember-cli-shims` patch version ([diff](https://github.com/ember-cli/ember-cli-shims/compare/0.1.0...v0.1.1)), all other version upgrades are within the `^` versioning already in place for these blueprints, so there should be no material difference other than updating the numbers.

All tests pass locally.

@rwjblue

### DIFF
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -1,8 +1,8 @@
 {
   "name": "<%= name %>",
   "dependencies": {
-    "ember": "~2.4.1",
-    "ember-cli-shims": "0.1.0",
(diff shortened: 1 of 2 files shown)


elwayman02:
Seems like there were a bunch of timeouts in the builds, but the Node 4 build passed. Can someone re-run these?

stefanpenner:
rebuilding, ya not sure what happened with that build.

stefanpenner:
@homu r+

stefanpenner:
thanks señor

PR #5655, opened 2016-03-21
Move "ember-cli" version printing into "ember --version" command

Turbo87:
/cc @stefanpenner

### DIFF
--- a/tests/unit/cli/cli-test.js
+++ b/tests/unit/cli/cli-test.js
@@ -79,10 +79,6 @@ afterEach(function() {
   commands = argv = ui = undefined;
 });
 
-function assertVersion(string, message) {
-  expect(/version:\s\d+\.\d+\.\d+/.test(string), message || ('expected version, got: ' + string)).to.be.true;
-}
-
 describe('Unit: CLI', function() {
   this.timeout(10000);
   it('exists', function() {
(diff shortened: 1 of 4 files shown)


kellyselden:
This removes the version printing from every command? I believe this is desired.

stefanpenner:
I think we can remove this, we used to have it to prevent common confusions. But maybe we can get rid of it...

rwjblue:
FWIW, as it stands today this console output is confusing (`version: 2.4.2`) doesn't tell me really anything (is this the current version?  of what thing? Ember? EmberCLI? etc).  I'm definitely in favor of removing this on every command...

Turbo87:
@rwjblue I've changed it to output `ember-cli: 2.4.2` on `ember --version` to be more explicit

tests/unit/cli/cli-test.js
```diff
@@ -79,10 +79,6 @@ afterEach(function() {
   commands = argv = ui = undefined;
 });
 
-function assertVersion(string, message) {
```
kellyselden: I don't think the version output is ever tested anymore after removing this, because I expected to see 'version: ' => 'ember-cli: ' in the diff.

Turbo87: fixed with 16d998f

Turbo87:
@kellyselden @stefanpenner r?

kellyselden:
@homu r+

kellyselden:
homu is thinking hard again.

kellyselden:
@homu r+

kellyselden:
@homu r+

kellyselden:
Well dang, I guess we merge the old fashioned way once CI finishes

PR #5661, opened 2016-03-21
Use "update-notifier" dependency to notify users of ember-cli updates

Turbo87:
bildschirmfoto 2016-03-21 um 23 09 52

The version will be checked at most once a day and once the notice has been shown it won't be shown until the next day either to not bother the users too much.

Also this will only check the globally installed `ember` unless it's explicitly called via `node_modules/.bin/ember` from some project outside of any NPM scripts.

(v2.0.0 was faked for the screenshot)

---

Before landing:
- [ ] Never `stderr`.
- [ ] Confirm it works behind a firewall.
- [ ] Check interaction with `--json`.

### DIFF
--- a/bin/ember
+++ b/bin/ember
@@ -7,6 +7,11 @@ process.title = 'ember';
 var resolve = require('resolve');
 var exit = require('exit');
 
+var updateNotifier = require('update-notifier');
+var pkg = require('../package.json');
+
+updateNotifier({ pkg: pkg }).notify({ defer: false });
+
 resolve('ember-cli', {
   basedir: process.cwd()
 }, function(error, projectLocalCli) {
--- a/bin/ember
+++ b/bin/ember
@@ -10,7 +10,7 @@ var exit = require('exit');
 var updateNotifier = require('update-notifier');
 var pkg = require('../package.json');
 
-updateNotifier({ pkg: pkg }).notify({ defer: false });
+notifyUpdate(updateNotifier({ pkg: pkg }).update);
 
 resolve('ember-cli', {
   basedir: process.cwd()
@@ -38,3 +38,16 @@ resolve('ember-cli', {
     exit(exitCode);
   });
 });
+
+function notifyUpdate(update) {
+  if (!process.stdout.isTTY || !update || require('is-npm')) {
+    return;
+  }
+
+  var chalk = require('chalk');
+
+  var message = '\nA new ' + chalk.yellow('ember-cli') + ' release is available (' + update.current + ' → ' + chalk.yellow(update.latest) + ')\n' +
+    'Visit ' + chalk.yellow.underline('https://github.com/ember-cli/ember-cli/releases/tag/v' + update.latest) + ' for update instructions\n';
+
+  console.error(message);
+}
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "inflection": "^1.7.0",
     "inquirer": "^0.12.0",
     "is-git-url": "^0.2.0",
+    "is-npm": "^1.0.0",
     "isbinaryfile": "^3.0.0",
     "leek": "0.0.21",
     "lodash": "^4.6.1",
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "through": "^2.3.6",
     "tiny-lr": "0.2.1",
     "tree-sync": "^1.1.0",
+    "update-notifier": "^0.6.3",
     "walk-sync": "^0.2.6",
     "yam": "0.0.19"
   },


kellyselden:
* Is it testable?
* How does it handle our beta releases, does it check @latest?
* I'm a little confused, how does it work when running in an out of date local project, does it still only check global and print the "-g"?

Turbo87:
> - Is it testable?

we might be able to mock it in some way, but I don't think it's worth it. the API is so simple you can't actually make many mistakes.

> - How does it handle our beta releases, does it check @latest?

yes, it checks `latest` from https://registry.npmjs.org/ember-cli/

> - I'm a little confused, how does it work when running in an out of date local project

the check is in the `ember` script itself, not in the `cli` folder. that means running in a local project will first invoke the global `ember` (which checks its version), and that will invoke the local `cli/index.js` which does not have a version check.

> does it still only check global and print the "-g"?

unless I misunderstood the question: yes

kellyselden:
> we might be able to mock it in some way, but I don't think it's worth it. the API is so simple you can't actually make many mistakes.

I agree probably not worth it.

> that will invoke the local cli/index.js which does not have a version check.

Sounds good.

kellyselden:
We have a home grown version checker [here](https://github.com/ember-cli/ember-cli/blob/caa24640bd986c8879291d6fd0b99c9d61fc1a86/lib/models/update-checker.js) that IIRC was more trouble than it was worth and disabled by default, so I would like to get @stefanpenner or @rwjblue approval as well on this. Perhaps a sister PR to remove that code would make sense.

bin/ember
```diff
@@ -7,6 +7,11 @@ process.title = 'ember';
 var resolve = require('resolve');
 var exit = require('exit');
 
+var updateNotifier = require('update-notifier');
```
stefanpenner: we aim to rarely (if ever) make changes to this file, as it is shared between all (old/new) ember apps on a system.

This should likely be somwhere in lib/cli

stefanpenner:
> We have a home grown version checker

this was most likely the problem

stefanpenner:
can we provide an alternative set of upgrade instructions?

Turbo87:
> We have a home grown version checker here

Oh, interesting. I didn't see that one.

> can we provide an alternative set of upgrade instructions? As our instructions can be more complicated -> https://github.com/ember-cli/ember-cli/releases/tag/v2.4.2
> 
> I think we should likely link to our release page for more information, or maybe all the instructions

that is possible. the `notify()` method just uses the default rendering but we can replace it if we want something different.

bin/ember
Turbo87: > This should likely be somwhere in lib/cli

moving it to `lib/cli` has two issues:

- it would check the version of the project-specific `ember-cli` dependency instead of the globally installed `ember` command
- if you fix that and pass the version/package info of the global command to `lib/cli` you would only run the version check for the global command if the project-specific `ember-cli` was `^2.5.0` which doesn't make much sense IMHO

that is essentially why I put it there instead of in `lib/cli`, otherwise I fully agree that changes should usually happen there instead (e.g auto completion)

stefanpenner: Yes, that makes sense. I think this is the correct direction now, we just need to make sure the UX around this isn't confusing.

Turbo87:
@stefanpenner I'm open to suggestions for the alternate update message and layout

btw what's the reason for the complicated update instructions?

stefanpenner:
> btw what's the reason for the complicated update instructions?

they have grown out of issues we have seen, bugs reports, and confusion. This doesn't mean we can't refactor them though.

Turbo87:
Updated the notification styling and text:

screenshot

Turbo87:
@stefanpenner @kellyselden any opinions on the new proposal?

kellyselden:
I think I'm a little confused. This message only triggers for the globally run instances, but the upgrade instructions in our tagged releases tells you how upgrade a local install.

Turbo87:
@kellyselden on https://github.com/ember-cli/ember-cli/releases/tag/v2.4.2 under "Setup" it says `npm install -g ember-cli@2.4.2` which is the global instance

kellyselden:
@Turbo87 Ah I should have read it instead of going from memory. I am :+1: on this.

stefanpenner:
Im liking this. But would like to try it out before I :+1: , our last update notifier left something thing to be desired.

Turbo87:
@stefanpenner you can try it like this:

- set `version` in `package.json` to e.g. `2.0.0`
- adjust https://github.com/yeoman/update-notifier/blob/master/index.js#L36-L38 to `0` instead of `Date.now()`
- `rm ~/.config/configstore/update-notifier-ember-cli.json` if it exists
- run `ember version` a few times until the notification pops up

Turbo87:
@stefanpenner did you have a chance to try it out?

stefanpenner:
@Turbo87 not yet (been swamped with ember-conf) but added it to my queue for today.

Turbo87:
@stefanpenner cool. let me know if you need any help.

Turbo87:
@stefanpenner ping :wink:

bin/ember
```diff
@@ -33,3 +38,16 @@ resolve('ember-cli', {
...
+  if (!process.stdout.isTTY || !update || require('is-npm')) {
+    return;
+  }
+
+  var chalk = require('chalk');
+
+  var message = '\nA new ' + chalk.yellow('ember-cli') + ' release is available (' + update.current + ' → ' + chalk.yellow(update.latest) + ')\n' +
+    'Visit ' + chalk.yellow.underline('https://github.com/ember-cli/ember-cli/releases/tag/v' + update.latest) + ' for update instructions\n';
+
+  console.error(message);
```
nathanhammond: Logging to stderr seems dangerous, especially if running on a build system. Could confuse a test runner.

Turbo87: there is `if (!process.stdout.isTTY)` above which should filter out most CI systems. also this is for the `ember` executable which you shouldn't use in tests anyway.

nathanhammond:
:+1: on the concept, but:

- We have people running builds behind firewalls with no network access. How does update-notifier deal with that?
- How does update-notifier deal with internal artifact systems?
- What does this do to our generic command performance?

bin/ember
nathanhammond: I'm not 100% sure that our test setup at work doesn't shell out to `ember test` since we mate a whole bunch of commands to run our tests and rely on `&&` to join them. `npm run test` could also blow up with an error code exit.

Turbo87:
> We have people running builds behind firewalls with no network access. How does update-notifier deal with that?

it forks itself to a different process to do the actual update check so it would just fail silently and not display anything

> How does update-notifier deal with internal artifact systems?

it checks the official NPM registry and won't work with internal system as far as I'm aware

> What does this do to our generic command performance?

I haven't explicitly measured it, but as mentioned above it basically forks itself out of the process and does the update check there, so I'm expecting no significant performance degradation.

to be honest all the `require()` calls (600ms here) most likely take much longer than the HTTP request to the NPM registry :wink:

nathanhammond:
Let's table for discussion at our next weekly meeting.

bin/ember
rwjblue: stderr should only be used for actual errors (which this is definitely not). We should also ensure that we honor the global --silent flag to squelch all but required output.

sindresorhus:
> it checks the official NPM registry and won't work with internal system as far as I'm aware

It should in theory support internal registries, as `update-notifier` => `latest-version` => [`package-json`](https://github.com/sindresorhus/package-json) reads the local `.npmrc`, but it's untested.

-

Open issues if there's anything that could be improved with `update-notifier`.

bin/ember
nathanhammond: Check JSON output.

kellyselden:
We should look into how this plays with our --json output

stefanpenner:
> it forks itself to a different process to do the actual update check so it would just fail silently and not display anything

in node if the child process times out, would that cause the parent to hang around until the child  also exists?

Turbo87:
AFAIK only if you *don't* call `unref()` which I think this lib does.

stefanpenner:
> AFAIK only if you don't call unref() which I think this lib does.

:+1:

bin/ember
stefanpenner: > We should also ensure that we honor the global --silent flag to squelch all but required output.

👍

Turbo87:
closing this for now. I'm no longer sure if this is really the right approach...

PR #5673, opened 2016-03-23
Fixes leading slash being added to component-test module name

### LINKED ISSUE #5518: component-test blueprint adds '/' in front of the component name
Using pods.

If I type `ember g component-test test` the test is generated incorrectly with the component name having a '/' added in front.

    import { moduleForComponent, test } from 'ember-qunit';
    import hbs from 'htmlbars-inline-precompile';

    moduleForComponent('/test', 'Integration | Component | test', {
      integration: true
    });

    test('it renders', function(assert) {
      // Set any properties with this.set('myProperty', 'value');
      // Handle any actions with this.on('myAction', function(val) { ... });"

      this.render(hbs`{{/test}}`);

      assert.equal(this.$().text().trim(), '');

      // Template block usage:"
      this.render(hbs`
        {{#/test}}
          template block text
        {{//test}}
      `);

      assert.equal(this.$().text().trim(), 'template block text');
    });

lpaluszk:
Fixes #5518

### DIFF
--- a/blueprints/component-test/index.js
+++ b/blueprints/component-test/index.js
@@ -42,7 +42,7 @@ module.exports = {
     var testType = options.testType || 'integration';
     var friendlyTestDescription = testInfo.description(options.entity.name, 'Integration', 'Component');
 
-    if (options.pod && options.path !== 'components' && options.path !== '') {
+    if (options.pod && options.path && options.path !== 'components') {
       componentPathName = [options.path, dasherizedModuleName].join('/');
     }
 


stefanpenner:
@trabus @Turbo87 we likely need to port this to the new externalized helpers

Turbo87:
@lpaluszk good catch! could you open similar PRs for the following two projects too? 

- https://github.com/emberjs/ember.js/blob/master/blueprints/component-test/index.js
- https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/component-test/index.js

Turbo87:
@homu r+

PR #5679, opened 2016-03-24
"app" blueprint: Update TravisCI config

Turbo87:
This PR updates the TravisCI config file in the `app` blueprint:

- npm@2 not installed explicitly
- "npm i bower" in before_install block
- PhantomJS installed via npm
- Unnecessary quotes removed
- Node v4.x instead of 0.12.x

### DIFF
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "0.12"
+  - "4"
 
 sudo: false
 
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -12,7 +12,6 @@ cache:
 before_install:
   - export PATH=/usr/local/phantomjs-2.0.0/bin:$PATH
   - "npm config set spin false"
-  - "npm install -g npm@^2"
 
 install:
   - npm install -g bower
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -10,7 +10,7 @@ cache:
     - node_modules
 
 before_install:
-  - "npm config set spin false"
+  - npm config set spin false
   - npm install -g bower
   - npm install phantomjs-prebuilt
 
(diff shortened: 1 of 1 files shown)


stefanpenner:
:+1:

stefanpenner:
@homu r+

mike-north:
FWIW, It's now easy for addon authors to introduce problems for node 0.12 users. For example, if they use enhanced object literals in `ember-cli-build.js`, everything will appear to be fine in CI since it is run against node 4. I've tripped over my own shoelaces on this multiple times since ember-cli 2.5 was released.

It's worthwhile to help (and encourage) addon authors to maintain compatibility (when possible) for node 0.12 and all other officially supported node versions that the ember ecosystem is supposed to work with. Particularly in corporate orgs, there may be officially blessed versions of node that are allowed in production, and those may not include recent releases (looking at you @stefanpenner ;) )

We should consider adding some stuff to the addon blueprint `index.js` to alter `.travis.yml` so node 0.12 is used for CI.

nathanhammond:
@mike-north Your concerns are legitimate. Can you open a PR with 0.12 added back and we'll have the discussion there? Note that regardless of the conclusion in that discussion we _will_ be dropping support for 0.10 and 0.12 on their respective end of support dates (October 1 2016 and April 1 2017, respectively).

nathanhammond:
Update: [it appears that the Node LTS schedule has changed](https://github.com/nodejs/LTS/pull/104). This means that we will drop support for 0.12 on December 31 2016.

As an aside, was there any discussion on this LTS schedule in public?

/cc @ErisDS

mike-north:
@nathanhammond I don't see any problem with apps using node v4, so my thinking is to leave that as-is. 

To be clear, this has nothing to do with a debate as to **when** support for 0.10/12 is dropped. It has not been dropped yet, and although we're all rightfully chomping at the bit to move forward, we need to support it (and help encourage the ecosystem to support it) until that day comes.

Some SemVer discussion needs to be had around this change IMO, because things that used to work will potentially break once this transition happens -- potentially it's addons themselves that need to major-version-incremented when starting to use stuff that's not compatible with node 0.x.

rwjblue:
It is also possible to deal with this at a linting level.  There is at least one eslint plugin that attempts to catch things that are not valid on a given node version (I have used [this one](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md)), this technique would allow us to encourage best practices by specifying supported version while still taking advantage of recent node engines. I personally would prefer to use node 4 or 6 in CI simply due to the massively improved error messages...

PR #5689, opened 2016-03-27
[ENHANCEMENT] Added return statement to `beforeEach` in `moduleForAcceptance`

Fed03:
In order to make use of QUnit ability of handling async tasks in `beforeEach` I added the proper return statement to `moduleForAcceptance`

I left `afterEach`untouched because of #5348, but maybe we could do the same thing.

### DIFF
--- a/blueprints/app/files/tests/helpers/module-for-acceptance.js
+++ b/blueprints/app/files/tests/helpers/module-for-acceptance.js
@@ -8,7 +8,7 @@ export default function(name, options = {}) {
       this.application = startApp();
 
       if (options.beforeEach) {
-        options.beforeEach.apply(this, arguments);
+        return options.beforeEach.apply(this, arguments);
       }
     },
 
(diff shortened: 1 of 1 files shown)


stefanpenner:
> I left afterEachuntouched because of #5348, but maybe we could do the same thing.

can you add ` afterEach` as well?

@trentmwillis r?

trentmwillis:
:+1: this is good and to adding it to `afterEach` as well. You could use `assert.async`, but Promises are better IMO.

Fed03:
I do the same in `afterEach` too.
The code is gross in my opinion but it works. If you find a better way let me know

trentmwillis:
@Fed03 I think you'd actually want to wait to run `destroyApp` until after the `afterEach` has finished running. So something like:

```js
let afterEach = options.afterEach && options.afterEach.apply(this, arguments);
return Promise.resolve(afterEach).then(() => destroyApp(this.application));
```

Fed03:
@trentmwillis You're right, done

Fed03:
I really don't understand why tests are failing....

rwjblue:
I restarted the node5 build.

stefanpenner:
@homu r+

PR #5701, opened 2016-03-28
Add test to check for project name collisions

ryanlabouve:
During Ember contributor workshop, wrote a test to help us explore a related issue.

Add test to specifically test for creating a new ember project with a name that conflicts with an existing folder in the current working directory.

e.g.
```
cd ~/
mkdir foo
ember new foo
```
should fail.

This test turned out not to be strictly required for the issue, but we still felt like it added value.

/cc @lukemelia


rwjblue:
LGTM, thank you!

stefanpenner:
:+1:

Issue #5079, opened 2015-11-12
restore disabled tests


stefanpenner:
something sideways on test cleanup...

https://github.com/ember-cli/ember-cli/commit/d813ceda3d12f11e9baf58046f070792fc250173

PR #5706, opened 2016-03-30
Uncommenting test

twokul:
fixes https://github.com/ember-cli/ember-cli/issues/5079

### DIFF
--- a/tests/acceptance/smoke-test-slow.js
+++ b/tests/acceptance/smoke-test-slow.js
@@ -119,19 +119,18 @@ describe('Acceptance: smoke-test', function() {
       });
   });
 
-  // TODO: re-enable, something is funky with test cleanup...
-  // it('ember test exits with zero when tests pass', function() {
-  //   return copyFixtureFiles('smoke-tests/passing-test')
-  //     .then(function() {
-  //       return ember(['test'])
-  //         .then(function(result) {
-  //           expect(result.code).to.equal(0);
-  //         })
-  //         .catch(function() {
-  //           expect(false, 'should NOT have rejected with a failing test');
-  //         });
-  //     });
-  // });
+  it('ember test exits with zero when tests pass', function() {
+    return copyFixtureFiles('smoke-tests/passing-test')
+      .then(function() {
+        return ember(['test'])
+          .then(function(result) {
+            expect(result.exitCode).to.equal(0);
+          })
+          .catch(function() {
+            expect(false, 'should NOT have rejected with a failing test').to.be.ok;
+          });
+      });
+  });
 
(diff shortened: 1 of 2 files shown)


Turbo87:
looks like the test is actually failing for some reason

twokul:
@Turbo87 yeah, looking into it

tests/acceptance/smoke-test-slow.js
```diff
@@ -119,19 +119,18 @@ describe('Acceptance: smoke-test', function() {
...
-  //         })
-  //         .catch(function() {
-  //           expect(false, 'should NOT have rejected with a failing test');
-  //         });
-  //     });
-  // });
+  it('ember test exits with zero when tests pass', function() {
+    return copyFixtureFiles('smoke-tests/passing-test')
+      .then(function() {
+        return runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'test')
```
Turbo87: why is `ember(['test'])` not working correctly here?

stefanpenner:
The point of the ember helper is to invoke programmatically. The test was uncommented so we would eventually find out what underlying bug is preventing it from working. Changing this to be just an exec, side-steps the goal.

nathanhammond:
We have something that is tremendously trolling us here and shows up in multiple issues. I've spent days trying to track it down to no avail, worked with @stefanpenner to no avail, and I've run out of energy to invest in this. This is likely the same thing that is blocking the multi meta modules PRs.

stefanpenner:
Good job tracking this done, i believe though we should fix this in testem.

rwjblue:
I'm unsure if this is anything but a test concern. Normally, we would want to follow normal require rules for testem.js (and we do not support changes to testem.js at runtime and restart anyways). 

It seems completely reasonable for testem to be requiring testem.js (FWIW we also require ember-cli-build.js exactly the same way as testem is requiring its own configuration).

stefanpenner:
> I'm unsure if this is anything but a test concern.

yes, but it would be better if we didn't have to remember in each test that potentially hits it. IMHO

rwjblue:
Confirm. Maybe we can clear in a general place that is fired for each test (for example: perhaps the ember helper could do this on cleanup when the command is 'test').

stefanpenner:
@rwjblue not a bad idea

twokul:
@rwjblue @stefanpenner I can change `ember` helper then

stefanpenner:
@twokul sounds good, thanks @rwjblue for the idea

Turbo87:
@stefanpenner @rwjblue r?

stefanpenner:
@homu r+

stefanpenner:
good catch @twokul

Issue #5482, opened 2016-02-17
add .github/{ISSUE_TEMPLATE,CONTRIBUTING}.md


stefanpenner:
https://github.com/blog/2111-issue-and-pull-request-templates

We will need to explore exactly what the ISSUE_TEMPLATE will suggest, but this is a good step in the right direction.

kellyselden:
:+1:

Turbo87:
FWIW the rust team decided against using those kinds of templates in https://github.com/rust-lang/rust/pull/31732, though I don't know if their reasons are applicable to our case as well

stefanpenner:
Thanks for the link, will review

stefanpenner:
Looks like we could benefit as long as the structure was extremely light. We should experiment and iterate with it, I think this could still be valuable, although the rust thread outlines the pitfalls we should avoid

kellyselden:
My thoughts are one of the most beneficial things to help us not waste time triaging is a demo repo and steps to reproduce. If a template nudged people to do this, I think it would make our lives better.

stefanpenner:
> If a template nudged people to do this, I think it would make our lives better.

my thoughts as well.

Turbo87:
another example: https://github.com/square/retrofit/blob/master/.github/ISSUE_TEMPLATE.md

kellyselden:
I like the mention of StackOverflow.

Turbo87:
https://www.talater.com/open-source-templates/ :joy:

Turbo87:
resolved by #5707

PR #5707, opened 2016-03-30
Issue template

twokul:
https://github.com/ember-cli/ember-cli/issues/5482

to start the conversation and refine issue template text

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,5 @@
+Thank you for taking the time to open an issue!
+
+For bug reports please include the following section in your issue details.
+If you include instructions on how to reproduce the bug or a failing test case
+it will make it easier for us to track down the issue you're having.


stefanpenner:
I would love to explore this, and find the most minimal subset that improves the quality of issues/pr's.

nathanhammond:
Hey @twokul! We discussed this in our team meeting today and we absolutely want this. But! We want to make sure it's as easy as possible. In order to limit the amount of work by a user we want to limit this to running one command, so we need to make it count. `ember version --verbose` comes close, but we may want to extend it to also capture things like `dependencies`, `peerDependencies`, and `devDependencies` of the currently active project as well as bower.json. That can be a secondary effort.

The other thing is that having exactly one issue template is pretty limiting as not all issues will manifest in a way that the template makes sense. We need to make sure that something inside of the template communicates its optional status.

For now, though, something like this might achieve our goals, but I'd love for you to help us wordsmith it better:

> Thank you for taking the time to open an issue! If you believe you've found a bug please include this section in your issue details. Including a link reproduction or failing test will make it easier for us to track down the issue you're having.
>
> ---
> Output from `ember version --verbose`:
> \`\`\`
> [Replace this line with the output.]
> \`\`\`

twokul:
@nathanhammond updated

nathanhammond:
@ember-cli/core Seem good?

kellyselden:
I agree it is now an easy, straight-forward template. Some areas it doesn't cover that I can think of:
* Build is really slow, please follow our perf guide first.
* How do I do this, post to stack overflow.
* I want this feature, follow the RFC process.

Are these common posts worth pointing out to cut back on issue triage, or not worth the confusion to new users? Can we word it well enough so it meets somewhere in the middle?

chadhietala:
We should point to the [perf guide](https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md) for "My build is slow".

Turbo87:
> Are these common posts worth pointing out to cut back on issue triage, or not worth the confusion to new users? Can we word it well enough so it meets somewhere in the middle?

I'd prefer to keep the template as short as possible. The main pain point right now seems to be that people don't include versions and reproductions in their bug reports which is solved by the proposal.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+Thank you for taking the time to open an issue! If you believe you've found a bug
+please include this section in your issue details. Including a link reproduction
+or failing test will make it easier for us to track down the issue you're having.
```
Turbo87: > Thank you for taking the time to open an issue!
>
> For bug reports please include the following section in your issue details. If you include instructions on how to reproduce the bug or a failing test case it will make it easier for us to track down the issue you're having.

opinions?

nathanhammond: I prefer @Turbo87's to mine.

nathanhammond:
@kellyselden @chadhietala I think I'd rather start with less information and add more as we notice new patterns. Also, Chad, maybe your `ember doctor` stuff would be worthwhile to bring back.

/cc @sivakumar-kailasam

sivakumar-kailasam:
Its up-to-date with master with fixed, passing tests at https://github.com/sivakumar-kailasam/ember-cli/tree/ember-doctor. I'll raise a PR for it.

I also wanted to extend the `ember doctor` rfc by adding a `--report` flag. So that we can open a partially filled issue template in the browser making it easier for folks to fill in the specifics while we take care of the rest. https://github.com/ember-cli/ember-cli/issues/new?title=My%20ember-cli%20issue&body=%0Anpm:%203.0.0%0Anode:%205.1.1%0Aos:%20darwin%0A%0Amore%20info%20from%20ember-doc%0A%0A%3Cfill%20more%20desc%20here%3E%0A%0A

Does this sound like something worth exploring?

Turbo87:
@nathanhammond this looks good to me now. r+ from my side

nathanhammond:
@sivakumar-kailasam Yes, the `--report` feature sounds wonderful. Let's make that a thing.

@homu r+

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Thank you for taking the time to open an issue!
+
+For bug reports please include the following section in your issue details.
```
joostdevries: what section?

Turbo87: good point, somehow that second part of @nathanhammond suggestion got lost 😕

nathanhammond: Good catch. Hold on.

PR #5708, opened 2016-03-30
Add blueprint prompts

knownasilya:
See [RFC #7](https://github.com/ember-cli/rfcs/pull/7) for more details.
- [x] Add `availableOptions.prompt` processing
- [x] Add `--skip-prompts`
- [x] Fix tests
- [ ] Add tests

### DIFF
--- a/lib/commands/generate.js
+++ b/lib/commands/generate.js
@@ -18,6 +18,7 @@ module.exports = Command.extend({
 
   availableOptions: [
     { name: 'dry-run',       type: Boolean, default: false, aliases: ['d'] },
+    { name: 'skip-prompts',  type: Boolean, default: false, aliases: ['sp'] },
     { name: 'verbose',       type: Boolean, default: false, aliases: ['v'] },
     { name: 'pod',           type: Boolean, default: false, aliases: ['p'] },
     { name: 'classic',       type: Boolean, default: false, aliases: ['c'] },
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -849,27 +849,27 @@ Blueprint.prototype._locals = function(options) {
   var moduleName = options.entity && options.entity.name || packageName;
   var sanitizedModuleName = moduleName.replace(/\//g, '-');
 
-  return this._processPrompts(options).then(function (prompts) {
-    return new Promise(function(resolve) {
-      resolve(this.locals(prompts));
-    }.bind(this)).then(function (customLocals) {
-      var fileMapVariables = this._generateFileMapVariables(moduleName, customLocals, options);
-      var fileMap = this.generateFileMap(fileMapVariables);
-      var standardLocals = {
-        dasherizedPackageName: stringUtils.dasherize(packageName),
-        classifiedPackageName: stringUtils.classify(packageName),
-        dasherizedModuleName: stringUtils.dasherize(moduleName),
-        classifiedModuleName: stringUtils.classify(sanitizedModuleName),
-        camelizedModuleName: stringUtils.camelize(sanitizedModuleName),
-        decamelizedModuleName: stringUtils.decamelize(sanitizedModuleName),
-        fileMap: fileMap,
-        hasPathToken: this.hasPathToken,
-        targetFiles: options.targetFiles,
-        rawArgs: options.rawArgs
-      };
+  return new Promise(function(resolve) {
+    return resolve(this._processPrompts(options));
+  }.bind(this)).then(function (prompts) {
+    return this.locals(prompts);
+  }.bind(this)).then(function (customLocals) {
+    var fileMapVariables = this._generateFileMapVariables(moduleName, customLocals, options);
+    var fileMap = this.generateFileMap(fileMapVariables);
+    var standardLocals = {
+      dasherizedPackageName: stringUtils.dasherize(packageName),
+      classifiedPackageName: stringUtils.classify(packageName),
+      dasherizedModuleName: stringUtils.dasherize(moduleName),
+      classifiedModuleName: stringUtils.classify(sanitizedModuleName),
+      camelizedModuleName: stringUtils.camelize(sanitizedModuleName),
+      decamelizedModuleName: stringUtils.decamelize(sanitizedModuleName),
+      fileMap: fileMap,
+      hasPathToken: this.hasPathToken,
+      targetFiles: options.targetFiles,
+      rawArgs: options.rawArgs
+    };
 
-      return merge({}, standardLocals, customLocals);
-    }.bind(this));
+    return merge({}, standardLocals, customLocals);
   }.bind(this));
 };
 
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -880,7 +880,7 @@ Blueprint.prototype._locals = function(options) {
   @return {Promise}
 */
 Blueprint.prototype._processPrompts = function (options) {
-  if (this.ui.ci || options.skipPrompts) {
+  if (options.skipPrompts) {
     return Promise.resolve(options);
   }
 
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -881,7 +881,7 @@ Blueprint.prototype._locals = function(options) {
 */
 Blueprint.prototype._processPrompts = function (options) {
   if (this.ui.ci || options.skipPrompts) {
-    return Promise.resolve({});
+    return Promise.resolve(options);
   }
 
   var availableOptions = this.availableOptions;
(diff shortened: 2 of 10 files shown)


tests/unit/models/blueprint-test.js
```diff
@@ -499,6 +499,7 @@ help in detail');
         .then(function() {
           var actualFiles = walkSync(tmpdir).sort();
           var output = ui.output.trim().split(EOL);
+          debugger;
```
Turbo87: I don't think that's supposed to stay here, right? :wink:

tests/unit/models/blueprint-test.js
```diff
@@ -1989,7 +1990,7 @@ help in detail');
     });
   });
 
-  describe('._locals', function() {
+  describe.only('._locals', function() {
```
Turbo87: and this neither

lib/models/blueprint.js
```diff
@@ -849,29 +849,80 @@ Blueprint.prototype._locals = function(options) {
...
+        decamelizedModuleName: stringUtils.decamelize(sanitizedModuleName),
+        fileMap: fileMap,
+        hasPathToken: this.hasPathToken,
+        targetFiles: options.targetFiles,
+        rawArgs: options.rawArgs
+      };
+
+      return merge({}, standardLocals, customLocals);
+    }.bind(this));
   }.bind(this));
```
Turbo87: why are you nesting the promises here instead of chaining them?

lib/models/blueprint.js
```diff
@@ -849,29 +849,80 @@ Blueprint.prototype._locals = function(options) {
...
 };
 
 /**
+  @private
+  @method _processPrompts
+  @param {Object} options
+  @return {Promise}
+*/
+Blueprint.prototype._processPrompts = function (options) {
+  if (this.ui.ci || options.skipPrompts) {
```
Turbo87: I'm not sure if we should do this by default on CI systems. IMHO it should be something explicit.

lib/models/blueprint.js
knownasilya: Chaining as in 

```js
return this._processPrompts(options).then(function (prompts) {
 return this.locals(prompts);
})
```

I have the wrapping `new Promise` per @stefanpenner request for preventing errors from being uncaught I think.

lib/models/blueprint.js
knownasilya: Well it is, you can set it with `CI=true ember g my-addon`

Turbo87: what I meant was I prefer explicitly using `ember g foo bar --skip-prompts` on TravisCI etc. instead of implying the option automatically when `CI=true`

knownasilya:
Looks like the travis build needs to be restarted.

lib/models/blueprint.js
knownasilya: Yeah, maybe. Just feel like people will forget, but maybe that's alright.

lib/models/blueprint.js
Turbo87: what I meant was this instead:

```js
return new Promise(function(resolve) {
    resolve(this._processPrompts(options));
}.bind(this)).then(function (prompts) {
    return this.locals(prompts);
}.bind(this)).then(function (customLocals) {
    var fileMapVariables = this._generateFileMapVariables(moduleName, customLocals, options);
    var fileMap = this.generateFileMap(fileMapVariables);
    var standardLocals = {
        dasherizedPackageName: stringUtils.dasherize(packageName),
        classifiedPackageName: stringUtils.classify(packageName),
        dasherizedModuleName: stringUtils.dasherize(moduleName),
        classifiedModuleName: stringUtils.classify(sanitizedModuleName),
        camelizedModuleName: stringUtils.camelize(sanitizedModuleName),
        decamelizedModuleName: stringUtils.decamelize(sanitizedModuleName),
        fileMap: fileMap,
        hasPathToken: this.hasPathToken,
        targetFiles: options.targetFiles,
        rawArgs: options.rawArgs
    };

    return merge({}, standardLocals, customLocals);
}.bind(this));
```

lib/models/blueprint.js
Turbo87: what use-case do you have for running `ember generate` on a CI system?

lib/models/blueprint.js
knownasilya: I think `locals` needs the `resolve`, not the prompt processor.

lib/models/blueprint.js
knownasilya: Good point. Think I should just remove it and if people later come back with a use-case, add it then?

lib/models/blueprint.js
Turbo87: why do you think that?

the `new Promise()` wrapping code was previously used to catch errors in `this.locals()`, but any code run in a `.then()` will automatically get caught and passed down the promise chain.

In my proposal I'm using the same trick to catch errors in your new `this._processPrompts()` method. Although since that method is supposed to be private I'm not sure it's even needed.

lib/models/blueprint.js
Turbo87: :+1:  sounds good to me

lib/models/blueprint.js
knownasilya: Will update that, since I didn't have it wrapped before.

lib/models/blueprint.js
```diff
@@ -850,7 +850,9 @@ Blueprint.prototype._locals = function(options) {
   var sanitizedModuleName = moduleName.replace(/\//g, '-');
 
   return new Promise(function(resolve) {
-    resolve(this.locals(options));
+    return resolve(this._processPrompts(options));
+  }.bind(this)).then(function (prompts) {
```
Turbo87: TBH I'd actually prefer a different syntax here but @stefanpenner will probably be able to tell if it has any disadvantages:

```js
return Promise.resolve().then(function() {
  return this._processPrompts(options);
}.bind(this)).then(function(options) {
  return this.locals(options);
}.bind(this)).then(/* ... */);
```

knownasilya: I feel like that's a hacky way to do it, almost an unnecessary step just to not use `new`.

Turbo87: I was actually surprised to see that an exception inside the `new Promise()` callback will reject the promise instead of propagating to the calling context. IMHO it looks a bit cleaner that way, but I'm fine with going either way.

stefanpenner: I prefer the constructor form, but maybe our promise subclass whats to add `Promise.invoke(context, method, [args])` ?

lib/commands/generate.js
```diff
@@ -18,6 +18,7 @@ module.exports = Command.extend({
 
   availableOptions: [
     { name: 'dry-run',       type: Boolean, default: false, aliases: ['d'] },
+    { name: 'skip-prompts',  type: Boolean, default: false, aliases: ['sp'] },
```
stefanpenner: i think other tools often refer to this as interactive or interactive=false, thoughts?

Turbo87: there's also `-y` and `--yes` in e.g. `apt`

stefanpenner: also good options

knownasilya: @Turbo87 how would the `--yes` apply to multiple yes/no prompts?

@stefanpenner I think it's less descriptive, and you have to set it explicitly, i.e. `--skip-prompts` vs `--interactive=false`.

Turbo87: > @Turbo87 how would the --yes apply to multiple yes/no prompts?

that is exactly why I hadn't suggested `--yes` before. in this case something like `--use-defaults` probably makes more sense unless all the defaults are `yes` anyway.

knownasilya: `--skip-prompts` would just use the defaults.

stefanpenner: im fine with --skip-prompts, just wanted us to explore some other ideas to be sure.

lib/models/blueprint.js
kellyselden: Constructor form feels more natural to me too.

stefanpenner: An alternative, which Im totally on-board with is adding. `Promise.invoke` to our promise subclass

```
Promise.invoke(this._processPrompts.bind(this), options).then
```

which would likely be implemented as:

```js
function invoke(method, ...args) {
return new this(function(resolve) {
  resolve(method(...args));
});
```

stefanpenner:
@knownasilya any cycles?

knownasilya:
@stefanpenner sure, what's up?

stefanpenner:
@knownasilya am I blocking completion of this or..? It seems like the above comment had a good resolution, im curious whats next or just rebase + test fixes

knownasilya:
Rebase and tests sound good. Any suggestions on what kind of tests would be most helpful? And maybe a pointer to similar tests in the code base.

knownasilya:
Ping @stefanpenner

Turbo87:
closing due to inactivity

PR #5709, opened 2016-03-31
Move test-loader to NPM

trentmwillis:
Pretty self-explanatory. Was unsure if this needed to be 100% backwards compatible (e.g., check if `ember-cli-test-loader` is installed as a bower or npm module and behave accordingly), so I took the simple route which relies on users upgrading their dependencies (addons like `ember-cli-qunit` will need to be updated accordingly, which I will be working on).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1310,7 +1310,6 @@ EmberApp.prototype.styles = function() {
  */
 EmberApp.prototype.testFiles = function(coreTestTree) {
   var testSupportPath = this.options.outputPaths.testSupport.js;
-  var testLoaderPath = this.options.outputPaths.testSupport.js.testLoader;
 
   testSupportPath = testSupportPath.testSupport || testSupportPath;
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@trentmwillis looks like there are a few failing tests due to this change. could you have a look?

trentmwillis:
I believe all the failures are due to this:

> Could not find module `ember-cli/test-loader` imported from `(require)` at http://localhost:7357/assets/vendor.js

I saw this with my local testing. The module now lives at `ember-cli-test-loader/test-support/index`, so `ember-cli-qunit` will need to be updated in order to get these passing. However, since the test-loader stuff is hardcoded currently, `ember-cli-qunit` tests will fail until this PR lands. So not sure how to proceed, the cyclic dependency is difficult to get around.

nathanhammond:
Cyclic dependency just means that we have coupling between `ember-cli` and `ember-cli-qunit`. That coupling is only in generated applications, so I'm comfortable with it (ember-cli itself isn't tied). This is another argument for extracting blueprints into an addon as well as refactoring `lib/broccoli/ember-app`.

I'm in favor of landing the [`ember-cli-qunit` change](https://github.com/ember-cli/ember-cli-qunit/pull/113) first (with broken tests) and then rerunning tests here before landing. @rwjblue @stefanpenner do you have any other thoughts on how to land this which wouldn't require something landing "broken"? (`require` pun intended.)

nathanhammond:
> I took the simple route which relies on users upgrading their dependencies

Note for me and @rwjblue: this needs to be documented in the changelog as part of the release process.

nathanhammond:
@rwjblue Thoughts?

trentmwillis:
So what's the plan here? Should I try to make this compatible with older versions of the test-loader as well?

Turbo87:
IMHO we should try to find a way to make ember-cli-qunit work both ways, then apply the same to ember-cli-mocha and only then merge this.

trentmwillis:
Okay, I think have a way to make it work. Will update later today.

rwjblue:
Awesome, thank you @trentmwillis

trentmwillis:
Updated this to check if `ember-cli-test-loader` is a Bower dependency, if it is it logs a deprecation but proceeds like it used to.

tests/unit/models/asset-size-printer-test.js
```diff
@@ -20,7 +20,6 @@ describe('models/asset-size-printer', function () {
     fs.writeFileSync(path.join(assetDir, 'some-project.css'), 'body { margin: 0 20px; }', {encoding: 'utf8'});
     fs.writeFileSync(path.join(assetDir, 'some-project.js'), 'module.exports = function () {};', {encoding: 'utf8'});
     fs.writeFileSync(path.join(assetDir, 'some-project.json'), 'module.exports = function () {};', {encoding: 'utf8'});
-    fs.writeFileSync(path.join(assetDir, 'test-loader.js'), 'module.exports = function () {};', {encoding: 'utf8'});
```
Turbo87: I think we should keep this to test the old bower case

trentmwillis:
Okay, I rebased this. Unfortunately I don't see an appropriate way to test these changes until a new release of `ember-cli-qunit` is out. I think the way forward should be:

1. Merge this change
2. Update ember-cli/ember-cli-qunit#113 to test against ember-cli#master,
3. Merge ember-cli/ember-cli-qunit#113
4. Release `ember-cli-qunit`
5. Update Ember-CLI app blueprint to use the new version of ember-cli-qunit and test-loader and update test cases

Does this make sense?

stefanpenner:
> Does this make sense?

Yup this sounds good to me. Thank you for thinking this through.

stefanpenner:
@homu r+

PR #5712, opened 2016-03-31
Update "mocha" and retry failed tests a second and third time

Turbo87:
This PR updates the `mocha` test runner to v2.4.5 and enables retrying of failed tests. `retries: 2` means that the test will be executed at most three times (first try and two retries).


stefanpenner:
@homu r+

Turbo87:
> Nice, i suspect the retry support will help ease some of the intermittent failures.

I hope so... although if they mess around with the file system and leave it in some inconsistent state we might not be as lucky as we hope to be.

stefanpenner:
Ya, its a band aide... We will continue to improve the stability of our tests alongside this.

PR #5723, opened 2016-04-01
Embed/include ember-try in ember-cli

kategengler:
Also:

- Remove ember-try from the addon package.json blueprints
- Update command set as the default test script for addons to `ember try:each` from `ember try:testall` (functionally equivalent).

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -37,9 +37,8 @@ module.exports = {
     // add `ember-disable-prototype-extensions` to addons by default
     contents.devDependencies['ember-disable-prototype-extensions'] = '^1.1.0';
 
-    // add `ember-try` to addons by default
-    contents.devDependencies['ember-try'] = '^0.2.2';
-    contents.scripts.test = 'ember try:testall';
+    // use `ember-try` as test script in addons by default
+    contents.scripts.test = 'ember try:each';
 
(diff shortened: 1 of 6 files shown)


rwjblue:
LGTM

chadhietala:
@homu r+

Turbo87:
do I understand it correctly that `ember-try` can't be updated independent from `ember-cli` anymore now? in other words: if `ember-try` gains any new features you can't use them unless you update `ember-cli` too? does that have any negative implications for our users?

kellyselden:
@Turbo87 semi true. Right now the hint will give us bugfixes, and once we 1.0 ember-try, which should be soon, then it will float until 2.0.
cc @kategengler

kategengler:
@Turbo87 @kellyselden Yep, but on the upside, `ember-try` commands will now be in every `ember-cli` install.

Turbo87:
since `ember-try` is now available for apps too, should we change the app blueprint Travis config to test with beta/canary like the addon blueprint does?

kategengler:
The apps would also need a `config/ember-try.js` generated to do that. 

None of the apps I have worked on have used Travis (they have used other CI services). 

Apps might run their CI build after every commit, sometimes dozens of times a day; it might be wasteful to run both canary and beta on every build.

PR #5721, opened 2016-04-01
Fix random races in tests.

rwjblue:
You are welcome.

PR #5724, opened 2016-04-01
[WIP] Speed up slow smoke tests

trabus:
Moved tests into smoke-tests to prevent duplicating createTestTarget between multiple smoke test files


rwjblue:
This looks pretty good, but I doubt we will get the speed boost that I got in https://github.com/ember-cli/ember-cli/pull/5721...

tests/acceptance/smoke-test-slow.js
```diff
@@ -1,13 +1,13 @@
...
 
-var path     = require('path');
-var fs       = require('fs');
-var crypto   = require('crypto');
-var expect   = require('chai').expect;
-var walkSync = require('walk-sync');
-var appName  = 'some-cool-app';
-var EOL      = require('os').EOL;
-var mkdirp   = require('mkdirp');
+var path       = require('path');
```
kellyselden: extra indentation not necessary?

trabus: I always like to line them up. ¯\_(ツ)_/¯

trabus:
So grouping them together does not appear to be any faster than https://github.com/ember-cli/ember-cli/pull/5725 (thanks for making that PR btw @twokul )
It seems that the `createTestTargets` method doesn't appear to be doing any additional work (at least not a noticeable amount) when run multiple times.

twokul:
@trabus splitting them into separate files makes it easier to debug but I wouldn't imagine it improves the performance (although it could be a nice side effect)

nathanhammond:
@twokul We're actively looking for ways to make this more performant, one of the paths we are investigating was actually _combining_ some tests, which is (amusingly) counter to the one you opened. We don't know yet what the correct solution is, but we're looking for it.

twokul:
@nathanhammond interesting, I didn't know that! feel free to close my PR then 😄

trabus:
Welp, got them passing finally (forgot some dependencies), but it's actually slower this time: 

Bundled: https://travis-ci.org/ember-cli/ember-cli/builds/120299433 (longest 29m 21s)
Split: https://travis-ci.org/ember-cli/ember-cli/builds/120211296 (longest 25m 30s)

Will keep looking at it and see if I can find any reason.

stefanpenner:
going to close this for now, we can reopen if further progress is made.

PR #5729, opened 2016-04-02
[ENHANCEMENT] Remove ember-data from generated addon package.json

trabus:
In 99% of addons, authors do not need or want ember-data in their `package.json`. This PR removes ember-data from an addon's `package.json` `devDependencies`.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -30,6 +30,9 @@ module.exports = {
     contents.dependencies['ember-cli-babel'] = contents.devDependencies['ember-cli-babel'];
     delete contents.devDependencies['ember-cli-babel'];
 
+    // 99% of blueprints don't need ember-data, make it opt-in instead
+    delete contents.devDependencies['ember-data'];
+
     if (contents.keywords.indexOf('ember-addon') === -1) {
       contents.keywords.push('ember-addon');
     }


stefanpenner:
:+1:

Turbo87:
@trabus @stefanpenner why do we actually reuse the package.json from the "app" blueprint? wouldn't it be easier to add a `package.json` to the "addon" blueprint?

nathanhammond:
@Turbo87 The degree to which they're different is small, which helps us make sure that we keep them in sync? Looking at blame, here is this commit: https://github.com/ember-cli/ember-cli/commit/772bfaf4a0f4e61f12643c9f5a3dc9aa3b3a60a4. I don't really have a strong preference here but without a compelling reason to do it a different way I'm not sure that changing it gets us anywhere. (Trading one problem for a different problem.)

Turbo87:
@nathanhammond I see, thanks!

stefanpenner:
I'm fine with either. As the differences are minimal, doing a programmatic patch did help prevent us from getting out of sink.

If we decide it's better to manage this manually or by another path, we can explore that.

trabus:
@Turbo87 The main reason was to make it easier to keep them in sync.

I'm going to catch a movie with my kids, but will be working on this later tonight

nathanhammond:
@trabus I'm :+1: once tests pass.

trabus:
So the failures appear to be due to the addon smoke tests being run first, setting up and caching the `.node_modules-tmp`, which now by not containing `ember-data` trips up the dependency-checker. 

This seems like a problem. :)

nathanhammond:
@trabus So what you're saying is that we use the same `npm install` to support both addons and apps?

Seems like something we should fix (and make our tests even SLOWER).

trabus:
@nathanhammond It seems so.

I just renamed the addon test to run later, we'll see if my hypothesis is correct if the tests pass. :)

trabus:
@nathanhammond That made it even weirder. https://travis-ci.org/ember-cli/ember-cli/builds/120373208

trabus:
@nathanhammond Yeah, this needs some refactoring for sure. https://github.com/ember-cli/ember-cli-internal-test-helpers/blob/master/lib/helpers/acceptance.js#L86

We definitely need to figure out a better way to get the dependencies for the tests that won't require duplication of effort.

nathanhammond:
@trabus Amusing how this issue so neatly dovetails with what you were working on during the hack session. :stuck_out_tongue:

Turbo87:
@trabus did you have time to tackle this problem?

kellyselden:
Might be nice to get some input from the ember-data team. @bmac @fivetanley do you agree?

fivetanley:
I'm all for it being removed.

Get Outlook for iOS




On Sun, Sep 18, 2016 at 2:11 PM -0500, "Kelly Selden" <notifications@github.com> wrote:












Might be nice to get some input from the ember-data team. @bmac @fivetanley do you agree?

stefanpenner:
👍

trabus:
Ugh. This is still having the issues with the caching difference between apps and addons. I'll try and dig back in again this week.

nathanhammond:
@trabus Dumb question: can we just change test order?

trabus:
@nathanhammond unfortunately it still breaks with the tests in the different order, just in a different way (see https://travis-ci.org/ember-cli/ember-cli/jobs/169930993).
```
    1) works in most common scenarios for an example addon
      Running: node_modules/ember-cli/bin/ember build in: /home/travis/build/ember-cli/ember-cli/tmp/some-cool-addon
Missing npm packages: 
Package: ember-disable-prototype-extensions
  * Specified: ^1.1.0
  * Installed: (not installed)
Run `npm install` to install missing dependencies.
```

trabus:
It didn't fix it with the test order either. It runs into a different issue when they're swapped (see previous Travis run for that issue).

nathanhammond:
Hokay I give up on simple fixes. :/

trabus:
homu r+

PR #5740, opened 2016-04-04
bump minimum version of ember-welcome-page

stefanpenner:
addresses:

* cache fixation issue
* live-reload issue


nathanhammond:
@homu r+

PR #5741, opened 2016-04-04
Resolve config directory relative to EmberApp project root

nickiaconis:
Similar in concept to #5614, but for the config directory.

### DIFF
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -60,7 +60,7 @@ describe('models/project.js', function() {
 
       var expected = path.normalize('tests/dummy/config/environment');
 
-      expect(project.configPath()).to.equal(expected);
+      expect(project.configPath()).to.match(new RegExp(expected + '$'));
     });
 
     it('calls getAddonsConfig', function() {
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -60,7 +60,7 @@ describe('models/project.js', function() {
 
       var expected = path.normalize('tests/dummy/config/environment');
 
-      expect(project.configPath()).to.match(new RegExp(expected + '$'));
+      expect(project.configPath().slice(-expected.length)).to.equal(expected);
     });
 
     it('calls getAddonsConfig', function() {
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -132,10 +132,12 @@ EmberApp.prototype._initTestsAndHinting = function(options, isProduction) {
   @param {Object} options
 */
 EmberApp.prototype._initProject = function(options) {
+  var app = this;
+
   this.project = options.project || Project.closestSync(process.cwd());
 
   if (options.configPath) {
-    this.project.configPath = function() { return options.configPath; };
+    this.project.configPath = function() { return app._resolveLocal(options.configPath); };
   }
 };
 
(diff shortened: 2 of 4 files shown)


stefanpenner:
:+1:

chadhietala:
Windows has opinions about paths.

nickiaconis:
@chadhietala I think you're referring to the "Starts with /" part, yes? I could do "Starts with project.root" instead. Thoughts?

tests/unit/models/project-test.js
```diff
@@ -60,7 +60,7 @@ describe('models/project.js', function() {
 
       var expected = path.normalize('tests/dummy/config/environment');
 
-      expect(project.configPath()).to.equal(expected);
+      expect(project.configPath()).to.match(new RegExp(expected + '$'));
```
Turbo87: can we use `expect(...).to.contain()` here?

chadhietala:
@nickiaconis I think you need to account for `\`. The errors in Appveyor seem to be issues with finding files.

tests/unit/models/project-test.js
nickiaconis: We could. I thought the more specific assertion, that the expected path exists at the very end of the output, was a better check than the more generic one, that the expected path exists somewhere within the output.

nickiaconis:
@chadhietala Yes, it looks like it was falling back to the `else` case on appveyor. I changed the condition for the middle case, and I think it should work for both Unix and non-Unix paths now.

nickiaconis:
Okay, I think I'm going about this wrong. `lib/model/project` worked fine as it was, so I'll undo those changes. This only affects custom config paths.

What is the expectation for custom config paths? If my config file lives at `project_root/ember-config/environment.js` what value should `emberAddon.configPath` take in `package.json`? `'ember-config'` or `'ember-config/environment'`? (My guess is the latter.)

nathanhammond:
It should take the latter, `ember-config/environment`, based upon the existing code paths and tests.

nickiaconis:
Okay, for my future self's benefit and the benefit of others, [the origin of the issue lies in the `EmberApp#_configTree` method](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L970). That line assumes that `Project#configPath` returns a fully resolved path. Therefore, `Project#configPath` must be prepended `Project#root` to the relative configuration path. This ensures uses of `Project#configPath` from methods other than `Project#config` work as expected.

nathanhammond:
Okay, so we might have this resolved. Kicking over AppVeyor.

nickiaconis:
@nathanhammond Has this been unblocked by #5774?

nickiaconis:
@nathanhammond How did I not see your message? /)_-

I think it needed a rebase to resolve the issue. I've done that, which kicked off a new build.

nathanhammond:
[It should eventually pass assuming your code works.](https://ci.appveyor.com/project/embercli/ember-cli/build/7821) Kicked it over a second ago for what looked like a transient issue.

nickiaconis:
Passed 4.2, but failed 5.0. I can't tell if it's a false-negative though; it doesn't look like the false-negatives I've seen in the past. @nathanhammond can you take a look?

nathanhammond:
I've not seen that before, seems like it might be a problem from your commit? Kicking it over again anyway.

nathanhammond:
@homu r+

PR #5742, opened 2016-04-05
add encrypted file system note

kellyselden:


stefanpenner:
Ya this would play a role but largely dependent on the type of encryption, for example my laptop has a fully encrypted drive and the overhead is negligible. I suppose that is because the MacBook Pro's have hardware support for this.

It may be worth briefly mentioning something to this^. Don't want people to assume fast ember-cli and encryption are mutually exclusive.

nathanhammond:
Add
> (depending upon system's implementation)

and I'm :+1:

stefanpenner:
We should likely explicitly say something about, that for example the MacBook Pro SSD hardware encryption is known to introduce negligible overhead.

I'll let someone better with words, wordsmith that.

kellyselden:
I'll update it to be more verbose and Windows focused.

Turbo87:
https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md#a-1 already seems to explain most of the encryption slowness. closing this PR for now.

PR #5752, opened 2016-04-07
BUGFIX Unable to name a add-on/project as a substring of "ember-core"

### LINKED ISSUE #5751: Unable to name a add-on/project as a substring of 'ember-core'
It seems that [`findAddonByName`](https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L531-L533) will return addons that are a substring of the `name` given to the function.

This means that [`_initVendorFiles`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L257) will load anything that is a substring of `ember-core`.

Which will make ember-cli crash on [this line](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L263).

I would say this behaviour is a bug and it should be removed.

martndemus:
Fixes #5751

I changed `findAddonByName` from matching on a substring of the requested name
to matching on the suffix.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -522,16 +522,17 @@ Project.prototype.reloadAddons = function() {
 Project.prototype.findAddonByName = function(name) {
   this.initializeAddons();
 
-  var exactMatch = find(this.addons, function(addon) {
-    return name === addon.name || (addon.pkg && name === addon.pkg.name);
-  });
+  var unqualifiedName = name;
+  if (name.indexOf('/') !== -1) {
+    unqualifiedName = name.slice(name.lastIndexOf('/') + 1);
+  }
 
-  if (exactMatch) {
-    return exactMatch;
+  function matchAddon(name, addon) {
+    return name === addon.name || (addon.pkg && name === addon.pkg.name);
   }
 
   return find(this.addons, function(addon) {
-    return name.indexOf(addon.name) > -1 || (addon.pkg && name.indexOf(addon.pkg.name) > -1);
+    return matchAddon(name, addon) || matchAddon(unqualifiedName, addon);
   });
 };
 
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -461,6 +461,16 @@ describe('models/project.js', function() {
       var addon = project.findAddonByName('not-an-addon');
       expect(addon).to.equal(undefined, 'not found addon should be undefined');
     });
+
+    it('should not return an addon that is a substring of requested name', function() {
+      var addon = project.findAddonByName('foo-ba');
+      expect(addon).to.equal(undefined, 'foo-ba should not be found');
+    })
+
+    it('should return an addon that is the unqualified name of requested name', function() {
+      var addon = project.findAddonByName('qux/foo');
+      expect(addon.name).to.equal('foo', 'should have found the foo addon');
+    })
   });
 
   describe('bowerDirectory', function() {


lib/models/project.js
```diff
@@ -529,7 +529,8 @@ Project.prototype.findAddonByName = function(name) {
   }
 
   return find(this.addons, function(addon) {
-    return name.indexOf(addon.name) > -1 || (addon.pkg && name.indexOf(addon.pkg.name) > -1);
+    return (name.slice(addon.name && -addon.name.length) === addon.name) ||
```
rwjblue: Off the top of my head, it is not clear why we aren't just checking if `name === addon.name || name === addon.pkg.name`.  Is that unreasonable?

martndemus: It is checked a few lines up. But it seems being able to match `ember-cli/ember-cli-qunit` with `ember-cli-qunit` is a requirement.

tests/unit/models/project-test.js
```diff
@@ -461,6 +461,16 @@ describe('models/project.js', function() {
...
     });
+
+    it('should not return an addon that is a substring of requested name', function() {
+      var addon = project.findAddonByName('foo-ba');
+      expect(addon).to.equal(undefined, 'foo-ba should not be found');
+    })
+
+    it('should return an addon that is a suffix of requested name', function() {
+      var addon = project.findAddonByName('bar');
+      expect(addon).to.equal(undefined, 'should have found the foo-bar addon');
```
nathanhammond: I don't know that this is our desired behavior. I kinda want to `split('\')` and test the last segment. Is there a scenario where we want to just match the suffix?

In short:
(`bar` !== `foo-bar`)

martndemus: I was thinking about that, but wasn't sure if that was desirable, but I sure can do that if that is more wanted.

lib/models/project.js
rwjblue: `ember-cli-qunit` has a `addon.name`, and even if it didn't its `pkg.name` isn't `ember-cli/ember-cli-qunit`.

martndemus: There is a [test](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/commands/install-test.js#L158) that causes `findAddonByName('ember-cli/ember-cli-qunit')` to be run.

rwjblue: Seems like a bad test :smile_cat:

martndemus: Ok, would you recommend killing/altering that test and then only matching on `===`?

rwjblue: yes

martndemus:
@rwjblue Updated!

martndemus:
Ok, I ended up putting the `ember-cli/ember-cli-qunit` test back, as this tests `ember install ember-cli/ember-cli-qunit`. The workaround I did is what @nathanhammond suggested: If there's a `/` then also match on the last part of the requested name.

tests/unit/models/project-test.js
```diff
@@ -461,6 +461,16 @@ describe('models/project.js', function() {
...
     });
+
+    it('should not return an addon that is a substring of requested name', function() {
+      var addon = project.findAddonByName('foo-ba');
+      expect(addon).to.equal(undefined, 'foo-ba should not be found');
+    })
+
+    it('should return an addon that is the unqualified name of requested name', function() {
+      var addon = project.findAddonByName('qux/foo');
+      expect(addon.name).to.equal('foo', 'should have found the foo addon');
```
nathanhammond: Can you also add a test searching for just `bar` with `qux/foo-bar`? (It's the inverse case of this test and should not match.)

nathanhammond:
One more test and I'm :+1:.

tests/unit/models/project-test.js
martndemus: I think the test on L465 covers that. What you describe would mean that an add-on actually has the name `qux/foo-bar` and the `bar` add-on is looked up.

nathanhammond: You're correct.

nathanhammond:
Waiting on master to be green on AppVeyor before approving.

Turbo87:
@martndemus could you rebase on master now that the AppVeyor builds have been fixed?

martndemus:
@Turbo87 rebased!

Turbo87:
@martndemus thanks!

@homu r+

PR #5754, opened 2016-04-08
Check for `ember-app` or `ember-addon` when looking for `package.json`

### LINKED ISSUE #5493: Check for `ember-app` or `ember-addon` when looking for `package.json` files
ember-cli@2.3.0

Example: cwd is `~/Documents/ember`
`ember new new-tool-ui`

The `new-tool-ui` folder is in `~/`, not `~/Documents/ember`

lukemelia:
In the event that a package.json file somehow ends up above your current
working directory, ember-cli would previously become very confused
because it assumed that the directory with the package.json file was
the root of an ember-cli app or addon. This commit updates the search
algorithm to verify that the package.json file found is indeed an
ember-cli app or addon, based on the package.json keywords.

A joint production of @ryanlabouve and @lukemelia, began at EmberConf
2016 Contributors Workshop.

Fixes #5493

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -571,6 +571,7 @@ Project.closest = function(pathName, _ui, _cli) {
   var ui = ensureUI(_ui);
   return closestPackageJSON(pathName)
     .then(function(result) {
+
       debug('closest %s -> %s', pathName, result);
       if (result.pkg && result.pkg.name === 'ember-cli') {
         return Project.nullProject(_ui, _cli);
@@ -578,7 +579,7 @@ Project.closest = function(pathName, _ui, _cli) {
 
       return new Project(result.directory, result.pkg, ui, _cli);
     })
-    .catch(function(reason) {
+    .catch (function(reason) {
       handleFindupError(pathName, reason);
     });
 };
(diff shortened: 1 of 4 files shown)


lib/models/project.js
```diff
@@ -597,29 +598,37 @@ Project.closest = function(pathName, _ui, _cli) {
...
+    });
   }
-
   debug('dir' + directory);
   debug('pkg: %s', pkg);
   if (pkg && pkg.name === 'ember-cli') {
     return Project.nullProject(_ui, _cli);
   }
-
+  pkg = require(path.join(directory, 'package.json'));
```
Turbo87: why did you switch from `JSON.parse(fs.readFileSync())` to `require()`?

Turbo87:
@lukemelia looks like the smoke tests are complaining about this PR.

can you verify that the algo will now search only for `package.json` files with those keyword and it will completely ignore other `package.json` files. I'm wondering if it would be easier to just seach for the first `package.json` file that is found and check that one for the keywords instead of continuing to search for a matching one after that.

lib/models/project.js
```diff
@@ -597,29 +598,37 @@ Project.closest = function(pathName, _ui, _cli) {
...
-  if (!pkg) {
-    pkg = JSON.parse(fs.readFileSync(path.join(directory, 'package.json')));
+  else {
+    directory = findupPath(pathName, function(directory) {
+      var packageJsonPath = path.join(directory, 'package.json');
+      var isPackageJsonPresent = existsSync(packageJsonPath);
+      if (!isPackageJsonPresent) {
+        return false;
+      }
+      var localPkg = require(packageJsonPath);
```
stefanpenner: lets readFileSync + json.parse or we will cache these files, which i don't think we want to.

lukemelia: Will do.

lukemelia:
Thanks. I'll take a look at the smoke tests and consider your suggestion @Turbo87.

stefanpenner:
needs rebase, but looks good.

stefanpenner:
@lukemelia ^

Turbo87:
I just tried this out and as mentioned in https://github.com/ember-cli/ember-cli/issues/5493 this is problematic since new apps don't get generated with any `ember-app` keyword. Closing this as a new approach is needed.

@lukemelia thanks for your efforts and sorry for the confusion!

PR #5756, opened 2016-04-08
Replace more internal assertion helpers with "chai-files" assertions

Turbo87:
This PR replaces `assertDirEmpty(...)` calls with `expect(dir(...)).to.be.empty` assertions that were introduced in `chai-files@1.2.0`


stefanpenner:
:+1:

Turbo87:
@homu r+

PR #5758, opened 2016-04-08
Revert #5746

Turbo87:
#5746 is under investigation for breaking the AppVeyor builds. Let's see what AppVeyor says to this PR...


nathanhammond:
Even reverted gives the same error. Probably not a problem. Closing for now.

PR #5761, opened 2016-04-08
Improved chai/expect assertion usage

Turbo87:


nathanhammond:
Strongly in favor, appears to have issues in 0.X Node for unidentified reasons.

Turbo87:
@nathanhammond looks like the "encoding" property on the `fs.readFileSync()` call was missing which caused `expected` to be a raw buffer instead of a string.

Turbo87:
@stefanpenner @nathanhammond @rwjblue r?

Turbo87:
merging due to the initial feedback and no objections

@homu r+

PR #5769, opened 2016-04-09
Update ember-ajax to 2.0.

rwjblue:


jaswilli:
It would be nice to get `status` code handling sorted out before a new major version bump is included in ember-cli. See: https://github.com/ember-cli/ember-ajax/pull/62

alexlafroscia:
@jaswilli I took care of that, `2.0.1` has that fix as @rwjblue suggested.

rwjblue:
Thanks @jaswilli for catching that, should be sorted now with 2.0.1. I'll update the version in this PR shortly.

rwjblue:
Updated to ember-ajax@2.0.1.

rwjblue:
@homu r+

PR #5786, opened 2016-04-13
Deprecate `Project.closest` in favor of `Project.closestSync`

### LINKED ISSUE #5755: Deprecate `Project.closest` in favor of `Project.closestSync`
While investigating https://github.com/ember-cli/ember-cli/issues/5493 we discovered that `Project.closest` was never used internally by ember-cli. Instead, `Project.closestSync` is used exclusively.

Would it be beneficial to deprecate `Project.closest` for future removal?

`Project.closest`:
https://github.com/ember-cli/ember-cli/blob/e016fbe/lib/models/project.js#L570

`Project.closestSync`:
https://github.com/ember-cli/ember-cli/blob/e016fbe/lib/models/project.js#L597

/cc @lukemelia

jeffjewiss:
This adds a deprecation message for `Project.closest`.

I tried searching the code and issues for examples of adding a deprecation, but didn't find an obvious example, so please forgive me if I missed some steps.

Resolves #5755

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -561,6 +561,7 @@ Project.prototype.generateTestFile = function(/* moduleName, tests */) {
   Returns a new project based on the first package.json that is found
   in `pathName`.
 
+  @deprecated
   @private
   @static
   @method closest
@@ -569,6 +570,9 @@ Project.prototype.generateTestFile = function(/* moduleName, tests */) {
  */
 Project.closest = function(pathName, _ui, _cli) {
   var ui = ensureUI(_ui);
+
+  ui.writeDeprecateLine('`Project.closest` is a private method that will be removed, please use `Project.closestSync` instead.');
+
   return closestPackageJSON(pathName)
     .then(function(result) {
       debug('closest %s -> %s', pathName, result);


Turbo87:
@jeffjewiss thanks for the PR! 

you should be able to use `ui.writeDeprecateLine('`Project.closest` is a private method that will be removed, please use `Project.closestSync` instead.')` instead of the itself deprecated `deprecate()` utility.

jeffjewiss:
@Turbo87 thanks for clarifying how the deprecation should be written, appreciate it! I've updated the commit to use `ui.writeDeprecateLine`.

lib/models/project.js
```diff
@@ -564,11 +564,14 @@ Project.prototype.generateTestFile = function(/* moduleName, tests */) {
   @private
   @static
   @method closest
-  @param  {String} pathName Path to your project
+  @param  {String} pathName Path to your project (deprecated)
```
Turbo87: @jeffjewiss sorry for nitpicking, but we should probably use the `@deprecated` annotation here roughly like this:

```js
/**
   Returns a new project based on the first package.json that is found
   in `pathName`.
 
    @deprecated
    @private
    @static
    ...
*/
```

jeffjewiss: No worries! Should I also remove `(deprecated)` from `@param`?

Turbo87: yes, I think that would make sense because it's the method that's deprecated, not the parameter

jeffjewiss: Okay cool, thanks again!

Turbo87:
@homu r+

PR #5789, opened 2016-04-13
Remove "home grown version checker"

Turbo87:
As described in https://github.com/ember-cli/ember-cli/pull/5661 the "home grown version checker" was creating more problems than it solved and was disabled by default. I don't think that anyone will really miss it...


Turbo87:
@stefanpenner @rwjblue r?

rwjblue:
I'd rather see this land together with #5661, but either way I'm in favor of removing...

Turbo87:
@homu r+

PR #5792, opened 2016-04-13
Add "rootURL" setting to deprecate "baseURL"

### LINKED ISSUE #417: rootURL deploy issues
Hey,

My use case: deploy to a directory inside an S3 bucket, say `http://s3bucket.com/foo`.

With the default rootURL of `'/'`, this wont work because it will try to resolve the assets at `http://s3bucket.com/app.js`

If I change the rootURL to `/foo/`, the assets load correctly, but now Ember's router gives double rootURL and opening your app takes you to `http://s3bucket.com/foo/foo`.

I think that for the defaults, all the paths should be relative so the above will work out of the box, and users can customize `environment.js` per their needs.

https://github.com/stefanpenner/ember-cli/issues/172

### LINKED ISSUE #2633: Kill `<base>` tag
The `<base>` tag has a number of unexpected side-effects. For instance it prevents SVG urls from working correctly. We should not use them by default.

### LINKED ISSUE #2957: baseUrl not set in built index.html if locationType is hash

### LINKED ISSUE #2989: background-image: url() breaks with ENV.baseURL = 'anything' if app is served on {base}/anything
## Problem ##

When ENV.baseURL is set, and your server is serving the production build of your ember app to a route other than the base, background-image: url()'s break

Here is my example:

https://github.com/Harris-Miller/cssUrlExample

pull that down, `ember install`, and `ember server`
I have 2 images being displayed on screen, one using an  tag, and the other using background-image. both load. and both are "/assets/images/tomster.png"

now do `ember build --environment=production`
then run `node app`

app.js is a simple express server that I whipped up and serves the /dist folder to localhost:3000/cssUrlExample

i also have in config/environments.js
```javascript
if (environment === 'production') {
  ENV.baseURL = 'cssUrlExample';
}
```
with that, the  tag works fine with just `src="/assets/images/tomster.png"` because of the added <base> tag
however the `background-image: url('/assets/images/tomster.png')` fails to load, because it's looking at "http://localhost:300/assets/images/tomster.png" and not "http://localhost:300/cssUrlExample/assets/images/tomster.png"

## Why this is an issue ##

I can see instances where serving an ember app like I am in my example could become actually quite common. Let's say I have a website built on an everyday server-side MVC. And lets say that we decide that a new sub-section of the site will be an ember app. So the server-side routes will server the production build of /dist from the ember app to that route (just like I am in my example express server). In this case, the background-image: url()'s will all fail if they are servered locally and not through a CDN

## Possible Solution ##

css url rewrites to append ENV.baseURL
e.g.
`ENV.baseURL = "cssUrlExample"`
so `background-image: url('/assets/images/tomster.png');`
becomes
`background-image: url('/cssurlExample/images/tomster.png');`

I'm still looking for the best solution myself on how to accomplish this (trying to find the right npm module to use) and I'll submit a pull request when I find it.

Would like to here thoughts on this and if you agree that this is an issue, or is there is a work around that I'm not aware of right now

The repo I posted about was build on:
```
version: 0.1.7
node: 0.10.35
npm: 2.1.8
```

### LINKED ISSUE #4507: Need more consistent handling of `baseURL` for testing apps in subdirectories
### Description
I am currently working on a production app that is served from a subdirectory (`/dashboards`). I'd like to be able to run tests in the following scenarios:
 * During development with `ember serve` by visiting `/tests`
 * After deploying the output of an `ember build` by visiting `/tests`
 * On the command-line using testem via  `ember test`
 * Through testem's server via `ember test --server`

I can't seem to find a combination of `baseURL` settings which enables all of these without further hacks.

---

### Actual

Here's the table of behaviors I am observing (using a just-generated clean app):

`baseURL`(development) | `baseURL` (test) | `ember serve` | `ember build` | `ember test` | `ember test --server`
--- | --- | :---: | :---: | :---: | :---:
`/dashboards` | `/`° | ✘¹ | ✘¹ | ✔ | ✔
`/dashboards` | `/dashboards` | ✔ | ✔ | ✘² | ✘²

° This is the default in `config/environment.js`
¹ Visiting `/tests` fails, visiting `/dashboards/tests` results in 404 errors for all assets
² Testing hangs due to 404 errors for all assets

---

### Expected

`baseURL`(development) | `baseURL` (test) | `ember serve` | `ember build` | `ember test` | `ember test --server`
--- | --- | :---: | :---: | :---: | :---:
`/dashboards` | ? | ✔ | ✔ | ✔ | ✔

---

### Version
```
$ ember -v
version: 1.13.1
node: 0.12.0
npm: 2.12.1
```

### LINKED ISSUE #4905: Use rootURL instead of baseURL
Long story short: they should have the same values so they shouldn't be set in different places, and the `<base>` tag causes problems, so let's get rid of baseURL.

Long story medium: @HeroicEric did a nice little writeup about how confusing the current situation is in an [issue on the Guides](https://github.com/emberjs/guides/issues/612#issue-101730315). @SaladFork had a conversation with @rwjblue about the current state of things and how to un-confuse the situation. He's going to clean up his notes and add them here shortly.

In the meantime, I'm closing out the issue on the Guides, since we don't want to encourage people to use a soon-to-be-deprecated feature of Ember CLI.

Turbo87:
This PR adds a new `rootURL` setting to `config/environment.js` that should be used instead of the old `baseURL` setting which had several issues related to the use of the `<base>` tag.

- [x] `{{rootURL}}` in `index.html` files is replaced with the actual value
- [x] `rootURL` is used instead of `baseURL` for `ember serve` (if it is set)
- [x] a `testem-url-rewriter` middleware addon is added to rewrite URLs with `rootURL` prefix to URLs without that prefix to make testem happy
- [x] adjust the app blueprint
- [x] deprecate the usage of `baseURL`
- [x] test all the things!
- [x] no, srsly, write unit tests
- [x] use all available label colors
- [x] Write deprecation guide (@nathanhammond)

#### Test Cases

- `baseURL: '/'`
- `baseURL: '/foo/bar'`
- `baseURL: '/', rootURL: '/'`
- `baseURL: '/foo/bar', rootURL: '/'`
- `baseURL: '/', rootURL: '/foo/bar'`
- `baseURL: '/foo/bar', rootURL: '/foo/bar'`

---

- `ember serve` and checking if `/tests` (or `<rootURL>/tests`) works
- `ember test`
- `ember test --environment=production`
- `ember test --server`

---

resolves #4905, resolves #4507, resolves #2989, resolves #2957, resolves #2633, resolves #417

/cc @rwjblue @stefanpenner @nathanhammond

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -43,6 +43,14 @@ module.exports = Command.extend({
       .then(function(commandOptions) {
         var config = this.project.config(commandOptions.environment);
 
+        this.ui.writeDeprecateLine(
+          'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
+          !(!config.rootURL && config.baseURL));
+
+        this.ui.writeWarnLine(
+          'The `baseURL` and `rootURL` settings should not be used at the same time.',
+          !(config.rootURL && config.baseURL));
+
         commandOptions = assign({}, commandOptions, {
           rootURL: config.rootURL,
           baseURL: config.baseURL || '/'
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -44,7 +44,8 @@ module.exports = Command.extend({
         var config = this.project.config(commandOptions.environment);
 
         commandOptions = assign({}, commandOptions, {
-          baseURL: config.rootURL || config.baseURL || '/'
+          rootURL: config.rootURL,
+          baseURL: config.baseURL || '/'
         });
 
         if (commandOptions.proxy) {
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -41,8 +41,10 @@ module.exports = Command.extend({
     return this._checkExpressPort(commandOptions)
       .then(this._autoFindLiveReloadPort.bind(this))
       .then(function(commandOptions) {
+        var config = this.project.config(commandOptions.environment);
+
         commandOptions = assign({}, commandOptions, {
-          baseURL: this.project.config(commandOptions.environment).baseURL || '/'
+          baseURL: config.rootURL || config.baseURL || '/'
         });
 
         if (commandOptions.proxy) {
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -9,16 +9,16 @@
 
     {{content-for "head"}}
 
-    <link rel="stylesheet" href="assets/vendor.css">
-    <link rel="stylesheet" href="assets/<%= name %>.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
 
     {{content-for "head-footer"}}
   </head>
   <body>
     {{content-for "body"}}
 
-    <script src="assets/vendor.js"></script>
-    <script src="assets/<%= name %>.js"></script>
+    <script src="{{rootURL}}assets/vendor.js"></script>
+    <script src="{{rootURL}}assets/<%= name %>.js"></script>
 
     {{content-for "body-footer"}}
   </body>
--- a/blueprints/app/files/app/router.js
+++ b/blueprints/app/files/app/router.js
@@ -2,7 +2,8 @@ import Ember from 'ember';
 import config from './config/environment';
 
 const Router = Ember.Router.extend({
-  location: config.locationType
+  location: config.locationType,
+  rootURL: config.rootURL
 });
 
 Router.map(function() {
--- a/blueprints/app/files/config/environment.js
+++ b/blueprints/app/files/config/environment.js
@@ -4,7 +4,7 @@ module.exports = function(environment) {
   var ENV = {
     modulePrefix: '<%= modulePrefix %>',
     environment: environment,
-    baseURL: '/',
+    rootURL: '/',
     locationType: 'auto',
     EmberENV: {
       FEATURES: {
@@ -29,7 +29,6 @@ module.exports = function(environment) {
 
   if (environment === 'test') {
     // Testem prefers this...
-    ENV.baseURL = '/';
     ENV.locationType = 'none';
 
     // keep test console output quieter
--- a/blueprints/app/files/tests/index.html
+++ b/blueprints/app/files/tests/index.html
@@ -10,9 +10,9 @@
     {{content-for "head"}}
     {{content-for "test-head"}}
 
-    <link rel="stylesheet" href="assets/vendor.css">
-    <link rel="stylesheet" href="assets/<%= name %>.css">
-    <link rel="stylesheet" href="assets/test-support.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/test-support.css">
 
     {{content-for "head-footer"}}
     {{content-for "test-head-footer"}}
@@ -21,12 +21,12 @@
     {{content-for "body"}}
     {{content-for "test-body"}}
 
-    <script src="testem.js" integrity=""></script>
-    <script src="assets/vendor.js"></script>
-    <script src="assets/test-support.js"></script>
-    <script src="assets/<%= name %>.js"></script>
-    <script src="assets/tests.js"></script>
-    <script src="assets/test-loader.js"></script>
+    <script src="{{rootURL}}testem.js" integrity=""></script>
+    <script src="{{rootURL}}assets/vendor.js"></script>
+    <script src="{{rootURL}}assets/test-support.js"></script>
+    <script src="{{rootURL}}assets/<%= name %>.js"></script>
+    <script src="{{rootURL}}assets/tests.js"></script>
+    <script src="{{rootURL}}assets/test-loader.js"></script>
 
     {{content-for "body-footer"}}
     {{content-for "test-body-footer"}}
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -693,6 +693,9 @@ EmberApp.prototype._templatesTree = function() {
 */
 EmberApp.prototype._configReplacePatterns = function() {
   return [{
+    match: /\{\{rootURL\}\}/g,
+    replacement: calculateRootURL
+  }, {
     match: /\{\{EMBER_ENV\}\}/g,
     replacement: calculateEmberENV
   }, {
@@ -1709,6 +1712,10 @@ function calculateBaseTag(config) {
   } else {
     return '';
   }
+}
+
+function calculateRootURL(config) {
+  return cleanBaseURL(config.rootURL) || '';
 }
 
 function calculateEmberENV(config) {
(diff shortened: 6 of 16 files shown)


lib/commands/serve.js
```diff
@@ -41,8 +41,10 @@ module.exports = Command.extend({
     return this._checkExpressPort(commandOptions)
       .then(this._autoFindLiveReloadPort.bind(this))
       .then(function(commandOptions) {
+        var config = this.project.config(commandOptions.environment);
+
         commandOptions = assign({}, commandOptions, {
-          baseURL: this.project.config(commandOptions.environment).baseURL || '/'
+          baseURL: config.rootURL || config.baseURL || '/'
```
rwjblue: I'd prefer to keep `baseURL` and `rootURL` as separate options to the command.

rwjblue:
Lookin' good so far!

lib/commands/serve.js
Turbo87: why?

rwjblue: Because `baseURL` will be deprecated, and I'd rather not have to think "is this referencing the old deprecated thing or the coerced thing" everytime I see the word `baseURL` in the repo.

Turbo87: adjusted in 8d2d839, though I'm not quite convinced if that really has any substantial benefit

rwjblue: @Turbo87 The @rwjblue from 3 months from now thanks you in advance. :smile_cat:

stefanpenner:
does this also need to do deep rewrites, or is index.html sufficient.

Turbo87:
@stefanpenner I've talked about this with @rwjblue and we figured that `rootURL` is available via the config module in JS files so any URLs there could be prefixed manually and for CSS files the `url()` directives are relative to the location of the CSS file so `rootURL` should be irrelevant.

Also note that `baseURL` still works the same as before until you specify a `rootURL` option.

Turbo87:
whoa, even AppVeyor is green on this! :trollface:

rwjblue:
@stefanpenner - I do not think we need deep rewrite support. 

* New apps to be generated without a base tag. 
* Existing apps can continue to use a base tag and make no changes (with a deprecation)
* Existing apps can update to remove base tag once they confirm any tweaks to asset paths that they had to make to work with base are reverted. Basically any relative path in html/templates/css will need to be confirmed since with base they are relative to base, but after removing base they are relative to the file they are included in.
* Existing apps can avoid the deprecation AND avoid the above mentioned relative asset review/audit by  removing the rootURL reference in app/router.js and adding this to their index.html:

    <base href="{{rootURL}}">

stefanpenner:
Awesome.

Do we have a  deprecation/upgrade guide/comment for them to follow?

rwjblue:
Not yet, but we will need to write a detailed deprecation/upgrade guide before landing this.

stefanpenner:
> Not yet, but we will need to write a detailed deprecation/upgrade guide before landing this.
>  Show all checks

:+1:

rwjblue:
@nathanhammond / @Turbo87 - Looks like the only thing left right now is the deprecation guide?  Is that correct? Any chance of landing this soonish in canary to let things simmer?

nathanhammond:
@rwjblue Confirm, I should probably write that tonight. I'll get everything done to merge this in < 24 hours. Five weeks on canary seems like a good maturation period.

Turbo87:
@rwjblue @stefanpenner @nathanhammond is this actually blocked on @nathanhammond writing the deprecation guide or can we merge/publish the deprecation guide after merging this PR into master/canary?

rwjblue:
I'm fine with it landing on canary, as long as we make sure to get the guide done and published before we branch to 2.7.0-beta.1 (~2-3 weeks)...

rwjblue:
@homu r+

nathanhammond:
@homu r-

nathanhammond:
Gimme a minute to test something. Update in thirty minutes.

nathanhammond:
Test worked, I'm in favor of this landing again. Deprecation blog post is mostly written. Please review (draft link in Slack).

Turbo87:
@homu r+

rwjblue:
I think @homu is stuck due to the `r-` above, but I'm not 100% certain (the website to list pending PR's state is down).

sandstrom:
@Turbo87 This is a great change! ⛵ 

One thing that I noticed when making the change is that `rootURL` has different meaning in ember-cli and ember. In ember it sets the mount-path for the client code. In ember-cli it determines where files are built or served.

In our case we serve/build on `/` but mount on `/client/`. This is trivial to support by not passing through the value from ember-cli into the router.

However, I wonder if you considered giving them different names to clarify this difference? E.g. `mountPath` for the ember.js value and rootURL for ember-cli.

nathanhammond:
@sandstrom At build time we set up the server to listen at the `rootURL`. We're actively trying to reduce confusion in this area and consolidate to a single name. The ability to `ember s` at a different path from the eventual path inside of the Ember application is a mis-feature that will troll users in innumerable ways now that we don't set a `baseURL`.

Can you explain your reasons for needing to `ember s` at `/` or is that simply your existing pattern?

sandstrom:
We have a proxy in front, forwarding different paths to different things. For example:

- `example.com/special/deep/admin` -> admin-client (build output at `/srv/www/admin-client`)
- `example.com/deep/path/blog` -> blog-client (build output at `/srv/www/blog`)
- `example.com` -> static website (build output at `/srv/www/static`)

(this goes for both development and production)

For simplicity it's nicer to build at `/`, otherwise we'd have, for example, the build output for admin-client at `/srv/www/admin-client/special/deep/admin` which is a tautology.

nathanhammond:
@sandstrom I think what you're looking for is `--output-path` which you can specify for both `ember build` and `ember serve`. Can you poke at that and see if it meets your needs?

sandstrom:
@nathanhammond I appreciate the suggestion, but `--output-path` does something else.

In the example above, if we couldn't have different values for `rootURL` in ember-cli and in ember.js we'd end up with this:

- `example.com/special/deep/admin` -> admin-client (build output at `/srv/www/special/deep/admin`)
- `example.com/deep/path/blog` -> blog-client (build output at `/srv/www/deep/path/blog`)

This is bad, because each app now need to know where it lives in our global routing structure.

Since we can give different values to `rootURL`, we don't end up with this. But that's exactly my point, that these two `rootURL` values are different things and it would be better if they had different names.

nathanhammond:
@sandstrom Find me in the ember community slack, let's chat. 😄

kanongil:
This feature means that `ember-cli` no longer supports apps where the compiled output can be deployed to multiple locations. Eg. `/one/` & `/two/`.

This is essential for eg. Cordova deployments, which serves the app using a plain `file:///...` URL which can change for each installation.

Did you miss this use-case? I see no mention of it in the deprecation notice.

rwjblue:
@kanongil - Please open an issue (commenting on a merged PR often gets missed), and explain in a bit greater detail (preferably with a reproduction).

kanongil:
@rwjblue Thanks. Issue #6178 has been posted.

mariogintili:
related to https://github.com/ember-cli/ember-cli/issues/6199 - if anyone could clear the air on this that would be great 😄

PR #5797, opened 2016-04-14
Use "fs-extra" convenience functions

Turbo87:
We already depend on the `fs-extra` package for some of our code, but also on other dependencies that do some of the same things. This PR removes a few of those other dependencies in favor of using `fs-extra` everywhere.

### DIFF
--- a/blueprints/in-repo-addon/index.js
+++ b/blueprints/in-repo-addon/index.js
@@ -20,7 +20,7 @@ module.exports = {
 
   afterInstall: function(options) {
     var packagePath = path.join(this.project.root, 'package.json');
-    var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
+    var contents    = fs.readJsonSync(packagePath, { encoding: 'utf8' });
     var name        = stringUtil.dasherize(options.entity.name);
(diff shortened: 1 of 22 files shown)


blueprints/in-repo-addon/index.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
 
   afterInstall: function(options) {
     var packagePath = path.join(this.project.root, 'package.json');
-    var contents    = JSON.parse(fs.readFileSync(packagePath, { encoding: 'utf8' }));
+    var contents    = fs.readJsonSync(packagePath, { encoding: 'utf8' });
```
stefanpenner: is the encoding then required?

Turbo87: I'm not sure. `fs-extra` itself is not adding an encoding by itself, so I just kept the existing encoding options for now.

stefanpenner: just checked, it appears to not be required.

Turbo87: I've removed them in 8c30c71

stefanpenner:
LGTM :shipit:

Turbo87:
@homu r+

PR #5238, opened 2015-12-20
Serve Task: Prevent Termination Confirmation on Windows

felixrieseberg:
* On Windows, the serve task can't be cancelled with a single Ctrl + C,
because Windows will ask for batch job termination confirmation.
* This commit adds a small windows-only listener for Ctrl + C, handling
process termination inside of Ember Cli.

Closes #5236

PR #5800, opened 2016-04-14
Rebased "Prevent Termination Confirmation on Windows"

Turbo87:
This is essentially https://github.com/ember-cli/ember-cli/pull/5238 rebased and with a few small test improvements. I've tested this with inquirer and didn't notice any problems.

/cc @felixrieseberg @stefanpenner @kellyselden

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -80,7 +80,24 @@ module.exports = Task.extend({
       process.on('SIGINT',  this.onSIGINT.bind(this));
       process.on('SIGTERM', this.onSIGTERM.bind(this));
       process.on('message', this.onMessage.bind(this));
+
+      if (/^win/.test(process.platform)) {
+        this.trapWindowsSignals();
+      }
+
       signalsTrapped = true;
+    }
+  },
+  
+  trapWindowsSignals: function () {
+    // This is required to capture Ctrl + C on Windows
+    if (process.stdin && process.stdin.isTTY) {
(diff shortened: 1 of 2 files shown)


felixrieseberg:
Thanks for staying on top of this 👍

Turbo87:
AppVeyor tests seem mostly green except for the usual suspects :shipit:

stefanpenner:
So this seems to trap for `builder` related stuff, but should this actually be more general? Allowing for this behavior in all scenarios instead of just during builds?

Turbo87:
@stefanpenner I tend to agree, having this everywhere would probably make sense.

stefanpenner:
I have tested it, and it works fine during the build-phase, but it is strange that other phases (even just before build, or just after, but before the process exists) that the prompt still appears.

So this is a good step in the right direction, as the build phases are longest, but it feels like it should be a more all covering.

Turbo87:
@stefanpenner IMHO we should merge this as a first step and then we can try to apply it more genetically in a later PR

stefanpenner:
> we can try to apply it more genetically

where you thinking of using https://en.wikipedia.org/wiki/Needleman%E2%80%93Wunsch_algorithm ?

:trollface: 

----



> @stefanpenner IMHO we should merge this as a first step

👍

Turbo87:
:joy: that should've said generically... damn phone keyboard

Turbo87:
@homu r+

PR #5803, opened 2016-04-15
Use Ember.assign instead of Ember.merge if possible

mike-north:
Newly created apps and addons throw deprecation warnings if used w/ `ember#release`. This is a :-1: UX. Unfortunately, I have no good ideas around how we can build tests around this, because `ember#release` is intentionally and inherently a moving target (zing!).

Additionally, we shouldn't set ourselves on a path where setting up an addon to test against older versions of Ember (i.e., the 2.4 LTS release, 1.13) would require *each* project maintainer to set up a similar adjustment.

**Note:** Typically this would be handled w/ a polyfill, but [it seems like that's not the plan in this case of Ember.assign](https://github.com/emberjs/ember.js/pull/12303#issuecomment-140040603)

## Failing case

```
ember new hello718 && \
  cd hello718 && \
  bower install --save ember#release && \
  ember g route hello && \
  ember g acceptance-test hello
```

Update your
**config/environment.js**
```
ENV.EmberENV.RAISE_ON_DEPRECATION = true;
```

And now run
```
ember test ci
```

You’ll see test fail due to deprecation warnings

### DIFF
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  let attributes = (Ember.assign | Ember.merge)({}, config.APP);
-  attributes = (Ember.assign | Ember.merge)(attributes, attrs); // use defaults, but you can override;
+  let attributes = (Ember.assign || Ember.merge)({}, config.APP);
+  attributes = (Ember.assign || Ember.merge)(attributes, attrs); // use defaults, but you can override;
 
   Ember.run(() => {
     application = Application.create(attributes);
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  let attributes = Ember.merge({}, config.APP);
-  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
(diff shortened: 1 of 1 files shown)


blueprints/app/files/tests/helpers/start-app.js
```diff
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  let attributes = Ember.merge({}, config.APP);
-  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
+  let attributes = (Ember.assign | Ember.merge)({}, config.APP);
```
topaxi: Shouldn't this be `||` instead of just one `|`?

blueprints/app/files/tests/helpers/start-app.js
```diff
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  let attributes = Ember.merge({}, config.APP);
-  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
+  let attributes = (Ember.assign || Ember.merge)({}, config.APP);
+  attributes = (Ember.assign || Ember.merge)(attributes, attrs); // use defaults, but you can override;
```
Turbo87: should we do something like `let assign = Ember.assign || Ember.merge; let attributes = assign(...)` instead to reduce code duplication?

rwjblue:
FWIW - The deprecation for `Ember.merge` has been removed from canary, beta, and release channels of Ember.  It will be included in v2.5.1 which will be released within a few days.

PR #5805, opened 2016-04-15
index.html and test blueprint tune-up suggestions

sheriffderek:
1. Set doctype to lowercase
  -doctype is case-insensitive / I see no reason to capitolize it
  -https://www.w3.org/TR/html5/syntax.html#the-doctype

2. Remove X-UA-Compatible" content="IE=edge" tag
  -The use cases for this are very very few
  -This meta-tag should be added by the user if needed - instead of a default
  -http://stackoverflow.com/questions/26346917/why-use-x-ua-compatible-ie-edge-anymore

3. Set to single quotes
  -I see no reason not to keep every bluprint as lean as possible
  -This is generally the only page in my app that has double quotes

I end up doing this for every project. Just throwing it out there. : )

### DIFF
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
     <meta charset="utf-8">
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title><%= namespace %></title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -1,24 +1,24 @@
 <!doctype html>
 <html>
   <head>
-    <meta charset="utf-8">
+    <meta charset='utf-8'>
     <title><%= namespace %></title>
-    <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name='description' content=''>
+    <meta name='viewport' content='width=device-width, initial-scale=1'>
 
-    {{content-for "head"}}
+    {{content-for 'head'}}
 
-    <link rel="stylesheet" href="assets/vendor.css">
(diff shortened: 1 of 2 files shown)


kellyselden:
@sheriffderek as for the quotes, we tend to prefer double quotes in html/hbs files. Don't know if it's explicitly called out somewhere though.

rwjblue:
Confirm, we had a change somewhere from 1.13 to 2.4 that actually modified those quotes to use doubles instead of singles. I don't think we should cause extra churn (again) here...

raido:
I agree with @rwjblue.

stefanpenner:
seems like we should keep the compat until we drop IE9 and IE10 ?

sheriffderek:
Certainly not trying to ruffle any feathers, but I'm unaware of any plus to double quotes besides maybe racking up more billable hours. Double quotes are one extra key stroke, and aren't single quotes easier on the server? Not trying to do a tab/space things - just honestly curious why that would be preferred. Thanks.

Added a link to some discussion about the X-UA-Compat

stefanpenner:
> just honestly curious why that would be preferred.

I think either is fine, which is why I'm inclined to leave it as is.

stefanpenner:
> Added a link to some discussion about the X-UA-Compat

If memory serves this was added to be defensive in scenarios with mis-configured servers or users systems that have an an in-compatible group policy set.

If the above mentioned concerns no-longer valid we can re-evaluate. The SO link provided doesn't indicate this, unless i mis read.

rwjblue:
> I'm unaware of any plus to double quotes besides maybe racking up more billable hours

Clearly, the HTML Engines parse double quotes faster. :smiling_imp: 

In the end, it doesn't really matter why I/we prefer double quotes over single quotes. There is basically no benefit to changing back to single quotes in this file, and forcing every user to tweak this when upgrading is just a waste of time.

rwjblue:
Also, RE:

> This is generally the only page in my app that has double quotes

You can always rescue your apps templates instead... :smiley_cat:

Turbo87:
I tend to agree with the others:

1. doctype is usually used in uppercase in all examples and even in the link you've posted so there is no need for us to change it

2. "X-UA-Compatible" see @stefanpenner's answer and feel free to write up an RFC on the removal of that thing

3. single quotes: most examples in the Ember world use double quotes in hbs files and we've just recently changed this so there is no reason to change it again

sheriffderek:
Thanks for checking it out! Glad to see the reasoning. Thanks.

stefanpenner:
:)

PR #5811, opened 2016-04-17
Fix Ember 2.5.0 deprecation warning for Ember.merge

jasonmit:


blueprints/app/files/tests/helpers/start-app.js
```diff
@@ -2,11 +2,13 @@ import Ember from 'ember';
...
 
+const assign = Ember.assign || Ember.merge;
+
 export default function startApp(attrs) {
   let application;
 
-  let attributes = Ember.merge({}, config.APP);
-  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
+  let attributes = assign({}, config.APP);
+  attributes = assign(attributes, attrs); // use defaults, but you can override;
```
jasonmit: We likely don't need to reassign `attributes` since `Ember.merge` and `Ember.assign` mutate the source object, `attributes`, but I kept that intact

stefanpenner:
This is no longer deprecated, and also the 3rd pr closed today. Thanks for the fix, but I think the deprecation was to hasty especially considering its part of the default template. Once we bundle blueprints correctly with ember versions we can revisit

jasonmit:
Oops, sorry -- I checked the opened PRs and issues before submitting and didn't see anything.

stefanpenner:
@jasonmit no problem, this was a strange scenario. Thank you for providing the PR though, in most cases this would be exactly what we want.

PR #5814, opened 2016-04-18
Deprecate and remove this.concatFiles

### LINKED ISSUE #5280: Deprecate and remove this.concatFiles
see https://github.com/ember-cli/ember-cli/issues/5277#issuecomment-167452827

jeffjewiss:
Closes #5280


stefanpenner:
LGTM

@rwjblue r?

rwjblue:
`this.concatFiles` also adds `options.sourceMapConfig = this.options.sourcemaps;` to the options provided, does `broccoli-concat` do this by default or does that need to be added to each `concat` invocation here?

jeffjewiss:
As far as I can tell `broccoli-concat` does not do this by default. I'll add the sourcemap options to each invocation of `concat`.

Turbo87:
@stefanpenner @rwjblue r?

tests/fixtures/dummy-project-outdated/package.json
```diff
@@ -27,4 +27,4 @@
     "express": "^4.1.1",
     "glob": "^3.2.9"
   }
-}
\ No newline at end of file
+}
```
stefanpenner: ?

Turbo87: most likely an editor that ensures new line at EOF

stefanpenner:
@rwjblue / @Turbo87 do we see anyone using this.concatFiles as a way to monkey patch? Do we care?  If we do, we may need a different deprecation strategy.

nathanhammond:
@stefanpenner Me! I use it to monkey-patch things! I don't disagree with this, but I'm in favor of a different deprecation strategy which still uses `this.concatFiles` for each of the existing invocations.

rwjblue:
We can make a private string that we can pass in options, that is not exported.  That way internal invocations can be "safe", but external ones are not.

stefanpenner:
> We can make a private string that we can pass in options, that is not exported. That way internal invocations can be "safe", but external ones are not.

ya, something like this would do the trick.

Turbo87:
should we close this PR in favor of the new suggested approach?

stefanpenner:
> should we close this PR in favor of the new suggested approach?
>  Hide all checks

should be easy enough to retrofit this PR, @jeffjewiss got any spare cycles for that coming up?

jeffjewiss:
Yeah, I'd be happy to take this on.

I'm having a little trouble understanding the means of deprecating something using a private string that can be passed in to the options. Is there an example of this you could point me to as a reference? Thanks!

Turbo87:
closing in favor of #5926

PR #5819, opened 2016-04-19
Fix for disabling hinting for developing an addon

### LINKED ISSUE #5594: Cannot disable hinting in addon when isDevelopingAddon is true
I have an app which includes an `npm link`ed addon in its `package.json`. Both the app and the addon have `hinting: false` in their `ember-cli-build.js` files, and the addon has `isDevelopingAddon: () => true` in its `index.js`. If I `ember serve` the app, the hinting option is ignored for the addon.

jeffjewiss:
Please let me know if these aren't the types of tests you're looking for to cover this scenario. I figured it was worth trying to write them to get some feedback. Thanks!

Closes #5594

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -131,8 +131,9 @@ var Addon = CoreObject.extend({
   hintingEnabled: function() {
     var isProduction = process.env.EMBER_ENV === 'production';
     var testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND || !isProduction;
+    var explicitlyDisabled = this.app && this.app.options && this.app.options.hinting === false;
 
-    return testsEnabledDefault;
+    return testsEnabledDefault && !explicitlyDisabled;
   },
 
   /**
--- a/tests/unit/models/addon-test.js
+++ b/tests/unit/models/addon-test.js
@@ -407,6 +407,99 @@ describe('models/addon.js', function() {
       });
     });
 
+    describe('hintingEnabled', function() {
+      /**
+        Tests the various configuration options that affect the hintingEnabled method.
+
+       | configuration | test1 | test2 | test3 | test4 | test5 |
+       | ------------- | ----- | ----- | ----- | ----- | ----- |
+       | hinting       | true  | true  | true  | false | unset |
+       | environment   | dev   | N/A   | prod  | N\A   | N\A   |
+       | test_command  | set   | set   | unset | set   | set   |
+       | RESULT        | true  | true  | false | false | true  |
+
+        @method hintingEnabled
+       */
+
+      var originalEnvValue, originalTestCommand, addon, project;
+
+      beforeEach(function() {
+        var MyAddon = Addon.extend({
+          name: 'test-project',
+          root: 'foo'
+        });
+
+        var projectPath = path.resolve(fixturePath, 'simple');
+        var packageContents = require(path.join(projectPath, 'package.json'));
+
+        project = new Project(projectPath, packageContents);
+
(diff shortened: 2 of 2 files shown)


tests/unit/models/addon-test.js
```diff
@@ -407,6 +407,86 @@ describe('models/addon.js', function() {
...
+
+      it('returns false when `EMBER_ENV` is set to production, `EMBER_CLI_TEST_COMMAND` is unset and options.hinting is true', function() {
+        process.env.EMBER_ENV = 'production';
+        delete process.env.EMBER_CLI_TEST_COMMAND;
+
+        addon.app = {
+          options: { hinting: true }
+        };
+
+        expect(addon.hintingEnabled()).to.eql(false);
```
Turbo87: you can use `expect(...).to.be.false;` for better failure messages

jeffjewiss: Would it make sense to change all the `expect`s to use `.ok()` and `.not.ok()` since they're all tests as to whether or not the value is truthy?

Turbo87: it would be `.ok;`, not `.ok();`. chai is using properties here instead of functions and using them as function isn't actually running any assertions. I don't agree with using properties there, but that's clearly not my decision to make 😉 

I'm fine with `.ok` here too, although if we can be more specific with `.true/false` then we probably should use that. I'm not sure if there are cases where that method can return something trueish which isn't actually `true` itself.

lib/models/addon.js
```diff
@@ -131,8 +131,9 @@ var Addon = CoreObject.extend({
   hintingEnabled: function() {
     var isProduction = process.env.EMBER_ENV === 'production';
     var testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND || !isProduction;
+    var hintingDisabled = this.app && this.app.options && this.app.options.hinting === false;
```
nathanhammond: Pedantic, but can we call this `explicitlyDisabled`? Having `hintingEnabled` and `hintingDisabled` in the same method made me triple-check.

tests/unit/models/addon-test.js
```diff
@@ -407,6 +407,86 @@ describe('models/addon.js', function() {
       });
     });
 
+    describe('hintingEnabled', function() {
```
nathanhammond: Can you throw a doc block here describing each permutation you're testing? It's a matrix and it'd be useful to have a picture.

| configuration  | test 1 | test 2 |
| ------------- | ------------- | ---------- |
| hinting  | true  |
| environment  | false  |
| test_command  | set  |
| RESULT | true |

nathanhammond:
This looks great, I've added a couple of comments primarily for clarity and those of us who are going to be looking at this a year from now wondering what was going on. I look forward to landing this!

tests/unit/models/addon-test.js
jeffjewiss: Thanks for clarifying, I was looking at the Chai docs and not the other tests as a reference.

Yeah, I agree, I'll update them to follow the `.to.be.true` syntax.

lib/models/addon.js
jeffjewiss: Yeah no problem, I'll update this!

jeffjewiss:
Thanks @nathanhammond! I've made some updates based on the feedback, but I'm not sure if I structured the doc comment block correctly. Please let me know if I need to update it.

nathanhammond:
@jeffjewiss Thank you for your work on this!

nathanhammond:
@homu r+

PR #5831, opened 2016-04-21
Update "inquirer" to v1.0.2

Turbo87:
Now that SBoudrias/Inquirer.js#362 is fixed we can update our "inquirer" dependency to the v1.x release series.

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var ora              = require('ora');
-var Promise          = require('../ext/promise');
 var EOL              = require('os').EOL;
 var chalk            = require('chalk');
 var writeError       = require('./write-error');
@@ -192,17 +191,17 @@ UI.prototype.stopProgress = function() {
   }
 };
 
+/**
+ * @method writeLevelVisible
+ * @param {Object} questions
+ * @param {Function} [callback]
+ * @return {Promise}
+ */
 UI.prototype.prompt = function(questions, callback) {
   var inquirer = require('inquirer');
 
   // If no callback was provided, automatically return a promise
-  if (callback) {
-    inquirer.prompt(questions, callback);
-  } else {
-    return new Promise(function(resolve) {
-      inquirer.prompt(questions, resolve);
-    });
-  }
+  return inquirer.prompt(questions, callback);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/ui/index.js
```diff
@@ -194,11 +193,13 @@ UI.prototype.prompt = function(questions, callback) {
...
   // If no callback was provided, automatically return a promise
   if (callback) {
-    inquirer.prompt(questions, callback);
-  } else {
-    return new Promise(function(resolve) {
-      inquirer.prompt(questions, resolve);
+    inquirer.prompt(questions).then(function(result) {
+      callback(result);
+    }).catch(function() {
+      callback(null);
```
stefanpenner: do we really want to catch and swallow all exceptions?

Turbo87: this is essentially just mirroring the previous behavior, but I'm open to other ideas.

stefanpenner: i think we should rethrow / crash if there is an exception. No?

Turbo87: I tend to agree, I'm just not sure *how* to rethrow in this case. The old `callback` argument is not using the node `function(err, result)` signature so we can't pass any errors to the callback like that.

Turbo87: @stefanpenner any ideas?

twokul: @Turbo87 @stefanpenner it seems with new `1.0.3` release, we should be able to re-throw ([commit link](https://github.com/SBoudrias/Inquirer.js/commit/19f5e1a4cadb1cc0c73f54c68691f9b61d241f0a)).

so then something like this should work:

```js
function rethrow(reason) {
  setTimeout(function() {
    throw reason;
  }, 0);
}
```

Turbo87:
@stefanpenner apparently I missed that inquirer is still supporting the old callback interface too. this means that inquirer now has the exact same interface as our `prompt()` method so we can just pass along the arguments and the return value. I updated the PR to that and added some type info for the API docs.

stefanpenner:
@Turbo87 cool, seems like the tests are unhappy. Am i blocking something or ...

Turbo87:
@stefanpenner looks  like AppVeyor is somewhat flaky at the moment, but Travis seems fine with these changes

stefanpenner:
@Turbo87 :shipit:

stefanpenner:
@homu r+

Turbo87:
@homu are you going to test this any time soon?!

stefanpenner:
i <3 @homu but its having some troubles lately...

PR #5833, opened 2016-04-22
Simplify yuidoc invocation

Turbo87:
This PR replaces the previously complicated `yuidoc` invocation with a plain shell command. This was made possible by the `project_version_preprocessor.js` file, that injects the `emberCLIVersion()` output into the project options.

This also allows us to run `npm run docs -- --server` 🎉


rwjblue:
@homu r+

PR #5837, opened 2016-04-23
[ENHANCEMENT] Use lodash map to parse arguments + extract isProduction Check

ddrmanxbxfr:
Hi,

I've done two simple changes to help our code climate issues in the ember-app file.

1. Extract the production environment check into a simple function and use it.
1.1. Cleanup of yoda conditions on production check. !!isProduction is the same as checking for true.
2. Extract the arguments using lodash map instead of a for loop.

Thanks for looking into it.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -52,8 +52,8 @@ module.exports = EmberApp;
     - storeConfigInMeta, defaults to `true`,
     - autoRun, defaults to `true`,
     - outputPaths, defaults to `{}`,
-    - minifyCSS, defaults to `{enabled: !!isProduction,options: { relativeTo: 'assets' }},
-    - minifyJS, defaults to `{enabled: !!isProduction},
+    - minifyCSS, defaults to `{enabled: isProduction,options: { relativeTo: 'assets' }},
+    - minifyJS, defaults to `{enabled: isProduction},
     - sourcemaps, defaults to `{}`,
     - trees, defaults to `{},`
     - jshintrc, defaults to `{},`
@@ -63,21 +63,9 @@ module.exports = EmberApp;
   @constructor
   @param {Object} options Configuration options
 */
+
 function EmberApp() {
-  var args = [];
-  var options = {};
-
-  for (var i = 0, l = arguments.length; i < l; i++) {
-    args.push(arguments[i]);
-  }
-
-  if (args.length === 1) {
-    options = args[0];
-  } else if (args.length > 1) {
-    args.reverse();
-    options = defaults.apply(null, args);
-  }
-
+  var options = this._extractOptions(arguments);
   this._initProject(options);
 
   this.env  = EmberApp.env();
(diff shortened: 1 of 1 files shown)


lib/broccoli/ember-app.js
```diff
@@ -64,13 +65,9 @@ module.exports = EmberApp;
   @param {Object} options Configuration options
 */
 function EmberApp() {
-  var args = [];
+  var args = map(arguments, function(item) { return item });
```
stefanpenner: this would deopt the function (not that it really matters, but it sorta hurts my soul...)

lib/broccoli/ember-app.js
```diff
@@ -114,13 +109,20 @@ function EmberApp() {
 }
 
 /**
+ @private
+ @method _isProduction
+*/
+EmberApp.prototype._isProduction = function() {
```
stefanpenner: 👍

lib/broccoli/ember-app.js
ddrmanxbxfr: Let's rollback to the old implementation but extract them into separate function.

Let me know if it's better.

ddrmanxbxfr:
Alright, I gave an other go to make it cleaner while keeping the old structure for arguments.

Thanks for looking into it.

lib/broccoli/ember-app.js
```diff
@@ -63,21 +63,9 @@ module.exports = EmberApp;
...
-
-  if (args.length === 1) {
-    options = args[0];
-  } else if (args.length > 1) {
-    args.reverse();
-    options = defaults.apply(null, args);
-  }
 
+function EmberApp() {
+  var options = this._extractOptions(arguments);
```
stefanpenner: Unfortunately merely passing arguments to another function will cause the issue.

Again, in this case it really doesn't matter, my brain is just hard wired to avoid those cases... It's hard for me to let go.

@rwjblue I am over thinking this right?

ddrmanxbxfr: Doesn't shrinking the big function into smaller ones helps to inline them and optimize it ?

rwjblue: It is definitely not a hot path, but it makes sense to unroll the arguments manually as is done currently (lines 70 through 72).

stefanpenner: > Doesn't shrinking the big function into smaller ones helps to inline them and optimize it ?

It isn't that straight forward.

Ultimately this function doesn't need to be fast, so it should just be w/e is most clear to read. Ignore my above concerns about unrolling args here.

lib/broccoli/ember-app.js
```diff
@@ -114,13 +100,53 @@ function EmberApp() {
...
+  }
+  return args;
+}
+
+/**
+ @private
+ @method _extractOptions
+ @param {Object} rawArguments arguments passed to the cli
+*/
+EmberApp.prototype._extractOptions = function(rawArguments) {
```
stefanpenner: now that this is extracted, can we ad a unit test?

lib/broccoli/ember-app.js
```diff
@@ -114,13 +100,53 @@ function EmberApp() {
...
+  }
+
+  return options;
+}
+
+/**
+ @private
+ @method _isProduction
+*/
+EmberApp.prototype._isProduction = function() {
```
stefanpenner: ^ unit test

lib/broccoli/ember-app.js
ddrmanxbxfr: Yeup, i`ll look at it and ping you back when it`s ready :+1:

Turbo87:
this has been obsoleted by https://github.com/ember-cli/ember-cli/pull/5898 and https://github.com/ember-cli/ember-cli/pull/5928

Issue #5745, opened 2016-04-05
2.4.2 Regression involving treeForAddon


jskeate:
Seems like a Broccoli Filter runs in <=2.4.1, but stops working in 2.4.2.

Set up two new projects, one app, one addon.

In the app: change `app/templates/application.hbs` to `{{imported-component}}`

In the addon, generate `imported-component`:

```javascript
import Ember from 'ember';
import layout from '../templates/components/imported-component';

export default Ember.Component.extend({
  layout,

  modified: false
});
```

```handlebars
In imported-component... {{unless modified 'not'}} modified
```

Addon index.js:

```javascript
/* jshint node: true */
'use strict';

var Filter = require('broccoli-filter');

class TestFilter extends Filter {
  canProcessFile(relativePath) {
    return relativePath.indexOf('imported-component') !== -1;
  }

  processString(content, relativePath) {
    return content.replace(/false/g, 'true');
  }
}

module.exports = {
  name: 'xyz',

  isDevelopingAddon() {
    return true;
  },

  treeForAddon(tree) {
    tree = new TestFilter(tree);
    return this._super.treeForAddon.call(this, tree);
  }
};
```

ember-cli 2.4.1: get a page saying "In imported-component... modified"

ember-cli 2.4.2: get a page saying "In imported-component... not modified"

Turbo87:
we've traced this down to https://github.com/ember-cli/ember-cli/pull/5561 and https://github.com/ember-cli/ember-cli/blob/17619b8f949a200122df62b7a14c66ea7e68f932/lib/models/addon.js#L684-L686 in particular

PR #5840, opened 2016-04-25
Fix treeForAddon example

jasonmit:
The example below does not work in it's current form as the tree you pass into `this._super.treeForAddon` doesn't actually get processed and instead gets replaced.

This was introduced I believe in 2.4.2 and once the bug is fixed this example will continue to work as expected.

Related to https://github.com/ember-cli/ember-cli/issues/5745

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -436,8 +436,8 @@ var Addon = CoreObject.extend({
 
     @example
     ```js
-    treeForAddon: function(tree) {
-      tree = this._super.treeForAddon.call(this, tree);
+    treeForAddon: function() {
+      var tree = this._super.treeForAddon.apply(this, arguments);
       var checker = new VersionChecker(this);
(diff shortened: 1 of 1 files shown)


lib/models/addon.js
```diff
@@ -437,14 +437,15 @@ var Addon = CoreObject.extend({
     @example
     ```js
     treeForAddon: function(tree) {
+      tree = this._super.treeForAddon.call(this, tree);
```
Turbo87: this should probably be `this._super.treeForAddon.apply(this, arguments);` instead to make sure it won't forget to pass any additional arguments

jasonmit:
Done feel free to squash in the git merge

Turbo87:
@jasonmit awesome, thanks!

@homu r+

jasonmit:
Test failure looks unrelated

Turbo87:
@homu retry

jasonmit:
Thanks

PR #5842, opened 2016-04-25
Replace "findup" with "find-up"

Turbo87:
"find-up" by @sindresorhus seems to be better maintained and is more commonly used. also "findup" bundles a CLI tool which we don't use, but we still need to pay the cost for downloading their CLI dependencies.


rwjblue:
Seems good to me

Turbo87:
@homu r+

PR #5847, opened 2016-04-25
tests/errors/silent: Move deprecated require() call into test

Turbo87:
This moves the deprecation warning output closer to the corresponding test.

### DIFF
--- a/tests/unit/errors/silent-test.js
+++ b/tests/unit/errors/silent-test.js
@@ -1,11 +1,11 @@
 'use strict';
 
 var SilentError    = require('silent-error');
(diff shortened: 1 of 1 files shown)


rwjblue:
👍

rwjblue:
It would also be possible to allow customizing https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/deprecate.js#L5-L9 from the test (to avoid the console output).

Turbo87:
@rwjblue yeah, I thought about it, but decided it wasn't worth the effort for something that will probably be removed soon anyway

Turbo87:
@homu r+

PR #5851, opened 2016-04-27
Fix "find-up" return value handling

Turbo87:
#5842 unfortunately broke some code because it returns `null` if a file was not found instead of throwing an error like before. This PR adjusts the surrounding code to that changed behavior.

Unfortunately testing this is rather hard since we run our test suite inside the `ember-cli` folder which itself has a `package.json` file. That is also the reason why this bug wasn't caught by our current test suite.

/cc @nathanhammond

### DIFF
--- a/lib/models/addon-discovery.js
+++ b/lib/models/addon-discovery.js
@@ -133,7 +133,11 @@ var AddonDiscovery = CoreObject.extend({
 
       var entryModulePath = resolve.sync(packageName, { basedir: root });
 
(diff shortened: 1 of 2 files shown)


nathanhammond:
I really, really want to find a way to test this. Does something like `process.chdir('..')` allow us to do a smoke test?

nathanhammond:
I'm going to open an issue to track testing this and approve since canary is broken.

nathanhammond:
@homu r+

PR #5859, opened 2016-04-28
[BUGFIX beta] Bump Ember & Ember Data version to 2.6.0-beta

rwjblue:


Turbo87:
I thought we decided to stay on the stable versions?

rwjblue:
I discussed with @stefanpenner a few days ago about it. It is possible that I misunderstood, but since ember-cli@2.6.0 will install ember@2.6.0 it seems reasonable that using the betas of ember-cli@2.6 would also use the beta versions of Ember.

Turbo87:
does that mean that our canary should also use ember canary?

rwjblue:
In theory yes, but Ember canary is a bit too unstable at the moment (due to all the work being done on glimmer2) so I think we should stick to tagged releases.

Turbo87:
shouldn't this PR target `ember-cli:beta`?

rwjblue:
@Turbo87 - All changes to beta go through master first.

Turbo87:
hmm okay... the way I've usually worked in most other projects is regularly merging `release/stable/beta` back into `master` to prevent them from falling out of sync. any advantages/disadvantages I should be aware of?

rwjblue:
@Turbo87 - Due to our train model where all changes flow from canary down into the final release, we need to ensure that changes also follow the same pattern. In this case, I made a mistake when doing the initial 2.6.0-beta.1 release as these changes should have been included there to begin with (and no cherry-picking or merging would have been needed).  In most other cases, we want the fixes to land on canary so they can be tested by those brave souls running canary before rolling things into beta or release branches.

rwjblue:
I'm happy to discuss a different pattern for doing these sorts of things going forward, but I'd like to land this for now....

rwjblue:
@homu r+

Turbo87:
> I'm happy to discuss a different pattern for doing these sorts of things going forward, but I'd like to land this for now....

sure, I didn't intend to block this. was just curious :)

PR #5867, opened 2016-04-29
models/addon: Kill the addonJsFiles() cache

Turbo87:
addonJsFiles() is called twice, by jshintAddonTree() and by treeForAddon(). in one case it gets passed the addonPath directly, while the other case might be overwritten and passes something else. If the Funnel in the first invocation is cached the second invocation parameter will be discarded which resulted in a regression.

Resolves https://github.com/ember-cli/ember-cli/issues/5745

/cc @rwjblue


rwjblue:
👍

rwjblue:
@homu r+

PR #5869, opened 2016-04-29
models/project: Remove "testing" code path from closestSync()

Turbo87:
Instead we check if pathName is in the "tmp" folder of the parent project and in this case we assume we're *not* inside a project.

This codepath was originally created for the blueprint-test-helpers and the proposed solution works for them as well.


stefanpenner:
@trabus will this work for you codez?

trabus:
LGTM :+1:

stefanpenner:
@homu r+

PR #5870, opened 2016-04-30
Ignore fastboot-dist

jonnii:
I've been experiment with fast boot and one of the first thing I need to do in every repo is add fastboot-dist to the gitignore.


stefanpenner:
Addons should bring there own additions to these files

jonnii:
@stefanpenner makes sense. what's the mechanism for that to happen? I'll try to get that added to `ember-fastboot-cli`.

PR #5877, opened 2016-05-03
Uniform import API

ef4:
This PR proposes a uniform API for importing assets into both apps and addons. It solves the most common pain point when using an addon from another addon, which is that `app.import` in the `included` hook fails.

This API gives us future flexibility to implement `import` differently depending on target you're importing into. In particular, an engine may choose to behave differently  so that vendored assets are in the appropriate bundles.

I'm opening this for discussion now, and will add tests if people think this is mergeable.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -391,6 +391,27 @@ var Addon = CoreObject.extend({
   },
 
   /**
+     Imports an asset into this addon.
+
+     Options:
+     - type - Either 'vendor' or 'test', defaults to 'vendor'
+     - prepend - Whether or not this asset should be prepended, defaults to false
+     - destDir - Destination directory, defaults to the name of the directory the asset is in
+
+     @public
+     @method import
+     @param  {(Object|String)}  asset   Either a path to the asset or an object with environment names and paths as key-value pairs.
+     @param  {Object=} options Options object
+  */
+  import: function(asset, options) {
+    var app = this.app;
+    while (app.app) {
+      app = app.app;
+    }
+    app.import(asset, options);
+  },
+
+  /**
     Returns the tree for all app files
 
     @public


rwjblue:
I am in favor of this. If we decide to land it, we should make a polyfill that works back through the 2.x series that folks can use so that we can start moving forward sooner.

stefanpenner:
The reason why we don't support nested add-ons importing, is we have no way to handle dupes.

ef4:
@stefanpenner I agree, but I think the API I'm proposing here is congruous with that future better implementation. The problem I'm trying to solve right now is that addon authors are commonly doing

```js
included(app) {
  if (typeof app.import !== 'function' && app.app) {
    app = app.app;
  }
  app.import(...);
}
```

which is future-hostile, because it effectively hides the nested import from ember-cli, and will cause ecosystem-wide pain once ember-cli adds first-class support. Whereas the API in this PR gives ember-cli the opportunity to implement the right thing without asking module authors to all make changes.

stefanpenner:
@ef4 so the idea is, people are doing this anyways. Lets give them a public API so we can detect it? If so this seems reasonable.

stefanpenner:
@ef4 / @rwjblue lets not forget about the compat shim

ef4:
@rwjblue were you picturing an addon that monkey-patches this API into older ember-cli versions? I can do that.

stefanpenner:
> @rwjblue were you picturing an addon that monkey-patches this API into older ember-cli versions? I can do that.

i believe so

rwjblue:
Yes, exactly. A polyfill to monkey patch down to all ember-cli@2 versions...

stefanpenner:
@ef4 we should make sure it doesn't add if not needed, or double add etc.

ef4:
@rwjblue @stefanpenner please see https://github.com/ember-cli/ember-cli-import-polyfill

stefanpenner:
Looks good

elwayman02:
How can this method work with situations where you want to pull `bowerDirectory` from `app` for the asset path?

Before you might do `app.import(app.bowerDirectory + 'foo/dist/foo.js');`, but now in situations where `app` is not guaranteed to be the root, is `app.bowerDirectory` still going to be applicable for `this.import(app.bowerDirectory + 'foo/dist/foo.js');`?

knownasilya:
> How can this method work with situations where you want to pull bowerDirectory from app for the asset path?

Just ran into this same situation here https://github.com/dgeb/ember-engines/issues/226

stefanpenner:
I'm pretty sure we should seriously consider deprecating and reverting this, as (as i was worried about originally and i really regret conceding to this...) we merely enable many other 👣  🔫 's

nathanhammond:
I'm in agreement with @stefanpenner. The more I work with nested addons the more I think that we need to revisit `.import`'s use (and existence). Note that I make no distinction between `this.import` and `app.import` as they're equivalent: you can reach `app.import` using only public APIs.

Unfortunately I don't think we can call for deprecation of `.import` holistically without having an incredibly complete specification of how people should accomplish the tasks for which we presently rely on it.

ef4:
I think all the footguns were already present and this didn't make it any worse. Before this the alternative was to just call app.app.app.import manually.

I would be hesitant to deprecate this before providing a fully-designed alternative. As we've learned the hard way before, it's not helpful to deprecate a thing unless there is a good alternative to tell people to migrate to.

@stefanpenner I could work on a nested addons RFC, but I would need to pick your brain first.

PR #5883, opened 2016-05-06
Fix reading npm package names with a version specifier

### LINKED ISSUE #5880: "ember install package@version" fails after npm
### What went wrong

I (and I think most users) expect `ember install` to match reasonably with `npm install` in terms of support for the way packages are defined. For example, I would expect both of these to result in the same version of the package being added to my project:

```sh
npm install ember-cp-validations@~2.7.0
ember install ember-cp-validations@~2.7.0
```

I'm now seeing that what happens if you attempt the `ember install` above is that the modification is made to `package.json`, but then the user gets an error

```
Install failed. Could not find addon with name: ember-cp-validations@~2.7.0
```

### Suspected cause

To me, this smells like we're losing the distinction between name and "name@target" at the time we attempt to run an installation blueprint. Sure enough, even after you get the above error, if you run the blueprint explicitly

```
ember g ember-cp-validations
```

Everything is run as you would expect.

### This used to work fine

- I tested with ember-cli `1.13.8` and `2.5.0`, and this bug doesn't appear to be present

### To reproduce

```sh
ember new ember123 && \
   cd ember123 && \
   npm install --save-dev ember-cli@2.6.0-beta.2 && \
   ember install ember-cp-validations@~2.7.0
```
 
```
ember-cli: 2.6.0-beta.2
http_parser: 2.6.2
node: 5.10.1
v8: 4.6.85.31
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 47
openssl: 1.0.2g
os: darwin x64
```

martndemus:
Fixes #5880

### DIFF
--- a/lib/utilities/get-package-base-name.js
+++ b/lib/utilities/get-package-base-name.js
@@ -1,12 +1,21 @@
 'use strict';
 
+var npa = require('npm-package-arg');
+
+
 module.exports = function (name) {
-  var packageParts;
+  var parsed, parsedName
 
   if (!name) {
     return null;
   }
 
-  packageParts = name.split('/');
-  return packageParts[(packageParts.length - 1)];
+  parsed = npa(name);
+  parsedName = parsed.name;
+
+  if (parsed.scope) {
+    parsedName = parsedName.replace(parsed.scope, '').slice(1);
+  }
+
+  return parsedName;
 };
--- a/tests/unit/commands/install-test.js
+++ b/tests/unit/commands/install-test.js
@@ -156,16 +156,6 @@ describe('install command', function() {
       });
     });
 
-    it('runs the package name blueprint task when given github/name and args', function() {
-      return command.validateAndRun(['ember-cli/ember-cli-qunit']).then(function() {
-        var generateRun = tasks.GenerateFromBlueprint.prototype.run;
-        expect(generateRun.calledWith[0][0].ignoreMissingMain).to.be.true;
-        expect(generateRun.calledWith[0][0].args).to.deep.equal([
-          'ember-cli-qunit'
-        ], 'expected generate blueprint called with correct args');
-      });
-    });
-
     it('gives helpful message if it can\'t find the addon', function() {
       return command.validateAndRun(['unknown-addon']).then(function() {
         expect(false, 'should reject with error').to.be.ok;
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -465,11 +465,6 @@ describe('models/project.js', function() {
       var addon = project.findAddonByName('foo-ba');
       expect(addon).to.equal(undefined, 'foo-ba should not be found');
     })
-
-    it('should return an addon that is the unqualified name of requested name', function() {
-      var addon = project.findAddonByName('qux/foo');
-      expect(addon.name).to.equal('foo', 'should have found the foo addon');
-    })
   });
 
   describe('bowerDirectory', function() {
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -525,17 +525,14 @@ Project.prototype.reloadAddons = function() {
 Project.prototype.findAddonByName = function(name) {
   this.initializeAddons();
 
-  var unqualifiedName = name;
-  if (name.indexOf('/') !== -1) {
-    unqualifiedName = name.slice(name.lastIndexOf('/') + 1);
-  }
+  var baseName = getPackageBaseName(name);
 
   function matchAddon(name, addon) {
     return name === addon.name || (addon.pkg && name === addon.pkg.name);
   }
 
   return find(this.addons, function(addon) {
-    return matchAddon(name, addon) || matchAddon(unqualifiedName, addon);
+    return matchAddon(name, addon) || matchAddon(baseName, addon);
   });
 };
 
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "node-uuid": "^1.4.3",
     "nopt": "^3.0.1",
     "npm": "2.15.5",
+    "npm-package-arg": "^4.1.1",
     "ora": "^0.2.0",
     "portfinder": "^1.0.3",
     "promise-map-series": "^0.2.1",
--- a/tests/unit/utilities/get-package-base-name-test.js
+++ b/tests/unit/utilities/get-package-base-name-test.js
@@ -11,4 +11,8 @@ describe('getPackageBaseName', function() {
   it('should return the package name without its scope', function() {
     expect(getPackageBaseName('@scope/my-addon')).to.equal('my-addon');
   });
+
+  it('should strip away version numbers', function() {
+    expect(getPackageBaseName('my-addon@~1.2.0')).to.equal('my-addon');
+  });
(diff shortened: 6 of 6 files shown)


Turbo87:
@mike-north can you verify that this works as expected now?

mike-north:
Ok, looks like the following works now

```
ember install ember-cp-validations@~2.7.0
OR
ember install ember-cp-validations@mike-north/ember-cp-validations#master
```

But, I think a couple forms of URL and repo-based target URLS still cause problems.
For example, I believe that the following should be valid:

```
ember install ember-cp-validations@git+http://git@github.com/mike-north/ember-cp-validations.git#master
```
And it seems like we still get the incorrect name resolved 

```
Install failed. Could not find addon with name: ember-cp-validations@git+http://git@github.com/mike-north/ember-cp-validations.git#master
```

see: https://docs.npmjs.com/files/package.json#git-urls-as-dependencies for examples of test cases. I'm not sure if all of these worked in ember-cli 2.5, but we should get them all working for parity with all npm package target flavors

martndemus:
So basically for this one `ember-cp-validations@git+http://git@github.com/mike-north/ember-cp-validations.git#master` I would have to somehow pick the `ember-cp-validations` from the leading part?
Is that leading name with an `@` a Ember CLI thing?

rwjblue:
No, the following works properly: 

```
% npm install --save glob@git+http://git@github.com:isaacs/node-glob.git#master
```

mike-north:
@martndemus you should make sure that if I have something like this in a blueprint's `afterInstall` hook

```js
this.addAddonsToProject({packages: [
  { name: 'ember-cp-validations',
    target: 'git+http://git@github.com/mike-north/ember-cp-validations.git#master'
  }
]});
```
or this

```js
this.addAddonToProject('ember-cp-validations@git+http://git@github.com/mike-north/ember-cp-validations.git#master');
```

or run this in the terminal

```sh
ember install ember-cp-validations@git+http://git@github.com:mike-north/ember-cp-validations.git#master
```

That the installation blueprint ember-cli will attempt to run is _still_ the one called `ember-cp-validations`

martndemus:
@mike-north Updated to now use `npm-package-arg`, with a little extra code this should now reasonably reliably get the right package name.

Turbo87:
LGTM

@rwjblue @stefanpenner r?

stefanpenner:
seems good, would like @mike-north r? as he actually helped us with the original code.

Turbo87:
@mike-north any comments?

stefanpenner:
I chatted with mike earlier, this week. He is quite busy, and most likely wont be able to dig in.

stefanpenner:
I believe this should most likely be a `[Bugfix release]`

stefanpenner:
> I chatted with mike earlier, this week. He is quite busy, and most likely wont be able to dig in.

@mike-north will have some cycles this afternoon, then we can wrap this up.

Turbo87:
I'd recommend keeping this on beta, otherwise we'd have to backport quite a few other things too. Beta will be promoted to release soon anyway...

mike-north:
This works for the use cases that originally motivated me to report the bug.

lib/utilities/get-package-base-name.js
```diff
@@ -1,12 +1,34 @@
...
+var npa = require('npm-package-arg');
+
+function fromHostedName(name) {
+  var packageParts, namePart, nameAndVersionParts;
+
+  packageParts = name.split('/');
+  namePart = packageParts[(packageParts.length - 1)];
+  nameAndVersionParts = namePart.split(/[#]/);
+
+  return nameAndVersionParts[0].replace(/.git$/, '');
```
mike-north: Is there any way to avoid string splitting? This is the kind of thing that has bitten us in the past -- typically we forget to cover some officially-supported use cases

nathanhammond: The npm methodology for testing this has already given up and not figured out a name. Next steps is that it would try to install it and see if there is a name specified in a package.json inside the .tgz or directory, or bail.

If we care a ton we can implement that. Thoughts?

nathanhammond:
At this point we should either do it right, or call this "good enough." We literally can't assume anything about the package name from a path spec.

```bash
mkdir foo
cd foo
npm init # name = "troll"
cd ../
mkdir test
npm install --save ../foo
cat package.json
# name = troll
```

Thoughts?

Turbo87:
@nathanhammond is this code only used for figuring out which blueprint to run after install? if so then we should be able to just show a warning that we couldn't figure out the package name and that the user might have to run the blueprint manually via `ember generate <addon-name>`.

nathanhammond:
@Turbo87 I'm in favor of that solution, which would mean we can drop the entire `fromHostedName` function. If we can't determine a package name we just fail to install the blueprint, and then provide a notice that they will have to run `ember generate <addon-name>`.

@martndemus Can you make all of your hard work less clever and write a good error message?

martndemus:
@nathanhammond Is there a preferred way of raising an error in Ember-CLI?

Turbo87:
@martndemus see https://github.com/ember-cli/ember-cli/blob/master/lib/ui/index.js#L107-L121

martndemus:
Removed the fancy stuff, still need to add an error message.

tests/unit/commands/install-test.js
```diff
@@ -156,16 +156,6 @@ describe('install command', function() {
       });
     });
 
-    it('runs the package name blueprint task when given github/name and args', function() {
-      return command.validateAndRun(['ember-cli/ember-cli-qunit']).then(function() {
```
stefanpenner: can you briefly describe why these are missing, i may be missing sufficient context

tests/unit/models/project-test.js
```diff
@@ -465,11 +465,6 @@ describe('models/project.js', function() {
       var addon = project.findAddonByName('foo-ba');
       expect(addon).to.equal(undefined, 'foo-ba should not be found');
     })
-
-    it('should return an addon that is the unqualified name of requested name', function() {
```
stefanpenner: can you briefly describe why these are missing, i may be missing sufficient context

nathanhammond: The goal is to only run the default blueprint if we can successfully parse a name out of the specified string using `node-package-args`.

We'll still run the `npm install` for things we don't recognize, but throw a console warning you that, "hey, we don't know your addon name, so you'll need to manually invoke your addon's generator blueprint."

nathanhammond: (This was testing package name inference that we don't do anymore.)

nathanhammond:
@martndemus Can you take a swing at the error message today? This is the last thing on our list for a 2.6 release. If not I can take it from here.

martndemus:
@nathanhammond Best I can do is tomorrow, but :+1: if you want to pick this up and finish it!

tests/unit/commands/install-test.js
Turbo87: we should probably keep the test, but then test for the yet-to-be-implemented warning message

Turbo87:
Resolved by #5963

PR #5891, opened 2016-05-09
Use jshint's esversion option instead of esnext option

san650:
esnext option is deprecated and will be removed. See http://jshint.com/docs/options/#esnext


stefanpenner:
@homu r+

Issue #3225, opened 2015-02-07
ember init should not error if no bower.json is present


stefanpenner:
it just should skip the bower install phase.

abuiles:
It looks like it doesn't error but adds again the file

stefanpenner:
we should defer this to a larger remove bower story.

PR #5892, opened 2016-05-10
Update bower-install.js

### LINKED ISSUE #5889: Install addons behind a proxy
I'm having trouble with installing addons through ember-cli.  
I tried to install the addon ember-cli-pace.  
Due to the restrictions of my company i am forced to use a private bower cache with proxy  
However `ember install ember-cli-pace` generates the following output:  

```
ember install ember-cli-pace                                                                                     [15:20:14]
(node:16657) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
WARNING: WARNING: Node v6.0.0 has currently not been tested against Ember CLI and may result in unexpected behaviour.
Could not start watchman; falling back to NodeWatcher for file system events.
Visit http://ember-cli.com/user-guide/#watchman for more info.
installing pace
  install bower package pace
Request to https://bower.herokuapp.com/packages/pace failed: getaddrinfo ENOTFOUND bower.herokuapp.com bower.herokuapp.com:443
Error: Request to https://bower.herokuapp.com/packages/pace failed: getaddrinfo ENOTFOUND bower.herokuapp.com bower.herokuapp.com:443
    at createError (/home/maurice/workspace/secretStuf/node_modules/bower/lib/node_modules/bower-registry-client/lib/util/createError.js:2:15)
    at Request._callback (/home/maurice/workspace/secretStuf/node_modules/bower/lib/node_modules/bower-registry-client/lib/lookup.js:97:29)
    at self.callback (/home/maurice/workspace/secretStuf/node_modules/bower/lib/node_modules/request/request.js:198:22)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at Request.request.emit (/home/maurice/workspace/secretStuf/node_modules/bower/lib/node_modules/request-replay/index.js:69:33)
    at Request.onRequestError (/home/maurice/workspace/secretStuf/node_modules/bower/lib/node_modules/request/request.js:820:8)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at TLSSocket.socketErrorListener (_http_client.js:306:9)

```

Using bower manually (`bower install pace`) works fine. But it doesn't solve my problem as ember-cli wants to install it anyway.  

```
bower i pace --save --verbose
bower pace#*                not-cached git://mirror.secret.xxx:6789/pace#*
bower pace#*                   resolve git://mirror.secret.xxx:6789/pace#*
bower pace#*                  checkout v1.0.2
bower pace#*              invalid-meta The "name" is recommended to be lowercase, can contain digits, dots, dashes
bower pace#*                  resolved git://mirror.secret.xxx:6789/pace#1.0.2
```

The proxy in bower is set through the .bowerrc file which is located in the project root folder.
The "proxy" is just a private registry which act as an proxy.

```
cat .bowerrc

{
  "directory": "bower_components",
  "analytics": false,
  "registry": "http://mirror.secret.xxx:5678"
}
```

Is there any config I missed?

---

```
ember version --verbose  
(node:12494) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
WARNING: WARNING: Node v6.0.0 has currently not been tested against Ember CLI and may result in unexpected behaviour.
Could not start watchman; falling back to NodeWatcher for file system events.
Visit http://ember-cli.com/user-guide/#watchman for more info.
ember-cli: 2.5.0
http_parser: 2.7.0
node: 6.0.0
v8: 5.0.71.35
uv: 1.9.0
zlib: 1.2.8
ares: 1.10.1-DEV
modules: 48
openssl: 1.0.2g
os: linux x64

```

mohlek:
Passing the working dir fixes #5889


lib/tasks/bower-install.js
```diff
@@ -21,7 +22,7 @@ module.exports = Task.extend({
 
     ui.startProgress(chalk.green('Installing browser packages via Bower'), chalk.green('.'));
 
-    var config = bowerConfig.read();
+    var config = bowerConfig.read(path.resolve('.'));
```
Turbo87: I'm wondering if we should use `this.project.root` here instead? Would that work for you too?

mohlek: Yeah that works great. I didn't knew that exists. I saw that path.resolve pattern somewhere else in the ember-cli project and adopted it.

Turbo87:
it looks like this is failing our test suite. I guess that it breaks for `ember new` when there is no `bower.json` file present yet.

mohlek:
I don't how this change conflicts with the bower.json as my change only affects the lookup path for the .bowerrc file. Strange

mohlek:
Seems like this is a somewhat known bug: #3225
Although I don't know why it was only triggered after my changes.

mohlek:
Apparently `this.project.root` is `~/ember-cli/common-tmp` and not

- `~/ember-cli/common-tmp/some-cool-app` as it should be when running `npm run test tests/acceptance/smoke-test-slow.js` or
- `~/ember-cli/common-tmp/some-cool-addon` when running `npm run test tests/acceptance/addon-smoke-test-slow.js`

I suppose this is a pre-existing bug in ember-cli, or am I misinterpreting the purpose of this variable?

Turbo87:
> I suppose this is a pre-existing bug in ember-cli

yeah, I tend to agree with you on that... I guess `path.resolve('.')` or `process.cwd()` might be our best option for now

mohlek:
But won't this lead to problems when running ember-cli from a project subdirectory (like `app/components/x-foo`)?

Turbo87:
that could indeed be a problem, yeah...

@stefanpenner @rwjblue any thoughts on this issue?

rwjblue:
Why is `this.project.root` the parent dir for NULL_PROJECT?  I am guessing that it is because we have not created the new directory yet.  Perhaps we should update `project.root` once we have created the directory?

Turbo87:
actually we should probably replace the whole project once it has been created... this looks like a rather involved cleanup though

stefanpenner:
@mike-north can we have your 👀  on this?

Turbo87:
tbieniek [19:42]  
did you understand the problem with the current state of the PR? (edited)

locks [19:42]  
barely

tbieniek [19:42]  
okay

[19:42]  
so

[19:42]  
CLI can be run in two "modes"

[19:42]  
inside of a project, and outside

[19:43]  
if it's outside it will use the `NULL_PROJECT` as `this.project`

[19:43]  
otherwise it will create a project based on the project you're in

[19:44]  
now during `ember new` you are operating outside of a project, so `NULL_PROJECT` is used

[19:44]  
and in that case `this.project.root` points to the folder in which you're creating the new project

[19:45]  
and the `bower install` step is in this case still operating against the `NULL_PROJECT`

[19:45]  
but when you `ember install something` inside a project you could also trigger `bower install`, in that case it's using the actual project

[19:46]  
and that is why `this.project.root` is not always pointing to the correct location during `bower install`

[19:47]  
one potential solution would be to change `this.project` to the newly created project, but since some addon may cache that somewhere that might not work so great

[19:47]  
was that roughly understandable?

nathanhammond [20:00]  
Can you copy that into the issue so we don't lose it to the backlog?

Turbo87:
Bower support in Ember CLI is now somewhat frozen and shouldn't be used anymore if possible so I'll go ahead and close this PR

PR #5895, opened 2016-05-11
Generate JSHint files only if JSHint is used

Turbo87:
one more step towards deprecating JSHint...

### DIFF
--- a/blueprints/lib/index.js
+++ b/blueprints/lib/index.js
@@ -1,5 +1,22 @@
+var fs = require('fs-extra');
+
 module.exports = {
   description: 'Generates a lib directory for in-repo addons.',
 
-  normalizeEntityName: function(name) { return name; }
+  normalizeEntityName: function(name) { return name; },
+
+  beforeInstall: function() {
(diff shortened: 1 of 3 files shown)


blueprints/lib/index.js
```diff
@@ -1,5 +1,22 @@
...
+
 module.exports = {
   description: 'Generates a lib directory for in-repo addons.',
 
-  normalizeEntityName: function(name) { return name; }
+  normalizeEntityName: function(name) { return name; },
+
+  beforeInstall: function() {
+    // make sure to create `lib` directory even if .jshintrc is not created
+    fs.mkdirsSync('lib');
```
stefanpenner: should this add a `.gitkeep` ?

Turbo87: It should, but it can't because gitkeeps have special handling currently and won't be generated (which is a bug in itself IMHO)

Turbo87:
@homu r+

Turbo87:
@homu retry

PR #5912, opened 2016-05-20
remove slow express-server-restart tests

stefanpenner:
* We have sufficient unit test coverage.
* if issues are discovered regression tests should remove unit covered.


Turbo87:
Since the test have been disabled for some time already I think we can safely remove it. 

@homu r+

stefanpenner:
> Since the test have been disabled for some time already I think we can safely remove it.

yup.

Turbo87:
the breaking tests seem completely unrelated... merging manually

PR #5914, opened 2016-05-21
Set rootURL to / in test environment

jasonmit:
Now that this was merged https://github.com/emberjs/ember.js/pull/13520 it's important people know that setting the `rootURL` on line 7 will likely cause some breakage in tests.  Everyone likely relied on `rootURL` to not be implemented in `Ember.NoneLocation` and now that it is it will likely cause some confusion.


Turbo87:
I'm not sure if it's really necessary, but I'm fine with merging this.

jasonmit:
I think you're right, just ran through all the scenarios and all seem to be fine.

PR #5918, opened 2016-05-22
models/project: Look only for projects with "ember-cli" dependency

Turbo87:
resolves https://github.com/ember-cli/ember-cli/issues/5493


stefanpenner:
@homu r+

PR #5920, opened 2016-05-22
move jQuery to ember-core for now

stefanpenner:
this is backwards compat, merely if someone has ember-core installed we will prefer that jQuery


stefanpenner:
@homu r+

Kerrick:
@stefanpenner Is there no way to pin a certain jQuery version anymore if we're using ember-core?

PR #5931, opened 2016-05-25
Extract standalone insertIntoFile() function

Turbo87:
This PR extracts the `insertIntoFile()` method from the `Blueprint` class into its own utility file to reduce the complexity of the class.

Ultimately we might want to extract this function into its own project and deprecate its use through the `Blueprint` class but this seems like a good first step for now.

/cc @trabus


nathanhammond:
I'm 50/50 on this. It's clearly better-factored, but I'm hesitant to create a de-facto public API that somebody can use. The fewer public APIs we have the easier it will be to change things later if we don't see this as the "perfect" solution.

Turbo87:
I don't understand, this is already public API which is actually one of the things I don't like about it...

rwjblue:
Agreed. We have to support this method either way, this PR seems like a clear win over the current implementation...

Turbo87:
@homu r+

nathanhammond:
My main thought was that by extracting it into a util people might import and use it in places where we don't want them to do so. Seems fine. :)

Turbo87:
> Ultimately we might want to extract this function into its own project and deprecate its use through the Blueprint class but this seems like a good first step for now.

@nathanhammond if people want to use it then they should be able to use it elsewhere too. what I don't like about it though is that it clutters the already rather complex Ember CLI codebase while it is actually not related to the CLI tool at all.

@homu retry

Turbo87:
wtf... how can the `controller` blueprint be missing and only in that single Travis run?!?

@homu retry

Turbo87:
💣 

@homu retry

Turbo87:
🙏

PR #5909, opened 2016-05-20
Handle all valid URL path characters for rootURL/baseURL in tests-server

les2:
The previous implementation used a regular expression for matching the
beginning of the request path against the baseURL + 'tests' but did
not escape the baseURL. Because RFC 3986 allows some characters that
are not valid in regexes, this caused problems when using them for
the baseURL.

Closes #5876

PR #5933, opened 2016-05-26
"tests-server" test cleanup

Turbo87:
This PR cleans up the "tests-server" unit test now that #5909 is merged.

### DIFF
--- a/tests/unit/tasks/server/middleware/tests-server-test.js
+++ b/tests/unit/tasks/server/middleware/tests-server-test.js
@@ -31,13 +31,17 @@ describe('TestServerAddon', function () {
           watcher: Promise.resolve()
         },
         finally: function() {
-          expect(nextWasCalled).to.true;
-          done();
+          try {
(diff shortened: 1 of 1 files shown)


tests/unit/tasks/server/middleware/tests-server-test.js
```diff
@@ -37,13 +31,17 @@ describe('TestServerAddon', function () {
           watcher: Promise.resolve()
         },
         finally: function() {
-          expect(nextWasCalled).to.true;
-          done();
+          try {
```
stefanpenner: good catch

stefanpenner:
@homu r+

tests/unit/tasks/server/middleware/tests-server-test.js
Turbo87: there might be a better pattern for this, but I couldn't think of one while doing those commits...

PR #5947, opened 2016-06-02
Tell greenkeeper to ignore default packages

leo:
After using ember-cli in combination with greenkeeper on my projects for a while, I've noticed that it's getting very irritating that greenkeeper is trying to update the default packages of the blueprint.

If you ask me, this shouldn't be the case. Greenkeeper should only be responsible for packages that have been added manually and that aren't part of the default blueprint.

This will ensure that people use the [update instructions](https://github.com/ember-cli/ember-cli/releases/latest) when a new version comes out (which prevents a lot of errors, compared to updating one of the default packages separately).

But if someone wants to do it like that nevertheless, they could still remove the package from the "ignore" property added in the commit of this PR.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -40,5 +40,30 @@
     "ember-resolver": "^2.0.3",
     "ember-welcome-page": "^1.0.1",
     "loader.js": "^4.0.1"
+  },
+  "greenkeeper": {
+    "ignore": [
+      "broccoli-asset-rev",
+      "ember-ajax",
+      "ember-cli",
+      "ember-cli-app-version",
+      "ember-cli-babel",
+      "ember-cli-dependency-checker",
+      "ember-cli-htmlbars",
+      "ember-cli-htmlbars-inline-precompile",
+      "ember-cli-inject-live-reload",
+      "ember-cli-jshint",
+      "ember-cli-qunit",
+      "ember-cli-release",
+      "ember-cli-sri",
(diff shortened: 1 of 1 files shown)


rwjblue:
We should be properly following SemVer for all of these packages, and therefore it should be fine to allow GreenKeeper to confirm new versions within the allowed range pass our tests.  

I suppose we could attempt to instruct Greenkeeper to not look for versions outside the allowed range for these specific packages, but I'm not sure that is something they support.

leo:
@rwjblue I'm mostly worried that the version of the CLI installed on your device (globally) could different from the instance within your application (after greenkeeper updated it). So if your global instance of ember-cli is older than the packages in your app, you'll get problems.

Or does the global instance pass control to the local one (like gulp does it)? If not, I think we should fix that problem by adding the "ignore" property...

rwjblue:
> I'm mostly worried that the version of the CLI installed on your device (globally) could differ from the instance within your application (after greenkeeper updated it).

The global version of ember-cli always defers to the local one.

> So if your global instance of ember-cli is older than the packages in your app, you'll get problems.

Since the project local version is always used, this is absolutely not a problem.

leo:
@rwjblue So it's recommended to update the packages separately, if there's a new version available? If so, I guess we can close this issue... :blush:

And no, I also don't think they support not looking for versions outside the allowed range.

Turbo87:
I tend to agree with @rwjblue that this shouldn't be necessary

PR #5952, opened 2016-06-03
`insertInto` - insert at the beginning of the file

pablobm:
### Use case

I'm working on a blueprint that (among other things) inserts an `import` statement at the beginning of a file. At the moment, the options for `insertIntoFile` don't allow doing this cleanly.

For example, take an original file such as this one:

```js
export default function() {
  // ...
}
```

I want my generator to be able to turn it into this:

```js
import SomeModule from 'some-module';

export default function() {
  // ...
}
```

I could do it by doing `insertIntoFile(path, myInsert, { before: 'export default' })`, but the risk is that the user may have changed the `export` line in some way that leaves me looking for a red herring (eg: if the user added support functions before the export).

The ability to simply insert at the beginning of the file is much more natural and less error prone in this case.

### Implementation

Using the approach on this PR, client code can simply specify an empty string as `before` option:

```js
insertIntoFile(path, myInsert, { before: '' })
```

Since an empty string is always found at the beginning of another string (by `indexOf` anyway), the code for this behaviour is almost there. We only need to be a bit more picky with what we consider falsy when it comes to reading `before`.

So what do you think?

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -4,7 +4,6 @@ var fs = require('fs-extra');
 var existsSync = require('exists-sync');
 var EOL = require('os').EOL;
 var Promise = require('../ext/promise');
-var util = require('util');
 
 var writeFile = Promise.denodeify(fs.outputFile);
 
@@ -66,8 +65,8 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
   var insert            = !alreadyPresent;
   var insertBehavior    = 'end';
 
-  if (util.isString(options.before)) { insertBehavior = 'before'; }
-  if (util.isString(options.after))  { insertBehavior = 'after'; }
+  if (typeof options.before === 'string') { insertBehavior = 'before'; }
+  if (typeof options.after === 'string')  { insertBehavior = 'after'; }
 
   if (options.force) { insert = true; }
 
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -14,10 +14,13 @@ var writeFile = Promise.denodeify(fs.outputFile);
   is passed.
 
   If `options.before` is specified, `contentsToInsert` will be inserted before
-  the first instance of that string.  If `options.after` is specified, the
-  contents will be inserted after the first instance of that string.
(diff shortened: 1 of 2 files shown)


stefanpenner:
im unsure about overloading before after this way, feels abit too clever

Turbo87:
how about this instead:

```js
var fs = require('fs');

var content = fs.readFileSync(this.project.root + '/path/to/your/file.js', 'utf8');
content = "import foo from 'bar';\n" + content;
fs.writeFileSync(this.project.root + '/path/to/your/file.js', content);
```

pablobm:
I agree it does look a bit too clever. I chose this approach for two reasons. One, it was a simple way to start a conversation on whether this functionality should exist at all. Two, `before: ''` looked sort of natural because it was the first thing I tried when I had this problem (only to see it fail). I wondered if other people had tried the same.

@Turbo87: I understand you mean not adding this functionality, and just doing it manually? Definitely could. I just found it a bit disappointing that `insertIntoFile` can't do what one could think is the simplest insertion possible. (Ok, maybe second to insertion at the end).

In my mind, a developer would want to do this, would try to make it work (including researching/asking online if there's something wrong), only to ultimately implementing it manually, having spent an hour in the process. That's what happened to me, but of course that makes me biased :-)

I'm happy with either option: implementing it differently (perhaps with an specialized option instead of `before`), or dropping it altogether.

Turbo87:
> @Turbo87: I understand you mean not adding this functionality, and just doing it manually?

yes, that's what I meant. as mentioned in https://github.com/ember-cli/ember-cli/pull/5931 the function should ultimately go into an independent package since it's not directly related to Ember CLI at all.

pablobm:
Right, I wasn't aware of #5931. The way I see it, there are two different conversations here:

1. Should this functionality exist?
2. If yes, when should it be added?
  1. Now, in this PR (either with this syntax or with an alternative one)
  2. After `insertIntoFile` is moved out to its own package

Opinions?

Turbo87:
> Should this functionality exist?

IMHO it's much simpler to just append/prepend using the usual Node.js API calls and there's no need to make `insertIntoFile` more complicated than it already is.

pablobm:
OK then, closing. Thanks for your comments 👍

PR #5963, opened 2016-06-09
Fix reading npm package names with a version specifier

### LINKED ISSUE #5880: "ember install package@version" fails after npm

Turbo87:
This PR finishes https://github.com/ember-cli/ember-cli/pull/5883 and moves it to the `beta` branch.

Resolves #5880

/cc @martndemus @nathanhammond @mike-north @stefanpenner @rwjblue

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -521,17 +521,12 @@ Project.prototype.reloadAddons = function() {
 Project.prototype.findAddonByName = function(name) {
   this.initializeAddons();
 
-  var unqualifiedName = name;
-  if (name.indexOf('/') !== -1) {
-    unqualifiedName = name.slice(name.lastIndexOf('/') + 1);
-  }
-
   function matchAddon(name, addon) {
     return name === addon.name || (addon.pkg && name === addon.pkg.name);
   }
 
   return find(this.addons, function(addon) {
-    return matchAddon(name, addon) || matchAddon(unqualifiedName, addon);
+    return matchAddon(name, addon);
   });
 };
 
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -55,18 +55,28 @@ module.exports = Task.extend({
     return packageNames.reduce(function(promise, packageName) {
       return promise.then(function() {
         blueprintName = addonInstall.findDefaultBlueprintName(packageName);
-        taskOptions = merge({
-          args: [blueprintName],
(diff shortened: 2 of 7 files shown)


martndemus:
:+1:

stefanpenner:
:shipit:

nathanhammond:
🚢, waiting on tests for homu.

nathanhammond:
(once landed I'll release our last beta.)

Turbo87:
It would be awesome if you could give this a quick test drive before merging. It should work but I'm not sure I covered all the corner cases.

Turbo87:
Travis is green now 🎉

nathanhammond:
@homu r+

Turbo87:
Looks like @homu *can* handle non-master branches :)

cbfx:
strange.

```
› ember install sinon@^9.0.2
```

```
🚧  Installing packages... This might take a couple of minutes.
Yarn: Installed sinon@^9.0.2
Install failed. Could not find addon with name: sinon@^9.0.2
```

PR #5976, opened 2016-06-11
Anonymous AMD Support

ef4:
This is a second attempt at #5512.

It adds an option named `amdModule` to app.import. It causes any anonymous AMD module in the imported file to be registered under the given name.

    app.import('bower_components/some-dep/some-dep.js', { amdModule: 'some-dep' });

This should work correctly with a large number of libraries that:

 - check for `typeof define === 'function' && define.amd`
 - call `define` with just a factory function, or a dependency list and factory function.

PR #5512 was reverted because the `stew.map` was running over every file in bower and vendor, which was prohibitvely expensive, and which caused binary files to get rewritten as ascii. To address those concerns, I switched to using Funnel to pull out only the explicitly named amdModules into a separate tree first.

In addition, apps that don't use this feature should incur zero cost, because we skip the extra trees entirely.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -31,6 +31,7 @@ var concat = require('broccoli-concat');
 var ConfigReplace = require('broccoli-config-replace');
 var ConfigLoader  = require('broccoli-config-loader');
 var mergeTrees    = require('./merge-trees');
+var shimAmd    = require('./amd-shim');
 var WatchedDir    = require('broccoli-source').WatchedDir;
 var UnwatchedDir  = require('broccoli-source').UnwatchedDir;
 
@@ -122,6 +123,7 @@ function EmberApp(defaults, options) {
 
   this._styleOutputFiles       = {};
   this._scriptOutputFiles      = {};
+  this.amdModuleNames          = null;
 
   this.legacyFilesToAppend     = [];
   this.vendorStaticStyles      = [];
@@ -980,9 +982,22 @@ EmberApp.prototype._processedExternalTree = function() {
     trees.unshift(bower);
   }
 
-  return this._cachedExternalTree = mergeTrees(trees, {
+  var externalTree = mergeTrees(trees, {
     annotation: 'TreeMerger (ExternalTree)'
   });
+
+  if (this.amdModuleNames) {
+    var anonymousAmd = new Funnel(externalTree, {
+      files: Object.keys(this.amdModuleNames),
+      annotation: 'Funnel (named AMD)'
+    });
+    externalTree = mergeTrees([externalTree, shimAmd(anonymousAmd, this.amdModuleNames)], {
+      annotation: 'TreeMerger (named AMD)',
+      overwrite: true
+    });
+  }
+
+  return this._cachedExternalTree = externalTree;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


stefanpenner:
awesome, will review.

Turbo87:
might be a good idea to write a few lines about this new feature in the `import()` API doc comment

stefanpenner:
@Turbo87 we actually decided at the F2F this will require an RFC and the ember core team will participate for final +1 during a final review process (similar to how the rust subteams and core-team work)

Turbo87:
@stefanpenner should we close the PR for now until the RFC is merged?

stefanpenner:
@Turbo87 nah its ok, although we may want to add a more appropriate RFC.

les2:
+1. i can clean out my vendor dir with hacked amd modules!

les2:
could this be provided as simple plugin one adds to app?

```
ember install amd-module-shims
```

Every transformation needed can't be shipped with ember-cli itself.

and even the included transformations could be implemented as plugins, simplifying the core.

plugins could be field tested without requiring full ember RFC. the best plugins become adopted and included with the core.

what hooks would we need to make this PR implementable as plugin? IMHO, that plugin api, assuming it doesn't exist, should be the RFC that is opened.

nathanhammond:
@les2 That's something we're exploring here: https://github.com/ember-cli/rfcs/issues/39

A lot of the head-to-tail handling of Broccoli trees is hard-coded in Ember CLI right now which is something that we want to investigate how to improve. However, that's not coming any time soon; so this is a pragmatic stepwise solution.

stefanpenner:
> Every transformation needed can't be shipped with ember-cli itself.

Agreed, this one though is so very common. (We can always extract)

> and even the included transformations could be implemented as plugins, simplifying the core.
> 
> plugins could be field tested without requiring full ember RFC. the best plugins become adopted and included with the core.
> 
> what hooks would we need to make this PR implementable as plugin? IMHO, that plugin api, assuming it doesn't exist, should be the RFC that is opene

Each type requires something different, which complicates maters. The above one is quick simple. Actually @chadhietala packager/linker work (we hope to resume this summer) is most likely what  provides the most reasonable API for this.

ef4:
Based on discussion here I updated https://github.com/ember-cli/rfcs/pull/55 to an API that looks like this:

```js
app.import('/path/to/module.js', {
  using: [
    { transformation: 'amd', as: 'some-dep' }
  ]
});
```

in order to support arbitrary future transformations. Details in the RFC, review appreciated.

lib/broccoli/ember-app.js
```diff
@@ -31,6 +31,7 @@ var concat = require('broccoli-concat');
 var ConfigReplace = require('broccoli-config-replace');
 var ConfigLoader  = require('broccoli-config-loader');
 var mergeTrees    = require('./merge-trees');
+var shimAmd    = require('./amd-shim');
```
stefanpenner: my OCD!

tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -351,6 +351,30 @@ describe('Acceptance: brocfile-smoke-test', function() {
...
+        });
+
+        (function() {
+          function define(name, deps, factory) {
+            expect(name).to.equal('hello-world');
+            expect(deps).to.deep.equal([]);
+            expect(factory()()).to.equal('Hello World');
+          }
+          /* eslint-disable no-eval */
+          eval(outputJS);
```
stefanpenner: If this code fails to invoke `define` the test also passes.

stefanpenner:
@ef4 left one test related comment + this needs rebase

stefanpenner:
I believe I would like to aim for this week to be the last week of the final review period for the corresponding RFC. That means, friday after the core team meeting, we will either merge this. or have action items to address issues.

I suspect we wont have any objections, it is largely just part of our process we want to improve. This aims to help ensure we have a cohesive  end2end experience.

stefanpenner:
no objects, once rebased this is ready to be merged to canary.

stefanpenner:
rebased, tests need running.

ef4:
This needs to be updated to match the latest in the RFC. It's a tiny change to the API naming, but I'm off to a wedding at this moment and can't do it myself, so please don't merge unless somebody makes that change first.

(It should be "using" instead of "amdModule" as described in the RFC.)

stefanpenner:
@ef4 ah, thanks. Obviously i jumped the gun.

ef4:
This is ready to go now.

stefanpenner:
@ef4 not quite, what are your thoughts on https://github.com/ember-cli/ember-cli/pull/5976/files#r74230651 ?

ef4:
@stefanpenner fixed that test.

hhff:
amazing work @ef4 !!!

PR #5979, opened 2016-06-11
Use `ember try:one` in travis

san650:
Removes deprecation warning

```
DEPRECATION: tryCmd command is deprecated in favor of `ember try:one`.

Example: `ember try default-scenario serve --port 4201`, becomes:
`ember try:one default-scenario --- ember serve --port 4201`
```


nathanhammond:
@homu r+

Turbo87:
@nathanhammond should we pull this down to `beta`?

nathanhammond:
Why not!

PR #5984, opened 2016-06-14
Replace "stub" helper with "testdouble"

Turbo87:


Turbo87:
looks like I need to fix a few Windows-related things...

stefanpenner:
cc @searls

searls:
How cool is this! I'll take this for a spin tomorrow and provide feedback on style/conventions

Turbo87:
I'm pretty sure @searls will kill me for this once he had a closer look... 😂 

This is meant as an intermediate step. I'm pretty sure that most of the time this is not using testdouble in the way it's supposed to be used, but it feels better than what it did before so I'm happy with it for now.

searls:
@Turbo87 your caveats will not save you, Tobias!

Turbo87:
@searls did you have a look? any feedback?

searls:
Hi Tobias, this got slightly delayed on my end. I probably won't be able to look at it until I'm back in the states Wednesday

stefanpenner:
@Turbo87 / @searls I think its safe to merge this now, although I would love review (but most likely in our code-base rather then this PR) when @searls has time.

stefanpenner:
@homu r+

Turbo87:
@searls can I interpret the silence as me not having done anything totally crazy and you're happy with it? 😉

searls:
Haha, you can interpret it as me not finding the time to actually do it. It's on my list, but the list is long and full of terrors. :(

searls:
I finally am getting the chance to scan this. A few things stand out so far:

## Using `ignoreExtraArgs: true` frequently

This option is really only appropriate for stubbings which receive a dynamic number of arguments and you aren't in control of what they'll be, which should be rare in a unit test because you ought to have complete control (or at least knowledge) of who is calling the test double. 

I suspect the reason you did this was to replace some test double that ignored-args by default, so I won't belabor the point but to say that I suspect most of these tests would keep passing if you deleted everything being passed to them. That means the tests are probably [insufficient](https://github.com/testdouble/contributing-tests/wiki/Necessary-&-Sufficient).

## Frequent use of `td.captor`

In general, repeated use of `td.captor` is usually a smell that a more well-suited matcher could provide clearer, terser assertions, as opposed to the multi-staged assertions you need with a `captor` -> `captor.capture()` -> assert series.

For example:

``` js
captor = td.captor()
td.verify(emberApp.addonPostprocessTree('template', captor.capture()));
expect(captor.value.description).to.equal('template', 'should be called with consolidated tree');
```

Could be shortened using the built-in contains matcher:

```js
td.verify(emberApp.addonPostprocessTree('template', td.matchers.contains({
  description: 'should be called with consolidated tree'
})));
```

searls:
Hey look it only took 3 months for me to do this.

PR #5987, opened 2016-06-15
[BUGFIX release] allow more then 10 exit handlers

stefanpenner:
We can decide on a more appropriate number (if others feel that is required). 

Resolves https://github.com/babel/ember-cli-babel/issues/76

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -13,6 +13,15 @@ var debug         = require('debug')('ember-cli:cli/index');
 var merge         = require('lodash/merge');
 var path          = require('path');
 
+// ember-cli and user apps have many dependencies, many of which require
+// process.addListener('exit', ....) for cleanup, by default this limit for
+// such listeners is 10, recently users have been increasing this and not to
+// their fault, rather they are including large and more diverse sets of
+// node_modules.
+//
+// https://github.com/babel/ember-cli-babel/issues/76
+process.setMaxListeners(1000);
+
 var version      = packageConfig.version;
 var name         = packageConfig.name;
 var trackingCode = packageConfig.trackingCode;


Turbo87:
LGTM 👍

Turbo87:
do we want this on `release` though or should we put this through a `beta` cycle?

side note: as discussed in https://github.com/ember-cli/ember-cli/pull/5955 we should think about targeting these PRs directly at the `release` or `beta` branches to reduce the amount of duplicated commits and have TravisCI/AppVeyor test for these branches directly.

stefanpenner:
> do we want this on release though or should we put this through a beta cycle?

i suspect release.

rwjblue:
I don't feel strongly, but this doesn't seem like a critical bugfix.  Seems safe for both beta and release, but it also isn't a giant deal and I think our default answer should be "beta" for all but critical fixes...

nathanhammond:
Doesn't seem critical to me–everything still works, you just get these messages. I'd aim for beta on principle.

Turbo87:
Cherry picked this to `beta`. I agree with @rwjblue that this isn't critical enough for a stable release.

lmcardle:
We upgraded to ember-cli 2.4.3, and all associated dependencies with the exception of ember data, and I did not hit this warning. Later we upgraded ember data to 2.3 (can't move past that version at this point) and this warning suddenly appeared. Is there any way we can change the maxListeners?

stefanpenner:
@lmcardle this is slated for 2.7 release, the diff shows how one could manually add it to there project: https://github.com/ember-cli/ember-cli/pull/5987/files#diff-3386b9e4d97ad97b803f4678beb66c5aR23 just put that line in your ember-cli-build.js file

lmcardle:
@stefanpenner you sir are awesome. That appears to work great!

stefanpenner:
@lmcardle nice, I'm glad to hear it worked for you. I don't think nodes default value here needs to change eventually, or maybe the approach at-together, but this will buy us some time.

nathanhammond:
> this will buy us some time.

Nomination for understatement of the year.

hugoruscitti:
For those who can't update to ember 2.7 right now, it's possible to fix this warning message by adding this line to the end of the "ember-cli-build.js" file:

```
process.setMaxListeners(1000);
```

stefanpenner:
@hugoruscitti thanks or sharing the tip, remember though the build tooling can be upgraded independently. Although it may require a more granular update.

PR #5992, opened 2016-06-17
Adds the e(mber) favicon per default.

fpauser:
The favicon.ico is multi-size and includes a 16x16 and a 32x32 pixels version.


stefanpenner:
LGTM

rwjblue:
I am unsure about adding more things that have to be removed from a newly generated app...

stefanpenner:
we should likely split  lib/models/file-info.js into a seperate PR

stefanpenner:
@rwjblue I think a favicon is fine, users merely need to update it with their own favicon which seems like a common thing to do?

kellyselden:
Maybe this is something for the ember-cli/ember-welcome-page addon, if at all possible.

nathanhammond:
Considering how many Sun, Drupal, H5BP, and other default favicons I've seen I'd rather not add Ember to that list. Also, what if your Ember application is not served at `/` and this now clobbers the reference to `/favicon.ico`? People know to add it if they need it, I don't want us to get into the business of implementing H5BP for our users.

I'm opposed to this.

stefanpenner:
> Considering how many Sun, Drupal, H5BP, and other default favicons I've seen I'd rather not add Ember to that list. Also, what if your Ember application is not served at / and this now clobbers the reference to /favicon.ico? People know to add it if they need it, I don't want us to get into the business of implementing H5BP for our users.

Seems pretty compelling @fpauser thoughts?

fpauser:
I came across this because after playing around with angular-cli - that also serves on port 4200 by default - all my ember-apps started showing the cached angular favicon, which somehow feels just not right...

Regarding the expressed "default-icon-annoyence": I see what you mean - but on the other hand: Whats so bad about a default that promotes the fine framework behind?

Regarding path issues: As the favicon is referenced relatively this should not be a problem, or do I oversee something?

Maybe a better solution could be to change the express-server to serve the ember-favicon by default as long as no custom favicon is present in `public`.

fpauser:
Ok, convinced - closing it for compelling reasons.

fpauser:
Moved the usable stuff to #5994.

Turbo87:
> Maybe a better solution could be to change the express-server to serve the ember-favicon by default as long as no custom favicon is present in public. This would at least ensure a correct favicon in local-mode and prevent wrong/missleading favicons from the browser-cache.

I like that idea...  :)

fpauser:
@Turbo87 I'll try to prepare something.

nathanhammond:
@fpauser I love the idea of the Express server having it locally. I look forward to that PR. (Seems like the perfect compromise, and once the user adds in a favicon it will do the right thing.)

mike-north:
I'm going to addon-ify this if the consensus is to not add it to ember-cli. Considering that there are multiple CLI tools that use localhost:4200 by default, you can end up with some odd behavior whose behavior is difficult to clear from the cache on some browsers. Particularly annoying if you're not a pure "ember developer" and use a bunch of different things.

screen shot 2016-10-11 at 2 42 55 pm

nathanhammond:
@mike-north We're not going to add it to ember-cli by default. As an addon it could be set up via middleware or inclusion. If we make it nigh impossible to get wrong I'm game for inclusion in the default blueprint. However, there are considerations...

## Middleware

- Express only allows exactly on middleware to send a body.
- The last one in our chain, `broccoli-middleware` (wrapped by `serve-files`), is the only one which truly knows if we don't have a `/favicon.ico`. This means we need it to either handle this task (nope), accept middleware of its own (complicated), or ...
- For bonus points some people actually use the Ember CLI server in production against all advice, so this is likely to still show up on people's prod websites.

## Inclusion

- Only be included in the development environment, ensure it gets clobbered by the root application.
- Different priorities of assets for `/favicon.ico` and `rel=icon` need to be investigated and ensure that no matter what happens once a user adds something that can trigger a favicon that it wins.
- But for bonus points people accidentally publish dev builds all the time so it's no guarantee that we avoid the problem of this showing up on people's prod websites by only inlining it into dev builds.

nathanhammond:
In other news, I consider defaults that could accidentally leak the wrong thing into a user's application user-hostile which goes against the type of tool we're trying to build here. Unless [angular-cli](https://github.com/angular/angular-cli) has solved all of these problems (which we could then just port back) I disagree with their decision.

PR #5994, opened 2016-06-17
No 'diff' option when binaries are involved (confirm-dialog).

fpauser:


stefanpenner:
mind adding a unit test?

lib/models/file-info.js
```diff
@@ -33,18 +33,30 @@ function diffHighlight(line) {
 }
 
 FileInfo.prototype.confirmOverwrite = function(path) {
+  var info = this;
```
Turbo87: what's the reason for that extra variable?

fpauser: You'r right - no reason/need to save the context there, will update the PR later.

nathanhammond:
Can you do some git-fu to make sure that just your change appears in this PR? Not sure what else changed that you have non-passing tests, so hoping to reduce the surface area of this test.

fpauser:
Did some gitfu to get it nice and dandy ;)
Locally I have one failing test that also breaks on master - did not investigate this further.

```
923 passing (21s)
  5 pending
  1 failing

1) Acceptance: ember generate passes custom cli arguments to blueprint options:
     AssertionError: expected "app/foo.js" to exist
```

Turbo87:
@fpauser could you rebase this branch and see if the tests are still failing? Otherwise I'd like to get this merged soon.

nathanhammond:
@homu r+

nathanhammond:
@fpauser Thank you for making the user experience better!

PR #5995, opened 2016-06-17
Allow custom config paths

bryanaka:
If you are running an Ember App **not** in the root directory of your app, you need to be able to specify where your config is located in order for Ember CLI to run correctly.

This adds the option to allow configuring the configPath via package.json - specifically via package.directories.configPath.

This is a follow up from #4042

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -172,7 +172,7 @@ EmberApp.prototype._initProject = function(options) {
   this.project = options.project || Project.closestSync(process.cwd());
 
   if (options.configPath) {
-    this.project.configPath = function() { return app._resolveLocal(options.configPath); };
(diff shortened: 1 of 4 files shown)


Turbo87:
> If you are running an Ember App not in the root directory of your app

I'm not sure I understand what you're doing there. Can you give an example of how that would look like?

nathanhammond:
So, this is actually already possible (which you may be familiar with having just gone through this code). Inside of `package.json`:

```
{
  "name": "app-name",
  "ember-addon": {
    "configPath": "path/to/config"
  }
}
```

It's fair game to do that inside of an Ember application as well, not just inside of addons. It's how we handle this for our application. I'm closing this PR on the assumption that this approach will meet your needs. If it doesn't _please feel encouraged to reopen_!

PR #5996, opened 2016-06-18
broccoli: Add "__esModule" property to "./config/environment" export

Turbo87:
This adds some Babel 6 compatibility.

/cc @rwjblue @stefanpenner


nathanhammond:
LGTM. (Multi meta modules made me super-familiar with this code.)

@homu r+

PR #6003, opened 2016-06-20
Pin "testem" to v1.8.1

Turbo87:
let's see if this fixes the builds...

/cc @rwjblue

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "silent-error": "^1.0.0",
     "symlink-or-copy": "^1.0.1",
     "temp": "0.8.3",
-    "testem": "^1.6.0",
+    "testem": "1.8.1",
     "through": "^2.3.6",
     "tiny-lr": "0.2.1",
     "tree-sync": "^1.1.0",


Turbo87:
Looks like this is indeed fixing things... Maybe we should keep testem pinned in the future.

rwjblue:
Yeah, but that is also unfortunate. We have been able to take advantage of the loose constraint much more than it has bitten us.

Perhaps we could test with ember-cli's CI before releasing `testem`?  I have mixed feelings about that...

johanneswuerbach:
Strange, after noticing the test failure I marked the package as new version deprecated, but it seems to still install it.

I'll go back to releasing a rc next time again.

johanneswuerbach:
Published testem v1.9.1, which is v1.8.1 re-published until all issue have been resolved, sorry about that 😢 .

rwjblue:
@johanneswuerbach - Thank you for jumping on it!

Turbo87:
Guess we can close this now. Thanks @johanneswuerbach :)

PR #6005, opened 2016-06-20
[BUGFIX] Fixes broccoli errors when `tests` dir is not present

### LINKED ISSUE #5946: Broccoli errors when tests/ are deleted
According to this person on Reddit: https://www.reddit.com/r/emberjs/comments/4m49rv/disable_tests_with_ember_cli/

I would advocate that if someone decides to opt-out of the test suite for whatever reason and deletes the directory the build should not error.

I have not confirmed this as I'm on a plane and don't want it to crash due to npm install.

twokul:
resolves #5946

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -198,7 +198,9 @@ EmberApp.prototype._initOptions = function(options) {
   }
 
   var appTree = new WatchedDir(this._resolveLocal('app'));
-  var testsTree = new WatchedDir(this._resolveLocal('tests'));
+
+  var testsPath = this._resolveLocal('tests');
+  var testsTree = existsSync(testsPath) ? new WatchedDir(testsPath) : null;
 
   // these are contained within app/ no need to watch again
   // (we should probably have the builder or the watcher dedup though)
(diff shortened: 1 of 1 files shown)


twokul:
cc @rwjblue

rwjblue:
Seems good.  You also confirmed locally?

twokul:
@rwjblue I did

stefanpenner:
So to be clear this just makes it possible to start/stop ember server/test without the existence of `tests` but will still fail if `tests` is removed mid-build. (Which i think we are ok with)

twokul:
@stefanpenner correct

rwjblue:
I'm 👍 here, but CI is blocked until https://github.com/ember-cli/ember-cli/pull/6003 or testem fixes the underlying issue...

rwjblue:
@homu r+

nathanhammond:
Giving up on @homu here.

Turbo87:
since this is a bugfix: do we want to backport this onto `release` or `beta`?

nathanhammond:
It has always been this way and it took until June of this year for it to show up. I'd call it a rare enough scenario that it goes through the normal lifecycle. If somebody shows up here clamoring for it sooner we can revisit.

rwjblue:
Agreed. I'm in favor of letting it ride on master through a full cycle...

PR #6008, opened 2016-06-21
log when an add ons command takes more then 20ms

stefanpenner:
- [ ] provide actual gist/readme in warning to help users/authors fix these issues.

screen shot 2016-06-20 at 11 50 14 pm

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -397,8 +397,16 @@ Project.prototype.initializeAddons = function() {
  */
 Project.prototype.addonCommands = function() {
   var commands = {};
+  var ui = this.ui;
   this.addons.forEach(function(addon) {
+    var start = Date.now();
     var includedCommands = (addon.includedCommands && addon.includedCommands()) || {};
+    var total = Date.now() - start;
+
+    if (total > 20 /* ms */) {
+      ui.writeWarnLine('Addon[' + addon.name + ']\'s commands took: ' + total + 'ms to load, see: http://some-link.com/help');
+    }
+
     var addonCommands = {};
 
     for (var key in includedCommands) {


nathanhammond:
This is going to be tough to choose a number for: slow computers, Windows, etc. Based on the fallout from the other "addon is doing the wrong thing!" `_super` stuff alerting the end-user of slow addons is going to dramatically increase support load.

lib/models/project.js
```diff
@@ -397,8 +397,16 @@ Project.prototype.initializeAddons = function() {
...
 Project.prototype.addonCommands = function() {
   var commands = {};
+  var ui = this.ui;
   this.addons.forEach(function(addon) {
+    var start = Date.now();
     var includedCommands = (addon.includedCommands && addon.includedCommands()) || {};
+    var total = Date.now() - start;
+
+    if (total > 20 /* ms */) {
+      ui.writeWarnLine('Addon[' + addon.name + ']\'s commands took: ' + total + 'ms to load, see: http://some-link.com/help');
```
rwjblue: Note for when this is ready to land: `some-link.com` needs to be changed

stefanpenner:
> This is going to be tough to choose a number for: slow computers, Windows, etc. Based on the fallout from the other "addon is doing the wrong thing!" _super stuff alerting the end-user of slow addons is going to dramatically increase support load.

Ya, i may add an option to disable the warning in `.emberrc` and document it from the link I have yet to create

lib/models/project.js
stefanpenner: already in the checklist

rwjblue: HAHA, sorry about that I didn't see the checklist...

rwjblue:
I'm not in favor of this as an "on by default" thing until we can massage over the relative differences in platform performance.  I'd be fine with this as enabled by default in beta or canary, but super noisy consoles in a release version is a no-go for me.

stefanpenner:
We can decide on a threshold that doesn't trip to often and a way to disable, having ember-cli be slow for people because addons are doing eager is also not a good thing.

rwjblue:
@stefanpenner - I agree that there is some threshold that we can use (perhaps we can grab the cost of requiring one of our own files, and use some sort of multiple).

rwjblue:
I think we should also allow addons to be able to completely avoid the lazy require shenanigans by returning the string path to their commands (instead of the actual object).  Something like:

```js
includedCommands: function() {
  return {
    'ember-try:each': path.join(__dirname, 'lib/commands/each'),
    'ember-try:one': path.join(__dirname, 'lib/commands/one')
  };
}
```

stefanpenner:
@rwjblue im not sure how that helps commands like `help`

rwjblue:
@stefanpenner - Confirm, but warning like this PR is doing doesn't "solve" `help` either. `help` is not the most commonly used command (things like `ember build`, `ember serve`, `ember test`, `ember version` would all be helped tremendously if addons could use the string method I mentioned above) and will always be slower than others (due to the fact that it traverses all commands and builds up their help output).

There is no silver bullet that solves all issues. I am suggesting that we:

* land a warning with the time be based on some known `require` time (not a hard coded number of milliseconds)
  * for invoking `includedCommands` (this PR)
  * for `require`ing the actual addons themselves
* add the ability to specify strings as the values in the object returned from `includedCommands`

stefanpenner:
> land a warning with the time be based on some known require time (not a hard coded number of milliseconds)

Im not sure why, 20ms to load a command is totally reasonable budget. 

> There is no silver bullet that solves all issues. I am suggesting that we:

Enforcing a budget actually would help this

> Confirm, but warning like this PR is doing doesn't "solve" help either

It actually would help dramatically

rwjblue:
> Im not sure why, 20ms to load a single command is totally reasonable budget on any system.

I doubt this very much. Weren't you just mentioning to me the other day about running ember-cli on a mobile device? Basing the number on a known require time seems both simple and obviously better than a random "magic number".

> Enforcing a budget actually would help this

Agreed.  I never suggested it wouldn't help. I am suggesting that it is only a portion of a larger solution.

> The goal of this is to put set a budget, to insure we don't continuously bitrot initial startup.

Yep, understood. Definitely an admirable goal.  We can do better. 😸

Turbo87:
> `help` is not the most commonly used command

@rwjblue making `help` fast is needed to finish https://github.com/ember-cli/ember-cli/pull/4709 (unless we use a cache, but that approach is a bit brittle)

nathanhammond:
We discussed this a bit in our core team meeting on 6/23 as things we could do that might be less jarring than a list of "slow addon" alerts. Things we could do:

- Opt-in only, via command line flag.
- Only output on `canary` or `beta` builds.
- Track performance ourselves using analytics and reach out behind the scenes to slow addons (usually nothing the consumer will be able to do to help, so this moves the responsibility to us while getting additional information).
- Write a command that scores the perf of addons; push that to Ember Observer.

Thoughts?

stefanpenner:
> Opt-in only, via command line flag.
> Only output on canary or beta builds.

seems good, would like to add:

* ember-observer could include this in there rating of an add-on (cc @kategengler i suspect you have some thoughts/ideas)

stefanpenner:
We should most likely (for now) transition this to a heimdall logger. Improve the ecosystem, then set thresholds for warning.

cc @hjdivad

Turbo87:

PR #6012, opened 2016-06-21
switch to a rollup subset of lodash and shave off 20 - 30%+ boot time

stefanpenner:


bin/ember
```diff
@@ -4,6 +4,7 @@
 // Provide a title to the process in `ps`
 process.title = 'ember';
 
+// require('time-require');
```
rwjblue: I'd actually like to leave this in, behind an env var flag.  What do you think?

rwjblue:
Does this just turn into another thing we have to maintain over time? Are there any actionable things that lodash itself can do here?

stefanpenner:
> Does this just turn into another thing we have to maintain over time? Are there any actionable things that lodash itself can do here?

It basically uses lodash public API lodash-es + rollup to automate the work: see: https://github.com/stefanpenner/ember-cli-lodash-subset

@jdalton may have some further ideas..

jdalton:
Have y'all tried [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin)?

stefanpenner:
> Have y'all tried lodash-webpack-plugin?

is this for node apps or browser apps?

rwjblue:
A small checklist of things that would make me more comfortable with this:

- [ ] change floating dependency on `lodash-es` to specific pinned version dependency.  This will allow Greenkeeper to properly notify us when upstream bugfixes need a release on our part.
- [ ] Move to ember-cli org?
- [ ] give others commit & release so that we can tag team the maintenance 
- [ ] enable greenkeeper.io for the repo so that we can keep up to date with `lodash-es` releases

stefanpenner:
> Move to ember-cli org?

GH wont let me for some reason, will try again later

stefanpenner:
> enable greenkeeper.io for the repo so that we can keep up to date with lodash-es releases

once we move to this org

jdalton:
@stefanpenner 
> Is this for node apps or browser apps? ^ is for the CLI itself not for the apps its builds.

It's the same idea as using rollup but would use webpack and generate a smaller build.

stefanpenner:
> give others commit & release so that we can tag team the maintenance

others have release, will give them commit once its moved (easier)

jdalton:
@stefanpenner 
> Is this for node apps or browser apps? ^ is for the CLI itself not for the apps its builds.

It's the same idea as using rollup but would use webpack and generate a smaller build.

I can create a branch that shows what I mean.

jdalton:
Something like https://github.com/jdalton/ember-cli-lodash-subset/commit/2d2b921e41e90c521fe17fb98f31814f984bfedc.

stefanpenner:
> It's the same idea as using rollup but would use webpack and generate a smaller build.

awesome thank you. It is indeed smaller, although somewhat slower. Will dig in a-bit more..

bin/ember
Turbo87: that would yet again increase the `npm install` time just for debugging purposes. I'm not really a huge of that.

lib/commands/test.js
```diff
@@ -35,7 +35,7 @@ module.exports = Command.extend({
   init: function() {
     this._super.apply(this, arguments);
 
-    this.assign    = require('lodash/assign');
+    this.assign    = assign;
```
Turbo87: we probably won't need `this.assign` anymore if we `require` above

lib/models/command.js
```diff
@@ -9,15 +9,16 @@ var getCallerFile       = require('get-caller-file');
...
-var union               = require('lodash/union');
-var uniq                = require('lodash/uniq');
-var uniqBy              = require('lodash/uniqBy');
-var map                 = require('lodash/map');
-var reject              = require('lodash/reject');
-var filter              = require('lodash/filter');
-var defaults            = require('lodash/defaults');
-var assign              = require('lodash/assign');
-var keys                = require('lodash/keys');
+var lodash = require('ember-cli-lodash-subset');
```
Turbo87: should we assign to `_` instead of `lodash` and use the methods as e.g. `_.map()` instead of `map()` plus all the destructuring?

stefanpenner: unsure, i can go either way. Any strong feelings?

lib/models/file-info.js
```diff
@@ -17,7 +16,7 @@ function processTemplate(content, context) {
     interpolate: /<%=([\s\S]+?)%>/g,
     escape:      /<%-([\s\S]+?)%>/g
   };
-  return template(content, options)(context);
+  return require('lodash.template')(content, options)(context);
```
stefanpenner: ya, i figured it didn't need to be part of the fast case, as it is only required in already slow scenarios  (blueprinting)

lib/models/command.js
Turbo87: seems to be the more conventional way of using lodash

stefanpenner:
@Turbo87 i have updated re: your comments, let me know if I addressed them.

Turbo87:
@stefanpenner that lodash destructuring was happening in a few other places as well. might be worth it to replace those too.

stefanpenner:
@rwjblue transfered, but required hackery: https://github.com/ember-cli/ember-cli-lodash-subset

* https://twitter.com/stefanpenner/status/745659020568977408
* https://twitter.com/stefanpenner/status/745659497503297536

stefanpenner:
@rwjblue your points are addressed, going to add travis now and we should be good

@Turbo87 i think i also addressed yours, but let me know.

Turbo87:
seems good, r+ from my side

stefanpenner:
@homu r+

stefanpenner:
@homu r+

PR #6030, opened 2016-06-30
[ENHANCEMENT] Asset Sizes I moved the creation of asset sizes to an object.

kiwiupover:
Moved the creation of asset sizes to an object.
Then print command iterates over that object.

I talked a little to @rwjblue and @kellyselden about the refactor. 

If we agree this is worth while I will write a test for the `makeAssetSizesObject` function. 

My use case I want to persist the asset-size data to a graphing server to track an apps size over time.

### DIFF
--- a/lib/models/asset-size-printer.js
+++ b/lib/models/asset-size-printer.js
@@ -9,56 +9,54 @@ function AssetPrinterSize(options) {
   assign(this, options);
 }
 
-AssetPrinterSize.prototype.print = function() {
+AssetPrinterSize.prototype.print = function () {
   var ui = this.ui;
 
-  return Promise.resolve()
-    .then(function () {
-      var filesize = require('filesize');
-      var fs = require('fs');
-      var zlib = require('zlib');
-      var gzip = Promise.denodeify(zlib.gzip);
-      var files = this.makeFileGlob();
-      var testFileRegex = /(test-(loader|support))|(testem)/i;
-      var promises = [];
+  return this.makeAssetSizesObject().then(function (files) {
+    this.validateAssetPath(files);
 
-      this.validateAssetPath(files);
-      ui.writeLine(chalk.green('File sizes:'));
+    ui.writeLine(chalk.green('File sizes:'));
+    return files.forEach(function (file) {
+      var sizeOutput = file.size;
+      if (file.showGzipped) {
+        sizeOutput += ' (' + file.gzipSize + ' gzipped)';
+      }
 
-      files
-        // Skip test files
-        .filter(function (file) {
-          var filename = path.basename(file);
-          return !testFileRegex.test(filename);
-        })
-        // Print human-readable file sizes (including gzipped)
-        .forEach(function (file) {
-          var filename = path.basename(file);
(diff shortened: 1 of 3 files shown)


Turbo87:
@kiwiupover looks like there is a failing test for this PR, can you check?

lib/models/asset-size-printer.js
```diff
@@ -9,56 +9,60 @@ function AssetPrinterSize(options) {
...
+  this.validateAssetPath(files);
+
+  files
+    // Skip test files
+    .filter(function (file) {
+      var filename = path.basename(file);
+      return !testFileRegex.test(filename);
+    })
+    // Print human-readable file sizes (including gzipped)
+    .forEach(function (file) {
```
stefanpenner: we should make this a `map` , that way we don't need to also push onto the promise array manually

lib/models/asset-size-printer.js
```diff
@@ -9,56 +9,60 @@ function AssetPrinterSize(options) {
...
+          gzipSize: filesize(buffer.length),
+          showGzipped: contentsBuffer.length > 0
+        };
       });
+
+      promises.push(gzipPromise);
     });
+
+  return Promise.all(promises).then(function (assetObject) {
+    return assetObject;
```
stefanpenner: assetObject would actually be the fullfillment of all the promises, is this variable named incorrectly or.. ?

kiwiupover: I will update that name.

lib/models/asset-size-printer.js
kiwiupover: I will make that change.

So @stefanpenner are saying this refactor is worth landing?

lib/models/asset-size-printer.js
```diff
@@ -9,56 +9,60 @@ function AssetPrinterSize(options) {
...
-        var sizeOutput = file.size;
-        if (file.showGzipped) {
-          sizeOutput += ' (' + file.gzipSize + ' gzipped)';
-        }
-
-        ui.writeLine(chalk.blue(' - ' + file.name + ': ') + chalk.white(sizeOutput));
+AssetPrinterSize.prototype.print = function () {
+  return this.makeAssetSizesObject().then(function (files) {
+
+    var ui = this.ui;
```
kellyselden: could the bind be removed if you just move this to outside scope?

lib/models/asset-size-printer.js
kellyselden: @kiwiupover I think a change that makes the code less complex, and also makes it easier for you to accomplish something, is a win-win for both.

nathanhammond:
@kiwiupover I'd call this a good approach. Better-factored code is always better. Can you address the outstanding comments to that we can land it? 😄

kiwiupover:
@nathanhammond @stefanpenner @kellyselden This is ready for review. 

Thanks guys for all your hard work.

nathanhammond:
@homu r+

Maybe?

Turbo87:
whoa! 🎉

PR #6036, opened 2016-07-03
Add information about inverting the test filter

jrjohnson:
QUnit accepts a prefixed ! to invert the filter.  I'm not sure if this is supported by other test frameworks or even what test frameworks ember-cli officially documents and supported.  I'm also not sure where else to put this information which makes it trivial to run non acceptance tests, but is fairly difficult to discover.

eg `ember test --filter='!acceptance'`

Maybe there isn't a good place for this info to go if it isn't globally supported.

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -21,7 +21,7 @@ module.exports = Command.extend({
     { name: 'server',      type: Boolean, default: false,           aliases: ['s'] },
     { name: 'host',        type: String,                            aliases: ['H'] },
     { name: 'test-port',   type: Number,  default: defaultPort,     aliases: ['tp'], description: 'The test port to use when running with --server.' },
-    { name: 'filter',      type: String,                            aliases: ['f'],  description: 'A string to filter tests to run' },
+    { name: 'filter',      type: String,                            aliases: ['f'],  description: 'A string to filter tests to run. Invert filter by prefixing with !.' },
     { name: 'module',      type: String,                            aliases: ['m'],  description: 'The name of a test module to run' },
     { name: 'watcher',     type: String,  default: 'events',        aliases: ['w'] },
     { name: 'launch',      type: String,  default: false,                            description: 'A comma separated list of browsers to launch for tests.' },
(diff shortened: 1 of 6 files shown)


lib/commands/test.js
```diff
@@ -21,7 +21,7 @@ module.exports = Command.extend({
     { name: 'server',      type: Boolean, default: false,           aliases: ['s'] },
     { name: 'host',        type: String,                            aliases: ['H'] },
     { name: 'test-port',   type: Number,  default: defaultPort,     aliases: ['tp'], description: 'The test port to use when running with --server.' },
-    { name: 'filter',      type: String,                            aliases: ['f'],  description: 'A string to filter tests to run' },
+    { name: 'filter',      type: String,                            aliases: ['f'],  description: 'A string to filter tests to run.  Invert filter by prefixing with !.' },
```
kellyselden: We use single space after periods.

jrjohnson: Thanks, fixed.

Turbo87:
> I'm not sure if this is supported by other test frameworks or even what test frameworks ember-cli officially documents and supported.

The default is obviously QUnit, but we have somewhat official support for Mocha too.

nathanhammond:
I'm not sure that the negated filter works for Mocha so I'm unsure how to best handle documenting something like this. It belongs in some guide somewhere, maybe? Also, this command should probably be extracted to `ember-cli-qunit` and `ember-cli-mocha` to solve the problem once and for all.

/cc @locks @acorncom @rwjblue @trentmwillis

Turbo87:
> Also, this command should probably be extracted to ember-cli-qunit and ember-cli-mocha to solve the problem once and for all.

That would potentially introduce quite a bit of code duplication though...

nathanhammond:
@Turbo87 It's a tough trade. Code duplication vs. "correct" separation of concerns. If we're aiming to create an "abstract" core-cli I feel like we should pull it out. Then the trick here is to create an abstraction that ember-cli-qunit and ember-cli-mocha can share. But I agree: the naïve approach here results in tons of duplicated code. Also not something a person can do without a lot of context.

nathanhammond:
@jrjohnson Hey! I'm going to close this without landing it for unfortunate reasons which are in no way your fault. Can I encourage you to open a PR against the Ember Guides to land information about this? We should absolutely do a better job communicating that this is available. https://guides.emberjs.com/v2.6.0/testing/

Unfortunately the way that command is set up we have to make sure its documentation applies just to the things it itself implements, not the things the downstream testing library provides. I really appreciate the time you've spent documenting this and look forward to improving our guides to talk about how this is set up. 😄

trentmwillis:
Official documentation for this was recently added to QUnit: https://github.com/jquery/api.qunitjs.com/pull/128. It'll be published in a patch release soon.

I imagine the Ember guides will want to defer to those guides once they're up.

jrjohnson:
@nathanhammond no worries.  I kind of figured that might be true, but I find it is usually easier to discuss a PR than an issue.  I will look at adding this to the guide.

PR #6039, opened 2016-07-04
Use ES6 spread operator instead of apply

dunnkers:
Use a spread operator instead of using apply. Some users might have had JSCS lint warnings because using apply in this blueprint.

### DIFF
--- a/blueprints/app/files/tests/helpers/module-for-acceptance.js
+++ b/blueprints/app/files/tests/helpers/module-for-acceptance.js
@@ -11,12 +11,12 @@ export default function(name, options = {}) {
       this.application = startApp();
 
       if (options.beforeEach) {
-        return options.beforeEach.apply(this, arguments);
+        return options.beforeEach(...arguments);
       }
     },
 
     afterEach() {
-      let afterEach = options.afterEach && options.afterEach.apply(this, arguments);
+      let afterEach = options.afterEach && options.afterEach(...arguments);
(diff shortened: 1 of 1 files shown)


trentmwillis:
Not sure what the JSCS rule is, but the use of `apply` is valid in this case. Using spread here would actually be a bug as the callback won't be invoked with the proper context.

dunnkers:
I'm referring to this rule: [http://jscs.info/rule/requireSpread](http://jscs.info/rule/requireSpread)

dunnkers:
Closing since looking at the failed tests tells us that functionality is indeed dependent on passing a context with `apply`.

nathanhammond:
@trentmwillis is correct that not applying `this` changes the result of the function call.

You would need to first `bind` the function which results in different browser support than `apply` and making this change a regression (phantom).

This is not a change we're willing to make for the default blueprint so enabling that rule in JSCS will require you to exclude this file from being linted.

rwjblue:
IMO, the JSCS rule is simply wrong here...

trentmwillis:
I feel like something is wrong with JSCS as [the current rule implementation](https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/require-spread.js#L56) looks like it should only throw an error if the first param is the same as the object it's being invoked on (e.g., `obj.func.apply(obj, arguments)`). Either way, seems to need fixing elsewhere.

Issue #5990, opened 2016-06-16
`this.import` may have bugs.


nathanhammond:
Not all `app`s are guaranteed to have `.app`: https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L422-L428

We may instead need to snag `.parent`: https://github.com/machty/ember-maybe-import-regenerator/blob/master/index.js#L38-L50

/cc @machty @rwjblue @zackthehuman

rwjblue:
Thanks for recording this issue @nathanhammond.

zackthehuman:
Indeed; this is a problem I have commonly run in to while trying to get the app's options via `this.app.options`, which sometimes bombs. Resolving by walking the parents seems like the most reliable thing to do.

Turbo87:
fixed by #6043

PR #6043, opened 2016-07-06
[BUGFIX beta] Test nested addon import

xcambar:
refs #5990 

This PR updates `Addon.prototype.import` (https://github.com/ember-cli/ember-cli/pull/6043/files#diff-f1e1923c0440a385f51d2a81c0e53b64R448) as per @machty's implementation (see https://github.com/machty/ember-maybe-import-regenerator/blob/master/index.js#L38-L50).

It also adds a smoke test that builds an app with one addon that itself requires 2 more (hence nested) addons. These addons imports a file each, one with the new `this.import(asset)`, the other with the current `app.import(asset)`.

The PR makes sure that independently from the addon developer's preference (or history), the import binds to the correct app, thanks to @machty's implementation.

@Turbo87 Though I let the smoke test at its initial place, I am more than willing to move it to a more appropriate location, your recommandations are welcome.

As a follow up to #5990, ping @rwjblue and @nathanhammond.


tests/acceptance/nested-addons-smoke-test-slow.js
```diff
@@ -0,0 +1,59 @@
...
+var file = chai.file;
+var dir = chai.dir;
+
+var appName  = 'some-cool-app';
+
+describe('Acceptance: nested-addons-smoke-test', function() {
+  this.timeout(360000);
+
+  before(function() {
+    return createTestTargets(appName);
```
Turbo87: since this part here takes quite a long time it might be beneficial to combine this with other "slow" tests. e.g. if there is something like an `addons-smoke-test-slow` already then you should probably just add it there instead.

Turbo87:
@xcambar I just checked that the current `master` branch is still passing on Travis. It seems that these test failures in the PR are real and caused by the proposed changes. I haven't looked at what exactly changed yet but it seems to influence some other code paths.

xcambar:
Thanks for investigating. FWIW, I'm currently trying to split the changes into smaller chunks and see when/why it breaks master.

rwjblue:
Looks like it is green.  Can you squash down to a single commit, and prefix with `[BUGFIX beta]` to help ensure we pull this down into the beta branch before releasing 2.7?

xcambar:
done

rwjblue:
@homu r+

PR #6045, opened 2016-07-06
[Enhancement] Return raw asset-size as data instead of strings

kiwiupover:
Before:
The asset-size object returned a string `"259.1 KB"` for the file size.

After: 
The file size is returned as raw data `265318` then as it is printed to the command line it is turned into a string. 

The print out is still the same:
image

### DIFF
--- a/lib/models/asset-size-printer.js
+++ b/lib/models/asset-size-printer.js
@@ -10,6 +10,7 @@ function AssetPrinterSize(options) {
 }
 
 AssetPrinterSize.prototype.print = function () {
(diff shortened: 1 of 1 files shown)


rwjblue:
LGTM

kiwiupover:
@rwjblue I changed the this PR a little to return raw bits for a file size then have the function that the returns the data to the user will format to human readable data.

kiwiupover:
Looks like the appveyor tests need a kick .

nathanhammond:
@homu r+

PR #6052, opened 2016-07-09
Turn on strict mode for tests.

### LINKED ISSUE #5389: Cannot execute tests when babel does not transpile 'let' keyword
Happening on 2.3.0-beta.1  in chromium 47

I am getting following exception when executing tests in case babel transpilation of 'let' keyword is disabled.  

`Uncaught SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`

My `ember-cli-build.js` config for babel

```
        babel: {
            comments: false,
            optional: [
                //"es7.asyncFunctions"
                //"es7.functionBind"
            ],
            blacklist: [
                'es6.templateLiterals',
                'es6.blockScoping',
                'es6.constants'
            ]
        },
```

Generally problem is that each non jshint module included in compiled tests.js is missing 'use strict' statement

nathanhammond:
Fixes #5389.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1032,6 +1032,7 @@ EmberApp.prototype._processedEmberCLITree = function() {
     'app-boot.js',
     'test-support-prefix.js',
     'test-support-suffix.js',
+    'tests-prefix.js',
     'tests-suffix.js'
   ];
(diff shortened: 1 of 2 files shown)


rwjblue:
LGTM

nathanhammond:
@homu r+

PR #6077, opened 2016-07-14
[BUGFIX beta] ensure concat order of app.import(x)

stefanpenner:
Files provided to app.import should always be ordered based on insertion order (either append, or prepend).
For this broccoli-concat explicitly provides `headerFiles` and `footerFiles`. `inputFiles` previously used is explicitly un-ordered.
This patch ensures the underlying libraries are correctly used, correctly enforcing expected behavior.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1250,11 +1250,11 @@ EmberApp.prototype.javascript = function() {
 
   var vendorFiles = [];
   for (var outputFile in this._scriptOutputFiles) {
-    var inputFiles = this._scriptOutputFiles[outputFile];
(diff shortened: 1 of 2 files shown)


stefanpenner:
- [ ] tests

rwjblue:
I think we should target `beta` here instead of `release` (2.7.0 final will release in just a week or two), but other than that this LGTM.

stefanpenner:
@rwjblue sounds good, will adjust

stefanpenner:
failing tests may be addressed by:

Turbo87:
as discussed yesterday we should probably target this PR on the `beta` branch

stefanpenner:
retargeting to beta

PR #6082, opened 2016-07-15
Change blueprint .travis.yml files to match travis gem output

backspace:
During the Advanced Ember Workshop at Wicked Good Ember 2016, @mike-north had us run `travis setup heroku` to have successful runs on Travis trigger a deployment to Heroku. That resulted in commit with a diff that was [mostly full of trivial changes](https://github.com/backspace/wge-advanced-2016/commit/7f2bd8c6c23255d680c156606a1a3394a82ba120#diff-354f30a63fb0907d4ad57269548329e3). My suggestion is that Ember CLI include .travis.yml files that match the output generated by the travis gem command.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -1,38 +1,28 @@
----
 language: node_js
 node_js:
-  - "4"
-
+- '4'
 sudo: false
-
 cache:
   directories:
-    - node_modules
-
+  - node_modules
 env:
-  - EMBER_TRY_SCENARIO=default
-  - EMBER_TRY_SCENARIO=ember-1.13
-  - EMBER_TRY_SCENARIO=ember-release
-  - EMBER_TRY_SCENARIO=ember-beta
-  - EMBER_TRY_SCENARIO=ember-canary
-
+- EMBER_TRY_SCENARIO=default
+- EMBER_TRY_SCENARIO=ember-1.13
+- EMBER_TRY_SCENARIO=ember-release
+- EMBER_TRY_SCENARIO=ember-beta
+- EMBER_TRY_SCENARIO=ember-canary
 matrix:
   fast_finish: true
   allow_failures:
-    - env: EMBER_TRY_SCENARIO=ember-canary
-
+  - env: EMBER_TRY_SCENARIO=ember-canary
 before_install:
-  - npm config set spin false
-  - npm install -g bower
-  - bower --version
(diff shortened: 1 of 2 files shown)


kellyselden:
This output doesn't match the indention in the Travis guides https://docs.travis-ci.com/user/customizing-the-build/, so I am not in favor of this.

backspace:
hmm, thanks for pointing that out. I’ll ask why that is, maybe the `travis` gem should be generating output that matches the documentation.

kellyselden:
> maybe the travis gem should be generating output that matches the documentation

Yeah most likely. Thanks for the PR though!

backspace:
@kellyselden, the [documentation](https://docs.travis-ci.com/user/customizing-the-build/) now has similar indentation to the gem. Should I submit another pull request or is there anything else to consider?

nathanhammond:
Following along at the upstream, I'm quite amused by this exchange. 😛 (Take a look @kellyselden.)

This is the really annoying type of change for users when doing a re-init on an Ember CLI upgrade. Lots of blueprint changes, but literally nothing of real value. If we're doing this I want to make 100% sure that we've got it right the first time and don't have to subject our users to this change multiple times.

Can we wait for confirmation of intentional vs. accidental behavior in the gem? Other than that, I'm in favor of standardizing.

stefanpenner:
> If we're doing this I want to make 100% sure that we've got it right the first time and don't have to subject our users to this change multiple times.

👍 

@drogus can you chime in re: ^

kellyselden:
We also should look to and .yml file standards or best practices. I'm not familiar if there are any, but those may trump all.

drogus:
I think that I'd be against that change. The problem with `travis` command currently is that it uses default `to_yaml` formatting. It also means that it will remove comments, which I think are valuable in a lot of cases. I think that maybe we should look into fixing `travis` command to work better when it comes to diffs? I'm not sure how hard it is, but I think that we discussed making it less obtrusive in the past. I also don't think that we could guarantee that the output won't change - it could change even by using a different version of YAML formatter I think (although I'm not sure about that).

stefanpenner:
> I think that I'd be against that change.

If our favorite travis-ci developer says 👎  I believe we should not do this.

backspace:
You can of course accept and reject contributions as fits your project, but I recommend not publicising your favourites within an organisation, as it’s alienating and exclusionary.

PR #6090, opened 2016-07-18
Enable long stack stitching

stefanpenner:
not just for RSVP, but all forms of async (even for those deep in our dependency graph)

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+require('longjohn');
+
 // Main entry point
 var requireAsHash = require('../utilities/require-as-hash');
 var packageConfig = require('../../package.json');
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "isbinaryfile": "^3.0.0",
     "leek": "0.0.23",
     "lodash.template": "^4.2.5",
+    "longjohn": "^0.2.11",
     "markdown-it": "8.0.0",
     "markdown-it-terminal": "0.0.4",
     "minimatch": "^3.0.0",


rwjblue:
For those interested, `longjohn` is here: https://github.com/mattinsler/longjohn

I'm 👍  on this, seems like a good addition.

stefanpenner:
failure seems unrelated...

stefanpenner:
@rwjblue we can configure the separator between the original and stitched stack. Current it is `--------------------------------` is that fine?

Turbo87:
> While it is very helpful in development and testing environments, it is not recommended to use longjohn in production. The data collection puts a lot of strain on V8's garbage collector and can greatly slow down heavily-loaded applications.

are we sure we want to do this? or is this just meant for `canary`/`beta`?

stefanpenner:
@Turbo87 test failures suggest not, but i have hope :P

stefanpenner:
rebased, going to investigate the previous test failures next.

stefanpenner:
This may be causing a hefty slow-down. (Which isn't surprising) we could enable this as a flag or env variable?


```js
DEBUG_LONG_STOCK=true
```

Turbo87:
question is if we should add the dependency for the 1% of all people that actually discover that env var. iirc NPM has optional dependencies though, so maybe envvar + optional dep might be a good idea here.

stefanpenner:
> question is if we should add the dependency for the 1% of all people that actually discover that env var. iirc NPM has optional dependencies though, so maybe envvar + optional dep might be a good idea here.

todays optionalDependencies are very noisy, and the weight of this dependency is low. So I believe either we include it by default \w flag, or we don't include it at all.

nathanhammond:
This is a pretty small change. The people who are likely to need it will also likely be debugging something else. Can we collapse this under an existing flag?

stefanpenner:
This likely needs API design. I don't really need this, I can easily add it as i require it. If someone wants this to be easier for non-power users a design for how to enable it should be provided and we can revive.

PR #6092, opened 2016-07-20
Remove ember-qunit-notifications

trentmwillis:
Should no longer be needed with the newer versions of `ember-cli-qunit` (ref: https://github.com/ember-cli/ember-cli-qunit/pull/119).

Forgot to do this sooner and now confusion is ensuing (ref: https://github.com/DockYard/qunit-notifications/issues/27).


twokul:
@homu r+

PR #6095, opened 2016-07-21
[INTERNAL] Remove unused 'es3Safe' option

ursm:
This option is used from no one already, so I think it can be eliminated safely.


stefanpenner:
@ursm thanks!

@homu r+

PR #6096, opened 2016-07-21
Fix and improve Watcher.detectWatcher

stefanpenner:
- [x] add tests to detectWatcher
- [x] improve fallback scenarios (not all systems node works on support fs.watch)
- [x]  make sure we consistently do the right thing, regardless of which branches we take.
- [x] windows CI seems to be failing for unknown reasons....

Ok, so this refactors the watcher detection code (which was quite complicated, untestable, buggy, and nearly impossible to extend without causing more issues). The goal here is, a separate model which can be tested, extended and improved with confidence. 

I feel This has festered long enough, a refactor was long overdue.

This PR also adds the ability to detect if `node` watcher extremely broken (we could extend this to detect if change events are even possible, but that can come later). If it is extremely broken, we log it and fall back to `polling`. This hopefully will both provide useful information to the user, and attempt a grace-full fallback.


lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+  };
+}
+
+WatchPreference.prototype.watchmanWorks = function(details) {
+  this._watchmanInfo.enabled = true;
+  this._watchmanInfo.version = details.version;
+  this._watchmanInfo.canNestRoots = semver.satisfies(details.version, '>= 3.7.0');
+};
+
+Object.defineProperty(WatchPreference.prototype, 'watchmanInfo', {
```
rwjblue: What do we gain by adding this as a getter?  Just the ability to change in the future?

lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+  get: function() {
+    return this._watchmanInfo;
+  }
+});
+
+module.exports = WatchDetector;
```
rwjblue: This always breaks by brain.  When an export is above the function.  Yes, I know hoisting is a thing, but still my brain hurts.

lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+    return new WatchPreference(WATCHMAN);
+  }
+}
+
+/*
+ * @public
+ *
+ * Detect if `fs.watch` provided by node is capable of observing a directory
+ * within `process.cwd()`. Although `fs.watch` is provided as a part of the
+ * node stdlib, their exists platforms it ships to that do not implemented the
```
rwjblue: s/implemented/implement/

lib/models/watch-detector.js
stefanpenner: can't set to it from the outside. (its readOnly)

lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+ * required sub-system. For example, there exists `posix` targets, without both
+ * inotify and FSEvents
+ *
+ * @method testIfNodeWatcherAppearsToWork
+ * @returns {Boolean) outcome
+*/
+WatchDetector.prototype.testIfNodeWatcherAppearsToWork  = function() {
+  return new Promise(function(resolve) {
+    var key = path.dirname(path.dirname(process.cwd()));
```
rwjblue: Is there a specific reason to go two directories up? This is somewhat confusing to me.  I would think that you would want to use the actual `project.root` which is generally `process.cwd()`.

lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+      return;
+    }
+
+    this._doesNodeWorkCache[key] = false;
+    try {
+      // builds a tmp directory at process.cwd() + '/tmp/' + 'something-unique';
+      var tmpDir = quickTemp.makeOrRemake(this, 'tmp');
+
+      this.fs.watch(tmpDir, { persistent: false, recursive: false });
+      this.fs.unwatch(tmpDir);
```
rwjblue: Should we actually change the file, and confirm it works?

lib/models/watch-detector.js
```diff
@@ -0,0 +1,238 @@
...
+ *  * watchman executable may be something unexpected
+ *  * ???
+ *
+ *  @method checkWatchman
+ *  @returns {Object} watch preference + _watchmanInfo
+*/
+WatchDetector.prototype.checkWatchman = function() {
+  var result = new WatchPreference(null);
+
+  return this.childProcess.exec('watchman version').then(function(output) {
```
rwjblue: Might want a `debug` for this check, and its result (including the error case when non-JSON is emitted)

tests/unit/models/watch-detector-test.js
```diff
@@ -0,0 +1,273 @@
...
+        expect(ui.output).to.not.match(/Could not start watchman/);
+        expect(ui.output).to.not.match(/falling back to NodeWatcher/);
+        expect(ui.output).to.not.match(/ember-cli\.com\/user-guide\/\#watchman/);
+        expect(ui.output).to.not.match(/Looks like you have a different program called watchman/);
+        expect(ui.output).to.not.match(/Invalid watchman found/);
+      });
+    });
+
+    describe('fallse back to NODE', function() {
+      var iff = it;
```
rwjblue: O_o

rwjblue:
This looks great! I left a few comments, but mostly just things I was thinking as I reviewed.  Only one that seems like it may be wrong (though you may correct me here).

tests/unit/models/watch-detector-test.js
stefanpenner: the lines bellow this read better as "iff(xyz)"... "if and only if xyz"

lib/models/watch-detector.js
stefanpenner: Ya, going up 2 gives us the stable directory in all cases, But should be explore further in app usage (rather then test usage scenarios)

tests/unit/models/watch-detector-test.js
rwjblue: heh, they only read better if you know that `iif` means "if and only if". You CS people. 😈

stefanpenner:
> This looks great! I left a few comments, but mostly just things I was thinking as I reviewed. Only one that seems like it may be wrong (though you may correct me here).

https://en.wikipedia.org/wiki/If_and_only_if apparently its MATHs people

rwjblue:
HAHA, MATHS! 🍇

lib/models/watch-detector.js
stefanpenner: some future PR could, this one merely makes sure the call does throw. Like it does on platforms that don't have support.

lib/models/watch-detector.js
stefanpenner: I'm going to re-work this yet.

stefanpenner:
the following appears to be possible, but should not.
```
was unable to use: "events", fell back to: "watchman"
```

- [ ] fix this

stefanpenner:
crap forgot about this, must ship soon.

lib/models/watch-detector.js
stefanpenner: update to be project.root

stefanpenner:
lets see what CI says.

jonnii:
https://www.youtube.com/watch?v=AJQ3TM-p2QI

stefanpenner:
Looks like one of the windows CI runs passed, but one failed (node 4.x) but failed due to connection \w github issue. 

Re-running...

stefanpenner:
Looks like all is green. Any reviewers ?

PR #6097, opened 2016-07-21
[BUGFIX beta] ServerWatcher disregards --watcher=*

stefanpenner:
* models/server-watcher is now a subclass of models/watcher, delegating all arguments parsing to models/server (this should prevent them from diverging)
* models/server now makes constructing the watcher instance an optional refinement a subclass can add. This is done by extracting watcher construction to `constructWatcher`
* make models/server-watcher output more consistent

### DIFF
--- a/lib/models/server-watcher.js
+++ b/lib/models/server-watcher.js
@@ -1,62 +1,48 @@
 'use strict';
 
-var Task    = require('./task');
+var Watcher = require('./watcher');
 
-module.exports = Task.extend({
-  verbose: true,
+var ServerWatcher = Watcher.extend({
+  constructWatcher: function(options) {
+    return new (require('sane'))(this.watchedDir, options);
+  },
 
   init: function() {
-    this._super.apply(this, arguments);
-    this.watcher = this.watcher || new (require('sane'))(this.watchedDir, {
-      verbose: this.verbose,
-      poll: this.polling()
-    });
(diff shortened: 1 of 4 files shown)


stefanpenner:
This manifests fatally for systems that don't support NodeWatcher (basically there exists posix or posix like environments that have neither FSEvents or Inotify, and `fs.watch` just explodes). With this fix, such users can use `--watcher polling` as an interim solution, without this fix if they have a `server/` directory, the server watcher will disregard and watch likely falling back to NodeWatcher. Which causes tears.

This likely affects all users trying the "Bash on Ubuntu On Windows" stuff MS released earlier this year: https://msdn.microsoft.com/en-us/commandline/wsl/about

With this fix, it becomes possible to use ember-cli in that environment.

thanks goes to @wycats for letting me know about the issue, and letting me debug on his laptop.

stefanpenner:
I suspect `[BUGFIX beta]` is fine, as bash on windows isn't really a platform we claim to support, and this issue most likely does not affect the majority of our current user base.

rwjblue:
@stefanpenner - Beta seems good to me also.

stefanpenner:
> @stefanpenner - Beta seems good to me also.

Sounds good, i'll retarget to the beta branch then

stefanpenner:
retargeted to beta: https://github.com/ember-cli/ember-cli/pull/6098

PR #6105, opened 2016-07-25
Update Ember Data to v2.7.0

bmac:


rwjblue:
👍 

/cc @twokul this is needed for release

PR #6111, opened 2016-07-27
set bowerDirectory in Addon

xcambar:
__tl;dr;__ The new import syntax allows for nested addons, but nested addons don't have the `bowerDirectory` property that allows to locate bower assets.

Considering this:

```js
// my-addon/index.js
module.exports = {
  name: 'my-addon',
  included: function(app) {
    var assetPath = app.bowerDirectory + '/whatever/dist/assets.min.js';
    this.import(assetPath);
  }
};
```

* When included in an app, `assetPath === "bower_components/whatever/dist/assets.min.js"` (by default).
* When included in an addon, `assetPath === "undefined/whatever/dist/assets.min.js"`.

This PR fixes the second behaviour to return the correct path in both cases.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -91,6 +91,7 @@ var Addon = CoreObject.extend({
     this._super();
     this.parent = parent;
     this.project = project;
+    this.bowerDirectory = project.bowerDirectory;
     this.ui = project && project.ui;
     this.addonPackages = {};
     this.addons = [];
--- a/tests/unit/models/addon-test.js
+++ b/tests/unit/models/addon-test.js
@@ -23,6 +23,21 @@ var fixturePath = path.resolve(__dirname, '../../fixtures/addon');
 describe('models/addon.js', function() {
   var addon, project, projectPath;
 
+  describe('constructor', function() {
+    it('should set bowerDirectory for addon', function() {
+      projectPath = path.resolve(fixturePath, 'simple');
+      var packageContents = require(path.join(projectPath, 'package.json'));
+
+      project = new Project(projectPath, packageContents);
+      var addon = new Addon({
+        project: project
+      });
+
+      expect(addon.bowerDirectory).to.equal(project.bowerDirectory);
+      expect(addon.bowerDirectory).to.equal('bower_components');
+    });
+  });
+
   describe('root property', function() {
     it('is required', function() {
       expect(function() {


nathanhammond:
Hey @xcambar! There appear to be legitimate test failures on this PR. I'm pretty sure that this is a behavior we would like to support, but doing the right thing here is a bit more complicated and will require us to spend a bit of time thinking about the problem.

- Should nested addons be allowed to use their grandparents' `bower_components` directory?
- How would a nested addon ensure that the dependency that they need is present in that directory?
- Nested addons don't automatically have their blueprint invoked and don't exactly have a hook to run their own `bower` installations except during `included`.

What do you think we should do to solve this?

xcambar:
My thoughts currently tend to suggest hooks and blueprints should be run recursively. And that this discussion should be moved to an RFC (I'll do it if you confirm) 😄. Here are the details:

> Should nested addons be allowed to use their grandparents' bower_components directory?

I think they should. Ember-CLI has grown to allow addons to manipulate and automate bower dependencies. I would suggest that not breaking an app's `bower_components` or `bower.json`  is the responsibility of the addon developer, as, in its current state, ember-cli doesn't prevent unintentional breakage of everything bower-related, and I don't see a reason not to extend this power/responsibility to nested addons.

> How would a nested addon ensure that the dependency that they need is present in that directory?

Top-level addons ensure the presence by providing and running a default blueprint. Same should go with nested addons, IMHO.

> Nested addons don't automatically have their blueprint invoked and don't exactly have a hook to run their own bower installations except during included.

As you know better than me, included is here (mostly) to add deps to the tree while building the app. It's not a convenient place to install new stuff in the FS. I used to add a blueprint to my top-level addon that runs blueprints for the nested addons. See the following snippet for a rough example:

```js
// blueprints/my-addon/index.js
module.exports = {
  normalizeEntityName: function() {},
  afterInstall: function() {
    var that = this;
    var args = arguments;
    var promises = ['ember-cli-selectize', 'ember-cli-showdown', 'ember-pikaday']
      .map(function(name) {
        return `${name}/blueprints/${name}`;
      })
      .map(require)
      .map(function(mod) {
        return mod.afterInstall.apply(that, args);
      });
    return Promise.all(promises);
  }
};
```

I'm pretty confident this kind of logic can be adapted for a generic use case.

---

Questions:

* Does ember-cli warn about conflicting versions of Bower components (eg, `package: ~1.0` is specified in bower.json, addon wants to install `package: ~2.0`)? The risk is high(er) with nested addons.
* What should happen if/when the same package is used as both a top-level addon and a nested addon, with conflicting versions? Does it fall under npm's umbrella?

nathanhammond:
Let's definitely move this discussion to an RFC. Amusingly, setting a single property is going to require a really long and complicated review process to decide what the behavior should be in every scenario.

I think I'd start by specifying in some sort of table what the behavior should be in every single case: accounting for engines, addons, nesting, and more.

nathanhammond:
(Closing, expecting to see it reopened on the RFC repo.)

xcambar:
Please see https://github.com/ember-cli/rfcs/pull/63

PR #6112, opened 2016-07-27
More specific docs for included hook

xcambar:
According to the doc, the `included` hook receives an EmberApp instance as its first )and only) argument.

When nesting addons, the first argument can be an EmberApp (first-level addon) or an Addon (nested addon).

This PR makes the docs more specific about this.


lib/models/addon.js
```diff
@@ -412,16 +412,16 @@ var Addon = CoreObject.extend({
 
     @public
     @method included
-    @param {EmberApp} app The application object
+    @param {EmberApp|Addon} parent The parent object that included this addon
```
nathanhammond: I'm torn about the param type definition because of this: https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-addon.js#L52-L56

Edit:
```diff
- that
+ which
```

lib/models/addon.js
```diff
@@ -412,16 +412,16 @@ var Addon = CoreObject.extend({
...
     @example
     ```js
-    included: function() {
+    included: function(parent) {
       this.import(somePath);
     }
     ```
   */
-  included: function(/* app */) {
+  included: function(/* parent */) {
```
nathanhammond: The generic `parent` value here is probably for the best as we're looking at have `Engines`, `Addons`, and `Applications` soon.

nathanhammond:
I'm in favor of landing this PR with the edits I've added. :smile: Thanks for your attention to detail, @xcambar!

lib/models/addon.js
xcambar: I understand your concern about the param definition. Though as you can guess, I'm in favour of verbosity and clarity on this one.

nathanhammond:
@homu r+

xcambar:
uuh... what? https://travis-ci.org/ember-cli/ember-cli/jobs/147963030#L2592

nathanhammond:
@homu retry

nathanhammond:
CI is broken. I know this PR is good, merging it.

PR #6119, opened 2016-07-28
[ENHANCEMENT] Allow tests to be passed into moduleForAcceptance

ryanto:
One of the biggest complaints I hear about QUnit when compared to mocha is that you cannot nest tests. QUnit supports nested modules (https://api.qunitjs.com/QUnit.module/), but it's hard to do with ember-cli's current `moduleForAcceptance` blueprint.

This PR adds a non breaking change to `moduleForAcceptance` allowing the tester to pass in a nested block of tests. The following would now work.

```javascript
import { test, module } from 'qunit';
import moduleForAcceptance from 'my-app/tests/helpers/module-for-acceptance';

let a, b, c;

moduleForAcceptance('Acceptance | Nested 1', {
  beforeEach() {
    a = 'a';
  },
  afterEach() {
    a = undefined;
  }
}, function() {

  // pass tests into moduleForAcceptance in order
  // to use nested modules.

  test('a', function(assert) {
    assert.equal(a, 'a');
  });

  // example of nested module
  module('Nested 2', {
    beforeEach() {
      b = 'b';
    },
    afterEach() {
      b = undefined;
    }
  }, function() {

    // nested tests
    test('nested a', function(assert) {
      assert.equal(a, 'a');
    });

    test('b', function(assert) {
      assert.equal(b, 'b');
    });

    // old syntax will work as well
    module('Nested 3', {
      beforeEach() {
        c = 'c';
      },
      afterEach() {
        c = undefined;
      }
    });

    test('3 levels', function(assert) {
      assert.equal(a, 'a');
      assert.equal(b, 'b');
      assert.equal(c, 'c');
    });
  });
});
```

In the QUnit UI these tests will have nice names.

The dropdown will look like this:

and the runner output will look like this.

### DIFF
--- a/blueprints/app/files/tests/helpers/module-for-acceptance.js
+++ b/blueprints/app/files/tests/helpers/module-for-acceptance.js
@@ -5,8 +5,8 @@ import destroyApp from '../helpers/destroy-app';
 
 const { RSVP: { Promise } } = Ember;
 
-export default function(name, options = {}) {
-  module(name, {
+export default function(name, options = {}, tests = undefined) {
+  let hooks = {
     beforeEach() {
       this.application = startApp();
 
@@ -19,5 +19,7 @@ export default function(name, options = {}) {
       let afterEach = options.afterEach && options.afterEach.apply(this, arguments);
       return Promise.resolve(afterEach).then(() => destroyApp(this.application));
     }
-  });
+  };
+
+  module(name, hooks, tests);
 }


kellyselden:
I love it and want to see nested modules, but I think we need to flesh out the API in an RFC. We might have to make sure it is compatible with https://github.com/emberjs/rfcs/pull/119.

trentmwillis:
I would like for all the Ember testing helpers get to a point where they're essentially pass-throughs for the `module` methods. In that regard, I think this is a step in the right direction.

I would like to see this land ahead of the testing RFC since that has no specified timeline and this is an obvious improvement.

kellyselden:
> I would like to see this land ahead of the testing RFC since that has no specified timeline and this is an obvious improvement.

@trentmwillis Me too, but since we are getting really serious about SemVer, I don't want to be locked into an API if it hasn't been thoroughly vetted.

trentmwillis:
> I don't want to be locked into an API if it hasn't been thoroughly vetted.

I want us to lock in on the QUnit API, but I understand that might need to accommodate some special cases from the Ember community.

nathanhammond:
@trentmwillis Can you provide line commentary on what needs to change in the PR?

Turbo87:
wouldn't it be better to use https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-acceptance.js instead?

Turbo87:
@nathanhammond @kellyselden I've tried to wrap my head around the situation again, and my conclusion is that we should first improve the API in `ember-qunit` to take advantage of the `test-module-for-acceptance` stuff and after that this PR will most likely not be needed anymore because we'll be able to remove some of those files.

nathanhammond:
@ryanto We discussed this yesterday during our team meeting. [Our notes from that conversation.](https://github.com/emberjs/core-notes/blame/master/ember-cli/2016-09/september-15.md#L48-L55)

@Turbo87's further review of this confirmed that judgment call we made yesterday, so I'm closing this.

We're none the less grateful for your PR in spite of closing it. This spurred valuable conversation which will make this story better. Thank you very much!

PR #6125, opened 2016-07-29
fix command in release process

kellyselden:
It was generating app code into the addon repo. I am fixing all the bad commits it generated.


kellyselden:
fixed the bad commits in https://github.com/ember-cli/ember-addon-output

PR #6139, opened 2016-08-02
Updating app/addon blueprints to latest dependency versions

elwayman02:


stefanpenner:
awesome, thanks!

elwayman02:
Whoo passed!

twokul:
@homu r+

Turbo87:
other than the shims update what's the reason for this? since they use caret ranges the deps should use the most recent versions anyway and updating them just creates unnecessary work for people updating their apps.

elwayman02:
I would actually argue that it creates less work in the long run. Instead
of forcing individual developers to do this work manually, now they can
simply run `Ember init` to get this for free. Because it's still within the
karat versions, they don't have to worry about regressions. Additionally,
there is some functional value, as platforms like Travis-CI have been known
to aggressively cache packages and troll people by running older versions
that still satisfy the fuzzy versioning, rather than pulling in the latest
allowed release. Updating the minimum version will invalidate the cache so
that developers don't have to manually troubleshoot why their app didn't
pick up the latest package version that they have on their local
environment.

twokul:
@Turbo87 I run into so many deprecation (`this._super`) issues b/c version weren't resolved properly  or there were 300500 versions of `ember-cli-htmlbars` which differed in point version

Turbo87:
good point @twokul, I guess it's worth updating then.

lrsnider:
When I upgraded to ember 2.8.0, running `ember init` deleted a handful of dependencies that I am now in the process of checking compatibility with this latest stable ember release. Is there a quicker fix to cross-check compatibility? Also wondering why `ember init` (or whatever other step in the upgrade CLI process it was) caused such a significant diff in my `package.json` and `bower.json`. @stefanpenner @twokul TIA.

kellyselden:
@lrsnider ember init is not a perfect tool, it just resets your app state to the default. If you have a complex app, you can use diffs in the [changelog](https://github.com/ember-cli/ember-cli/blob/master/CHANGELOG.md) (ie https://github.com/ember-cli/ember-new-output/compare/v2.7.0...v2.8.0) to fine tune your upgrade (it's how I do it).

Issue #6042, opened 2016-07-06
Nested Addons - `app.import` is undefined


xcambar:
Following up on  #4616, it occurs that every addon developer that wants to nest addons faces this code snippet in their hooks: https://github.com/rwjblue/ember-cli-pretender/commit/33e5794f5aef545f00995435d08cb471e79a3bae#diff-168726dbe96b3ce427e7fedce31bb0bcR10, copied/pasted here:

```
if (app.app) {
  app = app.app;
}
```

Would that be an acceptable PR to move this snippet before the `included` hook, so addon developers don't have to forget-and-deal-with-it-again every now and then?

rwjblue:
We added support for `this.import` in each addon to avoid this issue, but there is an issue in https://github.com/ember-cli/ember-cli/issues/5990 that needs to be corrected.

Using `this.import` (instead of just providing access to the top level app instance) allows us to do more intelligent things (like track `.import`'s by addon).

nathanhammond:
I'm closing this issue as a duplicate. @xcambar If you'd like to open a PR to port some of the code that @rwjblue wrote in the linked issue into ember-cli core and write a few tests we'd be eternally grateful. This is the only issue blocking our 2.7 release.

PR #6148, opened 2016-08-03
Update to _findHost to use do/while.

nathanhammond:
I'm pretty sure that this is identical to the current implementation, but y'all make sure I'm not crazy.

/cc @Turbo87 @xcambar @rwjblue


nathanhammond:
Related to #6042 and #6043.

rwjblue:
LGTM

nathanhammond:
@homu r+

stefanpenner:
👍

PR #6161, opened 2016-08-08
Update .travis.yml

stefanpenner:


Turbo87:
👍

nathanhammond:
If we're doing this we should land #6082 at the same time.

Turbo87:
@nathanhammond for the reasons given in that PR I don't think #6082 should be merged...

PR #6171, opened 2016-08-11
Adding a test to cover historySupportMiddleware with unknown location type

jasonmit:
Just adding a simple test to cover a discussion around the behavior of `historySupportMiddleware`.  This validates it will work correctly if you provide a locationType that might extend from HistoryLocation and the user wants to force the user of the history support middleware.

/cc @stefanpenner


Turbo87:
@jasonmit thanks!

@homu r+

PR #6180, opened 2016-08-16
bump portfinder to v1.0.7

### LINKED ISSUE #6083: `ember s` fails with `EADDRNOTAVAIL` after updating to `port-finder@1.0.4`

eriktrom:
Will push upstream dep `portfinder@1.0.7` once we receive consensus from the crowd that this works on all machines. This PR is for book keeping.

See https://github.com/ember-cli/ember-cli/issues/6083 to help out with the testing process. It should only take a minute of your time. :)

fixes #6083

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "npm": "2.15.5",
     "npm-package-arg": "^4.1.1",
     "ora": "^0.2.0",
-    "portfinder": "^1.0.4",
+    "portfinder": "^1.0.7",
     "promise-map-series": "^0.2.1",
     "quick-temp": "0.1.5",
     "resolve": "^1.1.6",


eriktrom:
@stefanpenner - ready to go when ci passes (although this is just book keeping, so merge at will)

stefanpenner:
@homu r+

Turbo87:
@stefanpenner @nathanhammond FYI https://github.com/indexzero/node-portfinder/blob/master/package.json#L32

stefanpenner:
@Turbo87 hmmm, thanks for pointing that point. @eriktrom is that a hard dependency on node >= 0.10 ?

eriktrom:
@stefanpenner - sorry whats the concern? Does portfinder only work w/ node >= 0.12? It should work with 0.10, I believe I bumped that at some point for travis, but the way the code is written now should work with 0.10 unless we're seeing issues in the wild where this is not the case

Turbo87:
@eriktrom the only issue I noticed so far is that NPM prints a compatibility warning during install when used with Node 0.10. if the package works with Node 0.10 then everything is fine.

eriktrom:
@Turbo87 - gotcha - yes it'll print warnings but should work just fine AFAIK

eriktrom:
I can change that to 0.10 if we think that's ideal - let me know, no biggie

Issue #6178, opened 2016-08-15
New "rootURL" option is incompatible with Cordova


kanongil:
Before the `rootURL` option was introduced (#5792), it was possible to create compiled build products that could be deployed to any base URL. Unfortunately, the `rootURL` option ties the build output to a specific base path by resolving resources absolutely, which makes it incompatible with apps that rely on this specific deploy mechanism.

This failure can be demonstrated by directly opening the `dist/index.html` of a simple ember app in a browser, using its `file:` URL. This worked in 2.6 (when `locationType='hash'`), but fails in 2.7 since it tries to locate `vendor.js` etc. at `file:///vendor.js`.

Console output for freshly create app compiled using `ember build`:
```
[Error] Failed to load resource: The requested URL was not found on this server. (vendor.css, line 0)
[Error] Failed to load resource: The requested URL was not found on this server. (new.css, line 0)
[Error] Failed to load resource: The requested URL was not found on this server. (vendor.js, line 0)
[Error] Failed to load resource: The requested URL was not found on this server. (new.js, line 0)
```

I would like to note that eg. Cordova relies on rendering from a `file:` URL, and that the actual absolute path is not static. One example `file:///Users/gil/Library/Developer/CoreSimulator/Devices/924FAEC4-0F05-43BE-8939-ED00118627AF/data/Containers/Bundle/Application/38FE7E97-3198-4E04-B78A-17BE5E4C5090/MyApp.app/www/index.html`

---

Output from ember version --verbose:
```
ember-cli: 2.7.0
http_parser: 2.5.2
node: 4.4.7
v8: 4.5.103.36
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 46
openssl: 1.0.2h
os: darwin x64
```

rwjblue:
If all you need/want is for things to be referenced relative to the `index.html`, you do not need to use `{{rootURL}}` in your `index.html` at all.

Another option is to set `rootURL` in `config/environment.js` ([around here](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/environment.js#L7)) to `''` which would mean the generated asset paths in `index.html` after `{{rootURL}}` is processed would be something like `src="assets/foo.js"` and `src="assets/vendor.js"` both of which are now relative.

Does this address your concern? If not, what have I miss understood?

kanongil:
[...]

PR #6182, opened 2016-08-16
[BUGFIX] Allow empty string as rootURL

kanongil:
This enables the config parser & server to handle `rootURL` with an empty string `''`. See #6178 for details on why.

With this fix it is possible to create relative-only build products when `locationType = 'hash'`.

FYI, I have another PR pending on this, which enables the server to correctly serve this when `locationType = 'history'` by redirecting all non-root requests to the root location.

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -46,11 +46,11 @@ module.exports = Command.extend({
 
         this.ui.writeDeprecateLine(
           'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
-          !(!config.rootURL && config.baseURL));
+          !(!('rootURL' in config) && config.baseURL));
 
         this.ui.writeWarnLine(
           'The `baseURL` and `rootURL` settings should not be used at the same time.',
-          !(config.rootURL && config.baseURL));
+          !(('rootURL' in config) && config.baseURL));
 
         commandOptions = assign({}, commandOptions, {
           rootURL: config.rootURL,
--- a/lib/tasks/server/middleware/serve-files/index.js
+++ b/lib/tasks/server/middleware/serve-files/index.js
@@ -25,7 +25,7 @@ ServeFilesAddon.prototype.serverMiddleware = function(options) {
     autoIndex: false // disable directory listings
   });
 
-  var baseURL = cleanBaseURL(options.rootURL || options.baseURL);
+  var baseURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL || options.baseURL);
 
   logger.info('serverMiddleware: baseURL: %s', baseURL);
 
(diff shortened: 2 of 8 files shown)


lib/commands/serve.js
```diff
@@ -46,11 +46,11 @@ module.exports = Command.extend({
 
         this.ui.writeDeprecateLine(
           'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
-          !(!config.rootURL && config.baseURL));
+          !((typeof config.rootURL !== 'string') && config.baseURL));
```
rwjblue: These warnings should change to `in` checks (as we don't really care about the values, just that the property is present).

lib/tasks/server/middleware/testem-url-rewriter/index.js
```diff
@@ -17,11 +17,11 @@ TestemUrlRewriterAddon.prototype.testemMiddleware = function(app) {
 
   this.project.ui.writeDeprecateLine(
     'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
-    !(!config.rootURL && config.baseURL));
+    !((typeof config.rootURL !== 'string') && config.baseURL));
```
rwjblue: Should use `in`

lib/tasks/server/middleware/testem-url-rewriter/index.js
```diff
@@ -17,11 +17,11 @@ TestemUrlRewriterAddon.prototype.testemMiddleware = function(app) {
...
 
   this.project.ui.writeWarnLine(
     'The `baseURL` and `rootURL` settings should not be used at the same time.',
-    !(config.rootURL && config.baseURL));
+    !((config.rootURL || config.rootURL === '') && config.baseURL));
```
rwjblue: `in`

lib/commands/serve.js
kanongil: Ok, I initially considering doing an `in` check. Will update.

kanongil:
PR updated. The test failure seems intermittent.

nathanhammond:
@homu r+

lib/tasks/server/middleware/serve-files/index.js
```diff
@@ -25,7 +25,7 @@ ServeFilesAddon.prototype.serverMiddleware = function(options) {
     autoIndex: false // disable directory listings
   });
 
-  var baseURL = cleanBaseURL(options.rootURL || options.baseURL);
+  var baseURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL || options.baseURL);
```
Turbo87: is this necessary? cleanBaseURL should automatically make it `/`, right?

kanongil: Yes, but on `rootURL=''` it will evaluate to false and use `baseURL` instead.

Turbo87: is that a problem? `baseURL` is deprecated and shouldn't be used and especially not in combination with `rootURL`

nathanhammond: I prefer the precision here until we fully deprecate it.

kanongil: > is that a problem?

Yes. `cleanBaseURL('' || undefined)` resolves to `undefined`and not `'/'`.

PR #6186, opened 2016-08-17
[ENHANCEMENT] Warn when empty rootURL is used with history addon

kanongil:
With `rootURL = ''` and `locationType='auto'/'history'` the app doesn't work for nested entry paths like `http://localhost:4200/my/test`.

This is due to failing to resolve the relative location of the basic assets referenced in `index.html` like `assets/vendor.js`.

This patch fixes it by catching such requests on the server, and redirecting them to the `rootURL` appended with the hash-based location. Once the page is loaded the built-in router will rewrite the URL to the originally requested URL.

I have configured this feature to be triggered automatically when `rootURL = ''`.

### DIFF
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -7,7 +7,7 @@ var cleanBaseURL = require('clean-base-url');
 
 /**
  * This addon is used to serve the `index.html` file at every requested
- * URL that begins with `baseURL` and is expecting `text/html` output.
+ * URL that begins with `rootURL` and is expecting `text/html` output.
  *
  * @class HistorySupportAddon
  * @constructor
@@ -31,6 +31,10 @@ HistorySupportAddon.prototype.shouldAddMiddleware = function(environment) {
 
 HistorySupportAddon.prototype.serverMiddleware = function(config) {
   if (this.shouldAddMiddleware(config.options.environment)) {
+    this.project.ui.writeWarnLine(
+      'Empty `rootURL` is not supported. Disable history support, or use an absolute `rootURL`',
+      config.options.rootURL !== '');
+
     this.addMiddleware(config);
   }
 };


nathanhammond:
@kanongil An approach like this would be a tremendous boon to users trying to maintain consistency across multiple environments. However, I don't believe that we can do this because of the _chance_ that somebody has defined routes which collide with asset paths. This is legal (if unwise) and I believe it breaks in your proposed approach when accounting for FastBoot:

```
this.route('entry', function() {
  this.route('assets', function() {
    this.route('vendor.js');
  });
});
```

=> `http://localhost:4200/entry/`

Even assuming we could do something reasonable for all of the assets inside of `index.html` we would need to know _all_ of the assets which could be loaded in using a relative path throughout the entire application lifecycle to guarantee no collisions to opt in to this. This suffers from the halting problem. 😜

Not closing yet–I'd still like to be proven wrong–but I want to see a lot of tests showing really trolling collisions and support for FastBoot.

kanongil:
@nathanhammond Thank you for taking the time to review this, and enlightening me on possible conflicts.

First of all this is a new mode, triggered by setting `rootURL = ''`. For any other configuration this will have no effect.

For non-fastboot, your example should go like this:

1. Navigate to `http://localhost:4200/entry/`
1. Server redirects to `http://localhost:4200/#/entry/`
1. Server serves `index.html` at `http://localhost:4200/`.
1. All relative references are loaded correctly based of the location of `index.html`.
1. App routes based on the `#/entry/` fragment and rewrites url to `http://localhost:4200/entry/`

For fastboot, it is already broken and doesn't change anything. Maybe just add a warning?

Turbo87:
IMHO `rootURL = ''` is a special cases that is only meant to support cordova and the `hash` location type. Instead of the proposed solution I'd rather see a warning printed to the console if someone is trying to combine other location types with the empty `rootURL`.

kanongil:
@Turbo87 I am fine with that solution as well. I don't actually need this myself.

nathanhammond:
@kanongil We talked about this today and don't really want to try to have a "perfect" story for every imaginable use case. In this case, since we know it's broken, we'd rather just throw an error and provide a notice to the user.

Can you modify this PR to implement that behavior?

kanongil:
Updated to just throw a warning, as requested.

nathanhammond:
@homu r+

PR #6194, opened 2016-08-18
[BUGFIX beta] Prevent Ember Data from overriding Date.parse.

bmac:
This updates the defaults for `config/environment.js` so Ember Data no longer logs a deprecation warning when it overrides `Date.parse`.

Crosslink: https://github.com/emberjs/data/issues/4501

### DIFF
--- a/blueprints/app/files/config/environment.js
+++ b/blueprints/app/files/config/environment.js
@@ -10,6 +10,10 @@ module.exports = function(environment) {
       FEATURES: {
         // Here you can enable experimental features on an ember canary build
         // e.g. 'with-controller': true
+      },
+      EXTEND_PROTOTYPES: {
+        // Prevent Ember Data from overriding Date.parse.
+        Date: false
       }
     },
 


rwjblue:
Looks good to me. Do we need to backport this to beta branch?

bmac:
Sure, whats the process for that in the `ember-cli` repo? Should I update the commit message to say `[BUGFIX beta]`?

nathanhammond:
Unfortunately the "change base branch" feature is not very good right now.

@Turbo87 Want to come up with a magic `git` command we can start pasting in that will try to do the right thing? (Feeling lazy right now.)

Turbo87:
@nathanhammond What I'm doing is usually `git rebase --interactive <new-target-branch>` and then remove all the commits that are unrelated to what I've actually done. There may be ways to do that removing automatically, but I'm afraid I don't know any yet.

Turbo87:
@homu r+

nathanhammond:
@homu retry

morhook:
A pretty similar code on `Ember 2.6.2` took the prototypes extension of Array, Functions and String prototypes. Isn't this doing the same? I think the generated code should be:

```
    EXTEND_PROTOTYPES: {
      Date: false,
      Array: true,
      Function: true,
      String: true,
    }
```

cibernox:
I can confirm that this change disabled all prototype extensions, not only the date one (at least in Ember 2.5).

I'll try later to see if that is the case in 2.8 and if so, it can be fixed by changing this blueprint to the snippet shared above by @morhook, or by changing Ember in a few places (like [here](https://github.com/emberjs/ember.js/blob/de4d6b6f2de3a82d3ec787f95c1b0044cc86111f/packages/ember-runtime/tests/system/string/decamelize_test.js#L6) to check explicitly assert that it's false, not undefined)

nathanhammond:
@cibernox Can you open this as a new issue? I want to have a conversation about explicit truthiness, falsiness, and deprecation plans (if any) for prototype extensions.

Issue #6147, opened 2016-08-03
reduce bite-size/file-count of ember-cli and its dep.


stefanpenner:
# why does this matter?

npm installs are dominate by:
- download times
- local IO moving/copying etc files around

By reducing this overhead, we help reduce both on-disk bytes, number on files, and improve install times. Etc.

It is also the case that many of our dependencies publish accidental files, or include extra dependencies. At scale, the cost becomes non-trivial

This will likely be a long running issue, we should likely get a task force together to address this.
## Stuff this aims to address
- download times
- untar/move times
- deletion times
- on-disk bytesize cost
- complexity?

somewhat related: #6137

Big offenders, we should make more aspects of ember pay as you go. For example:
- bower (accounts for like 20-30% of the dep size)

Another effort is to:
- tools/techniques to identify
- tests to prevent regressions cc @hjdivad 
## List of top-level deps + cost to us in # files:
- [x] broccoli-asset-rev 896: https://github.com/rickharrison/broccoli-asset-rev/issues/101
- [ ] ember-ajax 30
- [ ] ember-cli  27247 https://github.com/ember-cli/ember-cli/issues/6149
- [x] ember-cli-app-version 178  EmberSherpa/ember-cli-app-version#28
- [x] ember-cli-babel 5430 https://github.com/babel/ember-cli-babel/pull/83
- [ ] ember-cli-dependency-checker 98
- [x] ember-cli-htmlbars 1091 https://github.com/ember-cli/ember-cli-htmlbars/pull/87
- [ ] ember-cli-htmlbars-inline-precompile 21
- [ ] ember-cli-inject-live-reload 6
- [x] ember-cli-jshint 2063
- [ ] ember-cli-qunit 6167
- [ ] ember-cli-release 558
- [x] ember-cli-sri 515
- [ ] ember-cli-test-loader 8
- [x] ember-cli-uglify 1131 https://github.com/ember-cli/ember-cli-uglify/issues/11
- [ ] ember-data 6660 https://github.com/emberjs/data/issues/4492
- [ ] ember-export-application-global  15
- [ ] ember-inflector  27
- [ ] ember-load-initializers 11
- [ ] ember-resolver 83
- [x] ember-welcome-page  2288 https://github.com/ember-cli/ember-welcome-page/issues/33
- [ ] loader.js 36

discovered via npm v2 (although not totally accurate re: dupes, it helps pin-point costly roots):

```
for x in (ls) do
  echo $x
  find $x | wc -l
[...]

PR #6195, opened 2016-08-18
Add .travis and .github folders to npmignore

knownasilya:
These are added to all the different versions of ember-cli in every addon you install, so it adds up.

Should we also ignore the `/assets` folder?

See #6147

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,4 +4,6 @@
 /docs
 /tests/!(helpers)
 /tmp
+/.travis
+/.github
 /*.*


kellyselden:
LGTM

> Should we also ignore the /assets folder?

that folder is already ignored in the file.

.npmignore
```diff
@@ -4,4 +4,6 @@
 /docs
 /tests/!(helpers)
 /tmp
+/.travis
+/.github
```
Turbo87: isn't that covered already by the `/*.*` glob below?

kellyselden: Ah you are probably right.

knownasilya:
Yes, good catch..

PR #6198, opened 2016-08-19
display cleanup progress.

stefanpenner:
Users otherwise may assume something is frozen, and ctrl-c early


chadhietala:
@homu r+

stefanpenner:
i'll couple this with: https://github.com/broccolijs/broccoli-plugin/issues/16 to mitigate the "issue" where people are killing the process before cleanup of `tmp` is complete.

PR #6205, opened 2016-08-24
Revert "chore(package): update diff to version 2.2.2"

nathanhammond:
This reverts commit a68fe1c35e744ec1489b8be4c44cf545f54a0d3d.

`diff ^2.0` no longer supports Node 0.10 which means we can't use it until LTS support for 0.10 is dropped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "configstore": "^2.0.0",
     "core-object": "^2.0.2",
     "debug": "^2.1.3",
-    "diff": "^2.2.2",
+    "diff": "^1.3.1",
     "ember-cli-broccoli": "0.17.2",
     "ember-cli-get-component-path-option": "^1.0.0",
     "ember-cli-is-package-missing": "^1.0.0",


nathanhammond:
@homu r+

Turbo87:
@nathanhammond we should probably backport this onto the `release` branch or at least `beta`, right?

nathanhammond:
It has the interesting behavior of only applying when you upgrade to a version so I think we can get away with just porting it to beta and telling 0.10 users to skip the 2.7 upgrade. 2.8 will be the last version to support Node 0.10 and on Oct. 1 we can throw a party.

Turbo87:
GitHub claims that https://github.com/ember-cli/ember-cli/commit/a68fe1c35e744ec1489b8be4c44cf545f54a0d3d was introduced in 2.4.3, which makes me think that in theory we would even need to port this back to a 2.4.x branch if we wanted to. I personally don't care since I'm using Node 6, but if someone wants to do it...

nathanhammond:
I _really_ don't want to do a bunch of releases to port it back considering we've not heard a single request for it and that we can direct them to: `cd node_modules/ember-cli && npm install diff@1.3.2`

2.8 will work out of the box which I feel like meets our social contract of having a stable release that works on Node 0.10.

Note that for ease of backporting to previous branches we don't get to land (for example) `let` in master until Oct. 1 when we officially don't support it.

PR #6086, opened 2016-07-16
Use heimdalljs for structured instrumentation & structured logging

hjdivad:
- [x] plug leak
- [x] perf docs
- [x] release heimdall 0.1.4 (see https://github.com/heimdalljs/heimdalljs-lib/pull/3)
- release deps & related
  - [x] [broccoli-merge-trees](https://github.com/broccolijs/broccoli-merge-trees/pull/46)
  - [x] [ember-cli-broccoli-sane-watcher](https://github.com/ember-cli/broccoli-sane-watcher/pull/5)
  - [x] [broccoli-funnel](https://github.com/broccolijs/broccoli-funnel/pull/73)
  - [x] [broccoli-persistent-filter](https://github.com/stefanpenner/broccoli-persistent-filter/pull/71)
  - [x] [heimdalljs-fs-monitor](https://github.com/stefanpenner/heimdall-fs-monitor/pull/1)
  - [x] [ember-cli-broccoli](https://github.com/ember-cli/broccoli/pull/6)
  - [x] [broccoli-slow-trees](https://github.com/rwjblue/broccoli-slow-trees/pull/7)
  - [x] [heimdalljs-logger](https://github.com/heimdalljs/heimdalljs-logger/pull/2)
- [ ] run tests after heimdall 0.1.4 released

cc @stefanpenner

PR #6206, opened 2016-08-24
Remove debug from package.json

marpo60:
We no longer need `debug`, we are using `heimdalljs-logger`

Related: https://github.com/ember-cli/ember-cli/pull/6086


stefanpenner:
good catch, thanks!

PR #6208, opened 2016-08-25
Replace "ember-cli-broccoli" with "broccoli-{brocfile-loader, builder, middleware}"

Turbo87:
I'm pretty sure we could get by without `brocfile-loader` and its subdependency tree, but I've included it for now to keep the diff small. This is a potential future optimization though.

/cc @stefanpenner @hjdivad @chadhietala


stefanpenner:
👍 

@homu r+

cc @joliss we have split up the various broccoli deps, will gladly transfer the sub-parts to the broccoli org.

joliss:
I like the idea, but I don't really like the timing.

The Broccoli 1.0 master branch that has diverged significantly, and I have some further changes sitting in my clone. I'd like to finish that up, get Broccoli 1.0 released, and then do the split again.

In particular, I'd like to caution against making changes to any of the broccoli-{brocfile-loader, builder, middleware} repos -- getting those changes merged back will be messy and take much longer than splitting Broccoli 1.0 into packages.

stefanpenner:
> In particular, I'd like to caution against making changes to any of the broccoli-{brocfile-loader, builder, middleware} repos -- getting those changes merged back will be messy and take much longer than splitting Broccoli 1.0 into packages.

It's to late, those branches have diverged nearly year ago. We simply could not wait that long for our changes to land. We are aware this would cause grief upgrading but given limited time, we made the call. We will gladly help work to re-unite.

PR #6224, opened 2016-09-02
consolidate package.json normalization

kellyselden:
I noticed when generating an in-repo addon, the terminating new line was getting removed from `package.json`. This bug was fixed in addon generation, so the code was consolidated so everyone gets the new line. Then I added tests.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -8,6 +8,7 @@ var SilentError = require('silent-error');
 var date        = new Date();
 
 var normalizeEntityName = require('ember-cli-normalize-entity-name');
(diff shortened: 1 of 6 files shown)


nathanhammond:
> "Ugh, this is annoying." /@kellyselden goes and fixes it.

Awesome. My only comment at first pass is that this isn't limited to `package.json` ... it's really any JSON file. So let's name it something different?

kellyselden:
@nathanhammond How about `stringifyAndNormalize`?

nathanhammond:
`stringifyAndNormalize`: 👍

kellyselden:
@nathanhammond: updated.

nathanhammond:
@homu r+

PR #6227, opened 2016-09-02
add tests for alphabetize-object-keys

kellyselden:
I was extracting it out to use elsewhere, then it turns out I didn't need it after all.

End result is better factored/tested code!


nathanhammond:
This looks good; thanks!

@homu r+

PR #6228, opened 2016-09-02
in-repo-addon: sort additions to ember-addon/paths

kellyselden:
We should probably alphabetize additions to `ember-addon/paths` like we do to `dependencies` [here](https://github.com/ember-cli/ember-cli/blob/79a4b40a1e7c0c9cb356fc2bcf3f0348a0191eb3/blueprints/addon/index.js#L46). For consistency, and also deterministic is always better for git diffs.

### DIFF
--- a/blueprints/in-repo-addon/index.js
+++ b/blueprints/in-repo-addon/index.js
@@ -38,6 +38,7 @@ module.exports = {
     if (isInstall) {
       if (paths.indexOf(newPath) === -1) {
         paths.push(newPath);
+        contents['ember-addon']['paths'] = paths.sort();
       }
     } else {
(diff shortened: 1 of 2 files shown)


nathanhammond:
So this actually has consequences on the order in which addons run because of how we walk things. Namely, you could avoid all of `before` and `after` from the DAG map if you ordered things properly in your `package.json`.

As a result this _could_ be considered a breaking change.

Since this is the default behavior for full addons I believe it's worthwhile to be consistent at risk of surprising a few people using in-repo-addons. I suspect that the people using in-repo-addons will be comfortable specifying their DAG map and therefore consider the risk of this to be low enough that I'm going to land it.

@homu r+

PR #6229, opened 2016-09-02
include global npm version

kellyselden:
This would be useful for users that didn't use `ember install` or those that cloned then `npm install`ed. It would be nice to include this in code, but in code, every use of npm resolves to the locally installed 2.x version.


nathanhammond:
Yeah, and a large number of our filed bugs are npm-related.

@homu r+

PR #6231, opened 2016-09-05
suggest testing addons against LTS

kellyselden:
probably a good idea to default testing addons against LTS versions.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -12,6 +12,7 @@ cache:
 env:
   - EMBER_TRY_SCENARIO=default
   - EMBER_TRY_SCENARIO=ember-1.13
+  - EMBER_TRY_SCENARIO=ember-lts-2.4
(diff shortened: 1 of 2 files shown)


rwjblue:
Seems good to me. When 2.8 becomes an LTS, we can drop 1.13 and keep 2.4 and 2.8.

kellyselden:
@rwjblue Or keep 1.x until a 3.x comes out? or is 2.x LTS an indefinite thing?

nathanhammond:
This is a reasonable change for now; 2.8 doesn't go LTS for another release cycle at which point we can add it in and drop 1.13. Also, it's covered by `default` and `release` for newly-init'd addons.

@kellyselden Can you target this at the `beta` branch? Otherwise this won't land in time for the above to be true.

kellyselden:
@nathanhammond retargeted.

kellyselden:
@homu r+

Issue #6110, opened 2016-07-26
ember addon test-addon -> EACCES: permission denied, open ...package.json


rfbarber:
I receive the following error when attempting to create a new addon:

```
EACCES: permission denied, open '/usr/local/lib/node_modules/ember-cli/blueprints/addon/files/package.json'
Error: EACCES: permission denied, open '/usr/local/lib/node_modules/ember-cli/blueprints/addon/files/package.json'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.writeFileSync (fs.js:1156:15)
    at CoreObject.writeContentsToFile (/usr/local/lib/node_modules/ember-cli/blueprints/addon/index.js:170:6)
    at CoreObject.module.exports.generatePackageJson (/usr/local/lib/node_modules/ember-cli/blueprints/addon/index.js:47:25)
    at CoreObject.module.exports.files (/usr/local/lib/node_modules/ember-cli/blueprints/addon/index.js:96:10)
    at CoreObject.extend.install (/usr/local/lib/node_modules/ember-cli/lib/models/blueprint.js:433:43)
    at CoreObject.module.exports.Task.extend.run (/usr/local/lib/node_modules/ember-cli/lib/tasks/install-blueprint.js:56:24)
    at CoreObject.module.exports.Command.extend.run (/usr/local/lib/node_modules/ember-cli/lib/commands/init.js:102:29)
    at /usr/local/lib/node_modules/ember-cli/lib/commands/new.js:91:12
```

package.json does not exist in the blueprints/addon/files directory:

```
rfb@slate:~/Projects$ ls -la /usr/local/lib/node_modules/ember-cli/blueprints/addon/files/
total 44
drwxr-sr-x 5 nobody rfb 4096 Jul 26 13:03 .
drwxr-sr-x 3 nobody rfb 4096 Jul 26 13:03 ..
drwxr-sr-x 2 nobody rfb 4096 Jul 14 20:37 addon
drwxr-sr-x 2 nobody rfb 4096 Jul 26 13:03 addon-config
drwxr-sr-x 2 nobody rfb 4096 Jul 14 20:37 app
-rw-r--r-- 1 nobody rfb  527 Jul 14 20:38 ember-cli-build.js
-rw-r--r-- 1 nobody rfb   96 Jul 14 20:37 index.js
-rw-r--r-- 1 nobody rfb 1073 Jul 14 20:37 LICENSE.md
[...]

PR #6234, opened 2016-09-06
Display helpful message for write permissions denied exception

kamilogorek:
It's not necessarily the best solution that we can have, but it'll do the job for now and clarify the issue for some developers.

Fixes https://github.com/ember-cli/ember-cli/issues/6110

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -161,7 +161,23 @@ function alphabetizeDependencies(contents) {
 }
 
 function writeContentsToFile(contents, fileName) {
-  var packagePath = path.join(this.path, 'files', fileName);
+  var dirPath = path.join(this.path, 'files');
+  var packagePath = path.join(dirPath, fileName);
 
-  fs.writeFileSync(packagePath, stringifyAndNormalize(contents));
+  try {
+    fs.writeFileSync(packagePath, stringifyAndNormalize(contents));
+  } catch (exception) {
+    checkExceptionCode(exception, dirPath);
+  }
+}
+
+function checkExceptionCode(exception, dirPath) {
+  if (exception.code === 'EACCES') {
+    throw new Error(
+      'Directory path ' + dirPath + ' is not writeable.\n' +
+      'Please follow npm guide to resolve this issue: https://docs.npmjs.com/getting-started/fixing-npm-permissions'
+    );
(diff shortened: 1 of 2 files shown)


nathanhammond:
@kamilogorek This appears to have failed legitimately on Windows.

In general I'm in favor of providing this feedback to users but I'm unsure when exactly we should perform this test. @stefanpenner @rwjblue Any thoughts?

kamilogorek:
@nathanhammond that's true, but it's failing in the same way in all recent PRs. An issue is hidden somewhere deeper.

blueprints/addon/index.js
```diff
@@ -161,7 +161,23 @@ function alphabetizeDependencies(contents) {
 }
 
 function writeContentsToFile(contents, fileName) {
-  var packagePath = path.join(this.path, 'files', fileName);
+  var dirPath = path.join(this.path, 'files');
+  var packagePath = path.join(dirPath, fileName);
 
-  fs.writeFileSync(packagePath, stringifyAndNormalize(contents));
+  try {
+    fs.writeFileSync(packagePath, stringifyAndNormalize(contents));
```
stefanpenner: @rwjblue is this that file that we accidentally write too in peoples `node_modules` ?

rwjblue: Yes, it is that file that we have discussed.  Yes, we should absolutely fix it.

stefanpenner: @kamilogorek although you PR improves the error message, I suspect we can get rid of the problem in most cases if we don't incorrectly write to this file in some scenarios.

kamilogorek: Most definitely. Do you think those files should be written directly into generated folder or stored in an os temp folder first and then moved after being generated?

kamilogorek:
@Turbo87 API for this module has been completely rewritten and I'd need to get familiar with it from scratch. Content creating and writing is in a different place now, so it cannot be rebased.

Turbo87:
Sorry about that! Not sure why this PR was never merged last year... 😞 

I guess in that case we should probably close the PR and if you feel like porting it over to the new code we can open another PR for it, okay?

kamilogorek:
Sure, no worries! I'll try to tackle this one in my spare time :)

PR #6235, opened 2016-09-06
remove `default` ember try scenario

### LINKED ISSUE #6233: remove `default` ember-try scenario for addons
I feel like this is the least useful config to test against. It uses the version of ember that you used to run your dummy app. That doesn't have much to do with appropriate ranges for consuming apps. I feel like `ember-release` is often testing the same version. Thoughts?

kellyselden:
Resolves #6233.


nathanhammond:
I don't think CI is generally broken? I am confused as to why this is messing things up. I wonder if it has to do with something we do for caching npm modules during tests?

kellyselden:
@homu r+

PR #6240, opened 2016-09-08
remove unused path utility

kellyselden:


stefanpenner:
👍

nathanhammond:
Can we guarantee that nobody is including this `util` when depending on Ember CLI? Do we care if we take that away?

Turbo87:
@nathanhammond it hasn't been marked as public API so I'm 👍  on removing it if it's unused. if people depends on this then it should be extracted into a separate module, but I haven't seem this being used anywhere so far.

kellyselden:
After some research, it looks like this is left over from the move to `ember-cli-path-utils`. A search on github looks like everyone is using the `ember-cli-path-utils` version, so I deem this good to go.

kellyselden:
@homu r+

PR #6241, opened 2016-09-08
destroy deletes empty folders

kellyselden:
Wanted to get some feedback while I finish the tests.

This has been reported as a bug a few times before, so here's the fix. When a blueprint destroy removes a file, it checks if the dir is now empty and deletes. Then recurses up the tree and stops at `app`.

I see this as a bugfix and not something that needs to go through RFC. Thoughts?


stefanpenner:
I don't believe we should do this. Folder emptiness does not indicate we can safely remove it.

kellyselden:
@stefanpenner Do you know of an example when this is bad?

nathanhammond:
At least recently (don't know if this is still the case) deleting folders out from underneath Broccoli would result in an error. @stefanpenner and @hjdivad can confirm if they believe this is fixed.

Other bad scenarios have to do with addons relying on a certain folder existing and doing `new Funnel('/boom')`.

kellyselden:
@stefanpenner @nathanhammond Is there any compromise on this? Should it go to RFC? Personally I find it annoying whenever I destroy a route or whatever that I have to clean up the empty folders, just from an OCD/cleanliness perspective. It makes the `destroy` command "incomplete" for my goal of getting the state back to before the `generate`.

On the other hand, I do realize the risk of deleting something the user wasn't expecting.

hjdivad:
> At least recently (don't know if this is still the case) deleting folders out from underneath Broccoli would result in an error. @stefanpenner and @hjdivad can confirm if they believe this is fixed.

That should be fixed in `fs-tree-diff@^0.5.0` IIRC

stefanpenner:
> At least recently (don't know if this is still the case) deleting folders out from underneath Broccoli would result in an error. @stefanpenner and @hjdivad can confirm if they believe this is fixed.

ya we fixed this.

If the general consensus here is, blueprint destroy should destroy empty folders then we can move forward.

nathanhammond:
This is on the agenda for Thursday. :smile:

hjdivad:
> If the general consensus here is, blueprint destroy should destroy empty folders then we can move forward.

I agree with @kellyselden on this.  It seems weird for generate/destroy to not leave you back in the state you started from.  I think @kellyselden's suggestion will make the common cases less surprising for people.

Removing all empty dirs up to `app` will have some pathological edge cases, but `.gitkeep` and equivalents seem like a reasonable way to deal with this.

nathanhammond:
- [ ] Add test for checking symlink behavior.

kellyselden:
I ended up adding an additional property to `FileInfo` called `outputBasePath` to help in assembling new paths when traversing up.

kellyselden:
I finished my testing. The CLI doesn't let you put in root paths or '../..', so I think we are covered there. I wrote a symlink test, and it worked just like a normal file preserving it. It didn't seem all that useful a test to me. Thoughts @nathanhammond @stefanpenner ? Maybe I'm missing the scenario where a symlink test would be good?

Other than that, this is ready to go.

lib/utilities/clean-remove.js
```diff
@@ -0,0 +1,50 @@
...
+          });
+        });
+      }, Promise.resolve());
+
+      return linkedPromise;
+    });
+  }).catch(function() {
+    // you tried to destroy a blueprint without first generating it
+    // instead of trying to read dirs that don't exist
+    // swallow error and carry on
```
kellyselden: It's not a a user error, just a `fs.stat` error in this case. This is a supported thing in the CLI, if you destroy something that doesn't exist, the UI pretends to delete it:

```
$ ember d template jfsd/sdfljks/sdfsjl
uninstalling template
  remove app/jfsd/sdfljks/sdfsjl/template.hbs
```

This worked in the past because `fs.remove` is a no-op for non-existent files. Now we are calling `fs.readdir` to get folder contents, so we have to handle the exception somewhere. Instead of failing on the first `fs.readdir` deep in promises and short-circuiting, I return as soon as possible on the failure case using `fs.stat`.

lib/utilities/walk-up-path.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var path = require('path');
+
+function walkUp(thePath) {
```
nathanhammond: Hokay, so:
- Let's move away from recursion, can be done in a while loop.
- Doesn't address leading slashes, only relative paths.

Here's a quick sketch of a loop that does the right thing but in which I didn't actually sort out the operations to make it sensible:

```
function walkUp(thePath) {
  var paths = [];
  var lastPath = thePath;
  var nextPath;

  while (true) {
    nextPath = path.dirname(lastPath);
    if (nextPath === lastPath) {
      paths.pop();
      break;
    } 
    paths.push(nextPath);
    lastPath = nextPath;
  }

  return paths;
}
```

nathanhammond: Added two more loop choices. :P

tests/unit/utilities/walk-up-path-test.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var expect = require('chai').expect;
+var walkUp = require('../../../lib/utilities/walk-up-path');
+
+describe('walk-up-path', function() {
+  it('walks up paths', function() {
+    var paths = walkUp('app/templates/my/test.hbs');
```
nathanhammond: Need to add a test with a leading `/`.

nathanhammond:
Haven't reviewed `cleanRemove` ... will get to it after your next update.

lib/utilities/walk-up-path.js
kellyselden: > Doesn't address leading slashes, only relative paths.

Do we want to worry about that complexity if it's never being called that way? Are you thinking of the future?

lib/utilities/clean-remove.js
hjdivad: Okay so a couple of things.

1. It seems error prone to pretend to the user that we deleted something.  Seems like it would be easy to self troll with typos?  I'm likely missing context on the motivation for this.
2. Swallowing all errors is very often a trap.  Is it feasible to check the `err` `readdir` provides so that we only swallow what's intended? eg rethrow if `err.code !== 'ENOENT'`

lib/utilities/walk-up-path.js
nathanhammond: We're including it into ember-cli and people may try and use our utils, we should account for that edge case when it's relatively simple to handle.

lib/utilities/clean-remove.js
kellyselden: > It seems error prone to pretend to the user that we deleted something. Seems like it would be easy to self troll with typos? I'm likely missing context on the motivation for this.

Agreed. But this is existing functionality so out of scope of this.

> Swallowing all errors is very often a trap. Is it feasible to check the err readdir provides so that we only swallow what's intended? eg rethrow if err.code !== 'ENOENT'

Added.

lib/utilities/walk-up-path.js
kellyselden: removed recursion and handled leading slash

tests/unit/utilities/walk-up-path-test.js
kellyselden: added

lib/utilities/walk-up-path.js
```diff
@@ -0,0 +1,20 @@
...
+  var paths = [];
+
+  var currentPath = thePath;
+  while (true) { // eslint-disable-line no-constant-condition
+    currentPath = path.dirname(currentPath);
+    if (currentPath.match(/^[\./]$/)) {
```
nathanhammond: 1. Save off regex outside of loop.
2. `regex.test(currentPath)`

tests/unit/models/blueprint-test.js
```diff
@@ -32,6 +32,17 @@ var beforeInstallCalled;
 var afterInstallCalled;
 var afterUninstallCalled;
 
+function resetCalled() {
```
nathanhammond: Just `reset`?

tests/unit/utilities/walk-up-path-test.js
```diff
@@ -0,0 +1,26 @@
...
+
+    expect(paths).to.deep.equal([
+      'app/templates/my',
+      'app/templates',
+      'app',
+    ]);
+  });
+
+  it('walks up paths beginning with a slash', function() {
```
nathanhammond: Add a test for `/foo/foo/file.txt` to protect us from the bug I accidentally included in my first bit of feedback.

lib/utilities/clean-remove.js
```diff
@@ -0,0 +1,54 @@
...
+
+function cleanRemove(fileInfo) {
+  // see if file exists to avoid wasting time
+  return stat(fileInfo.outputPath).then(function() {
+    return remove(fileInfo.outputPath).then(function() {
+      var paths = walkUp(fileInfo.displayPath).map(function(thePath) {
+        return path.join(fileInfo.outputBasePath, thePath);
+      });
+
+      var linkedPromise = paths.reduce(function(linkedPromise, thePath) {
```
nathanhammond: Can you name the two `linkedPromise`s something a bit more clear? Maybe `completionPromise` and `chainedPromise`? Also add a note that you're using `reduce` to continually attach these head to tail. (The fact that it is shadowed makes it a bit confusing.)

nathanhammond:
| --- | --- |
| thing | thing2 |
| --- | --- |

lib/utilities/walk-up-path.js
kellyselden: done

tests/unit/models/blueprint-test.js
kellyselden: not sure I agree. I don't want to imply I'm doing more than just resetting the "called" variables.

lib/utilities/clean-remove.js
kellyselden: done

tests/unit/models/blueprint-test.js
nathanhammond: I understand now, this is only a _partial_ reset.

nathanhammond:
@homu r+

PR #6243, opened 2016-09-08
clean up npm output

kellyselden:


kellyselden:
actually this isn't windows friendly.

PR #6244, opened 2016-09-08
clean up npm output

kellyselden:
this version tested on both mac and windows


kellyselden:
ugh my unix tools were getting picked up in windows.

Issue #6190, opened 2016-08-17
Deprecation warnings should output more details as to what library/package is the cause 


RyanNerd:
Enhancement request:

When starting `ember s` deprecation notices are listed. But there isn't much information as to what module is causing the warning. For example:

DEPRECATION: `ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead

There is no indicator as to what component/addon/library is using `silent.js`. I `greped` my directory and you wouldn't believe how many instances of the word "silent" there are. So I am still digging though this archaeological mess of JS files trying to figure out where the problem is.

nathanhammond:
@RyanNerd Unfortunately that deprecation is coming from something that is (by rule) out of date. I'm not sure if we have a useful reference to the calling object to provide a better error message.

@stefanpenner Can you confirm/deny if we can do better with the `silent-error` deprecation warning?

rwjblue:
@nathanhammond - We could include a stack [here](https://github.com/ember-cli/ember-cli/blob/master/lib/errors/silent.js#L8).

stefanpenner:
@rwjblue we could include the entire stack, or maybe we could try only the previews 2 frames or something?

nathanhammond:
@RyanNerd Would you be willing to open a PR to add the stack trace to the file that @rwjblue pointed to?

RyanNerd:
@nathanhammond Things got busy for me. I will create a PR on Saturday. Thanks for your attention to this enhancement request.

RyanNerd:
Just as an aside:

The module throwing the deprecation issue for me is: ember-deploy-s3-index

This is not an easy fix:
The NPM repo for this package indicates a [version of 0.4.0](https://www.npmjs.com/package/ember-deploy-s3-index) . 
However the github [version is 0.5.0](https://github.com/Kerry350/ember-deploy-s3-index/releases/tag/v0.5.0).

Attempting to set the version in package.json to 0.5.0 makes NPM really angry because it is convinced the latest version is 0.4.0.

nathanhammond:
@RyanNerd That just means that the author didn't `npm publish`. Ping them directly!

PR #6245, opened 2016-09-08
Call stack output for the silent.js deprecation notice (Issue 6190)

RyanNerd:
This is a PR for [issue 6190](https://github.com/ember-cli/ember-cli/issues/6190)


lib/errors/silent.js
```diff
@@ -3,9 +3,12 @@
 var SilentError = require('silent-error');
 var deprecate   = require('../utilities/deprecate');
 
+// Get the call stack so we can let the user know what module is using the deprecated function.
+var stack = new Error().stack;
+
 Object.defineProperty(module, 'exports', {
   get: function () {
-    deprecate('`ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead.', true);
+    deprecate("`ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead.  Module: \n" + stack.toString(), true);
```
stefanpenner: i dont think this stack will be from the right place, it will be from line 7 of silent.js, rather then from the caller of `get`

RyanNerd: Here's the output:

```
DEPRECATION: `ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead. Module: 
Error
    at Module.Object.defineProperty.get [as exports] (/home/ryan/RxTrax/rxtrax-n/node_modules/ember-cli/lib/errors/silent.js:9:17)
    at Function.Module._load (module.js:411:16)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/ryan/RxTrax/rxtrax-n/node_modules/ember-deploy-s3-index/lib/s3-adapter.js:5:19)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
```

The ` Object.<anonymous> (/home/ryan/RxTrax/rxtrax-n/node_modules/ember-deploy-s3-index/lib/s3-adapter.js:5:19)` cued me in on what package was causing the deprecation notice.

What may be better is to do is a text replace for the stack output: `Error at` replaced with `Call stack`

nathanhammond: So, the trace isn't from the "right place" but it actually contains the information that we need nonetheless since all of that is still in the same stack.

Is that "good enough" @stefanpenner?

RyanNerd: Is this still an issue?

nathanhammond:
@RyanNerd Let's modify this slightly:

```diff
- Module:
+ The following stack items should include the module which needs to be updated:
```

```javascript
stack = stack.toString().replace("Error\n", "");
```

With these changes made I'll land it. Also, can you confirm if the usage point is always at the stack "index" of `4`? If so it'd be nice to make the stack smaller and more likely to identify the culprit. Alternatively removing things which are inside of `**/module.js` seems like a good way to minimize the stack.

nathanhammond:
(Bonus points for rebasing.)

nathanhammond:
@RyanNerd Do you have time to get back to this?

RyanNerd:
@nathanhammond Sorry things got really busy at my work. I'll see if I can grab some time on Halloween to do this.

PR #6249, opened 2016-09-11
Update to ember-cli-qunit@3.0.1.

rwjblue:
This updates apps to use QUnit 2.0 by default.

/cc @trentmwillis


nathanhammond:
In favor; want 👍 from @trentmwillis and we'll merge.

trentmwillis:
I think this should be fine, been out for a little bit now and haven't had any bugs filed (that I'm aware of). So :+1:

PR #6254, opened 2016-09-13
[BUGFIX] Do not rely on ember-resolver, detect bower package instead

### LINKED ISSUE #6253: Can't add a custom resolver
The following piece of code will fail a build if `ember-resolver` is not in package.json and `ember-resolver.js` is not bower.json:

https://github.com/ember-cli/ember-cli/blob/8c84c6c6188cddb35fb6723f50c9f229439073b3/lib/broccoli/ember-app.js#L328-L333

``` js
  if (this.registry.availablePlugins['ember-resolver']) {
    // if the project is using `ember-resolver` as an addon
    // remove it from `vendorFiles` (the NPM version properly works
    // without `app.import`s)
    delete this.vendorFiles['ember-resolver.js'];
  }
```

I think this should be inverted to check if the bower dep exists and then load the bower dev, else never even add `ember-resolver.js` to `vendorFiles`.

martndemus:
Fixes #6253

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -294,7 +294,8 @@ EmberApp.prototype._initVendorFiles = function() {
   }
 
   var handlebarsVendorFiles;
-  if ('handlebars' in this.project.bowerDependencies()) {
+  var bowerDeps = this.project.bowerDependencies();
+  if ('handlebars' in bowerDeps) {
     handlebarsVendorFiles = {
       development: this.bowerDirectory + '/handlebars/handlebars.js',
       production:  this.bowerDirectory + '/handlebars/handlebars.runtime.js'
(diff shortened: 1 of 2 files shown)


rwjblue:
A few constraints here (cross posting parts from #6253):

1) Today, you can avoid errors *and* avoid using `ember-resolver` addon via this config in your `ember-cli-build.js`. This works intentionally, and should continue to "work" (for compat reasons).

```js
new EmberApp(defaults, {
  vendorFiles: {
    'ember-resolver.js': null
  }
});
```

2) If both `bower.json` and `package.json` contain `ember-resolver` the NPM version *must* win.  

I'm super happy to remove the requirement for the `'ember-resolver.js': null` hack described above in #1. But we need to ensure that #2 is still addressed.

martndemus:
Bump, I have updated this PR and it now also does'nt crash `ember-cli-shims` is missing.

nathanhammond:
@rwjblue this is on you. 😜

martndemus:
@rwjblue Updated!

rwjblue:
Thanks!

@homu r+

Turbo87:
@martndemus it seems that one of the tests in this PR is broken. for details on why CI didn't catch it see #6486.

Issue #6149, opened 2016-08-04
This project appears to introduce 27247 files via itself and its deps


stefanpenner:
[PLAN OF ATTACK](https://github.com/ember-cli/ember-cli/issues/6147#issuecomment-237439212)

big offenders:
--------------

- [ ] bower    5021 we are on an old version... we should likely  (extract as addon, update, fork our own subset)??
- [ ] broccoli-babel-transpiler 3725 https://github.com/babel/broccoli-babel-transpiler/issues/90
- [ ] npm 4230 (we are still on an old version, we should likely work hard to upgrade asap, then investigate)

the rest (long tail but adds up) :
-----

- [x] amd-name-resolver 13 https://github.com/chadhietala/amd-name-resolver/pull/8
- [ ] bower-config 700 
- [ ] bower-endpoint-parser 11
- [x] broccoli-concat 546 https://github.com/ember-cli/broccoli-concat/issues/72
- [ ] broccoli-config-loader 213 https://github.com/ember-cli/broccoli-config-loader/issues/5
- [ ] broccoli-config-replace 408 https://github.com/ember-cli/broccoli-config-replace/issues/6
- [ ] broccoli-funnel 343 https://github.com/broccolijs/broccoli-funnel/issues/72
- [x] broccoli-funnel-reducer 5 https://github.com/ember-cli/broccoli-funnel-reducer/pull/1
- [x] broccoli-merge-trees 325 https://github.com/broccolijs/broccoli-merge-trees/issues/47
- [x] broccoli-sane-watcher 14 https://github.com/broccolijs/broccoli-sane-watcher/issues/33
- [x] broccoli-source 8 https://github.com/broccolijs/broccoli-source/pull/2
- [x] broccoli-viz 13 https://github.com/stefanpenner/broccoli-viz/pull/23
- [x] chalk 38
- [x] clean-base-url 4 https://github.com/ember-cli/clean-base-url/pull/1
- [x] compression 71
- [x] configstore 160
- [ ] core-object 30 https://github.com/ember-cli/core-object/issues/41
- [ ] debug 20
- [ ] diff 38
- [ ] ember-cli-broccoli 778 helped by https://github.com/ember-cli/broccoli/pull/5
- [x] ember-cli-get-component-path-option 5 https://github.com/ember-cli/ember-cli-get-component-path-option/pull/2
- [x] ember-cli-is-package-missing 3
[...]

PR #6271, opened 2016-09-16
Addons: Replace most of .npmignore with a `files` key in the package.json file

Turbo87:
This should hopefully help with things like https://github.com/ember-cli/ember-cli/issues/6149 in the future.

/cc @stefanpenner @rwjblue @locks

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -1,16 +1,2 @@
-/bower_components
 /config/ember-try.js
-/dist
-/tests
-/tmp
 **/.gitkeep
-.bowerrc
-.editorconfig
-.ember-cli
-.gitignore
-.jshintrc
-.watchmanconfig
-.travis.yml
-bower.json
-ember-cli-build.js
-testem.js
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -40,6 +40,16 @@ module.exports = {
     // use `ember-try` as test script in addons by default
     contents.scripts.test = 'ember try:each';
 
+    contents.files = [
+      'addon',
+      'app',
+      'blueprints',
+      'config',
+      'lib',
+      'vendor',
+      'index.js',
+    ];
+
     contents['ember-addon'] = contents['ember-addon'] || {};
     contents['ember-addon'].configPath = 'tests/dummy/config';
 
--- a/tests/unit/blueprints/addon-test.js
+++ b/tests/unit/blueprints/addon-test.js
@@ -98,6 +98,15 @@ describe('blueprint - addon', function() {
   "keywords": [\n\
     "ember-addon"\n\
   ],\n\
+  "files": [\n\
+    "addon",\n\
+    "app",\n\
+    "blueprints",\n\
+    "config",\n\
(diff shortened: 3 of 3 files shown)


Turbo87:
Where I'm not sure about is the `config` folder. I've had a look at a few plugins listed at EmberObserver but haven't found any plugin that would need its `config` folder in a user project so I left it out.

stefanpenner:
> Where I'm not sure about is the config folder. I've had a look at a few plugins listed at EmberObserver but haven't found any plugin that would need its config folder in a user project so I left it out.

yes this sounds correct.

stefanpenner:
> This should hopefully help with things like #6149 in the future.

👍

rwjblue:
> Where I'm not sure about is the config folder. I've had a look at a few plugins listed at EmberObserver but haven't found any plugin that would need its config folder in a user project so I left it out.

The [default implementation](https://github.com/ember-cli/ember-cli/blob/v2.8.0/lib/models/addon.js#L896-L904) uses `config/` folder.

stefanpenner:
> The default implementation of the config hook in addons uses config/ folder.
> Add your review

how is this merged with the host app?

rwjblue:
> how is this merged with the host app?

Done here: https://github.com/ember-cli/ember-cli/blob/v2.8.0/lib/models/project.js#L217-L229

stefanpenner:
@rwjblue so for all addons (including nested) they are recursively merged or..

kellyselden:
I love where this is going. Would like to see it rebased on #6272 (once that's merged) since I believe they share a lot of changes.

- [ ] rebase on #6272

Turbo87:
Yeah, I'll rebase on whatever gets merged first :wink:

nathanhammond:
@rwjblue is correct, `config` needs to be present.

@stefanpenner `this.addons` is already a flattened list of all of the addons in the application at the point of the `reduce`/`merge` operation.

Turbo87:
okay. I've rebased, added the `config` folder, kept `/config/ember-try.js` in `.npmignore` and adjusted the expected test output.

rwjblue:
> I've rebased, added the config folder, kept /config/ember-try.js in .npmignore 

Hmm. I was under the impression that `files` completely overrides `.npmignore`. Can you confirm via an `npm pack` of a generated addon with these updates, and confirm the files included in the tarball are correct?

Turbo87:
https://docs.npmjs.com/files/package.json#files states: 

> You can also provide a ".npmignore" file in the root of your package or in subdirectories, which will keep files from being included, even if they would be picked up by the files array.

I've just tried it though and the `.npmignore` file doesn't seems to have any effect. Maybe I'm doing something wrong or my NPM is broken, but it doesn't seem to work according to its documentation... Investigating...

Turbo87:
https://github.com/npm/npm/issues/7435

Conclusion: NPM has been broken since the dawn of time and we can't use this approach. The Module Unification RFC will unfortunately make things even worse when `*-test.js` files are included in the `src` folder.

nathanhammond:
`ember publish` coming soon to an addon near you.

Issue #6274, opened 2016-09-16
CI is broken


Turbo87:
It appears that our CI tests are currently broken, most likely due to some dependency update. The error looks roughly like this:

```
not ok 6 PhantomJS 1.9 - pods based templates: the application boots properly with pods based templates with a podModulePrefix set\n    ---\n        actual: >\n            null\n        stack: >\n            TypeError: 'undefined' is not a function (evaluating '_ember['default'].assign({}, _someCoolAppConfigEnvironment['default'].APP, attrs)')\n        message: >\n            beforeEach failed on the application boots properly with pods based templates with a podModulePrefix set: 'undefined' is not a function (evaluating '_ember['default'].assign({}, _someCoolAppConfigEnvironment['default'].APP, attrs)')\n        Log: |\n    ...\n"
```

From what I can tell the error is most likely coming from this line: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/tests/helpers/start-app.js#L9

Turbo87:
This might be related to https://github.com/ember-cli/ember-cli/pull/6193 but interestingly the CI tests for that PR passed back then...

nathanhammond:
Ohai, we tried to do this once before and it blew up: https://github.com/ember-cli/ember-cli/pull/5803

Going to try reverting #6193.

Turbo87:
https://embercommunity.slack.com/archives/-testing/p1474058140001083 seems relevant to this discussion too

nathanhammond:
Want to turn this into a tracking issue for understanding?

PR #6193, opened 2016-08-18
Changed the start-app test helper to use `Ember.assign`.

workmanw:
Some discussion here: ember-cli/ember-new-output#5 . I mistakenly submitted the PR to that repo not knowing it was generated.

PR #6276, opened 2016-09-16
Revert #6193

nathanhammond:
Optimistic change, didn't actually do any testing. RE: https://github.com/ember-cli/ember-cli/issues/6274

### DIFF
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  // use defaults, but you can override
-  let attributes = Ember.assign({}, config.APP, attrs);
+  let attributes = Ember.merge({}, config.APP);
+  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
 
   Ember.run(() => {
     application = Application.create(attributes);


nathanhammond:
@homu r+

Turbo87:
> Appears to work

Unfortunately not...

nathanhammond:
Intermittent failures, not the 100% red of before.

nathanhammond:
Just restarted the job, about to ask homu for a retry which I know will pass.

Turbo87:
did you see my comment in https://github.com/ember-cli/ember-cli/issues/6274? I have a feeling something else might be wrong and this would just hide the symptoms.

Turbo87:
FWIW after reading the backlog I'm fine with this change but I'm still not quite comfortable not knowing what was actually causing those issues.

Turbo87:
@nathanhammond this needs to be backported to the beta branch

nathanhammond:
Backported; we've been broken for so long that master and beta hadn't diverged.

rwwagner90:
@nathanhammond I just threw together a quick polyfill for this, to fix my 1.13 failures. If you would like, I could submit a PR to use it. https://github.com/shipshapecode/ember-assign-polyfill

nathanhammond:
TODO: Confirm if we have checked in an old `bower.json` as a fixture.

rwjblue:
Likely caused by the error fixed in https://github.com/emberjs/ember.js/pull/14366

Turbo87:
makes sense. should we revert the revert?

rwjblue:
Ya, we should give it a whirl.

nathanhammond:
https://github.com/ember-cli/ember-cli/pull/6305

PR #6279, opened 2016-09-18
blueprints/{addon,app}/.travis.yml: Cache the cache folders

Turbo87:
as suggested in https://github.com/ember-cli/ember-cli/pull/5528

(related: https://github.com/mike-north/ember-ci/issues/4)

/cc @stefanpenner @rwjblue


nathanhammond:
This seems like a good change though it may make runs a touch slower because It still has to do file system work.

Trading perf for correctness.

@homu r+

PR #6292, opened 2016-09-23
Closes #6290 - Use bower-config to get bower directory

### LINKED ISSUE #6290: ember-cli does not respect `bower_directory` environment variable
I'm currently in docker hell, and I need to have my npm and bower packages in the directory _above_ my app code. If my app is in `/usr/code/app` I need my bower_components to be in `/usr/code/bower_components`.

Bower lets you specify a `bower_directory` environment variable, and if I set this to `../bower_components` then `bower` works just fine. However `ember-cli` seems to ignore this and _only_ looks inside the `.bowerrc` file, but _only_ in the current directory and not in any parent directories.

Pretty much it would be great if ember-cli could respect the config rules described here: https://bower.io/docs/config/

```
ember-cli: 2.6.3
http_parser: 2.7.0
node: 6.6.0
v8: 5.1.281.83
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2h
os: linux x64
3.10.3

```

orf:
Use `bower-config` to read the correct bower directory, following the bower config spec here: https://bower.io/docs/config/

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -13,6 +13,7 @@ var _               = require('ember-cli-lodash-subset');
 var logger          = require('heimdalljs-logger')('ember-cli:project');
 var nodeModulesPath = require('node-modules-path');
 var versionUtils    = require('../utilities/version-utils');
+var bowerConfig     = require('bower-config');
 var emberCLIVersion = versionUtils.emberCLIVersion;
 
 /**
@@ -73,20 +74,17 @@ function Project(root, pkg, ui, cli) {
   @method setupBowerDirectory
  */
 Project.prototype.setupBowerDirectory = function() {
-  var bowerrcPath = path.join(this.root, '.bowerrc');
+  var config;
+  try {
+    config = bowerConfig.read(this.root);
+  } catch (e) {
+    logger.warn('%s: %s', e.code, e.message);
+    logger.warn('Defaulting to %s/bower_components', this.root);
 
-  logger.info('bowerrc path: %s', bowerrcPath);
-
-  if (existsSync(bowerrcPath)) {
-    try {
-      this.bowerDirectory = fs.readJsonSync(bowerrcPath).directory;
-    } catch (exception) {
-      logger.info('failed to parse bowerc: %s', exception);
-      this.bowerDirectory = null;
-    }
+    config = {directory: 'bower_components'};
   }
 
-  this.bowerDirectory = this.bowerDirectory || 'bower_components';
+  this.bowerDirectory = config.directory;
   logger.info('bowerDirectory: %s', this.bowerDirectory);
 };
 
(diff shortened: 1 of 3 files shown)


Turbo87:
how does this relate to https://github.com/ember-cli/ember-cli/pull/5892 and does it have the same issues?

orf:
I don't think it has the same issues, all the tests passed unlike #5892. I don't know enough about ember-cli's internals to answer confidently, but this change shouldn't matter if `bower.json` is present or not. It just finds the `.bowerrc` file and handles environment variable overrides.

lib/models/project.js
```diff
@@ -73,20 +74,14 @@ function Project(root, pkg, ui, cli) {
...
-      this.bowerDirectory = fs.readJsonSync(bowerrcPath).directory;
-    } catch (exception) {
-      logger.info('failed to parse bowerc: %s', exception);
-      this.bowerDirectory = null;
-    }
+  var config;
+  try {
+    config = bowerConfig.read(this.root);
+  } catch (e) {
+    config = {directory: 'bower_components'};
```
stefanpenner: i would prefer to only catch non-exceptional errors her. For example, if the config file is malformed, we should error in a way that is actionable. Falling back on error, will lead to tricky heisenbugs.

orf: I was merely following the existing code, that falls back to `bower_components` if the `.bowerrc` is malformed?

How could we handle this otherwise?

nathanhammond:
> by default `bower-config` walks up the directory tree to find a `.bowerrc` file, so in nested projects it might find a parent projects `.bowerrc` and use a nested `bower_components` directory.

Ember CLI doesn't allow you to have nested `bower_components` directories. Addons installed in an application won't have their own, we have exactly one.

orf:
Oh right, well in this case this MR should be fine unless there are any code problems?

lib/models/project.js
nathanhammond: @stefanpenner's comment is simply to make sure that we investigate the error and provide a decent error message based on whatever `e` is instead of simply jumping to a default behavior that may be wrong.

nathanhammond:
Note, without looking at code: I think you want to start the search for `.bowerrc` files in the project's root as opposed to anywhere else down tree.

orf:
So it currently starts the search in the project root, and then if it doesn't find it it continues down the tree.

lib/models/project.js
orf: Ok, I added this just now. It logs two warnings, one with the error code and message and the other saying that it is defaulting to `this.root/bower_components`

locks:
What's blocking this?

orf:
I'm not sure, I think I remember someone mentioning that they didn't want to end up having a million environment variables. I think this isn't a great argument against this though, and should be merged. It's incredibly useful in some specific cases.

orf:
I can rebase this and fix some of the conflicts if this has any hope of being merged, but if not please let me know and close it :+1:

locks:
Given the general move off of bower and the uncertainty of the future of the package manager, I believe that we don't want to bring in this extra dependency to the project.
Thank you for your work @orf, I'm sorry it took this long to make a decision either way :/

Issue #6118, opened 2016-07-28
`ember asset-sizes` should include relative path


rwjblue:
For example, when using `ember-cli-fastboot` there are multiple files starting with `vendor` you don't know which is in `fastboot/vendor` vs `dist/assets/vendor`.

I believe that this https://github.com/ember-cli/ember-cli/blob/master/lib/models/asset-size-printer.js#L50-L55 needs to be changed to add the relative path to the `name` (or add another property for it).

Example output:

```
➜  web-client git:(use-beta-builds) ✗ ember b -prod
Built project successfully. Stored in "dist/".
File sizes:
 - web-client-0f1637eedf4932d7ca1476fa3acb8632.js: 24.94 KB (6.63 KB gzipped)
 - web-client-cfe3e1c6f474b9086dbf1de3ffe86109.css: 816 B (412 B gzipped)
 - vendor-7082175b7f710bee72cbd5e0e828fe5b.js: 556.93 KB (154.33 KB gzipped)
 - vendor-d41d8cd98f00b204e9800998ecf8427e.css: 0 B
 - web-client-65ca4e49508d84ecabc59211abc15ad5.js: 25.82 KB (6.9 KB gzipped)
 - vendor-7d3f9a190a5cb696e8441304a5e57384.js: 509.46 KB (142.02 KB gzipped)
```

rwjblue:
@kiwiupover - You recently tackled some changes in the asset size printer, you think you may have some time to take a look at this?

PR #6296, opened 2016-09-24
Include relative path on ember asset-sizes

josemarluedke:
This implements #6118. 

The result looks like this:

```
$ ember b -prod
Built project successfully. Stored in "dist/".
File sizes:
 - dist/assets/test-app-05b6e5ef3de6d96abb221d931be3a037.js: 9.58 KB (2.8 KB gzipped)
 - dist/assets/test-app-d41d8cd98f00b204e9800998ecf8427e.css: 0 B
 - dist/assets/vendor-d41d8cd98f00b204e9800998ecf8427e.css: 0 B
 - dist/assets/vendor-d55002550e35ed1beba09da66fc57f99.js: 656.38 KB (175.79 KB gzipped)
 - dist/fastboot/test-app-1703e9c3df242d64c82c9962e8ce8d33.js: 10.32 KB (3.01 KB gzipped)
 - dist/fastboot/vendor-d55002550e35ed1beba09da66fc57f99.js: 656.38 KB (175.79 KB gzipped)
```

### DIFF
--- a/lib/models/asset-size-printer.js
+++ b/lib/models/asset-size-printer.js
@@ -44,11 +44,10 @@ AssetPrinterSize.prototype.makeAssetSizesObject = function () {
     })
     // Print human-readable file sizes (including gzipped)
     .map(function (file) {
-      var filename = path.basename(file);
       var contentsBuffer = fs.readFileSync(file);
       return gzip(contentsBuffer).then(function (buffer) {
         return {
-          name: filename,
+          name: file,
           size: contentsBuffer.length,
           gzipSize: buffer.length,
           showGzipped: contentsBuffer.length > 0


lib/models/asset-size-printer.js
```diff
@@ -39,16 +38,14 @@ AssetPrinterSize.prototype.makeAssetSizesObject = function () {
   var assets = files
     // Skip test files
     .filter(function (file) {
-      var filename = path.basename(file);
-      return !testFileRegex.test(filename);
+      return !testFileRegex.test(file);
```
nathanhammond: I don't believe that this section needs to change? We really want that to compare at the file name level so that we don't get false positives on path matches.

josemarluedke:
@nathanhammond PR modified. Thanks for reviewing it!

lib/models/asset-size-printer.js
josemarluedke: Yes, this is not necessary. I figured we could remove the `path`, but your argument is compelling. Changed.

chadhietala:
This LGTM

nathanhammond:
@homu r+

nathanhammond:
@josemarluedke Thank you for taking the time to address this for everybody!

PR #6297, opened 2016-09-25
add test to verify running destroy without generate doesn't error

kellyselden:
You can destroy anything without generating first and no error occurs, but it's an untested feature. This fixes that.

This test needs to exist before #6241 lands to prevent a regression.

### DIFF
--- a/tests/unit/models/blueprint-test.js
+++ b/tests/unit/models/blueprint-test.js
@@ -806,6 +806,10 @@ help in detail');
         expect(afterUninstallCalled).to.be.false;
       });
     });
+
+    it('doesn\'t throw when running uninstall without installing first', function() {
+      return blueprint.uninstall(options);
+    });
   });
 
   describe('basic blueprint uninstallation', function() {


tests/acceptance/destroy-test.js
```diff
@@ -245,6 +245,18 @@ describe('Acceptance: ember destroy', function() {
     return assertDestroyAfterGenerate(commandArgs, files);
   });
 
+  it('does\'t error when running destroy without generating first', function() {
+    var commandArgs = ['foo', 'foo'];
```
nathanhammond: `['component', 'foo']`

tests/acceptance/destroy-test.js
```diff
@@ -245,6 +245,18 @@ describe('Acceptance: ember destroy', function() {
     return assertDestroyAfterGenerate(commandArgs, files);
   });
 
+  it('does\'t error when running destroy without generating first', function() {
```
nathanhammond:
```diff
- does\'t
+ doesn\'t
```

kellyselden: shoot...

tests/acceptance/destroy-test.js
kellyselden: fixed

tests/acceptance/destroy-test.js
kellyselden: fixed

nathanhammond:
Optimistically assuming tests will pass now. Given that:

@homu r+

stefanpenner:
can this be unit tested instead? Full end to end seems both slower and marginal value.

nathanhammond:
lol at the test failure. I'm pretty sure that the `component` destroyer shouldn't care if there is a hyphen or not. Can we address that as well?

@stefanpenner I'm unsure how you would test in a non-acceptance scenario. But rather than creating a new test we could piggyback on top of an existing one of these tests.

@kellyselden Thoughts on the latter?

stefanpenner:
> @stefanpenner I'm unsure how you would test in a non-acceptance scenario. But rather than creating a new test we could piggyback on top of an existing one of these tests.

what is the minimal viable test to ensure the destroy command runs? I don't think that requires setting up an entire app.

kellyselden:
@stefanpenner Thanks for catching and reminding me about acceptance test costs. I created a unit (integration) test version of the same test, and verified it still tests the same scenario.

nathanhammond:
@homu r+

PR #6298, opened 2016-09-26
Closes #6290 - Allow ember's bower directory to be overridden by EMBER_BOWER_DIRECTORY.

### LINKED ISSUE #6290: ember-cli does not respect `bower_directory` environment variable

orf:
This is an alternative MR to #6292, it's a bit simpler and might be better to include.

The problem with #6292 is that by default `bower-config` walks up the directory tree to find a `.bowerrc` file, so in nested projects it might find a parent projects `.bowerrc` and use a nested `bower_components` directory.

Fixing this might be tricky, so perhaps this is a better idea: just have an `EMBER_BOWER_DIRECTORY` environment variable to override the `bower_components` directory.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -74,6 +74,7 @@ function Project(root, pkg, ui, cli) {
  */
 Project.prototype.setupBowerDirectory = function() {
   var bowerrcPath = path.join(this.root, '.bowerrc');
+  var envDirectory = process.env.EMBER_BOWER_DIRECTORY;
 
   logger.info('bowerrc path: %s', bowerrcPath);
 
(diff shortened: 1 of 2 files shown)


nathanhammond:
We don't want to consistently add in additional flags to work around problems every time we encounter them. That gets us into a situation where the complexity of the application ends up exploding with permutations. I want to continue with the solution from #6292 and figure out how to address the (harder) but still worthwhile problems. Will comment over there.

PR #6301, opened 2016-09-27
[Fixes #6300] consistent concat, regardless of system EOL

### LINKED ISSUE #6300: [Investigate] EOL in concat separator
as per: https://github.com/ember-cli/ember-cli/pull/6081#pullrequestreview-1482054
it is strange that we use EOL not '\n' for builds: https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1272

I believe in a perfect world, builds from win and builds from *nix produce the same outputs.

So open questions:

* is there a reason why we did this?
* is there any reason for us not to change it to `\n` in all cases.

stefanpenner:
@Turbo87 r?


Turbo87:
Seems good to me. Let's get this into canary! 

@homu r+

PR #6307, opened 2016-09-30
Node.js LTS updates.

nathanhammond:
This updates the versions of Node.js we support, improves our messaging to users, and makes the platform checker alerts dynamic (and correct, for the first time ever).

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -98,15 +98,17 @@ CLI.prototype.run = function(environment) {
 
     var PlatformChecker = require('../utilities/platform-checker');
     var platform = new PlatformChecker(process.version);
-    if (!platform.isValid && !this.testing) {
+    var recommendation = ' We recommend that you use the most-recent "Active LTS" version of Node.js.';
+
+    if (!this.testing) {
       if (platform.isDeprecated) {
         this.ui.writeDeprecateLine('Node ' + process.version +
-                                   ' is no longer supported by Ember CLI. Please update to a more recent version of Node');
+                                   ' is no longer supported by Ember CLI.' + recommendation);
       }
 
-      if (platform.isUntested) {
+      if (!platform.isTested) {
         this.ui.writeWarnLine('Node ' + process.version +
-                              ' has currently not been tested against Ember CLI and may result in unexpected behaviour.');
+                              ' is not tested against Ember CLI on your platform.' + recommendation);
       }
     }
 
--- a/lib/utilities/platform-checker.js
+++ b/lib/utilities/platform-checker.js
@@ -2,33 +2,23 @@
 
 var semver = require('semver');
 var logger = require('heimdalljs-logger')('ember-cli:platform-checker:');
-
-var findup = require('find-up');
-var fs = require('fs');
-var yaml = require('js-yaml');
-
-function getConfigFile(path) {
-  var configPath = findup.sync(path, { cwd: __dirname });
-  return fs.readFileSync(configPath, 'utf8');
-}
+var loadConfig = require('./load-config');
 
 var testedEngines;
 if (process.platform === 'win32') {
-  testedEngines = yaml
-    .safeLoad(getConfigFile('appveyor.yml'))
+  testedEngines = loadConfig('appveyor.yml')
     .environment.matrix
     .map(function(element) {
       return element.nodejs_version;
     })
     .join(' || ');
 } else {
-  testedEngines = yaml
-    .safeLoad(getConfigFile('.travis.yml'))
+  testedEngines = loadConfig('.travis.yml')
     .node_js
     .join(' || ');
 }
 
-var supportedEngines = JSON.parse(getConfigFile('package.json')).engines.node;
+var supportedEngines = loadConfig('package.json').engines.node;
 
 module.exports = PlatformChecker;
 function PlatformChecker(version) {
--- a/lib/utilities/platform-checker.js
+++ b/lib/utilities/platform-checker.js
@@ -3,32 +3,59 @@
 var semver = require('semver');
 var logger = require('heimdalljs-logger')('ember-cli:platform-checker:');
 
-var LOWER_RANGE = '0.12.0';
-var UPPER_RANGE = '7.0.0';
+var findup = require('find-up');
+var fs = require('fs');
+var yaml = require('js-yaml');
+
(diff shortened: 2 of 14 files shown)


tests/unit/utilities/platform-checker-test.js
```diff
@@ -4,59 +4,5 @@ var expect = require('chai').expect;
 var PlatformChecker = require('../../../lib/utilities/platform-checker');
 
 describe('platform-checker', function() {
```
Turbo87: now that this is empty... should we remove the `platform-checker` module?

nathanhammond: You beat me to adding the tests back in.

tests/unit/utilities/platform-checker-test.js
```diff
@@ -4,59 +4,22 @@ var expect = require('chai').expect;
...
-  });
-
-  it('should return isValid for Node v6', function() {
-    expect(new PlatformChecker('v6.0.0').isValid).to.be.equal(true);
-    expect(new PlatformChecker('v6.1.0').isValid).to.be.equal(true);
-    expect(new PlatformChecker('v6.99.0').isValid).to.be.equal(true);
+  it('checkIsTested', function() {
+    expect(new PlatformChecker('v0.10.1').checkIsTested('4 || 6')).to.be.equal(false, 'versions not in range are untested');
+    expect(new PlatformChecker('v9.0.0').checkIsTested('4 || 6')).to.be.equal(false, 'versions not in range are untested');
+    expect(new PlatformChecker('v4.5.0').checkIsTested('4 || 6')).to.be.equal(true, 'versions in the range are valid');
```
Turbo87: I'd suggest something like this:

```js
expect(new PlatformChecker('v4.5.0').checkIsTested('4 || 6'), 'v4.5.0 is tested for `4 || 6` constraint').to.be.true;
```

Turbo87: reason is that I consider `equal(true|false)` an antipattern in chai and I should probably add a rule for that to https://github.com/Turbo87/eslint-plugin-chai-expect

lib/utilities/platform-checker.js
```diff
@@ -2,33 +2,50 @@
...
+var testedEngines;
+if (process.platform === 'win32') {
+  testedEngines = loadConfig('appveyor.yml')
+    .environment.matrix
+    .map(function(element) {
+      return element.nodejs_version;
+    })
+    .join(' || ');
+} else {
+  testedEngines = loadConfig('.travis.yml')
```
rwjblue: Do we actually include these files in our published tarballs?  I'm not opposed, but I would be surprised if they were not ✂️ in the recent march to reduce file counts...

package.json
```diff
@@ -138,7 +139,7 @@
     "yuidocjs": "0.10.2"
   },
   "engines": {
-    "node": ">= 0.10.0"
+    "node": "0.12.* || 4.5.* || 6.*"
```
rwjblue: The 4.5.* strikes me as odd.  In theory, there could well be a `4.6` that would be part of the 4.x LTS series.

From https://github.com/nodejs/LTS#lts-plan:

> Note that while it is possible that critical security and bug fixes may lead to semver-major changes landing within an LTS stream, such situations will be rare and will land as semver-minor bumps in the LTS covered version.

Is your thought that these "major" bumps from critical bugfixes are not supported?  It seems like that would be unlikely...

lib/utilities/platform-checker.js
nathanhammond: We do now.

package.json
nathanhammond: Actually I forgot that was a possibility. My goal was to set the lower bound (4.5.0) but I missed that the upper bound could be anything less than `5`.

tests/unit/utilities/platform-checker-test.js
nathanhammond: The reason I dislike that pattern is simply this:
screen shot 2016-09-30 at 7 57 16 am

It's hard to tell what is being asserted as it's pushed _way_ to the right by the description. Don't care a ton, so somebody not you or me can chime in with their thoughts and this will adopt the style of the third person's recommendation.

lib/cli/cli.js
```diff
@@ -98,15 +98,17 @@ CLI.prototype.run = function(environment) {
...
         this.ui.writeDeprecateLine('Node ' + process.version +
-                                   ' is no longer supported by Ember CLI. Please update to a more recent version of Node');
+                                   ' is no longer supported by Ember CLI.' + recommendation);
       }
 
-      if (platform.isUntested) {
+      if (!platform.isTested) {
         this.ui.writeWarnLine('Node ' + process.version +
-                              ' has currently not been tested against Ember CLI and may result in unexpected behaviour.');
+                              ' is not tested against Ember CLI on your platform.' + recommendation);
```
Turbo87: I'm wondering why we differentiate between tested and deprecated here. IMHO the deprecation message based on the engine compat field in `package.json` should be enough.

nathanhammond: We've always had the distinction, but I've made it "useful" now.

Then:
- You receive the "untested" warning if your version is greater than the current hardcoded value.
- You receive the "deprecated" warning if your version was lower than the current hardcoded value.

Now:
- You receive the deprecated warning when your version is no longer supported per the `engines` key.
- You receive the untested warning when your version is not tested on your platform. This catches all unsupported Node versions in addition to the undeprecated but untested future releases.

If we keep both messages this is how we should do it. I like that it encourages people to stick to the happy path. It'll be noisy for a bit, and then it will calm down as people start to coalesce to a smaller set of versions.

rwjblue:
@homu r+

rwjblue:
@homu r+

rwjblue:
@homu retry

kellyselden:
@homu retry

stefanpenner:
looks like node folk, decided to rollback the EOL to the 31st of Oct. As this only landed in canary, it isn't the end of the world for us, as our canary wont land on master until after that date.

As long as they don't bump it back again, we should be fine.

TheAlphaNerd:
@stefanpenner there is zero plan to change the date. Sorry for the confusion

stefanpenner:
@TheAlphaNerd thanks for confirming :)

PR #6309, opened 2016-10-02
Cleanup WatchDetector

### LINKED ISSUE #2642: Watchman should not be looked for unless needed.
It seems that we are initializing `watchman` before any invocation when most do not need watching.

A few examples:

```
% ember version

version: 0.1.3-master-0dc44fb0a8
valid watchman found, version: [3.0.0]
node: 0.10.33
npm: 2.1.10
```

```
% ember build
version: 0.1.3-master-0dc44fb0a8
valid watchman found, version: [3.0.0]
Built project successfully. Stored in "dist/".
```

```
% ember generate component foo-bar
version: 0.1.3-master-0dc44fb0a8
valid watchman found, version: [3.0.0]
installing
  create app/components/foo-bar.js
  create app/templates/components/foo-bar.hbs
installing
  create tests/unit/components/foo-bar-test.js
```

We only use file watching for a two commands: `serve` and `test` (as far as I can tell).  We should start watchman in them instead of all the time.

stefanpenner:
- [x] exec -> execSync (now that we have node 0.12.x)
- [x] only run when actually needed [fixes #2642]
- [x] depends on #6307 (to drop node 0.10.x support)

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -232,12 +232,30 @@ var Command = CoreObject.extend({
 
       var options = commandOptions.options;
 
-      // do stuff to try and provide a good experience when it comes to file watching
-      var watchPreference = detector.findBestWatcherOption(options);
-      this.project._watchmanInfo = watchPreference.watchmanInfo;
-      options.watcher = watchPreference.watcher;
+      if (this.hasOption('watcher')) {
+        // do stuff to try and provide a good experience when it comes to file watching
+        var watchPreference = detector.findBestWatcherOption(options);
+        this.project._watchmanInfo = watchPreference.watchmanInfo;
+        options.watcher = watchPreference.watcher;
+      }
       resolve(this.run(options, commandOptions.args));
     }.bind(this));
+  },
+
+  /**
+    Reports if the given command has a command line option by a given name
+
+    @method hasOption
+    @param {String} name
+    @return {Boolean}
+  */
+  hasOption: function(name) {
+    for (var i = 0; i < this.availableOptions.length; i++) {
+      if (this.availableOptions[i].name === name) {
+        return true;
+      }
+    }
+    return false;
   },
 
   /**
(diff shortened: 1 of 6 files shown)


lib/models/command.js
```diff
@@ -192,54 +192,73 @@ var Command = CoreObject.extend({
...
   },
 
   /**
+    Reports if the given command has a command line option by a given name
+
+    @method hasOption
+    @param {String} name
+    @return {Boolean}
+  */
+  hasOption: function(name) {
```
nathanhammond:
```javascript
return this.availableOptions.some(function(option) {
  return option.name === name;
})
```

stefanpenner: we can do that once we drop node 0.12.x support (december)

nathanhammond: Not in 0.12. Nevermind.

lib/models/watch-detector.js
```diff
@@ -160,32 +155,31 @@ WatchDetector.prototype.findBestWatcherOption = function(options) {
...
+    // * not all platforms that run node, support node.watch
+    // * not all file systems support node watch
+    //
+    var appearsToWork = this.testIfNodeWatcherAppearsToWork();
+    actual = new WatchPreference(appearsToWork ? NODE : POLLING);
+  } else {
+    actual = bestOption;
+  }
+  debug.info('foundBestWatcherOption, preference was: %o, actual: %o', options, actual);
+  if (actual /* if no preference was initial set, don't bother informing the user */ &&
```
nathanhammond: This whole thing is bonkers. Can we create intermediate variables along the way?

stefanpenner: this is pre-existing

stefanpenner:
> Since this will be the only time we revisit this file in the near future I'd like to go ahead and clean up this absurd boolean algebra section.

This PR is changing enough, that can come in a later one.

PR #6319, opened 2016-10-06
Use --save-dev by default when installing addons

### LINKED ISSUE #6318: Installing addons from blueprint fails: "Could not find addon"
When running `addAddonsToProject` in a blueprint:

```
installing ember-cli-zesty-app-base
  identical testem.js
  install addons ember-cli-eslint@^1.7.0, ember-cli-template-lint@^0.4.12, ember-concurrency@^0.7.8, ember-test-selectors@^0.0.3
Install failed. Could not find addon with name: ember-cli-eslint@^1.7.0
```

We have an Ember addon for installing other addons that is currently running into issues. The `afterInstall` hook in the blueprint looks a like this:

```js
  afterInstall: function() {
    var blueprint = this;
    var project = this.project;

    return blueprint.addAddonsToProject({
      packages: [
        { name: 'ember-cli-eslint', target: '^1.7.0' },
        { name: 'ember-cli-template-lint', target: '^0.4.12' },
        { name: 'ember-concurrency', target: '^0.7.8' },
        { name: 'ember-test-selectors', target: '^0.0.3' }
      ]
    }).then(function() {
      return blueprint.addPackagesToProject([
        { name: 'babel-eslint', target: '^6.1.2' },
        { name: 'ember-cli-deploy', target: '^0.6.4' },
      ])
    })
```

I can reproduce the issue with [this pared down addon](https://github.com/binhums/bug-add-addons-to-project).

This looks similar to #5880, but I believe it is distinct.

Verified against master, I have a potential fix incoming.

---

```
ember-cli: 2.8.0
http_parser: 2.5.2
node: 4.4.5
v8: 4.5.103.35
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 46
openssl: 1.0.2h
os: darwin x64
2.15.5
```

binhums:
Currently, installing an addon with `this.addAddonsToProject` in a blueprint will fail. See #6318 for details.

Having stepped through the code it looks like the issue is that by default npm is not run with the `save-dev` flag. The consequence is that when ember-cli looks for the installed addon it can't find it.

This PR makes npm be called with a `--save-dev` argument if no specific options are specified. Namely by making `blueprintOptions` default to `{ saveDev: true}`.

My understanding is that the following happens:

1. `this.addAddonsToProject` is called with a list of addons (say in the `afterInstall` hook).
2. These are passed to the `addon-install` task.
3. This triggers `npm install` with the list of packages without `--save-dev`.
4. The addon is not added to `packages.json`
5. The addons are reloaded
6. Ember-cli tries to run the default blueprints for the newly installed addons.
7. Ember tries to find the addon in `project.addons` but it isn't there.
8. Error message, install failed.

I'm not 100% convinced this is the best place to set a default, we may equivalently want to set this up in the `addon-install` task.

I've added a unit test, we could potentially add an acceptance test here too.

Fixes #6318

### DIFF
--- a/tests/unit/tasks/addon-install-test.js
+++ b/tests/unit/tasks/addon-install-test.js
@@ -1,0 +1,127 @@
+'use strict';
+
+var AddonInstallTask = require('../../../lib/tasks/addon-install');
+var MockProject      = require('../../helpers/mock-project');
+var expect           = require('chai').expect;
+var CoreObject       = require('core-object');
+var Promise          = require('rsvp').Promise;
+
+describe('addon install task', function() {
+  var ui;
+  var project;
+
+  beforeEach(function() {
+    ui = {
+      startProgress: function() {
+
+      }
+    };
+  });
+
+  afterEach(function() {
+    // ui.stopProgress();
+    ui = undefined;
+    project = undefined;
+  });
+
+  describe('when no save flag specified in blueprintOptions', function() {
+    it('calls npm install with --save-dev as a default', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.not.equal(true);
+          expect(options['save-dev']).to.equal(true);
+          return new Promise(function(resolve, reject) {
+            resolve();
+          });
+        }
+      });
+
+      var addonInstallTask = new AddonInstallTask({
+        ui: ui,
+        project: project,
+        NpmInstallTask: mockNpmInstallTask
+      });
+
+      addonInstallTask.run({});
+    });
+  });
+
+  describe('when save flag specified in blueprintOptions', function() {
+    it('calls npm install with --save', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.equal(true);
+          expect(options['save-dev']).to.not.equal(true);
+          return new Promise(function(resolve, reject) {
+            resolve();
+          });
+        }
+      });
+
+      var addonInstallTask = new AddonInstallTask({
+        ui: ui,
+        project: project,
+        NpmInstallTask: mockNpmInstallTask
+      });
+
+      addonInstallTask.run({
+        blueprintOptions: {
+          save: true
+        }
+      });
+    });
+  });
+
+  describe('when saveDev flag specified in blueprintOptions', function() {
+    it('calls npm install with --save-dev', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.not.equal(true);
+          expect(options['save-dev']).to.equal(true);
+          return new Promise(function(resolve, reject) {
+            resolve();
+          });
+        }
+      });
+
+      var addonInstallTask = new AddonInstallTask({
+        ui: ui,
+        project: project,
+        NpmInstallTask: mockNpmInstallTask
+      });
+
+      addonInstallTask.run({
+        blueprintOptions: {
+          saveDev: true
+        }
+      });
+    });
+  });
+
+  describe('when both save and saveDev flag specified in blueprintOptions', function() {
+    it('calls npm install with --save', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.equal(true);
+          expect(options['save-dev']).to.not.equal(true);
+          return new Promise(function(resolve, reject) {
+            resolve();
+          });
+        }
+      });
+
+      var addonInstallTask = new AddonInstallTask({
+        ui: ui,
+        project: project,
+        NpmInstallTask: mockNpmInstallTask
+      });
+
+      addonInstallTask.run({
+        blueprintOptions: {
+          save: true,
+          saveDev: true
+        }
+      });
+    });
+  });
+});
--- a/tests/unit/tasks/addon-install-test.js
+++ b/tests/unit/tasks/addon-install-test.js
@@ -26,20 +26,18 @@ describe('addon install task', function() {
 
   describe('when no save flag specified in blueprintOptions', function() {
     it('calls npm install with --save-dev as a default', function() {
-      var mockNpmInstallTask = CoreObject.extend({
+      var MockNpmInstallTask = CoreObject.extend({
         run: function(options) {
-          expect(options.save).to.not.equal(true);
-          expect(options['save-dev']).to.equal(true);
-          return new Promise(function(resolve, reject) {
-            resolve();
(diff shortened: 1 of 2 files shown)


lib/models/blueprint.js
```diff
@@ -1061,10 +1061,13 @@ var Blueprint = CoreObject.extend({
...
   */
   addAddonsToProject: function(options) {
+    var defaultBlueprintOptions = {
+      saveDev: true
+    };
     var taskOptions = {
       packages: [],
       extraArgs: options.extraArgs || [],
-      blueprintOptions: options.blueprintOptions || {}
+      blueprintOptions: options.blueprintOptions || defaultBlueprintOptions
```
nathanhammond: This should be a `merge`.

tests/unit/models/blueprint-test.js
```diff
@@ -1528,6 +1528,27 @@ help in detail');
       expect(bluOpts).to.equal('-foo');
     });
 
+    it('sets blueprintOptions to saveDev by default', function() {
```
nathanhammond: Can we also add a test that checks the result of an actual `npm install` from a blueprint to see if the addon appears in `package.json`?

binhums: We can indeed. I'll see what I can throw together.

nathanhammond:
The default behavior for this should be described here:
https://github.com/ember-cli/ember-cli/pull/6017#issuecomment-229852012

That it isn't working implies that we broke something. If it works on the command line but not for `addAddonsToProject` can you identify if there is a more fundamental fix for this problem?

binhums:
Both the command line and `addAddonsToProject` use the `addon-install` task:

https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/addon-install.js#L20

so yeah we could definitely combine the common behavior in this task.

binhums:
Really the logic is `tasks.addon-install` always calls npm with either `--save` or `--save-dev` but never both. It uses `--save-dev` by default.

nathanhammond:
So why is it that it's not `save`ing when invoked via `addAddonsToProject` if they both use the same task? (I've not traced this yet.)

I'm a bit surprised that this is a bug based upon my review–I believe that the default behavior from the task should kick in.

xcambar:
The `install` commands default to `save-dev`, (https://github.com/ember-cli/ember-cli/blob/master/lib/commands/install.js#L13-L16) and these values are passed to the task.

The task itself has no default (https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/addon-install.js#L40-L41), which defaults to neither `save` nor `save-dev`.

And it does not save when using a blueprint just because there's no such option passed to the task (https://github.com/ember-cli/ember-cli/blob/9648d675500757160f07ce77bc9817e0e7a6a356/lib/models/blueprint.js#L1096).


Possible outcomes:

* Offer a default option to the install talk
* Pass the relevant default from the blueprint to the task

binhums:
This would definitely be an easy one to miss, especially since there are no tests for the install addon task.

I've decided to specify the sensible default in the [task](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/addon-install.js#L40-L41) that is invoked by both the command and the blueprint method.

In the process I've added a test for the task.

nathanhammond:
This appears to be the correct solution and the code looks correct. I'm unfamiliar with this portion of the test harness and don't know what should be expected to appear. The tests cover the correct cases.

@Turbo87 Can you confirm the tests look good, and if so, ship?

tests/unit/tasks/addon-install-test.js
```diff
@@ -0,0 +1,127 @@
...
+    project = undefined;
+  });
+
+  describe('when no save flag specified in blueprintOptions', function() {
+    it('calls npm install with --save-dev as a default', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.not.equal(true);
+          expect(options['save-dev']).to.equal(true);
+          return new Promise(function(resolve, reject) {
```
Turbo87: this could be simplified to `Promise.resolve()` right?

tests/unit/tasks/addon-install-test.js
```diff
@@ -0,0 +1,127 @@
...
+
+  afterEach(function() {
+    // ui.stopProgress();
+    ui = undefined;
+    project = undefined;
+  });
+
+  describe('when no save flag specified in blueprintOptions', function() {
+    it('calls npm install with --save-dev as a default', function() {
+      var mockNpmInstallTask = CoreObject.extend({
```
Turbo87: since this is essentially a class we might want to call this `MockNpmInstallTask` with a captial `M`

tests/unit/tasks/addon-install-test.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+      });
+
+      var addonInstallTask = new AddonInstallTask({
+        ui: ui,
+        project: project,
+        NpmInstallTask: mockNpmInstallTask
+      });
+
+      addonInstallTask.run({});
```
Turbo87: I'm not convinced that this test will actually fail if things don't work as planned as it doesn't account for async behavior (there is neither a `done` callback, nor is a Promise returned by the test function)

tests/unit/tasks/addon-install-test.js
```diff
@@ -0,0 +1,127 @@
...
+    ui = undefined;
+    project = undefined;
+  });
+
+  describe('when no save flag specified in blueprintOptions', function() {
+    it('calls npm install with --save-dev as a default', function() {
+      var mockNpmInstallTask = CoreObject.extend({
+        run: function(options) {
+          expect(options.save).to.not.equal(true);
+          expect(options['save-dev']).to.equal(true);
```
Turbo87: we can use `expect(...).to.be.true;` here to make the assertion message more explicit

binhums:
Agreed with all feedback. I'll be away from my computer until Tuesday 25 October. Will implement changes then.

If this needs merging with more urgency do feel free to add to my commits.

tests/unit/tasks/addon-install-test.js
binhums: done

tests/unit/tasks/addon-install-test.js
binhums: done

binhums:
Thanks for reviewing @Turbo87, assuming tests pass these should all be addressed now.

nathanhammond:
@binhums This is excellent! Thanks for coming back to it and wrapping it up.

@homu r+

PR #5921, opened 2016-05-22
(WIP) remove ember from bower

stefanpenner:
- [ ] don't rely on hard-coded path on stefs machine
- [ ] unit tests covering this new functionality

PR #6324, opened 2016-10-07
Moves ember dependency to `ember-source` npm package

locks:
Still to do:
- [ ] Unit tests covering this new functionality. (Should also cover ensuring that we have all assets needed published.)
- [x] Properly set up Ember Try.

All other comments in this thread [before this one](https://github.com/ember-cli/ember-cli/pull/6324#issuecomment-255447801) have been addressed.

---

Supersedes https://github.com/ember-cli/ember-cli/pull/5921
- [x] removes `ember` bower dependency
- [x] adds `ember-source` npm dependency
- @nathanhammond 
  - [x] Add `ember-source` to the default addon/app blueprints.
  - [x] Drop `ember-cli-legacy-blueprints` from the default addon/app blueprints.
  - [x] Relatedly we must change `ember-cli-legacy-blueprints`'s `before` definition to be `ember-source`. https://github.com/ember-cli/ember-cli-legacy-blueprints/pull/62
  - [x] Bump up `ember-cli-legacy-blueprints` dependency

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -10,6 +10,11 @@ module.exports = {
         resolutions: {
           'ember': 'lts-2-4'
         }
+      },
+      npm: {
+        devDependencies: {
+          'ember-source': null
+        }
       }
     },
     {
@@ -20,6 +25,11 @@ module.exports = {
         },
         resolutions: {
           'ember': 'lts-2-8'
+        }
+      },
+      npm: {
+        devDependencies: {
+          'ember-source': null
         }
       }
     },
@@ -32,27 +42,26 @@ module.exports = {
         resolutions: {
           'ember': 'release'
         }
+      },
+      npm: {
+        devDependencies: {
+          'ember-source': null
+        }
       }
     },
     {
       name: 'ember-beta',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#beta'
-        },
-        resolutions: {
-          'ember': 'beta'
+      npm: {
+        devDependencies: {
+          'ember-source': 'git://github.com/emberjs/ember.js.git#beta'
         }
       }
     },
     {
       name: 'ember-canary',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#canary'
-        },
-        resolutions: {
-          'ember': 'canary'
+      npm: {
+        devDependencies: {
+          'ember-source': 'git://github.com/emberjs/ember.js.git#master'
         }
       }
     }
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "<%= name %>",
   "dependencies": {
-    "ember-cli-shims": "0.1.3"
   }
 }
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "<%= name %>",
   "dependencies": {
-    "ember": "~2.10.0",
     "ember-cli-shims": "0.1.3"
   }
 }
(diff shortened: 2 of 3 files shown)


locks:
@nathanhammond https://github.com/emberjs/ember.js/pull/14396 is merged

stefanpenner:
We may want to [BUGFIX beta] this.

nathanhammond:
With only one release left in the beta cycle I'm opposed to [BUGFIX beta]. We'll land on master and let it ride the trains (only 8 weeks).

nathanhammond:
@locks This needs to:

- [ ] Add `ember-source` to the default addon/app blueprints.
- [ ] Drop `ember-cli-legacy-blueprints` from the default addon/app blueprints.
- [ ] Relatedly we must change `ember-cli-legacy-blueprints`'s `before` definition to be `ember-source`.

nathanhammond:
I've dropped my blocking of this, I _believe_ we have all of the pieces in place. Kicked over the test suites.

nathanhammond:
@locks Actually, still blocked on a new release of `ember-cli-legacy-blueprints`. That version bump should also appear in this PR. (Will be v0.1.2 if you want to add it optimistically.)

@trabus / @rwjblue / @stefanpenner can you grant me npm ownership of `ember-cli-legacy-blueprints`?

nathanhammond:
Also we'll need a `ember-source@2.10.0-alpha.2` so that it's named correctly.

jonnii:
Maybe we can also not generate `bower.json`? I suppose one reason for keeping it is because otherwise `ember init` won't show the changes?

locks:
@jonnii my reasoning for not removing it is that I think it'll be confusing, especially for newcomers who might need to bring in some bower dependency and can't find the file to declare it. Also, while we are moving every Ember dependency into npm-land, we do have to support bower until the next major version. My suggestion would be to remove it then :)

stefanpenner:
> Maybe we can also not generate bower.json? I suppose one reason for keeping it is because otherwise ember init won't show the changes?

I'm not sure we can easily do that today. Revisiting the generators at some point is likely a good idea, but I am not sure we can block this onthat.

jonnii:
@stefanpenner by that time we'll all be using yarn anyway.

stefanpenner:
> @stefanpenner by that time we'll all be using yarn anyway.

@jonnii one can dream -> https://github.com/yarnpkg/yarn/issues/650

blueprints/addon/files/.travis.yml
```diff
@@ -17,6 +17,7 @@ env:
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
+  # - EMBER_TRY_SCENARIO=ember-source
```
rwjblue: Why is this commented?

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -55,6 +55,19 @@ module.exports = {
...
+    {
+      name: 'ember-source',
+      bower: {
+        dependencies: {
+          'ember': null
+        }
+      },
+      npm: {
+        dependencies: {
+          'ember-source': 'latest'
```
rwjblue: Seems like this isn't needed? Since we already depend on the "right" version of ember-source in package.json...

blueprints/addon/files/.travis.yml
rwjblue: If we don't add it (since we removed `default`) we will never _actually_ run tests with ember-source in CI...

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -32,27 +42,31 @@ module.exports = {
...
       name: 'ember-beta',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#beta'
-        },
-        resolutions: {
-          'ember': 'beta'
+      npm: {
+        devDependencies: {
+          'ember-source': 'beta'
```
kategengler: This works?

nathanhammond:
Closing, this was landed in a different way.

Issue #6323, opened 2016-10-06
ember install from git


ming-codes:
Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.8.0
http_parser: 2.5.2
node: 4.3.2
v8: 4.5.103.35
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 46
openssl: 1.0.2g
os: darwin x64
2.14.12
```

---

```
>  ember install git+ssh://git@<host>:<name>/<repo>.git 
Installed packages for tooling via npm.
WARNING: Could not figure out blueprint name from: "git+ssh://git@<host>:<name>/<repo>.git". Please install the addon blueprint via "ember generate <addon-name>" if necessary.
Installed addon package.
```

The install seems to be working until it tries to find the default blueprint.

nathanhammond:
This is not a bug. The message you're seeing is telling you what to do:
```
WARNING: Could not figure out blueprint name from: "git+ssh://git@<host>:<name>/<repo>.git". Please install the addon blueprint via "ember generate <addon-name>" if necessary.
```

There is no way to actually know the addon's name from a git path.

ming-codes:
`package.json` offers a name that's generated when the addon is first created. The addon's name can be inferred from this.

nathanhammond:
Yes, it's possible to do this by inspecting state of `package.json` before and after we delegate to `npm install`, running a diff, `require.resolve`ing these new packages, and then figuring out their addon name. Inference is _not_ good enough and something we got out of the business of doing because of the associated errors. It must be correct, or we don't try.

Doing the above is incredibly low on our priority list since it doesn't represent the common case and requires manually running blueprints only once. If this is something you'd like to undertake we're glad to help guide you through the process of contributing it! :smile:

(Steps to do it already described.)

ming-codes:
Sure, I'll give it a shot.

The name of the addon is also in `index.js`. Would this be a better choice?

> Doing the above is incredibly low on our priority list since it doesn't represent the common case.

I'm surprised by this statement. How does enterprise distribute internal Ember addons then? It's a common scenario to install npm modules directly from git repo. I would imagine it's the same case for Ember addons.

nathanhammond:
I work at LinkedIn, we have our own internal npm registry. 😄

[...]

PR #6317, opened 2016-10-06
Force npm install to automatically retry per Travis CI recommendations.

nathanhammond:
`npm install` is failing on the 0.12 boxes.

PR #6325, opened 2016-10-07
Move to bundled install test.

### LINKED ISSUE #6322: `ember install` test uses local file system package.

nathanhammond:
Closes #6322.

Caveats:
- Includes a large .tgz file into the repo. (bundled the dependencies)
- Doesn't actually make us 100% offline capable, but reduces network exposure for `npm`.
- No longer tests for automatically invoking blueprints because package wasn't installed by package name.
- Protects us from random external addons breaking our test suite.
- We could do what [I describe here](https://github.com/ember-cli/ember-cli/issues/6323#issuecomment-252120185) in some future state and get the auto-invocation test back. (Though since likely nobody will claim this task we should consider this PR as if it will never happen.)
- [Would allow us to revert this line.](https://github.com/ember-cli/ember-cli/pull/6317#discussion_r82209740)

Discuss.

### DIFF
--- a/tests/acceptance/install-test-slow.js
+++ b/tests/acceptance/install-test-slow.js
@@ -44,20 +44,27 @@ describe('Acceptance: ember install', function() {
     ]);
   }
 
-  function installAddon(args) {
-    var generateArgs = ['install'].concat(args);
+  function installAddon(addon) {
+    addon = path.resolve(path.join(__dirname, '..', 'fixtures', 'install', addon));
+
+    var commandArgs = ['install'].concat(addon);
 
     return initApp().then(function() {
-      return ember(generateArgs);
+      return ember(commandArgs);
     });
   }
 
   it('installs addons via npm and runs generators', function() {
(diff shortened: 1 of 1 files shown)


kellyselden:
> Protects us from random external addons breaking our test suite.

We can mitigate this by moving to an addon we control, if this PR is not to land.

stefanpenner:
> We can mitigate this by moving to an addon we control, if this PR is not to land.
> Add your review

i would actually recommend we create an add-on with the soul purpose of providing something stable to point too

nathanhammond:
1. We should clearly create a simple test addon.
2. Even if we move to an addon we control we're still hitting the network during tests which is a great way to make our test suite nondeterministic and the source of the 0.12 problems.

I'm uncomfortable making 0.12 an allowed failure as that may mask a real underlying issue and I may forget to check the actual status page because habits.

nathanhammond:
Closing in favor of creating our own addon. Anybody want dibs on that?

PR #6327, opened 2016-10-07
fail fast semantics for test runs

stefanpenner:
For failures, we want to:

* encourage non-flakey tests (retrying just hides this)
* fail fast, not 3x as long


nathanhammond:
@ember-cli/core With this landing please pay _CLOSE_ attention and manually review all test runs in 0.12. You may have to restart them. We need to see at least one pass of the test suite in 0.12.

PR #6340, opened 2016-10-10
extract podTemplates, and small test

stefanpenner:


stefanpenner:
> I love the frugality of the test.

Ya... the setup require for ember-app is heavy... This at-least ensures it is invoked and not removed if that code remains the same. It also aims that removal has removed something one should not.

stefanpenner:
this is so i can have "public" api for: https://github.com/stefanpenner/ember-cli-no-pods

olleolleolle:
(Would it be reasonable to also extract the other `new Funnel` at:

https://github.com/ember-cli/ember-cli/blob/52fc04568f4a0202fae5ecc1162c44c86ec02f44/lib/broccoli/ember-app.js#L704

)

stefanpenner:
> (Would it be reasonable to also extract the other new Funnel as its own method? It was at:

Im not sure what that would do, can you share your thoughts/goals?

stefanpenner:
@nathanhammond which release channel can we ship this one on (i want to use it in our app @ work)

olleolleolle:
@stefanpenner Oh, just for readability and harmony in that method. Doing it the same way for both appends to that list. (BUT: Adding API surface is a larger concern than I was seeing when I made my comment.)

PR #6343, opened 2016-10-12
fix documentation for app-boot contentFor

chrisjshull:
I don't see any 'app' contentFor. But there *is* an 'app-boot' contentFor that's mysteriously missing.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1679,7 +1679,7 @@ EmberApp.prototype.toTree = function(additionalTrees) {
   Currently supported types:
   - 'head'
   - 'config-module'
-  - 'app'
+  - 'app-boot'
   - 'head-footer'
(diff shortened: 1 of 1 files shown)


stefanpenner:
this is most likely correct, good catch.

nathanhammond:
It exists, and it's not deprecated, however: https://github.com/ember-cli/rfcs/pull/30

I _personally_ would prefer to simply remove the `app` item and start a "passive" deprecation process where usage fades by virtue of lack of documentation–effectively turning it into a power-user feature. People who then elect to use it won't have issues migrating to the new thing because of their familiarity with the codebase.

I suspect that this is why @stefanpenner tagged this as discussion. @ember-cli/core thoughts?

nathanhammond:
Closing in favor of #6360. Thanks for your attention to detail, @chrisjshull!

PR #6344, opened 2016-10-12
original case for app/addon name in readme

kellyselden:
I think most people do this themselves after initial commit. I prefer this as well.


stefanpenner:
seems good to me, anyone else with strong opinions?

Turbo87:
very much 👍  on this!

PR #6363, opened 2016-10-19
[ENHANCEMENT] Add 2.8-lts scenario to default ember-try config

BrianSipple:
Now that the [2.8 LTS release is out](http://emberjs.com/blog/2016/10/17/ember-2-9-released.html), it may be beneficial to add this to the blueprint for an addon's `ember-try` scenarios.


rwjblue:
:+1: - I'd also like to consider dropping 1.13, but that probably should be a different PR/debate...

kellyselden:
@homu r+

PR #6371, opened 2016-10-21
blueprints/app: Update Ember and Ember Data to v2.9.0

Turbo87:
This will most likely need a v2.9.1 release... sorry...

/cc @nathanhammond


stefanpenner:
> This will most likely need a v2.9.1 release

ya this should likely happen.

Issue #6269, opened 2016-09-16
Remove remaining coupling to JSHint


Turbo87:
Currently our `app` and `addon` blueprints still include `.jshintrc` files (see https://github.com/ember-cli/ember-cli/tree/master/blueprints/app/files), which are now also provided by the default blueprint in `ember-cli-jshint` (see https://github.com/ember-cli/ember-cli-jshint/pull/21). To remove the duplication we should remove the `.jshintrc` files from our blueprint and run the `ember-cli-jshint` blueprint once the dependency is installed.

Unfortunately it's not quite as simple as adding the `lookupBlueprint().install()` call to the `afterInstall()` hook of the `app` and `addon` blueprints, because that hook will run _before_ `npm install` was executed. So currently possible is only:
1. `app` blueprint adds a `package.json` file including the `ember-cli-jshint` dependency
2. `app` blueprint tries to install `ember-cli-jshint` blueprint but fails because the dependency is not yet installed
3. `npm install` will install the `ember-cli-jshint` dependency but not run the corresponding blueprint

Now I am looking for ways to install the blueprint _after_ `npm install` has finished without breaking the world. Is there a good way to do this in the current code base?

Should we maybe implement a public `installAddon()` method that only resolves once `npm install` for that addon has finished?

/cc @rwjblue @stefanpenner

nathanhammond:
`installAddon()` would neatly solve the running of default blueprints issue for non-module-name-identified packages. I'm in favor of doing something like that. This is a pattern we should rely upon more, so I like the proposal.

aureliosaraiva:
How is this project making eslint default?
Can you detail how to reproduce the problem?
The problem was not clear to me!
It makes me willing to help in everything

PR #6378, opened 2016-10-23
Prepares Ember CLI for new version of ember-welcome-page

locks:
- Update ember-welcome-page dependency
-  Re-introduce application.hbs template to include the new
welcome-page component

- [x] Release new `ember-welcome-page` version

### DIFF
--- a/blueprints/app/files/app/templates/application.hbs
+++ b/blueprints/app/files/app/templates/application.hbs
@@ -1,0 +1,5 @@
+{{!-- The following component displays Ember's default welcome message. --}}
+{{welcome-page}}
+{{!-- Feel free to remove this! --}}
+
+{{outlet}}
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -34,7 +34,7 @@
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.5.1",
     "ember-resolver": "^2.0.3",
-    "ember-welcome-page": "^1.0.3",
+    "ember-welcome-page": "^2.0.1",
     "loader.js": "^4.0.10"
   },
   "engines": {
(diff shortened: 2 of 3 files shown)


nathanhammond:
@ember-cli/core Thoughts on having Ember CLI run the default blueprint of any addon present after `ember new` or `ember addon`? This would avoid us having to hardcode what amount to blueprint-level changes into our `app`/`addon` blueprints.

I'm hesitant to enshrine what feel like blueprint related tasks into the blueprints. If y'all feel like that's a good feature I'll build it today (so that welcome-page may adopt it).

Turbo87:
@nathanhammond 💯 👍  (see #6269)

kellyselden:
Sounds good to me. `ember init` will probably want this feature too.

nathanhammond:
I'm less sure about `ember init` as it's intended to be run in projects that are already set up.

locks:
> I'm less sure about ember init as it's intended to be run in projects that are already set up.

We should have this as a different thing 🤔

kellyselden:
> I'm less sure about ember init as it's intended to be run in projects that are already set up.

What about the scenario, you're doing an ember-cli upgrade via `ember init`. The default blueprint for package.json updated `ember-cli-eslint` from ^1.0.0 to ^2.0.0. There was a big default blueprint change in `ember-cli-eslint` that you need to run.

Isn't that the same problem as with `new` and `addon`?

nathanhammond:
@kellyselden You make a good point. But unfortunately that implies that default blueprints must be idempotent/safe to rerun which is not something that we require at this time since they're only ever invoked once. This is a larger problem than I was thinking and I no longer want to block this PR on that effort.

kellyselden:
Good point as well, I wasn't thinking of the blowing away of previous custom changes to blueprint. Carry on, no changes to `ember init`.

nathanhammond:
Hokay, so, @locks, this is not blocked by CLI people's opinions. Just tests passing once the new version is released.

nathanhammond:
@homu r+

nathanhammond:
@locks Thank you for your tireless efforts on this!

PR #6385, opened 2016-10-25
Respect testem exit code

johanneswuerbach:
Currently the UX around failing tests is bad as devs see a large
stack trace when tests failed.

This changes the behaviour to log a silent error depending on the
testem exit code.

The previous "No tests were run" handling has been removed as this
is already handled by testem upstream:
https://github.com/testem/testem/blob/be8348bf/lib/app.js#L439

While both checks will work with all testem versions, new versions won’t return not all tests passed / run cancelled errors any longer.
https://github.com/testem/testem/pull/1001

### DIFF
--- a/lib/tasks/test-server.js
+++ b/lib/tasks/test-server.js
@@ -3,15 +3,21 @@
 var TestTask = require('./test');
 var Promise  = require('../ext/promise');
 var chalk    = require('chalk');
+var SilentError = require('silent-error');
 
 module.exports = TestTask.extend({
   invokeTestem: function(options) {
     var task = this;
 
     return new Promise(function(resolve, reject) {
-      task.testem.startDev(task.testemOptions(options), function(code, error) {
-        if (error) { reject(error); }
-        else       { resolve(code); }
+      task.testem.startDev(task.testemOptions(options), function(exitCode, error) {
+        if (error) {
+          reject(error);
+        } else if (exitCode !== 0) {
+          reject(new SilentError('Testem finished with non-zero exit code. Tests failed.'));
+        } else {
+          resolve(exitCode);
+        }
       });
     });
   },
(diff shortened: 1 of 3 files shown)


lib/tasks/test-server.js
```diff
@@ -9,10 +9,13 @@ module.exports = TestTask.extend({
     var task = this;
 
     return new Promise(function(resolve, reject) {
-      task.testem.startDev(task.testemOptions(options), function(code, error) {
-        if (error) { reject(error); }
-        else       { resolve(code); }
-      });
+      if (error) {
+        reject(error);
+      } else if (exitCode !== 0) {
```
stefanpenner: where does `exitCode` come from?

lib/tasks/test-server.js
```diff
@@ -9,10 +9,13 @@ module.exports = TestTask.extend({
     var task = this;
 
     return new Promise(function(resolve, reject) {
-      task.testem.startDev(task.testemOptions(options), function(code, error) {
```
johanneswuerbach: My bad :-(

nathanhammond:
@johanneswuerbach

1. Do we need to wait until https://github.com/testem/testem/issues/1001 lands?
2. Legitimate test failures because error messages changed. Can you adjust those tests in this PR as well?

This looks like a good change to make. I like having that upstreamed.

johanneswuerbach:
@nathanhammond

1. No, I'll wait merging the testem PR until this PR gets released as otherwise the exit code of ember test might be 0 even when tests failed. Would there be any chance in bringing this fix directly into the beta channel?
2. Fixed!

nathanhammond:
Considering the change I don't want to land it directly in the beta channel. I want to give it time for people to identify the change in behavior and get user feedback.

nathanhammond:
@homu r+

PR #6387, opened 2016-10-26
Adds Node 7 to testing matrix

twokul:


kellyselden:
hmm node 7 not supported yet on appveryor

stefanpenner:
> hmm node 7 not supported yet on appveryor

maybe we should leave it off appveyor for now?

@twokul awesome!

kellyselden:
> maybe we should leave it off appveyor for now?

yes, lets split into 2 PRs so we can get travis coverage at least.

twokul:
sure, let me quickly change

stefanpenner:
👍

nathanhammond:
We need to add this to the engines key in package.json as well.

Issue #6348, opened 2016-10-14
`analytics.trackTiming()` always receives the category "rebuild"


mdebbar:
I would expect the category "build", "cold-build" or "warm-build" when I run a new build. But I always get a "rebuild" category. Is that the intended behavior? I would like to be able to log and track the timing of different categories separately "cold-build", "warm-build" and "rebuild".

Some more details:

`ember build` produces the following:
--
```javascript
{
  category: "rebuild",
  variable: "build time",
  label: "broccoli build time",
  duration: 12345
}
```
Code [here](https://github.com/ember-cli/ember-cli/blob/375f3a32f4564465d2eccc3815cb61b570ce29f0/lib/tasks/build.js#L31-L36). I was expecting to get `category: "build"` (or `"cold-build"`/`"warm-build"` if we want to be more specific).

`ember build --watch` produces:
--
```javascript
{
  category: "rebuild",
  variable: "rebuild time",
  label: "broccoli rebuild time",
  duration: 12345
}
```
Code [here](https://github.com/ember-cli/ember-cli/blob/ab70f424494245e82b1b5f3a25265abbd5f3f325/lib/models/watcher.js#L52-L57). I was expecting to get `category: "build"` for the first build, then `"rebuild"` on subsequent builds.

I know how to fix the first one (`ember build`) and I can send a PR. But I'm not sure about the second (`ember build --watch`).

---

Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.6.3
http_parser: 2.7.0
node: 6.7.0
v8: 5.1.281.83
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
3.10.7
```

nathanhammond:
This is likely because you had a persistent cache and it identified that it wasn't a completely cold build. Can you confirm by removing `/tmp/*async-disk-cache-*`?

nathanhammond:
Oh look at that, it's hardcoded. I simply assumed that we tried to do the right thing.

rwjblue:
haha, never assume that!

nathanhammond:
@rwjblue I feel like :trollface: should be promoted to a reaction.

stefanpenner:
we could in theory differentiate between first rebuild, and subsequent rebuilds. But beyond that, the rest is tricky.

twokul:
[...]

PR #6396, opened 2016-10-30
Document analytics collection.

nathanhammond:
~~The current analytics timing for the `build` task provides the `rebuild` category every time (per #6348). This should be `build` as the build task is only ever invoked once.~~

~~`ember build --watch` uses a different task.~~

/cc @twokul


nathanhammond:
Note: probably wrong about the previous message. 😛

nathanhammond:
So, after reviewing the reports this is correct. The switch is not the `category` but the `variable`.

Closes #6348.

/cc @mdebbars

stefanpenner:
thanks !

PR #6405, opened 2016-11-01
Node.js 0.12 Under Control. Whitelist build branches.

nathanhammond:
Reduce the number of spurious builds we do so that Travis and AppVeyor (particularly) don't get as far backed up.

Also, Node.js 0.12 appears to have stabilized on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,11 @@ node_js:
   - "6"
   - "7"
 
-matrix:
-  allow_failures:
-    - node_js: "0.12"
-    - node_js: "node"
- 
 branches:
-  except:
-    - canary
+  only:
+    - master
+    - beta
+    - release
 
 cache:
   directories:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,8 +8,10 @@ environment:
     - nodejs_version: "6"
 
 branches:
-  except:
-    - canary
+  only:
+    - master
+    - beta
+    - release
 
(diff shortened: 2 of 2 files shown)


nathanhammond:
@homu r+

.travis.yml
```diff
@@ -7,14 +7,11 @@ node_js:
...
-    - node_js: "0.12"
-    - node_js: "node"
- 
 branches:
-  except:
-    - canary
+  only:
+    - master
+    - beta
+    - release
```
Turbo87: does that still build PRs too?

rwjblue: Yes, it will build PRs but not branch builds when pushing to origin. This avoids running tests twice for the same PR (Travis/push and Travis/pr).

Turbo87: what about Greenkeeper branches?

rwjblue: Confirm, green keeper and homu branches need to be included or they won't run (likely why @nathanhammond couldn't get homu to run on a PR last night after this landed).

nathanhammond: I'm unsure about homu and greenkeeper... I'll monitor and confirm that today. I was mostly annoyed that I was backed up on test runs and trying to get out of that mess.

kellyselden: here is what you need for greenkeeper https://github.com/kellyselden/package-hint-historic-resolver/blob/master/.travis.yml#L11

PR #6407, opened 2016-11-01
Improve silent.js Deprecation

### LINKED ISSUE #6190: Deprecation warnings should output more details as to what library/package is the cause

nathanhammond:
Supersedes #6245. Closes #6190.

Example error message (added it to the version command for testing because I'm lazy):
```
DEPRECATION: `ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead. Required here: 
  Object.<anonymous> (/Users/nathanhammond/Repos/ember-cli/lib/commands/version.js:5:11)
```

@RyanNerd thanks for making this happen!

### DIFF
--- a/lib/errors/silent.js
+++ b/lib/errors/silent.js
@@ -3,9 +3,12 @@
 var SilentError = require('silent-error');
 var deprecate   = require('../utilities/deprecate');
 
+// Get the call stack so we can let the user know what module is using the deprecated function.
+var stack = new Error().stack;
+
 Object.defineProperty(module, 'exports', {
   get: function () {
-    deprecate('`ember-cli/lib/errors/silent.js` is deprecated, use `silent-error` instead.', true);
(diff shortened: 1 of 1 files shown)


nathanhammond:
(Note @Turbo87 @rwjblue: I intend to use this to test Homu, which is why I opened this PR.)

nathanhammond:
@homu r+

nathanhammond:
Homu appears to work in the "this is still against `HEAD`" scenario.

PR #6408, opened 2016-11-01
Fix greenkeeper.

nathanhammond:


nathanhammond:
@homu r+

rwjblue:
@nathanhammond - I think we also need to add `auto` which is what `homu` uses.

nathanhammond:
@homu r+

nathanhammond:
@homu retry

PR #6417, opened 2016-11-06
[BUGFIX] Makes sure that only internal commands are tracked

twokul:
Before this PR we were tracking all of commands (ember-cli + custom). We probably want to be smarter about it and think of a better way to expose tracking info. For now, we likely should limit tracking only for internal commands.


lib/utilities/analytics-utils.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+var ANALYTICS_WHITELISTED_COMMANDS = [
```
nathanhammond: Is it possible to pull this list off of `this.commands`? We do that in `ember help` to get the list: https://github.com/ember-cli/ember-cli/blob/master/lib/commands/help.js but I don't know where it is populated.

I'd prefer not to hardcode our list of commands.

lib/models/command.js
```diff
@@ -200,10 +202,12 @@ var Command = CoreObject.extend({
         return resolve('callHelp');
       }
 
-      this.analytics.track({
-        name:    'ember ',
-        message: this.name
-      });
+      if (isWhitelisted(this.name)) {
```
nathanhammond: Can we make this extensible so that our own internal addons with commands (fastboot, others) can opt in to tracking?

nathanhammond:
I like this pattern, but I want additional perspective on it. It masks that we're pulling a model by laundering it through the `utils` folder. @ember-cli/core thoughts?

Turbo87:
I'm not entirely convinced that having a simple boolean flag justifies having a new base class. IMHO we should just add that flag to the existing commands without the need for the new base class.

nathanhammond:
@Turbo87 I think it's also a question of if we expect the number of things we do distinctly for internal commands to grow. Right now it's just analytics. Could be a case of YAGNI.

twokul:
@Turbo87 yeah, I thought about it and I'm actually with you on this one. The only reason I added a new class was b/c it was that internal command should likely have the same consistent behaviour (even if it is just a boolean flag). In the case when someone else wants to add a new internal command, she/he should just extend a class w/o thinking too much. I also kind of expect this new class to grow.

Having said all that, I can refactor in favour of a boolean flag.

lib/utilities/internal-command.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var Command = require('../models/command');
+
+/*
```
twokul: sure, I can add a comment

nathanhammond:
@twokul Don't do that yet!

Turbo87:
closing due to inactivity

twokul:
@Turbo87 do we not want to get this in? I'd be happy to rebase and land this.

Turbo87:
we should discuss this in the next call. in any case I'd much rather see an `internal` or `tracked` flag or something like that, instead of the separate base class.

twokul:
@Turbo87 I don't have a strong preference here apart from the fact that separate class enforces separation w/o having to know about special property (`internal` or `tracked`)

PR #6429, opened 2016-11-14
WIP: Cleanup some addon discovery related code

stefanpenner:
* Mimic node, only store realpaths for modules (add ons in our case)

- [x] demonstrate working 
  - 10x reduction in folders in tmp
  - very nice (not sharing until totally green tests) initial build and rebuild time improvements
- [ ] finish and publish related rfc
- [ ] make real, not hacky (maybe split up across mulitiple PR)
- [ ] fix tests

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -496,6 +496,17 @@ EmberApp.prototype.initializeAddons = function() {
 EmberApp.prototype.addonTreesFor = function(type) {
   return this.project.addons.map(function(addon) {
     if (addon.treeFor) {
+
+      var treeName = type;
+      var key = addon.pkg.name + '@' + addon.pkg.version + '#' + treeName;
+      var hasBeenBuilt = addon.project.treeForState[key];
+      if (hasBeenBuilt) {
+        // console.log('hit:', key);
+        return;
+      }
+      // console.log('miss:', key);
+      addon.project.treeForState[key] = true;
+
       return addon.treeFor(type);
     }
   }).filter(Boolean);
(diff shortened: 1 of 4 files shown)


rwjblue:
:+1:

lib/broccoli/ember-app.js
```diff
@@ -41,6 +41,15 @@ var isNull        = require('ember-cli-lodash-subset').isNull;
 var Funnel        = require('broccoli-funnel');
 var funnelReducer = require('broccoli-funnel-reducer');
 
+function mergeTrees(_trees, options) {
```
nathanhammond: Can we upstream this into `mergeTrees`?

lib/models/addon.js
```diff
@@ -248,8 +261,39 @@ var Addon = CoreObject.extend({
     this.initializeAddons();
 
     var invokeArguments = args || [];
-
+    var parent = this;
     return this.addons.map(function(addon) {
+      if (methodName === 'treeFor') {
```
nathanhammond: I'm tempted that, if the addon has non-default implementations for these hooks, we just build it. (Rather than looking it up.) I'd guess that this still gets us the majority of the win and _also_ makes it less likely to break things.

lib/broccoli/ember-app.js
stefanpenner: @nathanhammond i started with that, but it confuses the internal node wrappers, and annotations for wired up incorrectly.  I do think eventually it should be merely a broccoli primitive (maybe in BMT or maybe???), which elides pointless things.

rwjblue:
Closing in favor of https://github.com/ember-cli/ember-cli/pull/6573

stefanpenner:
@rwjblue thanks making my mega-hack a real thing :)

PR #6437, opened 2016-11-16
[BUGFIX beta] more resilient signal trapped cleanup

stefanpenner:
Cooperative signal handling is complicated, in-order to provide the
experience we want, we take control of this process entirely.

See: ://github.com/sindresorhus/ora/issues/27 for more context

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -7,6 +7,10 @@ var logger        = require('heimdalljs-logger')('ember-cli:cli/index');
 var merge         = require('ember-cli-lodash-subset').merge;
 var path          = require('path');
 
+// work around misbehaving libraries, so we can correctly cleanup before
+// actually exiting.
+require('capture-exit').captureExit();
+
 // ember-cli and user apps have many dependencies, many of which require
 // process.addListener('exit', ....) for cleanup, by default this limit for
 // such listeners is 10, recently users have been increasing this and not to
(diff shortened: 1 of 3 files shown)


lib/cli/index.js
```diff
@@ -7,6 +7,10 @@ var logger        = require('heimdalljs-logger')('ember-cli:cli/index');
 var merge         = require('ember-cli-lodash-subset').merge;
 var path          = require('path');
 
+// work around misbehaving libraries, so we can correctly cleanup before
+// actually exiting.
+require('capture-exit').captureExit();
```
nathanhammond: Note why I'm not concerned about this: even if things were to be interrupted before this point it doesn't suffer the negative consequences that we are trying to prevent.

lib/models/builder.js
```diff
@@ -84,6 +86,9 @@ module.exports = Task.extend({
       process.on('SIGINT',  this.onSIGINT.bind(this));
       process.on('SIGTERM', this.onSIGTERM.bind(this));
       process.on('message', this.onMessage.bind(this));
+      exit.onExit(function(code) {
+        return this.cleanupAndExit(code);
```
nathanhammond: `this.cleanup.bind(this);`

lib/cli/index.js
hjdivad: @nathanhammond what are you not concerned about?  @Turbo87's suggestion to move this before other requires?

His point seems valid; we want to `captureExit` as soon as possible; this is unrelated to when an interruption occurs because we want to get head not only of interruptions but also people who pluck off `process.exit`

nathanhammond: Didn't think about other picking off `process.exit` in advance. Manual review says that isn't a concern here, but it could accidentally become a problem in the future if not hoisted. Seems good to be addressed.

PR #6443, opened 2016-11-21
Fix preProcessTree API docs.

kratiahuja:
`preProcessTree` doesn't appear in addon hooks [here](https://ember-cli.com/api/classes/Addon.html)

cc: @stefanpenner @rwjblue


rwjblue:
@homu r+

PR #6449, opened 2016-11-21
Nested Preprocessor Tests, (App|Addon|InRepoAddon)Fixture

nathanhammond:
Writing acceptance tests in Ember CLI has been fraught with peril. For many months we were unable to remove Ember Data from the addon `package.json` blueprint for reasons of test state leakage. With `PackageCache` now landed we've managed to reduce the runtime of our test suite by four minutes and prevented package state from leaking between tests. (And landed the Ember Data PR!)

Now, rather than relying on a shared non-composable fixture provided by `tests/helpers/acceptance.js`, I've created `AppFixture`, `AddonFixture`, and `InRepoAddonFixture`. These are built on top of `PackageCache` and `fixturify` to provide a DSL for programmatically building fixtures.

These fixture classes and associated DSL should be considered a first step toward writing tests which are able to run against our Broccoli build pipeline and incrementally build and test.

TODO:

- [x] Write preprocessor tests.
- [x] Documentation of the new test helpers.
- [x] Ensure that `dir` is read-only.
- [x] Guarantee that the directory is always present (even if empty) so that we don't accidentally clobber it.
- [x] Guarantee `serialize` is idempotent.
  - [x] Support Node 4.
  - [x] Make ESLint happy.
  - [x] Make sure that re-linking the PackageCache is safe.
  - [x] Mark fixture as dirty if there are any modifications. If dirty it will re-`serialize` on next serialization call.
  - [x] Make `serialize` a no-op if already serialized and a "clean" fixture.
- [x] Provide a helper to run `ember` commands in the context of a fixture?
- [ ] Allow for setting custom `PackageCache` cache names, and `clone`ing from an existing cache.
- [ ] Tests of the new test helpers.

OPTIONAL:

- [ ] Provide an API in `PackageCache` to make the `node_modules` symlinking a single method call.
- [ ] Migrate `tests/helpers/acceptance.js` to use these under the hood.
- [ ] Smarter reuse of existing package caches.
- [ ] Streamline API warts.
  - [ ] https://github.com/joliss/node-quick-temp/pull/16

---

<details>
<summary>Original description for posterity.</summary>
The CSS behavior inside of Ember CLI is a bit like the wild wild west. We want to get to the point where there is _no_ default CSS handling hardcoded into Ember CLI. In order to accomplish that we _must_ fully understand what it does so that we can successfully extract it into an addon. Once extracted into an addon that makes it possible for experimentation in this space to be more effective by virtue of not having to fight the built-in behavior.

---

This is step one. There will be no new features added, no change of behavior, possible refactoring to enable better testability. This will simply test and capture the output of the build for every conceivable CSS file layout, arrangement, and naming scheme.

Step two (in a followup PR) will be to remove all CSS behavior inside of Ember CLI and achieve identical results from an addon. I ...erm... may or may not be squatting on [`ember-cli-css`](https://www.npmjs.com/package/ember-cli-css). Dev will be done at my personal GitHub account, once the PR for this step is ready to land I'll transfer it to the `ember-cli` org.

Step three (contemporaneously with step two) will identify if this addon plays nicely with the existing Sass and LESS preprocessors. If not they'll be subjected to the same battery of tests to guarantee that they're compatible.

Once this has been proven to work I'll write an RFC which should be nearly rubber-stampable as it will state "there are no changes to the outcomes, only the method in which it is accomplished." The largest caveat is that there may possibly be a mechanical-only change for addon authors, which is to declare a dependency on `ember-cli-css`.
</details>


stefanpenner:
> This is step one. There will be no new features added, no change of behavior, possible refactoring to enable better testability.

I <3 it

stefanpenner:
Since we are locking down behavior, this likely needs thorough review.

nathanhammond:
So, this has in part turned into "make it easy to write good Ember CLI acceptance tests." API goals:

- Ability to compose `app` and `addon` fixtures.
- Ability to safely reuse an existing `node_modules` or `bower_components` directory across acceptance tests.
- Automatic setup and teardown.
- Significantly less boilerplate.
- Ability to leverage warm builds.
- Faster.

stefanpenner:
It would be great to get to a point, where we can use broccoli test infrastructure similar to: https://github.com/chadhietala/broccoli-rollup/blob/master/tests/index.js

To test `EmberApp`. I suspect this could dramatically help sanity/perf etc.

nathanhammond:
Output:
```
ananke:self-bRL6X2TR.tmp nathanhammond$ tree .
.
├── README.md
├── app
│   ├── app.js
│   ├── index.html
│   ├── resolver.js
│   ├── router.js
│   ├── styles
│   │   ├── _import.css
│   │   ├── addon.css
│   │   ├── alpha.css
│   │   ├── app.css
│   │   └── zeta.css
│   └── templates
│       └── application.hbs
├── bower.json
├── bower_components -> /Users/nathanhammond/Repos/ember-cli/tmp/bower_components-POBznhu5.tmp/bower_components
├── config
│   └── environment.js
├── ember-cli-build.js
├── node_modules -> /Users/nathanhammond/Repos/ember-cli/tmp/node_modules-Qlr1I9uY.tmp/node_modules
├── package.json
├── public
│   ├── crossdomain.xml
│   └── robots.txt
├── testem.js
└── tests
    ├── helpers
    │   ├── destroy-app.js
    │   ├── module-for-acceptance.js
    │   ├── resolver.js
    │   └── start-app.js
    ├── index.html
    └── test-helper.js

9 directories, 24 files
```

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,206 @@
...
+    useGlobalPackages: true
+  };
+
+  this.fixture = {};
+  this._fixtureCache = {};
+  this.dirs = {};
+  quickTemp.makeOrRemake(this.dirs, 'self');
+
+  this.blueprintPath = path.resolve(__dirname, '../../blueprints/app');
+  this.blueprintOptionsShim = {
```
chadhietala: Probably be a better idea to create `MockBlueprintOptions`

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,206 @@
...
+        return name;
+      }
+    }
+  };
+
+  this.loadBlueprint();
+}
+
+AppFixture.prototype = {
+  _bowerInstall: function() {
```
chadhietala: Is there a reason why these use `execSync`? I feel like you could create an API that this like this that should cut down the serial aspect of install deps for tests:

```
installDeps() {
  return Promise.all([this._bowerInstall(), this._npmInstall()]);
}
```

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,206 @@
...
+    var cursor = root;
+    for (var i = 0; i < keyPath.length - 1; i++) {
+      cursor = cursor[keyPath[i]] = {};
+    }
+    cursor[keyPath[i]] = contents;
+
+    merge(this.fixture, root);
+    return this;
+  },
+  generateCSS: function(fileName) {
```
chadhietala: Nit: requires whitespace.

tests/helpers/blueprint-shim.js
```diff
@@ -0,0 +1,60 @@
+var fs = require('fs-extra');
+var path = require('path');
+
+var stringifyAndNormalize = require('../../lib/utilities/stringify-and-normalize');
+var transformer = require('../../blueprints/addon/index');
+
+var processTemplate = require('../../lib/utilities/process-template');
+
+// The blueprint paths for `bower.json` and `package.json` so we're using the right things.
+var paths = {
```
chadhietala: Can you get away with just doing `process.cwd()` instead of `join`ing and walking back to the root of the project?

tests/helpers/blueprint-shim.js
```diff
@@ -0,0 +1,60 @@
...
+    'package.json': fs.readFileSync(paths['package.json'], 'utf8')
+  },
+  addon: {
+    'bower.json': '',
+    'package.json': ''
+  }
+};
+
+// 2. Stub out the shape of `this` expected by the transforms in the addon blueprint.
+var BlueprintShim = {
```
chadhietala: Perhaps create a `MockBlueprint` class.

tests/helpers/package-cache.js
```diff
@@ -0,0 +1,153 @@
+var fs = require('fs-extra');
```
chadhietala: Would be great if we could pull this out for now.

tests/helpers/sync-blueprint.js
```diff
@@ -0,0 +1,115 @@
...
+
+    var reduce;
+    reduce = beforeHook.bind(self, options, locals)();
+    reduce = process.bind(self, intoDir, locals)(reduce);
+    reduce = reduce.map(self._commit.bind(self));
+    reduce = afterHook.bind(self, options)(reduce);
+
+    return {
+      finally: function() {
+        // return reduce;
```
chadhietala: ???

tests/helpers/blueprint-shim.js
nathanhammond: I'm setting this up to allow us to run tests outside of the context of the Ember CLI folder. I'm intentionally using root-relative paths where at all possible.

tests/helpers/app-fixture.js
nathanhammond: I'm going to refactor this work on top of `blprnt`, this was a first pass to prove things out. I'm coming back to this tomorrow-ish.

tests/helpers/package-cache.js
nathanhammond: Yep, that's https://github.com/ember-cli/ember-cli/pull/6461

tests/helpers/sync-blueprint.js
nathanhammond: I'm mocking out the `Promise` shape. This is going away.

chadhietala:
QQ: Is there a reason why core-object is not used for the classes?

nathanhammond:
@chadhietala Because ES6 classes are a thing we get to use in five days!

package.json
```diff
@@ -78,6 +78,7 @@
     "express": "^4.12.3",
     "filesize": "^3.1.3",
     "find-up": "^1.1.2",
+    "fixturify": "^0.3.0",
```
rwjblue: Are we actually using this outside of tests?  Seems like this should be in `devDependencies`?

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+  var stew = require('broccoli-stew');
+  var addon = this;
+
+  // We're going to add a marker. The `postprocessTree` function will remove
+  // the marker if present and identify whether or not it found the marker.
+  var marker = '// ' + addon.name + '-preprocessTree(' + type + ')' + ' /';
+
+  tree = stew.map(tree, function(content, relativePath) {
+    var localMarker = marker + relativePath;
+    addon._findHost().project.ui.writeLine(localMarker);
```
rwjblue: `addon.project.ui` should work without needing to couple this to `_findHost`.

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+  var addon = this;
+
+  var preprocessTreeMarker = '// ' + addon.name + '-preprocessTree(' + type + ')';
+  var marker = '// ' + addon.name + '-postprocessTree(' + type + ')';
+
+  // We're going to inspect state and add an appropriate marker.
+  tree = stew.map(tree, function(content, relativePath) {
+
+    // Skip assets we know that the CSS preprocessor adds.
+    var cssName = addon._findHost().name;
```
rwjblue: `addon.project.name` unless you are trying use the same thing to ultimately test enginey things?

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+    var localPreprocessTreeMarker = preprocessTreeMarker + ' /' + preprocessPath;
+    var preprocessTreeMarkerIndex = content.indexOf(localPreprocessTreeMarker);
+    var preprocessTreeMarkerLastIndex = content.lastIndexOf(localPreprocessTreeMarker);
+
+    var localMarker = marker;
+
+    if (preprocessTreeMarkerIndex === -1) {
+      localMarker += '-no-preprocessTree';
+      localMarker += ' /' + relativePath;
+      addon._findHost().project.ui.writeLine(localMarker);
```
rwjblue: `addon.project.ui.writeLine`

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+      return localMarker + '\n' + content;
+    } else {
+      localMarker += '-removed-preprocessTree';
+      localMarker += ' /' + relativePath;
+      addon._findHost().project.ui.writeLine(localMarker);
```
rwjblue: `addon.project.ui.writeLine`

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+      return content.replace(localPreprocessTreeMarker, localMarker);
+    }
+  });
+
+  return tree;
+};
+
+describe('Acceptance: nested preprocessor tests.', function() {
+  this.timeout(1000 * 60 * 10);
```
rwjblue: I'd prefer to just use the whole number and add a comment for human form, thoughts?

```js
this.timeout(600000); // 10 minutes
```

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+    var name, type, ext;
+    var inRepoAddons = {};
+    var nestedInRepoAddons = {};
+    for (var i = 0; i < types.length; i++) {
+      type = types[i];
+      ext = extensions[i];
+      name = type + '-addon';
+
+      inRepoAddons[type] = new InRepoAddonFixture(name);
+      inRepoAddons[type].addMethod('preprocessTree', preprocessTreeFixture.toString());
```
rwjblue: `toString` here seems odd to me (just an observation)

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+    inRepoAddons['css'].generateCSS('addon/styles/alpha.css');
+    inRepoAddons['css'].generateCSS('addon/styles/zeta.css');
+
+    inRepoAddons['js']._npmAddonInstall({ name: 'ember-cli-babel' });
+    inRepoAddons['js'].generateJS('app/routes/hat.js');
+    inRepoAddons['js'].generateJS('app/routes/cat.js');
+    inRepoAddons['js'].generateJS('addon/components/thing-one.js');
+    inRepoAddons['js'].generateJS('addon/components/thing-two.js');
+
+    inRepoAddons['template']._npmAddonInstall({ name: 'ember-cli-htmlbars' });
```
rwjblue: Why underscored?  Can we have a method that is just `.addonInstall` (instead of tying to `npm`)?

tests/acceptance/nested-preprocessor-test-slow.js
```diff
@@ -0,0 +1,332 @@
...
+    expect(appJS.indexOf('js-addon-js-preprocessor-transform-self')).to.equal(-1);
+
+    // We have two hoisted templates and the application template.
+    var templatePreprocessorCount = appJS.split('template-addon-template-preprocessor-transform-parent').length - 1;
+    expect(templatePreprocessorCount).to.equal(3);
+
+    // There should be no `template` `self` preprocessor transforms.
+    expect(appJS.indexOf('template-addon-template-preprocessor-transform-self')).to.equal(-1);
+
+
```
rwjblue: extra whitespace

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+  this.name = name;
+  this._installedAddons = [];
+
+  this._init();
+}
+
+AppFixture.prototype = {
+
+  _init: function() {
+    process.chdir(root);
```
rwjblue: this seems pretty odd, I'm guessing its due to quickTemp API?  I'd prefer to simply avoid that and generate our own tmp dirs...

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+
+    fixturify.writeSync(this.dir, this.fixture);
+
+    var packageCache = new PackageCache(root);
+
+    var from, to;
+    if (this.fixture['package.json'] || npmLinks.length) {
+      var nodePackageCache;
+      if (isChild) {
+        process.env.NODE_ENV = 'production';
```
rwjblue: this is to avoid devDeps?

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+      '--skip-git'
+    );
+
+    this.fixture = fixturify.readSync(this.dir);
+
+    // Clean up after the generator.
+    fs.removeSync(this.dir);
+  },
+
+  serialize: function(isChild) {
```
rwjblue: isChild what?

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+      symlinkOrCopySync(from, to);
+    }
+
+    inRepoLinks.forEach(function(link) {
+      fs.mkdirsSync(path.dirname(link.from)); // Just in case the path doesn't exist.
+      fs.mkdirsSync(path.dirname(link.to)); // Just in case the path doesn't exist.
+      symlinkOrCopySync(link.to, link.from);
+    });
+
+    return this;
```
rwjblue: do we actually need/want to support chaining like this?

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+  },
+
+  clean: function() {
+    this._installedAddons.forEach(function(addon) {
+      addon.clean(true);
+    });
+
+    // Build up object to pass to quickTemp.
```
rwjblue: the code is simple, we should just steal what we want and avoid the oddness from quickTemp

tests/helpers/app-fixture.js
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    if (addon.type === 'in-repo-addon') {
+      return this._inRepoAddonInstall(addon);
+    }
+
+    throw new Error('Cannot install addon.');
+  },
+
+  _npmAddonInstall: function(addon) {
```
rwjblue: Mentioned above, but we should maybe change this to `addonInstall` and avoid "npm"

tests/acceptance/nested-preprocessor-test-slow.js
nathanhammond: [`project.name()` is not treated as the canonical value inside of `ember-app`.](https://github.com/ember-cli/ember-cli/blob/e64c27a651295af8cd1da907eb55fd8f35c0de6f/lib/broccoli/ember-app.js#L112) It's only used for `isDevelopingAddon` and inside of the constructor. I'd rather stop one short of the project and rely on it to tell me what name will eventually be used.

tests/acceptance/nested-preprocessor-test-slow.js
nathanhammond: Don't care, this is always the way I communicate it.

tests/acceptance/nested-preprocessor-test-slow.js
nathanhammond: I'm turning an actual honest-to-goodness function into a string, could rely on coercion or do it inside of `addMethod` but that makes it too magical. I consider this a test implementation detail.

tests/acceptance/nested-preprocessor-test-slow.js
nathanhammond: I'm working around the limitations of the API I have so far. `install` is used to approximate `ember install your-ember-addon` and delegates to underscored functions. You would pass it a fixture which is either an in-repo-addon, or a "full" addon. A "full" addon gets "installed" into the `node_modules` folder, the `_npmAddonInstall` method name is to distinguish from `_inRepoAddonInstall`

What I needed here was to modify the fixture to add a new dependency. I elected to leverage something that I already knew did it rather than add a new helper function, or do it manually in multiple locations because I'm not sure what that API should look like yet.

tests/acceptance/nested-preprocessor-test-slow.js
nathanhammond: Intentional to make the clusters stand out more.

tests/helpers/app-fixture.js
nathanhammond: https://github.com/joliss/node-quick-temp/pull/16

tests/helpers/app-fixture.js
nathanhammond: This is extended by `addon-fixture` which is extended by `in-repo-addon-fixture`. If any of these items has a `parent` it changes the way in which it does serialization.

tests/helpers/app-fixture.js
nathanhammond: Yep, no need to unnecessarily install libraries while running our test suite.

tests/helpers/app-fixture.js
nathanhammond: I want the DSL for working with these fixtures to feel expressive, chaining has a sense of composability that appeals to me and feels both "functional" and "imperative" at the same time in different ways. I'm not currently using this pattern but I want to keep playing with it and see if it's valuable.

tests/helpers/app-fixture.js
nathanhammond: `quickTemp` falls apart when you need a singular directory since its API is designed for making many, many directories. Trying to keep the directory creation and deletion logic wrapped in one bit of code. At this point I bet we're the sole user of that lib, I nominate fork and improve its API, not copypasta code.

package.json
nathanhammond: Fixed.

Turbo87:

PR #6452, opened 2016-11-22
Lazily install fs monitor

hjdivad:
Only install fs monitor if `process.env.BROCCOLI_VIZ`.  There's no reason to
monitor fs operations if we're not going to actually output them.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -12,14 +12,25 @@ var chalk       = require('chalk');
 var attemptNeverIndex = require('../utilities/attempt-never-index');
 var findBuildFile = require('../utilities/find-build-file');
 
-var FSMonitor = require('heimdalljs-fs-monitor');
-var monitor = new FSMonitor();
-monitor.start();
-
 var Sync = require('tree-sync');
 
 var signalsTrapped = false;
 var buildCount = 0;
+
+function _enableFSMonitorIfVizEnabled() {
+  var FSMonitor = require('heimdalljs-fs-monitor');
+  var monitor = new FSMonitor();
+  if (vizEnabled()) {
+    monitor.start();
+  }
+  return monitor;
+}
+
+_enableFSMonitorIfVizEnabled();
+
+function vizEnabled() {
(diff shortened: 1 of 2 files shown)


lib/models/builder.js
```diff
@@ -36,6 +36,12 @@ function outputViz(count, result) {
   });
 }
 
+if (vizEnabled()) {
```
stefanpenner: lets move this to a function we can unit test.

stefanpenner:
:shipit:

lib/models/builder.js
```diff
@@ -12,15 +12,26 @@ var chalk       = require('chalk');
...
-var monitor = new FSMonitor();
-monitor.start();
-
 var Sync = require('tree-sync');
 
 var signalsTrapped = false;
 var buildCount = 0;
 
+function _enableFSMonitorIfVizEnabled() {
+  var FSMonitor = require('heimdalljs-fs-monitor');
```
rwjblue: We should move these inside the guard also.  There is no need to require / instantiate FSMonitor instance if we are going to throw them away unused...

rwjblue: Hmm. OK.  Seems fine either way, just didn't seem like we needed to even require it if it wasn't needed.

rwjblue:
@homu r+

lib/models/builder.js
stefanpenner: @rwjblue ya, we started down that path. Then choose this trade-off. Not perfect, but reasonable balance maybe?

nathanhammond:
Getting reports that this PR may have broken CI. https://github.com/ember-cli/ember-cli/issues/6454#issuecomment-262401891

Once fixed we can land https://github.com/nodejs/citgm/pull/243

stefanpenner:
@nathanhammond let me know when you got a chance what/if we foobar'd

Turbo87:
one of the tests in this PR is indeed broken, see https://github.com/ember-cli/ember-cli/issues/6486

PR #6453, opened 2016-11-22
Avoid creating extraneous merge-trees.

rwjblue:
This change does three main things:

* Ensure that we always use `lib/broccoli/merge-trees.js` over requiring `broccoli-merge-trees` directly.
* When `lib/broccoli/merge-trees.js` is called with a single input tree, simply return it (instead of creating a `broccoli-merge-tree` instance).
* When `mergeTrees` is called with an empty array, use a single `EMPTY_MERGE_TREES` constant instance for all invocations.  This prevents us from creating many useless operations to essentially support an "empty" tree.  It is possible that we might want to teach Broccoli itself about empty trees, but that is left as a future excercise.

The results of these changes even in a newly created empty app are fairly significant. Here are the counts of `tmp` directories created before and after:

* 952 - Before changes
* 274 - After changes

In a real app, these numbers would be even higher (due to addons at each level of nesting).

---

Credit goes to @stefanpenner for the idea and initial implementation (which was stolen from https://github.com/ember-cli/ember-cli/pull/6429).

### DIFF
--- a/lib/broccoli/merge-trees.js
+++ b/lib/broccoli/merge-trees.js
@@ -1,13 +1,42 @@
 'use strict';
 
 var upstreamMergeTrees = require('broccoli-merge-trees');
+var EMPTY_MERGE_TREE;
 
-module.exports = function(inputTree, options) {
-  options = options || {};
-  options.description = options.annotation;
-  var tree = upstreamMergeTrees(inputTree, options);
+function overrideEmptyTree(tree) {
+  EMPTY_MERGE_TREE = tree;
+}
 
-  tree.description = options && options.description;
+function getEmptyTree() {
+  if (EMPTY_MERGE_TREE) {
+    return EMPTY_MERGE_TREE;
+  }
 
-  return tree;
+  return EMPTY_MERGE_TREE = upstreamMergeTrees([], {
+    annotation: 'EMPTY_MERGE_TREE',
+    description: 'EMPTY_MERGE_TREE'
+  });
+}
+
+module.exports = function mergeTrees(_inputTrees, options) {
+  var inputTrees = _inputTrees.filter(function(input) {
+    return input && input !== getEmptyTree();
+  });
+
+  switch (inputTrees.length) {
+  case 0:
+    return getEmptyTree();
+  case 1:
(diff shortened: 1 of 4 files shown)


stefanpenner:
nice!

@rwjblue may complement this nicely https://github.com/broccolijs/broccoli-merge-trees/pull/43

lib/broccoli/merge-trees.js
```diff
@@ -1,13 +1,26 @@
...
+  annotation: 'EMPTY_MERGE_TREE',
+  description: 'EMPTY_MERGE_TREE'
+});
 
-module.exports = function(inputTree, options) {
-  options = options || {};
-  options.description = options.annotation;
-  var tree = upstreamMergeTrees(inputTree, options);
+module.exports = function mergeTrees(_inputTrees, options) {
+  var inputTrees = _inputTrees.filter(Boolean);
```
stefanpenner: we may also want to filter out other EmptyMergeTrees here.

lib/broccoli/merge-trees.js
```diff
@@ -1,13 +1,26 @@
...
+var EMPTY_MERGE_TREE = upstreamMergeTrees([], {
+  annotation: 'EMPTY_MERGE_TREE',
+  description: 'EMPTY_MERGE_TREE'
+});
 
-module.exports = function(inputTree, options) {
-  options = options || {};
-  options.description = options.annotation;
-  var tree = upstreamMergeTrees(inputTree, options);
+module.exports = function mergeTrees(_inputTrees, options) {
```
stefanpenner: we should likely test this guy.

rwjblue:
@homu r+

rwjblue:
@homu r-

rwjblue:
@stefanpenner - Updated with tests and better filtering when input array contains the `EMPTY_MERGE_TREE` constant (this reduced the number of `tmp` dirs further down to 274).

rwjblue:
@homu r+

rwjblue:
Tested this in Ghost, the tmp dirs created during `ember s` went from 3444 to 966.

PR #6460, opened 2016-11-23
Refactor processTemplate.

nathanhammond:
Simple change to unlock broader usage of this function.

### DIFF
--- a/lib/utilities/process-template.js
+++ b/lib/utilities/process-template.js
@@ -1,0 +1,10 @@
+var _template = require('lodash.template');
+
+module.exports = function processTemplate(content, context) {
+  var options = {
+    evaluate:    /<%([\s\S]+?)%>/g,
+    interpolate: /<%=([\s\S]+?)%>/g,
+    escape:      /<%-([\s\S]+?)%>/g
+  };
+  return _template(content, options)(context);
+}
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -10,15 +10,6 @@ var EOL          = require('os').EOL;
 var isBinaryFile = require('isbinaryfile').sync;
 var canEdit      = require('../utilities/open-editor').canEdit;
 
-function processTemplate(content, context) {
-  var options = {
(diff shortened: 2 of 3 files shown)


Turbo87:
this changes the lazy to eager require() though. was that intended?

nathanhammond:
The lazy require in this case I'm _pretty sure_ didn't net us much because I believe that code path gets hit nearly every time. I [originally had it still lazy](https://github.com/ember-cli/ember-cli/pull/6449/commits/5d2b1b16f80beb2e7712a4339e4cc53431a338f2) but I elected against that.

Don't have strong opinions.

Turbo87:
> I believe that code path gets hit nearly every time

does `ember help` render templates?

stefanpenner:
seems good, mind adding 1 quick unit test covering this function (now that it is extracted). Then :shipit:

kellyselden:
> does ember help render templates?

I'm also curious if this regresses `ember help` speed.

stefanpenner:
> I'm also curious if this regresses ember help speed.

@kellyselden say more. The above PR looks like a extraction, I wouldn't expect a difference.

kellyselden:
@stefanpenner I believe moving the `require`s to lazily loaded instead of eagerly was to let `ember h` and `ember v` run faster. At first glance, this appears to revert some of that.

nathanhammond:
1. Moved back to lazy `require`.
2. Added a test.

Ready to ship.

lib/utilities/process-template.js
```diff
@@ -0,0 +1,10 @@
+var _template = require('lodash.template');
```
Turbo87: @nathanhammond with lazy require I meant moving the above line into the `processTemplate()` function, instead of the more complicated require() above. this solution will work too though. :shrug:

nathanhammond: Eh, by the time you've required the `processTemplate` helper you're gonna need it. 😜

I really truly don't care about this, so somebody tell me what to do! (Too many people with opinions.)

Turbo87:
@homu r+

PR #6461, opened 2016-11-23
Package Cache

nathanhammond:
I'm piecing out a lot of the work I've been doing for the CSS testing stuff. This is _just_ the improved package caching layer.

- [x] Adopt `execa`.
- [x] Wire it into the existing acceptance tests.
- [x] Add Node.js version and npm version to the cache hit calculation.
- [x] Auto-downgrade to `npm` on 0.12.
- [x] Lock out upgrade calls to once-per-process.
- [x] Only compare things which have a bearing on the installed packages.
- [x] Tests. (Who will test the test helpers?)
- [x] Land https://github.com/ember-cli/ember-cli/pull/6460.
- [x] Correctly handle linking Ember CLI.
- [x] Create API for dynamic package cache creation.
- [x] Documentation.


nathanhammond:
# Cache performance.
SSD MBP 13", Gigabit 8ms ping network.

## All Caches Nonexistent
Likely distorted by raw network performance at test locale.

```
bower cache clean
yarn cache clean
rm -rf tmp && mkdir tmp && rm -f ~/.config/configstore/package-cache.json
time node tests/helpers/package-cache.js
```
```
real	1m6.746s
user	0m53.145s
sys	0m26.788s
```

## Package manager caches present.

```
rm -rf tmp && mkdir tmp && rm -f ~/.config/configstore/package-cache.json
time node tests/helpers/package-cache.js
```
```
real	0m54.868s
user	0m40.457s
sys	0m18.402s
```

## Ember CLI Package Cache turned on with package manager caches present.

```
time node tests/helpers/package-cache.js
```
```
real	0m26.714s
user	0m22.858s
sys	0m4.904s
```

package.json
```diff
@@ -115,7 +115,8 @@
     "tiny-lr": "^1.0.3",
     "tree-sync": "^1.1.4",
     "walk-sync": "^0.3.0",
-    "yam": "0.0.22"
+    "yam": "0.0.22",
+    "yarn": "^0.17.8"
```
rwjblue: I don't think this can be a dep, right?

nathanhammond: It definitely works... not sure why it can't be?

nathanhammond: I can move it to `devDependencies` to avoid shipping it unnecessarily. We want to use it because it saves us a ton of time in our test suite, I'll set it up to fall back to `npm` for older versions of Node. (Which already works.) That meets our Node LTS commitment.

But, I just realized that we can use this same approach to make `ember new/addon/init` faster for our users. So I may leave it in `dependencies` and wire it into those as well.

locks: @nathanhammond can we move it to `devDependencies`, ship, then iterate on wiring it into `ember new/addon/init`?

nathanhammond: Regarding "iterate" ... it appears to be remarkably simple to do, Ember CLI is well factored for that change.

I'm more likely to be blocked on the testing side than I am the user-facing side.

nathanhammond:
Updated numbers for one real test, not a synthetic benchmark:

`it.only('generates package.json and bower.json with proper metadata')`

### Existing Strategy
```
$ time npm run test-slow
real	1m31.394s
user	1m4.133s
sys	0m31.892s
```

### Package Cache First Run (Yarn)
```
$ time npm run test-slow
real	0m40.900s
user	0m31.812s
sys	0m11.833s
```

### Package Cache Second Run (Yarn)
```
$ time npm run test-slow
real	0m26.463s
user	0m23.912s
sys	0m3.235s
```

**Test suite will run at least one minute faster.**

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,385 @@
...
+ * @param {String} [...arguments] Arguments to be passed into the yarn subcommand.
+ * @param {Object} [options={}] The options passed into child_process.execSync.
+ *   (https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options)
+ */
+var yarn = new CommandGenerator(require.resolve('yarn/bin/yarn.js'), {
+  networkCommands: ['install', 'upgrade']
+}).invoke;
+
+// This lookup exists to make it possible to look the commands up based upon context.
+var originals;
```
stefanpenner: it would be abit less risky, if this was a state on `PackageCache`. That way, a random test failure wont pollute future test runs. (Or if we invoke programmatically.)

lib/utilities/command-generator.js
```diff
@@ -0,0 +1,89 @@
+var execSync = require('child_process').execSync;
+var merge = require('ember-cli-lodash-subset').merge;
+
+/**
+ * A simple tool to make behavior consistent between the package manager commands.
```
stefanpenner: It would be useful to describe why and when one would use this class.

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,385 @@
...
+ * @method translate
+ * @param {String} type Either 'bower', 'npm', or 'yarn'.
+ * @param {String} lookup Either 'manifest', 'path', or 'upgrade'.
+ */
+function translate(type, lookup) { return lookups[lookup][type]; }
+
+
+/**
+ * The PackageCache wraps all package management functions. It also
+ * handles initial global state setup.
```
stefanpenner: It would be useful to describe why and when one would use this class.

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,385 @@
...
+/**
+ * The `bower` command helper.
+ *
+ * @method bower
+ * @param {String} subcommand The subcommand to be passed into bower.
+ * @param {String} [...arguments] Arguments to be passed into the bower subcommand.
+ * @param {Object} [options={}] The options passed into child_process.execSync.
+ *   (https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options)
+ */
+var bower = new CommandGenerator(require.resolve('bower/bin/bower'), {
```
stefanpenner: Generator made me think of blueprints + generation, thoughts on renaming this to `CommandBuilder` or `CommandFactory` or ..

nathanhammond:
Okay, so, imagine if you had a cache with an API that guaranteed you would always get the behavior of `rm -rf node_modules && npm install` each time. But did it as performantly as possible by leveraging a persistent backing cache. That's what this is. For bonus points you can pretend like it's not a cache and use it idempotently:

```
var cache = new PackageCache();
var dir1 = cache.create('cacheName', 'yarn', '{ "dependencies": ... }');
var dir2 = cache.create('cacheName', 'yarn', '{ "dependencies": ... }');

dir1 === dir2;
// => true
```

This makes it possible to assign a name to a cache and pull it back out. If the dependencies change it will rebuild that cache by invoking the `upgrade` command, minimizing the amount of effort required to make the change.

Making it a persistent cache allows for `ember new` (and `addon`, and possibly `init`) to be optimized:
- First run (leveraging `yarn`) drops from 90 seconds to 40 seconds. All performance win is from `yarn`.
- Second run (leveraging `yarn upgrade`) drops from 40 seconds to 26 seconds by pulling the last installation run (stored at a persistent cache named `ember-new-<version-number>`) out of cache and then running the faster `upgrade` command. This will opt you into SemVer drift.
- For the the full offline experience, `ember new --cached-dependencies` will take about **3 seconds** as it won't attempt `yarn upgrade` and will instead simply copy the files over. The only cost is no SemVer drift.

---

Next use case, tests. We currently have this abstract acceptance test helper which creates a single cache which we use for _all_ acceptance tests. This has numerous issues, not the least of which is that it is relatively magical.

In order to make it easier to test nested addons and build behavior I want to define a DSL around creating `App` and `Addon` fixtures so that they don't have to be hardcoded on disk, don't have to use the `copyFixtureFiles` function, and can spend less time doing disk IO. This DSL will also make it possible to specify a `PackageCache` and use that as the backing set of of packages for that addon. This will make it easy to build separate caches for both major types, `clone`, and otherwise modify them in a cheap manner. This is something which is incredibly expensive and slow in our current testing story. These fixture helpers will be composable and have a `serialize` hook which will automatically symlink in a specified `PackageCache` and write the assets to disk all at once.

This will make it easier to prevent test leakage (@trabus' problem) as well as prevent ordering problems since it will be much simpler to isolate.

For reasons of needing to support SemVer drift we can't move to full `--cached` only for CI, but we can still do it for local test runs to speed them up. (90 second win on test run time as well as making them wholly offline-capable after a single online run.)

---

So the wins:
- Performance of tests and installation.
- Offline installation support for the first time ever.
- A wrapper around `yarn`/`npm` so that we can begin adoption.
- Clean API which is useful as a foundation for the [next things I want to build](https://github.com/ember-cli/ember-cli/pull/6449/files#diff-3267d52422c53ab25f9eaf094b15a722).

stefanpenner:
So it is improved way of handled `npm/yarn` dependencies during test runs? Sounds good, I'm curious what deficiencies yarn has that prevent it from requiring this (although the compat layer for npm makes sense).


Questions for ConfigStore usage. is the data being stored intended as long-term state? How much data is usually stored. Is it more as a semi-persistent cache. Should a restart prune it out, should it be pruned if inactive for several days?

tests/unit/utilities/command-generator-test.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('gets clever on ci', function() {
+    var original;
+    var yarn = new Command('yarn', { networkCommands: ['install'] });
+
+    original = process.env.TRAVIS;
+    process.env.TRAVIS = 'true';
+    expect(yarn.ci('install')).to.deep.equal(['travis_retry']);
+    process.env.TRAVIS = original;
```
Turbo87: we should put those in `afterEach()` blocks to make sure the original is restored even if the assertion fails.

```js
describe('in CI environments', function() {
  var originalTravis, originalAppVeyor;

  beforeEach(function() {
    originalTravis = process.env.TRAVIS;
    originalAppVeyor = process.env.APPVEYOR;
  });

  afterEach(function() {
    process.env.TRAVIS = originalTravis;
    process.env.APPVEYOR = originalAppVeyor;
  });

  it ('gets clever on TravisCI', function() {
    process.env.TRAVIS = 'true';

    var yarn = new Command('yarn', { networkCommands: ['install'] });
    expect(yarn.ci('install')).to.deep.equal(['travis_retry']);
  });

  it ('gets clever on AppVeyor', function() {
    process.env.APPVEYOR = 'True';

    var yarn = new Command('yarn', { networkCommands: ['install'] });
    expect(yarn.ci('install')).to.deep.equal(['appveyor-retry']);
  });
});
```

lib/utilities/package-cache.js
nathanhammond: Resolved.

lib/utilities/command-generator.js
```diff
@@ -0,0 +1,89 @@
...
+ * @class CommandGenerator
+ * @constructor
+ * @param {Path} program The path to the command.
+ * @param {Object} options The options for the command invocation.
+ * @return {Function} A command helper.
+ */
+function CommandGenerator(program, options) {
+  this.program = program;
+  this.options = merge({
+    networkCommands: []
```
nathanhammond: Agreed. Changed.

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,385 @@
...
+/**
+ * The `npm` command helper.
+ *
+ * @method npm
+ * @param {String} subcommand The subcommand to be passed into npm.
+ * @param {String} [...arguments] Arguments to be passed into the npm subcommand.
+ * @param {Object} [options={}] The options passed into child_process.execSync.
+ *   (https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options)
+ */
+var npm = new CommandGenerator(require.resolve('npm/bin/npm-cli.js'), {
```
nathanhammond: Doesn't matter, that's our internal dependency version of `npm`. If it changes we'll break in CI and that will catch it.

lib/utilities/command-generator.js
nathanhammond: Resolved.

lib/utilities/package-cache.js
nathanhammond: I was trying hard to make the original implementations not leak, thus putting it in the closure only. This guarantees that even if we `throw` that it will reset:

```
afterEach(function() {
  testPackageCache.__resetForTesting();
});
```

If somebody accidentally breaks something outside of tests by calling the double-underscored, wholly undocumented, obviously intended for tests, `testPackageCache.__setupForTesting()` it's their own damn problem. 😜

lib/utilities/package-cache.js
nathanhammond: I try really hard not to care about names. Happy to bikeshed every single identifier in this PR but we should do so via real-time chat (video, in-person, Slack).

tests/unit/utilities/command-generator-test.js
nathanhammond: Directly copy/pasted. Much nicer than my pattern.

nathanhammond:
> So it is improved way of handled npm/yarn dependencies during test runs?
At its simplest, yes. It can also be used in lots of clever ways, see the documentation from my most-recent commit.

> I'm curious what deficiencies yarn has that prevent it from requiring this.
There aren't deficiencies in `yarn` here, there's nothing that `yarn` can do to resolve and link something faster than I can compare a cache key and copy over a directory's worth of content. I don't have to do linking, resolution, anything. This sort of behavior has to (by default) be one layer up from the package manager itself. Only we can really know when it is or isn't safe to use a cache. /cc @thejameskyle @kittens

> Is the data being stored intended as long-term state? How much data is usually stored. Is it more as a semi-persistent cache. Should a restart prune it out, should it be pruned if inactive for several days?

`PackageCache` + `Configstore` is actually a user-space concern of this tool. For example, in testing, we will never want the persistent cache to be deleted since `PackageCache` provides guarantees against stale cache reads. We don't particularly care about size or disk usage on CI. Further, people running Ember CLI tests locally will also likely not be concerned about using some disk to make tests run significantly faster. (I'll make that trade every time.)

Once wired into user-facing code such as `ember new` or `ember addon` we'll have consistent cache names for that as well:
- `ember-new-node`
- `ember-new-bower`
- `ember-addon-node`
- `ember-addon-bower`

Since we only use one set of caches it prevents these from growing unbounded. So, once set up properly, after the first run the size will be wholly consistent forever at whatever the size of our dependencies are.

nathanhammond:
I've pushed [a demonstration of the sort of API cleanup we can do for the existing generic acceptance test helper](https://github.com/ember-cli/ember-cli/pull/6461/commits/411abf9674a51b3b6880820a893df22200075dc3) which also hints at how we can make it easier to create fixtures due to the number of things we can simply avoid caring about which are automagically handled by `PackageCache`.

stefanpenner:
> There aren't deficiencies in yarn here, there's nothing that yarn can do to resolve and link something faster than I can compare a cache key and copy over a directory's worth of content.

This does seem like something yarn could do, but that doesn't mean we couldn't do it here.

> PackageCache + Configstore is actually a user-space concern of this tool. For example, in testing, we will never want the persistent cache to be deleted since PackageCache provides guarantees against stale cache reads. We don't particularly care about size or disk usage on CI. Further, people running Ember CLI tests locally will also likely not be concerned about using some disk to make tests run significantly faster. (I'll make that trade every time.)

As a maintainer, I think I would care about stale data remaining on disk that will never be cleaned up. An alternative is https://github.com/stefanpenner/sync-disk-cache or https://github.com/stefanpenner/async-disk-cache which store the data compressed in system tmp. restarts (or inactivity) after 3 days (OS specific) will evict the cache. This continues the existing guarantees, but minimizes issues they may arise. (We already use ^ for other persistent data, so it should be safe to use here.

tests/helpers/acceptance.js
```diff
@@ -141,29 +114,25 @@ function linkDependencies(projectName) {
   return tmp.setup(targetPath).then(function() {
     return copy('./common-tmp/' + projectName, targetPath);
   }).then(function() {
```
stefanpenner: looks much better

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,480 @@
...
+    if (this.options.linkEmberCLI === true && (type === 'npm' || type === 'yarn')) {
+      var emberCLIPath = path.resolve(__dirname, '../..');
+      commands[type].invoke('link', { cwd: emberCLIPath });
+    }
+
+    // Compare any existing manifest to the ideal per current blueprint.
+    var identical = this._checkManifest(label, type, manifest);
+
+    if (identical) {
+      // Use what we have, but opt in to SemVer drift.
```
MiguelMadero: later: could this be a configuration at least for yarn? One of the benefits of yarn is that we don't  semver drift unless we explicitly upgrade locally and commit a new yarn.lock. Here it seems like CI would ignore my yarn.lock and always SemVer drift regardless of what I had locally, leading to different dependencies being used. This sounds normal for npm, but I feel it changes the expected behavior of yarn. Even with npm, if we had shrinkwrap present, I would expect that creating a new cache wouldn't semver drift.

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,480 @@
...
+ *
+ * ```
+ * var dir3 = cache.update('your-cache', 'yarn', '{
+ *   "dependencies": {
+ *     "": "*",
+ *     "lodash": "*",
+ *     "ember-cli": "*"
+ *   }
+ * }');
+ * // => /somewhere/tmp/your-cache-A3B4C6D
```
MiguelMadero: Where is `A3B4C6D` coming from?

MiguelMadero:
@nathanhammond great stuff. I'm really happy to see this in CLI. 

It is a little bit unclear where the cache is stored, is it on the file system on a tmp dir that then gets cached/reused by CI on a system like travis? 

We have something similar, but with two layers, one is the filesystem, but we also compress it and push it to s3 after `npm i` and then pull them based on the shasum if it's not found locally. This allows us to re-use this cache across containers and branches. 

Travis has good support for specifying `cache.directories`, but other CI system don't have that. Also, for what I understand, Travis does this per branch, so even for travis it may make sense to have a cache that is easier to share based on the manifest as opposed (or in addition) to do it based on a folder per branch. 
This can be an iteration on this, but it may be important to have hooks to share this cache. For large apps the numbers are much worst than what we have here and the savings of a shared cache are significant to justify optimizing the cache hits.

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -50,7 +50,7 @@ describe('Acceptance: addon-smoke-test', function() {
     });
   });
 
-  it('generates package.json and bower.json with proper metadata', function() {
+  it.only('generates package.json and bower.json with proper metadata', function() {
```
MiguelMadero: minor: remove `only`

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,480 @@
...
+    var parsedCached = JSON.parse(cachedManifest);
+    var parsedNew = JSON.parse(manifest);
+
+    // Only inspect the keys we care about.
+    var keys = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies'];
+
+    var key;
+    for (var i = 0; i < keys.length; i++) {
+      key = keys[i];
+      if (JSON.stringify(parsedCached[key]) !== JSON.stringify(parsedNew[key])) {
```
rwjblue: I believe this should use `json-stable-stringify`

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,480 @@
...
+  bower: bower,
+  npm: npm,
+  yarn: yarn
+};
+
+// The definition list of translation terms.
+var lookups = {
+  manifest: {
+    bower: 'bower.json',
+    npm: 'package.json',
```
nathanhammond: No, see my next (as-of-yet-unwritten) comment. 😜

lib/utilities/package-cache.js
nathanhammond: It's a randomly-generated string by `quick-temp`.

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,480 @@
...
+    var parsedNew = JSON.parse(manifest);
+
+    // Only inspect the keys we care about.
+    var keys = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies'];
+
+    var key;
+    for (var i = 0; i < keys.length; i++) {
+      key = keys[i];
+      if (JSON.stringify(parsedCached[key]) !== JSON.stringify(parsedNew[key])) {
+        return false;
```
nathanhammond: My tests for the fixtures were unnecessarily invalidating because of this and it was really frustrating. So, solved. 😄

lib/utilities/package-cache.js
nathanhammond: We rely on SemVer drift to keep Ember CLI up-to-date. The default behavior for libraries is _not_ to commit the lockfile. Note that we only use `PackageCache` for brand new apps or addons (no lockfile) and for Ember CLI tests (where we need to ensure that SemVer drift doesn't break us).

nathanhammond:
> It is a little bit unclear where the cache is stored, is it on the file system on a tmp dir that then gets cached/reused by CI on a system like travis?

I haven't 100% sorted this out yet. 😛 Currently I'm generating folders like `ember-cli/tmp/cache-name-RANDOMSTRING` The end state is _not_ achieved until these are cached across runs for our CI boxes. If you're familiar with this I'd very much welcome assistance on that effort.

nathanhammond:
> I'm unsure how this wont grow unbounded, as users upgrade versions of CLI or run multiple versions. Or as transitives change over time. Or if a user toggles between versions, that things will be constantly evicted. I may be missing something, so share as you have time. - @stefanpenner

`PackageCache` stores and reuses caches by name which means that we cannot grow the cache larger than the number of named caches. Ember CLI on the user-facing side will only use four names:
- `app-bower`
- `app-node`
- `addon-bower`
- `addon-node`

These caches will be generated and used for `ember addon` and `ember new` only, which means that they will always map to the manifests from the global version of `ember-cli`. Even in the less-than-ideal scenario of a user switching their global `ember-cli` version frequently (e.g. `nvm use`) we'll still get a perf win by being able to call `yarn upgrade` on the cache instead of `yarn install` with a new manifest (40s for `install`, 26s for `upgrade` = 14s win).

The test behavior will be the same. Once we create the extent of our named caches (first run) it will cease to grow and instead continue to reuse the existing caches.

By virtue of this being `yarn` the caches literally never go stale and will _always_ provide some performance benefit.

lib/utilities/package-cache.js
MiguelMadero: > We rely on SemVer drift to keep Ember CLI up-to-date. The default behavior for libraries is not to commit the lockfile.

That makes sense as a general rule. I'm probably missing some context. I was under the impression that the PackageCache will be used for apps as well, in which case, I think the default behavior is to commit the lockfile and avoid SemVer drift. Is the plan to use this for apps as well?

MiguelMadero:
> If you're familiar with this I'd very much welcome assistance on that effort.

We use different systems. For [circle](https://circleci.com/docs/how-cache-works/), [appveyor](https://www.appveyor.com/docs/build-cache/#saveupdate-cache-before-build-finishes) and [travis](https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F) do pretty much the same thing. We also use buildkite, whic AFAIK doesn't have anything equivalent. Travis, appveyor and circle will upload the contents of whatever folder you specify to s3 and then download it on the next run. Not sure if AppVeyor uses s3, but it has something equivalent, it seems a bit smarter about cache invalidation based on changes (e.g. if package.json changed it can invalidate one folder). Travis and Circle have a separate cache per branch with master and *any* as a fallback, I'm not sure about AppVeyor, but I believe it's a single cache. 

The default caching behavior from AppVeyor and Travis should be enough for something like ` tests/acceptance/addon-smoke-test-slow.js`. I was thinking of a few other use cases where we have this for apps, but I feel like that's not the first target, so we can probably discuss those after this lands. 

For our apps, In order to increase cache-hits, instead of relying on a single per-branch cached folder, we push to s3 something like `host/{app-name}-{node|bower}-{manifest-shasum}.tgz`. Then we check in our local network if we have (saving a little bit form the s3 overhead), if not, we check s3, if not we install it. If we pulled it from cache, we don't `npm install` at all since we know it's accurate. That's similar to what you have with --cached-dependencies or just opting-out of SemVer drift.

lib/utilities/package-cache.js
ejsundstr: On a related note it would be useful to have the create function not take a single manifest string, but also work with an array of key generating strings.

Ala @MiguelMadero one might want to reuse this set of code for consideration of package.json + npm-shrinkwrap.json and asking the caller to concat "independent" manifest in the same way each time doesn't seem ideal.

I should mention, in my case, which also happens to be @MiguelMadero 's, we use package.json + npm-shrinkwrap.json + `node --version` + `npm --version` . We did something like this in the bash outside of ember to make our test suite viable.

nathanhammond: @MiguelMadero `PackageCache` is not yet intended to be used by the apps themselves. Once it gets extracted from Ember CLI, likely worth exploring how it could be leveraged by apps.

nathanhammond: @ejsundstr You make a good point that there may need to be additional cache key material. Using a cache across Node.js versions blows up the first time you have a native extension. Using a cache across `npm` versions _probably_ won't break you, but might. I'm going to abstract those two pieces directly into `PackageCache` so that the user doesn't have to think about it.

nathanhammond: (Still _mildly_ opposed to considering the shrinkwrap file as a cache key. This is designed to safely implement SemVer drift and clobber caches of the same name with the new SemVer-drifted versions. It is not in any way designed to lock things down.)

stefanpenner:
> These caches will be generated and used for ember addon and ember new only, which means that they will always map to the manifests from the global version of ember-cli. Even in the less-than-ideal scenario of a user switching their global ember-cli version frequently (e.g. nvm use) we'll still get a perf win by being able to call yarn upgrade on the cache instead of yarn install with a new manifest (40s for install, 26s for upgrade = 14s win).

Given:

* 2 apps
* one with local ember-cli 2.8
* one with local ember-cli 2.9
* global ember-cli is 2.10

How does the cache partitioned to avoid pollution between versions?

> Once we create the extent of our named caches (first run) it will cease to grow and instead continue to reuse the existing caches.

So one we have primed the cache, we can no longer add to it? Otherwise, changing versions would add to the cache, but we would never know to evict stale packages. Resulting in cache growth.


@MiguelMadero If your speaking of caching node_modules wholesale. We used to do that here, but it proved far to unstable/brittle. Instead we moved to preserving only `npm` system cache between CI runs. https://github.com/ember-cli/ember-cli/blob/master/.travis.yml#L20

> I haven't 100% sorted this out yet. 😛 Currently I'm generating folders like ember-cli/tmp/cache-name-RANDOMSTRING 

* npm's cache is already persistent between runs:  https://github.com/ember-cli/ember-cli/blob/master/.travis.yml#L20
* rather then creating  another way to cache files, can we use github.com/stefanpenner/sync-disk-cache or github.com/stefanpenner/async-disk-cache as those are both thoroughy tested/vetted If they are inteded to survive 3days lack of access, or restarts the location can be changed: https://github.com/stefanpenner/sync-disk-cache/blob/master/index.js#L84

> If the dependencies change it will rebuild that cache by invoking the upgrade command, minimizing the amount of effort required to make the change.

Does this mean we will no-longer get transitive dependency updates once we have a warm/persistent cache?

nathanhammond:
```
Question, Given:

    2 apps
    one with local ember-cli 2.8
    one with local ember-cli 2.9
    global ember-cli is 2.10

How does the cache partition to avoid pollution (or thrashing) when using any 1
of the 3 ember-cli versions on a given system?
```

Because `ember new` and `ember addon` are always executed with the global Ember CLI and those are the only two commands which will leverage `PackageCache`. Thus the cache is only ever used in one of your three scenarios. There may be a possibility to improve `ember init`, but that has to solve the problem you're thinking about, and I'm punting on that as I don't believe the return to be anywhere near as high and the complexity increases tremendously.

> So one we have primed the cache, we can no longer add to it only replace entirely? Otherwise, changing versions would add to the cache, but we would never know to evict stale packages. Resulting in cache growth. Or we when toggling between versions, we would constantly evict?

On each run we will always guarantee that you have a fully up-to-date cache based upon whatever your global Ember CLI version is. If you change versions often you'll not degrade worse than the performance you get presently: clean installation. If you are most people not on the @ember-cli team you'll get tremendous performance wins.

stefanpenner:
> Because ember new and ember addon are always executed with the global Ember CLI and those are the only two commands which will leverage PackageCache

ah ok this makes sense, thanks

> On each run we will always guarantee that you have a fully up-to-date cache based upon whatever your global Ember CLI version is.

Do we also ensure transitives dependencies are fully up to date?

> If you change versions often you'll not degrade worse than the performance you get presently: clean installation. If you are most people not on the @ember-cli team you'll get tremendous performance wins.

This sounds promising, but is it possible for us to decouple this particular goal and instead circle back later and then rexplore/implement it? That way, we can focus on that required scrutiny (via RFC or something) at a later time, and not block the test suite improvements on that level of scrutiny.

nathanhammond:
> Do we also ensure transitives dependencies are fully up to date?

Yep!

> This sounds promising, but is it possible for us to decouple this particular goal and instead circle back later and then rexplore/implement it? That way, we can focus on that required scrutiny (via RFC or something) at a later time, and not block the test suite improvements on that level of scrutiny.

Created an optional section in the top-post for this issue which includes `ember addon/new` inclusion.

chadhietala:
Couple things:

1. What is the purpose of this? I just see a checklist.
2. Where is the RFC to answer number 1

I've looked at this several times and still don't really understand what is going on here. Is this putting an additional cache in from of a package manager's cache?

nathanhammond:
> I've looked at this several times and still don't really understand what is going on here. Is this putting an additional cache in front of a package manager's cache?

Yes. It caches fully-resolved and linked versions. We already do this during our testing runs, that's exactly what this is: https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/acceptance.js

...except that it's nigh impossible to reason about and wholly untested.

---

Snagged out of https://github.com/ember-cli/ember-cli/pull/6461#issuecomment-263105123

So the wins:
- Performance of tests and installation.
- More stable CI behavior (because it's coded it in to `CommandGenerator`).
- Offline installation support for the first time ever.
- A wrapper around `yarn`/`npm` so that we can begin adoption of `yarn`.
- Clean API which is useful as a foundation for the [next things I want to build](https://github.com/ember-cli/ember-cli/pull/6449/files#diff-3267d52422c53ab25f9eaf094b15a722).

---

I don't consider an RFC to be required for testing-only improvements.

Turbo87:
@nathanhammond I'm not sure if this was discussed here already, but our use of `bower install` currently has a certain level of interactivity build into it when resolving version conflicts. I'm not sure how well this will play together with these changes here when plugged into the `BowerInstall` task.

nathanhammond:
`PackageCache` guarantees installed `node_modules` cache coherency for each test run, [proven](https://travis-ci.org/ember-cli/ember-cli/builds/182449482) by cherry-picking the commit from https://github.com/ember-cli/ember-cli/pull/5729 which is one of our oldest open PRs that was previously broken by incoherent caches.

This test run demonstrates that–even when using a single cache–we guarantee coherency (at the cost of time). Because we're still using a single cache (which we rewrite multiple times) test run times regressed, but I'm going to fix that next, as well as move our CI to `yarn`.

nathanhammond:
Script for local reproduction of Travis failure (run from inside the ember-cli repo directory):

```
rm -rf ~/.yarn
nvm use 6
nvm uninstall 7
nvm install 7
nvm alias default 7
rm -f ~/.config/configstore/package-cache.json
curl -o- -L https://yarnpkg.com/install.sh | bash
yarn global add phantomjs-prebuilt
yarn global remove phantomjs-prebuilt
yarn global add phantomjs-prebuilt
git clean -fdx
yarn install
yarn run test-all:cover
```

lib/utilities/package-cache.js
```diff
@@ -0,0 +1,524 @@
...
+  this.options = merge({
+    linkEmberCLI: false
+  }, options);
+
+  // Wrap the Configstore in a pretty interface.
+  Object.defineProperty(this, 'dirs', {
+    get: function() {
+      return this._conf.all;
+    },
+    set: function() {
```
nathanhammond: I went with this to make the error a bit more useful, don't feel too strongly though. Removed.

tests/unit/utilities/package-cache-test.js
```diff
@@ -0,0 +1,384 @@
...
+
+  describe('_upgrade', function() {
+    // We're only going to test the invocation pattern boundary.
+    // Don't want to wait for the install to execute.
+    var testCounter = 0;
+    var label;
+
+    // This accounts for the downgrade ability.
+    var command;
+    if (process.version.indexOf('v0.12') === 0) {
```
nathanhammond: This doesn't play nicely in a loop because each package manager is slightly different. It ends up becoming brittle. Split into three separate tests.

lib/utilities/command-generator.js
```diff
@@ -0,0 +1,51 @@
...
+
+    this._invoke(args, options);
+  },
+
+  _invoke: function(args, options) {
+    execa.sync(this.program, args, options);
+  }
+};
+
+module.exports = CommandGenerator;
```
nathanhammond: Done.

tests/helpers/acceptance.js
```diff
@@ -138,38 +77,33 @@ function teardownTestTargets() {
...
+  symlinkOrCopySync(path.join(packageCache.get('node'), 'node_modules'), nodeModulesPath);
 
-    // Need to junction on windows since we likely don't have persmission to symlink
-    // 3rd arg is ignored on systems other than windows
-    fs.symlinkSync(path.resolve('../..'), appsECLIPath, 'junction');
+  process.chdir(runFixture);
 
-    return targetPath;
-  });
+  return Promise.resolve(runFixture);
```
nathanhammond: Since I don't expect the future state version of this to be `async` I've moved away from promises. Keeping it as a promise was just a consequence of "it says it returns a promise!" and not thinking about it.

tests/helpers/acceptance.js
```diff
@@ -138,38 +77,33 @@ function teardownTestTargets() {
...
  * Clean a test run and optionally assert.
  * @return {Promise}
  */
 function cleanupRun(projectName) {
-  var targetPath = path.join(root, 'tmp', projectName);
-  return tmp.teardown(targetPath);
+  process.chdir(root);
+  quickTemp.remove(dirs, projectName + '-clone');
+
+  return Promise.resolve();
```
nathanhammond: Addressed.

stefanpenner:
Just noticed the following files appear dirty after a test run.

```sh
blueprints/addon/files/bower.json
blueprints/addon/files/package.json
```

I'm going to quickly check if it is preexisting. (Thought I would mention it in-case it rings a bell)

nathanhammond:
@stefanpenner Is pre-existing.

stefanpenner:
> @stefanpenner Is pre-existing.

tracking issue: https://github.com/ember-cli/ember-cli/issues/6560 for us to eventually fix

tests/unit/utilities/command-generator-test.js
```diff
@@ -0,0 +1,62 @@
...
+
+var expect = require('chai').expect;
+var Command = require('../../../lib/utilities/command-generator');
+
+describe('command-generator', function() {
+  it('invoke passes options', function() {
+    var yarn = new Command('yarn');
+
+    var passedOptions = {};
+    yarn._invoke = function(command, options) {
```
nathanhammond: Moved all spying tests to TestDouble.

stefanpenner:
@homu r+

nathanhammond:
<details>
<summary>This is what the code looked like prior to extraction. Yikes.</summary>

```javascript
var fs = require('fs-extra');
var path = require('path');
var quickTemp = require('quick-temp');
var execSync = require('child_process').execSync;
var symlinkOrCopySync = require('symlink-or-copy').sync;

var Configstore = require('configstore');
var conf = new Configstore('package-cache');

function buildCache() {
  // This is the object which will specify the path to the global cached packages.
  var dirs = {
    app: {},
    addon: {}
  };

  var contents = require('./blueprint-shim');

  quickTemp.makeOrReuse(dirs.app, 'bower_components');
  quickTemp.makeOrReuse(dirs.app, 'node_modules');
  quickTemp.makeOrReuse(dirs.addon, 'bower_components');
  quickTemp.makeOrReuse(dirs.addon, 'node_modules');

  fs.writeFileSync(path.join(dirs.app.bower_components, 'bower.json'), contents['app']['bower.json']);
  fs.writeFileSync(path.join(dirs.app.node_modules, 'package.json'), contents['app']['package.json']);
  fs.writeFileSync(path.join(dirs.addon.bower_components, 'bower.json'), contents['addon']['bower.json']);
  fs.writeFileSync(path.join(dirs.addon.node_modules, 'package.json'), contents['addon']['package.json']);

  // Generate the bower global caches.
  execSync('bower install', { cwd: dirs.app.bower_components });
  execSync('bower install --offline', { cwd: dirs.addon.bower_components });

  // Manually wire in Ember CLI.
  var emberCLIPath = path.resolve(__dirname, '../..');
  fs.mkdirSync(path.join(dirs.app.node_modules, 'node_modules'));
  symlinkOrCopySync(emberCLIPath, path.join(dirs.app.node_modules, 'node_modules', 'ember-cli'));
  fs.mkdirSync(path.join(dirs.addon.node_modules, 'node_modules'));
  symlinkOrCopySync(emberCLIPath, path.join(dirs.addon.node_modules, 'node_modules', 'ember-cli'));

  // Generate the npm global caches.
  execSync('npm install', { cwd: dirs.app.node_modules });
  execSync('npm install --cache-min 300', { cwd: dirs.addon.node_modules });

  conf.set('cached', true);
  conf.set('dirs', dirs);
}

module.exports = (function() {
  if (!conf.get('cached')) {
    buildCache();
  }
  return conf.get('dirs');
})();
```
</details>

Issue #6451, opened 2016-11-22
ember test not working with environmentConfig.rootUrl = null


Dejavs:
I have spent about 4 hours today in debugging ember test command and testem module. After executing command `ember test` i always get an error: `not ok 1 Browser "phantomjs [...] [...]" failed to connect. testem.js not loaded?`

Accidentaly i've found that the problem is with my config. I have set rootURL to null in environment.js config file. This is set to null because i'm developing mobile application with emberJS and i need only relative paths to files, so null value is perfect for me. But it breaks ember test command. I've found (after 4h research) that it can be easily fixed by adding:
`ENV.rootURL = '/';`
In the `environment === 'test' section.

---

Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.8.0
http_parser: 2.5.0
node: 4.2.6
v8: 4.5.103.35
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 55.1
modules: 46
openssl: 1.0.2f
os: linux x64
3.5.2
```

fpauser:
Same problem here. Created https://github.com/ember-cli/ember-cli/pull/6463 that should fix it.

nathanhammond:
@fpauser @Dejavs We actively removed that here: https://github.com/ember-cli/ember-cli/commit/20dfca0e0381b2899597b593a7e820b597468e90#diff-9588ab8abb99afa6cac30dd0e2b86041L32 in https://github.com/ember-cli/ember-cli/pull/5792

@Turbo87 Maybe you can provide some context for the decision here?

In general I don't want to get into the habit of hardcoding things for users differently in different environments. For example, that will make the test environment different from a person's production environment, possibly breaking things for people in a surprising manner. Removing that was one of the big wins of that PR.

Because of that I won't accept the corresponding PR https://github.com/ember-cli/ember-cli/pull/6463, but I would like to explore alternative solutions with y'all.

Turbo87:
@nathanhammond I believe it was removed because apps should be tested in the configuration that they are going to be deployed in, and testing with a different `rootURL` would defeat that purpose. Now obviously if you set `rootURL = null` things are a bit differently, but that sounds like something that should be explained in the same text that suggests to set `rootURL = null`.

nathanhammond:
[...]

PR #6463, opened 2016-11-24
Updated blueprint to force rootURL of '/' in tests

fpauser:


Turbo87:
@fpauser we have discussed this at the weekly CLI meeting yesterday and our conclusion is that we don't want to hardcode the `rootURL` at two places because it is likely that people will update one, but not the other.

what we would accept though is a PR that implements the suggestions in https://github.com/ember-cli/ember-cli/issues/6451#issuecomment-262832217, where `if (isEmpty(rootURL) && env === 'test') rootURL = '/'` or something like that.

PR #6468, opened 2016-11-25
Feature: Add current dir name to `process.title`

ro0gr:
Just an idea for enhancement. I haven't seriously tested it yet. Just interested if it's a way to go.

When I have several ember apps/addons started it's quite difficult to differentiate all the opened terminals and point to the right one from the first attempt. All of them have the same same title: "ember".

What if we provide the process title with a current directory name?

### DIFF
--- a/bin/ember
+++ b/bin/ember
@@ -2,7 +2,9 @@
 'use strict';
 
 // Provide a title to the process in `ps`
-process.title = 'ember';
+var path = require('path');
+var parsedPath = path.parse(process.cwd());
+process.title = 'ember:' + parsedPath.name;
 
 // require('time-require');
 var resolve = require('resolve');


rwjblue:
:+1: to the idea here! A few things to note:

* Because this file is required to work across newer and older versions (when installed globally this is the only file not used from the local project), and therefore changes here need very careful consideration. (This is just general caution and not directly related to concerns about the changes made in this PR).
* I'd prefer to actually move the setting of a more detailed `process.title` to after we have a project available. At that point we would use the project's name instead of the CWD name. In this scenario we would leave `bin/ember` alone and update [around here](https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L74) to set `process.title = 'ember: ' + environment.project.name;`. 

Thoughts?

ro0gr:
Totally agree. I'll make it

stefanpenner:
Would it make more sense to change the title later, and to the name of the project?

stefanpenner:
Specifically somewhere around here: https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L94 then we can use the project name if available.

Also, if we are changing something about the UI, we should doit via the [UI object](https://github.com/ember-cli/ember-cli/blob/master/lib/ui/index.js). That way it can be tested.

ro0gr:
@stefanpenner nice addition about ui object. Wasn't sure how to better test this. Thanks.

ro0gr:
> In this scenario we would leave bin/ember alone and update around here to set process.title = 'ember: ' + environment.project.name;.

Unfortunately I can't make work `process.title` to be set with space characters. It just rejects and remains the same as before an attempt to set. Can't find any related nodejs documentation about it.
Looks like an OS restriction(currently I'm on ubuntu). Will investigate.  

I've just realized we can also put the current command name to the UI title.
So instead `ember:serve` we can have something like `my-project:ember-serve` or `ember-serve(my-project)` in the titlebar. Looks like exact title format should be agreed.

@rwjblue @stefanpenner what do you think about adding of command name in the title? should I create an issue for this?

stefanpenner:
> @rwjblue @stefanpenner what do you think about adding of command name in the title? should I create an issue for this?

I think it could work, maybe something like?

* `ember:serve[ProjectName]`
* `ember:test[ProjectName]`

stefanpenner:
looks like the whitespace issue isn't on OSX. Maybe we fallback to no-whitespace only if the OS doesn't support it?

stefanpenner:
Im actually somewhat unsure about changing the process.title to drastically we do want it to show up reasonable in `ps` / `htop` etc and work with `killall -9 ember`

rwjblue:
@stefanpenner - Wouldn't your earlier 2 suggestions still work with those killall commands?

stefanpenner:
@rwjblue i don't believe killall would, since its exact-match only.

I think if the goal here is to format the titlebar of a terminal differently, terminals themselves usually have configuration to accomplish this.

ro0gr:
@stefanpenner you are right. killall doesn't act nicely with this change. 
It makes me think that playing with `process.title` is not the best idea in the world. 

Thanks for your feedback @stefanpenner @rwjblue, appreciate your help!

PR #6469, opened 2016-11-26
rename "LICENSE.MD" to "LICENSE"

kellyselden:
since this isn't really a markdown file anywho


Turbo87:
It does use a different rendering on GitHub though without the extension. Related question: why is there no such file in the app blueprint?

rwjblue:
I'm in favor of the change generally speaking, but don't see enough value to justify the blueprint churn. When changing file names we have to be careful since running the update will add the new file but not delete/move the existing one, leading to having both the new and old files (this was a significant pain point with the testem.json -> testem.js migration for example).

nathanhammond:
@Turbo87 No `LICENSE` in app blueprint: it's not a good guess that the application you're writing is open source. It's a decent guess that the addon you're writing is.

@kellyselden Given @rwjblue's concern I'm going to close this; I believe this ship has sailed. Reopen if you want to write the code to automagically do the transition for users of `ember init`.

More generally note that I have a preference of `.md` for improved rendering on GitHub. Closing this is a bit self-serving. 😜❤️

Turbo87:
@nathanhammond then we shouldn't set the license in the package.json file!

nathanhammond:
@Turbo87 Agreed. Should fix that.

PR #6475, opened 2016-11-28
extract ui to console-ui

stefanpenner:
The idea is console-ui to describe a common API for other libraries to confirm to. So that, there ui output can work nicely within a larger system.

* ui implementation
* mock ui implementation

This enables blueprint extraction

PR #6476, opened 2016-11-28
Blprnt

stefanpenner:
depends on: #6475

working towards actually decoupling blueprints from ember-cli:

* better/clearer boundaries between the two systems === good architecture
* people (including myself) want to use this elsewhere 

First step will be to keep the blueprint tests themselves as acceptance criteria. But the hope is to eventually remove most of them entirely. As most of the various permutation/combination tests would be better as very targetted unit tests, without only afew high-level smoke tests. Rather then many slow/brittle & hard to iterate right high level smoke tests.

- [x]  new project [blprint](ember-cli/blprnt)
- [x] extract tests
   - [x] many: https://github.com/ember-cli/blprnt/pull/8
   - [x] https://github.com/ember-cli/blprnt/issues/7
- [ ] more decoupling: https://github.com/ember-cli/blprnt/issues/6
- [ ] remove hacks
- [ ]  ???
- [ ] actually make this work


kellyselden:
Could we get #6473 resolved before this?

stefanpenner:
> Could we get #6473 resolved before this?

Would be great, I would love to have a better factoring of `blueprint` tests.

Turbo87:

PR #6479, opened 2016-11-28
docs: Blueprint:renamedFiles

les2:
This PR contains two small documentation updates broken into 2 commits.

The first commit simply corrects a typo in the property name used for `ember-cli/Blueprint:renameFiles`.

The second commit (which can be removed if folks object) adds some documentation for that property. I would like to contribute more documentation of this kind if folks are interested. The doc comment I have included is:

    /**
      Files that are renamed when installed into the target directory.
      This allows including files in the blueprint that would have an effect
      on another process, such as a file named `.gitignore`.

      The keys are the filenames used in the files folder.
      The values are the filenames used in the target directory.

      @static
      @property renamedFiles
      ```
    */

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1330,7 +1330,7 @@ Blueprint.list = function(options) {
 
 /**
   @static
-  @property renameFiles
+  @property renamedFiles
 */
 Blueprint.renamedFiles = {
   'gitignore': '.gitignore'
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1329,8 +1329,16 @@ Blueprint.list = function(options) {
 };
 
 /**
+  Files that are renamed when installed into the target directory.
+  This allows including files in the blueprint that would have an effect
+  on another process, such as a file named `.gitignore`.
+
+  The keys are the filenames used in the files folder.
+  The values are the filenames used in the target directory.
+
   @static
   @property renamedFiles
+  ```
 */
 Blueprint.renamedFiles = {
   'gitignore': '.gitignore'


les2:
just saw https://github.com/ember-cli/ember-cli/pull/6476 @stefanpenner 

Would you want me to reopen this PR to that repo?

The same typo exists there: https://github.com/ember-cli/blprnt/blob/master/lib/blueprint.js#L1335

lib/models/blueprint.js
```diff
@@ -1329,8 +1329,21 @@ Blueprint.list = function(options) {
...
-  @property renameFiles
+  @property renamedFiles
+
+  @example
+  ```js
+  // Rename the `gitignore` file to `.gitignore` upon installation
+  // in order to prevent it from interfering with [Git](https://git-scm.com/).
+  Blueprint.renamedFiles = {
+    'gitignore': '.gitignore'
+  };
```
Turbo87: @les2 this is a `static` property, not something that you can actually set on a custom blueprint, so I'm not sure how valuable that example actually is.

les2:
#Turbo87 i kind of agree. 

i do feel that the example makes the intent of the property even clearer, even if only for ember-cli developers.

theoretically - and I'm asking here - couldn't a build "plugin" / addon add an entry to this map if it really needed to? seems like a convenient escape hatch.

would you still object if the example *added* an entry instead of overwriting the entire property?

```
Blueprint.renamedFiles['jshintrc'] = '. jshintrc';
```

Or still in favor of nuking the example?

Personally, I think the example would assist others, but I also see your point about the property being static.

Let me know and I'll update the PR.

Turbo87:
> theoretically - and I'm asking here - couldn't a build "plugin" / addon add an entry to this map if it really needed to? seems like a convenient escape hatch.

`Blueprint.renamedFiles` is or should be a `private` property, so unless you have veeeery good reasons for doing so it should not be touched by anything outside of Ember CLI itself.

> i do feel that the example makes the intent of the property even clearer, even if only for ember-cli developers.

Since the example matches the actual implementation I don't see any immediate advantage. I do like the description text for the property though! 👍 

> Or still in favor of nuking the example?

I think that would be best, yes.

Thanks for your work on this! I hope I didn't discourage you to open more PRs like this.

les2:
definitely not discouraged! thanks for the review! I'll update shortly.

les2:
@Turbo87 added these lines to the comment:

    /**
      ...
      The keys are the filenames used in the files folder.
      The values are the filenames used in the target directory.
      ..
      ```
    */

That's the spirit of what I was aiming for with the `@example`.

thoughts?

Turbo87:
@homu r+

stefanpenner:
> just saw #6476 @stefanpenner
> 
> Would you want me to reopen this PR to https://github.com/ember-cli/blprnt and close this one? Or perhaps keep just this one? or open the PR to both repos?

If you could  that would be great.

les2:
> If you could that would be great.

@stefanpenner https://github.com/ember-cli/blprnt/pull/15

Issue #5336, opened 2016-01-12
Unable to override addon re-exports in the consumer application


ro0gr:
[repo](https://github.com/beatle/emcli-templates-collision-repro) with reproduction. This reproduction is based on a dummy app but it's also reproducible within a host app. 

Steps to reproduce:
```
$ ember g template test
$ ember g template test --dummy
```
produces following files:
- addon/templates/test.hbs
- app/templates/test.js
- tests/dummy/app/templates/test.hbs
- ...

Now if you run the build it will fail with an `EEXIST` exception: 
```
ember build
...
EEXIST, file already exists '/home/beatle/workspace/ember-cli-apps/c/tmp/template_compiler-output_path-beowNp4d.tmp/dummy/templates/test.js'
Error: EEXIST, file already exists '/home/beatle/workspace/ember-cli-apps/c/tmp/template_compiler-output_path-beowNp4d.tmp/dummy/templates/test.js'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:848:18)
    at symlink (/home/beatle/workspace/ember-cli-apps/c/node_modules/symlink-or-copy/index.js:82:14)
    at symlinkOrCopySync (/home/beatle/workspace/ember-cli-apps/c/node_modules/symlink-or-copy/index.js:58:5)
    at TemplateCompiler.Filter._handleFile (/home/beatle/workspace/ember-cli-apps/c/node_modules/broccoli-persistent-filter/index.js:101:12)
    at TemplateCompiler.<anonymous> (/home/beatle/workspace/ember-cli-apps/c/node_modules/broccoli-persistent-filter/index.js:88:21)
    at /home/beatle/workspace/ember-cli-apps/c/node_modules/promise-map-series/index.js:11:14
    at lib$rsvp$$internal$$tryCatch (/home/beatle/workspace/ember-cli-apps/c/node_modules/rsvp/dist/rsvp.js:493:16)
    at lib$rsvp$$internal$$invokeCallback (/home/beatle/workspace/ember-cli-apps/c/node_modules/rsvp/dist/rsvp.js:505:17)
    at lib$rsvp$$internal$$publish (/home/beatle/workspace/ember-cli-apps/c/node_modules/rsvp/dist/rsvp.js:476:11)
```

As I can see after some debugging it happens when broccoli tries to save compiled `tests/dummy/app/templates/test.hbs` to the `tests/dummy/app/templates/test.js` where the `app/templates/test.js` is merged already.

[...]

PR #6480, opened 2016-11-29
All JS templates to be used in addons

bcardarella:
#This patch will allow JS files to be used as templates in an addon.
Without the `overwrite: true` there is a merge error as the merging at
that point has a pre-compiled `.hbs` file in one tree and a
post-compiled `.js` file in the other tree. If, however, you are relying
on importing a template from another source and are using a `.js` file
as the origin no template compilation takes place and the two trees have
a similarly named file and a merge conflict results.

It may be best to actually remove the original `.hbs` file from the
original tree but that may add unecessary cost to the build step as those files are
already being filtered out later.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -736,6 +736,7 @@ var Addon = CoreObject.extend({
     var trees = [addonJs, templatesTree, reexported].filter(Boolean);
 
     return mergeTrees(trees, {
+      overwrite: true,
       annotation: 'Addon#compileAddon(' + this.name + ') '
     });
   },


rwjblue:
Can you give a more concrete example of what issue this is fixing for you (I'm trying to grok the set of files needed in a given addon to cause the error)?

rwjblue:
Is it basically:

```
addon/templates/foo.hbs
addon/templates/foo.js
```

?

bcardarella:
@rwjblue if I have a single file like:

```js
addon/template/components/foo.js
```

that is basically:

```js
export { default } from 'other/location';
```

bcardarella:
@rwjblue at this point there two trees at the LOC I patched has a copy of `template/components/foo.js` whereas if the original file was `template/components/foo.hbs` it would have `foo.hbs` in one tree and the compiled `foo.js` in the other. I didn't follow the compilation path completely but I suspect it is just a matter of the right-tree is post-compilation.

bcardarella:
Also, I can write a test for this but I didn't want to add another smoke test

stefanpenner:
we should likely also understand https://github.com/ember-cli/ember-cli/issues/5336 before merging this. As both this and that issue are seemingly related.

bcardarella:
@stefanpenner @ro0gr I'll try to address that issue with this PR if you'd like

rwjblue:
@bcardarella - I rebased and pulled this commit into #6538 because I realized that recent changes I made required it also, closing this one we will land the change (I kept your commit too) with that PR.

PR #6481, opened 2016-11-29
Update ember-load-initializers@0.6.0.

rwjblue:
This was a "breaking" change upstream, as we officially dropped support for IE8 (by using `Object.keys` / `String.prototype.lastIndexOf` / etc). No real changes...


rwjblue:
@homu r+

PR #6484, opened 2016-11-30
[BUGFIX] Fix remaining ember-source issues.

nathanhammond:
The removal from `bower` meant that we were erroneously removing it from `ember-source`. No longer.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -334,9 +334,11 @@ EmberApp.prototype._initVendorFiles = function() {
     delete this.vendorFiles['ember-resolver.js'];
   }
 
-  // don't crash if ember-cli-shims is not included
-  if (!bowerDeps['ember-cli-shims']) {
-    delete this.vendorFiles['app-shims.js'];
+  // Warn if ember-cli-shims is not included.
+  // `ember-source` bundles them by default, so we must check if that is the
+  // load mechanism of ember before checking `bower`.
+  if (!ember && !bowerDeps['ember-cli-shims']) {
+    this.project.ui.writeWarnLine('You have not included `ember-cli-shims` in your project\'s `bower.json`. This only works if you provide an alternative yourself and unset `app.vendorFiles[\'app-shims.js\']`.');
   }
 
   // this is needed to support versions of Ember older than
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -335,7 +335,7 @@ EmberApp.prototype._initVendorFiles = function() {
   }
 
   // don't crash if ember-cli-shims is not included
-  if (!bowerDeps['ember-cli-shims']) {
+  if (!ember && !bowerDeps['ember-cli-shims']) {
```
rwjblue: This change seems somewhat odd. The idea here was that a different addon could provide the shims. That would work even with using Ember from bower.

nathanhammond: In other words, what's the story of this: https://github.com/ember-cli/ember-cli/pull/6254#issuecomment-260717239

It's distinct from the `ember-resolver` issue but in that PR.

martndemus: What @rwjblue said, to be able to provide my own shims addon.

nathanhammond: I'm actually rather opposed to this check existing as it makes the failure for not including shims move to runtime where we can't control it as easily and manifests as:

> cannot find module 'ember'

Regardless, knowing the story makes it possible for me to address. Thank you!

nathanhammond: Okay, so the _only_ thing which requires this change is `[ember-new-modules-shim](https://github.com/DockYard/ember-new-modules-shim)`. The users of that addon should be assumed to be highly familiar Ember devs.

Given that, if users of `ember-new-modules-shim` wish to not include `ember-cli-shims` they will need to follow the _full_ set of instructions at https://github.com/DockYard/ember-new-modules-shim#optional-disabling-ember-cli-shims ... not just remove the `bower` dependency and rely on this change.

That is the only way we can make sure that shims are not included across both `ember-source`-provided and `bower`-provided shims while knowing user intent.

rwjblue:
👍 , seems good

nathanhammond:
Skipping @homu because CI is broken.

martndemus:
img_0364

PR #6487, opened 2016-11-30
Install JSHint files via "ember-cli-jshint" blueprint

### LINKED ISSUE #6269: Remove remaining coupling to JSHint

Turbo87:
This PR resolves #6269. It changes the steps in the `ember init` flow to be: install blueprint, `npm install`, `bower install`, `ember generate ember-cli-jshint`, `git init && git add && git commit`.

This will bring us closer to switching to `ember-cli-eslint` instead.


Turbo87:
As @rwjblue mentioned on Slack: this PR currently leaves the project in an inconsistent state when `--skip-npm` is used since the `ember-cli-jshint` addon is listed in `package.json`, but the blueprint won't be installed when `npm install` is run manually.

To resolve this we should remove `ember-cli-jshint` from the `package.json` file in the `app` blueprint and let NPM add it when installing the addon after the app itself has installed everything else. This is however currently blocked on #6502.

Turbo87:
@rwjblue I think I've addressed your initial feedback now. would love to get a second review :)

rwjblue:
LGTM

stefanpenner:
@homu r+

kellyselden:
A little late to the party, but should likes like [this](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/testem.js#L1) be removed as well and let the addon add them back? @Turbo87

nathanhammond:
@kellyselden That would be ridiculously complicated to do well. We'd need a better blueprint api, which I think we can get to via blprnt.

kellyselden:
@nathanhammond I may be missing something. Wouldn't it be easy to do with `Blueprint. insertIntoFile()`?

https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L1156

nathanhammond:
Yes-ish. If you knew which files existed. You end up introducing coupling between two separate projects without a specific API boundary and that gets a bit messy.

kellyselden:
I can see that. At the least shouldn't we remove the jshint artifacts in our blueprints, even if ember-cli-jshint doesn't choose to re-add?

nathanhammond:
Probably, open a tracking issue?

kellyselden:
#6540

PR #6491, opened 2016-11-30
Fix Builder._enableFSMonitorIfVizEnabled tests

hjdivad:
The monitor will only be active if it is the first instance to be constructed.
This is to avoid double counting.  This could be relaxed to be that only one
monitor can be active, but it still seems good to not even require the module
when viz is disabled.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -18,9 +18,9 @@ var signalsTrapped = false;
 var buildCount = 0;
 
 function _enableFSMonitorIfVizEnabled() {
-  var FSMonitor = require('heimdalljs-fs-monitor');
-  var monitor = new FSMonitor();
+  var monitor;
   if (vizEnabled()) {
+    monitor = new FSMonitor();
     monitor.start();
   }
   return monitor;
--- a/tests/unit/models/builder-test.js
+++ b/tests/unit/models/builder-test.js
@@ -1,13 +1,14 @@
 'use strict';
 
-var fs              = require('fs-extra');
+var fs              = require('fs');
+var fse             = require('fs-extra');
 var path            = require('path');
 var Builder         = require('../../../lib/models/builder');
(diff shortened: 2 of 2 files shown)


lib/models/builder.js
```diff
@@ -18,9 +18,10 @@ var signalsTrapped = false;
 var buildCount = 0;
 
 function _enableFSMonitorIfVizEnabled() {
-  var FSMonitor = require('heimdalljs-fs-monitor');
-  var monitor = new FSMonitor();
+  var monitor;
+    var FSMonitor = require('heimdalljs-fs-monitor');
```
nathanhammond: Whitespace.

tests/unit/models/builder-test.js
```diff
@@ -1,13 +1,14 @@
 'use strict';
 
-var fs              = require('fs-extra');
+var fs              = require('fs');
+var fse             = require('fs-extra');
```
nathanhammond: Why do we need both of these? I was pretty sure `fse` proxied to `fs`? Does `FSMonitor` not play nicely with the `fse` abstraction layer when being reset?

hjdivad: check the assert in the test

We want to assert that we have modified `fs.statSync`.  Because `fse.statSync !== fs.statSync` we can't trivially do this via `fs-extra`

hjdivad:
k it's a bit of a pain to not be able to rely on status but master has 48 failing tests, this branch has 47 and it's supposed to fix exactly 1 test.

lib/models/builder.js
hjdivad: thanks 👍

tests/unit/models/builder-test.js
nathanhammond: Seems good, thanks for the context! 😄

PR #6494, opened 2016-11-30
Fix more failing tests.

nathanhammond:
Not including it broke the tests when a new warning was added.


nathanhammond:
`ember-load-initializers@0.6.2` should resolve the last failing test. Not using `bower` here may unveil additional failures.

nathanhammond:
Merging this while still broken. Still trying to get CI green.

PR #6495, opened 2016-11-30
Make startApp functional

joeyespo:
Instead of using `let`, then setting the closure of `application` inside an immediately-run function callback, return the new application object, and then return that result directly.

This looks like a relic of [upgrading from `var`](https://github.com/ember-cli/ember-cli/commit/cc65668f9065d300e0daee49a3a201bfcca9c663) and the dated "put all `var` declarations at the top" [convention](https://github.com/ember-cli/ember-cli/commit/18a5f621b3f41b3a5f4139fc3d38118cde5e11df#diff-06761a3424465f8bb1a3e0fe86b96c31).

Note that if `Ember.run` ever becomes an asynchronous function, startApp will return `undefined` even without this change. `Ember.run` [does](https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-metal/lib/run_loop.js#L76-L78) run [immediately](https://github.com/ebryn/backburner.js/blob/v0.3.1/lib/backburner/index.js#L151) and [returns what is returned](https://github.com/ebryn/backburner.js/blob/v0.3.1/lib/backburner/index.js#L196) from the callback.

### DIFF
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -3,16 +3,13 @@ import Application from '../../app';
 import config from '../../config/environment';
 
 export default function startApp(attrs) {
-  let application;
-
   let attributes = Ember.merge({}, config.APP);
   attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
 
-  Ember.run(() => {
-    application = Application.create(attributes);
+  return Ember.run(() => {
+    let application = Application.create(attributes);
     application.setupForTesting();
     application.injectTestHelpers();
+    return application;
   });
-
-  return application;
 }


blueprints/app/files/tests/helpers/start-app.js
```diff
@@ -3,16 +3,13 @@ import Application from '../../app';
 import config from '../../config/environment';
 
 export default function startApp(attrs) {
-  let application;
-
   // use defaults, but you can override
-  let attributes = Ember.assign({}, config.APP, attrs);
+  const attributes = Ember.assign({}, config.APP, attrs);
```
joeyespo: Oh, ok. Changed.

Curious, is there rationale about that written anywhere?

kellyselden: https://github.com/ember-fastboot/fastboot/pull/100#discussion_r90163199

rwjblue: Better writeup than I can do: http://madhatted.com/2016/01/25/let-it-be

kellyselden: Thanks @rwjblue. Bookmarked for future use!

nathanhammond:
Sorry, just broke this PR. Can you rebase?

joeyespo:
@nathanhammond @homu It's all good! Rebased.

blueprints/app/files/tests/helpers/start-app.js
joeyespo: Hmm, while I disagree with the author's conclusion (he seems to be missing the point of constant *identifiers* (not values) and catering to common confusion instead of addressing it. [Read this for more in the context of identifier reassignment intent](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75)), I appreciate the reference 👍

joeyespo:
Tests are broken after the rebase. It's saying "Error: foo" @nathanhammond Any ideas?

stefanpenner:
👍 good change

nathanhammond:
@joeyespo That's a CI issue. I bet @homu will be happy with this.

@homu r+

nathanhammond:
@joeyespo Thank you for getting this in!

joeyespo:
Happy to! Thanks for all your hard work, Ember team(s) 😃

PR #6496, opened 2016-12-01
[BUGFIX release] Revert the reverted revert. Ember assign not available in all ember try scenarios 

webark:
Changing the start-app test helper to use `Ember.assign` instead of `Ember.merge`. Broke ember try as Ember 2.4 apparently doesn't have Ember.assign. Need to revert back to using the Ember.merge for now.

This reverts commit 7b282ae9afd46aef9943c765db8aa97a6efc3f83.

### DIFF
--- a/blueprints/app/files/tests/helpers/start-app.js
+++ b/blueprints/app/files/tests/helpers/start-app.js
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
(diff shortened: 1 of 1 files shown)


kellyselden:
This should released in a v2.10.1. Just came across this.

rwjblue:
Yep, is already on release branch and will be included in 2.10.1 when it is released.

nathanhammond:
Monday.

PR #6499, opened 2016-12-01
Ensure shared EMPTY_MERGE_TREE is reset after mocking in tests.

rwjblue:
The override functionality was being used in two different tests to ensure that the various exepected semantics were working properly, however this was not being cleaned up properly and therefore caused test failures for tests running afterwards that utilized the shared `EMPTY_MERGE_TREE`.


rwjblue:
Finally green:

```
  1013 passing (7m)
  8 pending
Mocha Tests Running Time: 435821.056ms
```

rwjblue:
@homu r+

PR #6506, opened 2016-12-01
Use `CoreObject` for builder and watcher

### LINKED ISSUE #6504: Builder should not extend `Task` but `CoreObject`
https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L60

twokul:
For some reason, `builder` and `watcher` inherited from `Task` which
isn't necessary. Switching to `CoreObject` instead.

Fixes #6504

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -6,7 +6,7 @@ var fs          = require('fs-extra');
 var existsSync  = require('exists-sync');
 var path        = require('path');
 var Promise     = require('../ext/promise');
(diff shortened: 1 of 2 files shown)


rwjblue:
Ya, looks good to me.  We just need to wait to land until after 2.11.0-beta.1 has been released.

rwjblue:
@homu r+

Turbo87:
@homu you're drunk!

PR #6507, opened 2016-12-01
Remove "ember-cli-release" from "app" and "addon" blueprint

### LINKED ISSUE #6462: Remove "ember-cli-release" from the default app blueprint
[`ember-cli-release`](https://github.com/lytics/ember-cli-release) has quite a few large [dependencies](https://github.com/lytics/ember-cli-release/blob/master/package.json#L23-L35) and I personally don't use it at all. Is this a plugin that people commonly use? It seems to me that the community has moved to `ember-cli-deploy` instead, so I propose removing `ember-cli-release` from the default app/addon blueprints. Those of us that use the plugin can just continue to use it, and everyone else will benefit from faster `npm install` times.

Turbo87:
Resolves #6462

/cc @kellyselden @nathanhammond @rwjblue

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -26,7 +26,6 @@
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-jshint": "^2.0.1",
     "ember-cli-qunit": "^3.0.1",
-    "ember-cli-release": "^0.2.9",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",


rwjblue:
To be clear, this also removes `ember-cli-release` from addons.  

We had a long conversation about this in this weeks ember-cli core team call, and generally believe that removal is the right path at this time.  The primary motivators are:

* `preversion` / `postversion` scripts along with `npm version` command often provide the same functionality.
* `ember-cli-release` is currently somewhat unmaintained (though I'm sure folks would step in if real issues were outstanding)
* `ember-cli-deploy` is the tool of choice for these sorts of things in apps
* `ember-cli-release` is likely not used by 50% of addons

Turbo87:
Thanks for the great summary @rwjblue!

nathanhammond:
The rationale for this decision is here: https://github.com/ember-cli/ember-cli/issues/6462#issuecomment-264322357

Locking conversation on this thread as I would like it to appear on the other thread.

rwjblue:
@homu r+

PR #6508, opened 2016-12-01
docs: remove bad markdown block

les2:
Fixes a documenation bug introduced by https://github.com/ember-cli/ember-cli/pull/6479

There was a dangling markdown block:

    ```

Removed that!

Note - this makes the documentation site render badly.

See https://ember-cli.com/api/classes/Blueprint.html and search for `renamedFiles`.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1338,7 +1338,6 @@ Blueprint.list = function(options) {
 
   @static
   @property renamedFiles
-  ```
 */
 Blueprint.renamedFiles = {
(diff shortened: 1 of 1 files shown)


les2:
@Turbo87 looks like i broke something here: https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L1341

Turbo87:
Oops, yeah, I missed that apparently. Good catch!

Turbo87:
@homu r+

PR #6509, opened 2016-12-01
[BUGFIX] Add missing ember-cli-shims to non ember-source scenarios

cibernox:
Without this, addons that update to 2.11 will break in CI


nathanhammond:
@homu r+

PR #6515, opened 2016-12-02
Avoid double walking of addon/templates in default scenario.

rwjblue:
By default `addon-templates` path is `addon/templates/` and would have already been walked when we check `addon/`. This is just wasteful work that is done on first build for each addon.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -293,6 +293,10 @@ var Addon = CoreObject.extend({
     return tree;
   },
 
+  _treePathFor: function _treePathFor(treeName) {
+    return path.normalize(path.join(this.root, this.treePaths[treeName]));
+  },
+
   /* @private
    * @method _warn
    */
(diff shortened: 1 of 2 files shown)


lib/models/addon.js
```diff
@@ -631,14 +631,15 @@ var Addon = CoreObject.extend({
     var templateExtensions = this.registry.extensionsForType('template');
     var addonTreePath = path.join(this.root, this.treePaths['addon']);
     var addonTemplatesTreePath = path.join(this.root, this.treePaths['addon-templates']);
+    var addonTemplatesTreeInAddonTree = addonTreePath.indexOf(addonTemplatesTreePath) === 0;
```
nathanhammond: If we're going to do this I _think_ we need to `path.normalize` it since you could find yourself traversing out of the context if you were passed `../asdf`.

rwjblue: Good point, will fix

rwjblue:
> Also, how easy is this to test?

Unsure ATM, its mostly just reducing duplicated work so it isn't _wrong_ per-se (just wasteful of I/O when using the default `addon-templates` path).  

I think the longer term solution to testing this would be to setup a heimdalljs-fs-monitor testing harness, simulate the specific scenario, and ensure it avoids the extra `readdir` / `stat` / etc calls.

rwjblue:
Updated to fix issue mentioned RE: `../foo` paths.  

I'd like to land and continue iterating on a better testing harness for this stuff.  I'm currently working on creating a harness to test addon broccoli methods more like you would test a broccoli-plugin, but am trying to split up unrelated refactors (so that I don't end up with a giant impossible to reason about PR)...

rwjblue:
Updated with some tests to confirm behavior.

stefanpenner:
👍

rwjblue:
@homu r+

PR #6520, opened 2016-12-03
Improve Windows Symlink Related messaging

stefanpenner:
* refactor to be more testable (but maintain same external API)
* improve tests
* don’t warn if canSymlink is already true (may have: https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/ setup, or some other mechanism)
* make sure the URL is formatted correctly

- [x] fix YUIDoc lint issue
- [x] get reviews

--- 

- [x] issue to update website: https://github.com/ember-cli/ember-cli.github.io/issues/113

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -5,7 +5,7 @@ var Command     = require('../models/command');
 var Promise     = require('../ext/promise');
 var SilentError = require('silent-error');
 var PortFinder  = require('portfinder');
-var win         = require('../utilities/windows-admin');
+var Win         = require('../utilities/windows-admin');
 var EOL         = require('os').EOL;
 
 PortFinder.basePort = 49152;
@@ -72,7 +72,7 @@ module.exports = Command.extend({
           project: this.project
         });
 
-        return win.checkWindowsElevation(this.ui).then(function() {
+        return Win.checkIfSymlinksNeedToBeEnabled(this.ui).then(function() {
           return serve.run(commandOptions);
         });
       }.bind(this));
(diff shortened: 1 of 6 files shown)


lib/utilities/windows-admin.js
```diff
@@ -2,40 +2,135 @@
...
+ *
+ * @method checkIfSymlinksNeedToBeEnabled
+ * @return {Promise<Object>} Object describing whether we're on windows and if admin rights exist
+ */
+WindowsSymlinkChecker.checkIfSymlinksNeedToBeEnabled = function(ui) {
+  return this._setup(ui).checkIfSymlinksNeedToBeEnabled();
+};
+
+/**
+ * @private
```
rwjblue: Should be below description

lib/commands/serve.js
```diff
@@ -72,7 +72,7 @@ module.exports = Command.extend({
           project: this.project
         });
 
-        return win.checkWindowsElevation(this.ui).then(function() {
+        return Win.checkIfSymlinksNeedToBeEnabled(this.ui).then(function() {
```
rwjblue: Is this a static method somehow? It looks from the refactor that it's only on the prototype but this doesn't create an instance. Am I missing something?

stefanpenner: @rwjblue there are 2 methods, one that is static and one that is on the prototype. The static one sets up the state, that the tests prefer to setup themselves.

stefanpenner:
also tested this manually in windows, all appears well.

felixrieseberg:
Jesus, @stefanpenner is a machine. Well done, sir.

rwjblue:
@homu r+

PR #6521, opened 2016-12-04
fix missing git scenarios

stefanpenner:
- [x] figure out if this fixes the issue I saw on windows, if not figure out what that issue was and fix it
- [ ] r?


rwjblue:
@homu r+

PR #6522, opened 2016-12-04
[BUGFIX beta] Ensure `--no-color` option is present for all commands.

rwjblue:
Using this option has always worked to disable `chalk` from actually emitting colored output (and therefore already has the desired effect), but it would emit a warning to the user when used suggesting that the option was not valid.

This adds `--no-color` as a global option to all commands, and adds a test to ensure that no warning is printed when it is used.

Addresses part of https://github.com/ember-cli/rfcs/issues/83.

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -26,6 +26,12 @@ var allowedWorkOptions = {
   everywhere: true
 };
 
+var globallyAllowedOptions = [
+  'argv',  // Why?
+  'h',     // -h
+  'help'   // --help
+];
+
 path.name = 'Path';
 // extend nopt to recognize 'gitUrl' as a type
 nopt.typeDefs.gitUrl = {
@@ -132,6 +138,13 @@ var Command = CoreObject.extend({
      * ```
      */
     this.availableOptions = this.availableOptions || [];
+    this.availableOptions.push({
+      name: 'no-color',
+      type: Boolean,
+      default: false,
+      aliases: [ ],
+      description: 'To disable usage of colors in the output.'
+    });
 
     /**
      * An array of anonymous options for the command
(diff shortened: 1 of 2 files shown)


lib/models/command.js
```diff
@@ -132,6 +138,7 @@ var Command = CoreObject.extend({
      * ```
      */
     this.availableOptions = this.availableOptions || [];
+    this.availableOptions.push({ name: 'no-color', type: Boolean, default: false, aliases: [ ], description: 'To disable usage of colors in the output.'});
```
stefanpenner: can this be multi-line?

rwjblue: ya, sounds good

rwjblue:
Updated to use multi-line (I do agree that it needed it, should've started that way).

stefanpenner:
@homu r+

stefanpenner:
heh, seems like the related tests are abit brittle...

rwjblue:
Ahh, I see. I have to update all the other command help output tests to include the new option. Thats going to have to wait for tomorrow 😴...

stefanpenner:
> Ahh, I see. I have to update all the other command help output tests to include the new option. Thats going to have to wait for tomorrow 😴...

I got some cycles, I'll doit now.

tests/unit/models/command-test.js
```diff
@@ -140,6 +140,13 @@ describe('models/command.js', function() {
     });
   });
 
+  it('parseArgs() should allow --no-color option', function() {
+    new ServeCommand(assign(options, {
+      settings: config.getAll()
+    })).parseArgs(['--no-color']);
```
stefanpenner: this test appears to be legitimately failing

stefanpenner:
> I got some cycles, I'll do some now.

Ok did one file, tired now...

Turbo87:

PR #6523, opened 2016-12-04
[fixes #6445] choose a more random port during test runs, this should…

### LINKED ISSUE #6445: ember-cli's own tests depend on port 4200
this is mega annoying when both working on ember-cli and an ember app at the same time.

We should use port-finder or something...

stefanpenner:
… allow running CLI tests \w an app running


rwjblue:
@stefanpenner -  👍 in general here, but we should make sure to use `@homu` for merging to help prevent a whole slew of issues...

stefanpenner:
i keep forgetting homu works again

PR #6524, opened 2016-12-04
Remove arbitrary *.js filtering for addon tree.

rwjblue:
This seems like a needless waste of time/effort. `broccoli-filter` inheritors already intelligently handle files from unknown types ( they simply pass through untouched) and filtering here like this is fairly costly.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -851,12 +851,7 @@ var Addon = CoreObject.extend({
   addonJsFiles: function(tree) {
     this._requireBuildPackages();
 
-    var includePatterns = this.registry.extensionsForType('js').map(function(extension) {
-      return new RegExp(extension + '$');
-    });
-
     return new Funnel(tree, {
-      include: includePatterns,
       destDir: 'modules/' + this.moduleName(),
       description: 'Funnel: Addon JS'
     });


stefanpenner:
also gets us closer to JS -> HBS a a single pipeline, rather then funnel + filter + merg

rwjblue:
Rebased on top of https://github.com/ember-cli/ember-cli/pull/6528 (which fixed AppVeyor).

@homu r+

rwjblue:
@homu retry

lib/models/addon.js
```diff
@@ -851,12 +851,7 @@ var Addon = CoreObject.extend({
   addonJsFiles: function(tree) {
     this._requireBuildPackages();
 
-    var includePatterns = this.registry.extensionsForType('js').map(function(extension) {
-      return new RegExp(extension + '$');
-    });
-
     return new Funnel(tree, {
-      include: includePatterns,
```
nathanhammond: The output tree here just dumped all of the assets in the tree into the `modules/foo-addon/` folder. That seems incredibly odd and isn't just a pass-through at that point; we've added new assets into a location that is [treated as JS-only](https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L298-L349).

I'm pretty concerned that this will break things but we can let it ride and see.

rwjblue: I do not see anything in your link above that would error if non-js assets are present.

nathanhammond: I didn't see anything either, other than the fact that it would be present. I am interested in letting this ride.

PR #6503, opened 2016-12-01
utilities/npm: Run NPM commands via "execa"

Turbo87:
Resolves #6502 

/cc @rwjblue @nathanhammond @stefanpenner @hjdivad

PR #6525, opened 2016-12-04
utilities/npm: Run NPM commands via "execa"

Turbo87:
backport of #6503 to the `beta` branch

> the NPM3 progress bar is screwing around with inquirer, e.g. when asking to overwrite files
>
> essentially the question is hidden behind the progressbar and it looks as if NPM froze

/cc @rwjblue @nathanhammond @stefanpenner

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -4,7 +4,6 @@
 
 var chalk = require('chalk');
 var Task  = require('../models/task');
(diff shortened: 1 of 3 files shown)


rwjblue:
@homu r+

nathanhammond:
Note that the intent of this is, "if it works in beta we consider promoting to release in 1-2 weeks."

PR #6528, opened 2016-12-04
Fix addon pod template detection on windows.

rwjblue:
When run under windows, the path separator used by `path.join` in `this._treePathFor` will be `\` which will reek havoc on later path manipulation logic (e.g. when we attempt to determine if `addon-templates` tree is within `addon` tree).


rwjblue:
@homu r+

PR #6529, opened 2016-12-04
finally add yarn.lock

stefanpenner:
* only for developers of ember-cli itself
* consumers will not be affected
* our CI does not yet use yarn


rwjblue:
> only for developers of ember-cli itself

👍 

> our CI does not yet use yarn

We need to be careful about exactly how to test when we do use `yarn` from CI.  I _believe_ that many of these problems have started to be ironed out in https://github.com/ember-cli/ember-cli/pull/6461.

stefanpenner:
> We need to be careful about exactly how to test when we do use yarn from CI. I believe that many of these problems have started to be ironed out in #6461.

i suspect it is just as simple as ignoring the lockfile in CI

nathanhammond:
> only for developers of ember-cli itself

This does not apply only to developers of Ember CLI. Anybody who `npm/yarn link`s Ember CLI will be impacted by it as well. They _kinda_ count as developers, but not everybody necessarily knows what's happening with this.

It also implies that _all_ Ember CLI developers _must_ adopt `yarn` which I believe to be a discussion that should happen more broadly than a quick one-off PR. (Otherwise that lockfile is meaningless or even actively harmful.)

stefanpenner:
> Anybody who npm/yarn links Ember CLI will be impacted by it as well.

only if they used `yarn` to install ember-cli's dependencies.

> They kinda count as developers, but not everybody necessarily knows what's happening with this.

Yup

> It also implies that all Ember CLI developers must adopt yarn which I believe to be a discussion that should happen more broadly than a quick one-off PR.

We should recommend all dependency changes come with the associated yarn lock file change, but beyond that users could in theory continue to use yarn or npm or there own custom gradle task ;)

> (Otherwise that lockfile is meaningless

Yup

> even actively harmful

I don't believe so, other projects like rails have taken this approach for quite some time with success (committed lock for dev, no-lock for CI)

stefanpenner:
the only downside to this is:

* transitive dependency related failures wont be immediately noticed, but CI will catch

The big upside :

* reproducible development environment
* debugging failures due to transitives changing/breaking can now use the lock file as a way to "bisect" the problem.

nathanhammond:
I feel like we must get all of the Ember CLI team onto `yarn` if we do this. Otherwise I'll add something to `package.json` (and not the lockfile) and it'll work just fine for me, for CI, and then go boom on your machine @stefanpenner.

Current status: still using `npm` after the absolute mess that was the 2.10/2.11 release process which required me to completely reset my `npm` and `yarn` environments because of weird linking behaviors.

> other projects like rails have taken this approach for quite some time

Those other projects didn't have two mostly-API-compatible tools both still in wide use. It's entirely reasonable for somebody to use `npm` right now which will break the `yarn` workflow like a champ.

stefanpenner:
> I feel like we must get all of the Ember CLI team onto yarn if we do this. Otherwise I'll add something to package.json (and not the lockfile) and it'll work just fine for me, for CI, and then go boom on your machine @stefanpenner.

yes, i recommend:

> We should recommend all dependency changes come with the associated yarn lock file change, but beyond that users could in theory continue to use yarn or npm or there own custom gradle task ;).

nathanhammond:
> We should recommend all dependency changes come with the associated yarn lock file change

My point was that we don't have any tooling to enforce this. So instead it will just troll us. See: Greenkeeper.

stefanpenner:
> and it'll work just fine for me, for CI, and then go boom on your machine @stefanpenner.

May actually be a yarn issue. For example, it may want to store a checksum of package.json at the time of locking in its lock file. Detecting if its lock is is valid or not.

nathanhammond:
> it may want to store a checksum of package.json at the time of locking in its lock file. Detecting if its lock is is valid or not.

This would solve my biggest complaint.

stefanpenner:
> We should recommend all dependency changes come with the associated yarn lock file change
> My point was that we don't have any tooling to enforce this. So instead it will just troll us. See: Greenkeeper.

We should be mindful to not let perfect be the enemy of the good.

For now the power of humans code-reviews already covers many other lack-of-tooling scenarios. We can graduate overtime to improved tooling such as:

* yarn aware green keeper
* yarn checksum of package.json in lockfile (unless there is a technical reason for this not to)

nathanhammond:
https://github.com/greenkeeperio/greenkeeper/issues/314

stefanpenner:
`yarn check` and `yarn check --integrity` may to the trick, if problems arise.

https://yarnpkg.com/en/docs/cli/check

Issue #5015, opened 2015-10-30
Babel 6.0


stefanpenner:
Would like to cut over as soon as possible

```
npm WARN deprecated babel-core@5.8.33: Babel 5 is no longer being maintained. Upgrade to Babel 6.
```
- [x] update broccoli-babel-transpiler (https://github.com/babel/broccoli-babel-transpiler/issues/62)
- [x] update babel-plugin-feature-flags (https://github.com/ember-cli/babel-plugin-feature-flags/pull/6)
- [x] update babel-plugin-filter-imports (https://github.com/ember-cli/babel-plugin-filter-imports/pull/6)
- [x] update babel-plugin-htmlbars-inline-precompile (https://github.com/pangratz/babel-plugin-htmlbars-inline-precompile/pull/8)
- [ ] update ember-cli-htmlbars-inline-precompile ([ember-cli/ember-cli-htmlbars-inline-precompile](https://github.com/ember-cli/ember-cli-htmlbars-inline-precompile))
- [ ] update ember-cli-babel (https://github.com/babel/ember-cli-babel/issues/62)
- [ ] update emberjs-build (https://github.com/emberjs/emberjs-build/pull/165)

---

- [x] Migrate `ember-cli-babel` specific settings out of `babel` config and into its own `ember-cli-babel` configuration (with support for older style with a deprecation). This removes one of the "double duties" that the `babel` config currently takes and _should_ help us traverse the path forward to babel@5 and babel@6 interop. (https://github.com/babel/ember-cli-babel/issues/102)
- [ ] Update ember-cli to **require** `ember-cli-babel` for things to function:
  - [ ] [lib/broccoli/ember-app.js#L1237](https://github.com/ember-cli/ember-cli/blob/dbf034b13cd25cfc401b55019f15d104d4289f09/lib/broccoli/ember-app.js#L1237) - This comment infers something that simply doesn't work anymore. We should likely remove that guard and just throw if we don't detect `ember-cli-babel`.
- [ ] Update `ember-cli-babel` to deprecate using the default of "blacklist modules"
  - [x] Change lib/models/addon.js to compile modules. This is used in ember-cli to ensure that we do a two-pass compilation of babel for addons.  The first is done by the addon itself, and the second is done by lib/broccoli/ember-app.js#L929.
[...]

PR #6530, opened 2016-12-05
Prevent running babel twice.

rwjblue:
We currently transpile the output of all addon's `addon/` trees twice.  Once inside the addon's `treeForAddon` step (this one excludes module transpilation by adding `es6.modules` to babel's blacklist) and the second time in [lib/broccoli/ember-app.js](https://github.com/ember-cli/ember-cli/blob/2df251ce018fd12570bdce72c167ff14a427fa74/lib/broccoli/ember-app.js#L929). 

This was originally done because we didn't require that addon's include their own preprocessors for JS (mostly due to performance reasons IIRC).  In the time since that decision was made, we have swapped from a few different systems `es6-module-transpiler` -> `esperanto` -> `babel`.  Now that our module transpilation step is using the same thing as we recommend addon's use (`ember-cli-babel`) what we are doing seems pretty whacky.  

This PR removes the double processing of addon's `addon/` files through babel, and includes a deprecation message for the scenario that originally prompted the second pass (where an addon didn't include a module transpiling preprocessor).

TODO: 

- [x] Confirm that we have tests that already confirm this is working properly (I believe that our existing nested addon tests cover these changes, but I'd like to confirm).
- [ ] Consider / discuss backporting the deprecation message for addons that do not have a transpiler and are relying on the global processing step.
- [X] Add helpful error if an addon overrides `this.options.babel.compileModules`
- [X] Implement logic to avoid deprecation being added in https://github.com/babel/ember-cli-babel/pull/105.
- [X] Split out `findAddonByName` into a utility method to be shared
- [ ] r?

Addresses parts of https://github.com/ember-cli/ember-cli/issues/5015

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -920,17 +920,10 @@ EmberApp.prototype._addonTree = function _addonTree() {
     annotation: 'TreeMerger (addons)'
   });
 
-  var addonES6 = new Funnel(addonTrees, {
+  var addonTranspiledModules = new Funnel(addonTrees, {
     srcDir: 'modules',
     allowEmpty: true,
     annotation: 'Funnel: Addon JS'
-  });
-
-  var transpiledAddonTree = new Babel(addonES6, this._prunedBabelOptions());
-  var trees = [ transpiledAddonTree ];
-
-  var reexportsAndTranspiledAddonTree = mergeTrees(trees, {
-    annotation: 'TreeMerger: (re-exports)'
   });
 
   return this._cachedAddonTree = [
@@ -941,7 +934,7 @@ EmberApp.prototype._addonTree = function _addonTree() {
       annotation: 'Concat: Addon CSS'
     }),
 
-    this._concatFiles(reexportsAndTranspiledAddonTree, {
+    this._concatFiles(addonTranspiledModules, {
       inputFiles: ['**/*.js'],
       outputFile: '/addons.js',
       allowNone: true,
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -28,6 +28,7 @@ var walkSync   = require('walk-sync');
 var ensurePosixPath = require('ensure-posix-path');
 var defaultsDeep = require('ember-cli-lodash-subset').defaultsDeep;
 var Babel = require('broccoli-babel-transpiler');
+var findAddonByName = require('../utilities/find-addon-by-name');
 
 function warn(message) {
   if (this.ui) {
@@ -143,13 +144,15 @@ var Addon = CoreObject.extend({
 
     this.options = defaultsDeep(this.options, {
       babel: {
-        // TODO: coordinate with @turbo87 on exactly how to do
-        // this without issuing a deprecation
-        compileModules: true,
         modules: 'amdStrict',
         moduleIds: true,
         resolveModuleSource: require('amd-name-resolver').moduleResolve
       }
+    });
+
+    var emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
+    this.options[emberCLIBabelConfigKey] = defaultsDeep(this.options[emberCLIBabelConfigKey], {
+      compileModules: true
     });
   },
 
@@ -331,6 +334,18 @@ var Addon = CoreObject.extend({
    * @method _warn
    */
   _warn: warn,
+
+  _emberCLIBabelConfigKey: function() {
+    // future versions of ember-cli-babel will be moving the location for its
+    // own configuration options out of `babel` and will be issuing a deprecation
+    // if used in the older way
+    //
+    // see: https://github.com/babel/ember-cli-babel/pull/105
+    var emberCLIBabelInstance = findAddonByName(this.addons, 'ember-cli-babel');
+    var emberCLIBabelConfigKey = (emberCLIBabelInstance && emberCLIBabelInstance.configKey) || 'babel';
+
+    return emberCLIBabelConfigKey;
+  },
 
   /**
     Returns a given type of tree (if present), merged with the
@@ -829,8 +844,13 @@ var Addon = CoreObject.extend({
   compileAddon: function(tree) {
     this._requireBuildPackages();
 
-    if (!this.options.babel.compileModules) {
-      throw new Error('y u clobber!?!!?!');
+    var emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
+    if (!this.options[emberCLIBabelConfigKey].compileModules) {
+      throw new SilentError(
+        'Ember CLI addons manage their own module transpilation during the `treeForAddon` processing. ' +
+          '`' + this.name + '` has overridden the `this.options.' + emberCLIBabelConfigKey + '.compileModules` ' +
+          'value which conflicts with the addons ability to transpile its `addon/` files properly.'
+      );
     }
 
     var addonJs = this.processedAddonJsFiles(tree);
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -26,6 +26,7 @@ var mergeTrees = require('../broccoli/merge-trees');
 var Funnel     = require('broccoli-funnel');
 var walkSync   = require('walk-sync');
 var ensurePosixPath = require('ensure-posix-path');
+var defaultsDeep = require('ember-cli-lodash-subset').defaultsDeep;
 
 function warn(message) {
   if (this.ui) {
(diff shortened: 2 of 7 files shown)


lib/models/addon.js
```diff
@@ -797,16 +831,23 @@ var Addon = CoreObject.extend({
   compileAddon: function(tree) {
     this._requireBuildPackages();
 
-    var reexported;
+    if (!this.options.babel.compileModules) {
+      throw new Error('y u clobber!?!!?!');
```
rwjblue: Remove this before landing, was using it as a debugging tool

lib/models/addon.js
```diff
@@ -878,9 +919,23 @@ var Addon = CoreObject.extend({
...
-    });
+    if (this._fileSystemInfo().hasJSFiles) {
+      var processedJsFiles = this.preprocessJs(this.addonJsFiles(tree), '/', this.name, {
+        registry: this.registry
+      });
+
+      if (!registryHasAddon(this.registry, 'ember-cli-babel')) {
+        this._warn('Addon files were detected in `' + this._treePathFor('addon') + '`, but `ember-cli-babel` ' +
+                              'is not present as a dependency in `' + this.name + '`. ' +
+                              'Please make sure to add `ember-cli-babel`) ' +
```
rwjblue: Weird `)` mid-sentence here

lib/broccoli/ember-app.js
```diff
@@ -920,19 +920,12 @@ EmberApp.prototype._addonTree = function _addonTree() {
...
   });
 
-  var addonES6 = new Funnel(addonTrees, {
+  var addonTranspiledModules = new Funnel(addonTrees, {
     srcDir: 'modules',
     allowEmpty: true,
     annotation: 'Funnel: Addon JS'
   });
 
-  var transpiledAddonTree = new Babel(addonES6, this._prunedBabelOptions());
```
stefanpenner: 🔥

lib/broccoli/ember-app.js
```diff
@@ -920,19 +920,12 @@ EmberApp.prototype._addonTree = function _addonTree() {
...
-  var trees = [ transpiledAddonTree ];
-
-  var reexportsAndTranspiledAddonTree = mergeTrees(trees, {
```
stefanpenner: :🔥

lib/models/addon.js
```diff
@@ -878,9 +937,22 @@ var Addon = CoreObject.extend({
...
+    if (this._fileSystemInfo().hasJSFiles) {
+      var processedJsFiles = this.preprocessJs(this.addonJsFiles(tree), '/', this.name, {
+        registry: this.registry
+      });
+
+      if (!registryHasPreprocessor(this.registry, 'js')) {
+        this._warn('Addon files were detected in `' + this._treePathFor('addon') + '`, but no JavaScript ' +
+                   'preprocessors were found for `' + this.name + '`. Please make sure to add a preprocessor ' +
+                   '(most likely `ember-cli-babel`) to in `dependencies` (NOT `devDependencies`) in ' +
+                   '`' + this.name + '`\'s `package.json`.');
```
stefanpenner: as it may be possible to have more then 1 addon by this name, should we also print its path on disk?

lib/models/addon.js
```diff
@@ -47,6 +50,18 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
   };
 }
 
+function processModulesOnly(tree) {
+  return new Babel(tree, {
+    modules: 'amdStrict',
+    moduleIds: true,
+    whitelist: ['es6.modules']
```
stefanpenner: do we need to provide the [amd-name-resolver](https://github.com/chadhietala/amd-name-resolver)? As this doesn't seem to get the config from 146

lib/models/addon.js
```diff
@@ -307,6 +335,18 @@ var Addon = CoreObject.extend({
    */
   _warn: warn,
 
+  _emberCLIBabelConfigKey: function() {
+    // future versions of ember-cli-babel will be moving the location for its
+    // own configuration options out of `babel` and will be issuing a deprecation
+    // if used in the older way
+    //
+    // see: https://github.com/babel/ember-cli-babel/pull/105
+    var emberCLIBabelInstance = findAddonByName(this.addons, 'ember-cli-babel');
```
stefanpenner: might be reason to add `this.findOwnAddonByName('ember-cli-babel')` as an API.

lib/models/addon.js
```diff
@@ -730,7 +777,7 @@ var Addon = CoreObject.extend({
     if (addonTemplates) {
       standardTemplates = new Funnel(addonTemplates, {
         srcDir: '/',
-        destDir: 'modules/' + this.name + '/templates'
+        destDir: this.name + '/templates'
```
stefanpenner: @nathanhammond mentioned some add-ons accidentally depend on this, we should likely see if any very popular ones absolutely do. And inform said authors to address.

lib/models/addon.js
```diff
@@ -797,16 +844,28 @@ var Addon = CoreObject.extend({
   compileAddon: function(tree) {
     this._requireBuildPackages();
 
-    var reexported;
+    var emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
+    if (!this.options[emberCLIBabelConfigKey].compileModules) {
+      throw new SilentError(
+        'Ember CLI addons manage their own module transpilation during the `treeForAddon` processing. ' +
+          '`' + this.name + '` has overridden the `this.options.' + emberCLIBabelConfigKey + '.compileModules` ' +
```
stefanpenner: as there may be more then one addon by this name in node_modules, should we print out its whole path?

lib/models/addon.js
rwjblue: Confirm, will update

lib/models/addon.js
rwjblue: Agreed, but I wanted to decouple new public API's from this PR. 

Technically, `project.findAddonByName` is also private (per YUIDocs IIRC), I'd suggest a future RFC to make that method public on both `Project` and `Addon`.

lib/models/addon.js
rwjblue: @stefanpenner - The `modules/` funneling is done a tad bit later. In this case, we need the paths on disk to match what we expect the named AMD moduleIds will be.

lib/models/addon.js
stefanpenner: sounds very reasonable.

lib/models/addon.js
rwjblue: Sounds good, will do

lib/models/addon.js
rwjblue: This is done around L865 in this file.

lib/models/addon.js
rwjblue: The full path absolute path to the addon's `addon/` directory is printed in this message, so I don't think changes are needed in this one for that purpose...

stefanpenner: 👍

rwjblue:
Updated from initial round of review...

lib/models/addon.js
```diff
@@ -47,6 +56,17 @@ function deprecatedAddonFilters(addon, name, insteadUse, fn) {
...
 
+function processModulesOnly(tree) {
+  var options = defaultsDeep({}, DEFAULT_BABEL_CONFIG);
+  options.whitelist = ['es6.modules'];
+
+  return new Babel(tree, options);
+}
+
+function registryHasPreprocessor(registry, type) {
+  return !!registry.load(type).length;
```
twokul: will `return registry.load(type).length > 0` be clearer and easier to parse?

lib/models/addon.js
twokul: I think both `ember-a11y` and `liquid-fire` depend on this

lib/models/addon.js
rwjblue: Yep, good idea!

rwjblue:
@twokul - RE: 

> I think both ember-a11y and liquid-fire depend on this 

Yes, `modules/` shifting isn't removed here, just moved around (now its done at the very end of `compileAddon` (called from `treeForAddon`).  There should be no semantic change here...

lib/models/addon.js
stefanpenner: @twokul do they depend on this, or the later `modules` funnel? I don't believe the later is removed (yet) although we do intend to explore this soon.

rwjblue:
@homu r+

mfazekas:
@rwjblue the issue with `ember-lodash` seems to be the condition added in the PR:

```
if (this._fileSystemInfo().hasJSFiles) {
      var processedJsFiles = this.preprocessJs(this.addonJsFiles(tree), '/', this.name, {
       ...
}
```

ember-lodash itself has no JSFiles, but it's generating it's tree from `.js` files in other modules - 
see 
https://github.com/mike-north/ember-lodash/blob/master/index.js#L15

PR #6532, opened 2016-12-06
DO NOT MERGE: Ensure proper failure

rwjblue:


tests/unit/make-it-fail-test.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+var expect = require('chai').expect;
+
+describe('die die die', function() {
```
twokul: [die die die my darling](https://www.youtube.com/watch?v=JoolQUDWq-k)

nathanhammond:
The repair here is broken. Closing this PR as it doesn't give us additional information.

PR #6533, opened 2016-12-06
blueprints/addon: Fix path to "ember" executable in ".travis.yml"

Turbo87:
Resolves the issues caused by travis-ci/travis-cookbooks#786

Note that I opened https://github.com/ember-cli/ember-cli-mocha/pull/155 to test if this change actually works.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -37,4 +37,4 @@ install:
 script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
-  - ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
+  - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup


blueprints/addon/files/.travis.yml
```diff
@@ -37,4 +37,4 @@ install:
 script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
-  - ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
+  - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
```
blimmer: This will certainly work, I proposed using `npm run` and created a `test-ci` script in `package.json`. It is more in line IMO with best practices, but this is a smaller change. I don't feel super strongly about it.

blimmer:
Travis fixed this upstream (for now) https://github.com/travis-ci/travis-build/pull/905

Turbo87:
yeah, we should still fix it though

blueprints/addon/files/.travis.yml
rwjblue: I agree with both cases actually.  

For a hot-fix in the release branch we should do the smallest possible change, which means changing this file directly to use `node_modules/.bin/ember` here. 

We can totally address a `package.json` `script` based system later in beta/canary...

nathanhammond:
@homu r+

PR #6536, opened 2016-12-07
fix phantom use on travis

kellyselden:
I noticed an issue where phantom 1.9 was being used even though we were expecting 2.1 https://travis-ci.org/ember-cli/ember-addon-output/jobs/180486782#L447-L468

turns out we needed to install it globally. Here it is in action: https://travis-ci.org/kellyselden/ember-datetimepicker/jobs/181879916#L472

I targeted release branch because this seemed like a pretty silly/low risk bug fix that we should get to people soon.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -25,10 +25,9 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g bower
+  - npm install -g bower phantomjs-prebuilt
   - bower --version
-  - npm install phantomjs-prebuilt
-  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
+  - phantomjs --version
 
 install:
   - npm install
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -12,10 +12,9 @@ cache:
 
 before_install:
   - npm config set spin false
-  - npm install -g bower
+  - npm install -g bower phantomjs-prebuilt
   - bower --version
-  - npm install phantomjs-prebuilt
-  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
+  - phantomjs --version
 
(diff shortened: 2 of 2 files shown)


blueprints/addon/files/.travis.yml
```diff
@@ -27,8 +27,8 @@ before_install:
   - npm config set spin false
   - npm install -g bower
   - bower --version
-  - npm install phantomjs-prebuilt
-  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
+  - npm install -g phantomjs-prebuilt
```
Turbo87: can you combine that with the `npm install -g bower` call above? no need to invoke NPM twice :)

rwjblue:
Hmmm. Doesn't testem find the local node modules version first? Or does it only work if node_modules/.bin is in $PATH?

Turbo87:
I guess we'd have to ask @johanneswuerbach that 😉

johanneswuerbach:
Newer versions of testem should prefer locally installed modules. Which testem version were you using?

blueprints/addon/files/.travis.yml
kellyselden: done

kellyselden:
So it looks like we have an issue with ember-try clearing out a local phantom install because it's not in the package.json.

This PR can be treated as a temp fix to get people back on phantom 2.1 while we figure out the real issue, or we can wait until the real fix comes along. I vote for temp fix. Thoughts?

Turbo87:
@kellyselden actually I think installing this globally is the right thing to do, not just a temp fix

kellyselden:
@nathanhammond or @Turbo87 can you verify I targeted this properly?

Turbo87:
@kellyselden yeah, should be good IMHO

rwjblue:
@homu r+

nathanhammond:
LGTM

PR #6537, opened 2016-12-07
Prevent deprecation from `ember-cli-babel` config options.

rwjblue:
Future versions of ember-cli-babel will be moving the location for its own configuration options out of `babel` and will be issuing a deprecation if used in the older way.

This changes things around to detect the correct (undeprecated) options location, and use it when possible.  Otherwise it falls back to `babel` (which is backwards compatible without a deprecation and forwards compatible with a deprecation).

/cc @Turbo87 

Reference: https://github.com/babel/ember-cli-babel/pull/105

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -185,18 +185,6 @@ EmberApp.prototype._initProject = function(options) {
   @param {Object} options
 */
 EmberApp.prototype._initOptions = function(options) {
-  var babelOptions = {};
-
-  if (this._addonInstalled('ember-cli-babel')) {
-    var amdNameResolver = require('amd-name-resolver').moduleResolve;
-    babelOptions = {
-      compileModules: true,
-      modules: 'amdStrict',
-      moduleIds: true,
-      resolveModuleSource: amdNameResolver
-    };
-  }
-
   var appTree = new WatchedDir(this._resolveLocal('app'));
 
   var testsPath = this._resolveLocal('tests');
@@ -217,8 +205,12 @@ EmberApp.prototype._initOptions = function(options) {
   var publicPath = this._resolveLocal('public');
   var publicTree = existsSync(publicPath) ? new WatchedDir(publicPath) : null;
 
-  this.options = defaultsDeep(options, {
-    babel: babelOptions,
+  var detectedDefaultOptions = {
+    babel: {
+      modules: 'amdStrict',
+      moduleIds: true,
+      resolveModuleSource: require('amd-name-resolver').moduleResolve
+    },
     jshintrc: {
       app: this.project.root,
       tests: this._resolveLocal('tests')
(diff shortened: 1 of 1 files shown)


nathanhammond:
I'm unsure if we want to target `beta` for this.

Turbo87:
@nathanhammond otherwise https://github.com/babel/ember-cli-babel/blob/master/index.js#L20 will blow up

nathanhammond:
(This supports the config location change, but I kinda want to split `ember-source` and the Babel updates releases.)

Turbo87:
@nathanhammond note that this will not be followed by anything in Babel 6 direction. this is just meant to get the deprecation out there ASAP.

rwjblue:
@nathanhammond - Confirm, we definitely do not want to try and land babel@6 in 2.11, that will (hopefully) be for 2.12.  But I do think we should get the deprecation out there ASAP, so folks have a little longer to digest it...

rwjblue:
@nathanhammond - Please confirm 👍 / 👎 on this, I don't want to land over your objection (if you are objecting)...

lib/broccoli/ember-app.js
```diff
@@ -217,8 +205,12 @@ EmberApp.prototype._initOptions = function(options) {
   var publicPath = this._resolveLocal('public');
   var publicTree = existsSync(publicPath) ? new WatchedDir(publicPath) : null;
 
-  this.options = defaultsDeep(options, {
-    babel: babelOptions,
+  var detectedOptions = {
```
nathanhammond: Why are we calling this `detectedOptions`? It feels like it's an extension to the default option set. The thing is that we're looking at project setup to figure these out. I just want a better name or a comment.

lib/broccoli/ember-app.js
```diff
@@ -250,7 +242,21 @@ EmberApp.prototype._initOptions = function(options) {
...
+
+  var emberCLIBabelInstance = this.project.findAddonByName('ember-cli-babel');
+  if (emberCLIBabelInstance) {
+    // future versions of ember-cli-babel will be moving the location for its
+    // own configuration options out of `babel` and will be issuing a deprecation
+    // if used in the older way
+    //
+    // see: https://github.com/babel/ember-cli-babel/pull/105
+    var emberCLIBabelConfigKey = emberCLIBabelInstance.configKey || 'babel';
+    detectedOptions[emberCLIBabelConfigKey] = detectedOptions[emberCLIBabelConfigKey] || {};
```
nathanhammond: Can we not just go ahead and set both `ember-cli-babel` and `babel` inside of the `detectedOptions` hash? This feels obfuscating.

rwjblue: @nathanhammond - Well, we can always set `detectedOptions['ember-cli-babel'].compileModules = true`, but we **cannot** always set `detectedOptions.babel.compileModules = true` (becuase this is exactly what triggers the deprecation).  The entire point here is that we have to figure out what config location to put ember-cli-babel's own config (defaulting to `babel` for backwards compat).

lib/broccoli/ember-app.js
rwjblue: Updated to what I mentioned above (`detectedDefaultOptions`)

nathanhammond:
:+1: once Travis is happy.

rwjblue:
CI is failing because ember-cli-babel@5.2.0 is broken when used with beta builds of ember-cli (which our CI "thinks" it is), upstream fix is https://github.com/babel/ember-cli-babel/pull/106.

Note: this will affect all PR's to beta or master branches.

rwjblue:
@homu r+

PR #6538, opened 2016-12-07
Fixups for recent two pass babel changes.

rwjblue:
We started testing out master in a very large app at work, and ran into a few small issues with the changes in #6530.  

This PR makes the following changes:

* Makes the warnings much more useful (and prevents an error that was occurring just when reading config to determine if we should warn).
* Cherry pick change from #6480, as it is now manifesting (due to internal changes in `modules/` shimming, the `overwrite` that was happening at a later stage, is now happening in `compileAddon`).
* Ensure a full pass of babel transpilation is done in the fallback case (when an addon with `addon/**/*.js` files has no JS preprocessors). This is required due to the way the migration from esperanto to babel occurred where the second pass (that was intended for only modules transpilation) was actually implemented as a full babel transpile step.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -147,12 +147,12 @@ var Addon = CoreObject.extend({
       throw new SilentError('An addon must define a `name` property.');
     }
 
-    this.options = defaultsDeep(this.options, {
+    this.__originalOptions = this.options = defaultsDeep(this.options, {
       babel: DEFAULT_BABEL_CONFIG
     });
 
     var emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
-    this.options[emberCLIBabelConfigKey] = defaultsDeep(this.options[emberCLIBabelConfigKey], {
+    this.__originalOptions[emberCLIBabelConfigKey] = this.options[emberCLIBabelConfigKey] = defaultsDeep(this.options[emberCLIBabelConfigKey], {
       compileModules: true
     });
   },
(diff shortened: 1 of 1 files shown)


Turbo87:
seems good on first glance

nathanhammond:
@bcardarella's fix is likely just fixing the symptom, not the problem. See: #5336. In favor of landing in spite of that.

rwjblue:
Yes, agreed there are more issues at play in #5336 for sure. I spent some time a while ago debugging that, and have a decent picture of what is wrong. I'll try to comment over there to give guidance.

rwjblue:
@homu r+

cibernox:
@rwjblue I lack some context, but I understand that some addons was (why?) deleting the babel configuration so the transpilation failed.

- what addon?
- Is this my impression or the change I propose in https://github.com/ember-cli/rfcs/pull/89 (passing a deep clone of the babel options to addons) would also fix this situation, since whatever the addon does will not affect any other addon/app?

rwjblue:
Before the changes in #6530 `this.options` was not set in an addon's `init`. We had some in-repo addons that were clobbering `this.options` completely in the `included` hook. I assumed that this might also be possible in public addons so I updated to gracefully handle this situation and provide a warning. 

The RFC wouldn't have solved this specific scenario.

bcardarella:
@nathanhammond I agree, I think however that the correct path forward is a more comprehensive refactor of the entire build pipeline. We have file concat's happening throughout that maybe aren't ideal. I believe it would be best to have all files exposed in their original directory nested form in the same tree before their final phases of transpilation, funneling, then concat. This final "mega tree" should allow for a hook that can be used to make any compile-time changes to the state of that tree. Right now if you want to make compile-time optimizations it is impossible to get a comprehensive view of all project files (vendor and app).

PR #6540, opened 2016-12-07
removing jshint reference in blueprints

kellyselden:
Response to https://github.com/ember-cli/ember-cli/pull/6487#issuecomment-265608329

There are still a lot of usages in the tests and fixtures, which may or may not be dead.

Removing these will probably make sticking with JSHint more difficult, unless that addon adds these ignores back. But the alternative is leaving our references to jshint forever?

cc @Turbo87


nathanhammond:
@kellyselden You got _way_ more than just blueprints here. I think we still want the ones for Ember CLI itself? I think eslint uses those too? (@turbo87 can you confirm?)

Turbo87:
@kellyselden in other places we've done it like [this](https://github.com/ember-cli/ember-cli/blob/master/blueprints/server/index.js#L32-L36). maybe we can put that into the locals and remove the line if JSHint is not used?

I'd also be fine with removing them completely though, now that we're installing ESLint by default.

kellyselden:
> You got way more than just blueprints here. I think we still want the ones for Ember CLI itself?

You're right @nathanhammond, I kept going when I saw that we still had jshint reference in our code as well after switching to eslint. I should have pointed that out. But since CI has passed and we lint our code in tests, all my changes should be safe?

twokul:
@Turbo87 @nathanhammond I rebased, is it good to 🚢 :shipit:

nathanhammond:
I'm in favor of this even more with our recent changes. Also, I believe that we also default remove JSHint on install now, so this should be safe? Want @Turbo87 to confirm.

Turbo87:
@nathanhammond we don't, but `ember-cli-eslint` does (see https://github.com/ember-cli/ember-cli-eslint/blob/master/blueprints/ember-cli-eslint/index.js#L36-L41)

Turbo87:
I'm wondering if we should replace `/*jshint node:true*/` with `/* eslint-env node */` instead...

blueprints/addon/files/npmignore
```diff
@@ -8,7 +8,6 @@
 .editorconfig
 .ember-cli
 .gitignore
-.jshintrc
```
Turbo87: we should probably add `.eslintrc.js` here

nathanhammond:
> `/* eslint-env node */`
Yeah, we should probably use that if browser/node is inter-mingled in the same directory. Otherwise we should do the right thing in `.eslintrc.js`.

> we don't, but ember-cli-eslint does - @Turbo87 
Confirm, and we run that blueprint on `ember init` in master, so I was counting it. 😜

lib/broccoli/app-config.js
```diff
@@ -1,7 +1,3 @@
-/* jshint ignore:start */
```
nathanhammond: I'm surprised that this gets through linting. Implies that maybe ESLint isn't running against 100% of Ember CLI. @Turbo87 Can you chime in here as to why this gets through linting?

Turbo87: Some files are ignored because they are not valid JS and they trip up the parser

nathanhammond: Oh, I looked for the ignore list in the wrong file. [Hello ignore list!](https://github.com/ember-cli/ember-cli/blob/master/.eslintignore)

blueprints/addon/files/index.js
```diff
@@ -1,4 +1,4 @@
-/* jshint node: true */
+/* eslint-env node */
```
nathanhammond: Is ignored.

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -1,4 +1,4 @@
-/*jshint node:true*/
+/* eslint-env node */
```
nathanhammond: Is ignored.

blueprints/addon/files/addon-config/environment.js
```diff
@@ -1,4 +1,4 @@
-/*jshint node:true*/
+/* eslint-env node */
```
nathanhammond: File is ignored.

blueprints/lib/files/lib/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
nathanhammond: Why are we adding this file? I feel like this belongs to `ember-cli-eslint`, but really we need a way for addons to hook into the base blueprints.

nathanhammond: Also, we've kept `.jshintrc` (Can't comment on a file with no changes.) Why?

twokul: why do we have `.jshintrc` there in the first place? :| I was trying to mimic what we had already going on with `JSHint`. Removing `.jshintrc` file is likely necessary but would require changing `blueprints/lib/index.js`. 

I can just remove `.eslintrc.js` and keep `.jshintrc` there and leave the change for another PR.

nathanhammond: I don't actually want a change here, we need extension points for blueprints that we don't have yet; I consider this a reasonable compromise.

nathanhammond:
@homu r+

Issue #6172, opened 2016-08-11
`validateAssetPath` should not throw error


luxferresum:
Hello

It seems like the feature was added to print the File sizes of the assets in a recent update.
We use an internal addin to postprocess the output in the `postBuild` hook to bundle everything into a single file for deployment. We do this in production builds only.

This change now makes ember throwing errors, which is a bit ugly.

Is there a reason that ember-cli throws the error and not just ignores it? Its not really an error to have no assets at that point.

Turbo87:
@luxferresum can you be a bit more specific when and why the function started throwing errors? how can we reproduce it?

luxferresum:
I'm talking about [this](https://github.com/ember-cli/ember-cli/blob/master/lib/models/asset-size-printer.js#L72) message.

To reproduce you just need a simple postBuild hook like this:

    postBuild(options) {
      // in reality here I would read all files and process them for deployment
      rimraf.sync(options.directory+'**/*');
      fs.writeFileSync(path.join(options.directory, 'deployfile.dat'), 'dummy file, in reality it includes all other files combined and packed ready for deployment');
    }

There are some usecases where the default folder structure can not be used, because of compatibility to legacy properitary software. So I really think this shouldnt be an error.

Turbo87:
@kiwiupover any comments on this?

PR #6541, opened 2016-12-08
Prints an error message instead of throwing an error

twokul:
fixes https://github.com/ember-cli/ember-cli/issues/6172

### DIFF
--- a/lib/models/asset-size-printer.js
+++ b/lib/models/asset-size-printer.js
@@ -14,17 +14,19 @@ AssetPrinterSize.prototype.print = function () {
   var ui = this.ui;
 
   return this.makeAssetSizesObject().then(function (files) {
-    this.validateAssetPath(files);
+    if (files.length !== 0) {
+      ui.writeLine(chalk.green('File sizes:'));
+      return files.forEach(function (file) {
+        var sizeOutput = filesize(file.size);
+        if (file.showGzipped) {
+          sizeOutput += ' (' + filesize(file.gzipSize) + ' gzipped)';
+        }
 
-    ui.writeLine(chalk.green('File sizes:'));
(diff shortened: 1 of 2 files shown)


stefanpenner:
Seems reasonable, although if possible a quick unit test would be lovely.

nathanhammond:
I suppose we're treating this as a "non-exceptional" error with this strategy. No, it didn't get you what you thought it would, but it _did_ run to completion.

I'm comfortable with that. @ember-cli/core?

Turbo87:
I'm fine with this changed, although given the new simplicity of the `validateAssetPath()` method I'd vote for inlining that method completely, otherwise the name might also be a bit misleading.

twokul:
@stefanpenner will add a test today

@Turbo87 sure or maybe rename to better communicate what it does

twokul:
@stefanpenner turns out we have decent code coverage and scenario with error message is tested. let me know if you want me to add additional tests.

stefanpenner:
@twokul good to hear there are lots of existing tests, is it possible though to add 1 additional scenario:

```
Prints an error message instead of throwing an error 
```

twokul:
@stefanpenner imo seems arbitrary? you wouldn’t write that kind of test if the original behaviour was to just print a message and keep going, basically testing behaviour that does not exist anymore. that said I can add a test, just wanted to make sure

stefanpenner:
@twokul alright, if you feel strongly the change in behavior isn't worth testing. I'm fine with skipping :)

twokul:
@stefanpenner added a test, good Sir. thanks for the review 😄

stefanpenner:
:shipit:  when green.

PR #6544, opened 2016-12-08
Update loader.js to newest version

ef4:
This release of loader.js improves the debugging experience when a module throws exceptions during its top-level eval. See https://github.com/ember-cli/loader.js/pull/90


stefanpenner:
👍

nathanhammond:
This is already in range, we don't typically push in changes that we get via SemVer drift. Given `yarn` adoption, however, we should probably start doing a better job.

nathanhammond:
@homu r+

PR #6545, opened 2016-12-08
Change livereload PortFinder.basePort to 49153

### LINKED ISSUE #6513: portfinder not working with Touchbar Macbook Pro
# Update

With help from @max & @nathanhammond this seems to have been narrowed down to some kind of incompatibility between portfinder and new Macbook Pros with the touchbar.

It seems to be reproducible by restarting the MBP, making sure that the lid is opened so the touchbar process (`eoshostd`) is using the default livereload port (49152).

I believe the intended behavior is that portfinder should detect that this port is in use and find another one that is available, which is not happening automatically or when forced using `--live-reload-port=0`.

# Original issue

After restarting my computer this morning I started running into the following whenever trying to run `ember server`:

```
Serving on http://localhost:4200/
Livereload failed on http://localhost:49152.  It is either in use or you do not have permission.
```

I'm not sure if something is running on this port or if it's a permissions error but I think it's a permissions error because running with `sudo` works.

Can somebody point me in the right direction for debugging this?

### Things I've Tried

- Restarting again
- Reinstalling node & anything node related
- Tried on multiple apps (with different ember-cli versions), including a brand new one
- Using sudo, it works.

##### Running live reload on a different port

This works but doesn't really fix anything.

```
> ember server --live-reload-port 49153

Livereload server on http://localhost:49153
Serving on http://localhost:4200/
```

##### Finding what's running on that port

I tried to find the process running on this port but I'm not sure if anything is? Since it works with `sudo`, it makes me think it's the permissions causing a problem.

```
> lsof -i:49152 # returns nothing
> lsof -n -i4TCP:49512 # returns nothing

# idk what this means or if it's helpful
> netstat -anp tcp | grep 49152
tcp6       0      0  fe80::aede:48ff:.49152 fe80::aede:48ff:.61000 ESTABLISHED
```

---

```
ember-cli: 2.10.0
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
3.10.8
```

OS X Sierra 10.12.1

eriktrom:
Fixes #6513

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -8,7 +8,7 @@ var PortFinder  = require('portfinder');
 var Win         = require('../utilities/windows-admin');
 var EOL         = require('os').EOL;
 
-PortFinder.basePort = 49152;
+PortFinder.basePort = 49153;
 
 var getPort = Promise.denodeify(PortFinder.getPort);
 var defaultPort = process.env.PORT || 4200;


eriktrom:
Manual test:

> ember s
Livereload server on http://localhost:49153

nathanhammond:
@HeroicEric Can you confirm that this works consistently if you make this change?

stefanpenner:
This doesn't hurt, so I'm merging. I suspect we may want to rejigger our automatic port finding, maybe @eriktrom in the new year we have a pairing/brainstorming session.

stefanpenner:
@homu

stefanpenner:
@homu r?

stefanpenner:
@homu r+

HeroicEric:
@nathanhammond seems to work. I tried restarting 3 times and haven't run into the issue. I'm not really sure how to test it any better.

Issue #6349, opened 2016-10-14
Add a clean way to get access to build timing


mdebbar:
Right now, the only way to get the build time information is by creating an addon and monkey patching `project.cli.analytics.trackTiming` like so:

``` javascript
module.exports = {
  name: 'build-time-tracker',
  init: function() {
    this._super.init && this._super.init.apply(this, arguments);

    var analytics = project.cli.analytics;
    var original = analytics.trackTiming;

    analytics.trackTiming = function(info) {
      original.apply(this, arguments);
      // Access `info` which contains meta data about current build, including build time.
    };
  },
};
```

It would be great to have a clean API to listen to build time info. Something as simple as this could work:

``` javascript
project.cli.onBuildComplete(function(info) {
  // Access the build info here..
});
```

Any thoughts?

hjdivad:
@mdebbar have you looked at the json output from using [`BROCCOLI_VIZ`](https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md#broccoli-viz).

You can see some examples of this being processed in [broccoli-slow-trees](https://github.com/rwjblue/broccoli-slow-trees/blob/master/calculate-summary.js#L20-L63).

What's your specific goal for which you want to monitor build times?

mdebbar:
I didn't know about that, thanks for the pointer. I tried it but I don't think I'll use broccoli viz, simply because it generates huge json and dot files (40+ MB) that I don't need. I just need the build time.

I found out that I can do`builder.on('start', function() { ... })` and `builder.on('end', function() { ... })` and capture `Date.now()` in both. Then calculate the build time as the difference.

My goal is to start tracking build times on all developers' machines. Then we can measure build time improvements more accurately.

stefanpenner:
we already have this tracking, but its likely not publicly augmentable. 

@twokul got any suggests on how to make leak pluggable here?

mdebbar:
@hjdivad, @stefanpenner in the future, we are planning to track the timing of sub-tasks (e.g. css compilation, babel, rollup, etc). Basically, the same info that's displayed as `Slowest Trees` by ember-cli. Is there a way to access that information without generating the gigantic broccoli-viz files?

stefanpenner:
@mdebbar I would recommend looking at where the slow tress are calculated today, that should most likely give you want you want.

twokul:
@stefanpenner let me think about it. it might require some changes to ember cli

nathanhammond:
[...]

PR #6546, opened 2016-12-09
Improve Internal build summary hooks

hjdivad:
## Todo
- [x] update to use `EMBER_CLI_INSTRUMENTATION` ~~instead of `BROCCOLI_VIZ`~~
- [x] ~~update perf doc re: `EMBER_CLI_INSTRUMENTATION`~~

## Questions
- [x] actually include `summary.output`? (seems fine, doesn't hurt anyone)
- [x] `summary.nodes` count all heimdall nodes or only broccoli nodes?
- [x] precompute `summary.totalTime`? ( its already coming from upstream)

## Overview

paired with @stefanpenner 

Improved analytics output, largely spike on missing infrastructure to make #6349
possible.

The intent is to provide a private API endpoint with a reasonable API, iterate,
and then see if we want to make this public.

## API

The single entry point will be an initially private api hook,
`_buildInstrumentation`, inside an addon's main entry file (typically
`index.js`).

This hook is invoked if the environment variable `EMBER_CLI_INSTRUMENTATION` is
set.

The hook is passed an analytics object that has two properties:
  
  1. `summary` - A POJO containing build information and
  2. `buildTree` A tree representation of the build pipeline, with a tree and
     stats iteration api.

### Hook

The hook is invoked if the environment variable `EMBER_CLI_INSTRUMENTATION` is
set.

```js
// index.js
module.exports = {
  // _ to indicate this hook is private and potentially unstable for now
  _buildInstrumentation(instrumentation) {
  }
}
```

### instrumentation argument

```js
{
  summary: /* described below */
  buildTree: /* described below */
}
```

#### Summary

The summary POJO has the following shape:

```js
{
  summary: {
    build: {
      type,
      count,
      outputChangedFiles,

      primaryFile,
      changedFileCount,
      changedFiles,
    },
    output,
    totalTime,
    buildSteps,
  }
}
```

The properties are as follows:

- `summary.build.type` one of `'initial'` or `'rebuild'`
- `summary.build.count` the number of the build (0 for initial build, > 0 for
  rebuilds).
- `summary.build.outputChangedFiles` an array of paths to output files that
  changed during this build.  These paths are relative to the `dist` directory.
- `summary.build.primaryFile` only present for rebuilds.  Indicates the first
  file the watcher noticed had changed.
- `summary.build.changedFileCount` only present for rebuilds.  The number of
  files the watcher had noticed changed before the build started.
- `summary.build.changedFiles` only present for rebuilds. The first 10 files
  the watcher had noticed changed before the build started.
- `summary.output` The temp directory containing the results of the build.
- `summary.totalTime`
- `summary.buildSteps` The number of build steps (broccoli nodes) in the build tree.

#### buildTree

The broccoli build pipeline can be thought of as a tree of broccoli nodes, as
well as more fine grained nodes (such as `applyPatches` and `computePatches`,
referring to fs-tree-diff patches).  The `buildTree` object represents the root
of this tree and provides APIs for iterating the tree and iterating stats

#### API

- `node.label`

Description: a POJO property that describes the node.  It will always include a `name`
property, and for broccoli nodes will also include a `broccoliNode` property.
Example:

```js
node.label == {
  name: 'TreeMerger (allTrees)',
  broccoliNode: true,
}
```

- `node.toJSON` 

Description: Returns a POJO that represents the serialized tree.  This POJO will
have a single key, `nodes`, an array of POJOs with the following properties:
- `id` an arbitrary node identifier.
- `children` an array of node identifiers for the children of this node
- `label` a POJO that describes this node.  Exactly as `node.label`.
- `stats` a POJO that describes the stats collected for this node.  It will
  include:
  - `time.self` - The time spent within this node (not including its
    descendants) in nanoseconds.
  - `fs` a POJO that includes the counts of all invocations of methods in
    `fs`, as well as the cumulative times spent in each method.
  - `own` any custom stats specific to this node.

Example:

  ```js
  // for a tree
  //  TreeMerger
  //    |- Babel_1
  //    |- Babel_2
  //    |--|- Funnel
  console.log(JSON.stringify(node.toJSON(), null, 2));
  // might print
  //
  {
    nodes: [{
      id: 1,
      children: [2,3],
      stats: {
        time: {
          self: 5000000,
        },
        fs: {
          lstat: {
            count: 2,
            time: 2000000
          }
        },
        own: {
        }
      }
    }, {
      // ...
    }]
  }
  ```

- `node.preOrderIterator(until)`
Description: An iterator that yields every node at the subtree rooted at `node`.
Nodes are yielded before their children.  If the optional matcher `until` is
passed, subtrees for which `until` returns something truthy will not be yielded.

Example:

  ```js
  // for a tree
  //  TreeMerger
  //    |- Babel_1
  //    |- Babel_2
  //    |--|- Funnel
  for (n of node.preOrderIterator()) {
    console.log(n.label.name);
  }
  // prints
  //
  // TreeMerger
  // Babel_1
  // Babel_2
  // Funnel

  
  for (n of node.preOrderIterator(x => x.label.name === 'Babel_2')) {
    console.log(n.label.name);
  }
  // prints
  //
  //  TreeMerger
  //  |- Babel_1
  ```

- `node.postOrderIterator(until)`
Description:  An iterator that yields every node at the subtree rooted at
`node`.  Nodes are yielded after their children.  If the optional matcher
`until` is passed, subtrees for which `until` returns something truthy will not
be yielded.

Example:
  ```js
  // for a tree
  //  TreeMerger
  //    |- Babel_1
  //    |- Babel_2
  //    |--|- Funnel
  for (n of node.postOrderIterator()) {
    console.log(n.label.name);
  }
  // prints
  //
  // Funnel
  // Babel_2
  // Babel_1
  // TreeMerger
  ```

- `node.ancestorsIterator`
Description:  An iterator that yields every ancestor node of `node`, in order
from `node` to the root node.  `node` itself is not yielded.

Example:

  ```js
  // for a tree
  //  TreeMerger
  //    |- Babel_1
  //    |- Babel_2
  //    |--|- Funnel
  //
  // on node Funnel
  for (n of node.ancestorsIterator()) {
    console.log(n.label.name);
  }
  // prints
  //
  // Babel_2
  // TreeMerger
  ```

- `node.statsIterator`
Description: An iterator that yields pairs of stat names and values.
Example:

  ```js
  //  for a typical broccoli node 
  for ([statName, statValue] of node.statsIterator()) {
    console.log(statName, statValue);
  }
  // prints
  //
  // "time.self" 64232794
  // "fs.statSync.count" 40
  // "fs.statSync.time" 401232123
  // ...
  ```


lib/models/builder.js
```diff
@@ -208,33 +227,41 @@ module.exports = CoreObject.extend({
     });
   },
 
+  _reportVizInfo: function (vizInfo) {
+    if (this.project && this.project.addons.length) {
+      this.project.addons.forEach(function (addon) {
+        if (addon._buildAnalytics) {
```
rwjblue: Should use a private symbol (something like Embers symbol) for now, until we have time for an RFC.

tests/unit/models/builder-test.js
```diff
@@ -19,7 +19,7 @@ var file = chai.file;
 var root            = process.cwd();
 var tmproot         = path.join(root, 'tmp');
 
-describe('models/builder.js', function() {
+describe.only('models/builder.js', function() {
```
rwjblue: .only

hjdivad: 👍

lib/models/builder.js
stefanpenner: lets just call this '__you_owe_stef_rob_and_dave_a_beer_or_something_if_you_invoke_thisBuildAnalytics`

rwjblue: 😂

mdebbar: Some companies use things like:
```javascript
'__DONT_USE_THIS_PRIVATE_KEY_OR_YOU_GET_FIRED'
```

lib/models/builder.js
```diff
@@ -33,19 +33,35 @@ function vizEnabled() {
...
       totalNodes: nodes.length,
     },
     nodes: nodes,
-  });
+  };
+
+  if (resultAnnotation.type === 'rebuild') {
+    vizJson.summary.build.primaryFile = resultAnnotation.primaryFile;
+    vizJson.summary.build.changedFileCount = resultAnnotation.changedFiles.length;
+    vizJson.summary.build.changedFiles = resultAnnotation.changedFiles.slice(0, 10);
```
mdebbar: I'm curious why not pass the whole list here?

lib/models/builder.js
```diff
@@ -208,33 +227,41 @@ module.exports = CoreObject.extend({
...
       })
+      .then(this.processAddonBuildSteps.bind(this, 'postBuild'))
+      .then(this.processBuildResult.bind(this))
       .then(function(result) {
-
         if (vizEnabled()) {
-          outputViz(buildCount++, result);
+          var vizJson = computeVizInfo(buildCount, result, resultAnnotation);
+          self._reportVizInfo(vizJson);
+          writeVizInfo(buildCount, vizJson);
```
mdebbar: It would be great if these 2 steps were controlled by different flags. If someone only needs to access the data programmatically, there is no reason to generate those files.

lib/models/builder.js
stefanpenner: In some cases (like a recrawl, or git branch change, of debase) it may be A very very large list (like every single file, which could be thousands or hundreds of thousands)

stefanpenner: Err rebase

lib/models/builder.js
stefanpenner: @rwjblue enabling this is a tad annoying in node. Do you have a recommendation?

lib/utilities/symbol.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+module.exports = function symbol(name) {
```
chadhietala: Is there a reason why we can't do a polyfill here?

lib/models/builder.js
stefanpenner: @rwjblue private symbol via `require('ember-cli/features/build-instrumentation');` 👍

lib/utilities/symbol.js
stefanpenner: The goal is for it to be a private symbol, so I copied what ember does. A full symbol polyfil, I believe is a tad trickier.

stefanpenner:
Some stuff:

- [ ] this changes the output of the ` broccoli-viz.<id>.json` files, should we make broccoli-viz support both? or..

lib/models/builder.js
stefanpenner: Now controlled by 2 flags:

* `BROCCOLI_VIZ` for the existing `broccoli-viz.<id>.json` file
* `EMBER_CLI_INSTRUMENTATION` for the new hook, once this becomes public. We will most likely make it a command line flag, so that it works via the `rc` files.

stefanpenner:
Tests:

- [x] _reportVizInfo
- [x] symbol
- [ ] instrumentationEnabled
- [ ] computeVizInfo
- [ ] build

hjdivad:
- [ ] rename features -> experiments (`ember-cli/lib/experiments.js`)

broccoli-viz.0.json
```diff
@@ -0,0 +1,20 @@
+{
```
rwjblue: Should be deleted

broccoli-viz.1.json
```diff
@@ -0,0 +1,20 @@
+{
```
rwjblue: ✂️

broccoli-viz.2.json
```diff
@@ -0,0 +1,20 @@
+{
```
rwjblue: 🔪

tests/unit/models/builder-test.js
```diff
@@ -11,18 +11,299 @@ var MockProject     = require('../../helpers/mock-project');
...
 var root            = process.cwd();
 var tmproot         = path.join(root, 'tmp');
 
-describe('models/builder.js', function() {
+var MockUI = require('console-ui/mock');
+var Heimdall = require('heimdalljs/heimdall');
+var walkSync = require('walk-sync');
+var itr2Array = require('../../helpers/itr2array');
+
+describe.only('models/builder.js', function() {
```
rwjblue: 💣

tests/unit/models/builder-test.js
```diff
@@ -293,6 +597,94 @@ describe('models/builder.js', function() {
...
+          expect(instrumentationArg).to.equal(mockVizInfo);
+        });
+      });
+
+      it('writes and invokes build instrumentation hook when BROCCOLI_VIZ=1', function() {
+        process.env.BROCCOLI_VIZ = '1';
+        var mockVizInfo = Object.create(null);
+        var mockResultAnnotation = Object.create(null);
+
+        builder._computeVizInfo = function (count, result, resultAnnotation) {
```
stefanpenner: I think we should use test double here:

how to use test double for this: https://github.com/ember-cli/ember-cli/pull/6461#discussion_r92225899

(future, we may want to document this in contributing.md)

hjdivad: 👍

rwjblue:
@homu r+

PR #6547, opened 2016-12-09
Removes deprecation messages from test suite

twokul:
"DEPRECATION: Overriding init without calling this._super is deprecated. Please call `this._super.init && this._super.init.apply(this, arguments);` addon: `undefined`"


nathanhammond:
Thanks @twokul! @homu r+

PR #6548, opened 2016-12-09
Properly set up package.json for later replacement.

nathanhammond:
We currently use the wrong information for setting up the package name inside of addons. We're generating a blueprint at invocation time in this blueprint, which means we should be populating it with the correct replacement placeholders.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -18,7 +18,7 @@ module.exports = {
     var contents = this._readContentsFromFile('package.json');
 
     delete contents.private;
-    contents.name = this.project.name();
+    contents.name = '<%= addonName %>';
     contents.description = this.description;
     contents.scripts = contents.scripts || {};
     contents.keywords = contents.keywords || [];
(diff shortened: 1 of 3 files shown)


Turbo87:
How is this actually different than before?

nathanhammond:
@Turbo87 If you specify `ember addon something --directory=something-else` `this.project.name() === 'something-else'`.

Turbo87:
I see, thanks!

kellyselden:
> If you specify ember addon something --directory=something-else then this.project.name() === 'something-else';.

It would be helpful to me and others if you can tell me why `something-else` is better choice for project name than `something`. Is it for module naming or importing purposes? What is the remaining use for `something` if not project name?

nathanhammond:
@kellyselden This fixes it so that we use `something`, not `something-else`. You're arguing for this fix. 😛

kellyselden:
Haha everyone carry on, it's Friday.

nathanhammond:
@homu r+

Turbo87:
I think what Kelly wanted to ask is why the project name was even something-else instead of just something

PR #6549, opened 2016-12-09
Allow using empty directories.

nathanhammond:
The usecase of `--directory` inside of Ember CLI allows you to specify where you wish to place the output from the `ember new` and `ember addon` commands. It's likely in those scenarios that the directory you're trying to populate already exists. Our goal was to make sure we played nicely with the existing assets. An empty directory meets that bar for safety and improves the ergonomics of the `--directory` command, also making it more suitable for testing.

### DIFF
--- a/lib/tasks/create-and-step-into-directory.js
+++ b/lib/tasks/create-and-step-into-directory.js
@@ -22,7 +22,7 @@ module.exports = Task.extend({
     var directoryName = this.directoryName = options.directoryName;
     if (options.dryRun) {
       return new Promise(function(resolve, reject) {
-        if (existsSync(directoryName)) {
+        if (existsSync(directoryName) && fs.readdirSync(directoryName).length) {
           return reject(this.warnDirectoryAlreadyExists());
         }
         resolve();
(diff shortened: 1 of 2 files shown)


kellyselden:
I like. Test?

nathanhammond:
Test already existed rolled up in the acceptance test world. Of course, I failed to edit it to prove lockout.

nathanhammond:
@Turbo87 Not according to the Internet or my manual review of the `fs` object. It makes sense that this is the case if you think about available POSIX functions.

nathanhammond:
This is a small, safe, change. It's tested. The scenario where we encounter the most cost should be the least common outcome of all possibilities. 

Here's [`readdir(3)`](http://man7.org/linux/man-pages/man3/readdir.3.html). Node.js doesn't actually implement [opendir(3)](http://man7.org/linux/man-pages/man3/opendir.3.html) which we could use to check the directory stream directly, so I'm shipping it for reasons of ergonomics.

@homu r+

stefanpenner:
If someone specified directory and it want empty I could even see us just applying regardless.  Or maybe that is a --force flag?

PR #6553, opened 2016-12-12
Fix wrong normalization in EmberApp._getAssetPath

jfhs:
I believe the intended behavior here was to replace all of `path.sep` to `/` but String.replace only replaces first match. This change should fix it.

I found it because on Windows, all assets imported with windows path separator were imported twice into output, and were in source map referenced by both windows and posix path. Maybe someone could look into why, if file has non-posix path, it's imported twice on Windows.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1636,7 +1636,7 @@ EmberApp.prototype._getAssetPath = function(asset) {
     return;
   }
 
-  assetPath = assetPath.replace(path.sep, '/');
+  assetPath = assetPath.split(path.sep).join('/');
 
   if (assetPath.split('/').length < 2) {
     console.log(chalk.red('Using `app.import` with a file in the root of `vendor/` causes a significant performance penalty. Please move `' + assetPath + '` into a subdirectory.'));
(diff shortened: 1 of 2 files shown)


stefanpenner:
@jfhs good find. We likely need to add a unit test to cover this. It might be a tad tricky to write a unit test as is. Maybe you can extract the line changed above to a function, which itself can be simply unit tested.

nathanhammond:
We can write a test that simply does `app.import` and checks the `outputFiles` object. Basically trigger the failing behavior that @jfhs discovered and rely on the fact that we do CI on Windows to confirm correctness.

It'll be _far_ easier to add an assertion to one of our existing acceptance tests.

jfhs:
Added test

nathanhammond:
@jfhs Thank you so much!

stefanpenner:
@homu r+

nathanhammond:
@homu r+

lib/broccoli/ember-app.js
```diff
@@ -1636,7 +1636,7 @@ EmberApp.prototype._getAssetPath = function(asset) {
     return;
   }
 
-  assetPath = assetPath.replace(path.sep, '/');
+  assetPath = assetPath.split('\\').join('/');
```
rwjblue: We should actually be using `ensure-posix-path` here, I believe (we already include it as a dependency).

Issue #6554, opened 2016-12-13
addon-test-support transpiled twice on master


trentmwillis:
Seems like there is an issue on `master` where `addon-test-support` files are getting transpiled twice. To reproduce:

```bash
ember new test-app
cd ./test-app
ember install ember-asset-loader # or any addon with addon-test-support contents
ember build
cat dist/assets/test-support.js # search for "ember-asset-loader"
```

You should find something like the following in the built output:

```js
define('ember-asset-loader/test-support/loaded-asset-state', ['exports'], function (exports) {
  'use strict';
  define('loaded-asset-state', ['exports', 'ember'], function (exports, _ember) {
    // elided...
  });
});
```

I'm going to dig in, but any tips on where to look would be helpful.

---

Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.11.0-beta.1-master-e7ec217113
http_parser: 2.7.0
node: 5.12.0
v8: 4.6.85.32
uv: 1.8.0
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 47
openssl: 1.0.2h
os: darwin x64
3.8.6

```

rwjblue:
Fixed by #6555

PR #6555, opened 2016-12-13
Don't run babel twice for addon-test-support

trentmwillis:
Addressing #6554.

~~I'm unsure this is the optimal solution. It seems like individual addons should handle transpiling their `addon-test-support` trees and the consuming app should just concatenate them. The problem with that approach is that it could be considered a breaking change at this point in time, if the consumer is hooking into `treeForAddonTestSupport` and not calling super (such as in [ember-cli-qunit](https://github.com/ember-cli/ember-cli-qunit/blob/master/index.js#L45-L48)).~~

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1394,9 +1394,8 @@ EmberApp.prototype.testFiles = function(coreTestTree) {
     overwrite: true,
     annotation: 'TreeMerger (addon-test-support)'
   });
-  var transpiledAddonTestSupportTree = new Babel(addonTestSupportTree, this._prunedBabelOptions());
 
-  var finalAddonTestSupportTree = new Funnel(transpiledAddonTestSupportTree, {
+  var finalAddonTestSupportTree = new Funnel(addonTestSupportTree, {
     allowEmpty: true,
     destDir: 'addon-test-support',
     annotation: 'Funnel: Addon Test Support'
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -638,14 +638,25 @@ var Addon = CoreObject.extend({
       return tree;
     }
 
-    var processed = this.preprocessJs(tree, '/', this.name, {
-      registry: this.registry
-    });
-
-    return new Funnel(processed, {
+    var namespacedTree = new Funnel(tree, {
       srcDir: '/',
       destDir: '/' + this.moduleName() + '/test-support'
     });
+
+    if (registryHasPreprocessor(this.registry, 'js')) {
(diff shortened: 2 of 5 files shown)


stefanpenner:
@rwjblue assigned you as a reviewer if thats ok.

lib/models/addon.js
```diff
@@ -638,11 +638,7 @@ var Addon = CoreObject.extend({
       return tree;
     }
 
-    var processed = this.preprocessJs(tree, '/', this.name, {
```
nathanhammond: We should leave the `preprocessJs` call in by default. This is definitely the path we want to take.

lib/broccoli/ember-app.js
```diff
@@ -1394,7 +1394,15 @@ EmberApp.prototype.testFiles = function(coreTestTree) {
     overwrite: true,
     annotation: 'TreeMerger (addon-test-support)'
   });
-  var transpiledAddonTestSupportTree = new Babel(addonTestSupportTree, this._prunedBabelOptions());
+
+  var transpiledAddonTestSupportTree
+  if (this._addonInstalled('ember-cli-babel')) {
```
nathanhammond: We want to call `preprocessJs` no matter what addons somebody has installed, so it was correct in `lib/models/addon`. We just need to have the back of people relying on the (previously) default babel transpile.

```javascript
var transpiledAddonTestSupportTree
if (!this._addonInstalled('ember-cli-babel') && "we didn't tell it to stop double transpiling") {
  if ("addonTestSupportTree has files") {
    this.ui.writeDeprecateLine('Must include ember-cli-babel if you want to transform the tree.');
  }
  transpiledAddonTestSupportTree = new Babel(addonTestSupportTree, this._prunedBabelOptions());
}
```

trentmwillis: The issue is that the transpilation either needs to happen only in `Addon#treeForAddonTestSupport` for all addons _or_ once only in `EmberApp#testFiles`, we can't do both.

If some addons have `ember-cli-babel` installed, then those addons will be double transpiled if we have to run `Babel` in the `EmberApp` again. The approach you outline above would trigger for a scenario where all but one addon is correctly preprocessing, causing everything except that one addon to double transpile.

trentmwillis:
@nathanhammond / @rwjblue updated. We'll now do the preprocessing in the addons and will just have to fix the few addons overriding the behavior (I'll open a PR for ember-cli-qunit soon).

rwjblue:
@homu r+

Thanks @trentmwillis!!!!

PR #6556, opened 2016-12-13
Acceptance Test Refactoring

nathanhammond:
This PR does a few things:

1. Makes all paths absolute. This is a first step toward enabling us to run tests outside of the Ember CLI folder.
2. Returns the location of the test fixture from the acceptance test helper to the acceptance test. This allows us to change the strategy without the test itself knowing how the helper functions. (Read: I plan to seamlessly swap in a faster fixture strategy.)
3. Makes the current tests pass using this new behavior changing as little as possible.

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -23,7 +23,7 @@ var expect = chai.expect;
 var dir = chai.dir;
 
 var addonName = 'some-cool-addon';
-var addonRoot = path.resolve(__dirname + '/../../tmp/' + addonName) + '/';
+var addonRoot;
 
 describe('Acceptance: addon-smoke-test', function() {
   this.timeout(450000);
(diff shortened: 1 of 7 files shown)


tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -69,7 +72,7 @@ describe('Acceptance: addon-smoke-test', function() {
 
   it('works in most common scenarios for an example addon', function() {
     return copyFixtureFiles('addon/kitchen-sink').then(function() {
-      var packageJsonPath = addonRoot + 'package.json'
+      var packageJsonPath = path.join(addonRoot, 'package.json');
```
stefanpenner: why are we adding the join here? `readJSONSync` already normalizes its input.

nathanhammond: I'm making this as consistent as possible. All paths are fully-qualified and normalized before being passed into anything upstream. It insulates us from future changes in a good way. For example, this was changed when it broke once the path I returned at `addonRoot` didn't have a trailing slash.

stefanpenner:
@homu r+

PR #6557, opened 2016-12-13
Fixes test-double warnings in our unit tests :christmas_tree:

twokul:
Warnings I was referring to:

```
Warning: testdouble.js - td.verify - test double was both stubbed and verified with arguments (), which is redundant and probably unnecessary. (see: https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double
```

[Why shouldn't we use `verify` and `when` at the same time?](https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double)

We were abusing `test-double` assertion a little bit. Of course, there's always [an option to turn warning off](https://github.com/testdouble/testdouble.js/blob/master/docs/C-configuration.md#tdconfig) but I feel like it's just hiding the problem. Besides, cleaning things up never hurts! 

cc @Turbo87

### DIFF
--- a/tests/unit/cli/cli-test.js
+++ b/tests/unit/cli/cli-test.js
@@ -43,7 +43,7 @@ function stubCallHelp() {
 
 function stubValidateAndRunHelp(name) {
   var stub = stubValidateAndRun(name);
-  td.when(stub(), {ignoreExtraArgs: true}).thenReturn('callHelp');
+  td.when(stub(), {ignoreExtraArgs: true, times: 1}).thenReturn('callHelp');
   return stub;
 }
 
@@ -154,7 +154,7 @@ describe('Unit: CLI', function() {
           var output = ui.output.trim();
           expect(output).to.equal('', 'expected no extra output');
 
-          td.verify(newCommand(), {ignoreExtraArgs: true, times: 1});
+          //td.verify(newCommand(), {ignoreExtraArgs: true, times: 1});
         });
       });
     });
--- a/tests/unit/commands/test-test.js
+++ b/tests/unit/commands/test-test.js
@@ -2,7 +2,7 @@
 
 var path           = require('path');
 var CoreObject     = require('core-object');
-var expect         = require('chai').expect;
+var expect         = require('../../chai').expect;
 var MockProject    = require('../../helpers/mock-project');
 var commandOptions = require('../../factories/command-options');
 var Promise        = require('../../../lib/ext/promise');
@@ -35,9 +35,13 @@ describe('test command', function() {
     td.replace(tasks.Test.prototype, 'run', td.function());
     td.replace(tasks.Build.prototype, 'run', td.function());
     td.replace(tasks.TestServer.prototype, 'run', td.function());
-    td.when(tasks.Test.prototype.run(), {ignoreExtraArgs: true}).thenReturn(Promise.resolve());
-    td.when(tasks.Build.prototype.run(), {ignoreExtraArgs: true}).thenReturn(Promise.resolve());
+    td.when(tasks.Test.prototype.run(), {ignoreExtraArgs: true, times: 1}).thenReturn(Promise.resolve());
+    td.when(tasks.Build.prototype.run(), {ignoreExtraArgs: true, times: 1}).thenReturn(Promise.resolve());
     td.when(tasks.TestServer.prototype.run(), {ignoreExtraArgs: true}).thenReturn(Promise.resolve());
+  });
+
+  afterEach(function() {
+    td.reset();
   });
 
   function buildCommand() {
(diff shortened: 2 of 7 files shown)


twokul:
what is the meaning of this `listen EADDRINUSE Error: listen EADDRINUSE`

tests/unit/commands/test-test.js
```diff
@@ -146,7 +147,11 @@ describe('test command', function() {
       return command.validateAndRun(['--path=tests']).then(function() {
         var captor = td.matchers.captor();
 
-        td.verify(tasks.Build.prototype.run(), {ignoreExtraArgs: true, times: 0});
+        tasks.Build.prototype.run().then(function() {
+          // dummy assertions to show that
```
stefanpenner: tricky promises

stefanpenner: @Turbo87 #someday

rwjblue:
IMHO, the warning is fundamentally wrong and forces tests to assume way too much about the system. I'll try to coordinate with @hjdivad and open a nicely detailed issue in the testdouble repo to start a more formal discussion.

tests/unit/cli/cli-test.js
```diff
@@ -154,7 +154,7 @@ describe('Unit: CLI', function() {
           var output = ui.output.trim();
           expect(output).to.equal('', 'expected no extra output');
 
-          td.verify(newCommand(), {ignoreExtraArgs: true, times: 1});
+          //td.verify(newCommand(), {ignoreExtraArgs: true, times: 1});
```
hjdivad: dead code

tests/unit/commands/test-test.js
```diff
@@ -50,10 +54,7 @@ describe('test command', function() {
     });
 
     it('builds and runs test', function() {
-      return command.validateAndRun([]).then(function() {
-        td.verify(tasks.Build.prototype.run(), {ignoreExtraArgs: true});
-        td.verify(tasks.Test.prototype.run(), {ignoreExtraArgs: true});
-      });
+      return command.validateAndRun([]);
```
hjdivad: What is this testing now?

It used to test that `validateAndRun` would call `run`, adding something like
```js
    if (true) {
      console.log('bye bye');
      return;
    }
```
to the top of `validateAndRun` does not cause this test to fail.

tests/unit/models/blueprint-test.js
```diff
@@ -365,20 +364,23 @@ describe('Blueprint', function() {
 
         expect(output).to.equal(testString);
 
-        td.verify(blueprint._printCommand(), {ignoreExtraArgs: true, times: 0});
+        td.verify(blueprint._printCommand(), {times: 0});
       });
 
       it('calls printCommand', function() {
```
hjdivad: Should we rename this test now?  It is no longer asserting that `printCommand` is called.

tests/unit/models/command-test.js
```diff
@@ -568,13 +572,16 @@ describe('models/command.js', function() {
         td.when(command._printCommand(), {ignoreExtraArgs: true}).thenReturn(' command printed');
       });
 
+      afterEach(function() {
+        td.reset();
+      });
+
       it('calls printCommand', function() {
```
hjdivad: Similarly we should probably rename this to something like `"prints the expected help string"` or something.

twokul:
@rwjblue I don't think the warning is necessarily wrong. There were [issues](https://github.com/testdouble/testdouble.js/issues/148) opened about exact same warning. General idea behind it (as far as I understood it) is to have isolated unit tests. It indeed doesn't make sense to stub something twice with `when` and `verify`. (`when` takes `{ times: 1 }` as a parameter).

tests/unit/commands/test-test.js
twokul: seemed to be broken in the first place, code below doesn't blow up either 😢 

```javascript
return command.validateAndRun([]).then(function() {
  expect(false).to.equal(true);
});
```

tests/unit/cli/cli-test.js
twokul: https://github.com/ember-cli/ember-cli/pull/6571

tests/unit/commands/test-test.js
twokul: if you run it individually, it breaks 🤔

tests/unit/cli/cli-test.js
hjdivad: 🎉

tests/unit/models/blueprint-test.js
hjdivad: makes sense! 👍

PR #6559, opened 2016-12-13
[ENHANCEMENT] Ignore jetbrains files

srvance:
I constantly finding myself adding JetBrains entries to `.gitignore` files in new and existing Ember projects. I know I'm not the only person using IntelliJ or WebStorm for Ember. This won't fix existing projects, but it will help eliminate this bit of recurring overhead for future individual and public projects.

I added it both to the project `.gitignore` and the `gitignore` in the app blueprint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ npm-debug.log
 .deps-tmp
 /coverage
 *.pem
+
+# JetBrains files
+.idea/
+*.iml
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -15,3 +15,7 @@
 /libpeerconnection.log
 npm-debug.log*
 testem.log
+
+# JetBrains files
+.idea/
+*.iml


twokul:
not sure I agree with this change, seems like JetBrains files should be a part of your global `.gitignore` if that's the editor of your choosing. 

@ember-cli/core thoughts?

srvance:
While there's no real cost to making it part of the blueprints, I can understand why you wouldn't want to get into the business of chasing editor-specific configs. Frankly, I forgot about the global `.gitignore`. I'd love to have it in, but totally understand and won't be heartbroken if you reject it. If it were more than a trivial change, I would have asked first.

stefanpenner:
@srvance thank you for your contribution, but I believe @twokul its correct.

Some history: At one point (in the early days) we did maintain a list of common user files to ignore. That became unmanageable and the current state is the attempt at a balance. It isn't perfect (im not sure why we include `sass` tmp files today...), but I believe we our stance on not including editor specific ignore files is reasonable.

Hopefully that seems reasonable?

srvance:
Completely.

PR #6567, opened 2016-12-14
Add heimdalljs as dev dependency

hjdivad:
It was already a transitive dev dependency, which is why tests passed
with yarn (or even npm3).


hjdivad:
@nathanhammond @rwjblue seems pretty straightforwardly good?

nathanhammond:
LGTM! @homu r+

PR #6573, opened 2016-12-15
Implement addon treeFor caching.

rwjblue:
In Ember CLI today, all addons at each level are built through the standard `treeFor` / `treeFor*` hooks. These hooks are responsible for preprocessing the JavaScript included by the tree returned from that specific hook (e.g., `treeForAddon` preprocesses the JS for the addon tree). [This RFC](https://github.com/ember-cli/rfcs/pull/90) proposes a mechanism that would allow these returned trees to be cached by default (when no build time customization is done) and expose proper hooks for addon authors to control the degree to which we dedupe these trees.

See [ember-cli/rfcs#90](https://github.com/ember-cli/rfcs/pull/90) for more context.

*Note:* This effort was derived from @stefanpenner's work in https://github.com/ember-cli/ember-cli/issues/6429.

### TODO:

- [x] Unit test `cacheKeyForTree` implementation
  - [x] Confirm that `treeFor*` method overrides opt-out of caching
  - [x] Confirm that modifying `this.treeForMethods` opts out of caching.
  - [x] Confirm that repeated calls to `cacheKeyForTree` return the same value.
- [x] Unit test `treeFor` caching mechanism
- [x] Extract basic `cacheKeyForTree` implementation to stand alone utility function
   - [x] test

### Future TODO:

- [ ] Ensure `mergeTrees` macro utility function is updated to properly dedupe when called like `mergeTrees([treeA, treeA])`. (https://github.com/ember-cli/ember-cli/pull/6574)
- [ ] Updates to Broccoli to ensure that "reused / previously built trees" are represented properly in heimdall tree.  This will included confirming that `buildSteps` properly handles the deduplicated results and shows an improvement.
- [ ] Extract the default `cacheKeyForTree` implementation into a separate package.
- [ ] Add API docs for `Addon.prototype.cacheKeyForTree` (in a separate PR after the corresponding RFC lands, so that ember-cli.com/api does not show the experimental API's until the RFC lands).


stefanpenner:
@rwjblue as I believe this then supersedes my [early mega-hack WIP PR on this topic](https://github.com/ember-cli/ember-cli/pull/6429) should we close #6429 ?

rwjblue:
Updated and immediate checklist items completed.

lib/models/addon.js
```diff
@@ -8,7 +8,9 @@ var existsSync   = require('exists-sync');
 var path         = require('path');
 var assign       = require('ember-cli-lodash-subset').assign;
 var SilentError  = require('silent-error');
-var logger       = require('heimdalljs-logger')('ember-cli:addon');
+var heimdallLogger = require('heimdalljs-logger');
+var logger = heimdallLogger('ember-cli:addon');
```
trentmwillis: This logger doesn't seem to be used anywhere.

lib/models/addon.js
```diff
@@ -8,7 +8,9 @@ var existsSync   = require('exists-sync');
...
+var cacheLogger = heimdallLogger('ember-cli:addon:tree-cache');
```
trentmwillis: Should we rename this `treeCacheLogger`? Just realized this may be a bit generic

lib/models/addon.js
```diff
@@ -29,13 +31,59 @@ var ensurePosixPath = require('ensure-posix-path');
...
+});
 
 var DEFAULT_BABEL_CONFIG = {
   modules: 'amdStrict',
   moduleIds: true,
   resolveModuleSource: require('amd-name-resolver').moduleResolve
 };
 
+var DEFAULT_TREE_FOR_METHODS = {
+  app:               'treeForApp',
```
trentmwillis: Now that we've moved this, can we change the spacing to be a single space?

lib/models/addon.js
```diff
@@ -376,6 +414,30 @@ var Addon = CoreObject.extend({
   treeFor: function treeFor(name) {
     this._requireBuildPackages();
 
+    var node = heimdall.start({
+      name: 'treeFor(' + this.name + ' - ' + name + ')',
```
trentmwillis: The number of times the word `name` appears on this line is a bit funny to me

lib/models/addon.js
```diff
@@ -387,10 +449,20 @@ var Addon = CoreObject.extend({
...
 
-    return mergeTrees(trees, {
+    var mergedTreesForType = mergeTrees(trees, {
       overwrite: true,
       annotation: 'Addon#treeFor (' + this.name + ' - ' + name + ')'
     });
+
+    if (cacheKeyForTreeType) {
+      addonTreeCacheStats.adds++;
+      cacheLogger.info('Cache Add: ' + ' - ' + cacheKeyForTreeType + ' - ' + !!tree);
```
trentmwillis: Can we either add a comment or use a variable name (e.g., `hasTree`) to denote why we're logging `!!tree`?

tests/unit/utilities/addon-calculate-cache-key-for-tree-test.js
```diff
@@ -0,0 +1,43 @@
...
+      pkg: { some: 'parsed', value: { goes: 'here' }}
+    };
+
+    var first = calculateCacheKeyForTree('app', addon);
+    var second = calculateCacheKeyForTree('app', addon);
+
+    expect(first).to.equal(second);
+  });
+
+  it('should return different value when different', function() {
```
trentmwillis: This test doesn't seem to match this description. Also, I feel like the description should be something to the effect of `should return different values for different addon instances`?

lib/models/addon.js
nathanhammond: Yes please always.

lib/models/addon.js
```diff
@@ -29,13 +31,59 @@ var ensurePosixPath = require('ensure-posix-path');
...
+  this.adds = 0;
+});
 
 var DEFAULT_BABEL_CONFIG = {
   modules: 'amdStrict',
   moduleIds: true,
   resolveModuleSource: require('amd-name-resolver').moduleResolve
 };
 
+var DEFAULT_TREE_FOR_METHODS = {
```
nathanhammond: We add `treeForEngine` inside of `ember-engines` which we then invoke. Other addons that we have written also do `eachAddonInvoke('treeForSomeMadeUpThing')`. Depending upon when that gets called it doesn't necessarily go through `treeFor` as a choke point and could invalidate cache strategy.

lib/models/addon.js
```diff
@@ -1445,4 +1442,112 @@ Addon.lookup = function(addon) {
...
+  Constructor._meta_ = {
+    modulePath: modulePath,
+    lookupDuration: lookupDuration,
+    initializeIn: 0
+  };
+
+  return Constructor;
+};
 module.exports = Addon;
+module.exports._resetTreeCache = _resetTreeCache;
```
nathanhammond: What happens with something like `persistentFilter` or anything that uses a disk cache to repopulate their trees? I can't come up with a failure mode, but maybe somebody else can? (This thought is separate from the line I'm commenting on other than the fact that this one made me think of it.)

lib/utilities/addon-calculate-cache-key-for-tree.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var crypto = require('crypto');
+
+module.exports = function calculateCacheKeyForTree(treeType, addonInstance) {
+  var cacheKeyParts = [
+    addonInstance.pkg,
+    addonInstance.name,
+    treeType,
```
nathanhammond: Oh how I dislike the trailing comma. 😛

lib/utilities/addon-calculate-cache-key-for-tree.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var crypto = require('crypto');
+
+module.exports = function calculateCacheKeyForTree(treeType, addonInstance) {
+  var cacheKeyParts = [
+    addonInstance.pkg,
```
nathanhammond: This seems a bit overkill, but easier than addressing all the things which could possibly invalidate.

tests/unit/models/addon-test.js
```diff
@@ -844,4 +852,100 @@ describe('models/addon.js', function() {
...
+        });
+
+        it('returns null if `treeForMethods` is modified', function() {
+          var addon = createAddon(Addon.extend({
+            name: 'test-project',
+            root: 'foo',
+            init: function() {
+              this._super && this._super.init.apply(this, arguments);
+
+              this.treeForMethods['app'] = 'treeForWTF_ARE_YOU_THINKING!?!';
```
nathanhammond: Let's rename. I know, :nofunnathan:.

lib/models/addon.js
```diff
@@ -1445,4 +1442,112 @@ Addon.lookup = function(addon) {
...
 */
 
+function methodsForTreeType(treeType) {
+  var treeMethods = DEFAULT_TREE_FOR_METHOD_METHODS[treeType];
+
+  return GLOBAL_TREE_FOR_METHOD_METHODS.concat(treeMethods);
+}
+
+if (experiments.ADDON_TREE_CACHING) {
+  addonProto[experiments.ADDON_TREE_CACHING] = function cacheKeyForTree(treeType) {
```
nathanhammond: Can we add a comment about how to move this when we promote it from `experiments`? I had a mild "hey, we didn't actually put the cache behavior on the prototype!" panic moment and had to look back through the code for this.

nathanhammond: `if (experiments.ADDON_TREE_CACHING) {` => implies boolean
`addonProto[experiments.ADDON_TREE_CACHING] =` => implies key

These feel very weird adjacent to each other. Can we at least `!!experiments.ADDON_TREE_CACHING` to make it a bit more clear?

lib/models/addon.js
rwjblue: It is definitely used (this PR just moved things around a tad to avoid calling `require('heimdalljs-logger')` multiple times) in `initializeAddons`.

lib/models/addon.js
rwjblue: In this case, the default `Addon.prototype.cacheKeyForTree` implementation would detect that there is no default method on the `addonProto` object and `return null;` (causing it to be uncached).  This is *precisely* the scenario where an addon (ember-engines in this case) would want to define a custom `cacheKeyForTree` and ensure that it incorporates whatever things would cause the cache key to be invalidated properly.

lib/models/addon.js
rwjblue: This is **absolutely** intended to be used as a canary feature. We discussed this conceptually yesterday  (in person), and the idea is to make `lib/experiments/index.js` simply `module.exports = {};` when we branch to `beta`.  This allows users that want to experiment with the feature to do so on master, but makes it very clear that it is experimental (until https://github.com/ember-cli/rfcs/pull/90 lands).

See a tad more details in [ember-cli/rfcs#91](https://github.com/ember-cli/rfcs/blob/addon-instrumentation-experimental-hooks/active/0000-addon-instrumentation-experimental-hooks.md#experiments)...

rwjblue: Updated the guard section to use `!!` to make it clear that it is a guard.

lib/models/addon.js
rwjblue: `broccoli-persistent-filter` is all about caching across restarts, however the cache that we are working with here is not persistent across restarts.  The cache that this PR implements is strictly to ensure that the same addon (version/name/dependencies/etc) at multiple levels of depth (in the addon hierarchy) return the same tree.

_tldr;_ This cache cannot interact with things using on disk caches (since it doesn't survive a restart).

lib/utilities/addon-calculate-cache-key-for-tree.js
rwjblue: It is much simpler to simply checksum the full `package.json` here for sure.  It costs us little, and reduces the complexity of this quite a bit.

One thing to note (I added an inline comment for this):

>  In a future refactor, this function will be extracted and migrated to a
>  shared package. This allows addon authors to easily utilize the "default"
>  cache key generation system (e.g. if they implement `treeFor*` in a way that
>  is totally safe), but still allows us to identify issues with the cache key
>  generation and iterate (since both addon and ember-cli could use `^1.0.0`
>  and float independently).

rwjblue:
@nathanhammond:

> I think the biggest remaining concern I have is people using eachAddonInvoke in a way which creates side effects.

Perhaps sharing an example would make this a bit more concrete, but I'm not terribly concerned with this.  It is possible that we might never be able to guarantee 100% cache key safety (there are **LOTS** of whacky things people _could_ do), but I believe that this PR strikes a really good balance of practicality and properly handles [all of the cases](https://github.com/search?q=eachAddonInvoke&type=Code&utf8=%E2%9C%93) that I found in addons / github.

>  Given that it's marked private I'd be comfortable moving it to receive a symbol (similarly to our _concat change) which would give us a way to help guide people away from the pattern.

IMHO, we should actually be looking to **make** it public instead (yes, I realize this is likely RFC territory).  I think it is an excellent primitive for us to provide to addons...

lib/models/addon.js
rwjblue: Also, I realized I didn't address this concern in my comment above:

> Depending upon when that gets called it doesn't necessarily go through treeFor as a choke point and could invalidate cache strategy.

Yes, you are absolutely correct! If someone does `this.eachAddonInvoke('treeForEngine')`, it completely bypasses our ability to cache.  This is totally fine, they just don't automatically get the benefits of the new system that we have in place to speed things (like engines / fastboot / etc) up.

nathanhammond: Note that it also invalidates the caching strategy not for itself, but for `eachAddonInvoke`d. Whatever gets invoked could have side effects that change how something gets built. I'd like for us to try and find a way to warn when we catch this occurring.

nathanhammond:
> IMHO, we should actually be looking to **make** it public instead (yes, I realize this is likely RFC territory).  I think it is an excellent primitive for us to provide to addons...

I like it as a tool, I want to keep it private until we land this story and we land `ember-modules-app` (or the majority of pieces that make up the incremental improvements). It makes it easier for us to justify tweaking its semantics if we must. (Off the top of my head I can think of Highlander rule tweaks that opt out of all calls past the first ...)

rwjblue:
> I like it as a tool, I want to keep it private until we land this story and we land ember-modules-app (or the majority of pieces that make up the incremental improvements).

Ya, seems totally fine to wait, I was mostly just reacting to the idea that essentially deprecate it (which I am generally opposed to right now).

nathanhammond:
Concern voiced off thread, capturing here:

I'd prefer that we were able to automatically identify which things can invalidate caching so that we don't inevitably miss adding some method to `DEFAULT_TREE_FOR_METHOD_METHODS` or fail to identify a property that impacts the default `treeFor` hooks. For the version that is behind the `experiment` symbol I believe that we can maintain this list manually. Before we promote out of `experiment` status I want us to explore at the very least the effort/costs to achieve automation.

lib/models/addon.js
trentmwillis: Ah, missed that it was pre-existing. Carry on.

rwjblue:
@homu r+

PR #6575, opened 2016-12-16
Add instrumentation to mergeTrees

trentmwillis:
Adds logging and stat collection via heimdall.

cc @rwjblue

### DIFF
--- a/lib/broccoli/merge-trees.js
+++ b/lib/broccoli/merge-trees.js
@@ -1,6 +1,7 @@
 'use strict';
-
 var upstreamMergeTrees = require('broccoli-merge-trees');
+var heimdallLogger = require('heimdalljs-logger');
+var logger = heimdallLogger('ember-cli:merge-trees');
 var EMPTY_MERGE_TREE;
 
 function overrideEmptyTree(tree) {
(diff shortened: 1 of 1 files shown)


lib/broccoli/merge-trees.js
```diff
@@ -1,8 +1,20 @@
 'use strict';
-
 var upstreamMergeTrees = require('broccoli-merge-trees');
+var heimdall = require('heimdalljs');
+var heimdallLogger = require('heimdalljs-logger');
+var logger = heimdallLogger('ember-cli:merge-trees');
 var EMPTY_MERGE_TREE;
 
+heimdall.registerMonitor('ember-cli-merge-trees', function MergeTreesSchema() {
```
hjdivad: since it looks like this monitor isn't used outside of the node created on line 48 it would also be fine to pass this in as a schema in `heimdall.start` instead of as a monitor

trentmwillis: Was unaware of this feature. Updated.

rwjblue:
Crappola :( I meant to use @homu for this.

Sorry 😢

PR #6576, opened 2016-12-16
Don't install things which will be linked.

nathanhammond:
This came out of a discussion of how to support the new `experiment` API in a way that didn't troll our release process. `PackageCache` now opts you out of default installation and manages it itself.

In short, this PR:
- Creates an API in PackageCache which prevents it from installing a package which we're instead going to link in.
- Enables us to not have a valid package-manager-installable version specified for packages which we're going to link in.
- Refactors `PackageCache` to be completely extractable.
- Makes it possible to use the system-installed package managers rather than bundled package managers.

/cc @rwjblue @hjdivad

### DIFF
--- a/tests/helpers/acceptance.js
+++ b/tests/helpers/acceptance.js
@@ -9,14 +9,6 @@ var root = path.resolve(__dirname, '..', '..');
 
 var PackageCache = require('../../tests/helpers/package-cache');
 var CommandGenerator = require('../../tests/helpers/command-generator');
-var npm = new CommandGenerator(require.resolve('npm/bin/npm-cli.js'));
-
-// Module scope for linking Ember CLI itself. Must happen once.
-var linked = false;
-if (!linked) {
-  linked = true;
-  npm.invoke('link', { cwd: root });
-}
 
 var quickTemp = require('quick-temp');
 var dirs = {};
@@ -92,7 +84,7 @@ function linkDependencies(projectName) {
   var nodeManifest = fs.readFileSync(path.join(runFixture, 'package.json'));
 
   var packageCache = new PackageCache(root);
-  packageCache.create('node', 'npm', nodeManifest, ['ember-cli']);
+  packageCache.create('node', 'npm', nodeManifest, [{ name: 'ember-cli', path: root }]);
 
   var nodeModulesPath = path.join(runFixture, 'node_modules');
   symlinkOrCopySync(path.join(packageCache.get('node'), 'node_modules'), nodeModulesPath);
--- a/tests/helpers/acceptance.js
+++ b/tests/helpers/acceptance.js
@@ -8,6 +8,16 @@ var Promise           = require('../../lib/ext/promise');
 var root = path.resolve(__dirname, '..', '..');
 
 var PackageCache = require('../../tests/helpers/package-cache');
+var CommandGenerator = require('../../tests/helpers/command-generator');
+var npm = new CommandGenerator(require.resolve('npm/bin/npm-cli.js'));
+
+// Module scope for linking Ember CLI itself. Must happen once.
+var linked = false;
+if (!linked) {
+  linked = true;
+  npm.invoke('link', { cwd: root });
+}
+
 var quickTemp = require('quick-temp');
 var dirs = {};
 
@@ -81,8 +91,8 @@ function linkDependencies(projectName) {
 
   var nodeManifest = fs.readFileSync(path.join(runFixture, 'package.json'));
 
-  var packageCache = new PackageCache({ linkEmberCLI: true });
-  packageCache.create('node', 'npm', nodeManifest);
+  var packageCache = new PackageCache(root);
+  packageCache.create('node', 'npm', nodeManifest, ['ember-cli']);
 
   var nodeModulesPath = path.join(runFixture, 'node_modules');
   symlinkOrCopySync(path.join(packageCache.get('node'), 'node_modules'), nodeModulesPath);
--- a/tests/helpers/package-cache.js
+++ b/tests/helpers/package-cache.js
@@ -506,7 +506,7 @@ PackageCache.prototype = {
     links = links || [];
 
     // Save metadata about the PackageCache invocation in the manifest.
-    var packageManagerVersion = require(type + '/package.json').version;
+    var packageManagerVersion = commands[type].invoke('--version').stdout;
 
     var jsonManifest = JSON.parse(manifest);
     jsonManifest._packageCache = {
--- a/tests/helpers/package-cache.js
+++ b/tests/helpers/package-cache.js
@@ -6,6 +6,7 @@ var quickTemp = require('quick-temp');
 var Configstore = require('configstore');
 var CommandGenerator = require('./command-generator');
 var stableStringify = require('json-stable-stringify');
+var symlinkOrCopySync = require('symlink-or-copy').sync;
 
 var originalWorkingDirectory = process.cwd();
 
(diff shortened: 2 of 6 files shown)


tests/helpers/acceptance.js
```diff
@@ -8,6 +8,16 @@ var Promise           = require('../../lib/ext/promise');
 var root = path.resolve(__dirname, '..', '..');
 
 var PackageCache = require('../../tests/helpers/package-cache');
+var CommandGenerator = require('../../tests/helpers/command-generator');
+var npm = new CommandGenerator(require.resolve('npm/bin/npm-cli.js'));
+
+// Module scope for linking Ember CLI itself. Must happen once.
+var linked = false;
+if (!linked) {
```
nathanhammond: I hilariously failed at this.

tests/helpers/acceptance.js
```diff
@@ -8,6 +8,12 @@ var Promise           = require('../../lib/ext/promise');
 var root = path.resolve(__dirname, '..', '..');
 
 var PackageCache = require('../../tests/helpers/package-cache');
+var CommandGenerator = require('../../tests/helpers/command-generator');
+
+// We must link `ember-cli`.
+var npm = new CommandGenerator(require.resolve('npm/bin/npm-cli.js'));
+npm.invoke('link', { cwd: root });
```
rwjblue: Would it be possible to instead defer this to the `PackageCache`?  I'd love to be able to test different `ember-cli` builds at the same time (or test locally without completely borking my ability to develop normal apps / addons)...

What if we had a `links` in the `PackageCache` manifest as an array of objects (with `name` and `destination` keys)?

tests/helpers/package-cache.js
```diff
@@ -3,20 +3,18 @@
...
+var originalWorkingDirectory = process.cwd();
 
 // Module scoped variable to store whether a particular cache has been
 // attempted to be upgraded.
 var upgraded = {};
 
-// Module scoped variable set on first Ember CLI link attempt per type.
-var isEmberCLILinked = {};
+// List of keys which can result in things being installed.
+var DEPENDENCY_KEYS = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies'];
```
nathanhammond: This array of items was already present in `_checkManifest`, now extracted as a "`const`". It's intended to be the complete set of things which could _possibly_ result in the package manager installing something. This is the "err on the side of caution" approach. It actually doesn't matter if something is or isn't automatically installed in any of the cases where we use this.

nathanhammond: Adding my explanation somewhat verbatim. 😛

tests/helpers/package-cache.js
```diff
@@ -183,48 +181,42 @@ function checkDowngrade(type) {
  *
  * ```
  * var CommandGenerator = require('./command-generator');
- * var yarn = new CommandGenerator(require.resolve('yarn/bin/yarn.js'), {
- *   retryCommands: ['install', 'upgrade']
- * });
+ * var yarn = new CommandGenerator(require.resolve('yarn/bin/yarn.js'));
```
nathanhammond: Good catch. This is the constant documentation/code drift problem.

tests/helpers/package-cache.js
```diff
@@ -309,11 +300,14 @@ PackageCache.prototype = {
    * @param {String} manifest The contents of the manifest file to write to disk.
    */
   _writeManifest: function(label, type, manifest) {
+    process.chdir(this.rootPath);
```
nathanhammond: `this.rootPath === originalWorkingDirectory` unless the user manually passes in a `rootPath` argument. I'm comfortable with things erroring out if the user specifies a non-existent folder. It'll be the correct error message and everything. I don't know what else might break? User deletes `cwd`? This will be the least of your problems.

I'd rather the API be something like passing in a `rootPath`, but I don't care too much. This is "fine."

tests/helpers/package-cache.js
```diff
@@ -309,11 +300,14 @@ PackageCache.prototype = {
...
     var outputDir = quickTemp.makeOrReuse(this.dirs, label);
+    process.chdir(originalWorkingDirectory);
+
     this._conf.set(label, outputDir);
 
     var outputFile = path.join(outputDir, translate(type, 'manifest'));
-    fs.writeFileSync(outputFile, manifest);
+    fs.outputFileSync(outputFile, manifest);
```
nathanhammond: This started erroring on me and I looked it up. "Surely `mkdirp` + `fs.writeFileSync` exists?"

tests/helpers/package-cache.js
```diff
@@ -429,26 +508,22 @@ PackageCache.prototype = {
    * @param {String} label The label for the cache.
    * @param {String} type The type of package cache.
    * @param {String} manifest The contents of the manifest file for the cache.
+   * @param {Array} links Packages to elide for install and link.
```
nathanhammond: elide => omit

tests/helpers/package-cache.js
```diff
@@ -429,26 +508,22 @@ PackageCache.prototype = {
...
-    // Easier to do this every time than try and lock it out.
-    if (!isEmberCLILinked[type] && this.options.linkEmberCLI === true && (type === 'npm' || type === 'yarn')) {
-      commands[type].invoke('link', { cwd: emberCLIPath });
-      isEmberCLILinked[type] = true;
-    }
+    links = links || [];
 
     // Save metadata about the PackageCache invocation in the manifest.
-    var packageManagerVersion = require(type + '/package.json').version;
+    var packageManagerVersion = commands[type].invoke('--version').stdout;
```
nathanhammond: I elected not to run it through my `translate` lookup since it is consistent across all three package managers currently supported. It seemed like an unnecessary layer of indirection. If we ever add a package manager that needs something different this will get updated.

Going with "YAGNI" here.

tests/helpers/acceptance.js
nathanhammond: The original version of `PackageCache` handled the linking internally. However, the refactor here makes it 100% extractable which means consuming code should be responsible for managing its own links. That should remain the case, but `PackageCache` should enable consumer to be successful.

As it stands you're correct: because it relies on a single global link it doesn't enable you to test two builds simultaneously. I believed I had addressed all of those issues with moving the `package-cache.json` config but didn't think through the global `link` use case.

[For those following along, `cd ~/myproject && npm link ~/repos/ember-cli` actually creates a global link for `ember-cli` and then in `myproject` runs `npm link ember-cli`.](https://docs.npmjs.com/cli/link)

The only way to support multiple links is to do it _outside_ of the package manager itself, but that also gets thorny since `npm` possibly runs an `install` beforehand:
- [`npm` will `install` and build your package.](https://github.com/npm/npm/blob/latest/lib/link.js#L157-L165)
- [`yarn` just symlinks you.](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/cli/commands/link.js#L58)
on the code on `npm link`
- [`bower` just symlinks you.](https://github.com/bower/bower/blob/master/lib/commands/link.js)

Given that I can't (safely) blindly `link` you I elected to defer to the package manager's CLI. Thoughts?

nathanhammond:
Some of this has been addressed elsewhere.

---

> `DEPENDENCY_KEYS` includes `optionalDependencies`, but no tests confirm it works (or how it interacts if you have the same thing at different versions for all of `dependencies`, `devDependencies`, and `optionalDependencies`). Adding `peerDependencies` to `DEPENDENCY_KEYS` is likely not what we want to do. `npm@3` and higher no longer auto-installs peer deps, and I _believe_ `yarn` is following the `npm >= 3` behavior here.

No special code path exists for any of those keys, they're collected as a list of things which _might_ install something. Doesn't matter if they do or not. It's used for manifest invalidation and prevention of installation if things will be `link`ed in.

> Do we simply want to `ln -s` instead of using `npm link` / `yarn link`?  It is much easier to reason about (for me at least) and allows linking things that are not globally linked already?

[The discussion for this should continue here.](https://github.com/ember-cli/ember-cli/pull/6576#discussion_r92963153)

> I am unsure of the format for `links` in the manifest.  It seems to me that you really want both the name and the actual target, otherwise how does `yarn link ${name}` / `npm link ${name}` actually work unless the system had a global version already linked?

[`cd ~/myproject && npm link ~/repos/ember-cli` actually creates a global link for `ember-cli` and then in `myproject` runs `npm link ember-cli`.](https://docs.npmjs.com/cli/link) There is only global linking if you rely on the package manager itself. As such, the format for links is just the name of the package since nothing else actually matters inside of the (current) `PackageCache` sphere of influence.

tests/helpers/acceptance.js
nathanhammond: This discussion has been concluded: implemented a custom manual linking story inside of `PackageCache` which doesn't aim to imitate the package managers, just get "close enough" with running linking. Consumer is still responsible for making all things work.

Barring needing a rebase, here's the implementation: https://github.com/ember-cli/ember-cli/pull/6576/commits/e8d1ce0432095f54e543fa907f1c3460073ebc9b

rwjblue: :+1:

rwjblue:
@homu r+

PR #6577, opened 2016-12-16
Streamlines the usage of `console-ui`

twokul:
Introduces a new model to serve as a singleton for `console-ui`.

cc @rwjblue

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -16,6 +16,7 @@ const findAddonByName = require('../utilities/find-addon-by-name');
 const heimdall = require('heimdalljs');
 const PackageInfoCache = require('./package-info-cache');
 const experiments = require('../experiments');
+const UI = require('./ui');
 
 let processCwd = process.cwd();
 
@@ -657,13 +658,13 @@ class Project {
     @static
     @method closestSync
     @param  {String} pathName Path to your project
-    @param  {UI} _ui The UI instance to provide to the created Project.
+    @param  {UI} optionalUI The UI instance to provide to the created Project.
     @return {Project}         Project instance
    */
-  static closestSync(pathName, _ui, _cli) {
+  static closestSync(pathName, optionalUI, optionalCLI) {
     logger.info('looking for package.json starting at %s', pathName);
 
-    let ui = ensureUI(_ui);
+    let ui = optionalUI || new UI();
 
     let directory = findupPath(pathName);
     logger.info('found package.json at %s', directory);
@@ -671,7 +672,7 @@ class Project {
     let relative = path.relative(directory, pathName);
     if (relative.indexOf('tmp') === 0) {
       logger.info('ignoring parent project since we are in the tmp folder of the project');
-      return Project.nullProject(_ui, _cli);
+      return Project.nullProject(optionalUI, optionalCLI);
     }
 
     let pkg = fs.readJsonSync(path.join(directory, 'package.json'));
@@ -679,10 +680,10 @@ class Project {
 
     if (!isEmberCliProject(pkg)) {
       logger.info('ignoring parent project since it is not an ember-cli project');
-      return Project.nullProject(_ui, _cli);
+      return Project.nullProject(optionalUI, optionalCLI);
     }
 
-    return new Project(directory, pkg, ui, _cli);
+    return new Project(directory, pkg, ui, optionalCLI);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/models/project.js
```diff
@@ -625,14 +620,18 @@ Project.closestSync = function(pathName, _ui, _cli) {
   @static
   @method projectOrnullProject
   @param  {UI} _ui The UI instance to provide to the created Project.
+  @param  {CLI} _cli The CLI instance to provide to the created Project.
   @return {Project}         Project instance
  */
-Project.projectOrnullProject = function(_ui, _cli) {
+Project.projectOrnullProject = function(ui, cli) {
+  this.ui = ui;
```
rwjblue: This seems _slightly_ odd here, because `this` is actually `Project.prototype` here.

twokul: yeah I know

stefanpenner: `this` is actually `Project` itself here, not `Project.prototype`. We should not put `state` such as `ui` on the `Project` class itself.

twokul: agreed, I was working on this over the weekend, decided to go with a singleton + a separate module; only 2 tests are failing, need to fix before pushing

lib/models/project.js
```diff
@@ -575,20 +576,22 @@ class Project {
     @static
     @method closest
     @param  {String} pathName Path to your project
+    @param  {UI} optionalUI The UI instance to provide to the created Project.
+    @param  {Object} optionalCLI The CLI instnace to provide to the created Project.
```
btecu: `instnace` -> `instance`

lib/models/project.js
```diff
@@ -600,32 +603,33 @@ class Project {
     @static
     @method closestSync
     @param  {String} pathName Path to your project
-    @param  {UI} _ui The UI instance to provide to the created Project.
+    @param  {UI} optionalUI The UI instance to provide to the created Project.
+    @param  {Object} optionalCLI The CLI instnace to provide to the created Project.
```
btecu: `instnace` ...

lib/models/project.js
```diff
@@ -640,12 +644,12 @@ class Project {
     @param  {UI} _ui The UI instance to provide to the created Project.
```
nathanhammond: Should fix documentation here as well.

lib/models/ui.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor(args) {
+    let options = args || {};
+
+    if (!singletonUI) {
+      singletonUI = new UI({
+        inputStream: options.inputStream,
+        outputStream: options.outputStream,
+        errorStream: options.errorStream || process.stderr,
+        errorLog: options.errorLog || [],
+        ci: process.env.CI || (/^(dumb|emacs)$/).test(process.env.TERM),
```
nathanhammond: I know this is just moved, I'm unsure why `dumb`, though?

lib/models/ui.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The class that serves as a singleton wrapper for `console-ui`.
+ *
+ * @private
+ * @module ember-cli
+ * @class CommandLineUI
+ * @constructor
+ */
+class CommandLineUI {
```
nathanhammond: Oh how I love this new year.

lib/models/ui.js
twokul: I believe if terminal has no ANSI support then `TERM` will be set to `dumb` (meaning there will be no colours); it is sort of a "last resort, when everything is broken" kind of thing.

lib/models/ui.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor(args) {
+    let options = args || {};
+
+    if (!singletonUI) {
+      singletonUI = new UI({
+        inputStream: options.inputStream,
```
kellyselden: Do these need to be `inputStream: options.inputStream || process.stdin,` to support this:
https://github.com/ember-cli/ember-cli/pull/6577/files#diff-3677132bf6240059b3a3637716019c84L711

nathanhammond: I think so.

twokul: I don't think so. We instantiate UI only once in `lib/cli/index.js` and `process.stdin` is passed via `options` from [`bin/ember`](https://github.com/ember-cli/ember-cli/blob/master/bin/ember#L44); in tests, we use a fake UI from `console-ui`.

kellyselden: In this location, https://github.com/ember-cli/ember-cli/pull/6577/files#diff-3677132bf6240059b3a3637716019c84R613, it used to use `process` in the fallback, but now it uses null options.

twokul: we use `closestSync` only in one place - [`lib/models/project`](https://github.com/ember-cli/ember-cli/search?utf8=%E2%9C%93&q=closestSync) and we do not pass in `ui` or `cli` parameters when we call it.

the reason the fallback was put in place was b/c we created 2 instances of the `UI`: one when `bin/ember` runs and the other one when we require `lib/broccoli/ember-app`.

with this change, `UI` is created only once and we no longer require this fallback.

perhaps I'm missing something obvious?

nathanhammond: Nope, it seems like you know this code better than the both of us. 😛

Turbo87:
@twokul can you rebased this? we should try to push this over the finish line

twokul:
@Turbo87 sure thing. let me rebase

twokul:
@Turbo87 done

locks:
Closing due to inactivity.

locks:
Closing due to inactivity.

PR #6578, opened 2016-12-16
Use package version of calculate-cache-key-for-tree.

rwjblue:
As mentioned in the PR that landed the initial implementation, this utility function has been extracted into a package that can be shared/used by addon authors.

https://github.com/ember-cli/calculate-cache-key-for-tree


rwjblue:
@homu r+

Issue #6579, opened 2016-12-17
`BROCCOLI_VIZ=1 ember build` is broken


rwjblue:
Currently, when linking against master and running `BROCCOLI_VIZ=1 ember build` the following error is triggered:

```js
Build failed.
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at Class._computeVizInfo (/Users/rjackson/src/open-source/ember-cli/lib/models/builder.js:96:31)
    at /Users/rjackson/src/open-source/ember-cli/lib/models/builder.js:315:26
    at tryCatch (/Users/rjackson/src/open-source/ember-cli/node_modules/rsvp/dist/rsvp.js:538:12)
    at invokeCallback (/Users/rjackson/src/open-source/ember-cli/node_modules/rsvp/dist/rsvp.js:553:13)
    at publish (/Users/rjackson/src/open-source/ember-cli/node_modules/rsvp/dist/rsvp.js:521:7)
    at flush (/Users/rjackson/src/open-source/ember-cli/node_modules/rsvp/dist/rsvp.js:2373:5)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

This was broken when we landed https://github.com/ember-cli/ember-cli/pull/6546.  In #6546 we added code that assumes that `resultAnnotation` is passed to our `Builder.prototype.build` method ([lib/models/builder.js#L96](https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L96)).  This is true when our build is triggered by `ember-cli-broccoli-sane-watcher` due to changes upstream in https://github.com/ember-cli/broccoli-sane-watcher/pull/8. 

The fundamental issue is that when running `ember build` we do not go through `ember-cli-broccoli-sane-watcher` at all, and we manually invoke our `Builder.prototype.build` method in the build task ([lib/tasks/build.js#L22](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/build.js#L22)) but pass in no options.

We either need to make all builds funnel through ember-cli-broccoli-sane-watcher (which seems _somewhat_ odd when we don't want a watcher) or make our build task emit the correct result annotation.

rwjblue:
**Note:** This does not affect `BROCCOLI_VIZ=1 ember serve`, which still works properly (since the `.build` method is invoked by ember-cli-broccoli-sane-watcher).

PR #6580, opened 2016-12-17
Fix `BROCCOLI_VIZ=1 ember build`.

rwjblue:
Ensure an annotation is passed into `Builder.prototype.build` from the build task (even when not using the watcher).

Fixes https://github.com/ember-cli/ember-cli/issues/6579.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -16,8 +16,6 @@ var _resetTreeCache = require('./addon')._resetTreeCache;
 
 var Sync = require('tree-sync');
 
-var signalsTrapped = false;
-
 function _enableFSMonitorIfInstrumentationEnabled() {
   var monitor;
   if (_instrumentationEnabled()) {
@@ -132,6 +130,7 @@ module.exports = CoreObject.extend({
     this._buildCount = 0;
     this.setupBroccoliBuilder();
     this.trapSignals();
+    this._instantiationStack = (new Error).stack.replace(/[^\n]*\n/, '');
   },
 
   /**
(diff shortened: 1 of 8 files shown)


rwjblue:
This is mostly a quick fix (but it seems reasonable as a long term solution also).  I'd like to follow up on this to add a smoke test that ensures this type of thing works properly in the future, but we mostly dislike smoke tests (due to massive time cost) so I am hesitant to block this on that test...

Thoughts?

twokul:
@rwjblue I wonder if we could add a small unit test and use `td.verify()`?

rwjblue:
This scenario is actually the intersection of at least three different "major" actors in the system (build command, build task, builder model). If we just test one of the parts and mock the rest we would likely end up in a similar situation to what got us here. 

When I have time this weekend, I'll try to see if I can make an integration test for the task + builder model part.

rwjblue:
OK, just spent the afternoon working on this.  It turned into a small refactor of our signal handling (for the better IMO).

High points of the last round of changes:

* Added integration test that exercises the `build` task (without resorting to mocking methods all over the place).
* Move the counter used for `broccoli-viz.X.json` files to the builder instance (makes testing around the file generated much much easier since it is in a stable location).
* Remove module scope guard around setting up signal handlers, and instead refactor to allow always setting up the handlers as needed. Adding the integration test above, broke some existing tests that confirmed the signal handlers were trapped because the `tests/integration/tasks/build-test.js` is now being ran before `tests/unit/model/builder-test.js`. After the refactor, the handlers are setup and removed for each builder instance (they were already tied to the specific builder instance due to usage of `.bind(this)` when registering them).  This actually makes scenarios where there may be multiple builders within the same process cleanup properly.

Sadly, the updates above make this somewhat more involved, and I do not want to land without review (the original version of the PR was trivial). Added a request for review for a few folks, but would love review from whomever has time...

tests/unit/models/builder-test.js
```diff
@@ -23,18 +23,101 @@ var MockUI = require('console-ui/mock');
...
+      });
+
+      var actualListeners = getListenerCounts();
+
+      expect(actualListeners).to.eql({
+        SIGINT: originalListenerCounts.SIGINT + 1,
+        SIGTERM: originalListenerCounts.SIGTERM + 1,
+        message: originalListenerCounts.message + 1,
+
+        // enable when https://github.com/ember-cli/capture-exit/pull/17 lands
```
nathanhammond: Are you willing to block on this? I can see us forgetting about it...

rwjblue: I'll create an issue here to make sure we don't forget, but I'd like to land independently (not sure when David/Stef will have time for review, and broccoli-viz has already been broken on master for a few days...).

nathanhammond: K, seems good with tracking issue.

lib/models/builder.js
```diff
@@ -130,8 +127,10 @@ module.exports = CoreObject.extend({
 
   init: function() {
     this._super.apply(this, arguments);
+    this._buildCount = 0;
     this.setupBroccoliBuilder();
     this.trapSignals();
+    this._instantiationStack = (new Error).stack.replace(/[^\n]*\n/, '');
```
rwjblue: I just found this helpful when I was tracking down the signal handlers that weren't being cleaned up properly (in tests mostly), I can remove it if others don't find it valuable..

tests/unit/models/builder-test.js
rwjblue: FYI - Stef reviewed + released the `capture-exit` changes, so I've updated this PR to also confirm exit handler setup/teardown. We won't need another issue to track that now...

rwjblue:
Thanks @homu, I rebased for you. 😸

lib/models/builder.js
```diff
@@ -163,17 +162,29 @@ module.exports = CoreObject.extend({
    * @method trapSignals
    */
   trapSignals: function() {
-    if (!signalsTrapped) {
```
stefanpenner: i believe we may want to guard `trapSignals` from being called unbalanced from `_cleanupSignals`. Likely throwing if done so.

rwjblue: @stefanpenner - We already add an `captureExit.onExit` callback to call `.cleanup` (which calls `_cleanupSignals`) for each builder instance, so it basically cannot be out of balance (since the final cleanup will be done at process exit when the `onExit` callbacks are ran).  Therefore, I do not think it is possible (short of someone manually getting the running builder instance and calling `offExit` in their own code) to end out unbalanced...

stefanpenner: :shipit:

rwjblue:
@homu r+

PR #6582, opened 2016-12-18
Use `tap` reporter in CI.

rwjblue:
This makes finding failed tests in CI output much much easier:

* you can search for `not ok` and jump directly to the failure
* all stdout / stderr is grouped in the reporter output
* the count of pass / fail / skip is emitted at the end

I still find `spec` to be much nicer to look at while running local tests, so I left that as the default for non-CI tests.

Compare the following CI builds:

* [spec output](https://travis-ci.org/ember-cli/ember-cli/jobs/184977601)
* [tap output](https://travis-ci.org/ember-cli/ember-cli/jobs/184987946)


stefanpenner:
@homu r+

PR #6583, opened 2016-12-19
Add test confirming ordering of buildInstrumentation hook.

rwjblue:
The underlying implementation was already correct, but I noticed that this test didn't have a variant with visualization enabled that showed the expected ordering.

This just adds that test...


rwjblue:
@homu r+

PR #6584, opened 2016-12-19
Ensure tests pass when experiments are not present.

rwjblue:
Add CI run ensuring that test suite runs with the experiments disabled (this helps prevent regressions from being detected only after branching the next beta), and set the stage for removing replacing `lib/experiments/index.js` when branching to beta with:

```js
'use strict';

var experiments = {};
Object.freeze(experiments);
module.exports = experiments;
```


lib/experiments/index.js
```diff
@@ -1,7 +1,17 @@
 'use strict';
 var symbol = require('../utilities/symbol');
+var version = require('../../package').version;
+var isCanaryVersion = version.match(/-alpha/);
```
Turbo87: since when do we use alpha tags on master?

rwjblue: Roughly when this PR lands 😜

rwjblue: The basic idea is to ensure that the experiments present on master are only available when using "canary" builds.

#6576 enables us to use version numbers that haven't been published to NPM yet (which is the only real reason that master uses the latest beta version number anyways).

If we want to use some other mechanism to accomplish this, I think we could. This just seems like the easiest/best path forward to me.

Happy to add to next meeting agenda to discuss more in depth...

nathanhammond: I'm tempted to label it `2.12.0-canary` since we don't actually publish individual releases. I've not though through what that would mean if we wanted to actually do a release off of that branch. It's too late at night for that.

.travis.yml
```diff
@@ -21,6 +21,15 @@ cache:
...
     - $HOME/.cache # includes bower's cache
 
+env:
+  - NPM_SCRIPT=test-all:cover
+
+matrix:
+  fast_finish: true
+  include:
+  - node_js: "6"
+    env: SIMULATE_EMBER_CLI_BETA=true NPM_SCRIPT=test
```
nathanhammond: Can you add an _additional commit_ to see how fast `test-all` is?

.travis.yml
```diff
@@ -21,6 +21,15 @@ cache:
     - $HOME/.npm
     - $HOME/.cache # includes bower's cache
 
+env:
+  - NPM_SCRIPT=test-all:cover
+
+matrix:
+  fast_finish: true
```
nathanhammond: Mildly opposed to `fast_finish` but since we have no "allowed_failures" this just means that the check mark gets updated a touch faster.

stefanpenner: > this just means that the check mark gets updated a touch faster.

when does that happen?

rwjblue: @stefanpenner - When this is enabled, the commit status is updated to failing once the _first_ job fails. Without this option, the commit status stays "pending" until all tests have concluded and is then marked as failing if one of the jobs failed.

stefanpenner: @rwjblue ah sounds good

rwjblue:
I trolled myself in the most newbish way possible, but this is finally ready for review.

Turbo87:
@rwjblue I might be missing something but how does this PR no accomplish "Allow experiments only on canary builds"?

rwjblue:
@Turbo87 - You are correct, it does not do that any longer.  I updated the description and title here to explain better what it does now.

nathanhammond:
@homu r+

PR #6586, opened 2016-12-19
Remove unused build-watch task

ro0gr:
Can't find anything who references this guy.


rwjblue:
It is used when passing the `--watch` option to `ember build` ([here](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/build.js#L43-L49)).  

I'm going to close this (since we do use it), but we obviously need some tests for this (since removing it didn't fail anything)...

ro0gr:
Oops, sorry

PR #6588, opened 2016-12-19
Fix incomplete API docs for Blueprint.lookupBlueprint

simonihmig:
https://ember-cli.com/api/classes/Blueprint.html#method_lookupBlueprint describes the dasherizedName param as an `Object` and does not specify the return value.


stefanpenner:
👍

PR #6589, opened 2016-12-19
Refactor addon tree cache to expose usable API

trentmwillis:
When using the `ADDON_TREE_CACHING` experiment, this adds `Addon.prototype.treeCache` so that consumers (such as ember-engines) can easily leverage the same cache with instrumentation available.

See https://github.com/ember-engines/ember-engines/pull/283 for an example of usage.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -79,9 +79,38 @@ var DEFAULT_TREE_FOR_METHOD_METHODS = {
   vendor: ['treeForVendor']
 };
 
-var ADDON_TREE_CACHE = Object.create(null);
+var ADDON_TREE_CACHE = {
+  __cache: Object.create(null),
+
+  getItem: function(key) {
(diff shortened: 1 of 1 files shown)


lib/models/addon.js
```diff
@@ -1486,6 +1506,13 @@ if (addonTreeCacheExperimentPresent) {
 
     return cacheKey; // profit?
   };
+
+  Object.defineProperty(addonProto, 'treeCache', {
```
rwjblue: Lets move this to be another export instead of adding to prototype.

Something like:

```js
if (experiments.ADDON_TREE_CACHING) {
  module.exports._treeCache = ADDON_TREE_CACHE;
}
```

I definitely don't want addon authors thinking this is just a run of the mill public API...

rwjblue:
@homu r+

PR #6591, opened 2016-12-20
produce error code only if build is not finished

### LINKED ISSUE #6421: [Windows] Ctrl+C produces Exit status 1
When interrupt an ember server I get the following output:
```
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "sta
rt"
npm ERR! node v5.3.0
npm ERR! npm  v3.10.6
npm ERR! code ELIFECYCLE
npm ERR! ember-cli-windows-exit-error@0.0.0 start: `ember server`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the ember-cli-windows-exit-error@0.0.0 start script 'ember server'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the ember-cli-windows-exit-error package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     ember server
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs ember-cli-windows-exit-error
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls ember-cli-windows-exit-error
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     D:\ruslan\ember-cli-windows-exit-error\npm-debug.log

---
```

I've tested this in different combinations: **powershell**/**cmd**, **windows7**/**windows8**, **node@5**/**node@6** and always get the same output as above.

The issue started to happen in **ember-cli@2.7** and still persist in master. As I can see [trapWindowsSignals](https://github.com/ember-cli/ember-cli/commit/6b0e17e9c70585c115ce6c37ea1b9276e2060451) revealed this.

I believe the problem is not related to trapWindowsSignals implemetation. Seems like it's [`process.exit(1)`](https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L260).
When I run `process.exit()` without error code termination works as expected at least on my local machine. 

Does anybody know is there any reason to exit with an error code after cleaning up? Can we safely remove it?

ro0gr:
An attempt to fix #6421. 
Instead of always throwing an error code we do it only if `Builder.build` is in progress.

Needs:
- [ ] tests
- [ ] testing under windows
- [ ] go/no-go feedback

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -162,13 +162,12 @@ module.exports = CoreObject.extend({
    * @method trapSignals
    */
   trapSignals: function() {
-    this._boundOnSIGINT = this.onSIGINT.bind(this);
-    this._boundOnSIGTERM = this.onSIGTERM.bind(this);
+    this._boundExit = this._exit.bind(this);
     this._boundOnMessage = this.onMessage.bind(this);
     this._boundCleanup = this.cleanup.bind(this);
 
-    process.on('SIGINT',  this._boundOnSIGINT);
-    process.on('SIGTERM', this._boundOnSIGTERM);
+    process.on('SIGINT',  this._boundExit);
+    process.on('SIGTERM', this._boundExit);
     process.on('message', this._boundOnMessage);
     exit.onExit(this._boundCleanup);
 
@@ -178,8 +177,8 @@ module.exports = CoreObject.extend({
   },
 
   _cleanupSignals: function() {
-    process.removeListener('SIGINT',  this._boundOnSIGINT);
-    process.removeListener('SIGTERM', this._boundOnSIGTERM);
+    process.removeListener('SIGINT',  this._boundExit);
+    process.removeListener('SIGTERM', this._boundExit);
     process.removeListener('message', this._boundOnMessage);
     exit.offExit(this._boundCleanup);
 
@@ -320,7 +319,7 @@ module.exports = CoreObject.extend({
     var self = this;
     attemptNeverIndex('tmp');
 
-    return this.processAddonBuildSteps('preBuild')
+    return this._build = this.processAddonBuildSteps('preBuild')
       .then(function() {
         return self.builder.build(willReadStringDir);
       })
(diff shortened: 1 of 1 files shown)


lib/models/builder.js
```diff
@@ -320,7 +319,7 @@ module.exports = CoreObject.extend({
     var self = this;
     attemptNeverIndex('tmp');
 
-    return this.processAddonBuildSteps('preBuild')
+    return this._build = this.processAddonBuildSteps('preBuild')
```
ro0gr: probably should be `this._isBuilding = true`. No need in a promise

lib/models/builder.js
```diff
@@ -400,30 +402,6 @@ module.exports = CoreObject.extend({
   },
 
   /**
-   * Handles the `SIGINT` signal.
-   *
-   * Calls {{#crossLink "Builder/cleanupAndExit:method"}}{{/crossLink}} by default.
```
ro0gr: This `crossLink` looks like outdated. Couldn't find where does it point to.

lib/models/builder.js
```diff
@@ -400,30 +402,6 @@ module.exports = CoreObject.extend({
...
-   * @method onSIGINT
-   */
-  onSIGINT: function() {
-    process.exit(1);
-  },
-
-  /**
-   * Handles the `SIGTERM` signal.
-   *
-   * Calls {{#crossLink "Builder/cleanupAndExit:method"}}{{/crossLink}} by default.
```
ro0gr: Looks like also outdated

rwjblue:
I like the refactoring (I should have noticed the SIGINT and SIGTERM were the same 😢 when I just refactored this), but I'm not sure this is quite the right behavior either.

You see, we don't attach any other signal handlers outside of the builder, so I'm not certain that its appropriate to consider the build state the only thing that is important exit code wise.  I'm thinking of things like `ember test` (where we do a build, then run tests, if a SIGINT happens mid-test suite we shouldn't `exit(0)`, right?)...

ro0gr:
You are right. Again :boom: 
Seems like builder doesn't(and probably should not) have enough knowledge about how to properly exit. 
I keep on iterate on this as time permits.

ro0gr:
Closing in favour of #6615

PR #6592, opened 2016-12-20
Use release branch when auto deploying API docs

### LINKED ISSUE #6585: Deploy ember-cli.com/api from release branch only.
We need to tweak the auto deploy script for `ember-cli.com/api` generation to only run for the `release` branch.  Currently, landing anything on master shows up there immediately even though it may not be usable until it releases as stable.

I _believe_ that we would only need to change [this line](https://github.com/ember-cli/ember-cli/blob/master/.travis/deploy.sh#L10), but @Turbo87 can likely confirm...

mike183:
Closes #6585

### DIFF
--- a/.travis/deploy.sh
+++ b/.travis/deploy.sh
@@ -7,7 +7,7 @@ if [[ $TRAVIS_PULL_REQUEST != "false" ]]; then
   exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


rwjblue:
@homu r+

rwjblue:
Thank you!

PR #6603, opened 2016-12-22
Prevent duplicate imports.

### LINKED ISSUE #6601: app.import will happily include the same file multiple times
If `app.import` is called on the same path multiple times, the file at that path will be concatenated into the final output once per call. I chatted briefly with @rwjblue about this in Slack and neither of us could come up with a real-world scenario where that would be desirable.

The most likely way for this to occur today is when two instances of an addon import the same library. Fun fact: if you have two different versions of that addon installed, you end up with two copies of _one version_ of the library (which clobbered the other when their `vendor` trees merged), which definitely isn't making anyone happy 😉 

I should have time in the next day or two to take a stab at fixing this, but wanted to open this to track in the meantime.

nathanhammond:
Closes #6601. /cc @rwjblue @dfreeman

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -122,6 +122,10 @@ function EmberApp(defaults, options) {
   this._initOptions(options);
   this._initVendorFiles();
 
+  // Not changing object shapes because people use these.
+  this._scriptOutputFilesBookkeeping = {};
+  this._legacyTestFilesToAppendBookkeeping = [];
+
   this._styleOutputFiles       = {};
   this._scriptOutputFiles      = {};
   this.amdModuleNames          = null;
@@ -1584,9 +1588,9 @@ EmberApp.prototype._import = function(assetPath, options, directory, subdirector
 
     if (options.type === 'vendor') {
       options.outputFile = options.outputFile || this.options.outputPaths.vendor.js;
-      addOutputFile(this._scriptOutputFiles, assetPath, options);
+      addOutputFile('firstOneWins', this._scriptOutputFiles, assetPath, options);
     } else if (options.type === 'test') {
-      if (!allowImport(this.legacyTestFilesToAppend, assetPath, options)) { return; }
+      if (!allowImport('firstOneWins', this.legacyTestFilesToAppend, assetPath, options)) { return; }
       if (options.prepend) {
         this.legacyTestFilesToAppend.unshift(assetPath);
       } else {
@@ -1598,9 +1602,9 @@ EmberApp.prototype._import = function(assetPath, options, directory, subdirector
   } else if (extension === '.css') {
     if (options.type === 'vendor') {
       options.outputFile = options.outputFile || this.options.outputPaths.vendor.css;
-      addOutputFile(this._styleOutputFiles, assetPath, options);
+      addOutputFile('lastOneWins', this._styleOutputFiles, assetPath, options);
     } else {
-      if (!allowImport(this.vendorTestStaticStyles, assetPath, options)) { return; }
+      if (!allowImport('lastOneWins', this.vendorTestStaticStyles, assetPath, options)) { return; }
       if (options.prepend) {
         this.vendorTestStaticStyles.unshift(assetPath);
       } else {
@@ -1841,7 +1845,7 @@ function calculateModulePrefix(config) {
   return config.modulePrefix;
 }
 
-function addOutputFile(container, assetPath, options) {
+function addOutputFile(strategy, container, assetPath, options) {
   var outputFile = options.outputFile;
 
   if (!outputFile) {
@@ -1851,8 +1855,7 @@ function addOutputFile(container, assetPath, options) {
   if (!container[outputFile]) {
     container[outputFile] = [];
   }
-
-  if (!allowImport(container[outputFile], assetPath, options)) { return; }
+  if (!allowImport(strategy, container[outputFile], assetPath, options)) { return; }
 
   if (options.prepend) {
     container[outputFile].unshift(assetPath);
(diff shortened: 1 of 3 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1845,6 +1845,30 @@ function addOutputFile(container, assetPath, options) {
...
+      'This can result in very strange behavior in your application as the',
+      'asset historically was included twice. Multiple versions of the same asset',
+      'are also clobbered by the last version added to the Broccoli tree, meaning',
+      'that it only ever included the version of the asset which was last added',
+      'to the Broccoli tree. This should be treated as a bug to address. This',
+      'typically occurs when an asset is imported by a nested addon. Ember CLI',
+      'no longer includes multiple copies by default and instead is using only',
+      'the _last_ occurrence imported.'
+    ];
+    this.project.ui.writeWarnLine(message.join(' '));
```
rwjblue: I don't think I'm a fan of warning here. We should basically just do the right thing. 

Warning here will be super noisy for folks, and due to our config + nested addon story there is very little they can actually do about it (other than possibly storing some flag on the project and checking for it, which would only work for newer versions of that addon).

lib/broccoli/ember-app.js
```diff
@@ -1845,6 +1845,30 @@ function addOutputFile(container, assetPath, options) {
     container[outputFile] = [];
   }
 
+  if (~container[outputFile].indexOf(assetPath)) {
```
rwjblue: I'd prefer a !== -1 here instead (until we can use includes). The bitwise operations always give me trouble when reading.

lib/broccoli/ember-app.js
stefanpenner: Ya, its unclear that this warning can even reasonably be actionable.

Likely:

* we should choose a dedupe strategy (last wins or first wins)
* `logger.info` that this happened

lib/broccoli/ember-app.js
hjdivad: It's hard to get used to because it's not a general trick; it pretty much only works with `indexOf`

nathanhammond: @rwjblue always seems to comment on my doing this; it's a habit I picked up and just idiomatic in my head. Rather than have this be a thing I've added us a nice little linting rule to enforce it.

lib/broccoli/ember-app.js
nathanhammond: > We should basically just do the right thing.

The "right thing" here is nebulous. If we're completely silent then the user goes about their merry way and is woefully unaware that we've enforced the Highlander Rule on them with `broccoli-merge-trees` and `{ overwrite: true }` (well before `app.import` gets run). Then, to prevent further trolling, we make it so that we at least don't include the asset multiple times in the bundled output. Basically we let you shoot yourself in the foot then we hand you a bandaid.

I'm modifying to `logger.info` but at some point we need to start making plans to move the ecosystem.
 
> we should choose a dedupe strategy 

Dedupe strategy is already defined and tested: last import wins (by array order, not invocation order because `prepend`).

stefanpenner: > The "right thing" here is nebulous.

Yup, I intended to leave "right thing" vague, so we could pencil it in as materializes clearly.

> Dedupe strategy is already defined and tested: last import wins (by array order, not invocation order because prepend).

Sounds good

lib/broccoli/ember-app.js
stefanpenner: My issue with it isn't just that it is a bitwise operation, it is that JavaScript's coercion is allowed to once again run wild.

In general, we encourage `===` over `==` and my feeling is that discouraging `~` is an extension of that.

My brain parses:
```js
if (~x.indexOf(y)) {

}
```

as:

```js
if (toBoolean(-(toNumber(x.indexOf(y)) + 1)) {

}
```
whereas

```

Comment (stefanpenner) [inline lib/broccoli/ember-app.js]: eslint rule sounds good.

Comment (stefanpenner) [inline lib/broccoli/ember-app.js]: that being said, `~` + coercion enables my favorite part of JavaScript

```js
-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-~-[]
> 42
```

nathanhammond: That's a nice array you have there.

nathanhammond:
I've modified this to lock out in a reasonable manner for AMD imports. If you attempt to import the same anonymous module with different module names it will now throw an error. If you attempt to import it with the same AMD module name it will dedupe correctly.

Also, while investigating, I discovered that [we didn't actually implement the API which was proposed in the Anonymous AMD RFC](https://github.com/ember-cli/rfcs/pull/55#pullrequestreview-14217214). Opened a tracking issue #6607.

I consider this fix complete given the limitations of our current API.

lib/broccoli/ember-app.js
```diff
@@ -1845,9 +1852,28 @@ function addOutputFile(container, assetPath, options) {
...
   } else {
     container[outputFile].push(assetPath);
   }
 }
+
+function allowImport(fileList, assetPath, options) {
+  if (fileList.indexOf(assetPath) !== -1) {
+    logger.info('Highlander Rule: duplicate `app.import(' + assetPath + ')`. Only including the last by order.');
+
+    if (options.prepend) {
```
rwjblue: This seems incorrect to me? The new import should be prepended and replace the existing one (because ordering matters 😜).

nathanhammond: Order does matter. But the last one always wins. Given `[1,2,3]` and adding `{ 3, prepend: true }` the _existing_ 3 is the one which would be last in the array, and thus the one which would apply in the end.

nathanhammond: We were talking past each other. New strategy detailed below.

nathanhammond:
# Recommendation

Which asset gets removed from the array is a confusing problem. Rob and I talked about it for a bit. My recommendation is first one wins in JS, and last one wins in CSS.

## CSS
For CSS we know for a fact that the current behavior in user apps is that the last copy is winning. We can therefore always select the last duplicate `assetPath` in the provided array.

## JS

In JS the asset which gets included can be one of three different things:

1. An AMD module definition.
2. Globals code.
3. Globals code which reifies an AMD module definition.

Given these three we should opt for a "first one wins" for JavaScript. No matter what there will be failure modes, but this minimizes the failure modes.

### Scenario 1 Failure Mode
None.

### Scenario 2 Modes
In a first one wins we end up hoisting globals code to their first occurrence. If the code does bootstrapping it may presently rely on being run at the later place as well, thus clobbering the global state with a "properly" set up global state which had access to later assets.

This also _fixes_ situations where globals code whose initialization results in duplicated event handlers or performs non-idempotent mutation of initial state (`globalArray.push('thing')`). Any globals code which is required by modules that get reified will also guaranteedly be present, which is also better.

### Scenario 3 Modes
Strange state could be caught in closures, but this seems exceedingly unlikely in an Ember application.

## Alternative
We could also do "first one wins" inside of the `prepend` subset, and "last one wins" inside of the non-`prepend` subset. And if the asset appears in both subsets it is included twice. But even deduplication can change what is present at different points. 

## Why We Can't Be Perfect

```javascript
app.import('a.js');
app.import('b.js');
app.import('a.js');
app.import('a.js');
app.import('b.js');

// first one wins: a, b
// last one wins: a, b
// But it's possible that "a" relied on being after "b".
```

rwjblue:
@homu r+

PR #6605, opened 2016-12-22
Lint for bitwise operations.

nathanhammond:
Considering we're enforcing this during code review let's make it official.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -115,6 +115,8 @@ module.exports = {
     // JSHint "quotmark"
     'quotes': [2, 'single'],
 
-    'semi': [2, 'always']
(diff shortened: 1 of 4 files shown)


nathanhammond:
@homu r+

.eslintrc.js
```diff
@@ -115,6 +115,8 @@ module.exports = {
     // JSHint "quotmark"
     'quotes': [2, 'single'],
 
-    'semi': [2, 'always']
+    'semi': [2, 'always'],
+
+    'no-bitwise': [2]
```
Turbo87: FWIW `'no-bitwise': 2` is enough here. the array notation is only needed when you want to set options on the rule.

PR #6606, opened 2016-12-22
Instrumentation

hjdivad:
experiment implementation of extended `instrumentation` hook.

see https://github.com/ember-cli/rfcs/pull/91 for some context although note that rfc needs to be updated.  This pr is more up to date currently (ie unifiy in a single instrumentation hook).

The idea is to use `instrumentation(name, graph)` instead of `buildInstrumentation` and to add instrumentation hooks so that there are four calls:

- `init`
- `command`
- `build` (only for commands that build)
- `shutdown`

We fire init as soon as we can (in `bin/ember` if possible, although we recover in cases where that didn't happen eg due to global version mismatch).  Init ends when command begins. command encompasses an entire successful command run, so will overlap with build.  The exception is for serve because we clean up build instrumentation to avoid leaking memory, so command will include only the last build.  shutdown covers everything from the command promise fulfillment to process exit.

Note that currently command and shutdown don't fire on SIGINT, which is likely desired behaviour for now but I am happy to paint this bikeshed with people who have opinions. 

### TODO

- [x] rename tree -> graph?
- [x] mark experiments private
- [x] cookie -> token
- [x] project.instrumentation ->   @@project._instrumentation
- [x] reenable retries 2
- [x] rename jsons to `instrumentation.${phase}.json` (`${phase}.${num}` for build)
- [x] consensus re module.exports location
- [x] update rfc
- [x] rebase

### Open Q / TODO

- [X] ~~We should try to ensure Builder.prototype.cleanup is included in shutdown of the build instrumentation~~ Added heimdalljs node specifically for tracking builder cleanup, and confirmed that it is included in `command` phase (but not `build` phase).  
- [X] ~~Solidify that all addon hooks are or are not included in build instrumentation. (I think they should be, as they contribute to "perceived" build time.)~~
  * The various addon hooks that are used to generate the build pipeline are included in `command` instrumentation (and not `build`).
- [x]  Add test in tests/unit/cli/cli-test.js that ensures that instrumentation is started if initInstrumentation is not passed in (might be tested already?)

### Followup Tasks

* change `BUILD_INSTRUMENTATION` from fallback to hard error
* test to insure against leak of the heimdall builder nodes when running `ember serve` 
* fix the way `SIGINT` works to ensure that `command` and `shutdown` instrumentation hooks are invoked for `ember serve` / `ember test --server`.
* clean up command instrumentation timing
* clean up build instrumentation  timing


tests/unit/models/addon-test.js
```diff
@@ -759,7 +772,8 @@ describe('models/addon.js', function() {
       var packageContents = require(path.join(projectPath, 'package.json'));
 
       ui = new MockUI();
-      project = new Project(projectPath, packageContents, ui);
+      var cli = new MockCLI({ ui: ui });
```
hjdivad: it does

hjdivad: https://github.com/ember-cli/ember-cli/blob/04cbe974a19beb61af25415901cd9ab78d08dbb4/tests/helpers/mock-cli.js#L9

stefanpenner: kk

lib/models/instrumentation.js
```diff
@@ -0,0 +1,277 @@
...
+function Instrumentation (options) {
+  this.ui = options.ui;
+
+  // project constructor will set up bidirectional link
+  this.project = null;
+
+  this.instrumentations = {
+    init: options.initInstrumentation,
+    build: {
+      cookie: null,
```
rwjblue: Should we call this something else? I feel like the cookie verbiage is removed from heimdall 0.3, and I'd prefer to avoid it here...

Maybe "token"?  What is it called in the 0.3 API?

lib/models/instrumentation.js
```diff
@@ -0,0 +1,277 @@
...
+
+  var node;
+  var statName;
+  var statValue;
+  var nodeItr;
+  var statsItr;
+  var nextNode;
+  var nextStat;
+
+  for (nodeItr = tree.preOrderIterator();;) {
```
rwjblue: I'd like to see this split out, it is a tad bit annoying to have to include the iterator twice, seems good for a future refactor (or just wait until we can use ES6?)...

lib/models/instrumentation.js
```diff
@@ -0,0 +1,277 @@
...
+        }
+      }
+    });
+  }
+};
+
+Instrumentation.prototype._writeInstrumentation = function (name, instrumentationInfo) {
+  if (!vizEnabled()) { return; }
+
+  var filename = 'broccoli-viz.' + name;
```
rwjblue: I think we should stop using `broccoli-viz` as part of the filename here. Only `build` instrumentation has anything to do with `broccoli-viz` and even for that one `broccoli-viz` isn't what drives the format (its heimdall output).

lib/models/instrumentation.js
```diff
@@ -0,0 +1,277 @@
...
+Instrumentation.prototype._instrumentationFor = function (name) {
+  var instr = this.instrumentations[name];
+  if (!instr) {
+    throw new Error('No such instrumentation "' + name + '"');
+  }
+  return instr;
+};
+
+Instrumentation.prototype._instrumentationTreeFor = function(name) {
+  return tree.loadFromNode(this.instrumentations[name].node);
```
rwjblue: We should throw if `!!this.instrumentations[name]`.  Specifically to prevent folks from doing `this.project.instrumentation.start('myaddonThingy')`.

lib/models/project.js
```diff
@@ -50,6 +50,9 @@ function Project(root, pkg, ui, cli) {
     version: null,
     canNestRoots: false
   };
+
+  this.instrumentation = cli.instrumentation;
```
rwjblue: Should be underscored

tests/runner.js
```diff
@@ -23,7 +23,7 @@ var reporter = process.env.MOCHA_REPORTER || (process.env.CI ? 'tap' : 'spec');
 var mocha = new Mocha({
   timeout: 5000,
   reporter: reporter,
-  retries: 2
+  // retries: 2
```
rwjblue: This burns my bippie too, but we should tweak in a later PR...

tests/unit/cli/cli-test.js
```diff
@@ -5,6 +5,8 @@ var MockUI        = require('console-ui/mock');
 var MockAnalytics = require('../../helpers/mock-analytics');
 var CLI           = require('../../../lib/cli/cli');
 var td = require('testdouble');
+var heimdall = require('heimdalljs');
+var chalk = require('chalk');
```
rwjblue: Why is `chalk` added here?  I didn't spot it used?

tests/unit/models/builder-test.js
```diff
@@ -800,22 +382,61 @@ describe('models/builder.js', function() {
...
+        it('invokes the instrumentation hook if it is preset', function() {
+          addon[experiments.INSTRUMENTATION] = function (instrumentation) {
+            hooksCalled.push('instrumentation');
+            instrumentationArg = instrumentation;
+          };
 
-        return builder.build(null, {}).then(function() {
-          expect(hooksCalled).to.deep.equal(['preBuild', 'build', 'postBuild', 'buildInstrumentation', 'outputReady']);
+          return builder.build(null, {}).then(function() {
+            expect(hooksCalled).to.deep.equal(['preBuild', 'build', 'postBuild', 'instrumentation', 'outputReady']);
```
rwjblue: I know I added this test originally, but I think we should probably change it so that `instrumentation` fires last (otherwise our build instrumentation doesn't include `outputReady` invocation times).

tests/unit/models/builder-test.js
```diff
@@ -800,22 +382,61 @@ describe('models/builder.js', function() {
...
+            instrumentationArg = instrumentation;
+          };
+
+          return builder.build(null, {}).then(function() {
+            expect(hooksCalled).to.deep.equal(['preBuild', 'build', 'postBuild', 'instrumentation', 'outputReady']);
+          });
+        });
+      }
+
+      if (experiments.BUILD_INSTRUMENTATION) {
```
rwjblue: I'd like to remove this (now or in a later PR) before enabling the experiment as public API...

tests/unit/models/instrumentation-test.js
```diff
@@ -0,0 +1,872 @@
...
+    var originalStatSync = fs.statSync;
+
+    beforeEach(function () {
+      expect(!!process.env.BROCCOLI_VIZ).to.eql(false);
+      expect(!!process.env.EMBER_CLI_INSTRUMENTATION).to.eql(false);
+    });
+
+    afterEach(function() {
+      td.reset();
+      delete process.env.BROCCOLI_VIZ;
```
rwjblue: This is done by the `afterEach` setup on L32, why do it again?

lib/models/instrumentation.js
```diff
@@ -0,0 +1,277 @@
...
+    this.ui.writeLine(chalk.yellow(
+      'No init instrumentation passed to CLI.  Please update your global ember or ' +
+      'invoke ember via the local executable within node_modules.  Init ' +
+      'instrumentation will still be recorded, but some bootstraping will be ' +
+      'omitted.'
+    ));
+  }
+}
+
+module.exports = Instrumentation;
```
rwjblue: I generally prefer to export after definition and find this pattern hard to read (yes, I know functions are hoisted):

```js
module.exports = Foo;
function Foo() { }
```

I do not feel super strongly (I have had to get used to the above style anyways...).

tests/unit/models/project-test.js
```diff
@@ -29,7 +52,9 @@ describe('models/project.js', function() {
             baseURL: '/foo/bar'
           });
 
-          project = new Project(projectPath, { }, new MockUI());
+          project = new Project(projectPath, { }, new MockUI(), cli);
```
rwjblue: Seems slightly odd to swap the pattern around from earlier tests, I was expecting to see:

```js
let cli = new MockCLI();
project = new Project(projectPath, { }, cli.ui, cli);
```

lib/models/instrumentation.js
hjdivad: nope, sounds good

tests/helpers/mock-cli.js
```diff
@@ -0,0 +1,13 @@
...
+var Instrumentation = require('../../lib/models/instrumentation');
+
+function MockCLI(options) {
+  options = options || {};
+
+  this.ui = options.ui || new MockUI();
+  this.instrumentation = options.instrumentation || new Instrumentation({});
+}
+
+module.exports = MockCLI;
```
hjdivad: will do

lib/models/instrumentation.js
hjdivad: We agree that `export default class Ohai {}` is the way to go post 4 jan right?  So whatever bikeshedding we do on this is for the short term right?

FWIW I don't have much of a dog in this hunt either way.  I find both unreadable and just want es6 classes.

lib/experiments/index.js
```diff
@@ -3,6 +3,7 @@ var symbol = require('../utilities/symbol');
 
 var experiments = {
   BUILD_INSTRUMENTATION: symbol('build-instrumentation'),
```
hjdivad: I think the plan for now is to check for `BUILD_INSTRUMENTATION` for a bit and hard error, telling the user to move to `INSTRUMENTATION`.  Later yeah definitely remove it from experiments entirely.

hjdivad:
@rwjblue 

>  We should try to ensure Builder.prototype.cleanup is included in shutdown of the build instrumentation

Why is it problematic for this to be included in `shutdown` instead of `command`?

>  Solidify that all addon hooks are or are not included in build instrumentation. (I think they should be, as they contribute to "perceived" build time.)

I'm trying to work out what the action item is here.  Addon hooks that are called during the build will be in the build instrumentation.  What's the specific thing that requires more solidificationizing?

>  Explicitly document lib/models/instrumentation.js as private API (unless we are planning to update the RFC and actually expose this to user-land).

I have updated the rfc for exposing this to userland but it seems reasonable to land the PR as experiment first anyway and promote to public after everyone has had a chance to raise any objections to the rfc.

But anyway 👍  on explicitly documenting it as private.  I presume you mean with yuidoc comments right?

> Add test in tests/unit/cli/cli-test.js that ensures that instrumentation is started if initInstrumentation is not passed in (might be tested already?)

This is unit tested in [models/instrumentation-test.js](https://github.com/ember-cli/ember-cli/blob/04cbe974a19beb61af25415901cd9ab78d08dbb4/tests/unit/models/instrumentation-test.js#L109-L124).  Do you feel it also needs a test in `cli-test.js`?

> Add test to ensure nodes are not leaked with multiple instrumentation.start('build') calls (we remove instr.node for every time we call .start, but I want to make sure that is under test)

👍

lib/cli/cli.js
```diff
@@ -125,9 +136,23 @@ CLI.prototype.run = function(environment) {
...
     command.beforeRun(commandArgs);
 
+    var instrumentation = this.instrumentation;
+
     return Promise.resolve().then(function() {
       loggerTesting.info('cli: command.validateAndRun');
+
+
+      instrumentation.stopAndReport('init');
+      instrumentation.start('command');
```
hjdivad: 👍 good call

lib/models/builder.js
```diff
@@ -300,25 +200,14 @@ module.exports = CoreObject.extend({
...
-    }
-  },
-
   /**
    * @private
    * @method build
    * @return {Promise}
    */
   build: function(willReadStringDir, resultAnnotation) {
+    this.project.instrumentation.start('build');
```
hjdivad: This sounds like a great improvement to me 👍

lib/models/instrumentation.js
hjdivad: yes token is much better

lib/models/instrumentation.js
hjdivad: My inclination is to leave this as an es6 refactor but I am happy to be persuaded otherwise.

lib/models/instrumentation.js
rwjblue: Yes, totally, we want ES6 classes here for sure.

lib/experiments/index.js
rwjblue: Seems good to me!

lib/models/instrumentation.js
rwjblue: Ahh, ok, I'm game for this.

rwjblue:
> Why is it problematic for this to be included in shutdown instead of command?

Sorry, I'm specifically saying that it _should_ be included in `shutdown`.  I'm unsure at the moment if it is included anywhere...

> I'm trying to work out what the action item is here. Addon hooks that are called during the build will be in the build instrumentation. What's the specific thing that requires more solidificationizing?

Action item is: making sure `outputReady` is ran before `addon.instrumentation('build', ...)` is called

> But anyway 👍 on explicitly documenting it as private. I presume you mean with yuidoc comments right?

Yes, exactly.  Also, I hadn't re-reviewed the RFC yet.  Will chime back in here after reading if things are still unclear on this.

lib/models/instrumentation.js
hjdivad: ` this.project.instrumentation.start('myaddonThingy').` already throws via `instrumentationFor`, which is tested: https://github.com/ember-cli/ember-cli/blob/04cbe974a19beb61af25415901cd9ab78d08dbb4/tests/unit/models/instrumentation-test.js#L325-L330

lib/models/instrumentation.js
hjdivad: @rwjblue i'm confused about your comment re `build`: i think `broccoli-viz` is generic enough to handle any of them, although it's certainly built for `build`.

That said I have zero opinion about the names of these files besides their indicating the phase and build number (for build), so am happy to bikeshed with you and @stefanpenner.

- `instrumentation.build.0.json`?

What sounds good to you?

lib/models/project.js
hjdivad: 👍

tests/runner.js
hjdivad: ah yeah accidentally left this in; will remove

tests/unit/cli/cli-test.js
hjdivad: holdover from a test that got moved to instrumentation-test.

i'll remove it here

tests/unit/models/builder-test.js
hjdivad: 👍

tests/unit/models/instrumentation-test.js
hjdivad: cruft.  it dies

tests/unit/models/project-test.js
hjdivad: this is just an error.  i will fix it to match your expectations

lib/models/instrumentation.js
rwjblue: :+1:

hjdivad:
> Sorry, I'm specifically saying that it should be included in shutdown. I'm unsure at the moment if it is included anywhere...

Ah I typo'd which has led to some confusion.  I was trying to ask why it's problematic for it to be included in `command` instead of `shutdown`.  This particular cleanup is part of the build command, no?

lib/models/instrumentation.js
rwjblue: > I think broccoli-viz is generic enough to handle any of them, although it's certainly built for build.

Broccoli is very specifically a build tool, and it is **only** related to the `build` instrumentation. Using the term "broccoli" in relation to ember-cli init/ commands / shutdown. 

> instrumentation.build.0.json?

Yes, I like this.

rwjblue:
>  I was trying to ask why it's problematic for it to be included in command instead of shutdown. This particular cleanup is part of the build command, no?

Yes, it is definitely part of the build command, but for `serve` / `test --server` it is part of `shutdown` (since it is done during process.exit).  I realize that the two server related commands will have different semantics (this is ok), but was just trying to: 1) ensure that it is included in _something_, and 2) nail down which instrumentation hook it would be included in for each of `ember b`, `ember s`, `ember test`, and `ember test --server`.  I think we are good with how things are here though, as long as we can generally guarantee that it is always included in **one** (and only one) of the hooks.

lib/models/instrumentation.js
hjdivad: @stefanpenner reminded me that we can't actually `export default class UWot {}` in node due to shenanigans, so the question is not quite as moot as i'd hoped, although I think the tl;dr is keep the exports near the bottom.

Or add module transpilation for ember-cli...

rwjblue:
> > Explicitly document lib/models/instrumentation.js as private API (unless we are planning to update the RFC and actually expose this to user-land).
> 
> I have updated the rfc for exposing this to userland but it seems reasonable to land the PR as experiment first anyway and promote to public after everyone has had a chance to raise any objections to the rfc.

Hmm, I just reread the current state of the RFC ([this version](https://github.com/ember-cli/rfcs/blob/dfd4f4d8cbd568cbf79a0e2438bab30d7bf077ca/active/0000-addon-instrumentation-experimental-hooks.md)) and I cannot see where it is talking about making the thing defined in `lib/models/intrumentation.js` public API. AFAICT, it is explicitly talking about the `addon.instrumentation` hook becoming public, nothing else (e.g. `this.project.instrumentation` does not become public api in that RFC)...

hjdivad:
> Hmm, I just reread the current state of the RFC (this version) and I cannot see where it is talking about making the thing defined in lib/models/intrumentation.js public API. AFAICT, it is explicitly talking about the addon.instrumentation hook becoming public, nothing else (e.g. this.project.instrumentation does not become public api in that RFC)...

Sorry I misread you again; thought you were talking about the experiment.  Yes absolutely `lib/models/instrumentation` is private

rwjblue:
Just pushed some commits to address some of the TODO items:

* rebase
* reset `retries`
* make `project.instrumentation` into `project._instrumentation` (makes it clearly private)
* replace usages of `cookie` with `token` (WRT heimdall usage)
* fix tests and update to heimdalljs-graph@0.3.0

rwjblue:
Just did another pass through the comments/requests here and tackled the following:

* Add `heimdall` node to track builder cleanup, and confirmed it is properly included in `command` instrumentation for `ember build`
* Adjusted various locations to ensure that `module.exports = SomeThing` happens above the constructor.
* Move build instrumentation hook to happen after `outputReady` hook.
* Refactor project unit tests to use `MockCLI` consistently.

rwjblue:
This is finally ready to land.  There are a few things left in the description that I have labeled as followup, and I created https://github.com/ember-cli/ember-cli/issues/6623 to track those tasks.

I'll follow-up with @stefanpenner and @hjdivad early this coming week, but (thanks to the new experiments features) we can iterate on master so I'm going to land this now to avoid continued rebase hell...

@homu r+

Turbo87:
@rwjblue @stefanpenner @hjdivad FWIW I'm not happy about merging this PR. the amount of unstructured WIP commits in this PR makes it *very* hard to figure out what's going on. I hope that in the future we can return to meaningful commits and commit messages.

rwjblue:
@Turbo87 - Yes, after looking back through the commits I totally agree with you. When I approved it I was looking through the diff but not the commits.  I should have totally squashed many of them together. I'm sorry, I'll try to do a better job of this going forward.

hjdivad:
@Turbo87 @rwjblue I completely agree.  Commits were WIP to make it easier to review incremental changes but were never intended to be merged as such.

Sorry about being offline the last week @rwjblue but thanks for pushing this through.

Turbo87:
cool, glad we're on the same page :)

rwjblue:
Yeah, I just mucked it up, I knew we needed to redo the commits but forgot 😢

PR #6608, opened 2016-12-22
Update RELEASE.md

nathanhammond:
I need to remember to check the experiments stuff during the next release. This addition is so that future me doesn't screw up a release and have to do it twice.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -95,7 +95,7 @@ git checkout -B master --track origin/master
 # Merge the new beta release into the "master" branch
 git merge vX.Y.0-beta.1
 
-# Confirm that experiments are turned on.
+# Confirm that experiments are turned on before pushing it back to master.
 # Push back upstream.
 git push origin
 ```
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -94,6 +94,10 @@ git checkout -B master --track origin/master
 
 # Merge the new beta release into the "master" branch
 git merge vX.Y.0-beta.1
+
+# Confirm that experiments are turned on.
```
kellyselden: How do you confirm? For future non-nathans

hjdivad: @kellyselden `Object.keys(require('ember-cli/lib/experiments')).length` or similar

kellyselden:
It reads to me the first warning says in beta experiments are on, then the second warning says in beta experiments are off.

hjdivad:
> It reads to me the first warning says in beta experiments are on, then the second warning says in beta experiments are off.

It reads the same way to me.  "Promoting canary to beta" should ensure experiments are off.  Releasing canary is the only time we should expect experiments to be on.

nathanhammond:
> It reads the same way to me. "Promoting canary to beta" should ensure experiments are off. Releasing canary is the only time we should expect experiments to be on.

This is the portion of directions at the tail end, _after_ the release, syncing the state back up.

nathanhammond:
@homu r+

PR #6613, opened 2016-12-26
Add missing annotations.

rwjblue:
These annotations help us track things down while visualization...

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1121,7 +1121,9 @@ EmberApp.prototype.test = function() {
   let appTestTree = this.appTests(coreTestTree);
   let testFilesTree = this.testFiles(coreTestTree);
 
-  return mergeTrees([appTestTree, testFilesTree]);
+  return mergeTrees([appTestTree, testFilesTree], {
+    annotation: 'TreeMerger (test)',
+  });
 };
 
 /**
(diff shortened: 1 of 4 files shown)


Turbo87:
@rwjblue any idea where the build failures are coming from?

rwjblue:
Yes, I added `annotation` to calls to `broccoli-babel-transpiler`, and (as you know) it passes its options directly through to `babel` so babel is erroring that `annotation` isn't a valid babel option. 😿

Turbo87:
@rwjblue should we deprecated that type of usage in `broccoli-babel-transpiler`? we could instead use a `babel` key in the options hash

rwjblue:
Yes, we should basically deprecate not having a `options.babel` key...

Turbo87:
Yeah, that's what I meant. Similar to `rollup` in https://github.com/chadhietala/broccoli-rollup.

nathanhammond:
Can we just drop the two babel annotations and do the rest? And then file a tracking issue for the rest of the work?

rwjblue:
@nathanhammond - Makes sense, yeah I can do that.

rwjblue:
Updated broccoli-babel-transpiler to allow annotations in https://github.com/babel/broccoli-babel-transpiler/pull/108 and https://github.com/babel/broccoli-babel-transpiler/pull/107 (6.0-alpha), and rebased (geez we've been busy).

Assuming Travis is 👍, this is finally good to go...

nathanhammond:
@homu r+

PR #6615, opened 2016-12-27
Command interruption

### LINKED ISSUE #6421: [Windows] Ctrl+C produces Exit status 1

ro0gr:
Fixes #6421 
During implementation I've also noticed that we don't clean [tests tmp](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/test.js#L53) when the process is interrupted. This pr [fixes](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:command-interruption?expand=1#diff-39de54119935553ddd431dac09b49eafR161) it as well.

Depends on https://github.com/ember-cli/capture-exit/pull/18

The problem was that the process interruption was handled in the `model/builder`. It always sent an **exit code 1** when the process has been stopped from outside. 
Actually it's not a case for `server` commands like `ember s` or `ember t -s`. Interrupting of these commands should not produce an error code. 

Unfortunately builder has not enough knowledge which exit code to produce. It should be responsible only for clean up of self produced artifacts on unexpected exits.

That's the reason why I moved handling of interruption to the [`lib/cli/cli.js`](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:command-interruption?expand=1#diff-16a41bfbe414387d773df099f941217eR138). 
If something has sent an interruption signal to a cli while the command is active then the [command's interrupt method is called](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:command-interruption?expand=1#diff-16a41bfbe414387d773df099f941217eR139). It [calls the current task's interrupt method](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:command-interruption?expand=1#diff-5346944aa3595cbf8116950d83f80cfeR188).
Return value of `interrupt()` is handled by the `capture-exit`. It allows us to reject interruption promise with a given exit code.

I've tested basic flows against windows and linux with these changes included. Couldn't find any issues.
`/tmp` is cleaned-up after successful/failed builds and interruption. 
exit code 1 produced only when the build and test commands are interrupted in a single(not server) run mode.

There is still some work left to be done:
  - [x] land capture-exit 
  - [x] fix current tests
  - [x] `ember t -s` acceptance test([nice starting point](https://github.com/ember-cli/ember-cli/blob/49479996571cfd838bd0c403cd08358d6591a344/tests/acceptance/smoke-test-slow.js#L284)) for `/tmp` cleanup and **exit code 0** when command interrupted 
 - [ ] ~~`interrupt()` unit tests~~
 - [x] unit tests for `Command.runTask()` 
 - [x] tests for `will-interrupt-process`
 - [x] rebase
 - [x] change interrupt exit code management to use process.exit(1) rather than a rejecting promise.
 - [x] change interrupt semantics so default is to exit 1 and the long-running commands explicitly opt-in to a zero exit code
 - [x] improve the handler setup test if it's feasible
 - [x] cleanup from inline comments eg
   - this._super in test command rather than copy super impl
   - a couple of comments are out of place from copying
   - super minor linting things

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -7,10 +7,10 @@ const getOptionArgs = require('../utilities/get-option-args');
 let logger = require('heimdalljs-logger')('ember-cli:cli');
 let loggerTesting = require('heimdalljs-logger')('ember-cli:testing');
 const Instrumentation = require('../models/instrumentation');
-const exit = require('capture-exit');
 const heimdall = require('heimdalljs');
 
 const Promise = RSVP.Promise;
+const onProcessInterrupt = require('../utilities/will-interrupt-process');
 
 class CLI {
   /**
@@ -145,6 +145,7 @@ class CLI {
         }
       }
 
+      let onCommandInterrupt = command.onInterrupt.bind(command);
       let instrumentation = this.instrumentation;
       let initCompleted = false;
 
@@ -154,6 +155,7 @@ class CLI {
         instrumentation.start('command');
 
         loggerTesting.info('cli: command.beforeRun');
+        onProcessInterrupt.addHandler(onCommandInterrupt);
 
         return command.beforeRun(commandArgs);
 
@@ -167,14 +169,14 @@ class CLI {
         }
         instrumentation.start('shutdown');
 
+        onProcessInterrupt.removeHandler(onCommandInterrupt);
         shutdownOnExit = function() {
           instrumentation.stopAndReport('shutdown');
         };
 
-        // schedule this with `capture-exit` to ensure that
-        // the shutdown instrumentation hook is invoked properly even if
-        // we exit before hitting the `finally` below
-        exit.onExit(shutdownOnExit);
+        // Ensure that the shutdown instrumentation hook is invoked properly
+        // even if we exit before hitting the `finally` below
+        onProcessInterrupt.addHandler(shutdownOnExit);
       }).then(result => {
         // if the help option was passed, call the help command
         if (result === 'callHelp') {
@@ -205,15 +207,15 @@ class CLI {
         });
       });
     })
-      .finally(() => {
-        if (shutdownOnExit) {
-          // invoke instrumentation shutdown and
-          // remove from `capture-exit` callbacks
-          shutdownOnExit();
-          exit.offExit(shutdownOnExit);
-        }
-      })
-      .catch(this.logError.bind(this));
+    .finally(() => {
+      if (shutdownOnExit) {
+        // invoke instrumentation shutdown and
+        // remove instrumentation interruption handlers
+        shutdownOnExit();
+        onProcessInterrupt.removeHandler(shutdownOnExit);
+      }
+    })
+    .catch(this.logError.bind(this));
   }
 
   /**
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,9 +1,5 @@
 'use strict';
 
-// work around misbehaving libraries, so we can correctly cleanup before
-// actually exiting.
-require('capture-exit').captureExit();
-
 // Main entry point
 const requireAsHash = require('../utilities/require-as-hash');
 const packageConfig = require('../../package.json');
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -128,12 +128,6 @@ module.exports = Command.extend({
       port: this._generateTestPortNumber(commandOptions),
     }, this._generateCustomConfigs(commandOptions));
 
-    let options = {
-      ui: this.ui,
-      analytics: this.analytics,
-      project: this.project,
-    };
-
     return Win.checkIfSymlinksNeedToBeEnabled(this.ui).then(() => {
       let session;
 
@@ -142,39 +136,32 @@ module.exports = Command.extend({
           throw new SilentError('Specifying a build is not allowed with the `--server` option.');
         }
 
-        let TestServerTask = this.tasks.TestServer;
-        let testServer = new TestServerTask(options);
         let builder = new this.Builder(testOptions);
 
-        testOptions.watcher = new this.Watcher(assign(options, {
+        testOptions.watcher = new this.Watcher(assign(this._env(), {
           builder,
           verbose: false,
           options: commandOptions,
         }));
 
-        session = testServer.run(testOptions)
-          .finally(() => builder.cleanup());
-
+        session = this.runTask('TestServer', testOptions).finally(() => builder.cleanup());
+      } else if (hasBuild) {
+        session = this.runTask('Test', testOptions);
       } else {
-        let TestTask = this.tasks.Test;
-        let test = new TestTask(options);
-
-
-        if (hasBuild) {
-          session = test.run(testOptions);
-        } else {
-          let BuildTask = this.tasks.Build;
-          let build = new BuildTask(options);
-
-          session = build.run({
-            environment: commandOptions.environment,
-            outputPath,
-          })
-          .then(() => test.run(testOptions));
-        }
+        session = this.runTask('Build', {
+          environment: commandOptions.environment,
+          outputPath,
+        })
+        .then(() => this.runTask('Test', testOptions));
       }
 
       return session.finally(this.rmTmp.bind(this));
     });
+  },
+
+  onInterrupt() {
+    this.rmTmp();
+
+    return this._super.onInterrupt.apply(this, arguments);
   },
 });
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -177,6 +177,46 @@ let Command = CoreObject.extend({
   },
 
   /**
+   * Called when command is interrupted from outside, e.g. ctrl+C or process kill
+   * Can be used to cleanup artifacts produced by command and control process exit code
+   *
+   * @method onInterrupt
+   * @return {Promise|undefined} if rejected promise then result of promise will be used as an exit code
+   */
+  onInterrupt() {
+    if (this._currentTask) {
+      return this._currentTask.onInterrupt();
+    }
+  },
+
+  _env() {
+    return {
+      ui: this.ui,
+      analytics: this.analytics,
+      project: this.project,
+    };
+  },
+
+  runTask(name, options) {
+    if (this._currentTask) {
+      throw new Error(`Concurrent tasks are not supported`);
+    }
+
+    let Task = this.tasks[name];
+    if (!Task) {
+      throw new Error(`Unknown task "${name}"`);
+    }
+
+    let task = new Task(this._env());
+
+    this._currentTask = task;
+
+    return Promise.resolve(task.run(options)).finally(() => {
+      delete this._currentTask;
+    });
+  },
+
+  /**
     Hook for extending a command before it is run in the cli.run command.
     Most common use case would be to extend availableOptions.
     @method beforeRun
--- a/lib/models/task.js
+++ b/lib/models/task.js
@@ -6,6 +6,17 @@ class Task extends CoreObject {
   run(/*options*/) {
     throw new Error('Task needs to have run() defined.');
   }
+
+  /**
+   * Interrupt comamd with an exit code
+   * Called when the process is interrupted from outside, e.g. CTRL+C or `process.kill()`
+   *
+   * @private
+   * @method onInterrupt
+   */
+  onInterrupt() {
+    process.exit(1);
+  }
 }
 
 module.exports = Task;
--- a/lib/utilities/will-interrupt-process.js
+++ b/lib/utilities/will-interrupt-process.js
@@ -1,0 +1,145 @@
+'use strict';
+
+// (un)Register process interruption signals(SIGINT, SIGTERM, process.kill()) handlers.
+//
+// When the process interrupted("CTR+C") all the registered handlers are being invoked via `capture-exit`.
+// The process finishes after the last interruption handler is settled.
+
+let exit = require('capture-exit');
+
+let windowsCtrlCTrap;
+let handlers = [];
+
+module.exports = {
+  /**
+   * Add process interruption handler
+   *
+   * When the first handler is added then automatically
+   * sets up process interruption signals listeners
+   *
+   * @private
+   * @method addHandler
+   * @param {function} cb   Callback to be called when process interruption fired
+   */
+  addHandler(cb) {
+    let index = handlers.indexOf(cb);
+    if (index > -1) { return; }
+
+    if (handlers.length === 0) {
+      setupSignalsTrap();
+    }
+
+    handlers.push(cb);
+    exit.onExit(cb);
+  },
+
+  /**
+   * Remove process interruption handler
+   *
+   * If there are no remaining handlers after removal
+   * then clean up all the process interruption signal listeners
+   *
+   * @private
+   * @method removeHandler
+   * @param {function} cb   Callback to be removed
+   */
+  removeHandler(cb) {
+    let index = handlers.indexOf(cb);
+    if (index < 0) { return; }
+
+    handlers.splice(index, 1);
+    exit.offExit(cb);
+
+    if (handlers.length === 0) {
+      teardownSignalsTrap();
+    }
+  },
+
+  /**
+   * Tears down all the handlers and cleans up all signal listeners
+   *
+   * @private
+   * @method teardown
+   */
+  teardown() {
+    while (handlers.length > 0) {
+      this.removeHandler(handlers[0]);
+    }
+  },
+};
+
+/**
+ * Sets up listeners for interruption signals
+ *
+ * When one of these signals is caught than raise process.exit()
+ * which enforces `capture-exit` to run registered interruption handlers
+ *
+ * @method setupSignalsTrap
+ */
+function setupSignalsTrap() {
+  process.on('SIGINT', _exit);
+  process.on('SIGTERM', _exit);
+  process.on('message', onMessage);
+  if (/^win/.test(process.platform)) {
+    trapWindowsSignals();
+  }
+
+  exit.captureExit();
+}
+
+/**
+ * Removes interruption signal listeners and tears down capture-exit
+ *
+ * @method teardownSignalsTrap
+ */
+function teardownSignalsTrap() {
+  process.removeListener('SIGINT', _exit);
+  process.removeListener('SIGTERM', _exit);
+  process.removeListener('message', onMessage);
+  if (/^win/.test(process.platform)) {
+    cleanupWindowsSignals();
+  }
+
+  exit.releaseExit();
+}
+
+/**
+ * Supresses "Terminate batch job (Y/N)" confirmation on Windows
+ *
+ * @method trapWindowsSignals
+ */
+function trapWindowsSignals() {
+  // This is required to capture Ctrl + C on Windows
+  if (process.stdin && process.stdin.isTTY) {
+    process.stdin.setRawMode(true);
+    windowsCtrlCTrap = function(data) {
+      if (data.length === 1 && data[0] === 0x03) {
+        process.emit('SIGINT');
+      }
+    };
+    process.stdin.on('data', windowsCtrlCTrap);
+  }
+}
+
+function cleanupWindowsSignals() {
+  if (windowsCtrlCTrap && process.stdin.removeListener) {
+    process.stdin.removeListener('data', windowsCtrlCTrap);
+  }
+}
+
+/**
+ * Handles the `message` event on the `process`.
+ *
+ * Calls `process.exit` if the `kill` property on the `message` is set.
+ *
+ * @method onMessage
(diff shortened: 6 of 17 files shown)


lib/commands/test.js
```diff
@@ -137,42 +132,36 @@ module.exports = Command.extend({
...
       }
 
       return session.finally(this.rmTmp.bind(this));
     }.bind(this));
+  },
+
+  onInterrupt: function() {
+    this.rmTmp();
+
+    if (this._currentTask) {
```
ro0gr: `super()` can be called

lib/tasks/build.js
```diff
@@ -61,5 +62,9 @@ module.exports = Task.extend({
 
         throw err;
       });
+  },
+
+  onInterrupt: function() {
+    return RSVP.Promise.reject(1);
```
ro0gr: Looks like `onInterrupt` should be able just to return an exit code without a promise if no async work is necessary

ro0gr:
I think current implementation contains several arguable places. Seems like splitting it to different PRs  might be a good idea.

lib/tasks/build.js
hjdivad: Does `onInterrupt` actually have any ability to control exit code?  I think any handler that rejects exist with code 1: https://github.com/ember-cli/capture-exit/blob/2a2ba681721f65f7213dfc67e621aaa922eb8320/index.js#L97

That said it does seem reasonable to allow handlers to reject with an exit code, although it raises the issue of what to do when you have `process.exit(a)` and a handler that rejects `b`.  Likely the right code in this case is `a`.

Anyway I bring it up here but the right place for hashing it out will be on https://github.com/ember-cli/capture-exit/pull/18

lib/utilities/will-interrupt-process.js
```diff
@@ -0,0 +1,140 @@
+'use strict';
+
+// work around misbehaving libraries, so we can correctly cleanup before actually exiting.
```
hjdivad: I wouldn't put this comment here with the code as-is.  It implies that we already capture exit by requiring, but we don't until we call `captureExit`.

That said I think there's a bigger issue mentioned above, which is that we want `captureExit` to be called much earlier (pretty much ASAP).

lib/utilities/will-interrupt-process.js
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    teardownSignalsTrap();
+  }
+};
+
+/**
+ * @public
+ * @method setup
+ */
```
hjdivad: these yuidocs don't look correctly any more; i think they're leftover from when this code was moved

lib/models/task.js
```diff
@@ -5,7 +5,17 @@ var CoreObject = require('core-object');
...
+
+  /**
+   * Called when the is interrupted from outside, e.g. CTRL+C or process kill
+   * Can be used to cleanup artifacts produced by the task
+   * Return value affects a process exit code
+   *
+   * @method onInterrupt
+   * @return {Promise|undefined} if rejected promise then result of promise will be used as an exit code
+   */
+  onInterrupt: function() {},
```
hjdivad: Is this the right default?  I would have thought that for most tasks we would expect a non-zero exit on `SIGINT` (eg npm install, addon install, test, build, generate from blueprint) and the zero-exits on `SIGINT` are the exceptions, for infinite running tasks (basically anything that uses a watcher).

lib/models/command.js
```diff
@@ -178,6 +178,38 @@ var Command = CoreObject.extend({
...
+   * Called when command is interrupted from outside, e.g. ctrl+C or process kill
+   * Can be used to cleanup artifacts produced by command and control process exit code
+   *
+   * @method onInterrupt
+   * @return {Promise|undefined} if rejected promise then result of promise will be used as an exit code
+   */
+  onInterrupt: function() {
+    if (this._currentTask) {
+      return this._currentTask.onInterrupt();
+    }
```
hjdivad: if we interrupt a command when there is no current task, should we expect a zero or nonzero exit code? If a user exists during init before task is started, it seems odd to me that we'd have a zero exit code in that scenario

hjdivad:
@ro0gr thanks so much for this contribution.  My review has lots of little nits but I hope it doesn't seem negative; I definitely feel this is on the right track and will make things better when it lands!

lib/tasks/build.js
hjdivad: Actually the more idiomatic way of doing this is probably just to call `process.exit(code)` and it saves us from having to figure out the various different cases in which we exit, possibly with a `process.exit`-provided code, and also with a handler-provided code.

ro0gr: I'll try `process.exit(code)` approach as you suggested. Seems like I need to better understand how does `capture-exit` actually works

lib/cli/index.js
```diff
@@ -1,9 +1,5 @@
 'use strict';
 
-// work around misbehaving libraries, so we can correctly cleanup before
-// actually exiting.
-require('capture-exit').captureExit();
-
```
ro0gr: As far as I understand we actually use `capture-exit` to ensure that all exit handlers are finished.
If that is true seems like if we have no exit handlers we don't need a captured exit.

That's why I've moved `captureExit()` to `will-interrupt-process` utility which automatically captures exit once the first exit handler is registered.

lib/commands/test.js
```diff
@@ -137,42 +131,40 @@ module.exports = Command.extend({
...
+      return this._currentTask.onInterrupt();
+    }
```
ro0gr: yes

lib/models/command.js
ro0gr: Agree, I think the same question is relevant for the case when the `cli/cli` is interrupted when there is still no active command.

lib/models/task.js
ro0gr: You are probably right. exitCode 1 looks better as a default here

lib/cli/index.js
hjdivad: Okay that's pretty convincing.  👍

hjdivad:
We would also want the serve task's `onInterrupt` to either resolve or reject its [hanging promise](https://github.com/ember-cli/ember-cli/blob/5f3b83fcbb41a63e93c24975cbeaef71f5e0947c/lib/tasks/serve.js#L60-L62) so that we can do other end-of-command things like invoke instrumentation hooks.

ro0gr:
> We would also want the serve task's onInterrupt to either resolve or reject its hanging promise 

Like the idea. 

Just to inform. At this moment I'm unable to invest enough time to move this pr further. 
I just don't want to hurry up. I suspect I can renew this effort in the beginning of the next week and provide something valuable in the mid of the next week.

ro0gr:
Also I believe this pause could give us enough time to finish https://github.com/ember-cli/capture-exit/pull/18 and to think more about 
> we should figure out whether or not handlers ought to be able to set exit  

and make a decision

hjdivad:
@ro0gr re: capture-exit I strongly suspect the answer is that handlers should not be able to set exit code except by calling `process.exit(code)`.  It's pretty finicky to deal with a `process.exit(code)` provided code as well as a handler's code.

hjdivad:
@ro0gr were you able to move this forward?  No worries if you don't have time, the current PR is already excellent and I can do the final refinements.

ro0gr:
@hjdivad yes, sorry for delay (:
just started looking in the capture-exit tests today. 

Feel free to make anything you need with this pr. I can try to help with follow up tests via separate prs.

hjdivad:
@ro0gr awesome.  I'm excited to have this in.  Will make things nicer and in any case we need it for the instrumentation hooks to be called properly in `ember serve`

ro0gr:
Should I close this pr?

lib/cli/cli.js
```diff
@@ -155,9 +160,12 @@ CLI.prototype.run = function(environment) {
           resolve(exitCode);
         }
       });
-    }.bind(this));
-
-  }.bind(this)).catch(this.logError.bind(this));
+    })
+    .finally(function() {
+      onProcessInterrupt.removeHandler(onCommandInterrupt);
+    });
```
hjdivad: nvm i forgot the ordering semantics of `finally` within `RSVP`

lib/utilities/will-interrupt-process.js
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @public
+   * @method addHandler
+   * @param {function} cb   Callback to be called when process interruption fired
+   */
+  addHandler: function(cb) {
+    var index = handlers.indexOf(cb);
+    if (index > -1) { return; }
+
+    if (!handlers.length) {
```
hjdivad: -> `handlers.length === 0` for clarity

lib/utilities/will-interrupt-process.js
```diff
@@ -0,0 +1,140 @@
...
+   * @param {function} cb   Callback to be removed
+   */
+  removeHandler: function(cb) {
+    var index = handlers.indexOf(cb);
+    if (index < 0) { return; }
+
+    handlers.splice(index, 1);
+    exit.offExit(cb);
+
+    if (!handlers.length) {
```
hjdivad: -> `handlers.length === 0` for clarity

tests/unit/cli/cli-test.js
```diff
@@ -133,6 +145,29 @@ describe('Unit: CLI', function() {
...
   });
 
+  describe('process signal listeners', function() {
+    var onCommandInterrupt;
+    beforeEach(function() {
+      onCommandInterrupt = td.matchers.isA(Function);
+    });
+
+    it('sets up signal listeners before command started', function() {
+      stubValidateAndRun('serve', function() {
```
hjdivad: Can we verify that the handler is added before the command is invoked?  I don't think it's critical, but it would be a nice improvement to the test if it doesn't cause too much churn elsewhere in the file

hjdivad:
@ro0gr This looks great.  I think the remaining things to do before the pr can land are:

- [ ] rebase
- [ ] change interrupt exit code management to use `process.exit(1)` rather than a rejecting promise.
- [ ] change interrupt semantics so default is to exit 1 and the long-running commands explicitly opt-in to a zero exit code
- [ ] minor cleanup from inline comments eg
  - `this._super` in test command rather than copy super impl
  - a couple of comments are out of place from copying
  - super minor linting things
  - improve the handler setup test if it's feasible

Although I think the minor cleanup can totally happen in a followup pr if needed.  The only important stuff is rebasing, changing the defaults and correct use of capture-exit.

Awesome work!

ro0gr:
> improve the handler setup test if it's feasible

@hjdivad what about acceptance tests for exit codes and /tmp cleanup?

> acceptance tests([nice starting point](https://github.com/ember-cli/ember-cli/blob/49479996571cfd838bd0c403cd08358d6591a344/tests/acceptance/smoke-test-slow.js#L284)) for /tmp cleanup and exit codes when command interrupted
>  - build [--watch]
>  - serve
>  - test [-serve]

I think it may be good idea to have this covered?

hjdivad:
@ro0gr yeah a single end-to-end smoketest on command interruption seems 👍  as long as we keep the high coverage area in unit tests.

lib/cli/cli.js
```diff
@@ -167,14 +169,15 @@ class CLI {
         }
         instrumentation.start('shutdown');
 
+        onProcessInterrupt.removeHandler(onCommandInterrupt);
         shutdownOnExit = function() {
           instrumentation.stopAndReport('shutdown');
         };
 
         // schedule this with `capture-exit` to ensure that
```
ro0gr: to be updated

lib/cli/cli.js
```diff
@@ -205,15 +208,15 @@ class CLI {
...
-          // invoke instrumentation shutdown and
-          // remove from `capture-exit` callbacks
-          shutdownOnExit();
-          exit.offExit(shutdownOnExit);
-        }
-      })
-      .catch(this.logError.bind(this));
+    .finally(() => {
+      if (shutdownOnExit) {
+        // invoke instrumentation shutdown and
```
ro0gr: comment to be updated

tests/unit/cli/cli-test.js
ro0gr: I think it could be easy to achieve if the `cli` raised some kind of events like "command:beforeRun".
Seems like [`instrumentation`](https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L154-L156) does something similar but as I understand it's not supposed to be a notifier.. 

Is it ok(from the aesthetic standpoint) to stub `instrumentation.start('command');` in order to know the moment when we can verify a handler?

hjdivad: Yeah `instrumentation` isn't a notifier.  But there is actually a `beforeRun` method that's invoked before a command is run.  see https://github.com/ember-cli/ember-cli/blob/master/tests/unit/cli/cli-test.js#L184-L221

ro0gr: exactly what's needed! thanks

ro0gr:
@hjdivad @stefanpenner @nathanhammond I believe this is ready for review

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,83 @@ describe('models/command.js', function() {
...
+        tasks: {
+          Async: AsyncTask,
+          Sync: SyncTask,
+          Failing: SyncTask,
+        },
+      }));
+    });
+
+    it('always handles task as a promise', function() {
+      command.runTask('Sync', { param: 'value' }).then(result => {
```
hjdivad: we need to return promises in mocha tests no?

I think this should be `return command.runTask( ...`  to avoid flaky races in tests

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,83 @@ describe('models/command.js', function() {
...
+        expect(result).to.eql({
+          param: 'value',
+        });
+      });
+    });
+
+    it('command environment should be shared with a task', function() {
+      command.runTask('Async', { param: 'value' });
```
hjdivad: return promises for async tests

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,83 @@ describe('models/command.js', function() {
...
+      expect(command._currentTask.ui).to.eql(command.ui);
+      expect(command._currentTask.analytics).to.eql(command.analytics);
+      expect(command._currentTask.project).to.eql(command.project);
+    });
+
+    it('_currentTask should store a reference to the current task', function() {
+      expect(command._currentTask).to.be.undefined;
+      command.runTask('Sync', { param: 'value' }).then(result => {
```
hjdivad: return promises for async tests

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,83 @@ describe('models/command.js', function() {
...
+        expect(command._currentTask).to.be.undefined;
+      });
+      expect(command._currentTask).to.be.an.instanceof(SyncTask);
+    });
+
+    it('_currentTask should cleanup current task on fail', function() {
+      command.runTask('Failing', { param: 'value' }).then(result => {
```
hjdivad: return promises for async tests

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,83 @@ describe('models/command.js', function() {
...
+        expect(command._currentTask).to.be.undefined;
+      });
+    });
+
+    it('throws on attempt to launch concurrent task', function() {
+      try {
```
hjdivad: return promises for async tests

I realize the second `runTask` will not create a promise, but the first one will

ro0gr:
@hjdivad updated per your request. waiting for CI

tests/unit/models/command-test.js
```diff
@@ -558,6 +561,96 @@ describe('models/command.js', function() {
...
+    });
+
+    it('_currentTask should cleanup current task on fail', function() {
+      return command.runTask('Failing', { param: 'value' })
+        .then(() => expect(false, 'should not be resolved').to.equal(true))
+        .catch(result => expect(command._currentTask).to.be.undefined);
+    });
+
+    it('throws on attempt to launch concurrent tasks', function() {
+      let asyncTaskRun, syncTaskRun;
```
hjdivad: `syncTaskRun seems unused?

hjdivad:
awesome thanks @ro0gr 

looks like you had a super minor lint-style issue both otherwise 👍

hjdivad:
appveyor seems to be a bit finnicky though :(

tests/unit/models/command-test.js
ro0gr: oops...

I've decided to handle all the tasks as promises to make tests bullet proof. I'm going to use this variable now.

ro0gr:
Not sure what can I do with appveyor. I believe it's not related

hjdivad:
@ro0gr appveyor is definitely not your fault; do you want to amend the commit to kill the stray var?  it's super minor so either way i'm going to merge this ^_^

ro0gr:
updated with this fix included

hjdivad:
@homu r+

hjdivad:
thanks for the contribution @ro0gr!

rwjblue:
@homu r+

lib/models/task.js
```diff
@@ -6,6 +6,17 @@ class Task extends CoreObject {
...
+
+  /**
+   * Interrupt comamd with an exit code
+   * Called when the process is interrupted from outside, e.g. CTRL+C or `process.kill()`
+   *
+   * @private
+   * @method onInterrupt
+   */
+  onInterrupt() {
+    process.exit(1);
```
stefanpenner: having nested exists seems problematic, it would likely be better if we just rejected a given commands run task. Then allowing the primary exit path handle it.

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/7019 addresses and issue where if a command fails the signal handler was not removed.

lib/models/task.js
hjdivad: We still want to communicate exit status.  Letting handlers just call `process.exit` seems like the most straightforward way of doing that, although I agree this is mega troll for lib usage.

Having said that, since we only have two statuses at the moment we could determine status from the final state of the command promise.  Otherwise seems we need some other out of band way to communicate exit status.

Turbo87: > Otherwise seems we need some other out of band way to communicate exit status

for zero and one it's obviously pretty easy, if we actually feel like we need more we could just add an `exitCode` property to the `Error` instance and use that instead if it exists.

hjdivad: @Turbo87 seems 👍 to me

ro0gr: Does anybody work to address process.exit issue for lib usage case? If not I would like to fix it this week

stefanpenner: nope not yet

note: https://github.com/ember-cli/ember-cli/pull/7019#pullrequestreview-36374109 does tidy things up but does not address ^

PR #6617, opened 2016-12-27
Remove wasted work around addon's addon trees.

rwjblue:
Prior to this change, after retrieving the `addon` trees for all addons we then immediately concat the assets down into a single `addons.{js,css}` file (after merging and funneling to a different location).

The long term objective is to keep "loose modules" throughout the entire pipeline and eventually replace the final concat stages that we are currently using with a higher level "packager". This takes a very small step towards that future...

The good news is that this is actually a win regardless of that future state (we reduced the number of build steps by around 4 with this change).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -911,32 +911,17 @@ EmberApp.prototype._addonTree = function _addonTree() {
     return this._cachedAddonTree;
   }
 
-  let addonTrees = mergeTrees(this.addonTreesFor('addon'), {
+  let addonTrees = this.addonTreesFor('addon');
+
+  let combinedAddonTree = mergeTrees(addonTrees, {
     overwrite: true,
-    annotation: 'TreeMerger (addons)',
+    annotation: 'TreeMerger: `addon/` trees',
   });
 
-  let addonTranspiledModules = new Funnel(addonTrees, {
-    srcDir: 'modules',
-    allowEmpty: true,
-    annotation: 'Funnel: Addon JS',
+  return this._cachedAddonTree = new Funnel(combinedAddonTree, {
+    destDir: 'addon-tree-output',
+    annotation: 'Funnel: addon-tree-output',
   });
-
-  return this._cachedAddonTree = [
-    this._concatFiles(addonTrees, {
-      inputFiles: ['**/*.css'],
-      outputFile: '/addons.css',
-      allowNone: true,
-      annotation: 'Concat: Addon CSS',
-    }),
-
-    this._concatFiles(addonTranspiledModules, {
-      inputFiles: ['**/*.js'],
-      outputFile: '/addons.js',
-      allowNone: true,
-      annotation: 'Concat: Addon JS',
-    }),
-  ];
 };
 
 /**
@@ -949,8 +934,7 @@ EmberApp.prototype._processedVendorTree = function() {
     return this._cachedVendorTree;
   }
 
-  let trees = this._addonTree();
-  trees = trees.concat(this.addonTreesFor('vendor'));
+  let trees = this.addonTreesFor('vendor');
 
   if (this.trees.vendor) {
     trees.push(this.trees.vendor);
@@ -982,14 +966,16 @@ EmberApp.prototype._processedExternalTree = function() {
 
   let vendor = this._processedVendorTree();
   let bower = this._processedBowerTree();
+  let addons = this._addonTree();
 
-  let trees = [vendor];
+  let trees = [vendor].concat(addons);
   if (bower) {
     trees.unshift(bower);
   }
 
   let externalTree = mergeTrees(trees, {
     annotation: 'TreeMerger (ExternalTree)',
+    overwrite: true,
   });
 
   if (this.amdModuleNames) {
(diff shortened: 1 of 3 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1275,17 +1251,21 @@ EmberApp.prototype.javascript = function() {
   }
 
   this.import('vendor/ember-cli/vendor-prefix.js', {prepend: true});
-  this.import('vendor/addons.js');
-  this.import('vendor/ember-cli/vendor-suffix.js');
```
nathanhammond: and if you add a setter to `_scriptOutputFiles` or overwrite `import`, seems fine? It's your own foot and you should have known better?

tests/unit/broccoli/ember-app-test.js
```diff
@@ -1095,40 +1093,23 @@ describe('broccoli/ember-app', function() {
 
         app.styles(); // run
 
-        expect(count).to.eql(3);
+        expect(count).to.eql(1);
 
         expect(args[0]).to.deep.eql({
-          allowNone: true,
-          annotation: 'Concat: Addon CSS',
```
nathanhammond: Nice. :+1:

lib/broccoli/ember-app.js
```diff
@@ -1341,15 +1321,17 @@ EmberApp.prototype.styles = function() {
     }.bind(this));
   }
 
-  this.import('vendor/addons.css');
-
   var vendorStyles = [];
   for (var outputFile in this._styleOutputFiles) {
+    var isMainVendorFile = outputFile === this.options.outputPaths.vendor.css;
     var headerFiles = this._styleOutputFiles[outputFile];
+    var inputFiles = isMainVendorFile ? ['modules/**/*.css'] : [];
```
rwjblue: This is specifically for addon/styles which comes from the addon tree (and was previously being concatted into addons.css (near one of your other comments). It definitely is namespaced to modules/.

rwjblue: bower_components is here as bower_components/ and vendor is at vendor/. This concat does not include either of them (which is why we use `modules/**`), but makes them available for you if you had app.import'ed from them (which is how app.import had always worked).

rwjblue: See https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L920-L951, which shows that this is the current semantics and not regressed by this PR.

lib/broccoli/ember-app.js
rwjblue: Yes, don't do that. 😜

rwjblue:
> I'd rather leave the CSS alone until I get it extracted into ember-cli-css.

I definitely understand the motivation here, but I disagree. Making things saner and less coupled can only make extraction easier.

lib/broccoli/ember-app.js
nathanhammond: 1. [Shows that we have a `modules` folder containing JS.](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L931)
2. [Pulls assets from _anywhere_ inside of the `treeForAddon` results.](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L937), not just the `modules` folder (note [which tree we feed it](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L925)).

It can do this because it is invoked before being combined with the `externalTree`. I _believe_ that your `concat` behavior is thus still broken. (I should actually confirm this rather than just statically analyzing the code but I'm on a plane and that's a pain in the ass to get a setup for since I've set up for my other PR.)

nathanhammond:
> I definitely understand the motivation here, but I disagree. Making things saner and less coupled can only make extraction easier.

In an earlier edit of that I had couched it with something about timeline and progress but because in my review it felt like the behavior as built was broken I went ahead and dropped it. I don't mind progress, but I'm a bit grumpy when I don't feel confident in it.

lib/broccoli/ember-app.js
nathanhammond: Also, anybody overrides `treeForAddon` (public!) and all bets are off.

rwjblue: > Also, anybody overrides treeForAddon (public!) and all bets are off.

We were already only concatting JS from `modules/` dir. `treeForAddon` is expected to return a tree with that directory (which is another thing that I dislike and want to kill, but takes time).  So we are definitely safe here WRT `*.js` assets. 

> Pulls assets from anywhere inside of the treeForAddon results.

Agreed, this is a great point (for those following along at home, @nathanhammond just correctly pointed out that we were only concatting `*.js` files from `modules/` but we would concat `*.css` files from anywhere in the tree returned by `treeForAddon`). I will address shortly.

rwjblue: OK, updated. Now includes all JS and CSS output from `treeForAddon`, and allowed for more build step removal.

rwjblue:
After last round of updates (thanks to a few passes of review by @nathanhammond) we have reduced build step count by 22 (for `travis-web`'s build).

rwjblue:
@nathanhammond - Would love for you to take another look after fixing that issue you brought up...

rwjblue:
@nathanhammond - Now that #6652 landed and reverted at least _some_ of the bad behavior could you double check your tests to see if this is unblocked?

nathanhammond:
@rwjblue I need to finish writing the tests for `treeForAddon` and `.import` (started here: https://github.com/ember-cli/ember-cli/pull/6449/files#diff-ae7bd994ed76e9420ab8f22034b3d5d2R21) which I didn't add in yet. I spent today writing documentation so that these tools can be landed and I don't have to keep rebasing everything on top of them to run tests. (Then moved into blog post drafting for tomorrow's release.)

Wednesday is still my target day to confirm/deny this.

rwjblue:
Now that ember-cli@2.12 is cut (and we are separated by another full canary + beta cycle from the other large build tooling changes) I'd like to push this forward...

stefanpenner:
@rwjblue re-reviewing now.

rwjblue:
@homu r+

nathanhammond:
:/ on this landing, my apologies for slowness. I'm trying to make it possible to land the testing tools when this isn't as high on my priority list anymore. This is the loose assets test we want: https://github.com/ember-cli/ember-cli/pull/6449/files#diff-ae7bd994ed76e9420ab8f22034b3d5d2

As it stands it is passing but there are a few more scenarios that I've not actually generated. @rwjblue I'd feel much more comfortable with additional testing scenarios. Trying to land by Monday. If I can avoid rebasing then the PR tests will be behind this, and @homu will test against head. 😛

PR #6618, opened 2016-12-28
Avoid extra build steps for `app/` with only `.gitkeep` file.

rwjblue:
The app tree's of each addon are gathered and then merged down into a single tree. The default addon blueprint includes writing to `app/.gitkeep`, which forces every addon to "have" an app tree (even though they may not have any files other than `.gitkeep`).

This change adds a detection for `app/.gitkeep` only (by using `fs.readdirSync`), and if only a `.gitkeep` is found we do not return the app tree.  We are mostly specializing this guard to check against what our default blueprint creates (which is a single `.gitkeep` based `app` tree).

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -34,6 +34,7 @@ var findAddonByName = require('../utilities/find-addon-by-name');
 var experiments = require('../experiments');
 var heimdall = require('heimdalljs');
 var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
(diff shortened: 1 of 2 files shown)


rwjblue:
After thinking about it and testing in an addon, this feels kinda footguney to me. Basically, this change makes the following:

```
ember addon foo
cd foo
ember serve
# separate terminal
ember generate component foo-bar
# `{{foo-bar}}` does not work because this PR noops the app tree
```

PR #6622, opened 2016-12-31
Remove support for Node 0.12.

rwjblue:
* Remove 0.12 from `.travis.yml`
* Update `engines` in `package.json` to indicate `4.5` is minimum version.
* Tweak language in `CONTRIBUTING.md` to suggest using "latest LTS" (instead of specific version).

---

Node 0.12 has already exited maintenance mode in a large percentage of the globe (Oceana, much of Asia, etc). Taking advantage of the one time timezones have worked in our favor...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 
 language: node_js
 node_js:
-  - "0.12"
   - "4"
   - "6"
   - "7"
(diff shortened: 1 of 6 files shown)


Turbo87:
@homu r+

🎉

nathanhammond:
I was hoping to do this after the next cutover. :/ Should still be fine since we haven't declared open season.

PR #6624, opened 2017-01-01
Update version of ember-cli-eslint used in new applications.

rwjblue:
New applications will now get ember-cli-eslint@3 which uses the latest version of eslint (we previously could not do this update due to eslint@3 dropping support for Node 0.12).

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -131,7 +131,7 @@ module.exports = Command.extend({
         if (!commandOptions.skipNpm) {
           return addonInstall.run({
             verbose: commandOptions.verbose,
-            packages: ['ember-cli-eslint@2']
+            packages: ['ember-cli-eslint@3']
           });
         }
       })


nathanhammond:
Given more time to think about it, I'm not 100% happy with how `ember-cli-eslint` isn't in the default blueprint anymore. Many users don't use `ember init` to handle upgrades which means that they end up "stuck" on whichever version they have installed, or worse, delete it without paying attention as to why [when they view the comparison](https://github.com/ember-cli/ember-cli/blob/master/dev/add-to-output-repos.sh#L35).

This whole setup exists for one reason: we need to be able to run all default blueprints for bundled addons once per application creation. We possibly want to run the default blueprint for every addon in the application each time that addon gets updated. But we can't run that blueprint _unless_ we actually have a `node_modules` folder. This complicates [programmatic fixture creation](https://github.com/ember-cli/ember-cli/pull/6449) as well.

/cc @Turbo87

Turbo87:
@nathanhammond I'm open to other suggestions...

PR #6625, opened 2017-01-01
Update dependencies previous avoided due to Node 0.12 support.

rwjblue:


Turbo87:
seems good 👍 

@homu r+

PR #6627, opened 2017-01-01
Lazily require `broccoli-babel-transpiler`.

rwjblue:
`broccoli-babel-transpiler` is only used during build related commands (and only in fallback scenarios), moving this to be a bit lazier shaves about 450ms from every `ember *` command invocation.

---

Before:

```
% \time node -e "require('./lib/models/addon');"
        0.61 real         0.51 user         0.10 sys
% cd empty-app
% \time ember v
        1.25 real         0.89 user         0.26 sys
```

After:

```
% \time node -e "require('./lib/models/addon');"
        0.16 real         0.14 user         0.03 sys
% cd empty-app
% \time ember v
        0.79 real         0.60 user         0.16 sys
```


rwjblue:
@homu r+

Turbo87:
Awesome! 🎉

PR #6628, opened 2017-01-01
Ensure `beforeRun` is included in `command` instrumentation.

rwjblue:
Prior to this change `command.beforeRun` was being included in `init` instrumentation. This change fixes that by moving `.beforeRun` into `command` instrumentation.

This change also ensures that `beforeRun` can return a promise (which we had previously documented in the API docs but was not actually supported until this commit) and adds a test.


rwjblue:
@homu r+

PR #6629, opened 2017-01-01
Enable more ESLint rules

Turbo87:
No ES6 stuff yet, just general cleanup.


Turbo87:
@ember-cli/core comments please :)

.eslintrc.js
```diff
@@ -10,113 +13,84 @@ module.exports = {
   rules: {
```
hjdivad: should we add [dot-location](http://eslint.org/docs/rules/dot-location)?

hjdivad:
> I reviewed the rule changes in .eslintrc.js and they look good to me, but I have not reviewed the full diff.

I reviewed the same as @rwjblue.  I could almost 👍  his comment except I'm enthusiastic about trailing commas ^_^

Turbo87:
@hjdivad since `dot-location` can be controversial I left it out of the PR. we should probably discuss it outside of this PR and then open another one if we feel like we should add it.

hjdivad:
> @hjdivad since dot-location can be controversial I left it out of the PR. we should probably discuss it outside of this PR and then open another one if we feel like we should add it.

My thinking was that while dot-location can be controversial, `consistency > mixed` would be uncontroversial even for consistency of someone's unpreferred approach.

That said, deferring to a follow up PR sounds sensible to me 👍

rwjblue:
> My thinking was that while dot-location can be controversial, consistency > mixed would be uncontroversial even for consistency of someone's unpreferred approach.

Agreed. I could definitely live with either, but prefer `property` personally...

Turbo87:
Rebased and fixed the conflicts. I'll merge once CI passes.

Turbo87:
actually... @homu r+

PR #6631, opened 2017-01-02
🏎 Lazily install "bower" if required

Turbo87:

### DIFF
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -8,6 +8,15 @@ var Task = require('../models/task');
 var formatPackageList = require('../utilities/format-package-list');
 
 module.exports = Task.extend({
+
+  ensureBower: function() {
+    if (this.bower) {
+      return Promise.resolve(this.bower);
+    }
+
+    return Promise.resolve(require('bower'));
+  },
+
   // Options: Boolean verbose
   run: function(options) {
     var bowerJson = path.join(this.project.root, '/bower.json');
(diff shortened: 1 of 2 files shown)


lib/tasks/bower-install.js
```diff
@@ -3,68 +3,97 @@
...
+              message.data.name + ' ' + pick.endpoint.target);
+          });
+        } else if (message.level === 'info' && options.verbose) {
+          // e.g.
+          //   cached git://example.com/some-package.git#1.0.0
+          ui.writeLine('  ' + chalk.green(message.id) + ' ' + message.message);
+        }
       }
-    }
+    }.bind(this));
```
rwjblue: Perhaps `() => {}`?

package.json
```diff
@@ -37,7 +37,6 @@
   },
   "dependencies": {
     "amd-name-resolver": "0.0.6",
-    "bower": "^1.3.12",
     "bower-config": "^1.3.0",
```
rwjblue: Should do the same treatment for the other bower packages (bower-config, bower-endpoint-parser)?

lib/tasks/bower-install.js
```diff
@@ -3,68 +3,97 @@
...
+    });
+  },
+
+  installBower: function() {
+    var ui = this.ui;
+    var chalk = require('chalk');
+
+    ui.startProgress(chalk.green('NPM: Installing bower ...'));
+
+    return Promise.cast(execa('npm', ['install', 'bower@^1.3.12'], {
```
rwjblue: > Promise.cast is dead, don't use it

- @stefanpenner in https://github.com/emberjs/ember.js/issues/13984

Turbo87: https://github.com/ember-cli/ember-cli/blob/9b3387123e84d0c05775490f1903e7c6f11f7b38/lib/utilities/npm.js#L21 ... ups 😉

lib/tasks/bower-install.js
Turbo87: yeah, I'll use that in the future. wanted to keep this compatible with the other stuff for now.

package.json
Turbo87: I thought about that, and decided to tackle the biggest issue/dependency (`bower`) first. In the future we can and should make those optional too.

Turbo87:
> I'm also curious how this interacts with blueprints using addBowerPackageToProject?

I've tested this by running `ember install ember-cli-mocha@0.12.0` in a fresh project with a linked CLI and it works quite well from what I can tell so far.

> I'm not sure doing nothing (if bower.json doesn't exist) is right?

That is the same logic that we've used before. I didn't change that part of the code, but I agree that we should probably have a look at that part too. Seems unrelated to this change though.

rwjblue:
> That is the same logic that we've used before. I didn't change that part of the code, but I agree that we should probably have a look at that part too. Seems unrelated to this change though.

Can you make an issue for that (so we don't loose track)? I suspect if we can get this right then we can remove bower.json from the default app/addon blueprint...

lib/tasks/bower-install.js
rwjblue: Haha, well we can fix this one usage here (since we have to change it anyways) and then go on a hunt...

Turbo87: changed `Promise.cast()` to `Promise.resolve()` and apparently it is still working well 😀

lib/tasks/bower-install.js
```diff
@@ -3,68 +3,97 @@
 // Runs `bower install` in cwd
 var path       = require('path');
 var existsSync = require('exists-sync');
+var execa = require('execa');
 var Promise    = require('../ext/promise');
 var Task       = require('../models/task');
 var formatPackageList = require('../utilities/format-package-list');
+var resolve = Promise.denodeify(require('resolve'));
```
stefanpenner: thoughts on `resolve.sync` ?

package.json
stefanpenner: 👍

lib/tasks/bower-install.js
Turbo87: why use sync when we can use the async version without blocking the progress indicator?

rwjblue: Ya, thats a good enough reason to use async version to me. 😺

rwjblue:
@homu r+

package.json
Glarregle: Coming from the future.. Was wondering why `ember-cli` still has `bower-config` has dependency.

PR #6634, opened 2017-01-03
Remove "ember-cli-app-version" from "addon" blueprint

Turbo87:
[ember-cli-app-version](https://github.com/EmberSherpa/ember-cli-app-version) only seems useful for apps, but I don't see any advantage for addons.


kellyselden:
@homu r+

PR #6639, opened 2017-01-04
Disable usage of `var`.

rwjblue:
* Prefer `const` for `require`'s
* Prefer `let` for everything else.


blueprints/addon/index.js
```diff
@@ -1,21 +1,23 @@
...
+
+const fs = require('fs-extra');
+const existsSync = require('exists-sync');
+const path = require('path');
+const walkSync = require('walk-sync');
+const stringUtil = require('ember-cli-string-utils');
+const uniq = require('ember-cli-lodash-subset').uniq;
+const SilentError = require('silent-error');
+const sortPackageJson = require('sort-package-json');
+let date = new Date();
```
stefanpenner: i would expect this also to be `const` does the linting prevent this?

rwjblue: @stefanpenner - The linting rules absolutely still allow us to use `const` there, but I used the basic heuristic of "if its a `require` then make it `const`" to automate these changes and this one just ended up being a `let`.

Once we land this we should start reviewing "whats left" and fixing things like this one...

stefanpenner: sounds great.

stefanpenner:
We shouldn't let this sit for long, but lets make sure we coordinate with @nathanhammond re: landing this in a way thats aligned with his plan here. (He thought about it thoroughly, don't want to mis some aspect of his plan) If that is already done, :shipit:

rwjblue:
@stefanpenner - Yep, @Turbo87 @nathanhammond and I circled the wagons last night and are on the same page.  We basically want to get all of this "large scale refactoring" done and out of the way before branch point into 2.12's beta cycle.

rwjblue:
@homu r+

Issue #6637, opened 2017-01-04
Access to podPath in Blueprint locals hook


simonihmig:
Use case: want to create a custom blueprint in an addon (component-test in this case), that imports (in the generated file) a file from the tests/ folder. 

**Solution A: relative path**
```js
// generated file foo-bar-test.js from blueprint
import PageObjectComponent from '../../pages/components/foo-bar';
``` 

This works when adding a key for the relative path to the hash returned from my `locals` hook, but only when using the classic file layout. For pods I cannot use something like [this](https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/component-test/index.js#L31-L33) as the options object you get in the `locals` hook does not contain `options.podPath`, and I don't see any other way to get that information.

**Solution B: absolute path**
```js 
// generated file foo-bar-test.js from blueprint
import PageObjectComponent from 'packageName/tests/pages/components/foo-bar';
``` 

This seems to work for apps, as can be seen in other blueprints like [this one](https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/blueprints/mixin-test/mocha-files/tests/unit/mixins/__name__-test.js#L4). But it fails in addons, as `import SomeThing from 'addonpackageName'` refers to the addons `addon` folder, so there is no way I can refer to the addons `tests` folder, get a 'Could not find module `addonpackageName/tests/pages/components/foo-bar`' error.

Any solutions possible for this use case? Would actually expect to have the `podPath` defined in the options, as it is the case for the fileMap tokens.

Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.10.0
http_parser: 2.7.0
node: 6.3.1
v8: 5.0.71.57
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2h
os: darwin x64
3.10.3
```

simonihmig:
Thanks, this was helpful!

Though I think it would be nice to have a bit more straightforward way there, like not hitting the `.config()` API (which is declared as private) and not duplicating the `podPath` extraction logic. Maybe this can be taken into account when the API might be changed anyways some time in the future?

But again, thanks for now!

rwjblue:
We should likely consider making project.config() public API. Would you mind opening a PR changing the YUIDoc designation so we can discuss?

simonihmig:
@rwjblue Sure, done!

PR #6644, opened 2017-01-05
Use  ES6 classes for internal classes

Turbo87:
I'd recommend using `?w=1` for review... (https://github.com/ember-cli/ember-cli/pull/6644/files?w=1)

PR #6645, opened 2017-01-05
Make project.config() public

simonihmig:
There are use cases like the one discussed in #6637 that require access to `project.config()`. Should it be made public therefore?

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -193,7 +193,7 @@ Project.prototype.configPath = function() {
 /**
   Loads the configuration for this project and its addons.
 
-  @private
+  @public
   @method config
   @param  {String} env Environment name
   @return {Object}     Merged confiration object


rwjblue:
IMO, this is already at the very least an intimate API and we should make it public.

rwjblue:
> Wouldn't that also imply that the returned configuration object should be "public"? In the sense that its structure should be documented and be part of the semver "contract", i.e. stick to this structure and not introduce BC (unless new major version of course).

No, I don't think so.  I think we should be able to specific specific keys in that object that are definitely public from ember-cli's perspective (e.g. `rootURL`), but we definitely do not need to make the entire config structure public.  The vast majority of things that end up in config (from my experience) are used and controlled by the app and are not generated/used by ember-cli.

simonihmig:
@rwjblue Yes sure. Probably was not very precise, but meant basically the same: document the config keys that are used by ember-cli and that everybody using that API should be able to rely on.

Turbo87:
this is essentially what comes out of `config/environment.js`, right?

rwjblue:
> this is essentially what comes out of config/environment.js, right?

confirm

Turbo87:
> ☔️ The latest upstream changes (presumably #6644) made this pull request unmergeable. Please resolve the merge conflicts.

whoops...

> > this is essentially what comes out of config/environment.js, right?
> confirm

seems fine to make this method public then, as long as we communicate clearly that the structure itself is not public/stable

nathanhammond:
Calling this good with three 🚢. @homu r+

Issue #6623, opened 2016-12-31
Instrumentation Feature


rwjblue:
This issue tracks the implementation work needed for https://github.com/ember-cli/rfcs/pull/91. That work is happening behind an experiment flag so that we can experiment and test while iterating on the feature.  

TODO:

* Added from https://github.com/ember-cli/ember-cli/pull/6606:
  - [x] change `BUILD_INSTRUMENTATION` from fallback to hard error
  - [x] test to insure against leak of the heimdall builder nodes when running `ember serve` 
  - [x] fix the way `SIGINT` works to ensure that `command` and `shutdown` instrumentation hooks are invoked for `ember serve` / `ember test --server`.
  - [x] clean up command instrumentation timing https://github.com/ember-cli/ember-cli/pull/6628
  - [x] clean up build instrumentation  timing

hjdivad:
Fixing `SIGINT` ought to depend on https://github.com/ember-cli/ember-cli/pull/6615

rwjblue:
I believe this is taken care of at this point.

@hjdivad - Any tasks you see left for implementation here?

PR #6646, opened 2017-01-05
Fix some of the issues in #6623

hjdivad:
- test that we don't leak heimdall nodes
- always try to invoke `command`, `shutdown` instrumentation events, although
  for the moment we're still waiting on #6615
- Add platform to instrumentation summaries
- change BUILD_INSTRUMENTATION from fallback to error


rwjblue:
@homu r+

PR #6647, opened 2017-01-05
Convert CoreObject classes to ES6 classes extending CoreObject

Turbo87:


lib/models/addon-discovery.js
```diff
@@ -21,11 +20,10 @@ const heimdall = require('heimdalljs');
   @extends CoreObject
   @constructor
 */
-let AddonDiscovery = CoreObject.extend({
-  init(ui) {
-    this._super.apply(this, arguments);
+class AddonDiscovery {
```
nathanhammond: Don't we need to extend `CoreObject` here? Or are we calling this internal-only?

lib/models/addons-factory.js
```diff
@@ -16,13 +15,11 @@ const heimdall = require('heimdalljs');
   @extends CoreObject
   @constructor
 */
-let AddonsFactory = CoreObject.extend({
-  init(addonParent, project) {
-    this._super();
-
+class AddonsFactory {
```
nathanhammond: (Also here.)

lib/models/addon-discovery.js
Turbo87: > are we calling this internal-only?

👍

stefanpenner:
yay nice things!

PR #6649, opened 2017-01-06
Make in-repo-addon blueprint 'use strict'.

nathanhammond:
Matches [other](https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/index.js#L2) places.


nathanhammond:
@homu r+

PR #6651, opened 2017-01-07
remove travis sudo check

kellyselden:
I ran a travis build without this sudo check, and it still uses the new container system, so I think this override is no longer necessary as it's the default now.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -3,8 +3,6 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
-
 cache:
   directories:
     - $HOME/.npm
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -3,8 +3,6 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
-
 cache:
   directories:
     - $HOME/.npm


Turbo87:
@kellyselden I think that depends on when the project was created on Travis. Removing this might troll older projects.

maybe @backspace can confirm?

kellyselden:
@Turbo87 My test was using a new project to travis, so you may be on to something.

backspace:
It’s true, as shown [in this table](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), `sudo: false` is only true by default for repositories enabled on Travis in 2015 or later.

Turbo87:
@kellyselden @stefanpenner I personally would be fine keeping sudo our of the blueprints, but I'm pretty sure some of our users would get trolled by this. I'd propose to keep it in there for now and revert this PR.

sorry...! 😞

kellyselden:
@Turbo87 I agree. I'm thinking we should put a comment above the line saying "Feel free to remove this line if your project is newer than 2015." @ember-cli/core Thoughts?

Turbo87:
@kellyselden I'd just keep it the way it is. Those people that care probably already know that they can remove it. Having it in there explicitly certainly doesn't hurt anyone.

kellyselden:
> Those people that care probably already know that they can remove it.

I didn't... 😢

PR #6660, opened 2017-01-10
Preserve user errors in instrumentation hook

hjdivad:
When we run `stopAndReport('init')` we invoke user hooks.  These might
error.  If they do the instrumentation start/stops will be unbalanced
and we will report the following unhelpful error instead of the actual
error in the user hook.

`Error: Cannot stop instrumentation "command".  It has not started.`

This commit fixes this issue.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -143,9 +143,11 @@ class CLI {
       }
 
       let instrumentation = this.instrumentation;
+      let initCompleted = false;
 
       return Promise.resolve().then(() => {
         instrumentation.stopAndReport('init');
+        initCompleted = true;
         instrumentation.start('command');
 
         loggerTesting.info('cli: command.beforeRun');
@@ -156,9 +158,10 @@ class CLI {
         loggerTesting.info('cli: command.validateAndRun');
 
         return command.validateAndRun(commandArgs);
-
       }).finally(() => {
-        instrumentation.stopAndReport('command', commandName, commandArgs);
+        if (initCompleted) {
+          instrumentation.stopAndReport('command', commandName, commandArgs);
+        }
         instrumentation.start('shutdown');
 
         shutdownOnExit = function() {
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -215,6 +215,9 @@ class Builder extends CoreObject {
       .then(result => {
         self.project._instrumentation.stopAndReport('build', result, resultAnnotation);
         return result;
+      }, reason => {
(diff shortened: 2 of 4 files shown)


hjdivad:
@nathanhammond not quite sure when we're branching for beta, but we probably want to land this before

hjdivad:
@stefanpenner hmm tests pass but looks like there might still be a bug; will have to investigate before merging

lib/models/builder.js
```diff
@@ -215,6 +215,9 @@ class Builder extends CoreObject {
       .then(result => {
         self.project._instrumentation.stopAndReport('build', result, resultAnnotation);
         return result;
+      }, reason => {
+        self.project._instrumentation.stopAndReport('build', null, resultAnnotation);
```
hjdivad: @stefanpenner i added this as well as `_buildSummary` tolerance for lack of result to ensure we don't mask an error here.

I was getting a local error b/c I had symlinked `ember-cli` and had an out of date `broccoli-babel-transpiler` (5.6.0 instead of 5.6.2).  In any case, seems good to not mask errors.

hjdivad:
@stefanpenner ok there was not actually a bug.  There was simply another spot where we would mask the root cause error.  The reason I was getting an error at all was simply that I had an out of date `broccoli-babel-transpiler` which yarn wasn't picking up b/c it came from a symlinked `ember-cli`.

hjdivad:
This seems good to merge from my POV, pending tests & reviews

lib/cli/cli.js
```diff
@@ -156,9 +158,10 @@ class CLI {
         loggerTesting.info('cli: command.validateAndRun');
 
         return command.validateAndRun(commandArgs);
-
       }).finally(() => {
-        instrumentation.stopAndReport('command', commandName, commandArgs);
+        if (initCompleted) {
+          instrumentation.stopAndReport('command', commandName, commandArgs);
+        }
```
rwjblue: Shouldn't we have an `else` here to `instrumentation.stopAndReport('init')`?

hjdivad: The only code wrapped inside a promise handler prior to the flag being set is

```js
        instrumentation.stopAndReport('init');
```

So the only way the flag is going to be unset is because we error'd when trying to invoke `stopAndReport('init')` (eg because the addon instrumentation hook errored).  What's the win in retrying `stopAndReport('init')` here?  Is there a case I've forgotten?

rwjblue: I see.  Yes, this makes sense thank you for explaining.

rwjblue:
@homu r+

PR #6661, opened 2017-01-11
Add new `search` command

thoov:
Adds a new command called `search`. It currently supports 3 search modes: addon discovery, code discovery, and code exploration which are explained below.

### Addon Discovery

The addon discovery mode allows you to search for addons based on a given keyword. This is a great way to discovery addons within the community:

```
ember search --addon 'i18n'
```

### Code Discovery

The code discovery mode allows you to search all addons that contain a given code snippet. This is great to find other addons that are using a method or hook that you would like to learn more about. As an example, hooks within your index.js file such as: `included`, `treeForApp`, `treeForPublic`, etc. are not well documented. Here we can search to find other addons which are using one of these hooks:

```
ember search --code 'treeForPublic'
```

### Code Exploration

The last mode is code exploration. This allows you to see actual code diffs based on that code snippet. Note that we must supply the exact addon name via --addon.

```
ember search --addon 'ember-intl' --code 'treeForPublic'
```

Slightly unrelated note: It would be interesting to explore (if the cli-core team hasn't already) a way to install global commands like this one via addons without having to add code to ember-cli's code base.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -347,6 +347,7 @@ class Project {
     let internalMiddlewarePath = path.join(__dirname, '../tasks/server/middleware');
     let legacyBlueprintsPath = require.resolve('ember-cli-legacy-blueprints');
     let emberTryPath = require.resolve('ember-try');
+    let emberSearchPath = require.resolve('ember-cli-search');
     return [
       path.join(internalMiddlewarePath, 'testem-url-rewriter'),
       path.join(internalMiddlewarePath, 'tests-server'),
@@ -356,6 +357,7 @@ class Project {
       path.join(internalMiddlewarePath, 'proxy-server'),
       path.dirname(legacyBlueprintsPath),
       path.dirname(emberTryPath),
+      path.dirname(emberSearchPath),
     ];
   }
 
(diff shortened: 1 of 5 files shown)


Cryrivers:
nice to see such a feature integrated to Ember CLI! 👍

rwjblue:
Might be nice to pull from emberobserver.com's API instead of npm directly, at which point we could provide ratings, categories, etc...

thoov:
@rwjblue Sounds good. I will use emberobserver's and bring in ratings + categories

thoov:
Using emberobservers code search allow us to do this:

kapture 2017-01-11 at 15 56 46

kategengler:
Just for awareness, Ember Observer's API is about to change to JSONAPI. I plan to support both APIs for a short time but the old version has some serious performance issues so it needs to go.

thoov:
@kategengler Do you have a sample JSON output ready yet or a dev environment with the new response by chance?

rondale-sc:
@thoov Would it be possible to include a link to the github repo with the line number when doing code search so that it would be command clickable right from the terminal?

thoov:
@rondale-sc Good idea. I will add that

thoov:
@rwjblue @stefanpenner r? Also Im not sure why the build is failing. It looks to be pulling an old commit that I rebased over. Not sure if one of you could clear the cache or something.

les2:
this is awesome

PR #6663, opened 2017-01-11
Add stats and logging for addon tree caching opt out

trentmwillis:
This will make it easier to identify addons that are opting out of tree caching. This won't catch addons that define their own `cacheKeyForTree` method, but we should assume that means those addons are handling caching on their own.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -11,6 +11,7 @@ const SilentError = require('silent-error');
 const heimdallLogger = require('heimdalljs-logger');
 let logger = heimdallLogger('ember-cli:addon');
 let treeCacheLogger = heimdallLogger('ember-cli:addon:tree-cache');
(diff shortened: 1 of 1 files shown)


lib/models/addon.js
```diff
@@ -41,6 +42,13 @@ heimdall.registerMonitor('addon-tree-cache', function AddonTreeCacheSchema() {
   this.adds = 0;
 });
 
+heimdall.registerMonitor('cache-key-for-tree', function CacheKeyForTreeSchema() {
+  this.optOut = {
+    modifiedMethods: 0,
+    treeForMethodsOverride: 0,
+  };
```
hjdivad: Can we change this schema to just have simple numeric properties (counts) rather than a pojo?

Aside from anything else it will make upgrading to v0.3 heimdall more straightforward

trentmwillis: Was just trying to make things descriptive; I'll flatten it out.

rwjblue:
@homu r+

PR #6665, opened 2017-01-11
Add stableCacheKeyForTree utility

trentmwillis:
This utility makes it easy for addon developers to opt back into addon
tree caching if they need to do simple overrides of some hooks.

Will be used liked:

```js
const cacheKeyForTree = require('ember-cli/lib/utilities/stable-cache-key-for-tree');

module.exports = {
  name: 'some-addon',
  treeForApp() { return someStableTree; },
  cacheKeyForTree
};
```

@rwjblue: if this seems good, should likely be added to the [RFC](https://github.com/ember-cli/rfcs/pull/90).


trentmwillis:
Closing in favor of moving this to the calculate-cache-key-for-tree package.

PR #6669, opened 2017-01-12
tasks/bower-install: Fix "bower" lookup

### LINKED ISSUE #6666: Lazy `bower` installation has issues.
Works:

```sh
git clone git@github.com:ember-engines/engine-testing.git
cd engine-testing
npm install --save-dev ember-cli/ember-cli#68ec21d9ee662a5532b043b3d71f34232c73192c
git clean -fdx
npm install && bower install

# Say yes to all the things.
ember init --skip-npm --skip-bower

npm install --save-dev ember-cli/ember-cli#68ec21d9ee662a5532b043b3d71f34232c73192c
git clean -fdx
npm install && bower install

# Say no to `package.json` (which should be the only one).
ember init
```

Doesn't work: replace the hash (the commit before #6631 landed) with `canary`.

We'll need to fix forward or revert depending on how much work it is. (AKA, want it resolved before release branch point which is "literally any minute.")

Turbo87:
tl;dr we are lazily installing `bower` into the `ember-cli` package folder to avoid conflicts with other bower dependencies, and our `resolve()` call was not looking in the right base directory.

Resolves #6666

### DIFF
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -9,6 +9,8 @@ const Task = require('../models/task');
 const formatPackageList = require('../utilities/format-package-list');
 let resolve = Promise.denodeify(require('resolve'));
 
+const cliPath = path.resolve(`${__dirname}/../..`);
+
 class BowerInstallTask extends Task {
 
   ensureBower() {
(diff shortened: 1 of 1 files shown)


rwjblue:
Does the fact thst our tests passed before this mean we just don't have coverage of this task at all? Is there a way to test this (in a follow up PR)?

rwjblue:
@homu r+

nathanhammond:
> Does the fact thst our tests passed before this mean we just don't have coverage of this task at all?

I don't know how easy it is to test something like this without doing the full reproduction steps I went through in #6666. That is relatively straightforward to accomplish with future `AppFixture` work...

PR #6671, opened 2017-01-12
add description to build environment option

kellyselden:
I was trying to make a build `ember b` to run tests on later `ember t --path dist`, but an `app.import` under `EmberApp.env() === 'test'` wasn't getting included.

Turns out `ember b -e test` works for this scenario, but the help text doesn't make it clear.

I will fix tests if others like this addition.

### DIFF
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -9,7 +9,7 @@ module.exports = Command.extend({
   aliases: ['b'],
 
   availableOptions: [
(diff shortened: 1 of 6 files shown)


rwjblue:
Seems good to me

kellyselden:
rebased and updated tests

kellyselden:
@homu r+

PR #6677, opened 2017-01-14
Move to outputFileSync

### LINKED ISSUE #6672: insertIntoFile isn’t working for ember-cli-mirage
Hello! Much appreciation from me for Ember CLI, as always.

I noticed when starting a new Ember CLI application recently that `ember install ember-cli-mirage` failed to insert the relevant lines in `.jshintrc` and `tests/.jshintrc`. You can see an issue about this [here](https://github.com/samselikoff/ember-cli-mirage/issues/953). It was trivial for me to fix because I’m familiar with a lot of what’s happening in Ember CLI applications, but it’s concerning to me because the guides have people installing Mirage, and this unexpected error could be confusing for newcomers.

In [this Travis log](https://travis-ci.org/backspace/insertIntoFile#L1278), you can see that after `ember install ember-cli-mirage`, the JSHint files haven’t changed.

As [pointed out by @imhuytq](https://github.com/samselikoff/ember-cli-mirage/issues/953#issuecomment-272071503), the underlying issue looks like it’s with `fs.exists`. I added some `console.log` statements and things looked as expected, but on [this line](https://github.com/jprichardson/node-fs-extra/blob/1.x/lib/output/index.js#L12), the callback is never getting called, not even with a false value for `itDoes`.

I gather that JSHint is going away, but this also affects [this use of `insertIntoFile`](https://github.com/samselikoff/ember-cli-mirage/blob/24f994f4419f1b175c82331b00c3089e3f65c038/blueprints/ember-cli-mirage/index.js#L41-L44).

---

```
ember-cli: 2.10.0
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
3.10.8
```

nathanhammond:
I'm not sure what's going wrong here but this fixes the problem. Closes #6672.

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -5,8 +5,6 @@ const existsSync = require('exists-sync');
 const EOL = require('os').EOL;
 const Promise = require('../ext/promise');
 
-let writeFile = Promise.denodeify(fs.outputFile);
-
 /**
   Inserts the given content into a file. If the `contentsToInsert` string is already
   present in the current contents, the file will not be changed unless `force` option
@@ -95,12 +93,10 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
   if (contentsToWrite !== originalContents) {
     returnValue.inserted = true;
 
-    return writeFile(fullPath, contentsToWrite)
-      .then(() => returnValue);
-
-  } else {
-    return Promise.resolve(returnValue);
+    fs.outputFileSync(fullPath, contentsToWrite);
   }
+
+  return Promise.resolve(returnValue);
 }
 
 module.exports = insertIntoFile;


backspace:
Excellent, thanks for looking into this!

Turbo87:
I'd rather not merge fixes for problems we don't fully understand. If this is an issue in fs-extra, we should fix there.

nathanhammond:
1. We have a completely broken public API feature, `insertIntoFile`. Which has been broken for much of the 2.X series per @backspace's review.
2. My bet is that it's an issue in `ext/promise` (which we wrapped it in) and we're also trying to kill (this removes another use of it).
3. [Given how widespread the use of `fs-extra` is](https://www.npmjs.com/package/fs-extra) it seems _incredibly_ unlikely that this is an upstream issue.

I don't like making our users pay the price for a research project. If we intend to block landing this based upon that research project I'd like one of you to own that effort, @Turbo87 / @stefanpenner.

...otherwise let's land this.

nathanhammond:
Note as well that at _some_ point presumably this worked and the last change to the file referenced by @backspace was being moved inside of `fs-extra` on Jun 24, 2015. And here is the history before then: https://github.com/jprichardson/node-fs-extra/commits/bef28cc11865e83da022616878da2971350b216a/lib/output.js

stefanpenner:
> We have a completely broken public API feature, insertIntoFile which has been broken for much of the 2.X series per @backspace's review

Interesting, sounds like there is a missing test. Please be sure to add such a test to this PR.

> I don't like making our users pay the price for a research project. If we intend to block landing this based upon that research project I'd like one of you to own that effort, @Turbo87 / @stefanpenner.

Root cause analysis here seems perfectly reasonable. Wouldn't want our users being affected by this issue in other ways (ways without a nice reproduction as this one).

nathanhammond:
Tests: https://github.com/ember-cli/ember-cli/blob/master/tests/unit/utilities/insert-into-file-test.js

Nothing appears to be missing.

stefanpenner:
@nathanhammond if this is a bugfix, shouldn't it have a regression test?

stefanpenner:
Just took a look, there does not appear to be an issue in ember-cli rather in mirage (and a slightly misleading example). More details -> https://github.com/ember-cli/ember-cli/issues/6672#issuecomment-272754538

nathanhammond:
There's literally nothing in the way this is constructed which implies bug in anything reasonable.

Pure Node.js REPL:
```
ananke:foo nathanhammond$ pwd
/Users/nathanhammond/tmp/foo
ananke-761:foo nathanhammond$ node
> var fs = require('fs');
undefined
> fs.exists('/Users/nathanhammond/tmp/foo', function() { console.log('here', arguments[0]); });
undefined
> here true
```

Modifications made to `fs-extra`:
```diff
- var fs = require('graceful-fs')
+ var fs = require('fs')
```

Node.js debugging:
```
ananke:foo nathanhammond$ node debug ./node_modules/.bin/ember g ember-cli-mirage
< Debugger listening on port 5858
connecting to 127.0.0.1:5858 ... ok
break in /Users/nathanhammond/Repos/ember-cli/bin/ember:2
  1 
> 2 'use strict';
  3 
  4 // Provide a title to the process in `ps`
c
< installing ember-cli-mirage
<   identical /mirage/config.js
<   identical /mirage/scenarios/default.js
<   identical /mirage/serializers/application.js
break in /Users/nathanhammond/Repos/ember-cli/node_modules/fs-extra/lib/output/index.js:12
 10 
 11   var dir = path.dirname(file)
>12   debugger;
 13   fs.exists(dir, function (itDoes) {
 14     if (itDoes) return fs.writeFile(file, data, encoding, callback)
s
break in /Users/nathanhammond/Repos/ember-cli/node_modules/fs-extra/lib/output/index.js:13
 11   var dir = path.dirname(file)
 12   debugger;
>13   fs.exists(dir, function (itDoes) {
 14     if (itDoes) return fs.writeFile(file, data, encoding, callback)
 15 
repl
Press Ctrl + C to leave debug repl
> fs.exists('/Users/nathanhammond/tmp/foo', function() { console.log('here', arguments[0]); });
> fs.existsSync('/Users/nathanhammond/tmp/foo');
> 
```

stefanpenner:
@nathanhammond I believe this is as described in https://github.com/ember-cli/ember-cli/issues/6672#issuecomment-272754538. 

For example:

```js
const fs = require('fs');
const p = Promise.resolve();

p.then(function a() {
  fs.exists('/', function c() {
    console.log('if an exist is scheduled, but no-one is left to hear it, does it happen?');
  });
   // in your example, you would be debugging in this context
  console.log('does / exist?', require('fs').existsSync('/'));
});

p.then(function b() { process.exit(); });
```

output:

```sh
does / exist?: true
```

---

The order of stuff ends up being:

1. `p` followed by `a`
2. `p` followed by `b`
3. invoke `a`
4. schedule exists '/' with `c`
5. invoke `existsSync('/'(` => `true` (which returns synchronously)
6. invoke `b`
7. `process.exit`

Step 8 will never happen, as task `c` will arrive to late.

stefanpenner:
let me try the above with a debugger to make sure

PR #6678, opened 2017-01-14
Avoid error upon registering a heimdall monitor twice.

rwjblue:
`heimdall` throws an error if the same monitor name is registered twice. In normal operations with npm@3 / yarn this is not an issue (because only one version of this file is required). However, when working in a libary where you are `npm link ember-cli` it is possible to actually end up with this file being evaluated twice (once from the `npm link` and another from the source location).

This change simply checks if a monitor has already been registered with a given name, and does nothing.


rwjblue:
@homu r+

PR #6680, opened 2017-01-14
Use global NPM with version check

Turbo87:
... and remove our `npm` dependency.

see https://github.com/ember-cli/rfcs/pull/88


twokul:
this is nice

lib/tasks/npm-task.js
```diff
@@ -13,47 +20,86 @@ class NpmTask extends Task {
     // The command to run: can be 'install' or 'uninstall'
     this.command = '';
 
-    this.npm = this.npm || require('../utilities/npm');
+    this.versionConstraints = '3 || 4';
+  }
+
+  npm(args) {
+    logger.info('npm: %j', args);
```
twokul: curious why `%j`? isn't `args` an array?

lib/tasks/npm-task.js
```diff
@@ -13,47 +20,86 @@ class NpmTask extends Task {
     // The command to run: can be 'install' or 'uninstall'
     this.command = '';
 
-    this.npm = this.npm || require('../utilities/npm');
+    this.versionConstraints = '3 || 4';
```
twokul: would it make sense to turn `versionConstraints` into a `const` on the top of the file or there was some reasoning behind having it as an instance property?

lib/tasks/npm-task.js
```diff
@@ -13,47 +20,86 @@ class NpmTask extends Task {
...
+      throw error;
+    });
   }
 
   run(options) {
-    let ui = this.ui;
-    let startMessage = this.formatStartMessage(options.packages);
-    let completeMessage = this.formatCompleteMessage(options.packages);
+    return this.checkVersion().then(() => {
+
```
twokul: extra line, intentional?

kategengler:
Assuming I was cc'd for `ember-try`, amusingly, it already relies on there being a global `npm` https://github.com/ember-cli/ember-try/blob/master/lib/dependency-manager-adapters/npm.js#L82

rwjblue:
@kategengler - Ya, likely wanted to ensure communicate that we are going to remove just in case it was setup like bower was before.

kategengler:
@rwjblue Yep, just confirming that ember-try should be fine with this change.

lib/tasks/npm-task.js
```diff
@@ -13,47 +20,86 @@ class NpmTask extends Task {
...
-    this.npm = this.npm || require('../utilities/npm');
+    this.versionConstraints = '3 || 4';
+  }
+
+  npm(args) {
+    logger.info('npm: %j', args);
+    return RSVP.resolve(execa('npm', args, { preferLocal: false }));
+  }
+
+  checkVersion() {
```
stefanpenner: is this function unit tested?

lib/tasks/npm-task.js
Turbo87: It would be possible, but it seemed more flexible this way.

lib/tasks/npm-task.js
Turbo87: args is an array, yes. %j uses JSON formatting, while the default toString of an array can hide information

lib/tasks/npm-task.js
Turbo87: Yeah, wanted to give the content a little more breathing room

lib/tasks/npm-task.js
Turbo87: not yet, but I guess we could change that

stefanpenner: with that, I am 👍  with this PR.

Turbo87: @stefanpenner unit tests for `checkVersion()` are added now

stefanpenner:
@homu r+

PR #6681, opened 2017-01-14
Run "ember-cli-eslint" blueprint on "ember init"

### LINKED ISSUE #6643: ember-cli-eslint doesn't appear in the default blueprint.
Given more time to think about it, I'm not 100% happy with how `ember-cli-eslint` isn't in the default blueprint anymore. Many users don't use `ember init` to handle upgrades which means that they end up "stuck" on whichever version they have installed, or worse, delete it without paying attention as to why [when they view the comparison](https://github.com/ember-cli/ember-cli/blob/master/dev/add-to-output-repos.sh#L35).

This whole setup exists for one reason: we need to be able to run all default blueprints for bundled addons once per application creation. We may possibly in the future want to run the default blueprint for every addon in the application each time that addon gets updated. The attachment to `init` is because we can't run any blueprints unless we actually have a `node_modules` folder...

This complicates [programmatic fixture creation](https://github.com/ember-cli/ember-cli/pull/6449) as well.

I don't have a solution yet just a set of constraints and a bit of uncertainty that we've found the global maxima.

/cc @Turbo87

Turbo87:
... instead of running "npm install" again

Resolves #6643 (at least partly)

### DIFF
--- a/tests/acceptance/addon-smoke-test-slow.js
+++ b/tests/acceptance/addon-smoke-test-slow.js
@@ -30,6 +30,7 @@ describe('Acceptance: addon-smoke-test', function() {
   before(function() {
     return createTestTargets(addonName, {
       command: 'addon',
+      createESLintConfig: true,
     });
   });
 
--- a/tests/acceptance/brocfile-smoke-test-slow.js
+++ b/tests/acceptance/brocfile-smoke-test-slow.js
@@ -25,7 +25,7 @@ describe('Acceptance: brocfile-smoke-test', function() {
   this.timeout(500000);
 
   before(function() {
-    return createTestTargets(appName);
+    return createTestTargets(appName, { createESLintConfig: true });
   });
 
   after(teardownTestTargets);
--- a/tests/helpers/acceptance.js
+++ b/tests/helpers/acceptance.js
@@ -3,12 +3,15 @@
 const symlinkOrCopySync = require('symlink-or-copy').sync;
 const path = require('path');
 const fs = require('fs-extra');
+const RSVP = require('rsvp');
 const runCommand = require('./run-command');
 
 let root = path.resolve(__dirname, '..', '..');
 
 const PackageCache = require('../../tests/helpers/package-cache');
 const CommandGenerator = require('../../tests/helpers/command-generator');
+
+const outputFile = RSVP.denodeify(fs.outputFile);
 
 const quickTemp = require('quick-temp');
 let dirs = {};
@@ -55,6 +58,18 @@ function createTestTargets(projectName, options) {
   options.command = options.command || 'new';
 
   return applyCommand(options.command, projectName, '--skip-npm', '--skip-bower', `--directory=${outputDir}`)
+    .then(() => {
+      if (options.createESLintConfig) {
+        let eslintConfig = `module.exports = {
+          root: true, 
+          parserOptions: { 
+            sourceType: 'module'
+          },
+        };`;
+
+        return outputFile(`${outputDir}/.eslintrc.js`, eslintConfig);
+      }
+    })
     .catch(handleResult);
 }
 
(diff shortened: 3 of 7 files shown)


Turbo87:
Looks like I need to fix a few tests...

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -30,6 +30,7 @@ describe('Acceptance: addon-smoke-test', function() {
   before(function() {
     return createTestTargets(addonName, {
       command: 'addon',
+      createESLintConfig: true,
```
twokul: trailing comma?

Turbo87: yes

https://github.com/ember-cli/ember-cli/blob/master/.eslintrc.js#L60

tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -25,7 +25,7 @@ describe('Acceptance: brocfile-smoke-test', function() {
   this.timeout(500000);
 
   before(function() {
-    return createTestTargets(appName);
+    return createTestTargets(appName, { createESLintConfig: true });
```
nathanhammond: Can we name this option differently? I suspect that we will eventually want this to be `runDefaultBlueprints` or something like that.

tests/helpers/acceptance.js
```diff
@@ -55,6 +58,18 @@ function createTestTargets(projectName, options) {
   options.command = options.command || 'new';
 
   return applyCommand(options.command, projectName, '--skip-npm', '--skip-bower', `--directory=${outputDir}`)
+    .then(() => {
+      if (options.createESLintConfig) {
```
nathanhammond: Everything else is deferring to the upstream Ember command, this is `mock`ing it. I'd _much_ prefer (given this is an acceptance test) to actually run the generator.

tests/acceptance/brocfile-smoke-test-slow.js
Turbo87: this option does not actually run the blueprint, it just generates a very basic `.eslintrc.js` file

tests/helpers/acceptance.js
Turbo87: I'd prefer that too, but I'm not sure how to do it as long as we use `--skip-npm` here :-/

nathanhammond: We use `--skip-npm` on the target creation but we link in a unified PackageCache: https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/brocfile-smoke-test-slow.js#L34

... the `npm` assets will be there.

nathanhammond: Hrm. We just need to do this _after_ `linkDependencies` meanwhile you were doing it at the `createTestTargets` point. I ... stand corrected. I'd like to make it actually run the blueprints, there's no particular reason that `linkDependencies` needs to be run `beforeEach` that I can tell in a quick glance while correcting my previous comment...

kellyselden:
I might be a little out of the loop, but what do I lose if I always run `ember new my-app -sb && cd my-app && npm link ember-cli && npm i`?

Turbo87:
@kellyselden that would be another logical step. I think I'd still prefer this PR to be merged so that we don't have to invoke npm twice.

kellyselden:
> that would be another logical step. I think I'd still prefer this PR to be merged so that we don't have to invoke npm twice.

👍

Turbo87:
@stefanpenner @rwjblue friendly reminder 😉

stefanpenner:
restarting appveyor, looks like it was unable to download the commit/pr in question for node 4. Lets see if a second attempt is successful (seems likely)

Turbo87:
@stefanpenner I had restarted that AppVeyor build multiple time in the past week and always had issues like that. It seems that AppVeyor is just broken right now... I don't see how the code changes could cause that in any way.

PR #6686, opened 2017-01-16
add note about sudo: false

kellyselden:
Result of https://github.com/ember-cli/ember-cli/pull/6651#issuecomment-272555394 and https://github.com/ember-cli/ember-cli/pull/6651#issuecomment-272758859

Basically I think it is a worthwhile comment, might save someone a few hours of headache.

@ember-cli/core Feel free to approve or deny, I won't be offended 😄

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - "6"
 
+# feel free to remove this line if your project is newer than 2015
 sudo: false
 
 cache:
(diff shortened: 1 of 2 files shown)


Turbo87:
A note like that implies something to do (read: work) for everyone that creates a new project, when it actually does work fine without removing that line. For that reason and to avoid unnecessary blueprint churn I'd suggest to keep it as it was. I don't feel strongly about this though so feel free to ignore me.

nathanhammond:
Don't care.

kellyselden:
I'll close for now while I think if this is really more benefit than hassle.

PR #6688, opened 2017-01-16
Replace custom Promise class with RSVP

Turbo87:
- `lib/ext/promise` -> 💀 
- `rsvp` -> 🎉


Turbo87:
@stefanpenner I'm not aware of any addons doing that, do you think that might actually be used somewhere? And actually our API only specifies that we return a `Promise`, not an `RSVP.Promise` or our custom class, so technically addons shouldn't rely on those methods being available.

stefanpenner:
> @stefanpenner I'm not aware of any addons doing that, do you think that might actually be used somewhere? And actually our API only specifies that we return a Promise, not an RSVP.Promise or our custom class, so technically addons shouldn't rely on those methods being available.

Ya, I agree they should not be. I am only mildly worried that they are. If they are, I would love to give them a more useful error.

I'm using https://emberobserver.com/code-search to try and find any usages, so far we may be in the clear...

stefanpenner:
```
tbieniek [7:31 AM] 
they will most likely notice during beta and the fix is relatively easy

stefanpenner [7:31 AM] 
sounds reasonable
```

rwjblue:
Many addons are requiring our promise class: https://emberobserver.com/code-search?codeQuery=ext%2Fpromise

We should put a file there that returns `require('rsvp').Promise` with a deprecation. Basically the same as what we do for silent-error module.

Turbo87:
@rwjblue wasn't sure how to get a deprecation message in there properly, but I've added a doc commented with `@deprecated`.

lib/ext/promise.js
```diff
@@ -1,62 +1,15 @@
...
+
+/**
+ * Using the `Promise` class from `ember-cli/ext/promise` is deprecated
+ * and private API. Please use RSVP.js directly instead.
+ *
+ * @class Promise
+ * @private
+ * @deprecated Please use RSVP.js directly
+ */
+module.exports = require('rsvp').Promise;
```
rwjblue: See [lib/errors/silent.js](https://github.com/ember-cli/ember-cli/blob/master/lib/errors/silent.js) for how to deprecate properly.

rough example:

```js
  get() {
    // deprecate here
    return require('rsvp').Promise;
  }
});
```

Turbo87:
@rwjblue thanks, updated.

lib/ext/promise.js
```diff
@@ -1,62 +1,16 @@
...
+const deprecate = require('../utilities/deprecate');
+
+Object.defineProperty(module, 'exports', {
+  get() {
+    // Get the call stack so we can let the user know what module is using the deprecated function.
+    let stack = new Error().stack;
+    stack = stack.split('\n')[5];
+    stack = stack.replace('    at ', '  ');
+
+    deprecate(`\`ember-cli/lib/ext/promise.js\` is deprecated, use \`rsvp\` instead. Required here: \n${stack.toString()}`, true);
```
stefanpenner: should we do this in 2 steps, e.g. deprecate before removing ^ ? Or are we pretty confident no-one depends on ext/promise functionality (i am dubious)

Turbo87: Most of the usage I've found used it as a standard Promise, but https://emberobserver.com/code-search?codeQuery=denodeify also shows a few instances where `denodeify()` is used. Can we export the RSVP `Promise` and monkey patch `denodeify()` on top?

stefanpenner: Any reason to not leave `ext/promise` for a release or two but with all its methods deprecated?

Turbo87: guess we could do that as well, yeah

stefanpenner: I would also like to 🔥 ext/promise (it is my invention so I can be negative about it) but we may need a deprecation cycle.

Turbo87: @stefanpenner fixed

Turbo87:
@homu r+

Issue #6679, opened 2017-01-14
Undo typing doesn't trigger live-reload


gandalfar:
It seems that undoing typing in text editor doesn't trigger live reload on errors.

Steps to reproduce:

1. `ember s`
2. In the template:  `{{var}}`
3. (live reload happens)
4. Change template to: `{{var}}}` (triple `}}}`)
5. (live reload happens with build error)
6. Change template back to `{{var}}`
_7. live reload fails to happen_ (also app.js isn't written do `dist/assets/app.js`)
8. Now make a bigger change: `{{var}} test`
9. (live reload happens again)

I'm reading through source code of ember-cli, but I'm still tracking where in the pipeline error happens.
---

Output from `ember version --verbose && npm --version`:
```
➜  simplesass git:(master) ✗ ember version --verbose && npm --version
ember-cli: 2.11.0-beta.1-master-13de36657d
http_parser: 2.7.0
node: 6.6.0
v8: 5.1.281.83
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2h
os: darwin x64
3.10.7

```

gandalfar:
Investigating a bit further, it doesn't seem connected to live-reload (I can reproduce it with `ember s --live-reload=false`). So the problem seems that while `ember-cli` console output says:

```
file changed templates/index.hbs

Build successful - 88ms.
```

The build of `dist/assets/app.js` doesn't happen (no changed timestamp on hard drive)

gandalfar:
Looking at this further, it's correct that `dist/assets/app.js` doesn't get touched as it was correctly compiled in a previous step 3 (before build error). So live reload should pick up on it.

gandalfar:
I think the problem is that livereload-server gets empty list of files that changed. So while didChange event triggers, there's no information on what actually changed. Nothing changed because `broccoli-persistent-filter` did a good job and just said that everything is cached.

I think `broccoli-persistent-filter` should invalidate cache key for `triggerError` on the file that triggered it.

gandalfar:
My initial propose for a fix: https://github.com/gandalfar/ember-cli/commit/aa466972619b76bf95f268e3333e776b718b8bfc

gandalfar:
It also seems to fix #6242

PR #6690, opened 2017-01-16
BUGFIX #6679 - workaround for tiny-lr not reloading on empty files arguments

### LINKED ISSUE #6242: Fixing SASS error doesn't reload the page from build error
Recently the ability to have changed styles hot-swapped without a reload was added (which is awesome!!).

I'm using [`ember-cli-sass`](https://github.com/aexmachina/ember-cli-sass), and a build error caused by invalid SASS, however, will cause the page to reload and display an error similar to:

> ### Build error
> 
> **tmp/sass_compiler-input_base_path-SJwgbQCL.tmp/0/app/styles/_some-style.scss, line 96**
> 
> **The Broccoli Plugin: [SassCompiler] failed with:**
> ---
> 
> ```
> Error: Invalid CSS after "}": expected "}", was ""
>     at options.error (node_modules/node-sass/lib/index.js:272:32)
> 
> The broccoli plugin was instantiated at: 
>     at SassCompiler.Plugin (node_modules/broccoli-plugin/index.js:7:31)
>     at SassCompiler.CachingWriter [as constructor] (node_modules/broccoli-sass-source-maps/node_modules/broccoli-caching-writer/index.js:18:10)
>     at new SassCompiler (node_modules/broccoli-sass-source-maps/index.js:20:17)
> ```

Fixing this error triggers a hot-swap reload rather than a full page reload. This unfortunately means you are still on the build error page rather than viewing your application. You have to manually reload the page to get back to your application.

It would be great if instead of requiring a manual reload, fixing such an error caused a full reload.

_(Originally posted as an [`ember-cli-sass` issue](https://github.com/aexmachina/ember-cli-sass/issues/135))_

---

```
$ ember version --verbose && npm --version
ember-cli: 2.7.0
http_parser: 2.7.0
node: 6.3.1
v8: 5.0.71.57
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2h
os: darwin x64
3.10.3

$ npm ls --depth=0 | grep ember-cli-sass
├── ember-cli-sass@5.5.1
```

gandalfar:
In a situation when you revert failed compile attempt in `broccoli-persistent-filter`'s `processString`, 

```
// lib/tasks/server/livereload-server.js

didChange(results) {
...
      this.tree = new FSTree.fromEntries(this.getDirectoryEntries(results.directory), {sortAndExpand: true});
      files = previousTree.calculatePatch(this.tree)
        .filter(isNotRemoved)
        .filter(isNotDirectory)
        .map(relativePath)
        .filter(isNotSourceMapFile);
```
the result of `files` is `[]`.

This gets passed to `this.liveReloadServer().changed({` which doesn't trigger new websocket push, because there's nothing to notify clients about.

This PR works around this by catching the fact that there was a compilation error and forcing reload on next `didChange`.

I'm not sure if it's possible to write test for this as `tiny-lr`'s [`changed`](https://github.com/mklabs/tiny-lr/blob/master/lib/server.js#L220) doesn't return. I've got test written for the Error handling portion of the functionality.

This might be a too complicated approach, but I'm not familiar with whole system enough to see any other options. Because it doesn't different between reasons for errors, it also fixes #6242.

### DIFF
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -137,7 +137,13 @@ class LiveReloadServerTask extends Task {
     let previousTree = this.tree;
     let files;
 
-    if (results.directory) {
+    if (results.stack) {
+      this._hasCompileError = true;
+      files = ['LiveReload due to compile error'];
+    } else if (this._hasCompileError) {
+      this._hasCompileError = false;
+      files = ['LiveReload due to resolved compile error'];
+    } else if (results.directory) {
       this.tree = new FSTree.fromEntries(this.getDirectoryEntries(results.directory), { sortAndExpand: true });
       files = previousTree.calculatePatch(this.tree)
         .filter(isNotRemoved)
--- a/tests/unit/tasks/server/livereload-server-test.js
+++ b/tests/unit/tasks/server/livereload-server-test.js
@@ -214,6 +214,38 @@ describe('livereload-server', function() {
         return watcherEventTest('not-an-event', 0);
       });
 
+      it('recovers from error when file is already cached in previous cache step', function() {
+        let compileError = function() {
+          try {
+            throw new Error('Compile time error');
+          } catch (error) {
+            return error;
+          }
+        }.apply();
+
+        subject.getDirectoryEntries = createStubbedGetDirectoryEntries([
+          'test/fixtures/proxy/file-a.js',
+        ]);
+
+        return subject.start({
+          liveReloadPort: 1337,
+          liveReload: true,
+        }).then(function() {
+          return watcher.emit('error', compileError);
+        }).then(function(result) {
+          expect(result).to.be.true;
+          expect(subject._hasCompileError).to.be.true;
+          expect(changedCount).to.equal(1);
+          return watcher.emit('change', {
+            directory: '/home/user/projects/my-project/tmp/something.tmp',
+          });
+        }).then(function(result) {
+          expect(result).to.be.true;
+          expect(subject._hasCompileError).to.be.false;
+          expect(changedCount).to.equal(2);
+        });
+      });
+
       describe('filter pattern', function() {
         it('shouldTriggerReload must be true if there are no liveReloadFilterPatterns', function() {
           subject.project.liveReloadFilterPatterns = [];


stefanpenner:
So TL;DR in-order to recover post error, if no files have changes we must do a full reload?

(this seems reasonable)

lib/tasks/server/livereload-server.js
```diff
@@ -137,8 +137,15 @@ class LiveReloadServerTask extends Task {
     let previousTree = this.tree;
     let files;
 
+    if ( results.stack ) {
+      this._hasCompileError = true;
+      files = ['LiveReload due to compile error'];
+    } else if ( this._hasCompileError) {
+      this._hasCompileError= false;
```
stefanpenner: missing whitespace between `_hasCompileError` and `=`

lib/tasks/server/livereload-server.js
```diff
@@ -137,8 +137,15 @@ class LiveReloadServerTask extends Task {
     let previousTree = this.tree;
     let files;
 
+    if ( results.stack ) {
```
stefanpenner: extra whitespace between `(` and `results.stack` and `)`;

lib/tasks/server/livereload-server.js
```diff
@@ -137,8 +137,15 @@ class LiveReloadServerTask extends Task {
...
+      this._hasCompileError = true;
+      files = ['LiveReload due to compile error'];
+    } else if ( this._hasCompileError) {
```
stefanpenner: unbalanced whitespace

gandalfar:
@stefanpenner your understanding of the issue is correct

gandalfar:
I've fixed whitespace problems and squashed commits into single one.

stefanpenner:
@gandalfar thanks!

This looks reasonable, but we likely need a unit test covering it. If you need help, let me know. We can likely lend a hand if needed.

gandalfar:
Can you suggest what should be part of the test? I don't think I can capture tiny-lr output. I can assert against error stack trace and this._hasError.

stefanpenner:
likely just ensure https://github.com/ember-cli/ember-cli/pull/6690/files#diff-60cfa3bc4236a25e0793052851917a47R164 gets invoked with the expected value.

gandalfar:
I've added test that checks for `changedCount` and `subject._hasCompileError`. To actually track `files` contents, I'd have to modify 

```
let stubbedChanged = function() {
      changedCount += 1;
    };
```

with custom `changed` method that tracks the value. I can do that if this is better approach (I was just looking at how other tests are written).

gandalfar:
I've squashed commits and made sure it still applies to current master. Is there anything else I can do?

stefanpenner:
@gandalfar this looks great! Sorry for letting this sit.

stefanpenner:
re-running appvevor, will merge when green.

Turbo87:
finally 🍏 !

@homu r+

PR #6692, opened 2017-01-17
tests/acceptance/generate: Fix flaky tests

Turbo87:
We have to wait for the first outputFile() too before we can continue.


rwjblue:
@homu r+

PR #6695, opened 2017-01-17
BUGFIX Scoped package names

andytompkins:
In the function `findDefaultBlueprintName`, the package name passed to `findAddonByName` has been processed by `getPackageBaseName`, which removes the scope (@scope/my-pkg-name becomes my-pkg-name). In `findAddonByName` a comparison is done:
```
function matchAddon(name, addon) {
    return name === addon.name || (addon.pkg && name === addon.pkg.name);
  }
```
In the case where you are using scoped names, addon.pkg.name will be the fully scoped package name and that second comparison can never be true. If your package name in package.json and index.js are the same, you would never notice this. We are transitioning from a naming standard like "prefix-my-package-name" to "@prefix2/my-package-name". To make it easier for users, we decided to update the package name in package.json, but leave the index.js name unchanged (so that users of our library can update their package.json but not have to update import and require statements.) Under ember-cli 2.5.1 this worked fine, but when we went to try an upgrade we found this issue.  The logic in the `matchAddon` function appears to support the notion of the names being different. Processing addon.pkg.name with `getPackageBaseName` alleviates the problem.

### DIFF
--- a/tests/unit/utilities/find-addon-by-name-test.js
+++ b/tests/unit/utilities/find-addon-by-name-test.js
@@ -14,7 +14,15 @@ describe('findAddonByName', function() {
     }, {
       name: 'foo-bar',
       pkg: { name: 'foo-bar' },
+    }, {
+      name: 'foo-bar-baz',
+      pkg: { name: '@foo/bar-baz' },
     }];
+  });
+
+  it('should return the foo-bar-baz addon when the unscoped package name is supplied', function() {
+    let addon = findAddonByName(addons, 'bar-baz');
+    expect(addon.name).to.equal('foo-bar-baz', 'should have found the foo-bar-baz addon');
   });
 
   it('should return the foo addon from name', function() {
--- a/lib/utilities/find-addon-by-name.js
+++ b/lib/utilities/find-addon-by-name.js
@@ -1,10 +1,11 @@
 'use strict';
 
+const getPackageBaseName = require('../utilities/get-package-base-name');
 const find = require('ember-cli-lodash-subset').find;
 
 module.exports = function findAddonByName(addons, name) {
   function matchAddon(name, addon) {
-    return name === addon.name || (addon.pkg && name === addon.pkg.name);
+    return name === addon.name || (addon.pkg && name === getPackageBaseName(addon.pkg.name));
   }
 
   return find(addons, addon => matchAddon(name, addon));


rwjblue:
Seems good to me...

tests/unit/utilities/find-addon-by-name-test.js
```diff
@@ -14,9 +14,17 @@ describe('findAddonByName', function() {
...
       pkg: { name: 'foo-bar' },
+    }, {
+      name: 'foo-bar-baz',
+      pkg: { name: '@foo/bar-baz' },
     }];
   });
 
+  it('should return the foo-bar-baz addon when the unscoped package name is supplied', function() {
+    let addon = findAddonByName(addons, 'bar-baz');
+    expect(addon.name).to.equal('foo-bar-baz', 'should have found the foo-bar-baz addon');
```
stefanpenner: Im curious why `pkg.name = '@foo/bar-baz'` trumps `pkg.name = 'foo-bar'`  here, how would one find `pkg.name = 'foo-bar'` now

andytompkins: I am not entirely sure what you are asking. 
The test cases seem to be missing this scenario, where addon.name and addon.pkg.name are both present but have different values. Our package name in package.json is something like "@scope/my-package-name". In index.js the name is "prefix-my-package-name". If I run `ember install @scope/my-package-name` it cannot find the addon, because the name passed in has had the scope removed, while the list of addons has not, and so in the comparison there is no match. (The addon.name is "prefix-my-package-name" and the addon.pkg.name is "@scope/my-package-name", while the base package name passed in is "my-package-name".)

landongn:
👍 nice work, Mike!

WarttHog:
Right.  npm scopes.  Good catch!

tests/unit/utilities/find-addon-by-name-test.js
stefanpenner: > In index.js the name is "prefix-my-package-name".

The fact that there is a difference here seems bad, as described the following two packages`@foo/bar` and `bar` will collide and cause grief.

rwjblue:
Sorry for letting this linger so long, I believe that we have addressed all of the issues around scoped packages at this point. Sorry that we didn't get this PR landed.

PR #6698, opened 2017-01-18
Remove "bower.json" and only create if necessary

Turbo87:
"if necessary" being if a blueprint specifically wants to install Bower dependencies.

This PR removes the need for Ember CLI to lazily install Bower during `ember new`, since the `bower.json` file is now gone in the `app` and `addon` blueprints.

### DIFF
--- a/blueprints/app/files/bower.json
+++ b/blueprints/app/files/bower.json
@@ -1,5 +1,0 @@
-{
-  "name": "<%= name %>",
-  "dependencies": {
-  }
-}
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1017,7 +1017,7 @@ let Blueprint = CoreObject.extend({
     return task.run({
       verbose: true,
       packages: packageNamesAndVersions,
-      installOptions,
+      installOptions: installOptions || { save: true },
     });
   },
 
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -35,8 +35,13 @@ class BowerInstallTask extends Task {
     }
 
     return this.resolveBower()
-      .catch(() => this.installBower()
-        .then(() => this.resolveBower()))
+      .catch(error => {
+        if (error.message.indexOf('Cannot find module \'bower\'') === -1) {
+          throw error;
+        }
+
+        return this.installBower().then(() => this.resolveBower());
+      })
       .then(bowerPath => this.importBower(bowerPath));
   }
 
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -51,12 +51,19 @@ class BowerInstallTask extends Task {
       .then(() => ui.writeLine(chalk.green('NPM: Installed bower')));
   }
 
+  get bowerJsonPath() {
+    return path.join(this.project.root, 'bower.json');
+  }
+
+  hasBowerJson() {
+    return existsSync(this.bowerJsonPath);
+  }
+
   // Options: Boolean verbose
   run(options) {
-    let bowerJson = path.join(this.project.root, '/bower.json');
     let ui = this.ui;
 
-    if (!existsSync(bowerJson)) {
+    if (!this.hasBowerJson()) {
       ui.writeWarnLine('Skipping bower install: bower.json not found');
       return Promise.resolve();
     }
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -8,6 +8,8 @@ const RSVP = require('rsvp');
 const Task = require('../models/task');
 const formatPackageList = require('../utilities/format-package-list');
 
+const logger = require('heimdalljs-logger')('ember-cli:tasks:bower-install');
+
 const Promise = RSVP.Promise;
 const resolve = RSVP.denodeify(require('resolve'));
 
@@ -16,6 +18,7 @@ const cliPath = path.resolve(`${__dirname}/../..`);
 class BowerInstallTask extends Task {
 
   resolveBower() {
+    logger.info('Resolving "bower" from %s ...', cliPath);
     return resolve('bower', { basedir: cliPath });
   }
 
@@ -27,10 +30,15 @@ class BowerInstallTask extends Task {
     return this.resolveBower()
       .catch(() => this.installBower()
         .then(() => this.resolveBower()))
-      .then(bowerPath => require(bowerPath));
+      .then(bowerPath => {
+        logger.info('Resolved "bower" to: %s', bowerPath);
+        return require(bowerPath);
+      });
   }
 
   installBower() {
+    logger.info('Installing "bower" via NPM into: %s', cliPath);
+
     let ui = this.ui;
     const chalk = require('chalk');
 
@@ -59,6 +67,8 @@ class BowerInstallTask extends Task {
 
       let startMessage = this.formatStartMessage(packages);
       let completeMessage = this.formatCompleteMessage(packages);
+
+      logger.info('Installing %j via Bower with options: %j', packages, installOptions);
 
       ui.startProgress(chalk.green(startMessage));
 
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -1,6 +1,7 @@
 'use strict';
 
 // Runs `bower install` in cwd
+const fs = require('fs-extra');
 const path = require('path');
 const existsSync = require('exists-sync');
 const execa = require('execa');
@@ -12,6 +13,7 @@ const logger = require('heimdalljs-logger')('ember-cli:tasks:bower-install');
 
 const Promise = RSVP.Promise;
 const resolve = RSVP.denodeify(require('resolve'));
+const writeJson = RSVP.denodeify(fs.writeJson);
 
 const cliPath = path.resolve(`${__dirname}/../..`);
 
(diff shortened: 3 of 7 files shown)


lib/tasks/bower-install.js
```diff
@@ -1,32 +1,48 @@
...
+    return this.resolveBower()
       .catch(() => this.installBower()
-        .then(() => resolve('bower', { basedir: cliPath })))
-      .then(bowerPath => require(bowerPath));
+        .then(() => this.resolveBower()))
+      .then(bowerPath => this.importBower(bowerPath));
   }
 
   installBower() {
+    logger.info('Installing "bower" via NPM into: %s', cliPath);
```
twokul: what's `cliPath` here? global or `node_modules/ember-cli`?

lib/tasks/bower-install.js
```diff
@@ -37,32 +53,61 @@ class BowerInstallTask extends Task {
...
 
-    if (!existsSync(bowerJson)) {
-      ui.writeWarnLine('Skipping bower install: bower.json not found');
+    let packages = options.packages || [];
+    let installOptions = options.installOptions || {};
+    let savePackages = installOptions.save || installOptions.saveDev;
+
+    // if we are running "bower install" from "ember init" and there is
+    // no "bower.json" we return early
+    if (!this.hasBowerJson() && !savePackages) {
```
twokul: is `!savePackages` necessary here?

tests/unit/tasks/bower-install-test.js
```diff
@@ -2,31 +2,60 @@
...
+      });
+    });
+
+    it('install "bower" if it can not be resolved', function() {
+      td.when(task.resolveBower()).thenReturn(Promise.reject(), Promise.resolve('path/to/bower'));
+      td.when(task.installBower()).thenResolve();
+      td.when(task.importBower('path/to/bower')).thenReturn('ok');
+
+      return task.ensureBower().then(ok => {
+        expect(ok).to.equal('ok');
```
twokul: 👌 is👌

twokul:
restarting Appveyor

lib/tasks/bower-install.js
Turbo87: it depends ™

if you're in `ember new` it will be the global, if you're in `ember install` it will be the local

lib/tasks/bower-install.js
Turbo87: yes, otherwise `addBowerPackagesToProject()` won't actually add anything to `bower.json` if it doesn't exist yet

tests/unit/tasks/bower-install-test.js
Turbo87: 🤷‍♂️

Turbo87:
I seriously have no idea anymore what's wrong with AppVeyor. The code itself seems to work fine, but CI is :trollface:ing us...

lib/tasks/bower-install.js
```diff
@@ -37,32 +53,61 @@ class BowerInstallTask extends Task {
...
+      logger.info('Skipping "bower install" since "bower.json" does not exist');
       return Promise.resolve();
     }
 
-    return this.ensureBower().then(bower => {
+    // resolve "bower" and install via NPM if necessary
+    // -> then check if a "bower.json" file exists and create if necessary
+    // -> then run the actual "bower install"
+    return this.ensureBower().then(() => this.ensureBowerJson()).then(() => {
```
nathanhammond:
```javascript
RSVP.all([this.ensureBower(), this.ensureBowerJson()])
  .then(...)
```

lib/tasks/bower-install.js
nathanhammond: This somewhat falls in line with our goals of using global package managers. Note that we don't trigger `--save`. @Turbo87 I'd prefer to _only_ do `npm install -g bower`...

blueprints/app/files/bower.json
```diff
@@ -1,5 +0,0 @@
-{
```
nathanhammond: I'm tempted to move this into a different generator rather than just omitting it. `bower init` is _way_ overkill for something that will never be run as a `bower` package.

```
$ bower install --save 'jquery'
bower cached        https://github.com/jquery/jquery-dist.git#3.1.1
bower validate      3.1.1 against https://github.com/jquery/jquery-dist.git#*
bower no-json       No bower.json file to save to, use bower init to create one
```

(For people manually installing `bower` packages instead of transitively via `ember install`.)

lib/models/blueprint.js
```diff
@@ -1017,7 +1017,7 @@ let Blueprint = CoreObject.extend({
     return task.run({
       verbose: true,
       packages: packageNamesAndVersions,
-      installOptions,
+      installOptions: installOptions || { save: true },
```
nathanhammond: This should be `merge`d. (I'm aware that the original didn't merge and that this is simply moved.)

lib/tasks/bower-install.js
```diff
@@ -37,32 +53,61 @@ class BowerInstallTask extends Task {
...
+  hasBowerJson() {
+    return existsSync(this.bowerJsonPath);
+  }
+
+  ensureBowerJson() {
+    if (this.hasBowerJson()) {
+      return Promise.resolve();
+    }
+
+    let projectName = this.project.name();
```
nathanhammond: - ✅ `in-repo-addon`
- ✅ `app`
- ✅ `addon` (since it will be the dummy app)

Also, not actually used or relevant.

lib/tasks/bower-install.js
nathanhammond: Note: rather than omitting the file we can do `Object.keys().length` of `dependencies` and `devDependencies` to determine if we need to install it. It's also less likely to break `ember-try`.

blueprints/app/files/bower.json
nathanhammond: (This comment only applies if we maintain the same approach of `hasBowerJson`.)

lib/tasks/bower-install.js
Turbo87: @nathanhammond we have no control over the installed bower version in that case

lib/tasks/bower-install.js
Turbo87: we could, but then again, why have an essentially empty file in the project?

lib/tasks/bower-install.js
Turbo87: in that case you might get a `bower.json` file without the dependency if the `npm install bower` fails

blueprints/app/files/bower.json
Turbo87: not sure how `bower init` is involved here

lib/models/blueprint.js
Turbo87: that would technically be a breaking API change

nathanhammond: I'd call it a bugfix. 😜 YOLO?

blueprints/app/files/bower.json
nathanhammond: This is an ergonomics question, paired with https://github.com/ember-cli/ember-cli/pull/6698#discussion_r96703772 ... a blank `bower.json` file means that users can blissfully use `bower`. Otherwise the user is presented with the above message asking them to `bower init` which is _way_ more than they need.

lib/tasks/bower-install.js
nathanhammond: Don't care too strongly. Can you promote the `.then` chain to the next line for clarity?

lib/tasks/bower-install.js
nathanhammond: @Turbo87 We can

1. Check to see if there is a global, and if it is above our version. If so, we use it.
2. If no global, install one, return.
3. If global and under our version, install into the current `cliPath`.

Note: I'd prefer to always dump this into our top-level `ember-cli` instead of into the project's copy. You'll counter that version matters local to the version of CLI and I'd agree, especially for `bower` since we interact with it programmatically instead of shelling out, and we'll make no changes.

Turbo87: There are several issues with what you're proposing since we're not shelling out to bower but requiring it instead. Also using global bower seems unrelated to this PR and can be discussed separately.

blueprints/app/files/bower.json
Turbo87: The user will not be presented with any message!? We automatically create the file if we have to, though not through bower init.

lib/tasks/bower-install.js
nathanhammond: This conversation went exactly as I expected, I'm comfortable with things as they stand. 😜 

> You'll counter that version matters local to the version of CLI and I'd agree, especially for bower since we interact with it programmatically instead of shelling out, and we'll make no changes.

Turbo87:
@stefanpenner @rwjblue friendly reminder 😉

blueprints/app/files/bower.json
stefanpenner: let me know if i misunderstood this comment thread, but it sounds like:

* @nathanhammond is saying: using a warning a user would get when using `bower` cli directly themselves
*  @Turbo87 is saying: when ember cli invokes `bower`, it handles this case already.

If so, it seems reasonable to me that a user intending to once again use bower, is warned to first use `bower init`.

lib/models/blueprint.js
stefanpenner: Sounds reasonable as a `[BUGFIX]` commit, independent of this PR.

lib/tasks/bower-install.js
```diff
@@ -1,32 +1,48 @@
...
+
   ensureBower() {
     if (this.bower) {
-      return Promise.resolve(this.bower);
+      return Promise.resolve();
     }
 
-    return resolve('bower', { basedir: cliPath })
+    return this.resolveBower()
       .catch(() => this.installBower()
```
stefanpenner: I would prefer if we only installed bower if a specific error occurs. For example, I would hate for some other issue (SyntaxError or something), from accidentally causing `bower` to be installed

lib/tasks/bower-install.js
stefanpenner: vanity nit-pick: if this was `let name = this.project.name()` the `writeJson` bellow could use concise object syntax `{ name }`

stefanpenner:
with https://github.com/ember-cli/ember-cli/pull/6698#discussion_r97881345 fixed, you have my 👍  The other is merely a nit-pick.

Turbo87:
- AppVeyor is broken as always
- @nathanhammond seems to agree with the PR now
- addressed @stefanpenner's comments

anyone please confirm and press the big (currently not so green) merge button 😉 

(do **not** use @homu to merge this, as I've changed the base branch for this PR)

lib/tasks/bower-install.js
```diff
@@ -1,32 +1,53 @@
...
+      return Promise.resolve();
     }
 
-    return resolve('bower', { basedir: cliPath })
-      .catch(() => this.installBower()
-        .then(() => resolve('bower', { basedir: cliPath })))
-      .then(bowerPath => require(bowerPath));
+    return this.resolveBower()
+      .catch(error => {
+        if (error.message.indexOf('Cannot find module \'bower\'') === -1) {
```
rwjblue: Is the error message platform dependent or guaranteed?

Turbo87: tl;dr this is the best we can do for now... (see https://github.com/substack/node-resolve/commit/c622aefeb286e479d536601e30bb828e69f86ec3#commitcomment-20625328)

rwjblue:
👏

PR #6700, opened 2017-01-18
dev: Fix clone URL in add-to-output-repos.sh

Turbo87:
Cloning via HTTPS sucks when you use 2FA


Turbo87:
that AppVeyor failure is just wrong... no idea what's going on there...

rwjblue:
@homu r+

PR #6702, opened 2017-01-18
AppVeyor: Use cache buster for NPM cache

Turbo87:
This should hopefully reduce our NPM issues on AppVeyor as long as we're not using yarn there (aka `master` doesn't care, but `release` and `beta` currently do)


rwjblue:
@homu r+

PR #6704, opened 2017-01-19
Update lockfile to use latest allowed versions.

rwjblue:
This is the result of a `yarn upgrade`...


twokul:
restarted appveyor

nathanhammond:
@homu r+

PR #6707, opened 2017-01-20
Change usage of shims for ember-source@2.11.0 final.

rwjblue:
See:

* https://github.com/ember-cli/ember-cli-shims/pull/82
* https://github.com/emberjs/ember.js/pull/14808

ember-source@2.11.0 will not have shims, this prepares us for that.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -278,10 +278,13 @@ EmberApp.prototype._resolveLocal = function(to) {
   @method _initVendorFiles
 */
 EmberApp.prototype._initVendorFiles = function() {
+  var bowerDeps = this.project.bowerDependencies();
   var ember = this.project.findAddonByName('ember-source');
+  var addonEmberCliShims = this.project.findAddonByName('ember-cli-shims');
+  var bowerEmberCliShims = bowerDeps['ember-cli-shims'];
   var developmentEmber;
   var productionEmber;
-  var emberShims;
+  var emberShims = null;
   var jquery;
 
   if (ember) {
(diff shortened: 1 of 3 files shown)


lib/broccoli/ember-app.js
```diff
@@ -278,10 +278,13 @@ EmberApp.prototype._resolveLocal = function(to) {
   @method _initVendorFiles
 */
 EmberApp.prototype._initVendorFiles = function() {
+  var bowerDeps = this.project.bowerDependencies();
   var ember = this.project.findAddonByName('ember-source');
+  var emberCliShimsFromAddon = this.project.findAddonByName('ember-cli-shims');
+  var emberCliShimsFromBower = bowerDeps['ember-cli-shims'];
```
twokul: `addonEmberCliShims` and `bowerEmberCliShims` perhaps?

lib/broccoli/ember-app.js
```diff
@@ -291,7 +294,11 @@ EmberApp.prototype._initVendorFiles = function() {
     jquery           = ember.paths.jquery;
   } else {
     jquery = this.bowerDirectory + '/jquery/dist/jquery.js';
-    emberShims = this.bowerDirectory + '/ember-cli-shims/app-shims.js';
+
+    if (emberCliShimsFromBower) {
+      emberShims = this.bowerDirectory + '/ember-cli-shims/app-shims.js';
```
twokul: `path.join`?

lib/broccoli/ember-app.js
rwjblue: Yeah, that reads better to me, thank you!

lib/broccoli/ember-app.js
rwjblue: this is just following what we did before, and I believe that we actually want to guarantee posix style paths (`/`).

twokul: path would guarantee posix styles paths, no? totally understand if you don't want to change as a part of this PR though.

rwjblue: No `path.join` would use system separator, so on windows it would be `\` and on posix it would be `/`.  We want it to always be `/`.

nathanhammond:
LGTM.

PR #6708, opened 2017-01-21
Adds in liveReloadFilters options

briangonzalez:
Addresses the following RFC: https://github.com/ember-cli/rfcs/pull/71


locks:
From what I could find the mentioned RFC was never transfered to the new RFC process, so I'll be closing the PR. Thanks for your effort!

PR #6716, opened 2017-01-25
Remove link to transition guide when ember-cli-build.js file is missing

### LINKED ISSUE #6709: Improve "No ember-cli-build.js found" error message
see https://github.com/ember-cli/ember-cli/blob/5176549db6138da5f3d9cd71243f8b306ca93b60/lib/models/builder.js#L56

This error message should IMHO not include the transition URL and message, since it only applies if you have an existing `Brocfile.js` in the same folder.

It should also use the `SilentError` class instead, because there is really no point in showing the stacktrace of the error here.

mike183:
Resolves #6709

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -53,7 +53,7 @@ class Builder extends CoreObject {
     } else if (buildFile) {
       this.tree = buildFile({ project: this.project });
     } else {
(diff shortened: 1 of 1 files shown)


rwjblue:
@homu r+

Turbo87:
@mike183 awesome, thanks for fixing that!

mike183:
@Turbo87 Not a problem, a couple of small PRs is the least I can do considering the amount I've used Ember in recent projects.

PR #6717, opened 2017-01-26
Make instrumentation more resilient to errors

hjdivad:
1. if `stopAndReport` is unable to `token.stop()` because the heimdall stack is unbalanced, log an error but do not throw.  This case is most likely due to some other error and we want the root cause to surface rather than masking it with an error from the instrumentation.

2. Don't try to report command instrumentation when the command promise has rejected.  This is error prone (although less so due to the change listed above) and in any case build timing is far more relevant for regular builds than builds which error (eg b/c we can't grab the port for `ember serve`).

### DIFF
--- a/lib/models/instrumentation.js
+++ b/lib/models/instrumentation.js
@@ -5,6 +5,7 @@ const chalk = require('chalk');
 const experiments = require('../experiments/');
 const heimdallGraph = require('heimdalljs-graph');
 const utilsInstrumentation = require('../utilities/instrumentation');
+const logger = require('heimdalljs-logger')('ember-cli:instrumentation');
 
 let vizEnabled = utilsInstrumentation.vizEnabled;
 let instrumentationEnabled = utilsInstrumentation.instrumentationEnabled;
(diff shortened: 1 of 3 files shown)


lib/models/instrumentation.js
```diff
@@ -244,7 +244,12 @@ class Instrumentation {
     if (!instr.token) {
       throw new Error(`Cannot stop instrumentation "${name}".  It has not started.`);
     }
-    instr.token.stop();
+    try {
+      instr.token.stop();
+    } catch (e) {
+      this.ui.writeLine(chalk.red(`Error reporting instrumentation '${name}'.  Stack: ${this._heimdall.stack}`));
```
rwjblue: should we also include e.stack (so we can actually triage from the output)? Possibly this could just be a logger.error though (if we want to keep the actual console clear of heimdall errors)...

hjdivad: 👍

rwjblue:
@homu r+

tests/unit/models/instrumentation-test.js
```diff
@@ -391,6 +393,17 @@ describe('models/instrumentation.js', function() {
       }).to.throw('Cannot stop instrumentation "init".  It has not started.');
     });
 
+    it('warns if heimdall stop throws (eg when unbalanced)', function() {
```
Turbo87: @hjdivad this test appears to be broken on AppVeyor. Can you check?

rwjblue: https://github.com/ember-cli/ember-cli/pull/6747 should fix this test

PR #6724, opened 2017-01-29
remove ember-cli-sri and ember-cli-uglify from addons

kellyselden:
I'd like to start the conversation on whether these are useful as defaults to addons. The only way I can think of is if they are deploying the dummy app to production.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -38,6 +38,10 @@ module.exports = {
     // 100% of addons don't need ember-cli-app-version, make it opt-in instead
     delete contents.devDependencies['ember-cli-app-version'];
 
+    // 100% of addons don't need these, make it opt-in instead
+    delete contents.devDependencies['ember-cli-sri'];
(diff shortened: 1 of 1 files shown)


rwjblue:
It seems pretty common for folks to deploy the dummy app (often used for demo apps and published to GH-pages for example). 

What is the cost of keeping things closer between an app and addon here?

kellyselden:
> What is the cost of keeping things closer between an app and addon here?

Not much. I was just auditing the footprint of a fresh addon. Much like ember-release, if people are using them, and if the deps are worth it.

Turbo87:
@kellyselden can you add this to the agenda for next week? we should discuss this and either merge or close this PR

kellyselden:
Done

nathanhammond:
In favor of closing–I definitely take advantage of these in publishing demo apps.

kellyselden:
Closing for now. We couldn't come to a conclusion whether this would be better or worse for most people, so leaving for now.

PR #6472, opened 2016-11-27
add ability to clean up old files in generators

kellyselden:
A response to https://github.com/ember-cli/ember-cli/pull/6469#issuecomment-263076360

This will probably need an RFC because it introduces a new blueprint member `oldFilesToRemove`. A blueprint can override the array and give a list of old file names that the generator (ember init or your own custom blueprint) will clean up for you.

PR #6731, opened 2017-01-30
Revert changes removing `bower.json` from default blueprints.

rwjblue:
This reverts commit be142aaf7801bf64f4322583c7d82ae7c7066c52, reversing changes made to b90f8602eff3e71e9f7ff61541a740bfd25ea503.

This is being reverted due to user experience issues when the `bower.json` file is no longer part of the blueprint.

As things exist prior to a revert, any user upgrading from a version of ember-cli prior to 2.11 would never be instructed to update their `bower.json` to remove entries for `ember`, `ember-cli-shims`, etc.

To bring the changes being reverted back, we need to land the ability for a blueprint to indicate that a file should be deleted as part of the `ember init` process.  This was RFC'ed in ember-cli/rfcs#92 and the implementation is pending in https://github.com/ember-cli/ember-cli/pull/6472 and should hopefully land in 2.13 (current ember-cli#master branch).

This was discussed at the ember-cli meeting on 2017-01-26.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -54,6 +54,14 @@ module.exports = {
     this._writeContentsToFile(sortPackageJson(contents), 'package.json');
   },
 
+  generateBowerJson() {
+    let contents = this._readContentsFromFile('bower.json');
+
+    contents.name = '<%= addonName %>';
+
+    this._writeContentsToFile(contents, 'bower.json');
+  },
+
   afterInstall() {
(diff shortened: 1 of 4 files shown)


rwjblue:
@Turbo87 - Ya, that seems reasonable.  I'll review and update.

rwjblue:
@Turbo87 - OK, updated to revert just those two commits.  Please review and make sure I didn't miss any?

rwjblue:
@homu r+

PR #6736, opened 2017-01-30
start using filesToRemove

kellyselden:
A good starting point with `Brocfile.js` and `testem.json` cleanup.

### DIFF
--- a/blueprints/app/index.js
+++ b/blueprints/app/index.js
@@ -5,6 +5,11 @@ const stringUtil = require('ember-cli-string-utils');
 module.exports = {
   description: 'The default blueprint for ember-cli projects.',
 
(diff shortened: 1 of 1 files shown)


rwjblue:
👍

rwjblue:
We should also revert https://github.com/ember-cli/ember-cli/pull/6731 and add `bower.json` to the array (on master only)

kellyselden:
> We should also revert #6731 and add bower.json to the array (on master only)

@rwjblue That one was a little more in flux, and didn't want to involve it with this uncontroversial PR.

rwjblue:
haha, yes, I was talking generally (not for this PR)

rwjblue:
@homu r+

PR #6739, opened 2017-02-01
Remove bower.json files again

Turbo87:
Reverts #6731 and adds `bower.json` to the `filesToRemove` arrays

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -17,6 +17,7 @@ module.exports = {
   description: 'The default blueprint for ember-cli addons.',
 
   filesToRemove: [
+    'bower.json',
     'Brocfile.js',
     'testem.json',
   ],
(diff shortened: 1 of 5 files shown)


blueprints/addon/index.js
```diff
@@ -16,6 +16,10 @@ const stringifyAndNormalize = require('../../lib/utilities/stringify-and-normali
 module.exports = {
   description: 'The default blueprint for ember-cli addons.',
 
+  filesToRemove: [
```
kellyselden: is having this in the app blueprint enough? I never tested an addon for existing filesToRemove.

Turbo87: I'm not sure, but I think the `addon` blueprint only inherits the `files` folder, but nothing else.

Turbo87:
rebased on top of #6741

rwjblue:
@homu r+

PR #6741, opened 2017-02-01
addon needs to mirror filesToRemove from app

kellyselden:
Came to light in #6739

@Turbo87 apologies if you need to rebase.


kellyselden:
tested it locally and it indeed the case.

kellyselden:
@homu r+

PR #6742, opened 2017-02-01
Promote cacheKeyForTree to public API

trentmwillis:
As discussed in ember-cli/rfcs#90. Should be good to go once the RFC is merged.

Pro tip: use `?w=1` when viewing the changes to tidy up the large amount of whitespace changes in the tests.


rwjblue:
https://github.com/ember-cli/rfcs/pull/90 has landed, so this is now ready to land

@homu r+

PR #6745, opened 2017-02-02
ensure SIGINT ember serve produces instrumentation

hjdivad:
Previously the command's run promise would not resolve, and subsequent then handlers responsible for producing instrumentation would never run.

Cleanup occurred due to capture-exit.

Now our exit handler not only cleans up, but also finishes the run promise chain.

Also ensure ServeCommand.run invokes its task via runTask so our onInterrupt hook works.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -145,10 +145,10 @@ class CLI {
         }
       }
 
-      let onCommandInterrupt = command.onInterrupt.bind(command);
       let instrumentation = this.instrumentation;
+      let onCommandInterrupt;
 
-      return Promise.resolve().then(() => {
+      let runPromise = Promise.resolve().then(() => {
         instrumentation.stopAndReport('init');
         instrumentation.start('command');
 
@@ -156,7 +156,6 @@ class CLI {
         onProcessInterrupt.addHandler(onCommandInterrupt);
 
         return command.beforeRun(commandArgs);
-
       }).then(() => {
         loggerTesting.info('cli: command.validateAndRun');
 
@@ -172,9 +171,6 @@ class CLI {
         shutdownOnExit = function() {
           instrumentation.stopAndReport('shutdown');
         };
-        // Ensure that the shutdown instrumentation hook is invoked properly
-        // even if we exit before hitting the `finally` below
-        onProcessInterrupt.addHandler(shutdownOnExit);
       }).then(result => {
         // if the help option was passed, call the help command
         if (result === 'callHelp') {
@@ -203,13 +199,16 @@ class CLI {
           }
         });
       });
+
+      onCommandInterrupt = () =>
+        Promise.resolve(command.onInterrupt())
+          .then(() => runPromise);
+
+      return runPromise;
     })
     .finally(() => {
       if (shutdownOnExit) {
-        // invoke instrumentation shutdown and
-        // remove instrumentation interruption handlers
         shutdownOnExit();
-        onProcessInterrupt.removeHandler(shutdownOnExit);
       }
     })
     .catch(this.logError.bind(this));
--- a/lib/commands/generate.js
+++ b/lib/commands/generate.js
@@ -38,14 +38,6 @@ module.exports = Command.extend({
                                             'blueprint name to be specified. ' +
                                             'For more details, use `ember help`.'));
     }
-    let Task = this.tasks.GenerateFromBlueprint;
-    let task = new Task({
-      ui: this.ui,
-      analytics: this.analytics,
-      project: this.project,
-      testing: this.testing,
-      settings: this.settings,
-    });
 
     let taskArgs = {
       args: rawArgs,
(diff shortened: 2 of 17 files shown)


hjdivad:
cc @ro0gr

lib/cli/cli.js
```diff
@@ -203,13 +199,14 @@ class CLI {
           }
         });
       });
+
+      onCommandInterrupt = () => Promise.resolve(command.onInterrupt()).then(() => runPromise);
```
rwjblue: I have a slight preference to add a newline or two here, to make it easier to read...

```js
onCommandInterrupt = () => 
  Promise.resolve(command.onInterrupt())
    .then(() => runPromise);
```

tests/unit/models/builder-test.js
```diff
@@ -181,6 +179,21 @@ describe('models/builder.js', function() {
...
+  describe('cleanup', function() {
+    beforeEach(function() {
+      builder = new Builder({
+        setupBroccoliBuilder,
+        project: new MockProject(),
+        processBuildResult(buildResults) { return Promise.resolve(buildResults); },
+      });
+    });
+
+    it('is idempotent', function() {
```
rwjblue: Minor nit-pick: Can we return the promise from the test so it can complete and avoid leaking?

tests/unit/tasks/build-watch-test.js
```diff
@@ -0,0 +1,77 @@
...
+      _watcher,
+    };
+    return task.run(options);
+  }
+
+  describe('onInterrupt', function() {
+    it('fulfills the run promise and cleans up the builder', function() {
+      let runPromise = runBuildWatchTask();
+
+      RSVP.resolve().then(() => task.onInterrupt());
```
rwjblue: What is happening here?

lib/commands/generate.js
```diff
@@ -61,7 +53,17 @@ module.exports = Command.extend({
...
-    return task.run(taskOptions);
+    return this.runTask('GenerateFromBlueprint', taskOptions);
+  },
+
+  _env() {
+    return {
+      ui: this.ui,
+      analytics: this.analytics,
+      project: this.project,
+      testing: this.testing,
```
ro0gr: Should we extend `Command._env()` with missing `settings` and `testing` props?

[Looks like](https://github.com/ember-cli/ember-cli/blob/ace2656c2637093d8443d907f70b45ae49b7ea30/lib/cli/cli.js#L103-L110) we always have these props inside commands.

lib/commands/init.js
```diff
@@ -141,8 +101,20 @@ module.exports = Command.extend({
...
       .then(() => {
         if (commandOptions.skipGit === false) {
-          return gitInit.run(commandOptions, rawArgs);
+          return this.runTask('GitInit', commandOptions, rawArgs);
         }
       });
   },
+
+  _env(name) {
+    if (name === 'GenerateFromBlueprint') {
```
ro0gr: Why don't we just pass all this stuff always?

lib/cli/cli.js
Turbo87: how about this:
```js
onCommandInterrupt() {
  return Promise.resolve(command.onInterrupt())
    .then(() => runPromise);
}

tests/unit/models/builder-test.js
hjdivad: Yes but I wouldn't personally call it a nit.  Good catch on this oversight ^_^

lib/commands/generate.js
hjdivad: ah good idea.  might be able to clean up a bunch of the `_env`s with this

lib/commands/init.js
hjdivad: i had it in there to preserve semantics (we were not previously passing `testing` to this task, but you've convinced me to just always pass those options

tests/unit/tasks/build-watch-test.js
hjdivad: The goal is just to call `onInterrupt` async; can do `setTimeout` if you think that would be clearer.

hjdivad:
@rwjblue @ro0gr updated

1. minor formatting
2. moved settings & testing to `lib/models/command` so we always pass them in and don't need all the `_env` overrides.

@rwjblue assuming tests 👍  i think it's good to merge but LMK if you want any more revision.

tests/unit/tasks/build-watch-test.js
rwjblue: The issue I had with it is that we do not actually wait for this to resolve. Wouldn't the test framework just move on to the next test before this is completed?

rwjblue: Now that I reread the test, it is possible that this is the entire point....

hjdivad: @rwjblue per our offline conversation we're 👍 on this test now yes?

hjdivad:
@rwjblue i think this is good to merge.

There were a couple of red tests that greened on a rebuild but they seem totally unrelated to this PR (eg one was linting timing out).

rwjblue:
@homu r+

PR #6746, opened 2017-02-02
WIP: remove ember-cli-legacy-blueprints

kellyselden:
now that ember-source provides our blueprints

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "ember-cli-broccoli-sane-watcher": "^2.0.3",
     "ember-cli-get-component-path-option": "^1.0.0",
     "ember-cli-is-package-missing": "^1.0.0",
-    "ember-cli-legacy-blueprints": "^0.1.2",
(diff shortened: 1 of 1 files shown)


Turbo87:
as mentioned on Slack: this would remove the blueprints for all users that are not using ember-source yet and I don't see a reason why we would need to do that any time soon. let's revisit that if we should ever release a v3.0.0.

kellyselden:
I ran `ember h g` and saw that this addon was winning vs ember-source

rwjblue:
yeah, that is definitely not right

Turbo87:
> I ran ember h g and saw that this addon was winning vs ember-source

then we should debug this.

https://github.com/ember-cli/ember-cli-legacy-blueprints/blob/master/package.json#L57-L60 and https://github.com/emberjs/ember.js/blob/master/package.json#L92 look correct to me

Turbo87:
I've opened #6751 to track this issue

PR #6748, opened 2017-02-02
Use yarn if yarn.lock exists or `--yarn` is used

### LINKED ISSUE #6341: Explore if YARN can be used as the bundled `npm`
Note: Yarn + ember-cli seem to work together, this issue is specifically for exploring ember-cli + yarn bundled together

---

Once the dust settles, we can continue this an RFC. For now, this is purely fact-finding to inform a future RFC
- [ ] currently not aligned with Node LTS: ~~https://github.com/yarnpkg/yarn/issues/650~~ https://github.com/yarnpkg/rfcs/pull/1 (but seems like it easily could be, and am volunteering to help with the effort).
- [ ] programmatic API (someone needs to check if this exists, and if it can be added to the "public" api)

Turbo87:
Resolves #6341 

- [x] rewrite `ember install` to use yarn if `yarn.lock` exists
- [x] rewrite `ember init/new/addon` to use yarn if `--yarn` is used
- [x] add `--yarn` (enforce yarn usage) and `--no-yarn` (enforce NPM usage) options to `ember install` 
- [x] write tests
- [x] rebase on top of https://github.com/ember-cli/ember-cli/pull/6745 to resolve conflict
- [x] remove `--no-yarn` default for `ember init` command (`yarn.lock` should be detected for project upgrades)
- [x] adjust `.travis.yml` files to use `yarn` if requested

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -5,11 +5,15 @@ node_js:
 
 sudo: false
 
-cache:
+<% if (yarn) { %>cache:
+  yarn: true
+  directories:
+    - $HOME/.cache # includes bowers cache
+<% } else { %>cache:
   directories:
     - $HOME/.npm
     - $HOME/.cache # includes bowers cache
-
+<% } %>
 env:
   # we recommend testing LTS's and latest stable release (bonus points to beta/canary)
   - EMBER_TRY_SCENARIO=ember-lts-2.4
@@ -24,7 +28,17 @@ matrix:
   allow_failures:
     - env: EMBER_TRY_SCENARIO=ember-canary
 
-before_install:
+<% if (yarn) { %>before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+  - yarn global add bower phantomjs-prebuilt
+  - bower --version
+  - phantomjs --version
+
+install:
+  - yarn install
+  - bower install
+<% } else { %>before_install:
   - npm config set spin false
   - npm install -g bower phantomjs-prebuilt
   - bower --version
@@ -33,7 +47,7 @@ before_install:
 install:
   - npm install
   - bower install
-
+<% } %>
 script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -82,24 +82,43 @@ class NpmTask extends Task {
   }
 
   findPackageManager() {
-    if (!this.hasYarnLock()) {
-      logger.info('yarn.lock not found -> using npm');
+    if (this.useYarn === undefined) {
+      if (!this.hasYarnLock()) {
+        logger.info('yarn.lock not found -> using npm');
+        return this.checkNpmVersion();
+      }
+
+      logger.info('yarn.lock found -> trying yarn');
+      return this.checkYarn()
+        .then(() => {
+          logger.info('yarn found -> using yarn');
+          this.useYarn = true;
+        })
+        .catch(() => {
+          logger.info('yarn not found -> using npm');
+          return this.checkNpmVersion();
+        });
+
+    } else if (this.useYarn) {
+      logger.info('yarn requested -> trying yarn');
+
+      return this.checkYarn().catch(error => {
+        if (error.code === 'ENOENT') {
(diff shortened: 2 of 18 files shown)


rwjblue:
Can you confirm that by adding `--yarn` as a flag, we allow the usage of project level and/or global `.ember-cli` files (e.g. if I have a `~/.ember-cli` with `{ "yarn": true }` I can just run `ember new foo` and it is the same as `ember new foo --yarn`)?

Turbo87:
@rwjblue yep, that seems to work fine too :)

lib/tasks/npm-task.js
```diff
@@ -17,6 +19,8 @@ class NpmTask extends Task {
   constructor(options) {
     super(options);
 
+    this.useYarn = ('useYarn' in options) ? options.useYarn : undefined;
```
rwjblue: This seems odd to me since `options.useYarn === ('useYarn' in options) ? options.useYarn : undefined` anyways.

lib/tasks/npm-task.js
```diff
@@ -2,21 +2,25 @@
...
 
 const Task = require('../models/task');
 
 class NpmTask extends Task {
 
-  constructor(options) {
+  constructor(options = {}) {
     super(options);
 
+    this.useYarn = ('useYarn' in options) ? options.useYarn : undefined;
```
rwjblue: Should be able to just do:

```js
this.useYarn = options.useYarn;
```

Basically `options.useYarn` is `undefined` if `useYarn` wasn't included in `options` originally...

lib/tasks/npm-task.js
Turbo87: you're right, seems to work just as fine without it

lib/tasks/npm-task.js
Turbo87: actually we can remove the line completely thanks to CoreObject 😉

lib/tasks/npm-task.js
```diff
@@ -28,7 +36,27 @@ class NpmTask extends Task {
     return RSVP.resolve(execa('npm', args, { preferLocal: false }));
   }
 
-  checkVersion() {
+  yarn(args) {
+    logger.info('yarn: %j', args);
+    return RSVP.resolve(execa('yarn', args, { preferLocal: false }));
```
stefanpenner: If `execa` throws, the `yarn` method will violate the promise guarentee of no throws. I think it would be best to wrap the call in the promise constructor, and call resolve on the result of `execa` so throws while calling `execa`

something like:

```js
return new Promise(resolve => resolve(execa('yarn', args, { preferLocal: false }));
```

this more closely mimics the async function semantics as well.

robclancy:
Is this expected to make the next release? Half of my job is fixing issues with NPM these days.

stefanpenner:
> Is this expected to make the next release?

It will go through the canary -> beta -> release cycle train. So it wont be in the next stable release, but will be usable if you are an early adopter soon.

lib/tasks/npm-task.js
Turbo87: the same is true for the `npm()` method that was previously introduced. have you been able to make `execa()` throw an exception? it seems like execa knows about that contract and converts potential exceptions to Promise rejections itself (see e.g. https://github.com/sindresorhus/execa/blob/master/index.js#L128-L133)

stefanpenner: 👍

Turbo87:
@homu r+

blueprints/addon/files/.travis.yml
```diff
@@ -24,7 +28,17 @@ matrix:
...
-before_install:
+<% if (yarn) { %>before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+  - yarn global add bower phantomjs-prebuilt
+  - bower --version
+  - phantomjs --version
+
+install:
+  - yarn install
```
Turbo87: @rwjblue I'm wondering if this should be `yarn install --no-lockfile` instead

rwjblue: @Turbo87 - Why? To force fresh deps? If that is the case, then `ember-try` scenarios will already test against latest.

I'm not opposed to changing to `--no-lockfile` though

Turbo87: > Why? To force fresh deps?

yes

> If that is the case, then ember-try scenarios will already test against latest.

that's only true for the variable dependencies though, right?

rwjblue: ya, good point

PR #6753, opened 2017-02-07
Interrupt command with an error if no _currentTask

ro0gr:
I think we've slightly changed interruption behavior after #6615 for external commands like:
 - https://github.com/isleofcode/ember-cordova/blob/master/lib/commands/build.js#L106
 - https://github.com/ember-cli-deploy/ember-cli-deploy/blob/master/lib/commands/deploy.js#L53

Previously we've always thrown an `exit code 1` for the commands where the builder is involved. Now we've delegated decision about an exit code to the [`_currentTask`](https://github.com/ro0gr/ember-cli/blob/196ab418013ed5b66c99e7b2a84fc29c917dcc84/lib/models/command.js#L187). 

The problem is that commands who don't use [private](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:external-commands-interruption?expand=1#diff-5346944aa3595cbf8116950d83f80cfeR208) `runTask` will always produce a zero exit code.

This change aims to solve it by producing an `exit code 1` if `_currentTask` is unknown.

@hjdivad @rwjblue

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -186,6 +186,9 @@ let Command = CoreObject.extend({
   onInterrupt() {
     if (this._currentTask) {
       return this._currentTask.onInterrupt();
+    } else {
+      // interrupt all external commands which don't use `runTask()` with an error
+      process.exit(1);
     }
   },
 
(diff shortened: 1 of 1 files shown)


lib/models/command.js
```diff
@@ -186,6 +186,9 @@ let Command = CoreObject.extend({
   onInterrupt() {
     if (this._currentTask) {
       return this._currentTask.onInterrupt();
+    } else {
+      // interrupt all external commands which don't use `runTask()` with an error
```
ro0gr: couldn't find a better term for external commands. Alternative was `3rd party commands`

ro0gr:
I believe test failure on travis should be fixed via #6745

The reason is that currently `serve` command [isn't aware](https://github.com/ember-cli/ember-cli/blob/85bb2ef1b261aab9ec9ba3ca5c58e8ccaf3c1910/lib/commands/serve.js#L78) of `runTask`

rwjblue:
Can you retarget this against the `hjdivad/instrumentation-sigint` branch (there is a button in the github UI for this), so we can land into that branch? I don't feel comfortable landing (and therefore breaking master) independently.

ro0gr:
Haven't found this magic button. Just re-created pr with `hjdivad/instrumentation-sigint` branch as a target

PR #6754, opened 2017-02-08
Interrupt command with an error if no _currentTask

ro0gr:
I think we've slightly changed interruption behavior after #6615 for external commands like:

- https://github.com/isleofcode/ember-cordova/blob/master/lib/commands/build.js#L106
- https://github.com/ember-cli-deploy/ember-cli-deploy/blob/master/lib/commands/deploy.js#L53

Previously we've always thrown an `exit code 1` for the commands where the builder is involved. Now we've delegated decision about an exit code to the `_currentTask`.

The problem is that commands who don't use private `runTask` will always produce a zero exit code.

This change aims to solve it by producing an `exit code 1` if `_currentTask` is unknown.

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -186,6 +186,9 @@ let Command = CoreObject.extend({
   onInterrupt() {
     if (this._currentTask) {
       return this._currentTask.onInterrupt();
+    } else {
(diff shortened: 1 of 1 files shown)


ro0gr:
seems like travis build is not triggered and re-used result of #6753

PR #6755, opened 2017-02-08
Make findHost public

asakusuma:
Seems reasonable for an addon to want access to the host app. Also, this method doesn't seem to be used internally at all.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -566,10 +566,10 @@ let addonProto = {
     This prevents previous addons (prior to `this.import`, ca 2.7.0)
     to break at importing assets when they are used nested in other addons.
 
-    @private
+    @public
     @method _findHost
   */
-  _findHost() {
+  findHost() {
     let current = this;
     let app;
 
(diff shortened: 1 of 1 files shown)


rwjblue:
We have discussed this a bit before, and I'm not nearly as confident that this should be public. 

Can you describe exactly what you are trying to do with the host? Remember that there is pretty little amount of public API on the host app instance, and I feel like we should address that first (before making it public API to get it).

nathanhammond:
Note that in most cases you should just use `this.import` and things will "just work." (This is the only reason that `_findHost` was introduced.)

asakusuma:
I believe I've found a way for the host to send the required info to the addon, rather than the addon querying the host, which solves my issue for now. Closing, I'll reopen if I think the opposite is cleaner.

nightire:
@asakusuma Can you please share more on what you've found to send info from host to addon?

asakusuma:
@nightire I have nested addons register themselves with a top-level addon, who's parent is the host.  Nested addons can find a top level addon via `findAddonByName`.

```
this.project.findAddonByName('some-top-level-addon');
```

PR #6758, opened 2017-02-10
[BUGFIX] - Fixes blueprints noop log removals

### LINKED ISSUE #6744: Blueprints log removals even when noop
When you run `ember d ...`, the console will show removing files even if they aren't there. This is now showing up in the new `filesToRemove` feature even if the files are not there. Should be a relatively easy fix.

gadogado:
This fixes a bug where blueprint removals were always logging as successful regardless if the designated file(s) for removal didn't exist.  I changed the `cleanRemove` utility to accept an optional callback function that runs when an `ENOENT` error occurs.  

I added a callback function to `remove` in the blueprint model which results in the ui indicating the`file does not exist` for each missing file. 

Not sure if this is the right approach and appreciate any feedback!

```
$ ember d route nothing-here
uninstalling route
  remove app/routes/nothing-here.js
  remove app/templates/nothing-here.hbs
  does not exist app/routes/nothing-here.js
  does not exist app/templates/nothing-here.hbs
updating router
  remove route nothing-here
uninstalling route-test
  remove tests/unit/routes/nothing-here-test.js
  does not exist tests/unit/routes/nothing-here-test.js
```

Resolves #6744

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1546,8 +1546,10 @@ function finishProcessingForInstall(infos) {
 }
 
 function finishProcessingForUninstall(infos) {
-  infos.forEach(markToBeRemoved);
-  return infos;
+  let validInfos = infos.filter(info => existsSync(info.outputPath));
+  validInfos.forEach(markToBeRemoved);
+
+  return validInfos;
 }
 
 module.exports = Blueprint;
--- a/tests/integration/models/blueprint-test.js
+++ b/tests/integration/models/blueprint-test.js
@@ -621,6 +621,19 @@ describe('Blueprint', function() {
         expect(actualFiles).to.contain('app/basics/mock-project.txt');
       });
     });
+
+    it('uninstall doesn\'t log remove messages when file does not exist', function() {
+      options.entity = { name: 'does-not-exist' };
+
+      return blueprint.uninstall(options)
+      .then(function() {
+        let output = ui.output.trim().split(EOL);
+        expect(output.shift()).to.match(/^uninstalling/);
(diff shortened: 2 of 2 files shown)


Turbo87:
@gadogado thanks for your solution here! there is one small issue though. in "dry" mode you no longer see that a file is supposed to be removed. maybe we should use `existsSync()` here to check if the file exists, and only then show the `remove` message and call `cleanRemove()` if not in dry mode.

gadogado:
@Turbo87 thanks for the good suggestion.  I'll get something in shortly.

gadogado:
@Turbo87 is this better?
```
13:23 $ ember d --dry-run route nothing-here
uninstalling route
You specified the dry-run flag, so no files will be deleted.
  does not exist app/routes/nothing-here.js
  does not exist app/templates/nothing-here.hbs
uninstalling route-test
You specified the dry-run flag, so no files will be deleted.
  does not exist tests/unit/routes/nothing-here-test.js
```

Turbo87:
@gadogado I think I'd drop the `does not exist` message completely and only show a message if the file actually exists

gadogado:
@Turbo87 I dropped the remove filePath messages if the file doesn't exist.

Turbo87:
@gadogado hmm, I'm not sure this works as expected. steps to reproduce:

- `npm link` in the `ember-cli` folder
- change to a temporary folder
- `ember new foo`

I'm still seeing e.g. `remove Brocfile.js` here

gadogado:
@Turbo87 good catch;  I was focusing on `ember d` and didn't check `ember new`.   You mind looking this over again when you get a sec?

rwjblue:
@homu r+

PR #6766, opened 2017-02-10
Remove flagging for `experiments.INSTRUMENTATION`.

rwjblue:
Now that the instrumentation hook is public API, this removes the
internal code guards and cleans things up a bit.

This PR likely should wait until after we branch 2.13 into the beta branch...


Turbo87:
@rwjblue looks like ESLint doesn't like this 😉

hjdivad:
reminder to ourselves about a couple of other things to do re: this effort

- [ ] clean up bin (#6769)
- [ ] update broccoli-viz

rwjblue:
@homu r+

PR #6768, opened 2017-02-10
Normalize end-of-line characters in strings to compare prior to diffing

### LINKED ISSUE #6767: ember init - diff shows each line changed when having non-uniform line endings
Diffing in `lib/models/file-info` seems to not ignore white space differences and this caused difficulties for me when working on systems with unequal line endings.
`init` will prompt for every single file to be overwritten. It will show that each line is different. 

Procedure:
1) Updated ember-cli on a windows machine
2) Checked out an ember app located on a mac and did all the usual pkg update procedures, finalizing with `ember init`
3) `ember init` complained that all files differ on every line, turns out because of the varying line endings

There might be some line ending normalization needed in this case prior to passing the input to createPatch. I think my git config is also as it should be: `core.autocrlf input` on mac and `core.autocrlf true` on windows, since on windows it is definitely preferred to have \r\n line endings in the same way it is not preferred to have \r\n on bsd or gnu/linux.

koopa:
Resolves #6767

### DIFF
--- a/lib/models/file-info.js
+++ b/lib/models/file-info.js
@@ -5,6 +5,7 @@ const RSVP = require('rsvp');
 const chalk = require('chalk');
 const EditFileDiff = require('./edit-file-diff');
 const EOL = require('os').EOL;
+const rxEOL = new RegExp(EOL, 'g');
 const isBinaryFile = require('isbinaryfile').sync;
 const canEdit = require('../utilities/open-editor').canEdit;
 const processTemplate = require('../utilities/process-template');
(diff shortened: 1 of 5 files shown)


stefanpenner:
looks good, can a unit test be added?

koopa:
I will gladly provide one as soon as I figure out how to properly (still quite new to this).
Also I messed up with the replace, I should probably wait until that is resolved and not try to replace a promise :)

btecu:
Thanks for fixing this!

stefanpenner:
it appears there is a legit test failure on posix:

```
not ok 770 Unit - FileInfo normalizes line endings before comparing files
  AssertionError: expected 'confirm' to equal 'identical'
      at tests/unit/models/file-info-test.js:177:23
      at tryCatch (node_modules/rsvp/dist/rsvp.js:538:12)
      at invokeCallback (node_modules/rsvp/dist/rsvp.js:553:13)
      at publish (node_modules/rsvp/dist/rsvp.js:521:7)
      at flush (node_modules/rsvp/dist/rsvp.js:2373:5)
```

koopa:
This should take care of that

Turbo87:
@homu r+

Turbo87:
@homu retry

PR #6770, opened 2017-02-13
models/addon: Add @since tag to this.import()

Turbo87:


rwjblue:
appveyor failure on node 4 seems related to bad checksum for tarball download

@homu r+

PR #6772, opened 2017-02-13
`ember new` with blueprints from npm

tomdale:
This PR adds support for installing blueprints from npm. Previously, we supported installing blueprints from the local filesystem or via an explicit `.git` URL.

This feature works by looking at the blueprint name and only attempting to install the blueprint from npm if:

1. The blueprint is a valid npm package name.
2. No blueprint with that name was found locally.

This ends up being nicer from a developer ergonomics perspective because I can just say something like "Type `ember new my-app -b pwa-blueprint`" instead of "Type `ember new my-app -b https://github.com/tomdale/pwa-blueprint.git`".

The feature is wrapped in the `NPM_BLUEPRINTS` experiment flag so hopefully it can be iterated on without too much risk if people think this is a good overall direction.

### DIFF
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const Blueprint = require('../models/blueprint');
-const Task = require('../models/task');
 const RSVP = require('rsvp');
 const SilentError = require('silent-error');
 const isGitRepo = require('is-git-url');
@@ -10,6 +8,10 @@ const temp = require('temp');
 const execa = require('execa');
 const path = require('path');
 const merge = require('ember-cli-lodash-subset').merge;
+
+const Blueprint = require('../models/blueprint');
+const Task = require('../models/task');
+const experiments = require('../experiments');
 
 // Automatically track and cleanup temp files at exit
 temp.track();
@@ -73,15 +75,17 @@ function lookupBlueprint(blueprintName, paths) {
   try {
     blueprint = Blueprint.lookup(blueprintName, { paths });
   } catch (err) {
-    // If we didn't find it locally, we'll see if it's a valid npm package name
-    // and install it from npm if so.
-    if (maybeNpmPackage(blueprintName)) {
-      return downloadBlueprint(blueprintName, fromNpm);
+
+    if (experiments.NPM_BLUEPRINTS) {
+      // If we didn't find it locally, we'll see if it's a valid npm package name
+      // and install it from npm if so.
+      if (maybeNpmPackage(blueprintName)) {
+        return downloadBlueprint(blueprintName, fromNpm);
+      }
     }
 
     // Blueprint wasn't found and it couldn't be an npm package, so re-throw the
     // original exception.
-
     return Promise.reject(err);
   }
 
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -7,7 +7,7 @@ const SilentError = require('silent-error');
 const isGitRepo = require('is-git-url');
 const validateNpmPackageName = require('validate-npm-package-name');
 const temp = require('temp');
-const childProcess = require('child_process');
+const execa = require('execa');
 const path = require('path');
 const merge = require('ember-cli-lodash-subset').merge;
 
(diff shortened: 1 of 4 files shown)


lib/tasks/install-blueprint.js
```diff
@@ -38,22 +46,100 @@ class InstallBlueprintTask extends Task {
...
+  let pkgPath = path.join(tempPath, 'node_modules', blueprintName);
+
+  return execa('npm', ['install', blueprintName], { cwd: tempPath })
+    .catch(err => throwNpmError(err))
+    .then(() => pkgPath);
+}
+
+const NOT_FOUND_REGEXP = /npm ERR! 404 {2}'(\S+)' is not in the npm registry/;
+
+function throwNpmError(err) {
```
stefanpenner: can we unit test this function?

lib/tasks/install-blueprint.js
```diff
@@ -38,22 +46,100 @@ class InstallBlueprintTask extends Task {
...
+// repository so it has all of its dependencies.
+function fromGit(blueprintName, tempPath) {
+  return execa('git', ['clone', blueprintName, tempPath])
+    .then(() => execa('npm', ['install'], { cwd: tempPath }))
+    .then(() => tempPath);
+}
+
+// Install the blueprint with npm. We skip running `npm install` inside like the
+// git version because npm handles this for us automatically.
+function fromNpm(blueprintName, tempPath) {
```
stefanpenner: can we unit test this function?

lib/tasks/install-blueprint.js
```diff
@@ -38,22 +46,100 @@ class InstallBlueprintTask extends Task {
...
+    return lookupBlueprint(blueprintName, blueprintPaths)
+      .then(blueprint => blueprint.install(installOptions));
+  }
+}
+
+// Given a blueprint name and set of paths to search, looks up the specified
+// blueprint. If it detects a git repository URL, it will clone the repository
+// and use that as the blueprint. If it detects an npm package name, it will
+// install the package and use that as the blueprint.
+function lookupBlueprint(blueprintName, paths) {
```
stefanpenner: can we unit test this function?

Turbo87:
@tomdale @stefanpenner I can take over on this if you want me to

tomdale:
@Turbo87 Seems legit, I don't think I'll be able to get to it today.

Turbo87:
@tomdale I'm wondering if we should use a `npm:` prefix for this (e.g. `ember new my-app -b npm:pwa-blueprint`). mixing up blueprint names, git urls and npm package names without some way to distinguish them feels a little confusing to me.

stefanpenner:
> @tomdale I'm wondering if we should use a npm: prefix for this (e.g. ember new my-app -b npm:pwa-blueprint). mixing up blueprint names, git urls and npm package names without some way to distinguish them feels a little confusing to me.

I like that! @tomdale any downside?

tomdale:
I don't love the idea of a prefix, because the motivation is to have something as concise as possible for people to type.

In practice I don't think there's actually that much ambiguity (at least, specifically for the `ember new` case):

1. No ambiguity in git URLs as we require them to be very explicit, including a `.git` prefix at the end.
2. The set of `new`-compatible blueprints is very bounded (just `app` and `addon`).
3. Everything else is an npm package.

I think there's a more worrying ambiguity case with `ember generate`, where the likelihood of collisions is higher. As a compromise, how would we feel about scoping npm-sourced blueprints to just the `ember new` case?

I'd also be open to a different flag that took an `npm install`-compatible package identifier (e.g. `ember-cli/template` for GitHub, `gist:ember-cli/template` for Gists, `bitbucket:foo/bar`, etc.)

rwjblue:
> I don't love the idea of a prefix, because the motivation is to have something as concise as possible for people to type.

I agree with @tomdale here, I'd prefer to avoid the prefix in what seems like the default case (npm packages) for `ember new` or `ember addon`.

> As a compromise, how would we feel about scoping npm-sourced blueprints to just the ember new case?

I think this would be fine, and as you mentioned we only support `app` and `addon` as blueprint names outside of a project context anyways so there is little worry of collisions...

PR #6776, opened 2017-02-15
Implement targets RFC

cibernox:
This PR implements https://github.com/ember-cli/rfcs/pull/95 (still unmerged).

API example

```js
app.project.targets; // { browsers: [...], node: '6.6.0' };
```

### DIFF
--- a/blueprints/app/files/config/targets.js
+++ b/blueprints/app/files/config/targets.js
@@ -1,0 +1,10 @@
+/* eslint-env node */
+
+module.exports = {
+  browsers: [
+    'ie 9',
+    'last 1 Chrome versions',
+    'last 1 Firefox versions',
+    'last 1 Safari versions'
+  ]
+};
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -17,6 +17,7 @@ let emberCLIVersion = versionUtils.emberCLIVersion;
 const findAddonByName = require('../utilities/find-addon-by-name');
 const Instrumentation = require('./instrumentation');
 const heimdall = require('heimdalljs');
+const experiments = require('../experiments');
 
 let processCwd = process.cwd();
 // ensure NULL_PROJECT is a singleton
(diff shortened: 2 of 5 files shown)


lib/models/project.js
```diff
@@ -216,6 +216,39 @@ class Project {
...
+
+    if (this.pkg['ember-addon'] && this.pkg['ember-addon']['configPath']) {
+      configPath = this.pkg['ember-addon']['configPath'];
+    }
+
+    let targetsPath = path.join(this.root, configPath, 'targets');
+
+    let appTargets;
+    if (existsSync(`${targetsPath}.js`)) {
+      appTargets = JSON.parse(fs.readFileSync(`${targetsPath}.js`, 'utf8'));
```
rwjblue: I believe this should use `require` since the targets file isn't valid JSON and we want to allow usage of node-land things anyways.

blueprints/app/files/config/targets.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-env node */
+
+module.exports = {
+  browsers: [] // ["last 2 versions", "safari >= 7"]
```
rwjblue: Let's inline the default here also (instead of the comment). This lets us update the default for new apps without potentially breaking existing ones.

lib/models/project.js
```diff
@@ -216,6 +216,39 @@ class Project {
...
 
   /**
+    Returns the targets of this project, or the default targets if not present
+
+    @public
+    @method targets
+    @return {Object}  Targets object
+   */
+  get targets() {
+    let configPath = 'config';
```
rwjblue: we should memoize this to avoid reading from disk on every access

cibernox: Done.

lib/models/project.js
cibernox: Done, but test fail. I've created the file with the right content, I can't `require` it. 
I suspect that must be something obvious related with how the fake filesystem works in tests.

cibernox:
All feedback applied. There is still the issue of the failing test because `require` can't find the `${projectPath}/config/targets`, while `fs.readFileSync(${projectPath}/config/targets.js)` can read it.

cibernox:
I've fixed the tests and validated its behaviour in app (installing it from my github).

Open implementation questions:

- At the moment the `config/targets.js` exports a simple POJO. `/config/ember-try.js` follows the same approach. `config/environment.js` instead exports a function that when called with the environment returns the POJO. If you think that function-that-returns-pojo provides some value now it's the time.

- When the package.json doesn't have an `engines` property the targets don't include the `node` property. I can leave it that way or I can return the minimum supported version of node, ATM node 4, although this supporter version afaik refers to ember-cli, and ember itself is not tested against any version of node so I lean towards just not including it.

- The POJO I return is currently memoized and it's not frozen. In its current implementation an addon could grab it and modify it. I'm sitting on the fence wether of not that is a weakness or a strenght. By example, I can imagine ember-fastboot being node 6+ despite of ember-cli being node4+. Also I can imagine ember-electron doing something similar. I'd rather leave it unfrozen for now.

cibernox:
@rwjblue Removed references to node since it was withdrawn from the RFC.

cibernox:
@rwjblue Added to experiments.

rwjblue:
@homu r+

PR #6777, opened 2017-02-15
Better error message when addon doesn't define a name property

briangonzalez:
Spent a good while trying to figure out which addon was failing. PR'ing this in for the next developer ;)

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -221,7 +221,7 @@ let addonProto = {
     p.setupRegistry(this);
 
     if (!this.name) {
-      throw new SilentError('An addon must define a `name` property.');
+      throw new SilentError(`An addon must define a \`name\` property (found at ${this.root}).`);
     }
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@briangonzalez could you try to add a test for this?

lib/models/addon.js
```diff
@@ -221,7 +221,7 @@ let addonProto = {
     p.setupRegistry(this);
 
     if (!this.name) {
-      throw new SilentError('An addon must define a `name` property.');
+      throw new SilentError(`An addon must define a \`name\` property (found at ${this.root}).`);
```
rwjblue: We should probably also try to grab its `package.json`s `name` (if present)

Turbo87: yeah, although that should probably be a different refactoring. but since we're planning such a fallback, testing this PR here might not actually make as much sense. 🤷‍♂️

PR #6778, opened 2017-02-15
Don't print heimdall stack on errors

hjdivad:
Currently this is error-prone.  The stack is useful for understanding error cases but it needs to be made more resilient first.

The error case is as follows

1. build_1
  a.  something in build adds to the stack and then errors, leaving the stack unbalanced.
  b.  stopAndReport(build_1) fails to stop the node.
2. build_2
  a.  start(build_2) now detachs a node it thought was not active, but because of the prior stop failure, it is still active.  When the new build node is added, it is a descendant of the build_1 node.
  b.  something in the build adds to the stacka nd then errors, leaving the stack unbalanced.
  c.  now the stack is detached, an error case.

The better solution here is for stop in 1b. to work even when unbalanced.  What should be done

  1.  stopping nodes autocloses the stack to the current node.
  2.  analysis tools (eg rob's visualizer) check for autoclosed nodes and highlight the nearest un-autoclosed ancestor

### DIFF
--- a/lib/models/instrumentation.js
+++ b/lib/models/instrumentation.js
@@ -248,7 +248,7 @@ class Instrumentation {
     try {
       instr.token.stop();
     } catch (e) {
-      this.ui.writeLine(chalk.red(`Error reporting instrumentation '${name}'.  Stack: ${this._heimdall.stack}`));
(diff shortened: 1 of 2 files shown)


rwjblue:
@homu r+

PR #6789, opened 2017-02-19
Support npm packages as `ember new` blueprints

Turbo87:
Closes #6772 

- [x] tries to install from NPM if a blueprint with a matching name is not found
- [x] only works for `ember new` and `ember addon`
- [x] fails if the npm package does not have the `ember-blueprint` keyword
- [x] hidden behind `NPM_BLUEPRINTS` experiment flag
- [x] has tests 😉

### DIFF
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -8,8 +8,14 @@ const temp = require('temp');
 const path = require('path');
 const merge = require('ember-cli-lodash-subset').merge;
 const execa = require('execa');
+const SilentError = require('silent-error');
+const validateNpmPackageName = require('validate-npm-package-name');
+
+const experiments = require('../experiments');
 
 const logger = require('heimdalljs-logger')('ember-cli:tasks:install-blueprint');
+
+const NOT_FOUND_REGEXP = /npm ERR! 404 {2}'(\S+)' is not in the npm registry/;
 
 // Automatically track and cleanup temp files at exit
 temp.track();
@@ -50,7 +56,8 @@ class InstallBlueprintTask extends Task {
     logger.info(`Resolving blueprint "${name}" ...`);
 
     if (!isGitRepo(name)) {
-      return this._lookupBlueprint(name);
+      return this._lookupBlueprint(name)
+        .catch(error => this._handleLookupFailure(name, error));
     }
 
     return this._createTempFolder()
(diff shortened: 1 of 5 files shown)


lib/tasks/install-blueprint.js
```diff
@@ -80,6 +113,30 @@ class InstallBlueprintTask extends Task {
     return execa('npm', ['install'], { cwd });
   }
 
+  _npmInstallModule(module, cwd) {
+    logger.info(`Running "npm install ${module}" in "${cwd}" ...`);
+    return execa('npm', ['install', module], { cwd })
```
rwjblue: Should use `yarn` if present (or based on the same rules as https://github.com/ember-cli/ember-cli/pull/6748).

lib/tasks/install-blueprint.js
```diff
@@ -80,6 +113,30 @@ class InstallBlueprintTask extends Task {
...
+      .then(() => path.join(cwd, 'node_modules', module));
```
rwjblue: Should we use `path.dirname(resolve('module/package', { baseDir: cwd })`? I'm unsure there are any cases where what you have now is not correct though, feel free to disregard...

lib/tasks/install-blueprint.js
```diff
@@ -80,6 +113,30 @@ class InstallBlueprintTask extends Task {
...
+      let packageName = match[1];
+      throw new SilentError(`The package '${packageName}' was not found in the npm registry.`);
+    }
+
+    throw error;
+  }
+
+  _validateNpmModule(modulePath, packageName) {
+    logger.info(`Checking for "ember-blueprint" keyword in "${packageName}" module ...`);
+    let pkg = require(path.join(modulePath, 'package.json'));
```
rwjblue: `require(path.join(modulePath), 'package')` should be enough (no need to include extension for `require`).

lib/tasks/install-blueprint.js
Turbo87: I don't think it's necessary in this case

lib/tasks/install-blueprint.js
Turbo87: I know, but it's probably faster since the list of extensions won't have to be tried sequentially, right?

lib/tasks/install-blueprint.js
Turbo87: @rwjblue I would like to do something like that in a separate follow-up PR. as you can see a few lines above the existing code of this feature already uses npm exclusively.

lib/tasks/install-blueprint.js
rwjblue: True, though speed probably isn't a concern here.  As I read this again though, I actually think what you have is easier to grok than what I wrote anyways. 

I can see my future self saying "WTF is `${modulePath}/package`?" 😝

lib/tasks/install-blueprint.js
rwjblue: gotcha, seems good, can you either confirm this is encompassed by an existing issue or create a new issue (so that we don't loose track)?

Turbo87: https://github.com/ember-cli/ember-cli/issues/6790

Turbo87:
@homu r+

PR #6795, opened 2017-02-21
Cleanup EmberApp class

Turbo87:


lib/broccoli/ember-app.js
```diff
@@ -388,12 +385,8 @@ class EmberApp {
     @return
   */
   concatFiles(tree, options) {
-    this.project.ui.writeDeprecateLine('EmberApp.concatFiles() is deprecated. Please use the `broccoli-concat` module directly.',
-      arguments[2] === SECRET_DEPRECATION_PREVENTION_SYMBOL);
-
-    options.sourceMapConfig = this.options.sourcemaps;
-
-    return concat(tree, options);
+    this.project.ui.writeDeprecateLine('EmberApp.concatFiles() is deprecated. Please use the `broccoli-concat` module directly.');
```
rwjblue: This unconditionally deprecates calling `this.concatFiles` (and I presume refactors the internals to avoid calling it and instead call `_concatFiles`). We originally still funneled work through `this.concatFiles` because of a ember-cli-fastboot [monkey patch](https://github.com/ember-fastboot/ember-cli-fastboot/blob/ee4bd3b2e32a8b3995a18322ef9fbd25e05047cd/lib/ext/patch-ember-app.js#L20-L25).  I'm happy to remove this, but we need to also update that monkey patch to continue to function after this lands.

FWIW, the fastboot team has a path forward to remove the need for this monkey patch, and @kratiahuja has been working hard on landing. Take a peek at the [meeting summary](https://gist.github.com/kratiahuja/fd073007e10abb9db0a2ec42bc1d7c17)  where we came to consensus...

lib/broccoli/ember-app.js
```diff
@@ -679,23 +674,21 @@ class EmberApp {
     @return {Tree}
   */
   _filterAppTree() {
-    if (this._cachedFilterAppTree) {
-      return this._cachedFilterAppTree;
+    if (!this._cachedFilterAppTree) {
```
rwjblue: Hehe, I preferred the simpler guard + return clause. No real reason to force rightward shift for such a simple thing.

Not blocking though, this is also fine...

lib/broccoli/ember-app.js
```diff
@@ -742,16 +733,16 @@ class EmberApp {
   */
   _configReplacePatterns() {
     return [{
-      match: /\{\{rootURL\}\}/g,
```
rwjblue: The escaping is not needed?

lib/broccoli/ember-app.js
Turbo87: I'm fine with either, but what I definitely don't like is stuff like `return this.cachedSomething = something;` aka. assignments in return statements. I actually thought I enabled the ESLint rule that prevented stuff like that

lib/broccoli/ember-app.js
Turbo87: apparently not, no. my IDE actually told me that it's not needed and it seems to work fine without the escaping.

lib/broccoli/ember-app.js
Turbo87: > and I presume refactors the internals to avoid calling it and instead call _concatFiles

we're actually already calling `_concatFiles()` everywhere internally. the only call to `concatFiles()` is in a unit test that tests that the deprecation message is being printed.

rwjblue: LOL, OK, WHOOPS!

lib/broccoli/ember-app.js
rwjblue: @Turbo87 - Seems good to disable return with assignment (I agree many folks are confused by these types of statements).

lib/broccoli/ember-app.js
rwjblue: Sweet!  Not sure why I would have put them there originally.  😿

rwjblue:
@homu r+

lib/broccoli/ember-app.js
kratiahuja: @rwjblue I guess once this lands we need to update the monkey patch until we ship the new fastboot build scheme right?

rwjblue: @kratiahuja - Yes, but as @Turbo87 pointed out it is already broken (because this PR didn't remove any calls to `this.concatFiles`, meaning that we had already broken the fastboot monkey patch).

Turbo87: I actually don't see how this change would result in different behavior for anyone monkeypatching these methods. It still works the same as before and it still shows the same deprecation message as before.

kratiahuja: Ohio sorry my bad, I didn't see you are still calling concatFiles on the next line. Yeah then it will work.

Issue #6779, opened 2017-02-16
ember try:one exits with zero on test failure


rwjblue:
When running `ember try:one some-scenario` the exit code will be `0` even if there are test failures. I noticed this while digging into an issue in ember-engines (which uses ember-cli/ember-cli#master to make sure it still works properly against master).

Steps to reproduce:

```
ember new exit-code-incorrect-demo
cd exit-code-incorrect-demo
npm install github:ember-cli/ember-cli # to get master installed

# emits "fail" properly
 ember test && echo 'pass' || echo 'fail'

# this should emit "fail" but does not
 ember try:one default && echo 'pass' || echo 'fail'
```

See https://github.com/rwjblue/exit-code-incorrect-demo for an easier to clone demo.

I ran a `git bisect` and identified 8cd5447ac3b71721e84b40d544b0c3bd50e78e8c as the culprit (though I haven't dug into why just yet).

rwjblue:
Did more digging into this the last couple days, my findings so far:

* ember-try uses `process.exit(1)` or `process.exit(0)` to force a desired exit code (see [here](https://github.com/ember-cli/ember-try/blob/master/lib/tasks/try-each.js#L129-L134) and [here](https://github.com/ember-cli/ember-try/blob/master/lib/tasks/try-each.js#L153-L159))
* the promise ember-try does not return an ultimate value from the command's run promise (see [here](https://github.com/ember-cli/ember-try/blob/master/lib/commands/try-one.js#L53-L71) and [here](https://github.com/ember-cli/ember-try/blob/master/lib/tasks/try-each.js#L36-L52))
* adding a return value to the command promise (where the return value is the exit code) and not calling `process.exit` ensures the proper exit code
* it seems calling `process.exit(<anything>)` does not actually change the exit code any longer?

The last few points are somewhat surprising, but returning the exit code is actually a much simpler system than what is being done now anyways so I think I'll refactor to that in ember-try...

ro0gr:
Can't see what exactly breaks 'process.exit()' in the command interruption stuff. Hopefully will take look on the reason in the begining of the March.

rwjblue:
I did more digging ( 👴 🔫 ) and here is more information (I believe I have identified the root cause):

[...]

PR #6796, opened 2017-02-21
Update dependencies to latest versions.

rwjblue:
* Update ember-try min version to 0.2.10.
* Update to diff@3 (removing from greenkeeper ignore).
* Update to configstore@3.
* Update to github@9.

Addresses #6779.
Closes #6788.


rwjblue:
@homu r+

package.json
```diff
@@ -149,7 +149,6 @@
   "trackingCode": "UA-49225444-1",
   "greenkeeper": {
     "ignore": [
-      "diff"
```
Turbo87: @rwjblue should we remove the whole greenkeeper section now?

rwjblue: I thought about it, but decided to leave it so that I remember where the ignores go for next time 😝

PR #6797, opened 2017-02-21
Remove "proxyquire" dependency

Turbo87:
`proxyquire` is hard to use and understand and we can do much better with a much simpler API in many cases


tests/unit/broccoli/merge-trees-test.js
```diff
@@ -4,20 +4,17 @@
...
 
 const MockUI = require('console-ui/mock');
 
-let mergeTreesStub;
-let mergeTrees = proxyquire('../../../lib/broccoli/merge-trees', {
-  'broccoli-merge-trees'() {
-    return mergeTreesStub.apply(this, arguments);
-  },
-});
+let mergeTrees = require('../../../lib/broccoli/merge-trees')
```
rwjblue: missing semi-colon, also we generally use `const` for required things...

Turbo87: whoops... fixed!

Turbo87:
@rwjblue fixed the failing tests. should be ready now.

kellyselden:
The addition of underscored member functions is what I was trying to avoid when adding this. That said, I'm not necessarily against this.

Turbo87:
@homu r+

PR #6799, opened 2017-02-21
Refactor `capture-exit` usage to avoid releasing exit.

### LINKED ISSUE #6779: ember try:one exits with zero on test failure

rwjblue:
Calling `exit.releaseExit()` (possibly _while exiting_) will lead to very difficult to track down bugs.

We should capture exit very early in our processing, and simply not release. There are no downsides to this, and calling `captureExit.releaseExit()` during exit will soon throw an error (see https://github.com/ember-cli/capture-exit/pull/22#issuecomment-281412875).

Fixes #6779

### DIFF
--- a/lib/utilities/will-interrupt-process.js
+++ b/lib/utilities/will-interrupt-process.js
@@ -6,6 +6,7 @@
 // The process finishes after the last interruption handler is settled.
 
 let exit = require('capture-exit');
+exit.captureExit();
 
 let windowsCtrlCTrap;
 let handlers = [];
@@ -83,8 +84,6 @@ function setupSignalsTrap() {
   if (/^win/.test(process.platform)) {
     trapWindowsSignals();
   }
-
-  exit.captureExit();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


rwjblue:
@homu r+

Issue #6769, opened 2017-02-11
bin/ember should remain forwards/backwards compat


stefanpenner:
https://github.com/ember-cli/ember-cli/commit/2809c3412dcde5ede0826b87b8d523470b650dbf introduces features that are not backwards compatible (the ember executable must work for many versions of ember-cli)

Lets move this to `ember-cili/lib/cli/index.js` instead

rwjblue:
@stefanpenner - The scenario where things are mismatched is actually handled (and tested). Did you run into a specific issue or are you just generally concerned?

hjdivad:
@rwjblue the mismatched versions is handled, but as @stefanpenner pointed out to me out of band, the initial require of utils/instrumentation does not handle the case where the bin gets disconnected from its `node_modules`.

rwjblue:
Ok. You may recall we had a specific conversation about this before landing the changes to bin/ember. It should be fine to move the logic into cli/index.

stefanpenner:
> It should be fine to move the logic into cli/index.

Lets do that

rwjblue:
https://github.com/ember-cli/ember-cli/issues/6801 addresses this

locks:
Addressed!

PR #6801, opened 2017-02-21
Move instrumentation out of `bin/ember`.

rwjblue:
As has generally been our preference, we try very very hard to avoid changing things in `bin/ember` that may negatively impact users using it as a launcher for many different ember-cli versions.

This moves the recent changes out of `bin/ember` into `lib/cli/index.js`. The downside is that it is _slightly_ later in the boot process, but the upside is that we get to keep `bin/ember` backwards and forwards compatible for a little while longer...

Fixes https://github.com/ember-cli/ember-cli/issues/6769.

### DIFF
--- a/bin/ember
+++ b/bin/ember
@@ -4,21 +4,6 @@
 // Provide a title to the process in `ps`
 process.title = 'ember';
 
-var instrumentation = require('../lib/utilities/instrumentation');
-
-var instrumentationEnabled = instrumentation.instrumentationEnabled;
-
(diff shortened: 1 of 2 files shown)


stefanpenner:
👍

hjdivad:
@rwjblue  we should be able to get rid of the warning in `lib/models/instrumentation` now.  The only case it was meant to handle was `bin/ember` version mismatches.

rwjblue:
@hjdivad - OK

PR #6802, opened 2017-02-22
Fix minifyCss options (incorrect CSS includes paths)

Exelord:
Hi,
I just noticed a bug according to importing fonts from css files.
I had sth like:
```sass
@font-face
  src: url('fonts/myfont-light.woff2') format('woff2')
```

and `ember-cli` with `minifyCSS` options generated a path `/assets/assets/fonts/myfont-light.woff2` which was inncorect.

I debug this problem it was caused by the latest update of https://github.com/shinnn/broccoli-clean-css where `relativeTo` option was changed to `rebaseTo`.

So, here is a fix for that :)

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -65,7 +65,7 @@ let DEFAULT_CONFIG = {
     },
   },
   minifyCSS: {
-    options: { relativeTo: 'assets' },
+    options: { rebaseTo: 'assets' },
   },
   sourcemaps: {},
   trees: {},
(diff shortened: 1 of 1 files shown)


rwjblue:
ember-cli-preprocessor-registry@3.1.1 was released reverting the update to clean-css@4.

rwjblue:
Closing, as the PR is not needed at this point (and we will likely be going in a different direction WRT how CSS minification is included).

Thank you very much for taking the time to research and submit a fixing PR, sorry about the troubles you had to deal with.

Exelord:
Ok, just remember about it when you will be releasing a new version: https://github.com/ember-cli/ember-cli-preprocessor-registry/issues/21

It will let us save a few hours :)

PR #6804, opened 2017-02-22
Remove missing init instrumentation warning

hjdivad:
This warning no longer makes sense after #6801.

The just-in-time creation of an init node is left in for testing convenience, but not anything else.


hjdivad:
cc @rwjblue

rwjblue:
@homu r+

PR #6805, opened 2017-02-23
more old file cleanup

kellyselden:
more files I noticed still lingering around in projects that have been upgraded from old versions.


rwjblue:
@homu r+

PR #6806, opened 2017-02-24
[ENHANCEMENT] - Adds eslint-plugin-mocha 

### LINKED ISSUE #6760: Use eslint-plugin-mocha
https://github.com/lo1tuma/eslint-plugin-mocha contains quite a few useful rules for Mocha tests. We should probably use it.

Example config: https://github.com/simplabs/eslint-config-simplabs/blob/v0.3.0/plugins/mocha.js

gadogado:
This adds ESLint rules for mocha via https://github.com/lo1tuma/eslint-plugin-mocha.

I used [this](https://github.com/lo1tuma/eslint-plugin-mocha/tree/master/docs/rules) list of rules and the simplelabs [example](https://github.com/simplabs/eslint-config-simplabs/blob/v0.3.0/plugins/mocha.js) as a reference for what might be appropriate; however, I might have missed a rule or configuration so any feedback is appreciated. 

There were also a few eslint errors that came up after adding the rules which have been fixed here.

Resolves #6760


rwjblue:
Adding `mocha/no-exclusive-tests` should allow us to remove `mocha-only-detector` from the project, can you add that to this PR?

rwjblue:
This looks great though, just that one modification and we should be good to go here...

gadogado:
@rwjblue thanks for looking at this.  I removed `mocha-only-detector`

rwjblue:
@homu r+

rwjblue:
Thank you!

Turbo87:
awesome, thanks @gadogado!

PR #6808, opened 2017-02-24
Use `_shouldCompileJS` to guard precompilation of addon JS.

rwjblue:
This makes it simpler to reason about when reviewing the `processedAddonJSFiles` method, and adds an extension point (still private) allowing addons that rely on processing addon JS files but do not have `addon/**/*.js` files (where they are generally shifting things from a node module into their addon path).


rwjblue:
@homu r+

PR #6812, opened 2017-02-27
Support CommonJS loaders

frederikbosch:
Maybe you want to have a RFC for this. I think this is rather a minor change to the existing `shim` feature.

### DIFF
--- a/lib/broccoli/amd-shim.js
+++ b/lib/broccoli/amd-shim.js
@@ -5,13 +5,31 @@ module.exports = function shimAmd(tree, nameMapping) {
   return stew.map(tree, (content, relativePath) => {
     let name = nameMapping[relativePath];
     if (name) {
-      return [
-        '(function(define){\n',
-        content,
-        '\n})((function(){ function newDefine(){ var args = Array.prototype.slice.call(arguments); args.unshift("',
-        name,
-        '"); return define.apply(null, args); }; newDefine.amd = true; return newDefine; })());',
-      ].join('');
+      return `
+(function(define, module){
+
+${content}
+
+  if (define.defined === false && module.exports) { 
+    define([], function () { 
+      return module.exports; 
+    });
+  }
+})((function(){ 
+  function newDefine(){ 
+    var args = Array.prototype.slice.call(arguments);
+    if (typeof args[0] === "string") { args.shift(); }
+    args.unshift("${name}");
+    define.defined = true; 
+    return define.apply(null, args);
+  }; 
+
+  newDefine.amd = true; 
+  newDefine.defined = false; 
+
+  return newDefine;
+})(), {});
+`;
     } else {
       return content;
     }
--- a/tests/acceptance/brocfile-smoke-test-slow.js
+++ b/tests/acceptance/brocfile-smoke-test-slow.js
@@ -297,6 +297,30 @@ describe('Acceptance: brocfile-smoke-test', function() {
     })();
   }));
 
+  it('can turn CommonJS loader into named AMD', co.wrap(function *() {
+    yield copyFixtureFiles('brocfile-tests/app-import-common-js');
+    yield runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'build');
+
+    let outputJS = fs.readFileSync(path.join(appRoot, 'dist', 'assets', 'output.js'), {
+      encoding: 'utf8',
+    });
+
+    (function() {
+      let defineCount = 0;
+      // eslint-disable-next-line no-unused-vars
+      function define(name, deps, factory) {
+        expect(name).to.equal('hello-world');
+        expect(deps).to.deep.equal([]);
+        expect(factory()()).to.equal('Hello World');
+        defineCount++;
+      }
+      /* eslint-disable no-eval */
+      eval(outputJS);
+      /* eslint-enable no-eval */
+      expect(defineCount).to.eql(1);
+    })();
+  }));
+
   // skipped because of potentially broken assertion that should be fixed correctly at a later point
   it.skip('specifying partial `outputPaths` hash deep merges options correctly', co.wrap(function *() {
     yield copyFixtureFiles('brocfile-tests/custom-output-paths');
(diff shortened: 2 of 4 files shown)


rwjblue:
Seems OK to me, but I'd like to see a test to confirm things don't regress...

frederikbosch:
@rwjblue You want me to execute such a test? What would you expect of such a test?

stefanpenner:
This seems reasonable, as @rwjblue mentioned we should get this under test as the details are nuanced. 

@frederikbosch i think the easiest way to test this, is to eval the code in a test and make sure it does the right things with the environment it expects.

`new Function(wrapperCode, ....`
* for the various different `define` `module` global states this aims to add support for.

frederikbosch:
@stefanpenner Will do.

frederikbosch:
@stefanpenner There were already tests for the anonymous AMD loader. I added a test for the `module.exports` syntax.

lib/broccoli/amd-shim.js
```diff
@@ -6,11 +6,14 @@ module.exports = function shimAmd(tree, nameMapping) {
     let name = nameMapping[relativePath];
     if (name) {
       return [
-        '(function(define){\n',
+        '(function(define, module){\n',
```
rwjblue: Sorry to be overly nit-picky here, but it would be _wonderful_ if we could clean this up with a bit of ES6 syntax.

* Use template strings with interpolation for `content`
* spread out for readability

Might look like:

```js
return `
(function(define, module){

${content}

  if (define.defined === false && module.exports) { 
    define([], function () { 
      return module.exports; 
    });
  }
})((function(){ 
  function newDefine(){ 
    var args = Array.prototype.slice.call(arguments);
    while (typeof args[0] === "string") { args.shift(); }
    args.unshift("${name}");
    define.defined = true; 
    return define.apply(null, args);
  }; 

  newDefine.amd = true; 
  newDefine.defined = false; 

  return newDefine;
})(), {});
`
```

rwjblue: In rewriting this for the last comment, I am slightly confused about this part:
```
while (typeof args[0] === "string") { args.shift(); }
```

What is this trying to do?  Why would `args[0]` (or an arbitrary number of initial arguments) be strings?

frederikbosch: @rwjblue Thanks for your feedback. I agree. The readability will increase when using templates. I will update my PR.

frederikbosch: @rwjblue Regarding shifting the first argument when it is a string: libraries may [define their own name and use it as first argument](https://github.com/jquery/jquery/blob/3.1.1/dist/jquery.js#L10180). This will be conflicting with our wrapping factory, because the result will be `define("jquery", "jquery", factory);`.

rwjblue: Right, but in that case it is only the first arg (not an arbitrary number of leading arguments), so I don't think we need a loop there...

frederikbosch: @rwjblue Right, I will update that too!

frederikbosch:
Updated the PR according the suggest changes by @rwjblue.

frederikbosch:
@rwjblue 

> What happens if the module uses require? Do we need to do anything special?

Nothing changes with respect to `require`. The only thing this PR does is the following.

- It [injects the `module` variable into the scope of module](https://github.com/frederikbosch/ember-cli/blob/patch-1/lib/broccoli/amd-shim.js#L9).
- It checks if the module [did not use `define` already](https://github.com/frederikbosch/ember-cli/blob/patch-1/lib/broccoli/amd-shim.js#L13)
- It checks if the module [used `module.exports`](https://github.com/frederikbosch/ember-cli/blob/patch-1/lib/broccoli/amd-shim.js#L13)
- It uses [`define` to register the module](https://github.com/frederikbosch/ember-cli/blob/patch-1/lib/broccoli/amd-shim.js#L13) via loader.js

As example, please see [this library](https://github.com/bfred-it/object-fit-images/blob/master/package.json#L33). Importing this requires an addon without this PR. With this PR, it works out of the box. I am sure there are more examples.

rwjblue:
OK, this is looking great, thank you for updating and answering my questions above.

The only remaining question I have is determining if we should overload the `transformations: 'amd'` bit to support cjs like this, or if we should make a `transformations: 'cjs'` that does this instead...

frederikbosch:
@rwjblue I don't think so because we do not present the user with a new option. This PR gives supports to CommonJS `module.exports` syntax, by factorizing it to AMD. So in the end, we are still using AMD. Hence, we do not need `transformations: 'cjs'`.

rwjblue:
I think I disagree 😺. If you read through [the original RFC for adding `transformations`](https://github.com/ember-cli/rfcs/blob/master/active/0055-anonymous-amd.md) the idea is that you are transforming _from_ `amd` (not to AMD though that is the current implementation).

rwjblue:
@stefanpenner / @Turbo87 - Thoughts?

frederikbosch:
@rwjblue You are right, it does not conform that RFC. However, should we give the user a decision when there actually is no decision?

rwjblue:
Not really sure what you mean.  The only "actor" in the system that has a chance of knowing what to do is the end user.  If a given library is authored to call `define(['dep', 'whatever'], function() {})` or `module.exports = {}` is totally something they need to tell us.

In theory, I am fine with a `transformations: 'detect'` or somesuch that does what is done now, but `transformations: 'amd'` processing  CJS inputs just seems odd/wrong to me...

frederikbosch:
Good point, that also requires code changes. I am happy to go for the following solution.

1. `detect` which uses the transformation shim of this PR
2. `amd` which uses the transformation shim before PR
3. `cjs` which uses the transformation shim of PR, but then only with `module.exports` factory

rwjblue:
I just published [ember-cli-cjs-transform](https://github.com/rwjblue/ember-cli-cjs-transform) which _should_ support this use case now (and leverages the APIs added for [ember-cli/rfcs#55](https://github.com/ember-cli/rfcs/blob/master/active/0055-anonymous-amd.md)). 

I'm going to close this for now, but please check it out and report any issues!

PR #6816, opened 2017-03-02
Resolve path when calling UnwatchedDir for Bower

arthirm:
The path passed to UnWatchedDir for Bower directory case was not resolved. Fixing that in this commit.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -195,7 +195,7 @@ class EmberApp {
     let templatesTree = existsSync(templatesPath) ? new UnwatchedDir(templatesPath) : null;
 
     // do not watch bower's default directory by default
-    let bowerTree = this.project._watchmanInfo.enabled ? this.bowerDirectory : new UnwatchedDir(this.bowerDirectory);
+    let bowerTree = this.project._watchmanInfo.enabled ? this.bowerDirectory : new UnwatchedDir(this._resolveLocal(this.bowerDirectory));
 
     let vendorPath = this._resolveLocal('vendor');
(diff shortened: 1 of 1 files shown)


arthirm:
@rwjblue @hjdivad @stefanpenner

lib/broccoli/ember-app.js
```diff
@@ -195,7 +195,7 @@ class EmberApp {
     let templatesTree = existsSync(templatesPath) ? new UnwatchedDir(templatesPath) : null;
 
     // do not watch bower's default directory by default
-    let bowerTree = this.project._watchmanInfo.enabled ? this.bowerDirectory : new UnwatchedDir(this.bowerDirectory);
+    let bowerTree = this.project._watchmanInfo.enabled ? this.bowerDirectory : new UnwatchedDir(this._resolveLocal(this.bowerDirectory));
```
rwjblue: This will only resolve when watchman is not present/enabled.  We need to resolve it either way.

I think we want something like this will work:

```js
let bowerDirectory = this._resolveLocal(this.bowerDirectory);
let bowerTree = this.project._watchmanInfo.enabled ? bowerDirectory : new UnwatchedDir(bowerDirectory);
```

Turbo87:
@arthirm why do you need to resolve it? are you sure that it should be resolved? and if so, why?

rwjblue:
Yeah, it should be resolved (just like all the other paths that we pass to UnwatchedDir / WatchedDir). This is to support having bower_components be resolved relative to the project root instead of `process.cwd()`.

rwjblue:
@homu r+

PR #6822, opened 2017-03-02
WIP: Transpile async/await in tests via Babel

Turbo87:
It seems to work quite well. Do we want this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "yam": "0.0.22"
   },
   "devDependencies": {
+    "babel-register": "^6.23.0",
     "broccoli-filter": "^1.2.2",
     "broccoli-plugin": "^1.2.0",
     "broccoli-test-helper": "^1.0.0",
@@ -132,6 +133,7 @@
     "ember-cli-internal-test-helpers": "^0.9.0",
     "eslint-plugin-chai-expect": "^1.1.1",
     "eslint-plugin-mocha": "^4.8.0",
+    "fast-async": "^6.2.1",
     "github": "^9.0.0",
     "istanbul": "^0.4.2",
     "mocha": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "yam": "0.0.22"
   },
   "devDependencies": {
+    "babel-preset-env": "^1.1.11",
     "babel-register": "^6.23.0",
     "broccoli-filter": "^1.2.2",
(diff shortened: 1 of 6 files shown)


rwjblue:
I'm on the fence.  Based on the way you have it setup to prevent transpilation from leaking into `lib/` code during out test suite, I think this is probably OK. However, we have generally been opposed to any/all transpilation so it is a fairly large departure.

We should definitely discuss at the CLI core team call...

stefanpenner:
> Based on the way you have it setup to prevent transpilation from leaking into lib/ code during out test suite, I think this is probably OK

^ is important, thanks.

---

Out of curiosity how much transpilation time does this add?

Turbo87:
@stefanpenner the overhead on my machine is barely noticeable compared to the time that we spend in our testing code itself

rwjblue:
I _think_ he meant the transpiled output (e.g. when stepping through the code during a debugging session). At least that is what I am also curious about 😜

stefanpenner:
> I am also wondering how bad the transpiled output is (e.g. when stepping through the code during a debugging session).

I am an outlier, i actually usually prefer the giant switch statement over callbacks :P...

Turbo87:
I've adjusted the PR to use https://github.com/babel/babel-preset-env now which has some up- and downsides:

- ➕ it only transpiles async/await if Node version is below 7.7.1 (or whatever introduced native async/await)
- ➖ it transpiles all other ESsomething features too unless we use a large blacklist
- ➖ it has a lot more dependencies (all potentially needed plugins)

please make sure to try out both approaches. it should also be possible to adjust the first approach (with fast-async) to only enable babel transpilation if Node is below a certain version.

Turbo87:
@ember-cli/core has anyone tried this?

stefanpenner:
@Turbo87 I have not, but can. Anything specific I should look at, or mostly just around debugging?

Turbo87:
> Anything specific I should look at

just the normal development workflows that you have. whether anything is more complicated or actually easier then before using this approach.

and as mentioned in https://github.com/ember-cli/ember-cli/pull/6822#issuecomment-283804129 I'm no longer certain if we should use `babel-preset-env` or if we should just use the `fast-async` plugin and disable Babel transpilation completely if the local Node version supports native async/await.

rwjblue:
Honestly, I'm happier with the `co` setup than with transpiling, so I think we should just close this...

Feel free to reopen if you disagree...

PR #6824, opened 2017-03-02
chore(package): update broccoli-merge-trees to version 1.2.3

kellyselden:
Is this good practice? To eliminate the potential of someone running 1.2.2 and hitting the error?


kellyselden:
Looks like v1.2.1 is failing now because of dependency version drift https://travis-ci.org/ember-cli/ember-cli/builds/207121932

We should probably merge this then?

kellyselden:
@homu r+

PR #6828, opened 2017-03-06
Use Babel 6

rwjblue:
This is the starting point for using babel@6 in ember-cli itself and apps. This mostly includes various fallback paths that we have provided that ensure "things are transpiled", without much guarantee about actual transpilation content.

It also paves the way for allowing the usage of ember-cli-babel@6 by providing `this.options.babel` as needed based on ember-cli-babel version in use.

Utilizes the new `project.targets` to provide custom transpilation configuration based on project targets.

---

- [ ] Release final stable version of ember-cli-babel@6 (https://github.com/babel/ember-cli-babel/pull/115)
- [ ] Release final stable version of ember-cli-htmlbars-inline-precompile for babel@6 support

Fixes https://github.com/ember-cli/ember-cli/issues/5015.


package.json
```diff
@@ -37,9 +37,11 @@
   },
   "dependencies": {
     "amd-name-resolver": "0.0.6",
+    "babel-plugin-transform-es2015-modules-amd": "^6.22.0",
+    "babel-preset-env": "^1.2.0",
     "bower-config": "^1.3.0",
     "bower-endpoint-parser": "0.2.2",
-    "broccoli-babel-transpiler": "^5.6.2",
+    "broccoli-babel-transpiler": "6.0.0-alpha.3",
```
rwjblue: This is the primary "merge blocker".  Need to release as something > .alpha to land (IMO).

lib/utilities/default-targets.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+module.exports = {
+  browsers: [
+    'ie 9',
+    'last 1 Chrome versions',
```
cibernox: This is totally outside the scope but here we go:
This matrix is the actual matrix of browsers in which ember itself is tedted. I believe that probably safari 9 should be added. Arguably today iOS9 is far more relevant than ie9 & 10 combined.

rwjblue: No argument there. In this PR I'm just taking advantage of the new targets, so I refactored to a separate util to avoid duplication.

lib/utilities/babel-config-for-targets.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+const presetEnv = require('babel-preset-env').default;
```
cibernox: Should the babel-preset-env live here or can it be part ember-cli-babel?

rwjblue: The idea is that ember-cli-babel is "just" transpiling with the provided config options. I am not really opposed to flip flopping if folks feel strongly...

rwjblue: This was good advice, moved this logic to ember-cli-babel (in https://github.com/babel/ember-cli-babel/commit/78b0342f1cf0e4aae3b22dfcde10d04bce74b253).

lib/broccoli/babel-process-modules-only.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+module.exports = function processModulesOnly(tree, annotation) {
+  let options = {
+    plugins: [
+      // this sucks, but :goodnews: once a new version of babel@6 is released
+      // we can remove this in favor of `babel-plugin-transform-es2015-modules-amd`
+      [require('rwjblue-custom-babel-6-amd-modules-no-interop'), { loose: true, noIterop: true }],
```
rwjblue: clearly not ideal, but seems ok so that we aren't blocked hard on a release...

rwjblue:
Things that are done:

* update the app blueprint to use ember-cli-babel@6 (alpha for now)
* change all fallback paths to only process modules (this was the only guarantee anyways)
* ensure the default babel options are populated based on ember-cli-babel version
* ensure ember-cli-babel@6 can be used with ember-cli@2.11.1  and ember-cli@2.12 (likely older though I haven't tested yet)
* ensure ember-cli-htmlbars-inline-precompiler@0.4.0-beta.1 (babel@6 version) is working properly in conjunction with ember-cli@2.11 and ember-cli@2.12
* published custom build of babel@6 (from babel's master branch) that includes the changes needed to emit modules that can interop between babel@5 and babel@6

This is ready to start reviewing (code wise), I'm still working through a couple test failures ATM though...

lib/broccoli/babel-process-modules-only.js
stefanpenner: cc @hzoo

rwjblue:
@homu r+

Issue #6819, opened 2017-03-02
Allow using styles from `src/ui/app.css`


Turbo87:
For the implementation of the [Module Unification RFC](https://github.com/emberjs/rfcs/pull/143) we need to change the `EmberApp` class to allow the `app.css` file to be placed in the `src/ui/styles/` folder instead of `app/styles/`.

see also https://github.com/emberjs/ember.js/issues/14882

PR #6825, opened 2017-03-03
EmberApp: Use "src/ui/index.html" if it exists

Turbo87:
Resolves #6818 

/cc @mixonic

PR #6835, opened 2017-03-07
EmberApp: Use "src/ui/styles" over "app/styles" if it exists

Turbo87:
This should be seen as a short-term enhancement, but that whole part of the code is in desperate need of a cleanup... 😟 

Resolves https://github.com/ember-cli/ember-cli/issues/6819

/cc @mixonic

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -193,7 +193,10 @@ class EmberApp {
 
     // these are contained within app/ no need to watch again
     // (we should probably have the builder or the watcher dedup though)
-    let stylesTree = new UnwatchedDir(this._resolveLocal('app/styles'));
+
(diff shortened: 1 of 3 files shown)


mixonic:
this is sufficient for my immediate needs. 👍  if @rwjblue or someone on ember-cli is also in agreement it is a good path for now.

stefanpenner:
I believe all the `modules` RFC stuff should behind an ember-cli feature flag.

rwjblue:
I think the only blocker is @stefanpenner's comment RE: flagging...

Turbo87:
@stefanpenner @rwjblue updated the code to use a feature flag. I've also adjusted the code introduced in https://github.com/ember-cli/ember-cli/issues/6825 to only be enabled via the same feature flag.

Turbo87:
@homu r+

PR #6839, opened 2017-03-08
Allow `ember new -b <blueprint> foo` to opt-in to yarn by default.

rwjblue:
This change removes the hardcoded default of `--yarn=false`, but does not change the way `ember new foo` or `ember addon foo` work with regards to `npm` vs `yarn`. The change simply allows a blueprint to include a `yarn.lock` to "hint" to our `NPMTask` that yarn should be used if possible.


rwjblue:
@homu r+

PR #6840, opened 2017-03-08
Add logging for `this.runTask` within commands.

rwjblue:
This makes tracking down errors thrown from commands that manage multiple tasks much easier.

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -228,6 +228,8 @@ let Command = CoreObject.extend({
       throw new Error(`Concurrent tasks are not supported`);
     }
 
+    logger.info(`\`${this.name}\` command running \`${name}\` task`);
+
     let Task = this.tasks[name];
     if (!Task) {
       throw new Error(`Unknown task "${name}"`);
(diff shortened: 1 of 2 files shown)


lib/models/command.js
```diff
@@ -237,9 +239,15 @@ let Command = CoreObject.extend({
 
     this._currentTask = task;
 
-    return Promise.resolve(task.run(options)).finally(() => {
-      delete this._currentTask;
-    });
+    return Promise.resolve(task.run(options))
+      .catch(error => {
+        logger.info(`An error occurred running \`${name}\` from the \`${this.name}\` command.`, error.stack);
```
stefanpenner: I'm surprised the existing error handler doesn't provide this?

rwjblue: Which existing error handler are you referring to?

stefanpenner: one of these guys? https://github.com/ember-cli/ember-cli/blob/de200a2dd57cba1542891fbb53672e66c062f864/lib/cli/cli.js#L265-L274

rwjblue: Ya. Those only show the original stack of the error. While spelunking a fairly annoying bug with @tomdale we had a situation where the stack was completely in native code land and nothing pointed to the actual command / task that was causing the error. 

In our case, this was due to native promises suffering from similar issues as we had with RSVP waaaay back (swallowed stacks and what not). It's likely that this will be corrected, but in the meantime I figured we could just log the error stack and mention which command / task triggered it.

stefanpenner: ok

rwjblue:
@homu r+

PR #6841, opened 2017-03-09
Update yarn.lock.

rwjblue:
* `yarn upgrade`
* Update `yarn.lock` for greenkeeper update of broccoli-merge-trees.


kellyselden:
Shoot. This is hard for me to remember. Does it hurt much to be out of sync like this? I hope GK implements lock file updating soon.

rwjblue:
No, in general it doesn't seem to be a massive problem to be out of sync. I only PR'ed because I ran `yarn` locally and ended up with uncommitted changes to yarn.lock.

PR #6842, opened 2017-03-09
Merge current environment vars before calling git commit

tomdale:
In at least Ember CLI 2.11, the call to `git commit` in the `GitInitTask` was silently failing if any of your git config files contained a path with a `~`. That's because we were overriding the child git process's environment, including the `$HOME` variable required to do path expansion for `~`.

In practice, this means that many people running `ember new my-app` today are getting apps that have all of the files but without the friendly default commit.

As of at Ember CLI 2.12, the refactoring of the app init process caused this error to be raised explicitly, which is how I discovered it. This PR changes the `exec` call to include the parent process's environment, including `$HOME`, so path expansion can happen normally and the commit is created successfully.

Big thanks to @rwjblue who helped me debug this and author the patch.


rwjblue:
:+1: I'll merge into beta branch (instead of master) once CI is green.

stefanpenner:
good find

rwjblue:
I borked things up while changing target of this PR so I merged manually into beta and master.

PR #6843, opened 2017-03-10
Use "co" to run async functions

Turbo87:
This is an alternative to https://github.com/ember-cli/ember-cli/pull/6822 which does not require any transpiling on Node 4+

If this works well we might also consider using it outside the testing code too.

see https://github.com/tj/co


stefanpenner:
Lgtm

cibernox:
fwiw, I also like it. No transpilation, almost identical advantages, easy to find-and-replace later.

PR #6847, opened 2017-03-10
print the `serving on http://host:port/basePath` after each rebuild

stefanpenner:
this has literally been killing me since the start of time.

- [x] fix tests

### DIFF
--- a/lib/models/watcher.js
+++ b/lib/models/watcher.js
@@ -3,6 +3,7 @@
 const chalk = require('chalk');
 let logger = require('heimdalljs-logger')('ember-cli:watcher');
 const CoreObject = require('core-object');
+const cleanBaseURL = require('clean-base-url');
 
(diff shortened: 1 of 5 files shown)


Turbo87:
hmm... if you have a lot of build output then I understand the need to see that line *after* the build. I could however also imagine repeating those lines over and over being a little annoying... 🤔

rwjblue:
I'm also in favor of this generally speaking.  I can't tell you how many times I have forgotten the port I was serving a given app on.

stefanpenner:
I could put it on the same line as `build success`

```
build sucessful, serving on: http://localhost:4200
```

stefanpenner:
screen shot 2017-03-10 at 3 19 37 pm

rwjblue:
@homu r+

PR #6851, opened 2017-03-12
Use "execa" in "git-init" task

Turbo87:


rwjblue:
LGTM, I restarted one of the travis jobs (it was failing due to random timeout).

PR #6862, opened 2017-03-16
Update minimum ember-try version.

rwjblue:
Fixes issues when using scenarios that have bower dependencies,
when the host app doesn't initially include a `bower.json`.


rwjblue:
@homu r+

PR #6865, opened 2017-03-17
Use `yarn install --no-lockfile` in travis for addons

### LINKED ISSUE #6844: Use `yarn install --no-lockfile` for CI in addon blueprints
This will make sure that tests are always run with the latest matching dependencies that addon users would also install.

https://github.com/ember-cli/ember-cli/blob/1c3c2c2d3a3c123522a5376e0d52d24d25f0404d/blueprints/addon/files/.travis.yml#L35

rwwagner90:
I'm not sure if this was as simple as the issue description made it sound, but it sounded like just changing the one line from `yarn install` to `yarn install --no-lockfile`, so I did that. Let me know if extra work is required and I'll be happy to update the PR.

Fixes #6844

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -32,7 +32,7 @@ matrix:
   - phantomjs --version
 
 install:
-  - yarn install
+  - yarn install --no-lockfile
 <% } else { %>before_install:
   - npm config set spin false
   - npm install -g phantomjs-prebuilt


stefanpenner:
LGTM, @Turbo87 r?

Turbo87:
@rwjblue @stefanpenner do you agree with cherry picking this on `beta` that introduced yarn support?

rwjblue:
Ya, seems good.

PR #6868, opened 2017-03-18
Add --welcome option to `new` and `init` so that it can be skipped with --no-welcome

### LINKED ISSUE #6444: Add `--no-welcome` option or similar to `new` & `init`
While `ember-welcome-page` is great for first experience, the first thing I do after generating a new app is remove it. It would be nice to be able to not install it in the first place.

romulomachado:
Closes #6444

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -93,6 +93,7 @@ module.exports = {
       emberCLIVersion: require('../../package').version,
       year: date.getFullYear(),
       yarn: options.yarn,
+      welcome: options.welcome,
     };
   },
 
--- a/lib/commands/new.js
+++ b/lib/commands/new.js
@@ -19,14 +19,15 @@ module.exports = Command.extend({
   works: 'outsideProject',
 
   availableOptions: [
-    { name: 'dry-run',    type: Boolean, default: false, aliases: ['d'] },
-    { name: 'verbose',    type: Boolean, default: false, aliases: ['v'] },
-    { name: 'blueprint',  type: String,  default: 'app', aliases: ['b'] },
-    { name: 'skip-npm',   type: Boolean, default: false, aliases: ['sn'] },
-    { name: 'skip-bower', type: Boolean, default: false, aliases: ['sb'] },
-    { name: 'skip-git',   type: Boolean, default: false, aliases: ['sg'] },
-    { name: 'yarn',       type: Boolean },
-    { name: 'directory',  type: String,                  aliases: ['dir'] },
+    { name: 'dry-run',     type: Boolean, default: false, aliases: ['d'] },
+    { name: 'verbose',     type: Boolean, default: false, aliases: ['v'] },
+    { name: 'blueprint',   type: String,  default: 'app', aliases: ['b'] },
+    { name: 'skip-npm',    type: Boolean, default: false, aliases: ['sn'] },
+    { name: 'skip-bower',  type: Boolean, default: false, aliases: ['sb'] },
+    { name: 'skip-git',    type: Boolean, default: false, aliases: ['sg'] },
+    { name: 'welcome',     type: Boolean, default: true },
+    { name: 'yarn',        type: Boolean },
+    { name: 'directory',   type: String,                  aliases: ['dir'] },
   ],
 
   anonymousOptions: [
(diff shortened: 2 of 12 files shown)


romulomachado:
Hey @Turbo87, thanks for the suggestions!

I tried sending the `welcome` option to `locals()` before, but it was putting the flag into the file instead of evaluating the `if`, I was missing the `{`, the example you provided was very helpful.

Thanks again for helping!

I'm working on fixing the tests now.

romulomachado:
@Turbo87 ✅  Just added the tests! LMK if there's anything else I should do.

romulomachado:
Just realized it's not very friendly to have only the `{{outlet}}` on the `template.hbs` if `--no-welcome` is true so I added the old `<h2 id="title">Welcome to Ember</h2>` back.

blueprints/addon/index.js
```diff
@@ -84,6 +84,8 @@ module.exports = {
     let addonName = stringUtil.dasherize(addonRawName);
     let addonNamespace = stringUtil.classify(addonRawName);
 
+    let { yarn, welcome } = options;
```
Turbo87: I think destructuring is not yet supported on Node 4

tests/acceptance/new-test.js
```diff
@@ -331,4 +331,30 @@ describe('Acceptance: ember new', function() {
...
+      'new',
+      'foo',
+      '--skip-npm',
+      '--skip-bower',
+      '--skip-git',
+      '--no-welcome',
+    ]);
+
+    expect(file('package.json'))
+      .not.to.match(/"ember-welcome-page"/);
```
Turbo87: you could check for "Welcome to Ember" here too

lib/commands/new.js
```diff
@@ -19,14 +19,15 @@ module.exports = Command.extend({
...
-    { name: 'skip-git',   type: Boolean, default: false, aliases: ['sg'] },
-    { name: 'yarn',       type: Boolean },
-    { name: 'directory',  type: String,                  aliases: ['dir'] },
+    { name: 'dry-run',     type: Boolean, default: false, aliases: ['d'] },
+    { name: 'verbose',     type: Boolean, default: false, aliases: ['v'] },
+    { name: 'blueprint',   type: String,  default: 'app', aliases: ['b'] },
+    { name: 'skip-npm',    type: Boolean, default: false, aliases: ['sn'] },
+    { name: 'skip-bower',  type: Boolean, default: false, aliases: ['sb'] },
+    { name: 'skip-git',    type: Boolean, default: false, aliases: ['sg'] },
+    { name: 'welcome',     type: Boolean, default: true },
```
Turbo87: since `welcome` by itself isn't that descriptive it would probably make sense to add a `description` here too

tests/acceptance/new-test.js
```diff
@@ -331,4 +331,36 @@ describe('Acceptance: ember new', function() {
...
+      '--skip-npm',
+      '--skip-bower',
+      '--skip-git',
+    ]);
+
+    expect(file('package.json'))
+      .to.match(/"ember-welcome-page"/);
+
+    expect(file('app/templates/application.hbs'))
+      .to.contain(/"{{welcome-page}}"/);
```
Turbo87: I *think* the issue is that you wrote `/"{{welcome-page}}"/` instead of `/{{welcome-page}}/`. Could you try this without the quotes? or just use `expect(...).to.contain('{{welcome-page}}');` instead.

romulomachado: @Turbo87 That was it! Thank you, just updated it!

romulomachado:
@Turbo87 Just rebased the branch (went from 18 commits to 5).

not bad

Turbo87:
there seems to be one failing test left, but I'm not sure why that is actually failing... 🤔 

any ideas?

romulomachado:
@Turbo87 Had a busy week last week. Hopefully, I'll have some time to fix that failing test this week. I have no idea why it's failing, but I'll investigate.

kellyselden:
I'm late to the party, and this idea can be a separate PR, but it seems like if we had a good extendable commands story, the welcome addon could just extend the new command and add it's option, and a hook system would allow it to no-op on that option.

Turbo87:
@romulomachado could you rebase the PR on top of the `master` branch?

romulomachado:
@Turbo87 ✅  Done!

rwjblue:
@homu r+

PR #6884, opened 2017-03-21
Remove guard in `treeForAddon` around `addon/**/*.js` files.

rwjblue:
This guard was added to avoid wasted effort when preprocessing the `addon/` directory when it only contains a single `.gitkeep` file (or is non-existent). The assumption I made was that if there are no `*.js` files inside the `addon/` directory, then we didn't need to attempt to transpile it.

Unfortunately, "back in reality" there are a number of addons that wrap existing libraries, that have no `addon/**/*.js` files but still need the `this._super.treeForAddon` invocation to process the provided tree.

This change, reverts the guard (it was only a relatively small savings anyways).

The diff is best viewed [without whitespace changes](https://github.com/ember-cli/ember-cli/pull/6884/files?w=1)...

Related issues:

* https://github.com/ember-redux/ember-redux/issues/105
* https://github.com/mike-north/ember-lodash/issues/89
* https://github.com/ciena-blueplanet/ember-dagre/issues/77

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -822,20 +822,6 @@ let addonProto = {
     return this._fileSystemInfo().hasPodTemplates;
   },
 
-  /**
-     Looks in the addon/ folder to determine if JS files exist that need to be
-     precompiled.
-
-     This is executed once during initial build but not on rebuilds.
-
-     @private
-     @method _shouldCompileJS
-     @return {Boolean} indicates if JS files exist and need to be compiled for this addon
-  */
-  _shouldCompileJS() {
-    return this._fileSystemInfo().hasJSFiles;
-  },
-
   _fileSystemInfo() {
     if (this._cachedFileSystemInfo) {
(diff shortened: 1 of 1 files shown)


rwjblue:
Note: this targets `release` branch, since the underlying issue exists for users of ember-cli@2.12

rwjblue:
@homu r+

toranb:
@rwjblue will this land in a patch release for ember-cli v2.12?

toranb:
@rwjblue mostly asking because we have a handful of PRs out for shims to implement `_shouldCompileJS` and I'm thinking to myself ... "do we really need to implement this?"

rwjblue:
@toranb - Yes, as mentioned in https://github.com/ember-cli/ember-cli/pull/6884#issuecomment-288209341, this PR targets 2.12 patch release.

toranb:
@rwjblue perfect :) thanks for the double clarification

Issue #6863, opened 2017-03-16
Command failed: npm install bower@^1.3.12


Turbo87:
#6631 introduces lazy bower installation, which seemed to work fine for everyone that tested it. not that this PR has hit the public we are seeing some people reporting issues that this feature is not working as intendend:

```
NPM: Installed dependencies
Command failed: npm install bower@^1.3.12
(node:10767) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
npm ERR! Darwin 15.6.0
npm ERR! argv "/usr/local/bin/node" "/Users/pc01835/Development/ember/docroot_polar_vision/trunk/pv-styleguide/node_modules/.bin/npm" "install" "bower@^1.3.12"
npm ERR! node v6.10.0
npm ERR! npm  v3.5.4

npm ERR! Cannot read property 'target' of null
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/pc01835/Development/ember/docroot_polar_vision/trunk/pv-styleguide/node_modules/ember-cli/npm-debug.log
```

this is actually a bug in npm, which got fixed in v3.8.3 according to the meta data in https://github.com/npm/npm/commit/473d3244a8ddfd6b260d0aa0d395b119d595bf97

right now we do not check the npm version when using npm to lazily install bower, but we should likely do that to make sure that at least v3.8.3 is used.

related #6790 

/cc @rstudner

Turbo87:
Resolved by https://github.com/ember-cli/ember-cli/pull/6885

PR #6885, opened 2017-03-21
Work around broken bower installation for old npm versions

Turbo87:
This PR:

- checks the error message returned from `npm install bower` and displays a helpful error message if the known bug was hit

- checks if the `bower.json` actually has any dependencies before trying to `npm install bower` so that the first case should become a rare edge case when used with a custom app blueprint that still includes bower dependencies

Resolves https://github.com/ember-cli/ember-cli/issues/6863


rwjblue:
@homu r+

PR #6886, opened 2017-03-21
Handle addon constructor errors gracefully

### LINKED ISSUE #6820: Cryptic error running `ember` when package has `ember-addon` keyword without being a full addon
I have an Ember utility package that I tagged with the `ember-addon` keyword for it to have EmberObserver visibility, but it is not a full addon. The package is `ember-cli-build-config-editor`. It has a similar function and role as `ember-router-generator` as a utility that supports blueprints. This causes `ember` to crash with a cryptic error.

## Recipe

* `ember new <name>`
* `cd <name>`
* Edit `package.json` and add `"ember-cli-build-config-editor": "0.1.1",` to the `devDependencies`
* `ember`

You will see the following output:

```
did not recognize object of type "range"
Error: did not recognize object of type "range"
    at Object.getFieldNames (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/types.js:663:19)
    at visitChildren (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:223:36)
    at Visitor.PVp.visitWithoutReset (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:204:20)
    at NodePath.each (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path.js:101:26)
    at visitChildren (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:219:18)
    at Visitor.PVp.visitWithoutReset (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:204:20)
    at NodePath.each (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path.js:101:26)
    at visitChildren (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:219:18)
    at Visitor.PVp.visitWithoutReset (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:204:20)
    at visitChildren (/Users/srvance/tests/e-c-addon-bug/node_modules/ast-types/lib/path-visitor.js:246:25)
```

I believe this is because `ember` scans the `node_modules` for things with the `ember-addon` keyword but does not verify that the package fulfills the addon contract before trying to invoke it.

## Workarounds

Two workarounds are:

1. Remove the `ember-addon` keyword, which I've done in version 0.1.2
2. Create an `ember-addon` entry point in `package.json` that fulfills the contract (mentioned by @rwjblue with reference to `ember-cli-htmlbars` but not tried yet. I'll rev the package to 0.2.0 if it works.)

---

```
ember-cli: 2.11.1
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
4.0.3
```

jsturgis:
This change adds a user friendly error message when an addon constructor
function throws an error during addon initialization.

Fixes #6820

### DIFF
--- a/lib/models/addons-factory.js
+++ b/lib/models/addons-factory.js
@@ -52,7 +52,16 @@ class AddonsFactory {
         });
         let start = Date.now();
         let AddonConstructor = Addon.lookup(addonInfo);
-        let addon = new AddonConstructor(addonParent, project);
+        let addon;
+
+        try {
+          addon = new AddonConstructor(addonParent, project);
+        } catch (e) {
+          project.ui.writeError(e);
+          const SilentError = require('silent-error');
+          throw new SilentError(`An error occured in the constructor for ${addonInfo.name} at ${addonInfo.path}`);
+        }
+
         if (addon.initializeAddons) {
           addon.initializeAddons();
         } else {
--- a/tests/fixtures/addon/invalid-constructor/lib/ember-invalid-addon/index.js
+++ b/tests/fixtures/addon/invalid-constructor/lib/ember-invalid-addon/index.js
@@ -1,0 +1,5 @@
+'use strict';
+
+module.exports = function() {
+  throw new Error('invalid');
+};
--- a/tests/fixtures/addon/invalid-constructor/lib/ember-invalid-addon/package.json
+++ b/tests/fixtures/addon/invalid-constructor/lib/ember-invalid-addon/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "ember-invalid-addon",
+  "keywords": [
+    "ember-addon"
+  ]
+}
--- a/tests/fixtures/addon/invalid-constructor/package.json
+++ b/tests/fixtures/addon/invalid-constructor/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "test-project",
+  "private": true,
+  "ember-addon": {
+    "paths": ["./lib/ember-invalid-addon"]
+  }
+}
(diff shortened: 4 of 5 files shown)


rwjblue:
Looks really good, thanks for working on this!

One thought that I had was that we may want to also wrap the `Addon.lookup` line as well (but keep track of which one errors). That `Addon.lookup` just above is where we `require` the actual addon module ([here](https://github.com/ember-cli/ember-cli/blob/6ff2aa283c3961b6bf8f7981f3453d1db35afa9b/lib/models/addon.js#L1602)), so we could provide nicer errors for both scenarios:

* `An error occured in the constructor for ${addonInfo.name} at ${addonInfo.path}`
* `An error occured while looking up ${addonInfo.name} at ${addonInfo.path}`

What do you think?

jsturgis:
Thanks! I think that sounds like a good idea but I noticed that `Addon.lookup` is already handling the most common error case [here](https://github.com/ember-cli/ember-cli/blob/6ff2aa283c3961b6bf8f7981f3453d1db35afa9b/lib/models/addon.js#L1617) and I can't think of another error case that could happen in lookup that it isn't already checking for.

rwjblue:
Seems good, thanks for checking into it!

@homu r+

PR #6888, opened 2017-03-22
Print out `yarn install` when yarn.lock file is present

samdemaeyer:
As some might know, when you are using yarn and you remove `/node_modules`, you will have a comment when you try to run `ember server`
```
node_modules appears empty, you may need to run `npm install`
```
screen shot 2017-03-22 at 16 21 42

I fixed something similar in [ember-cli-dependency-checker](https://github.com/quaertym/ember-cli-dependency-checker/pull/68).
It's quite confusing for people who are not that familiar with the package managers.

As this is my first PR to ember-cli, I am guessing that there might be a better way for this fix. But at least the idea is out there now.

In this PR, I will print out 
```
node_modules appears empty, you may need to run `yarn`
```
screen shot 2017-03-22 at 16 35 18

when there is a `yarn.lock` file present.

I know that yarn is not used by ember-cli by default yet, but it might be ok to check this as in the team I am working in, it got confusing for some people .. 😅

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -16,6 +16,7 @@ const WatchDetector = require('../models/watch-detector');
 const SilentError = require('silent-error');
 const execSync = require('child_process').execSync;
 const fs = require('fs');
+const existsSync = require('exists-sync');
 
 let cache = {};
 
@@ -289,7 +290,7 @@ let Command = CoreObject.extend({
       if (this.works === 'insideProject') {
         if (!this.project.hasDependencies()) {
           let installInstuction = '`npm install`';
-          if (this.project.hasYarnLockFile()) {
+          if (existsSync(path.join(this.project.root, 'yarn.lock'))) {
             installInstuction = '`yarn`';
           }
           throw new SilentError(`node_modules appears empty, you may need to run ${installInstuction}`);
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -288,7 +288,11 @@ let Command = CoreObject.extend({
 
       if (this.works === 'insideProject') {
         if (!this.project.hasDependencies()) {
-          throw new SilentError('node_modules appears empty, you may need to run `npm install`');
+          let installInstuction = '`npm install`';
+          if (this.project.hasYarnLockFile()) {
+            installInstuction = '`yarn`';
+          }
+          throw new SilentError(`node_modules appears empty, you may need to run ${installInstuction}`);
         }
       }
 
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -291,7 +291,7 @@ let Command = CoreObject.extend({
         if (!this.project.hasDependencies()) {
           let installInstuction = '`npm install`';
           if (existsSync(path.join(this.project.root, 'yarn.lock'))) {
-            installInstuction = '`yarn`';
+            installInstuction = '`yarn install`';
           }
           throw new SilentError(`node_modules appears empty, you may need to run ${installInstuction}`);
         }
--- a/tests/unit/commands/generate-test.js
+++ b/tests/unit/commands/generate-test.js
@@ -47,15 +47,26 @@ describe('generate command', function() {
     td.reset();
   });
 
-  it('runs GenerateFromBlueprint but with null nodeModulesPath', function() {
-    command.project.hasDependencies = function() { return false; };
-    let originalYarnLockPath = path.join(command.project.root, 'yarn.lock');
-    let dummyYarnLockPath = path.join(command.project.root, 'foo.bar');
-    fs.renameSync(originalYarnLockPath, dummyYarnLockPath);
+  describe('without yarn.lock file', function() {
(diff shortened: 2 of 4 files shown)


lib/models/command.js
```diff
@@ -288,7 +288,11 @@ let Command = CoreObject.extend({
 
       if (this.works === 'insideProject') {
         if (!this.project.hasDependencies()) {
-          throw new SilentError('node_modules appears empty, you may need to run `npm install`');
+          let installInstuction = '`npm install`';
+          if (this.project.hasYarnLockFile()) {
```
rwjblue: I'd prefer to check it here directly, and not add another method to `project`. 

If however, we do keep the project method around, we should update other locations that search for `yarn.lock` to use it ([example here](https://github.com/ember-cli/ember-cli/blob/bd2822868e3b9bf11cfa9f28b91cb4d021c6680b/lib/commands/init.js#L64)).

samdemaeyer: @rwjblue, we only use it in [/lib/commands/init.js:64](https://github.com/ember-cli/ember-cli/blob/bd2822868e3b9bf11cfa9f28b91cb4d021c6680b/lib/commands/init.js#L64) and [lib/tasks/npm-task.js:45](https://github.com/ember-cli/ember-cli/blob/bd2822868e3b9bf11cfa9f28b91cb4d021c6680b/lib/tasks/npm-task.js#L45)
I can take it out and check it directly if you like.
Than I just need another approach to test it, ATM, we overwrite the `hasYarnLockFile` function in the project class:
```js
command.project.hasYarnLockFile = function() { return true; };
```

samdemaeyer: @rwjblue, in [this commit](https://github.com/ember-cli/ember-cli/pull/6888/commits/ea2f6dcbd96783d7e5d67cf68de38022a8eafcfc) I will check it directly. To test this behaviour, I am renaming the `ember-cli` project's actual `yarn.lock` file before the test runs and after the test, rewrite it back to the original name.
Could you advise me if this is a good approach, or might it be to risky? If so, what would you suggest for testing?

lib/models/command.js
```diff
@@ -288,7 +289,11 @@ let Command = CoreObject.extend({
 
       if (this.works === 'insideProject') {
         if (!this.project.hasDependencies()) {
-          throw new SilentError('node_modules appears empty, you may need to run `npm install`');
+          let installInstuction = '`npm install`';
+          if (existsSync(path.join(this.project.root, 'yarn.lock'))) {
+            installInstuction = '`yarn`';
```
Turbo87: can we make this `yarn install` to make it a bit easier to understand?

samdemaeyer: @Turbo87, sure, no prob.

tests/unit/commands/generate-test.js
```diff
@@ -47,9 +49,20 @@ describe('generate command', function() {
 
   it('runs GenerateFromBlueprint but with null nodeModulesPath', function() {
     command.project.hasDependencies = function() { return false; };
+    let originalYarnLockPath = path.join(command.project.root, 'yarn.lock');
+    let dummyYarnLockPath = path.join(command.project.root, 'foo.bar');
+    fs.renameSync(originalYarnLockPath, dummyYarnLockPath);
 
     return expect(command.validateAndRun(['controller', 'foo'])).to.be.rejected.then(reason => {
       expect(reason.message).to.eql('node_modules appears empty, you may need to run `npm install`');
+    }).then(() => fs.renameSync(dummyYarnLockPath, originalYarnLockPath));
```
Turbo87: this `renameSync` will not run if the assertion above fails. you could wrap it in a describe() block though:

```js
describe('without yarn.lock file', function() {
  let originalYarnLockPath, dummyYarnLockPath;

  beforeEach(function() {
    originalYarnLockPath = path.join(command.project.root, 'yarn.lock');
    dummyYarnLockPath = path.join(command.project.root, 'foo.bar');
    fs.renameSync(originalYarnLockPath, dummyYarnLockPath);
  });

  afterEach(function() {
    fs.renameSync(dummyYarnLockPath, originalYarnLockPath);
  });

  it('runs GenerateFromBlueprint but with null nodeModulesPath', function() {
    command.project.hasDependencies = function() { return false; };

    return expect(command.validateAndRun(['controller', 'foo'])).to.be.rejected.then(reason => {
      expect(reason.message).to.eql('node_modules appears empty, you may need to run `npm install`');
    });
  });
});
``

Comment (samdemaeyer) [inline tests/unit/commands/generate-test.js]: Thanks @Turbo87, makes sense!
Question, would you typically destroy the global variables in the `afterEach` function?
```js
afterEach(function() {
  fs.renameSync(dummyYarnLockPath, originalYarnLockPath);
  originalYarnLockPath = dummyYarnLockPath = null;
});
```

Turbo87: since they are reassigned in every `beforeEach()` block there is IMHO no need to reset them in `afterEach()`

Turbo87:
@homu r+

PR #6889, opened 2017-03-22
Use const/let in all blueprints

simonihmig:
This will remove all left over usage of `var` off the blueprints.

When to use `const` vs. `let` is a matter of opinions obviously. After checking with @Turbo87 on Slack, I was assuming the `ember-suave`-like rules of `const` in module root, `let` otherwise would be appropriate. Although in some cases we have `require` statements in a function body, so used `const` there as well. Hope that makes sense...

There will be some test failures, as this is relying on https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/104 to be merged and released. So don't merge yet!

### DIFF
--- a/blueprints/addon/files/ember-cli-build.js
+++ b/blueprints/addon/files/ember-cli-build.js
@@ -1,8 +1,9 @@
+'use strict';
 /* eslint-env node */
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function(defaults) {
-  var app = new EmberAddon(defaults, {
+  let app = new EmberAddon(defaults, {
     // Add options here
   });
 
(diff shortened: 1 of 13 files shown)


rwjblue:
Changes look good.  I agree with the `const` / `let` split here. 

Not sure why the tests are failing, might need to dig into the travis output to see...

simonihmig:
@rwjblue The two failing tests are the ones I was mentioning above, requiring the changes in https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/104 to succeed. Once there is a new release of `ember-cli-blueprint-test-helpers`, I can add a commit to bump that dependency, which should fix the failing tests. (I tried that successfully on my local machine with `yarn link`!)

Turbo87:
@simonihmig I just published v0.17.1 of the blueprint-test-helpers

Turbo87:
@homu r+

Turbo87:
@homu retry

PR #6903, opened 2017-03-25
remove npm experiment refs

### LINKED ISSUE #6900: Remove the `npm-blueprints` experiment
The `npm-blueprints` experiment enables users to use blueprints published to npm directly when creating new applications. Since this is a new feature, at the moment there aren't many usable packages, but you can try `ember new some-good-name-here -b ember-blueprint-demo` if you're on canary.

The work entails two things:

* Remove `NPM_BLUEPRINTS` from the [/lib/experiments/index.js](https://github.com/ember-cli/ember-cli/blob/master/lib/experiments/index.js#L7) file
* Find any reference in the code to `experiments.NPM_BLUEPRINTS` and remove/change the code accordingly

If we check the [`install-blueprint` file](https://github.com/ember-cli/ember-cli/blob/65dac827d931f5c9e4739106c9afc39ab4f298a0/lib/tasks/install-blueprint.js#L78-L81), the following code would need to be removed:

```
    if (!experiments.NPM_BLUEPRINTS) {
      logger.info(`"NPM_BLUEPRINTS experiment disabled -> rethrowing original error`);
      throw error;
    }
```

Other references also need to be found and removed, and relevant tests need to be updated so the build is green.

tylerturdenpants:
Per issue description, removed and refactored npm experiment references

resolves #6900

### DIFF
--- a/lib/experiments/index.js
+++ b/lib/experiments/index.js
@@ -4,7 +4,6 @@ const getCallerFile = require('get-caller-file');
 const deprecate = require('../utilities/deprecate');
 
 let experiments = {
-  NPM_BLUEPRINTS: symbol('npm-blueprints'),
   MODULE_UNIFICATION: symbol('module-unification'),
   get INSTRUMENTATION() {
     deprecate(`\`experiments.INSTRUMENTATION\` is deprecated now that the feature has landed, use \`instrumentation\` method on your addon instead. Required from: \n${getCallerFile()}`, true);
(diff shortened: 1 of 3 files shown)


Turbo87:
IMHO we should test drive that at least until a couple of days after EmberConf before actually removing the feature flag. essentially we're not in any rush since the next branch point is still a few weeks out.

rwjblue:
@Turbo87 - Yep, agreed.  We should hold off on landing this for a week or two.

@tylerturdenpants - Thanks for taking the time and submitting the PR!

rwjblue:
We discussed at yesterdays ember-cli meeting, and are 👍 on moving this forward.  There has been quite a bit of real world testing/usage (mostly with `@glimmer/blueprint`).

rwjblue:
@homu r+

PR #6904, opened 2017-03-25
Update ember-cli-qunit to use Babel 6.

rwjblue:
The primary changes in ember-cli-qunit@4.0.0-beta.1 are:

* Requires Node 4 or higher.
* Updates to use Babel 6.
* Removes test auto-start behavior, in favor of user invoking `start()` when ready.


rwjblue:
@homu r+

PR #6907, opened 2017-03-26
Ignore files created by Ember-Try

elwayman02:
Ember-Try creates a number of temporary files within a project during its test runs. This can cause IDEs & Version Control Tools to try and scan the files, which uses unnecessary resources on the system. Additionally, if Ember-Try exits unexpectedly, some of these files may not be cleaned up correctly, leaving them around to be accidentally committed.

Adding these files to `.gitignore` will alleviate both problems, given the intentional temporary nature of the files.

### DIFF
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -15,3 +15,8 @@
 /libpeerconnection.log
 npm-debug.log*
 testem.log
+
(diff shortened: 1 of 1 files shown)


Turbo87:
that seems like a good idea, thanks!

@homu r+

PR #6908, opened 2017-03-27
Several fixes for the module unification feature flag

mixonic:
Several fixes here for the module unification flag, written mostly by @rwjblue with my very watchful supervision.

* Avoid depending on the `app` tree being present at all times
* Create a more realistic `src` build pipeline
* Property add `src` files to the linting system
* When the `src` tree is present, defer to `src/main.js` to declare an application instead of `app/app.js`.

/cc @rwjblue @Turbo87

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -197,7 +197,12 @@ class EmberApp {
     let srcTree = existsSync(srcPath) ? new WatchedDir(srcPath) : null;
 
     let appPath = this._resolveLocal('app');
-    let appTree = existsSync(appPath) ? new WatchedDir(appPath) : null;
+    let appTree;
+    if (experiments.MODULE_UNIFICATION) {
+      appTree = existsSync(appPath) ? new WatchedDir(appPath) : null;
+    } else {
+      appTree = new WatchedDir(appPath);
+    }
 
     let testsPath = this._resolveLocal('tests');
     let testsTree = existsSync(testsPath) ? new WatchedDir(testsPath) : null;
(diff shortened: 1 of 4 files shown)


lib/broccoli/ember-app.js
```diff
@@ -187,7 +187,8 @@ class EmberApp {
     let srcPath = this._resolveLocal('src');
     let srcTree = existsSync(srcPath) ? new WatchedDir(srcPath) : null;
 
-    let appTree = new WatchedDir(this._resolveLocal('app'));
+    let appPath = this._resolveLocal('app');
+    let appTree = existsSync(appPath) ? new WatchedDir(appPath) : null;
```
Turbo87: I'm a little worried about the potential side effects of this change since we don't have that many unit tests for the build pipeline yet and the change is not covered by the feature flag

rwjblue: Ya, we need to flag this, and add tests to cover these changes generally speaking.

mixonic:
This code is some of that last stuff we need to see landed in order to support MU behind feature flags.

rwjblue:
Finally got some time to work on this, and I think its ready for another round of reviews.

* Rebased on top of current master
* Removed one super bizarre (and misplaced) test
* Fixed up the previously failing tests
* Added a few new integration style tests around `lib/broccoli/ember-app.js` (we can now confirm that `EmberApp#appAndDependencies` returns the right things in a fast (< 100ms) broccoli-test-helper style test 🎉 )
* Guarded all new functionality behind the `MODULE_UNIFICATION` experiment

lib/broccoli/ember-app.js
```diff
@@ -1127,8 +1183,11 @@ class EmberApp {
     let config = this._configTree();
     let templates = this._processedTemplatesTree();
 
+    let srcTree = experiments.MODULE_UNIFICATION ? this._processedSrcTree() : null;
```
cibernox: Perhaps move the `experiments.MODULE_UNIFICATION` check to `_processedSrcTree` just call it here?

twokul:
@mixonic @rwjblue I must be reading the code wrong? `experiments.MODULE_UNIFICATION` always returns something (`__module-unification__ [id=EMBER_CLI91203924408]` for example). That means check like this won't work:

```javascript
if (!experiments.MODULE_UNIFICATION) {
  return null;
}
```

and you always have to pair it with another guard `!this.trees.src`

Turbo87:
@twokul experiments are disabled on the `beta` and `release` branches

Issue #6780, opened 2017-02-16
Document `Addon` model `init` method.


rwjblue:
A coworker noticed that `init` is not listed in the API docs for `lib/models/addon.js`.

We should add documentation for `init` (it is definitely public API) to:

https://ember-cli.com/api/classes/Addon.html

mixonic:
Closed by https://github.com/ember-cli/ember-cli/pull/6910

PR #6910, opened 2017-03-27
[BUGFIX] Add yuidocs for the addon:init method

mattmarcum:
Addresses [6780](https://github.com/ember-cli/ember-cli/issues/6780)

Just starting the documentation process.  Let me know if there's any other relevant info to add in here.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -175,6 +175,23 @@ function registryHasPreprocessor(registry, type) {
   @param {Project} project The current project (deprecated)
 */
 let addonProto = {
+
+  /**
+  Initializes the addon.  If you override this method make sure and call `this._super.init && this._super.init.apply(this, arguments);` or your addon will not work.
+
+  @public
+  @method init
+  @param {Project|Addon} parent The project or addon that directly depends on this addon
+  @param {Project} project The current project (deprecated)
+
+  @example
+  ```js
+  init(parent, project) {
+    this._super.init && this._super.init.apply(this, arguments);
+    this._someCustomSetup();
+  }
+  ```
+  */
   init(parent, project) {
     this._super();
     this.parent = parent;


lib/models/addon.js
```diff
@@ -175,6 +175,15 @@ function registryHasPreprocessor(registry, type) {
...
 */
 let addonProto = {
+
+  /**
+  Initializes the addon.  If you override this method make sure and call `this._super()` or your addon will not work.
+
+  @public
+  @method init
+  @param {Project|Addon} parent The project or addon that directly depends on this addon
+  @param {Project} project The current project (deprecated)
```
Turbo87: could you add an `@example` for this method? I think the `this._super()` call should actually include the `arguments` too (e.g. like in https://github.com/babel/ember-cli-babel/blob/4ab3d4a0929c6a4c9420a7bd6c15ed24f6d976f6/index.js#L23)

mattmarcum: ok, I used argument destructuring.

Turbo87: I think argument destructuring is not available on Node 4 though, and since we don't run Ember CLI itself through Babel we can't use it yet.

mattmarcum: ok

lib/models/addon.js
```diff
@@ -175,6 +175,23 @@ function registryHasPreprocessor(registry, type) {
...
+
+  @example
+  ```js
+  init(parent, project) {
+    this._super(parent, project);
+    this._someCustomSetup();
+  }
+  ```
```
Turbo87: could you change the example to:

```js
init: function() {
  this._super.init && this._super.init.apply(this, arguments);
  this._someCustomSetup();
}
```

Turbo87:
awesome, thanks. could you squash the commits?

mattmarcum:
squashed

Turbo87:
muchas gracias!

@homu r+

PR #6911, opened 2017-03-27
Fix dirty git state

### LINKED ISSUE #6909: Git status dirty after running test suite
On current master, an additional file file is present after running the test suite via `npm run test-all`: `blueprints/addon/files/package.json`.

After talking through this with @Turbo87, it's likely that this file is simply copied over from an app's `package.json` and not being cleaned up properly. I'll try to track this down and will submit a PR to fix if I find the offending test.

clekstro:
This fixes #6909. Inside of the `afterInstall` hook, we were using the async methods of fs-extra (`.remove()`) but not actually waiting for them to resolve. Instead, we use the sync variant to resolve the issue, as this ensures the file has been removed before we proceed.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -69,7 +69,7 @@ module.exports = {
     let bowerPath = path.join(this.path, 'files', 'bower.json');
 
     [packagePath, bowerPath].forEach(filePath => {
-      fs.remove(filePath);
+      fs.removeSync(filePath);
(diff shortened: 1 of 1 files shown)


Turbo87:
looks like this was caused by myself in https://github.com/ember-cli/ember-cli/pull/5797

Turbo87:
@homu r+

PR #6912, opened 2017-03-27
Stop creating recursive symlink (addon requiring itself)

clekstro:
When running the `addon/smoke-test-slow` test in isolation, I was getting an `EEXIST` error that the `node_modules/developing-addon` symlink that we were creating manually in this test is already present. Deleting the extra test setup resolved the issue.

This just impacts the test suite and was a precursor to #6911.  

Many thanks to @Turbo87 for helping to track this down :heart:


rwjblue:
@homu r+

Issue #6311, opened 2016-10-03
consider cleanup of watch-detector


stefanpenner:
https://github.com/ember-cli/ember-cli/pull/6309#discussion_r81490324

specially: https://github.com/ember-cli/ember-cli/blob/master/lib/models/watch-detector.js#L173-L181

It is documented (although those comments can be improved), but if someone feels they can express it in a clearer way a PR for review would be good.

locks:
Closing as a related PR with cleanup has been merged.

PR #6916, opened 2017-03-27
Watch-detector clean up

yogipatel:
Addresses #6311.

Refactors the watch detector code to be clearer to read. The logic
used to detect node/watchman/etc have not been changed much, they've
just been updated to not include any fallback logic and to not write
and text back to the UI.

The basic logic of selecting a watcher is to first try the user's
preference, if provided, otherwise watchman, node, and polling
watchers are tried, in order. The `events` type isn't documented
anywhere, AFAICT, so it might be a candidate for removal since the
default order covers the events type anyway.

### DIFF
--- a/lib/models/watch-detector.js
+++ b/lib/models/watch-detector.js
@@ -60,27 +60,6 @@ class WatchDetector {
   }
 
   /*
-   * @private
-   *
-   * implements input options parsing and fallback.
-   *
-   * @method extractPreferenceFromOptions
-   * @returns {WatchDetector)
-  */
-  extractPreferenceFromOptions(options) {
-    if (options.watcher === POLLING) {
-      debug.info('skip detecting watchman, "polling" was selected.');
-      return new WatchPreference(POLLING);
-    } else if (options.watcher === NODE) {
-      debug.info('skip detecting watchman, "node" was selected.');
-      return new WatchPreference(NODE);
-    } else {
-      debug.info('no watcher preference set, lets attempt watchman');
-      return new WatchPreference(WATCHMAN);
-    }
-  }
-
-  /*
    * @public
    *
    * Detect if `fs.watch` provided by node is capable of observing a directory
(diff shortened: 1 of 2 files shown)


lib/models/watch-detector.js
```diff
@@ -130,55 +109,49 @@ class WatchDetector {
...
-    debug.info('foundBestWatcherOption, preference was: %o, actual: %o', options, actual);
-    if (actual /* if no preference was initial set, don't bother informing the user */ &&
-      original /* if no original was set, the fallback is expected */ &&
-      actual.watcher !== original &&
-
-      // events represents either watcherman or node, but no preference
-      !(original === EVENTS && (actual.watcher === WATCHMAN || actual.watcher === NODE))) {
-      // if there was an initial preference, but we had to fall back inform the user.
-      this.ui.writeLine(`was unable to use: "${original}", fell back to: "${actual.watcher}"`);
+      this.ui.writeLine(`Selected ${bestWatcher.watcher} watcher`);
```
hjdivad: It's a bit chatty to always output the selected watcher.  Previously we only wrote something if we had to fall back.  What's the reason for changing this?

tests/unit/models/watch-detector-test.js
```diff
@@ -147,78 +133,32 @@ describe('WatchDetector', function() {
...
-
-        childProcess.execSync = function() {
-          throw new Error();
-        };
-        fs.watch = function() { return { close() { } }; };
-
-        let result = subject.checkWatchman();
-        expect(result).to.have.property('watcher', 'node');
-        expect(ui.output).to.eql('');
-      });
```
hjdivad: we still want this test for `getWatchmanWatcher` don't we?

tests/unit/models/watch-detector-test.js
```diff
@@ -147,78 +133,32 @@ describe('WatchDetector', function() {
...
-
-        childProcess.execSync = function() {
-          throw new Error();
-        };
-
-        let result = subject.checkWatchman();
-        expect(result).to.have.property('watcher', 'node');
-        expect(ui.output).to.match(/Could not start watchman/);
-        expect(ui.output).to.match(/Visit https:\/\/ember-cli.com\/user-guide\/\#watchman/);
-      });
```
hjdivad: we still want this test for `getWatchmanWatcher` don't we?

Turbo87:

PR #6919, opened 2017-03-29
Do not use `chalk.white` when displaying asset sizes

lucasmazza:
White text is not readable on white/light terminals.

Before:

before

After:

after


Turbo87:
that makes sense, thanks!

@homu r+

Turbo87:
@rwjblue should we pull that down to the release branch?

PR #6930, opened 2017-04-02
test node 4 instead of 6

kellyselden:
Scenario: I'm changing my addon's index.js. I've introduced node 6 syntax unknowingly. Since I'm running node 6 locally and our Travis blueprints test node 6 only, it remains broken until someone starts using the addon in node 4.

Downside: It slows down Travis runs. (We could manually install npm 3/4 in Travis to speed it up again)

Thoughts?

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "6"
+  - "4"
 
 sudo: false
 
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "6"
+  - "4"
 
 sudo: false
 <% if (yarn) { %>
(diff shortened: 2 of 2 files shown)


rwjblue:
You are basically saying that we should suggest folks test in the lowest supported Node version, right?

kellyselden:
Yes

kellyselden:
I accidentally put `this._super.included(...arguments);` in an addon and never realized it broke node 4.

rwjblue:
Ya, makes sense to me.

rwjblue:
@stefanpenner / @Turbo87 - Any objections?

Turbo87:
I think for apps it's fine to test on Node 6 actually. Since we will add that 6 to 4 migration to `ember-addon-output` we should likely add a comment explaining why we're using Node 4 here.

Turbo87:
@kellyselden could you update the PR?

Turbo87:
> Since we will add that 6 to 4 migration to ember-addon-output we should likely add a comment explaining why we're using Node 4 here.

@stefanpenner any opinions?

Turbo87:
@kellyselden there is one more issue with this: Node 4 comes with npm 2 and the more recent CLI releases require at least npm 3

PR #6932, opened 2017-04-02
make blueprint files public

kellyselden:
We use it ([here](https://github.com/ember-cli/ember-cli/blob/master/blueprints/server/index.js#L30)), and is more useful now that we have good support for choices in linting and testing (different config files).

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -231,6 +231,7 @@ let Blueprint = CoreObject.extend({
   /**
     Used to retrieve files for blueprint.
 
+    @public
     @method files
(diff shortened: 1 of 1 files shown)


twokul:
it seems like our tests rely on `.jshintrc` files; a lot of `ENOENT: no such file or directory, stat '/home/travis/build/ember-cli/ember-cli/blueprints/server/files/server/.jshintrc'` failures 😕

kellyselden:
@twokul Shoot I didn't mean to remove that file...

kellyselden:
Fixed.

PR #6933, opened 2017-04-02
sync api docs with website

kellyselden:
Blocked on https://github.com/ember-cli/ember-cli.github.io/pull/130

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -59,14 +59,14 @@ const writeFile = RSVP.denodeify(fs.outputFile);
   `files` contains templates for the all the files to be
   installed into the target directory.
 
-  The `__name__` token is subtituted with the dasherized
+  The __`__name__`__ token is subtituted with the dasherized
   entity name at install time. For example, when the user
   invokes `ember generate controller foo` then `__name__` becomes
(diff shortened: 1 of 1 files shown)


lib/models/blueprint.js
```diff
@@ -59,14 +59,14 @@ const writeFile = RSVP.denodeify(fs.outputFile);
   `files` contains templates for the all the files to be
   installed into the target directory.
 
-  The `__name__` token is subtituted with the dasherized
+  The __`__name__`__ token is subtituted with the dasherized
```
Turbo87: this looks strange...

Turbo87: also this should probably by `substituted` instead of `subtituted`, right?

trabus: I think that's just for bolding it __`__name__`__ vs `__name__`

Turbo87: well, I understand what it does, but what is not clear to me is why we would want it explicitly bold here. we could do the same using CSS if we wanted to.

Turbo87:
@kellyselden can you update and rebased this PR?

Turbo87:
@kellyselden friendly reminder 😉

Turbo87:

PR #6936, opened 2017-04-02
use RSVP.resolve shorthand

kellyselden:


kellyselden:
@twokul A friendly reminder that `@homu r+` is the preferred merging strategy.

PR #6938, opened 2017-04-02
Add ESLint config to "server" and "lib" blueprints

kellyselden:
@Turbo87 Am I on the right track? I made assumptions on what the fix was.


kellyselden:
@homu r+

PR #6941, opened 2017-04-03
[BUGFIX] Do not set committer for the initial git commit

### LINKED ISSUE #6920: ember new fails on commit when signing commits with GPG
Running `ember new ember-app` when global commit signing is enabled fails the ember new process..

Here's the issue in the output:

```
Error creating new application. Removing generated directory `./ember-app`
Command failed: git commit -m "Initial Commit from Ember CLI v2.12.1

                                                        _...,
                                                     ,:^;,...;
              -+===;.         ,,--++====++-,,,     .:  /.....,
            :::::~+++++#:,+#++++++++++++++++++#*..:   /,......
           (,,,,,,::=+++##++++++++++++++++++++++#.   :....../
            ...,,,,,::++++++++++++++++++++++++++++++*..,...:
            *..+...,#@@@@@@@@@++++++++++++++++++++++#*....*
              @#,;##############@@@+*+#@@@@@@@@@@#*++#..<
              *@##@@+,-*^^^*-+@####@@@######@@@#####@@,,,+
                @#@*            @#@@@@#@@+--*^^*--#@@@@@@#
                @#@.    @#      @##+++@#,           .@@#@@
                 #@#    @@     +@@++++#@@     @@     :@@
                 :@#*         @#@++++++@#*    #@     @@+
                :*+@@#;,.__.+@#@+,-^^.++@#          @@++
               ;*  :*@@@##@@@@;++r._j^.+@##@+,.__,,@@++.
              /*    ........+++++++++++++#@@@@@###@@#++++,
            ,:       ...,@@@#++===----==@@@####,,....+++++
           .:       ......@@##@\   ;   :@####@,,...... +++.
           ;       .........@###,   ;  ;xx#@;,,.....   *;+,
           |       ........,*;xxxx--^--=xxx,........   :+#;
           ;         ......,,;xxxxxxxxxxxxx;,.....     *+#
            ;          ......,::xxxx;.     ......       +.   .
             *;            .........  +###  ....       / ,. /:| ,.
               .+:             ...  ;##++##, .      ,#. (..v..;*./
                  **                ##  ###*    .:*&&&+. \.,....<,
                   #&+**==-..,,__  ;##  ###  :,*+&&&&&&&v+#&,,.._/
                  #&&&&*...,::,,.  ##; ,##* .*****;:&&&&&&&&&
                 ,+*+;~*..*** *.* ### ###* *******    *+#&;*
                                  ##,;##    ****    :,  **
         #####    ##   ###  ###,  ########       .#####   ;##  ##
        #######  ;##  #### ,###.  ##########    ########  ### ####
       ###  ###  ### ##########   ####  ####   ,##   ###  #######*
      ### ,###  ##############:   ##     ###  #### ,##   :#### ###  ##;
  ##########    ########### ##   .##    ,###  #######    ##### :######
    ######    .###### ####  ##   ###    ### ######*    :#####   ####
       #############  ####  ################    ######## ###
        #####*  *#*    #:   :###   *###*         *####    #*
"
error: gpg failed to sign the data
fatal: failed to write commit object

Error: Command failed: git commit -m "Initial Commit from Ember CLI v2.12.1

                                                        _...,
                                                     ,:^;,...;
              -+===;.         ,,--++====++-,,,     .:  /.....,
            :::::~+++++#:,+#++++++++++++++++++#*..:   /,......
           (,,,,,,::=+++##++++++++++++++++++++++#.   :....../
            ...,,,,,::++++++++++++++++++++++++++++++*..,...:
            *..+...,#@@@@@@@@@++++++++++++++++++++++#*....*
              @#,;##############@@@+*+#@@@@@@@@@@#*++#..<
              *@##@@+,-*^^^*-+@####@@@######@@@#####@@,,,+
                @#@*            @#@@@@#@@+--*^^*--#@@@@@@#
                @#@.    @#      @##+++@#,           .@@#@@
                 #@#    @@     +@@++++#@@     @@     :@@
                 :@#*         @#@++++++@#*    #@     @@+
                :*+@@#;,.__.+@#@+,-^^.++@#          @@++
               ;*  :*@@@##@@@@;++r._j^.+@##@+,.__,,@@++.
              /*    ........+++++++++++++#@@@@@###@@#++++,
            ,:       ...,@@@#++===----==@@@####,,....+++++
           .:       ......@@##@\   ;   :@####@,,...... +++.
           ;       .........@###,   ;  ;xx#@;,,.....   *;+,
           |       ........,*;xxxx--^--=xxx,........   :+#;
           ;         ......,,;xxxxxxxxxxxxx;,.....     *+#
            ;          ......,::xxxx;.     ......       +.   .
             *;            .........  +###  ....       / ,. /:| ,.
               .+:             ...  ;##++##, .      ,#. (..v..;*./
                  **                ##  ###*    .:*&&&+. \.,....<,
                   #&+**==-..,,__  ;##  ###  :,*+&&&&&&&v+#&,,.._/
                  #&&&&*...,::,,.  ##; ,##* .*****;:&&&&&&&&&
                 ,+*+;~*..*** *.* ### ###* *******    *+#&;*
                                  ##,;##    ****    :,  **
         #####    ##   ###  ###,  ########       .#####   ;##  ##
        #######  ;##  #### ,###.  ##########    ########  ### ####
       ###  ###  ### ##########   ####  ####   ,##   ###  #######*
      ### ,###  ##############:   ##     ###  #### ,##   :#### ###  ##;
  ##########    ########### ##   .##    ,###  #######    ##### :######
    ######    .###### ####  ##   ###    ### ######*    :#####   ####
       #############  ####  ################    ######## ###
        #####*  *#*    #:   :###   *###*         *####    #*
"
error: gpg failed to sign the data
fatal: failed to write commit object

    at ChildProcess.exithandler (child_process.js:210:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:194:7)
    at maybeClose (internal/child_process.js:899:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

merwan:
As discussed in #6920, setting the git committer for the initial commit can trigger a gpg error which prevents the project from being created.
Setting only the git author and not the git committer should fix this issue and is also semantically more correct as the first commit is authored by emberjs-cli but is committed by the user.

Resolves #6920

### DIFF
--- a/lib/tasks/git-init.js
+++ b/lib/tasks/git-init.js
@@ -12,7 +12,6 @@ const Promise = RSVP.Promise;
 class GitInitTask extends Task {
   run(_commandOptions) {
     let commandOptions = _commandOptions || {};
-    const template = require('lodash.template');
     const chalk = require('chalk');
     let ui = this.ui;
 
@@ -26,11 +25,7 @@ class GitInitTask extends Task {
 
       return this._gitInit()
         .then(() => this._gitAdd())
-        .then(() => {
-          let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
-          let commitMessage = template(commitTemplate)(pkg);
-          return this._gitCommit(commitMessage);
-        })
+        .then(() => this._gitCommit())
         .then(() => ui.writeLine(chalk.green('Successfully initialized git.')));
     })
       .catch(error => {
@@ -53,8 +48,19 @@ class GitInitTask extends Task {
     return execa('git', ['add', '.']);
   }
 
-  _gitCommit(commitMessage) {
-    return execa('git', ['commit', '-m', commitMessage], { env: this.buildGitEnvironment() });
+  _gitCommit() {
+    const template = require('lodash.template');
+    let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
+    let commitMessage = template(commitTemplate)(pkg);
+    let env = this.buildGitEnvironment();
+    return execa('git', ['commit', '-m', commitMessage], { env })
+      .catch(error => {
+        if (error.message.indexOf('git config --global user') > -1) {
+          env.GIT_COMMITTER_NAME = 'Tomster';
+          env.GIT_COMMITTER_EMAIL = 'tomster@emberjs.com';
+          return execa('git', ['commit', '-m', commitMessage], { env });
+        }
+      });
   }
 
   buildGitEnvironment() {
@@ -63,8 +69,6 @@ class GitInitTask extends Task {
     return Object.assign({}, process.env, {
       GIT_AUTHOR_NAME: 'Tomster',
       GIT_AUTHOR_EMAIL: 'tomster@emberjs.com',
-      GIT_COMMITTER_NAME: 'Tomster',
-      GIT_COMMITTER_EMAIL: 'tomster@emberjs.com',
     });
   }
 }
--- a/lib/tasks/git-init.js
+++ b/lib/tasks/git-init.js
@@ -55,7 +55,7 @@ class GitInitTask extends Task {
     let env = this.buildGitEnvironment();
     return execa('git', ['commit', '-m', commitMessage], { env })
       .catch(error => {
-        if (error.message.indexOf('git config --global user') > -1) {
+        if (error && error.message && error.message.indexOf('git config --global user') > -1) {
           env.GIT_COMMITTER_NAME = 'Tomster';
           env.GIT_COMMITTER_EMAIL = 'tomster@emberjs.com';
           return execa('git', ['commit', '-m', commitMessage], { env });
--- a/tests/unit/tasks/git-init-test.js
+++ b/tests/unit/tasks/git-init-test.js
@@ -52,13 +52,13 @@ describe('git-init', function() {
     td.when(task._gitVersion()).thenResolve();
     td.when(task._gitInit()).thenResolve();
     td.when(task._gitAdd()).thenResolve();
-    td.when(task._gitCommit(td.matchers.anything())).thenResolve();
+    td.when(task._gitCommit()).thenResolve();
 
     return task.run().then(function() {
       td.verify(task._gitVersion());
       td.verify(task._gitInit());
       td.verify(task._gitAdd());
-      td.verify(task._gitCommit(td.matchers.anything()));
+      td.verify(task._gitCommit());
 
       expect(task.ui.output).to.contain('Successfully initialized git.');
       expect(task.ui.errors).to.equal('');


twokul:
Travis is failing with:

```sh
*** Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
to set your account's default identity.
Omit --global to set the identity only in this repository.
fatal: empty ident name (for <travis@testing-docker-8b1ca034-f637-4627-8fd7-8e6cbf338ceb.(none)>) not allowed
Error: Command failed: git commit -m Initial Commit from Ember CLI v2.13.0-beta.3
```

I think we need to get user's name and email and use those.

merwan:
Not sure how this should be handled. It seems legit to assume that all users have their name and email set. I think we should set them specifically in the test environment.

twokul:
@merwan from a perspective of an experienced git user -- sure, that's a fair assumption; from a perspective of a novice, I have doubts.

two things:
+ we could use a fake user for testing purposes
+ we should likely check if user name and email are set to avoid the failure altogether and provide a useful error message.

thoughts?

kellyselden:
@twokul I like both your suggestions.

merwan:
I updated the PR. This checks that either the git user email and name are set either though environment variables or in a git configuration file. If not, an error message is displayed and the git init steps are not executed.

I'm new to nodejs development and I don't think I handled well the case when the git user is not configured. I would be happy to receive comments and suggestions on what should be done in this case. Same thing for the two functions that check that the git user email and name, it's probably not using the usual nodejs idioms, I would be happy to improve this!

lib/tasks/git-init.js
```diff
@@ -24,14 +24,19 @@ class GitInitTask extends Task {
...
         .then(() => this._gitAdd())
         .then(() => {
           let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
           let commitMessage = template(commitTemplate)(pkg);
           return this._gitCommit(commitMessage);
         })
-        .then(() => ui.writeLine(chalk.green('Successfully initialized git.')));
+        .then(() => ui.writeLine(chalk.green('Successfully initialized git.')))
+        .catch(() => {
+          ui.writeError('Git user email and name should be set.');
```
twokul: should we change this to:

```js
let errorMessage = 'You should set `user.email` and `user.name` via `git config`. For more information, visit https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration';
return Promise.reject(new SilentError(errorMessage));
```

lib/tasks/git-init.js
```diff
@@ -45,6 +50,20 @@ class GitInitTask extends Task {
     return execa('git', ['--version']);
   }
 
+  _gitEmailConfigured() {
+    if (process.env.GIT_COMMITTER_EMAIL) {
+      return Promise.resolve();
```
twokul: we might want to resolve with `process.env.GIT_COMMITTER_EMAIL`; it's always confusing to consuming promises with empty fulfilled value.

lib/tasks/git-init.js
```diff
@@ -45,6 +50,20 @@ class GitInitTask extends Task {
...
+    }
+    return execa('git', ['config', '--get', 'user.email']);
+  }
+
+  _gitNameConfigured() {
+    if (process.env.GIT_COMMITTER_NAME) {
+      return Promise.resolve();
```
twokul: same here

lib/tasks/git-init.js
```diff
@@ -24,14 +24,19 @@ class GitInitTask extends Task {
     return this._gitVersion().then(() => {
       gitVersionWorked = true;
 
-      return this._gitInit()
+      return this._gitEmailConfigured()
```
twokul: I'd change the name to something like `getCommiterEmail` & `getComitterName` but that's my personal preference. @kellyselden @Turbo87 thoughts?

twokul:
@merwan thanks for the PR!

Turbo87:
I'm wondering if we should just try committing it without checking the username/email, and if it fails and the error message contains `git config --global user` we retry committing with a default username/email. Should likely be simpler and faster compared to checking for username/email upfront.

twokul:
@Turbo87 what do you mean by default? Tomster? If so, my perception was that we don't want to use Tomster for the first commit as it causes issues with gpg signing. Am I totally off?

as to how to detect username/email -- I generally prefer to be explicit and not to grep output for strings; also, I think checking for environment variables might be better than running the command and anticipating failure. I'm not quite sure about faster and simpler.

having said that I'll let you make a call.

Turbo87:
> what do you mean by default? Tomster?

yes, the same one that we always used until now.

> If so, my perception was that we don't want to use Tomster for the first commit as it causes issues with gpg signing.

that is why we should first try it without overriding the committer information

> I think checking for environment variables might be better than running the command and anticipating failure

I'm not convinced that reading env vars will actually cover all the way of setting committer info

> I'm not quite sure about faster and simpler.

faster: what I'm proposing will work with just a single `git commit` in the majority of the cases and a second `git commit` in the other cases, which optimizes for making the common scenario fast (if that even matters...).

simpler: we're only running `git commit` once or twice instead of reading any env vars. conceptually that seems to be the same thing a human would do: try committing with default user info first, if that fails we follow the instructions and set the user info (although temporary in our case) and then commit again.

stefanpenner:
@turbo87 sounds good

twokul:
@Turbo87 sounds good, thanks for explaining

merwan:
@Turbo87 @twokul I updated the PR according to https://github.com/ember-cli/ember-cli/pull/6941#issuecomment-291776081

lib/tasks/git-init.js
```diff
@@ -53,8 +48,19 @@ class GitInitTask extends Task {
...
-  _gitCommit(commitMessage) {
-    return execa('git', ['commit', '-m', commitMessage], { env: this.buildGitEnvironment() });
+  _gitCommit() {
+    const template = require('lodash.template');
+    let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
+    let commitMessage = template(commitTemplate)(pkg);
+    let env = this.buildGitEnvironment();
+    return execa('git', ['commit', '-m', commitMessage], { env })
+      .catch(error => {
+        if (error.message.indexOf('git config --global user') > -1) {
```
stefanpenner: error may be `null` or undefined or something non-object

rwjblue: ahh, good point

So we should tweak the check to be something like:

```
if (error && error.message && error.message.indexOf('git config --global user') > -1)` {
```

Sound good?

lib/tasks/git-init.js
```diff
@@ -53,8 +48,19 @@ class GitInitTask extends Task {
...
+    let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
+    let commitMessage = template(commitTemplate)(pkg);
+    let env = this.buildGitEnvironment();
+    return execa('git', ['commit', '-m', commitMessage], { env })
+      .catch(error => {
+        if (error && error.message && error.message.indexOf('git config --global user') > -1) {
+          env.GIT_COMMITTER_NAME = 'Tomster';
+          env.GIT_COMMITTER_EMAIL = 'tomster@emberjs.com';
+          return execa('git', ['commit', '-m', commitMessage], { env });
+        }
```
Turbo87: since we're not rethrowing `error` here the error will be swallowed now, right?

stefanpenner: oops your are right, we should re-throw....

stefanpenner: PR inbound

merwan: I'm so frustrated I didn't handle this properly in the first place... Thanks for fixing the PR!

PR #6955, opened 2017-04-12
Discover dependencies of npm-linked addons

ef4:
This alters our addon resolution strategy to match node's behavior. Node always resolves symlinks.

This fixes the case where you have an npm-linked dependency with its own dependencies. Today those grandchild dependencies are silently dropped.

I can do a test too. Does it make sense to extend nested-addons-smoke-test-slow so it includes a symlinked addon?

### DIFF
--- a/lib/models/addon-discovery.js
+++ b/lib/models/addon-discovery.js
@@ -9,6 +9,7 @@ const existsSync = require('exists-sync');
 const path = require('path');
 const resolve = require('resolve');
 const heimdall = require('heimdalljs');
+const fs = require('fs');
 
 /**
   AddonDiscovery is responsible for collecting information about all of the
@@ -198,7 +199,7 @@ class AddonDiscovery {
         logger.info(' - is addon, adding...');
         let addonInfo = {
           name: addonPkg.name,
-          path: addonPath,
+          path: fs.realpathSync(addonPath),
           pkg: addonPkg,
         };
         return addonInfo;
--- a/tests/fixtures/addon/simple/node_modules/.gitignore
+++ b/tests/fixtures/addon/simple/node_modules/.gitignore
@@ -1,0 +1,1 @@
+symlinked-addon
--- a/tests/unit/models/addon-discovery-test.js
+++ b/tests/unit/models/addon-discovery-test.js
@@ -7,6 +7,7 @@ const AddonDiscovery = require('../../../lib/models/addon-discovery');
 let fixturePath = path.resolve(__dirname, '../../fixtures/addon');
 const MockUI = require('console-ui/mock');
 const MockCLI = require('../../helpers/mock-cli');
+const fs = require('fs');
 
 describe('models/addon-discovery.js', function() {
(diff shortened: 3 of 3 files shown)


stefanpenner:
LGTM

> I can do a test too.

Ya, we should get a test.

> Does it make sense to extend nested-addons-smoke-test-slow so it includes a symlinked addon?

My preference would be to unit test this function, rather then a larger acceptance test.

rwjblue:
Is there a practical scenario where what we are doing now fails? Can you give me some steps to try locally to understand the problem you are trying to solve?

stefanpenner:
@ef4 should we actually realpath sooner, that way the debug logging is also consistent? Maybe here ? https://github.com/ember-cli/ember-cli/pull/6955/files#diff-6a115371fcdc66c67f9710569fa7c15eR204

stefanpenner:
@ef4 is the drop happening here?
https://github.com/ember-cli/ember-cli/blob/28100bbdc1222778072c9148a3693601522a60d4/lib/models/addon.js#L1582-L1588

ef4:
@stefanpenner I originally put it there, but that is outside the existsSync guard, so it blows up if the path doesn't exist.

ef4:
@rwjblue yes, there are practical failures.  Scenario:

1. You are developing addonA and you have it npm linked into an app.
2. In addonA you `ember install  -S addonB`.
3. You rebuild the app, but addonB is missing.

stefanpenner:
> My preference would be to unit test this function, rather then a larger acceptance test. That is unless, appending it to that existing test reasonably simple.

@ef4 got any cycles for ^ ?

ef4:
Yes, working on this.

I think I can do an end to end test that proves the important semantics
(that grandchild modules make it into the build) by extending one of the
existing scenarios (no additional builds that would make the tests slower).

ef4:
This is turning out to be complicated. I am getting entangled in the PackageCache (I can't readily tweak the contents of node_modules in one acceptance test without leaking that change).

stefanpenner:
@ef4 I think your best bet is to unit test the function is question, that should be simpler then the larger acceptance test and the `PackageCache` stuff.

ef4:
Yes, I am curious whether this will be happy in windows, what with a symlink checked into the repo.

stefanpenner:
> Yes, I am curious whether this will be happy in windows, what with a symlink checked into the repo.

It looks un-happy.

kellyselden:
Would it be possible to create the symlink during the test, maybe in project tmp/?

rwjblue:
This is getting stale, but still seems to be some thing that we need.  @ef4 - Any chance you can bring this back up to date with master and kick off another test?

ef4:
@rwjblue this is ready to go. 

Git on windows does not check out symlinks unless it runs in an admin shell. But node can make the symlink either way.

ef4:
github's status is wrong and Travis succeeded a long time ago.

PR #6963, opened 2017-04-17
Preserve header key case when serving with proxy

### LINKED ISSUE #6960: Preserve header key case when running server with proxy option
When running `ember server` with the `--proxy` option [http-proxy](https://github.com/nodejitsu/node-http-proxy) is used as a middleware for those requests. By default `http-proxy` lower cases all header keys. Ran into issue where I was expecting `X-REQUEST-ID` sent in uppercase from the API, but when developing using `--proxy` it introduced that header has `x-request-id` instead.

`http-proxy` has a `preserveHeaderKeyCase` option which when set to `true` "fixes" this. I can go ahead and create a PR with this if there's any interest.

---

```
ember-cli: 2.12.1
http_parser: 2.7.0
node: 7.6.0
v8: 5.5.372.40
uv: 1.11.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 51
openssl: 1.0.2k
icu: 58.2
unicode: 9.0
cldr: 30.0.3
tz: 2016j
os: darwin x64

4.3.0
```

jpadilla:
Still missing a test for this but not entirely sure where.

---

Fixes #6960

Recap: When running `ember server` with the `--proxy` option [http-proxy](https://github.com/nodejitsu/node-http-proxy) is used as a middleware for those requests. By default `http-proxy` lower cases all header keys. Ran into issue where I was expecting `X-REQUEST-ID` sent in uppercase from the API, but when developing using `--proxy` it introduced that header has `x-request-id` instead.

### DIFF
--- a/lib/tasks/server/middleware/proxy-server/index.js
+++ b/lib/tasks/server/middleware/proxy-server/index.js
@@ -17,6 +17,7 @@ class ProxyServerAddon {
         secure: options.secureProxy,
         changeOrigin: true,
         xfwd: options.transparentProxy,
+        preserveHeaderKeyCase: true,
       });
 
       proxy.on('error', e => {


rwjblue:
Seems ok to preserve the case, but I was under the impression that headers were supposed to be case insensitive?

rwjblue:
https://tools.ietf.org/html/rfc2616 and then https://tools.ietf.org/html/rfc7230 say:

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

jpadilla:
@rwjblue yup that's true, but I've seen many clients just lowering the case of each header key to "normalize". I noticed this when handling headers in an `ember-data` adapter's `handleResponse()` hook. My initial idea was to raise the issue over there and try to have `ember-data/-private/utils/parse-response-headers` lowercase header keys as well.

rwjblue:
Ya, and just to be clear, I'm not opposed to preserving whatever case was specified. I guess I'm just surprised that other tooling is case dependent...

stefanpenner:
> Still missing a test for this but not entirely sure where.

Ya not sure, this area doesn't have great test coverage. Although adding tests would be lovely (if you have cycles)

jpadilla:
@stefanpenner honestly might be a while before I can write any tests for this. I was following a [similar](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/server/middleware/tests-server-test.js) one, but `next` isn't called in the `proxy-server` middleware so the approach is a bit different. With a bit of guidance I'm sure I can get something done if need be.

jpadilla:
@Turbo87 thank you! 🎆

PR #6965, opened 2017-04-19
Fallback for toYarnArgs in npm-task.js produces invalid yarn command

pichfl:
This was found while debugging a yarn related issue in Ember Electron. This was not very visible as most authors would probably include either the save or the save-dev flag, but as we had a typo in there in lead to an interesting bug hunt.

The else clause produced a no longer valid yarn invocation as `yarn install` is no longer deprecated but removed.


pichfl:
Missing test, will update soonish.

PR #6966, opened 2017-04-19
ENHANCEMENT: throw when converting `npm install foo` to `yarn install foo`

pichfl:
This was found while debugging a yarn related issue in Ember Electron. This was not very visible as most authors would probably include either the save or the save-dev flag, but as we had a typo in there in lead to an interesting bug hunt.

`npm install foo` installs a package without saving it to package.json which is not an option in yarn (at least none that I know of). Therefore `yarn install foo` is no longer an equivalent command. 

This changed rather recently as `yarn` printed a deprecation warning pointing to `yarn add` before, but this is no longer the case in 0.23.2.

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -196,11 +196,14 @@ class NpmTask extends Task {
 
   toYarnArgs(command, options) {
     let args = [];
+
     if (command === 'install') {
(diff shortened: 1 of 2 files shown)


kellyselden:
LGTM. Thanks!

@homu r+

PR #6969, opened 2017-04-19
Fix #6964 - Remove `.bowerrc` from blueprints

### LINKED ISSUE #6964: Remove `.bowerrc` file from app blueprint.
A newly blueprinted app should not include any `bower` related files as of 2.13.0, however we are still generating a `.bowerrc` file.  To fix this we should remove `.bowerrc` [here](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.bowerrc) and add `.bowerrc` to the list of `filesToRemove` in both the [`app`](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/index.js#L8) and [`addon`](https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/index.js#L19) blueprints.

A PR doing this work should target the beta branch (which will become 2.13.0 final).

pichfl:


rwjblue:
Awesome, thank you!

rwjblue:
@homu r+

PR #6982, opened 2017-04-24
Adds yarn-error.log to gitignore in the app blueprint

MiguelMadero:
While updating an app to ember-cli@2.13.0-beta.4 I noticed all/most of the blueprint files prefer yarn over npm now, however, the gitignore was missing this.


MiguelMadero:
thanks!

Turbo87:
@rwjblue @stefanpenner I guess we should backport that to the `beta` branch, correct?

PR #7002, opened 2017-04-30
update from npm 2 when using node 4

kellyselden:
Addresses https://github.com/ember-cli/ember-cli/pull/6930#issuecomment-298063946

Only thing worth debating is this will skip npm 3 and go straight to 4. This seems like a good thing, since we also get Travis's latest yarn if we go the alternate route.

cc @Turbo87

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -35,7 +35,8 @@ install:
   - yarn install --no-lockfile
 <% } else { %>before_install:
   - npm config set spin false
-  - npm install -g phantomjs-prebuilt
+  - npm install -g npm phantomjs-prebuilt
+  - npm --version
   - phantomjs --version
 
 install:


stefanpenner:
Should we support only the latest NPM or the bundled NPM for a given version of node?

rwjblue:
@stefanpenner - Currently our other functionality (like `ember install`) and `ember new` will issue a warning when attempting to run with `npm@2`.

Turbo87:
@kellyselden can you also add the comment explaining why we use Node 4 for addon CI?

kellyselden:
> can you also add the comment explaining why we use Node 4 for addon CI?

@Turbo87 done in #7026

kellyselden:
Merging for now. We can always change it to npm 3 before release if there is opposition.

@homu r+

kellyselden:
@homu retry

kellyselden:
@homu retry

PR #7003, opened 2017-05-01
Allow node 7.x on Windows

btecu:
Windows users have been getting `WARNING: Node v7.9.0 is not tested against Ember CLI on your platform. We recommend that you use the most-recent "Active LTS" version of Node.js.` even on the latest versions because `appveyor.yml` versions was not updated when `travis.yml` was.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ environment:
   matrix:
     - nodejs_version: "4"
     - nodejs_version: "6"
+    - nodejs_version: "7"
 
 branches:
   only:


kellyselden:
LGTM. It's kinda crazy that we scan these files for that integrity check. This means that if we can't necessarily use AppVeyor as a Windows smoke test, it has to test every version of node to get that message to go away. As a result, way slower pull requests.

@homu r+

stefanpenner:
@kellyselden does appveyor have optional failures?

stefanpenner:
Should this be bugfix beta or bugfix release? cc @rwjblue / @Turbo87

kellyselden:
@stefanpenner Looks like it does https://www.appveyor.com/docs/build-configuration/#allow-failing-jobs. Another option would be to just use travis.yml for the Windows messaging.

stefanpenner:
@kellyselden I believe @nathanhammond design was for us to use optional failures for this (maybe he can shed some light). In general I do think connecting CI + supported versions is pretty handy.

Turbo87:
@stefanpenner IMHO this should be backported to `release`

PR #7007, opened 2017-05-01
Updated npm version for ember-data to use ~ instead of ^

fushi:
Users should explicitly move from minor version to minor version, rather than automatically jumping to whatever is newest.


stefanpenner:
I'm 👍

pichfl:
yarn helps doing this and provides a `--tilde` flag. `yarn upgrade-interactive --tilde` turned out to be very useful in our team.

PR #7011, opened 2017-05-02
npmTask should throw when trying to convert `npm install foo` to `yarn install foo`

Turbo87:
backport of https://github.com/ember-cli/ember-cli/pull/6966

/cc @pichfl


rwjblue:
@homu r+

PR #7019, opened 2017-05-02
CLI: fixup shutdown/interrupt code

stefanpenner:
- [x] fix tests
- [x] add regression test

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -84,8 +84,6 @@ class CLI {
    * @return {Promise}
    */
   run(environment) {
-    let shutdownOnExit = null;
-
     return RSVP.hash(environment).then(environment => {
       let args = environment.cliArgs.slice();
       let commandName = args.shift();
@@ -146,31 +144,27 @@ class CLI {
       }
 
       let instrumentation = this.instrumentation;
-      let onCommandInterrupt;
+      let removeInterruptionHandler;
 
-      let runPromise = Promise.resolve().then(() => {
+      return Promise.resolve().then(() => {
+        removeInterruptionHandler = onProcessInterrupt.addHandler(() => command.onInterrupt());
         instrumentation.stopAndReport('init');
         instrumentation.start('command');
 
         loggerTesting.info('cli: command.beforeRun');
-        onProcessInterrupt.addHandler(onCommandInterrupt);
 
         return command.beforeRun(commandArgs);
       }).then(() => {
         loggerTesting.info('cli: command.validateAndRun');
 
         return command.validateAndRun(commandArgs);
+      }).finally(() => {
+        removeInterruptionHandler();
       }).then(result => {
         instrumentation.stopAndReport('command', commandName, commandArgs);
-
-        onProcessInterrupt.removeHandler(onCommandInterrupt);
-
         return result;
       }).finally(() => {
         instrumentation.start('shutdown');
-        shutdownOnExit = function() {
-          instrumentation.stopAndReport('shutdown');
-        };
       }).then(result => {
         // if the help option was passed, call the help command
         if (result === 'callHelp') {
@@ -199,19 +193,10 @@ class CLI {
           }
         });
       });
-
-      onCommandInterrupt = () =>
-        Promise.resolve(command.onInterrupt())
-          .then(() => runPromise);
-
-      return runPromise;
-    })
-    .finally(() => {
-      if (shutdownOnExit) {
-        shutdownOnExit();
-      }
-    })
-    .catch(this.logError.bind(this));
+    }).finally(() => {
+      onProcessInterrupt.teardown();
+      this.instrumentation.stopAndReport('shutdown');
+    }).catch(this.logError.bind(this));
   }
 
   /**
--- a/tests/unit/cli/cli-test.js
+++ b/tests/unit/cli/cli-test.js
@@ -6,6 +6,7 @@ const MockAnalytics = require('../../helpers/mock-analytics');
 const td = require('testdouble');
 const Command = require('../../../lib/models/command');
 const Promise = require('rsvp').Promise;
+const SilentError = require('silent-error');
 
 let ui;
 let analytics;
@@ -65,10 +66,16 @@ function stubRun(name) {
 }
 
 describe('Unit: CLI', function() {
+  let teardownHandler;
   beforeEach(function() {
+    teardownHandler = td.function();
+    let teardown = td.function();
+    let addHandler = td.function();
+    td.when(addHandler(td.matchers.isA(Function))).thenReturn(teardownHandler);
+
     willInterruptProcess = td.replace('../../../lib/utilities/will-interrupt-process', {
-      addHandler: td.function(),
-      removeHandler: td.function(),
+      addHandler,
+      teardown,
     });
 
     CLI = require('../../../lib/cli/cli');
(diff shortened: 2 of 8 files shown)


tests/unit/cli/cli-test.js
```diff
@@ -228,6 +235,31 @@ describe('Unit: CLI', function() {
     });
   });
 
+  describe.only('command crash', function() {
```
chadhietala: Remove the `only`

stefanpenner: Ooo. Will fix

stefanpenner:
Windows CI seems unhappy, re-running. If not I can take a look tonight when i have a windows machine handy.

tests/unit/cli/cli-test.js
```diff
@@ -228,6 +235,31 @@ describe('Unit: CLI', function() {
...
+      });
+
+      project.eachAddonCommand = function(callback) {
+        callback('custom-addon', {
+          custom: CustomCommand,
+        });
+      };
+
+      return ember(['custom']).then(function() {
+        expect(false, 'should have rejected').to.eql(false);
```
ro0gr: Shouldn't one of these false be a true? Seems like this line will always pass if a promise isn't rejected.

stefanpenner: 👍

Turbo87: actually we should use https://github.com/domenic/chai-as-promised for tests like that. e.g.

```js
return expect(ember(['custom'])).to.be.rejected;
```

tests/unit/cli/cli-test.js
```diff
@@ -228,6 +235,31 @@ describe('Unit: CLI', function() {
...
+      });
+
+      project.eachAddonCommand = function(callback) {
+        callback('custom-addon', {
+          custom: CustomCommand,
+        });
+      };
+
+      return ember(['custom']).then(function() {
+        expect(false, 'should have rejected').to.eql(true);
```
stefanpenner: switch to chai-as-promised.

ro0gr:
I've tested this pr against several known issues:
 - https://github.com/ember-fastboot/ember-cli-fastboot/issues/386
 - https://github.com/emberjs/data/pull/4965

The good news is that these issues are addressed.

The bad news is that it breaks testem integration on windows.

I've noticed it in ["ember new foo, test, SIGINT exits with error and clears tmp/"](https://github.com/ember-cli/ember-cli/blob/93c4fd5763fa69fcb11eb0bee1b6b38e57ad4d15/tests/acceptance/smoke-test-slow.js#L302) test failure. The test on itself passes but [afterEach](https://github.com/ember-cli/ember-cli/blob/93c4fd5763fa69fcb11eb0bee1b6b38e57ad4d15/tests/acceptance/smoke-test-slow.js#L43) hook fails with the following message:

>  √ ember new foo, test, SIGINT exits with error and clears tmp/ (145694ms)
>   1) "after each" hook for "ember new foo, test, SIGINT exits with error and clears tmp/"
>
>  1 passing (4m)
 > 1 failing
 >
 > 1) Acceptance: smoke-test "after each" hook for "ember new foo, test, SIGINT exits with error and clears tmp/":
     AssertionError: expected "C:\Users\kudgo\workspace\ember-cli\tmp\some_cool_app_clone-KeX18ufj.tmp" to not exist

Literally `afterEach` is unable to delete temporary application made for test. The reason for this is that testem doesn't kill `phantomjs` process anymore. This process locks test application directory.
 
<details>
<summary>testem.log - before this pr</summary>

> info Test'em starting..
info Seeking for config file...
info Starting ci
info Starting server
info Waiting for tests.
info Start triggered test run.
info Running tests...
info Tests running.
info spawning: phantomjs - [ 'C:\\Users\\kudgo\\workspace\\ember-cli\\node_modules\\testem\\assets\\phantom.js',
  'http://localhost:7357/5383/tests/index.html?hidepassed' ]
info New client connected: PhantomJS 2.1 5383
info tryAttach PhantomJS 2.1 5383
info received SIGINT signal
info Closing browser PhantomJS.
WARN PhantomJS closed 1
info Process PhantomJS terminated. 1 null
</details>

<details>
<summary>testem.log - this pr</summary>

> info Test'em starting..
info Seeking for config file...
info Starting ci
info Starting server
info Waiting for tests.
info Start triggered test run.
info Running tests...
info Tests running.
info spawning: phantomjs - [ 'C:\\Users\\kudgo\\workspace\\ember-cli\\node_modules\\testem\\assets\\phantom.js',
  'http://localhost:7357/5096/tests/index.html?hidepassed' ]
info New client connected: PhantomJS 2.1 5096
info tryAttach PhantomJS 2.1 5096
info
</details>



In the second log we miss `received SIGINT signal` entry which causes the problem.

I'm in process of investigation. Any ideas would be awesome.

ro0gr:
It's not just a test setup issue. It also reproducible by manual verification

ro0gr:
> The bad news is that it breaks testem integration on windows.

After some testing it turned out that it breaks only with `phantomjs`. `Chrome` launcher is disposed properly.
Seems like some kind of timing issue. Wrapping of interruption handler in timeout "fixes" the problem:
```js
        removeInterruptionHandler = onProcessInterrupt.addHandler(() => {
          return new RSVP.Promise(function(resolve) {
            setTimeout(() => {
              command.onInterrupt();
              resolve();
            }, 1000);
          });
```
But it seems like a bad solution.

I've checked how does "ctrl+c" work against `ember-cli@2.12` the problem persist there. The reason why this issue didn't appear earlier in tests is that we didn't have smoke test for `ember test` interruption.

This pr looks good for me but if we merge most likely it will break windows CI.

lib/cli/cli.js
```diff
@@ -199,19 +193,10 @@ class CLI {
...
-      return runPromise;
-    })
-    .finally(() => {
-      if (shutdownOnExit) {
-        shutdownOnExit();
-      }
-    })
-    .catch(this.logError.bind(this));
+    }).finally(() => {
+      onProcessInterrupt.teardown();
```
ro0gr: What is the reason for additional `teardown` here? https://github.com/ember-cli/ember-cli/pull/7019/files#diff-16a41bfbe414387d773df099f941217eR162 should do all the necessary cleanup?

ro0gr:
@stefanpenner I believe now with headless chrome this should pass under appveyor

kellyselden:
@ro0gr Rebased on master.

ro0gr:
@kellyselden great!

There is one failing test which breaks [here ](https://github.com/ember-cli/ember-cli/blob/7274291ae515c53a5de6924dfe27797568e7b20c/tests/unit/utilities/will-interrupt-process-test.js#L191).

I think it should be re-written to match updated `will-interrupt-process` api. 

**Before**:
```js
      // ...
      removewillInterruptProcess.addHandler(cb);
      td.verify(process.stdin.setRawMode(true));

      willInterruptProcess.removeHandler(cb);
      td.verify(process.stdin.setRawMode(false));

      td.verify(process.stdin.setRawMode(), {
        ignoreExtraArgs: true,
        times: 2,
      });
```

**After**:
```js
      // ...
      let removeHandler = willInterruptProcess.addHandler(cb); // `addHandler` now yields `removeHandler`
      td.verify(process.stdin.setRawMode(true));

      removeHandler(cb);
      td.verify(process.stdin.setRawMode(false));

      td.verify(process.stdin.setRawMode(), {
        ignoreExtraArgs: true,
        times: 2,
      });
```
I believe this should work.

-----
Saying the truth I find this test fragile. We could test it much more better if merge https://github.com/ember-cli/ember-cli/pull/7086 first.

stefanpenner:
@ro0gr its merged!

Turbo87:

PR #7028, opened 2017-05-05
add yarn missing default comment

kellyselden:
To match the default [here](https://github.com/ember-cli/ember-cli/blob/d3df9ffbfbcdf31e704f1c0dd8bb868927dc35aa/lib/commands/addon.js#L16)

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -23,7 +23,7 @@ module.exports = Command.extend({
     { name: 'blueprint',  type: String,                  aliases: ['b'] },
     { name: 'skip-npm',   type: Boolean, default: false, aliases: ['sn'] },
     { name: 'skip-bower', type: Boolean, default: false, aliases: ['sb'] },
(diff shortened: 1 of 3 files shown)


lib/commands/init.js
```diff
@@ -23,7 +23,7 @@ module.exports = Command.extend({
     { name: 'blueprint',  type: String,                  aliases: ['b'] },
     { name: 'skip-npm',   type: Boolean, default: false, aliases: ['sn'] },
     { name: 'skip-bower', type: Boolean, default: false, aliases: ['sb'] },
-    { name: 'yarn',       type: Boolean },
+    { name: 'yarn',       type: Boolean, default: false },
```
kellyselden: Alright, I'll update it with a comment instead.

kellyselden: @rwjblue What do you think now?

rwjblue:
@homu r+

PR #7033, opened 2017-05-08
Ensure addon blueprint calls `this.filesPath`

mike183:
Ensures that the default `addon` blueprint honours the `filesPath()` method that was introduced in #4837

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -125,7 +125,9 @@ module.exports = {
 
   files() {
     let appFiles = this.lookupBlueprint('app').files();
-    let addonFiles = walkSync(path.join(this.path, 'files'));
+
+    let addonFilesPath = this.filesPath(this.options);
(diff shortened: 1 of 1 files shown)


stefanpenner:
this will require a test + docs

mike183:
Wouldn't this be covered by the tests/documentation for the base blueprint model (where the method is actually defined)?

rwjblue:
Ya, this isn't changing any functionality, just honors the contract that is already there.  That being said, we do probably need a test to confirm that (for example) you could extend from the addon blueprint and have things "just work".

Turbo87:
CI failure seems unrelated

Turbo87:
@mike183 could you rebase this so that we can merge it?

mike183:
@Turbo87 Rebased.

Sorry for letting this sit, never managed to find the time to look at adding a new test.

mike183:
CI is red, unsure if the failure is related though.

PR #7036, opened 2017-05-10
Ensure `lintTree` results cannot clobber tests.

rwjblue:
The prior logic resulted in the result of `lintTree` being **after** the actual processed `tests/` tree. When those trees were then passed to `MergeTrees` the `lintTree` result would clobber the normal processed `tests/` tree. In general, when a linter is actually emitting linting tests, this is fine because the files that it emits are not the same names as the original (e.g. `tests/test-helper.js` becomes `tests/test-helper.lint-test.js`), but when the `lintTree` implementation does not emit different names than the input tree the `lintTree` versions "win".

In Ember CLI < 2.13, this setup still worked properly because we would fully process the result of `lintTree` through the consuming applications JS processor pipeline (e.g. babel). This _sounds_ neat but actually means that we were transpiling all of `tests/` twice.

In Ember CLI 2.13 and above, we only process the result of `lintTree` for modules (not a full "transpile down to ES5"). This means that any babel processing that would have been done for `tests/` normally is thrown away when the transpiled versions are clobbered by the untranspiled `lintTree`.

This change corrects the ordering of trees, so that the correctly proccessed `tests/` tree "wins" when the `lintTree` tree emits files with the same names.

### DIFF
--- a/tests/unit/broccoli/ember-app/test-test.js
+++ b/tests/unit/broccoli/ember-app/test-test.js
@@ -1,0 +1,97 @@
+'use strict';
+
+const co = require('co');
+const broccoliTestHelper = require('broccoli-test-helper');
+const expect = require('chai').expect;
+
+const EmberApp = require('../../../../lib/broccoli/ember-app');
+const MockCLI = require('../../../helpers/mock-cli');
+const Project = require('../../../../lib/models/project');
+
+const buildOutput = broccoliTestHelper.buildOutput;
+const createTempDir = broccoliTestHelper.createTempDir;
+const walkSync = require('walk-sync');
+const stew = require('broccoli-stew');
+
+describe('EmberApp#test', function() {
+  let input;
+
+  beforeEach(co.wrap(function *() {
+    input = yield createTempDir();
+
+    input.write({
+      'config': {
+        'environment.js': `module.exports = function() { return { modulePrefix: 'test-app' }; };`,
+      },
+    });
+  }));
+
+  afterEach(function() {
+    return input.dispose();
+  });
+
+  function createApp(options) {
+    options = options || {};
+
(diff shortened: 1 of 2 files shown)


lukemelia:
Nice work, @rwjblue! Thanks for tracking this down!

tests/unit/broccoli/ember-app/test-test.js
```diff
@@ -0,0 +1,95 @@
...
+          }
+        }
+      }
+    });
+
+    let app = createApp();
+
+    app.project.addons.push({
+      lintTree(type, tree) {
+        return stew.map(tree, function(string) {
```
stefanpenner: `=>` ?

rwjblue: Fixed (along with a few other linting errors).

tests/unit/broccoli/ember-app/test-test.js
```diff
@@ -0,0 +1,92 @@
...
+      },
+    });
+
+    let app = createApp();
+
+    app.project.addons.push({
+      lintTree(type, tree) {
+        return stew.map(tree, string => string.toUpperCase());
+      },
+    });
```
Turbo87: what is this doing? could use a short comment explaining it.

rwjblue: Good idea, will do.

rwjblue:
Updated with more comments around the test adding a fake `lintTree` (for observing side effects).  Also confirmed that without the changes made to `lib/broccoli/ember-app.js` the test fails in the same way as was reported (just to make sure the test was emulating the right thing).

rwjblue:
@homu r+

Issue #7035, opened 2017-05-09
test-support and addon-test-support don't lint files


trentmwillis:
I've been working on adding some linting support to an addon and noticed that test-support and addon-test-support trees don't seem to be linted.

For addons, it looks like we would need to add `addon-test-support` into the [`jshintAddonTree` function](https://github.com/ember-cli/ember-cli/blob/2afca8eee9393026b5af6ec02157b7c53cb67feb/lib/models/addon.js#L1097-L1121) (which should probably be renamed).

For applications, it looks like we should add `test-support` into the [`lintTestTrees` function](https://github.com/ember-cli/ember-cli/blob/2afca8eee9393026b5af6ec02157b7c53cb67feb/lib/broccoli/ember-app.js#L1182-L1185).

If the above seems good, I can work on it at some point soon.

stefanpenner:
> If the above seems good, I can work on it at some point soon.

👍

rwjblue:
`test-support` isn't a thing for apps

rwjblue:
Also, FWIW, I think this is basically a duplicate of https://github.com/ember-cli/ember-cli/issues/6667.  Though that one doesn't mention `test-support` (and we definitely should lint it too).

rwjblue:
Fixed by #7037 which was rebased / tweaked / landed in #7378.

PR #6668, opened 2017-01-12
Add `app` directory to lintJsTrees

kiwiupover:
Closes #6667 and https://github.com/ember-fastboot/ember-cli-fastboot/issues/307

PR #7037, opened 2017-05-10
Lint addon-test-support and test-support trees

trentmwillis:
Fixes https://github.com/ember-cli/ember-cli/issues/7035.

I'd like to add a test for this, but am unsure where to do so. Can someone recommend which test module to put it in?

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1180,38 +1180,35 @@ class EmberApp {
     @return {Array}
    */
   lintTestTrees() {
-    let lintedApp = this.addonLintTree('app', this._filterAppTree());
-    let lintedTests = this.addonLintTree('tests', this.trees.tests);
-    let lintedTemplates = this.addonLintTree('templates', this._templatesTree());
-
     let testSupportTree = this._mergeTrees(this.addonTreesFor('test-support'));
-    let lintedTestSupport = this.addonLintTree('test-support', testSupportTree);
 
-    lintedApp = processModulesOnly(new Funnel(lintedApp, {
+    return [
+      this.lintTestTree('app', this._filterAppTree()),
+      this.lintTestTree('tests', this.trees.tests),
+      this.lintTestTree('templates', this._templatesTree()),
+      this.lintTestTree('test-support', testSupportTree),
+    ].filter(Boolean);
+  }
+
+  /**
+    Lints the given tree, handling proper relocation of the files and processing of the modules. Used by `lintTestTrees`.
+
+    @private
+    @method lintTestTree
+    @param {String} treeName
+    @param {Tree} tree
+    @return {Tree}
+   */
+  lintTestTree(treeName, tree) {
+    let lintedTree = this.addonLintTree(treeName, tree);
+    let relocatedLintedTree = new Funnel(lintedTree, {
       srcDir: '/',
       destDir: `${this.name}/tests/`,
-      annotation: 'Funnel (lint app)',
-    }), 'Babel: lintTree(app)');
+      annotation: `Funnel (lint ${treeName})`,
+    });
+    let processedLintedTree = processModulesOnly(relocatedLintedTree, `Babel: lintTree(${treeName})`);
 
-    lintedTests = processModulesOnly(new Funnel(lintedTests, {
-      srcDir: '/',
-      destDir: `${this.name}/tests/`,
-      annotation: 'Funnel (lint tests)',
-    }), 'Babel: lintTree(tests)');
-
-    lintedTemplates = processModulesOnly(new Funnel(lintedTemplates, {
-      srcDir: '/',
-      destDir: `${this.name}/tests/`,
-      annotation: 'Funnel (lint templates)',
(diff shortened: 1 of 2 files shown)


stefanpenner:
Should we make this list pluggable? @kratiahuja I suspect would be happy about this for the additional fastboot trees.

rwjblue:
@trentmwillis - Please see https://github.com/ember-cli/ember-cli/pull/7036, which adds some rudimentary tests in this area.

lib/broccoli/ember-app.js
```diff
@@ -1180,29 +1180,35 @@ class EmberApp {
...
-    lintedApp = processModulesOnly(new Funnel(lintedApp, {
-      srcDir: '/',
-      destDir: `${this.name}/tests/`,
-      annotation: 'Funnel (lint app)',
-    }), 'Babel: lintTree(app)');
+    return [
+      this.lintTestTree('app', this._filterAppTree()),
+      this.lintTestTree('tests', this.trees.tests),
+      this.lintTestTree('templates', this._templatesTree()),
+      this.lintTestTree('test-support', testSupportTree),
```
rwjblue: This isn't quite right.  `lib/broccoli/ember-app.js` is for the consuming app this should be done in the addon, not here.

rwjblue: See https://github.com/ember-cli/ember-cli/pull/6668 for more detail on the approach that I'm suggesting.

trentmwillis: Bah, okay. I was following the existing pattern since `app` isn't linted in the addon.

rwjblue: Ya, but the way this is done now would only add linting for `tests/dummy/app/` and `tests/dummy/test-support`, which isn't what you are going for.

trentmwillis: That's not accurate. The implementation here collects the `test-support` trees from the various addons, so it does pick up what I intended (note: I also tested this locally).

rwjblue: I'd say we should likely just categorize this in with the build/rebuild work that has been underway for a while. We should continue reducing/removing LI specific customizations in this area (which is a major part of the core/extended refactoring work).

Closing...

rwjblue: I see, sorry about missing that part. Putting this here also has the issue of linting all addon's versions of `test-support`, instead of only the addon under test (generally done via `isDevelopingAddon` check).

trentmwillis: Yep yep. It definitely makes more sense for the addon to handle it.

lib/models/addon.js
```diff
@@ -1097,17 +1093,48 @@ let addonProto = {
...
+
+      let addonTemplates = this._addonTemplateFiles(addonPath);
+      let lintTemplateTrees = this._eachProjectAddonInvoke('lintTree', ['templates', addonTemplates]);
+
+      lintTrees = lintTrees.concat(lintAddonJsTrees, lintTemplateTrees);
+    }
+
+    let appPath = this._treePathFor('app');
+    if (existsSync(appPath)) {
+      let appJs = this.appJsFiles(appPath);
```
rwjblue: Do we need to filter by extension?  Would this work: `this._eachProjectAddonInvoke('lintTree', ['app', appPath])`?

kratiahuja:
Yes would love for this to be pluggable to allow external trees to be be able linted as well :)

lib/models/addon.js
trentmwillis: We need it because there could be styles or templates in the `app` directory.

trentmwillis:
> Some of the linted things are namespaced (e.g. always in this.moduleName() + '/test-support), but others aren't. It would be nice if they were consistent.

Yeah, I was unsure how best to handle it. If we expect this to only be run from the addon itself, then we don't need the namespacing. But, if it ever gets run with another app, then you could run into conflicts.

> Should we make this list pluggable?

Yes, but I'd like to punt on that for another PR.

lib/models/addon.js
rwjblue: Hmm. I think it wouldn't be harmful to not filter, but it doesn't matter much.

rwjblue:
> If we expect this to only be run from the addon itself, then we don't need the namespacing

Ya, I kinda like the namespacing, but I'd prefer to avoid adding more prototype methods for them.

lib/models/addon.js
trentmwillis: I guess it would depend on how the linting plugin works, right? If it assumes it only receives JS files then it'd be bad, but otherwise it's probably okay.

rwjblue: Ya, most of them are `broccoli-persistent-filter`'s with `extensions = ['js']` so it wouldn't really matter...

Turbo87:
@trentmwillis @rwjblue what's the status of this PR?

trentmwillis:
@Turbo87 I'm planning to get to this soon, the past week has been a bit crazy :)

rwjblue:
Rebased / tweaked / landed in #7378.

PR #7038, opened 2017-05-11
Update "ember-cli-htmlbars"

stefanpenner:


Turbo87:
@stefanpenner why is this targeted at `release`? what bug is it fixing?

stefanpenner:
I believe it may address: https://github.com/emberjs/data/pull/4947 

I can target beta instead... would love your and @rwjblue's opinion.

rwjblue:
Beta seems ok, but ultimately the ember data issue is resolved with an update in ember-data itself.

stefanpenner:
updated to beta

stefanpenner:
after chatting with @rwjblue we should do this a release, as addons testing blueprints will be affected.

Turbo87:
@stefanpenner it's still not clear to me what problem this is solving. can you elaborate on that?

stefanpenner:
@Turbo87 https://github.com/ember-cli/ember-cli-htmlbars/pull/110 causes issues for addons testing blueprints. Ember-data ran into it.

If we upgrade ^ this goes away, We may also want to revert that PR. I can't remember what @rwjblue and me decided.. maybe he can refresh my memory.

stefanpenner:
@rwjblue apparently already reverted the ember-cli-htmlbars stuff. I'll switch this back to beta.

Turbo87:
https://github.com/ember-cli/ember-cli-htmlbars/pull/110 is already included in v1.3.2 though, so not necessarily a need for the major bump 🤔

PR #7041, opened 2017-05-14
[BUGFIX release] Revert rawMode to original value during windows signals cleanup

ro0gr:
I've noticed `ember-data` node-tests hangs forever without an ability to _CTRL+C_ while running on Windows.

The reason for this is that `will-interrupt-process` doesn't revert original `isRaw`. This pr includes reverting of the `rawMode` to its original value which allows you to kill the process with _CTRL+C_

**How to test**:
First you need Windows. 
Put this script in the root of any ember app or addon:
```js
// win-hang.js
const ember = require('ember-cli/tests/helpers/ember');

// this ends up with an exception in `ember` test helper
ember(['g', 'ember']);
```
And then run it `node win-hang.js`.

**Expected**: the process finishes
**Actual**: the process hangs. Unable to kill the process even by pressing _CTRL+C_

tested on win10.

### DIFF
--- a/lib/utilities/will-interrupt-process.js
+++ b/lib/utilities/will-interrupt-process.js
@@ -8,7 +8,7 @@
 let exit = require('capture-exit');
 exit.captureExit();
 
-let windowsCtrlCTrap;
+let windowsCtrlCTrap, originalIsRaw;
 let handlers = [];
 
 module.exports = {
@@ -81,7 +81,8 @@ function setupSignalsTrap() {
   process.on('SIGINT', _exit);
   process.on('SIGTERM', _exit);
   process.on('message', onMessage);
-  if (/^win/.test(process.platform)) {
+
+  if (isWindowsTTY()) {
     trapWindowsSignals();
   }
 }
@@ -95,7 +96,8 @@ function teardownSignalsTrap() {
   process.removeListener('SIGINT', _exit);
   process.removeListener('SIGTERM', _exit);
   process.removeListener('message', onMessage);
-  if (/^win/.test(process.platform)) {
+
+  if (isWindowsTTY()) {
     cleanupWindowsSignals();
   }
 }
@@ -106,22 +108,31 @@ function teardownSignalsTrap() {
  * @method trapWindowsSignals
  */
 function trapWindowsSignals() {
+  const stdin = process.stdin;
+
+  originalIsRaw = stdin.isRaw;
+
   // This is required to capture Ctrl + C on Windows
-  if (process.stdin && process.stdin.isTTY) {
-    process.stdin.setRawMode(true);
-    windowsCtrlCTrap = function(data) {
-      if (data.length === 1 && data[0] === 0x03) {
-        process.emit('SIGINT');
-      }
-    };
-    process.stdin.on('data', windowsCtrlCTrap);
-  }
+  stdin.setRawMode(true);
+
+  windowsCtrlCTrap = function(data) {
+    if (data.length === 1 && data[0] === 0x03) {
+      process.emit('SIGINT');
+    }
+  };
+  stdin.on('data', windowsCtrlCTrap);
 }
 
 function cleanupWindowsSignals() {
-  if (windowsCtrlCTrap && process.stdin.removeListener) {
-    process.stdin.removeListener('data', windowsCtrlCTrap);
-  }
+  const stdin = process.stdin;
+
+  stdin.setRawMode(originalIsRaw);
+
+  stdin.removeListener('data', windowsCtrlCTrap);
+}
+
+function isWindowsTTY() {
+  return (/^win/).test(process.platform) && process.stdin && process.stdin.isTTY;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ro0gr:
I think it's ready for review @stefanpenner @hjdivad @rwjblue 

seems like travis node@4 failure is related to some eslint timeout. I'm not sure what it's saying to me. Can anybody restart the build please?

ro0gr:
To be clear `ember-data` node-tests still hang on my windows machine. But with this change it's possible to successfuly `CTRL+C` now

tests/unit/utilities/will-interrupt-process-test.js
```diff
@@ -160,16 +160,43 @@ describe('will interrupt process', function() {
...
+      });
+
+      willInterruptProcess.addHandler(cb);
+      td.verify(process.stdin.setRawMode(true));
+
+      willInterruptProcess.removeHandler(cb);
+      td.verify(process.stdin.setRawMode(false));
+
+      td.verify(process.stdin.setRawMode(), {
+        ignoreExtraArgs: true,
```
ro0gr: After quick check of `testdouble` docs I couldn't find a simple way to verify if `setRawMode` was called exactly 2 times, the first one with `true` and second one with `false`. 

I ended up with this approach. Incoming arguments and invocation order is tested above on the lines 189 and 192. The line with `ignoreExtraArgs` ensures that we call `setRawMode` exactly 2 times. 

I'm glad to improve here but not sure how to better express it with `testdouble`.

lib/utilities/will-interrupt-process.js
```diff
@@ -106,24 +108,37 @@ function teardownSignalsTrap() {
...
+  };
+  stdin.on('data', windowsCtrlCTrap);
 }
 
 function cleanupWindowsSignals() {
-  if (windowsCtrlCTrap && process.stdin.removeListener) {
-    process.stdin.removeListener('data', windowsCtrlCTrap);
+  const stdin = process.stdin;
+
+  if (typeof originalIsRaw === 'boolean') {
```
ro0gr: You are right. After `isWindowsTTY` was extracted this check become redundant.

lib/utilities/will-interrupt-process.js
```diff
@@ -106,24 +108,37 @@ function teardownSignalsTrap() {
...
+    stdin.setRawMode(originalIsRaw);
+  }
+
+  if (windowsCtrlCTrap) {
```
ro0gr: redundant

tests/unit/utilities/will-interrupt-process-test.js
```diff
@@ -160,16 +160,43 @@ describe('will interrupt process', function() {
...
-      td.verify(process.stdin.setRawMode(true));
       td.verify(process.stdin.on('data', windowsCtrlCTrap));
     });
 
+    it('adds and reverts rawMode on Windows', function() {
+      Object.defineProperty(process, 'platform', {
+        value: 'win',
+      });
+
+      Object.defineProperty(process, 'stdin', {
```
ro0gr: Sounds good to me.

tests/unit/utilities/will-interrupt-process-test.js
hjdivad: Ah yeah I see what you're doing.  I don't know how to make this any clearer with `testdouble` either.  This seems 👍  then.

ro0gr:
@hjdivad I think it's ready for review.
You were right, fake `process` allows to test `will-interrupt-process` much more better.

tests/unit/utilities/will-interrupt-process-test.js
ro0gr: I've tried to make it. But it messes up the pr. I will iterate on this in a separate pr.

hjdivad:
@ro0gr LGTM.  Is there anything else you want to do on this PR before it's merged?

stefanpenner:
I would like to also review this before we merge (if I don't tonight, then ignore my request :))

ro0gr:
@hjdivad yes, I would like to allow a customizable  process for will-interrupt-process in this pr before it gets merged.

hjdivad:
@ro0gr sounds good, although personally I think it would be fine for that to come in a followup pr.

ro0gr:
@hjdivad I'm ok to merge this. My original intention is fulfilled here.

hjdivad:
@stefanpenner did you get a chance to take a look?

stefanpenner:
We may want this to be a `[bugfix beta]` thoughts @Turbo87 ?

Turbo87:
I'm not familiar with the details of this PR, but if it's a bugfix it should probably be on `release`?

ro0gr:
Yes, it's a fix for bug introduced in 2.13

ro0gr:
I'll change pr title in few hours

ro0gr:
Can this pr get merged?

rwjblue:
Thank you @ro0gr!

PR #7043, opened 2017-05-16
Prevent duplicate AddonDiscovery work

stefanpenner:
Apps with nested add-ons and many engines are negatively affected by duplicate add-on discovery work.

This freezes + caches the return values, preventing much duplicate work.

### DIFF
--- a/lib/models/addon-discovery.js
+++ b/lib/models/addon-discovery.js
@@ -11,7 +11,10 @@ const path = require('path');
 const resolve = require('resolve');
 const findup = require('find-up');
 const heimdall = require('heimdalljs');
-
+const PACKAGE_CACHE = Object.create(null);
+const AT_PATH_CACHE = Object.create(null);
+const FROM_DEPS_CACHE = Object.create(null);
+const FROM_IN_REPO_DEPS_CACHE = Object.create(null);
 /**
   AddonDiscovery is responsible for collecting information about all of the
   addons that will be used with a project.
@@ -20,7 +23,7 @@ const heimdall = require('heimdalljs');
   @extends CoreObject
   @constructor
 */
-class AddonDiscovery {
+module.exports = class AddonDiscovery {
   constructor(ui) {
     this.ui = ui;
   }
@@ -110,8 +113,15 @@ class AddonDiscovery {
     @method discoverFromDependencies
    */
   discoverFromDependencies(root, nodeModulesPath, pkg, excludeDevDeps) {
+    let key = `${root}|${nodeModulesPath}|${pkg.name}|${excludeDevDeps}`;
+    let entry = FROM_DEPS_CACHE[key];
+
+    if (entry !== undefined) {
+      return entry;
+    }
+
     let discovery = this;
-    let addons = Object.keys(this.dependencies(pkg, excludeDevDeps)).map(function(name) {
+    let addons = Object.freeze(Object.keys(this.dependencies(pkg, excludeDevDeps)).map(function(name) {
       if (name !== 'ember-cli') {
         let addonPath = this.resolvePackage(root, name);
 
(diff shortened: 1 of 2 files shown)


lib/models/project.js
```diff
@@ -394,6 +394,8 @@ class Project {
     @method discoverAddons
    */
   discoverAddons() {
+    if (this.addonPackages) { this.addonPackages; }
```
lukemelia: Missing `return`?

lukemelia:
Good idea!

lib/models/addon-discovery.js
```diff
@@ -55,6 +59,7 @@ class AddonDiscovery {
     let addons = projectAsAddon.concat(cliAddons, internalAddons, dependencyAddons, inRepoAddons);
 
     token.stop();
+    console.timeEnd(project.name());
```
hjdivad: stray `console.{time,timeEnd}`

stefanpenner:
> Sweet, looks like it mainly just needs tests. ^_^

Yup, got some cleanup todo and tests to write.

lib/models/addon-discovery.js
```diff
@@ -11,7 +11,10 @@ const path = require('path');
 const resolve = require('resolve');
 const findup = require('find-up');
 const heimdall = require('heimdalljs');
-
+const PACKAGE_CACHE = Object.create(null);
```
rwjblue: I'm concerned with these caches being in module scope without a mechanism to reset them.  The most common thing that I can think of that this might be an issue for is when leveraging ember-cli as a library to test things like blueprints. In this scenario, it is not unreasonable to update the layout of `node_modules` between tests and expect discovery to "just work".

lib/models/addon-discovery.js
```diff
@@ -131,18 +141,25 @@ class AddonDiscovery {
           return addon;
         }
       }
-    }, this).filter(Boolean);
-    return addons;
+    }, this).filter(Boolean));
+
+    return (FROM_DEPS_CACHE[key] = addons);
```
rwjblue: Silly nitpick, but my preference would be either:

```js
return FROM_DEPS_CACHE[key] = addons;
```

or 

```js
FROM_DEPS_CACHE[key] = addons;
return addons;
```

lib/models/addon-discovery.js
rwjblue: I'd recommend exposing a way to reset the caches, and likely invoke that when `Project.prototype.initializeAddons` is called for the first time for each project instance...

stefanpenner: unsure, as when `require` is invoked the result is frozen. require'ing again with a different node_module topology is already not going to work.

If you can provide some legitimate scenarios where this is expected to work?

rwjblue: It isn't about this file itself being required though, its this caching currently assumes that during the life of the process none of the addon discovery can change for the same input paths.  This isn't true when (for example) using something like our own test runner where we would generate / regenerate files in node_modules with different contents (e.g. via `input.write` from `broccoli-test-helper`).

stefanpenner: I suspect our own tests rely on this. I'll make the cache live as long as the current project, and a cache per project. I think that should cover it?

rwjblue: Yep, that's what I suggested above.

lib/models/project.js
```diff
@@ -394,6 +394,10 @@ class Project {
     @method discoverAddons
    */
   discoverAddons() {
+    if (this.addonPackages) {
```
stefanpenner: @dcombslinkedin ^^

dcombslinkedin:
Thanks, Stef!
________________________________

Turbo87:

PR #7045, opened 2017-05-16
Make app.import() work with files inside `node_modules`

Turbo87:
Instead of funneling all of `node_modules` we build a list of npm packages that we want to funnel, and then only use those to reduce the necessary IO work.

### DIFF
--- a/tests/unit/broccoli/ember-app/import-test.js
+++ b/tests/unit/broccoli/ember-app/import-test.js
@@ -143,4 +143,79 @@ describe('EmberApp.import()', function() {
     expect(Object.keys(outputTree['assets']).sort()).to.deep.equal(['app-import.js', 'vendor.js']);
     expect(outputTree['assets']['vendor.js']).to.contain('verysmallmoment');
   }));
+
+  it('can import node dependencies into vendor.js', co.wrap(function *() {
+    input.copy(`${__dirname}/../../../fixtures/app-import`);
+    input.write({
+      'node_modules': {
+        'moment': {
+          'moment.js': 'window.moment = "what does time even mean?";',
+        },
+      },
+    });
+
+    let app = createApp();
+
+    app.import('node_modules/moment/moment.js');
+
+    let output = yield buildOutput(app.javascript());
+    let outputTree = output.read();
+    expect(Object.keys(outputTree)).to.deep.equal(['assets']);
+    expect(Object.keys(outputTree['assets']).sort()).to.deep.equal(['app-import.js', 'vendor.js']);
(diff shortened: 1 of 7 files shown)


stefanpenner:
Does this need to take into account the node algorithm to not be self-trollish?

Turbo87:
> Does this need to take into account the node algorithm to not be self-trollish?

I guess that would make sense, yeah. I have no clue how to test this though... 😞

tests/unit/broccoli/ember-app/import-test.js
```diff
@@ -0,0 +1,221 @@
...
+      sourcemaps: { enabled: false },
+    }, options);
+
+    app._processedTemplatesTree = () => '';
+
+    return app;
+  }
+
+  it('does not import bower dependencies if they are not explicitly imported', co.wrap(function *() {
+    input.copy(`${__dirname}/../../../fixtures/new-app`);
```
rwjblue: We shouldn't embed a full `ember new foo --skip-npm --skip-bower` into `tests/fixtures/`. If we need that (and it seems fine to generally have it), then we should generate it (the generator doesn't take long to run) or clone it from `ember-new-output`.

Turbo87:
@rwjblue @stefanpenner it's 🍏 now :)

tests/unit/broccoli/ember-app/import-test.js
```diff
@@ -0,0 +1,224 @@
...
+    return createTempDir().then(tempDir => (input = tempDir));
+  });
+
+  afterEach(function() {
+    process.chdir(cwd);
+    delete process.env.EMBER_ENV;
+    return input.dispose();
+  });
+
+  function createApp(options) {
```
rwjblue: Can you make an issue to refactor a shared create app helper, so we can remove the duplicates?

Turbo87:
@homu r+

@rwjblue done (https://github.com/ember-cli/ember-cli/issues/7050)

Turbo87:
@dfreeman identified multiple potential issues with this:

- scoped packages (eg. `@glimmer/foobar`) are causing all of `@glimmer` to be funneled instead of just `@glimmer/foobar` 

- calling `app.import()` with multiple files from the same package is causing the package to be funneled multiple times

nightire:
@Turbo87 BTW, `app.import` should also support wrap guard clause for FastBoot compatibility, otherwise users still have to funnel package(s) by hand.

rwjblue:
@nightire - seems unrelated to this PR.  There are other open PR's adding this functionality, perhaps you can review them and comment over there?

Issue #6999, opened 2017-04-30
removePackageFromProject fails with yarn if package is not present


blimmer:
I noticed this while switching a project over to ember-cli-mocha from ember-cli-qunit. Exact repro steps from a new ember-cli project:

1. `ember new my-app`
2. `cd my-app`
3. `yarn install` (to get lockfile)
4. `yarn remove ember-cli-qunit`
5. `ember install ember-cli-mocha`

When `yarn remove` is run, it throws an exception that breaks running the addon blueprint. Here's [the line in the blueprint](https://github.com/ember-cli/ember-cli-mocha/blob/c08eb9280296c2cf2afaf3d9cf661c5123925a52/blueprints/ember-cli-mocha/index.js#L17) that's running:

```
return addonContext.removePackageFromProject('ember-cli-qunit');
```

```
⠸ Yarn: Installing ember-cli-chai@^0.3.0 ...
Yarn: Installed ember-cli-chai@^0.3.0
  uninstall package ember-cli-qunit
Command failed: yarn remove ember-cli-qunit
error This module isn't specified in a manifest.
yarn remove v0.23.3
[1/2] Removing module ember-cli-qunit...
info Visit https://yarnpkg.com/en/docs/cli/remove for documentation about this command.

Error: Command failed: yarn remove ember-cli-qunit
error This module isn't specified in a manifest.
yarn remove v0.23.3
[1/2] Removing module ember-cli-qunit...
info Visit https://yarnpkg.com/en/docs/cli/remove for documentation about this command.

    at Promise.all.then.arr (/Users/blimmer/code/raffle/node_modules/execa/index.js:210:11)
    at process._tickCallback (internal/process/next_tick.js:109:7)
```

---

Output from `ember version --verbose && npm --version`:
```
ember-cli: 2.13.0
http_parser: 2.7.0
node: 6.10.2
v8: 5.1.281.98
uv: 1.9.1
zlib: 1.2.11
ares: 1.10.1-DEV
icu: 58.2
modules: 48
openssl: 1.0.2k
os: darwin x64
3.10.10
```

rwjblue:
Thanks for reporting! The question in my mind is which side has the "bug". Specifically, should we consider `removePackageFromProject` to be more of a "ensure package removed from project" (and idempotent) or that `removePackageFromProject` strictly removes a package from the project and errors if not found/present?  I can likely get behind either interpretation actually...

@Turbo87 / @stefanpenner - What do y'all think?

[...]

PR #7046, opened 2017-05-16
Pass only packages to npm uninstall task that exist

raido:
This is a PR for issue #6999.

`Blueprint.removePackagesFromProject` now checks if package actually exists in project dependencies, if not then does not include it for uninstall procedure.


stefanpenner:
Is this an urgent bugfix, or should it take the normal release trains?

Turbo87:
@homu r+

given that this has proper tests I'd be okay with port this back into the `release` branch

Turbo87:
@homu retry

PR #7058, opened 2017-05-20
Use beta ember-cli-shims

locks:
This beta brings https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md support, and deprecates the current shims.


Turbo87:
@homu r+

👍

PR #7059, opened 2017-05-20
Addon#setupPreprocessorRegistry should be invoked after `addon.app` is set.

stefanpenner:
- [x] get feedback re: if this is reasonable (cc @rwjblue) 
- [x] tests (more comprehensive life-cycle tests)

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -138,8 +138,8 @@ class EmberApp {
     this.trees = this.options.trees;
 
     this.populateLegacyFiles();
-    p.setupRegistry(this);
     this._notifyAddonIncluded();
+    p.setupRegistry(this);
 
     if (!this._addonInstalled('loader.js') && !this.options._ignoreMissingLoader) {
       throw new SilentError('The loader.js addon is missing from your project, please add it to `package.json`.');
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -138,8 +138,10 @@ class EmberApp {
     this.trees = this.options.trees;
 
     this.populateLegacyFiles();
-    this._notifyAddonIncluded();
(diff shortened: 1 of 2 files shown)


stefanpenner:
The more I look at this, the more reasonable it seems. For example, it is strange to add addons to the registry, that later are not even included.

stefanpenner:
Ya, so this ordering is depended on -> https://github.com/ember-cli/ember-cli-htmlbars-inline-precompile/blob/master/index.js#L71

But I believe we can at-least make `app` available before we call setupRegistry rather then after.

Turbo87:
@stefanpenner I'm not entirely following yet. what advantage does this create?

stefanpenner:
@Turbo87 I'll write a test, that makes this more obvious but TL;DR `addon.setupPreprocessorRegistry` is invoked before `addon.app` is set, this means if you need access to `addon.app.config` for your registry setup your are out of luck today.

stefanpenner:
fixed linting error.

stefanpenner:
will change this to bugfix beta.

rwjblue:
Updated to point to `beta`, and rebased.

Turbo87:
thanks to https://github.com/simplabs/ember-test-selectors/issues/116 I finally understand what this PR is doing... LGTM 👍

rwjblue:
Merging manually (changing targets makes @homu emo).

Turbo87:
@stefanpenner it looks like the implementation here is somewhat flawed. for app builds this seems to work, but when building addons `setupPreprocessorRegistry()` is apparently called before the `EmberApp/Addon` class is even instantiated.

I'm trying to fix https://github.com/simplabs/ember-test-selectors/issues/345 but it seems almost impossible at this point. If you or @rwjblue have 30min to pair with me on this next week that would be awesome!

PR #7060, opened 2017-05-22
Allow addon.include to change mutate built-in app.imports

stefanpenner:
- [ ] see if this works
- [ ] anyone apposed?
- [ ] tests

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -139,9 +139,9 @@ class EmberApp {
 
     this.trees = this.options.trees;
 
-    this.populateLegacyFiles();
     p.setupRegistry(this);
     this._notifyAddonIncluded();
+    this.populateLegacyFiles();
 
     if (!this._addonInstalled('loader.js') && !this.options._ignoreMissingLoader) {
       throw new SilentError('The loader.js addon is missing from your project, please add it to `package.json`.');


stefanpenner:
cc @mbutki

nathanhammond:
This probably unblocks the work done here: https://github.com/ef4/ember-browserify/pull/100

I believe I only found one addon that would be impacted by this. Possibly @zackthehuman's i18n work? Regardless, only likely to be encountered by skilled maintainers who can code around it.

stefanpenner:
@nathanhammond https://github.com/ember-cli/ember-cli/pull/7059

stefanpenner:
Looks like exploring this will require additional deep dive.

rwjblue:
I believe that the logic in `populateLegacyFiles` needs to be modified, so that the `app.import`'s that are done currently are inverted (and `prepend: true` option is added).  Without changing that, this change is likely a breaking change (as evidenced by the failing tests).

@stefanpenner is working on making the ordering (specifically for ember-cli-shims) less important in https://github.com/ember-cli/ember-cli-shims/pull/118, but updating ember-cli versions generally doesn't require updating other related blueprint packages so we should take care to ensure that this PR passes using the current version of ember-cli-shims (1.2.0-beta.2)...

Turbo87:

PR #7065, opened 2017-05-26
Set the basePort for livereload from 49153 -> 7020

eriktrom:
I have observed that touchbar starts in the range 40,000-60,000. 5020 is
far outside that range, and in the range of open ports in general.

change to basePort to 7020 to fix this

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -9,7 +9,7 @@ const EOL = require('os').EOL;
 
 const Promise = RSVP.Promise;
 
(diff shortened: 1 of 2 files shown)


eriktrom:
My bad. I thought the test was for 0-65,000 - will fix

eriktrom:
y u fail? 

closes #6513

eriktrom:
now passes, rebased, g2g (unless CI surprises me after merging master)

stefanpenner:
@homu r+

PR #7066, opened 2017-05-26
Switch PhantomJS to Firefox

kellyselden:
This is the implementation of https://github.com/ember-cli/rfcs/pull/86.

~~I'd like #7064 to be resolved first.~~

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@ We love pull requests. Here's a quick guide:
 
    * node (latest LTS recommended) -- *do not install node using sudo*
    * npm (3.x)
-   * phantomjs
+   * firefox
 
 3. Run the tests. We only take pull requests with passing tests, and it's great
    to know that you have a clean slate: `npm install && npm run test-all`.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ branches:
 install:
   - ps: Install-Product node $env:nodejs_version
   - appveyor-retry npm i -g npm@^3
-  - appveyor-retry choco install phantomjs
+  - appveyor-retry choco install firefox
   - appveyor-retry yarn
   - appveyor-retry yarn add mocha-appveyor-reporter # must be installed locally.
 
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -6,6 +6,9 @@ node_js:
   - "4"
 
 sudo: false
+
+addons:
+  firefox: latest
 <% if (yarn) { %>
 cache:
   yarn: true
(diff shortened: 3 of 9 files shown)


stefanpenner:
nice!

stefanpenner:
thoughts:

* CI times seems slightly slower.
* should appveyor get similar treatment?

kellyselden:
Ready for review now.

kellyselden:
> CI times seems slightly slower.

That was brought up in the RFC https://github.com/ember-cli/rfcs/pull/86/files#r97205765. I guess the tradeoff is worth it to get a real browser?

> should appveyor get similar treatment?

How so? We don't send down any appveyor assets to apps and addons. The reasoning I've always thought is "We test Windows so you don't have to."

cc @stefanpenner

stefanpenner:
> How so? We don't send down any appveyor assets to apps and addons. The reasoning I've always thought is "We test Windows so you don't have to."

Isn't one of the goals to eventually drop phantomjs altogether? If so, weening our appveyor.yml off it may also make sense?

kellyselden:
That makes sense. We can make the switch in our CI files too.

rwwagner90:
Should we perhaps use Chrome instead? I generally add this to my travis.yml:

```yml
- export CHROME_BIN=chromium-browser
- export DISPLAY=:99.0
- sh -e /etc/init.d/xvfb start
```

kellyselden:
@rwwagner90 There were some concerns brought up with Chrome in the RFC.

rwwagner90:
@kellyselden I took a look at the RFC, was the concern you are referring to that `sudo: required` does not use container based infrastructure?

All the concerns in the text of the RFC itself, I believe, are not a problem. 

You can run it without installing it, if you just use the following:

```yml
sudo: required
dist: trusty
...
before_install:
  - export CHROME_BIN=chromium-browser
  - export DISPLAY=:99.0
  - sh -e /etc/init.d/xvfb start
```

kellyselden:
@rwwagner90 Some things I can think of: open source issues from RFC, need sudo, need trusty, using chromium instead of chrome.

rwwagner90:
@kellyselden needing those things has not hindered me at all, but just my two cents. FF is okay with me, just wanted to ensure everyone was on the same page and realized Chrome was available without installing any addons or anything.

I don't know a lot about Travis, so if sudo and trusty cause issues, just please go ahead and disregard this suggestions 😄

rwjblue:
As an aside, if you are using `--headless` you don't use `  - sh -e /etc/init.d/xvfb start`. Using the virtual frame buffer approach is much slower than phantomjs, whereas using the new `--headless` flag to chrome should be faster...

rwwagner90:
@rwjblue I don't believe I am using `--headless`. I was not aware of it. Where does that go?

Turbo87:
@kellyselden since Chrome headless is available on Travis now we should use that directly instead of using `xvfb`

Example at https://github.com/rust-lang/crates.io/pull/813

rwwagner90:
@Turbo87  I tried out the headless stuff in an ember addon, and it worked as well, if you would like an ember example https://github.com/shipshapecode/ember-flatpickr/blob/master/testem.js and https://github.com/shipshapecode/ember-flatpickr/blob/master/.travis.yml

rwjblue:
Agreed with @turbo87. We should use the new --headless stuff...

rwwagner90:
Would a PR to do Chrome instead of Firefox be accepted? I'm happy to throw one together.

Turbo87:
see #7148 😉

rwwagner90:
Ahh, beat me to it!

stefanpenner:
I believe (now that chrome has `--headless` that this may no longer be needed.

@kellyselden c/d?

kellyselden:
Although I like rooting for the underdog, no need to run xvfb if we don't have to...

jrowlingson:
@kellyselden this may be worth revisiting now that Firefox (56.0) supports [headless mode](https://developer.mozilla.org/en-US/Firefox/Headless_mode) via the [`-headless`](https://developer.mozilla.org/en-US/Firefox/Releases/56) flag on all platforms? After running the latest version (57.0) against our large-ish test suite I am seeing the performance now seems to be on par with Chrome. Travis `firefox: latest` is resolving to 55.0 currently but this is okay since this was when Linux support was added for the headless flag. I can submit a PR if there is interest.

PR #7068, opened 2017-05-27
Remove reference to "lib/ext/promise" from docs

ro0gr:
it's deprecated now.

Should we promote `rsvp` instead?


rwjblue:
@homu r+

PR #7020, opened 2017-05-03
[WIP] Refactor livereload server

kolybasov:
PR for #6913.

- [ ] Refactor LiveReloadServer to LiveReloadAddon(middleware)
- [ ] Refactor tests for LiveReloadServer
- [ ] Update docs about live-reload port and host

I don't know how to refactor tests, they are a bit complicated for me.
Also I don't know how to test local version of ember-cli on real project.

Also live-reload port and live-reload-host will be redundant so we need to inform users about it. 

Any suggestions?

PR #7069, opened 2017-05-28
Reject on interruption

ro0gr:
Signal interruption history is very interesting but complex part of cli.
In short the less `process.exit` we call the better.

In this pr I've replaced `process.exit(1)` invocations in the `Command` and `Task` models with the `Promise.reject()`s. All the interruption promises are handled with `capture-exit`.

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -17,6 +17,7 @@ const SilentError = require('silent-error');
 const execSync = require('child_process').execSync;
 const fs = require('fs');
 const existsSync = require('exists-sync');
+const RSVP = require('rsvp');
 
 let cache = {};
 
@@ -197,7 +198,7 @@ let Command = CoreObject.extend({
       return this._currentTask.onInterrupt();
     } else {
       // interrupt all external commands which don't use `runTask()` with an error
-      process.exit(1);
+      return RSVP.reject();
     }
   },
 
@@ -212,15 +213,15 @@ let Command = CoreObject.extend({
   },
 
   /**
-   * Looks up for the task and runs
-   * It also keeps the reference for the current active task
-   * Keeping reference for the current task allows to cleanup task on interruption
+   * Lookup and run the task
+   * Stores the reference to the current task
+   * It allows us to handle task interruption
    *
    * @private
    * @method runTask
    * @throws {Error} when no task found
    * @throws {Error} on attempt to run concurrent task
-   * @param {string} name Task name from the tasks registry. Should be capitalized
+   * @param {string} name `PascalCase` task names
    * @param {object} options
    * @return {Promise} Task run
    */
--- a/lib/models/task.js
+++ b/lib/models/task.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const CoreObject = require('core-object');
+const RSVP = require('rsvp');
 
 class Task extends CoreObject {
   run(/*options*/) {
@@ -8,14 +9,15 @@ class Task extends CoreObject {
   }
 
   /**
-   * Interrupt comamd with an exit code
-   * Called when the process is interrupted from outside, e.g. CTRL+C or `process.kill()`
+   * Reject all the interrupted tasks by default
+   * Called when the process is interrupted from outside(CTRL+C)
    *
    * @private
    * @method onInterrupt
+   * @returns RSVP.Promise
    */
   onInterrupt() {
-    process.exit(1);
+    return RSVP.reject();
   }
 }
 


ro0gr:
This needs some testing. I think https://github.com/ember-cli/ember-cli/pull/7019 should be merged first.

ro0gr:
Currently the most complete(but outdated) documentation about `process.exit` in `ember-cli` is written here https://github.com/ember-cli/ember-cli/issues/6779#issuecomment-280940358.

lib/models/command.js
```diff
@@ -197,7 +198,7 @@ let Command = CoreObject.extend({
       return this._currentTask.onInterrupt();
     } else {
       // interrupt all external commands which don't use `runTask()` with an error
-      process.exit(1);
+      return RSVP.reject();
```
hjdivad: We care about exiting non-zero here.  Rejecting a promise on command does not affect the exit code.  This is why the tests are red for this PR as well.

Were you planning on making other changes to ensure that the process still exits non-zero and if so what would those changes be?

ro0gr: I see. Thank you for explanation. I've tried to pass an exit code but it leads to a strange result error message in the console.

I'm going to continue work here once #6779 is resolved.

ro0gr: Sorry I've meant #7020

ro0gr:
Writing from a phone could be a mess. I mean #7019. Sorry

lib/models/command.js
hjdivad: @ro0gr when you say pass an exit code, do you mean use the exit code as the rejection value?  something like `RSVP.reject(1)`?

We will `process.exit(1)` if the cleanup promise (ie the at-exit work) rejects, but the command run promise isn't part of cleanup.

We could change the way the runner deals with command promises but of course we'll want to maintain the semantics that a SIGINT on `ember serve` (and similar) exits 0 while a SIGINT on most other things (eg `ember build`) exit non-zero.

ro0gr: >  something like RSVP.reject(1)?
> We will process.exit(1) if the cleanup promise (ie the at-exit work) rejects, but the command run promise isn't part of cleanup.

yes. If we do `RSVP.reject(1)` then `capture-exit` just [outputs 1](https://github.com/ember-cli/capture-exit/blob/master/index.js#L97) to the console and exits with `exitCode=1`. As far as I understand this [`catch`](https://github.com/ember-cli/capture-exit/blob/master/index.js#L92) is designed to handle cleanup exceptions. Seems like it's better not to change this behavior. 

Now I think it might be better not to reject on cleanup. We could `resolve` with a desired `exitCode`, something like:
```js
onInterrupt() {
    return RSVP.resolve(1);
    // or even
    return 1;
}
```
Then `capture-exit` can remember the first recognized exitCode and call [`process.exit(firstExitCode)`](https://github.com/ember-cli/capture-exit/blob/master/index.js#L90) after all the interruption handlers are settled.

@hjdivad what do you think? Do you have any objections about `resolve(desiredExitCode)`?

hjdivad: > yes. If we do RSVP.reject(1) then capture-exit just outputs 1 to the console and exits with exitCode=1. As far as I understand this catch is designed to handle cleanup exceptions. Seems like it's better not to change this behavior.

Yes although the actual rejection reason is ignored; if we `RSVP.reject(4)` the exit code is still 1.

> @hjdivad what do you think? Do you have any objections about `resolve(desiredExitCode)`?

The problem is that a) you can have multiple handlers and b) the exit trigger can itself be a call to `process.exit` that passes in its own exit code.  The current implementation of capture-exit is meant to mimic node's behaviour as much as possible.

If you `process.exit(a)` and then  `process.exit(b)` again within a `process.on('exit')` handler, the exit code will be `a`; what happens in the `on('exit')` handlers doesn't affect the exit code.

---

Having the `command.onInterrupt` promise resolve to the desired exit code sounds good to me, but I don't particularly want to push this into capture-exit, which can already be a bit tricky to reason through.  I think I'd rather see the `onInterrupt` promise not added as an `onExit` handler but instead run within `will-interrupt-process` followed by a post-interrupt call (which by default would just `process.exit` but could avoid doing this in unit tests).

I'm not sure if this is the kind of thing @rwjblue had in mind in [this comment](https://github.com/ember-cli/ember-cli/issues/6779#issuecomment-280941092) or if he wanted exit handlers to more directly control the exit code.  @rwjblue what say you?

Turbo87:

PR #7074, opened 2017-05-31
[ENHANCEMENT] Fix eslint warning on generated config/environment.js

morhook:
When opening the generated file config/environment.js with my IDE (with eslint integrated), an empty if is showing an error (the production one)

side note: on the build in ember-cli an empty if is not an error


Turbo87:
@morhook sorry, apparently we missed this PR 😞

morhook:
No problem @Turbo87 ! Thanks a lot!

PR #7077, opened 2017-06-01
Add reasonable uglify-js options.

rwjblue:
These options should always have been specified, and are good defaults.

I could have sworn that these defaults were already present, but I could not find them in this repo, in ember-cli-uglify repo, or the broccoli-uglify-sourcemap repo...

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -233,6 +233,18 @@ class EmberApp {
       },
       minifyJS: {
         enabled: this.isProduction,
+        options: {
+          compress: {
+            // this is adversely affects heuristics for IIFE eval
+            'negate_iife': false,
+            // limit sequences because of memory issues during parsing
+            sequences: 30,
+          },
+          output: {
+            // no difference in size and much easier to debug
+            semicolons: false,
+          },
+        },
       },
       outputPaths: {
         app: {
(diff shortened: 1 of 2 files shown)


krisselden:
@rwjblue should we not deep merge?

rwjblue:
@krisselden - We do merge deeply via `lodash.defaultsDeep` [here](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/lib/broccoli/ember-app.js#L284).

stefanpenner:
@rwjblue `[BUGFIX release]` ?

rwjblue:
This targets beta at the moment. I'm fine with release also, can update the PR target in the morning.

stefanpenner:
> This targets beta at the moment. I'm fine with release also, can update the PR target in the morning.

seems likely good to get this out asap IMHO.

rwjblue:
retargeted and rebased

stefanpenner:
:shipit:

cibernox:
Out of curiosity, how does this affect performance?

stefanpenner:
@cibernox what performance are you curious about build/runtime?

cibernox:
Precisely. This is tagged as performance but it's not clear to me to which one it refers. I assume it refers to some sort IIF optimization, and I'm curious to read more about that.

stefanpenner:
@cibernox TL;DR

* uglify stripped the leading `(` here  `define('asdf',[], (function() { } ))`, preventing https://www.npmjs.com/package/ember-cli-optimize from working

krisselden:
@cibernox negate_iife hurts the performance at runtime of parsing IIFE to save one character.  Turning statements into sequences, like a series of define() statements is also costly at parsing, but @rwjblue would have needed to set it to sequences: false or 0

Issue #7073, opened 2017-05-30
`will-interrupt-process` and `process` testability issues


ro0gr:
**Issues**:
- There are [some ugly things happening](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/tests/unit/utilities/will-interrupt-process-test.js#L144) with the global `process` in the `will-interrupt-process` unit tests. [Related comment](https://github.com/ember-cli/ember-cli/pull/7041#discussion_r116785222)
- Currently we can work with a real `process` only. Even in a test mode when we use [ember test helper](https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/ember.js) `will-interrupt-process`[ still works with a real `process`](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/lib/utilities/will-interrupt-process.js#L81). 

While the first issue looks like a minor(if we agreed to ignore a possibility to accidentally corrupt the global `process`) the second one is bit more tricky:
- as soon as we do [process.stdin.on](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/lib/utilities/will-interrupt-process.js#L117) on Windows TTY screen we should likely ensure that stdio is fully drained before exiting[ like the `bin/ember` does](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/bin/ember#L33). But [it's not the case with ember test helper](https://github.com/ember-cli/ember-cli/blob/5d6a29aa821f307b74e13cddc23c6fd76ef88e58/tests/helpers/ember.js#L94).  
- we should not allow `will-interrupt-process` to make a real `process.exit` or subscribe to a real `process` events when somebody uses ember test helper in [their test suite](https://github.com/emberjs/data/blob/69cdc9e6373409b5fd5e0c248a7aebd3283773aa/node-tests/blueprints/adapter-test.js#L21)

**Proposal**
The proposal is to implement a `ProcessMock` test helper. 
It will be derrived from `EventEmitter` to emulate a real process more precisely. All the streams will be inherited from `EventEmitter` as well.
All the non-`EventEmitter` methods like `setRawMode` can be implemented as a test double functions.

`ProcessMock` instance will be created in the ember test helper and passed to the cli.
[...]

PR #7086, opened 2017-06-03
Mock process

ro0gr:
Solves #7073 

This pr introduces `MockProcess` test helper. `MockProcess` and `MockProcess.stdin` both inherit `EventEmitter`. Non-`EventEmitter` methods are implemented as `testdouble.function()`s.

It allows to test `process.exit`, `SIGINT`, `CTRL+C` behaviors in a more safe and flexible way.
```js
let process = new MockProcess({
  platform: 'win',
  stdin: {
    isTTY: true,
  },
});

willInterruptProcess.capture(process);

process.stdin.emit('data', [0x03]); // CTRL+C

td.verify(process.exit());
```

**General Changes:**

- `will-interrupt-process` now gets new ~~public~~ method `.capture(EventEmitter process)`
- `cli/index` captures the process once it is invoked
- `tests/helpers/ember` now passes `MockProcess` instance to the `cli/index`

**Todo:**

 - [x] unit tests for new ~~public~~ api in `will-interrupt-process`
 - [x] collapse/rework `.free()` and `.teardown()`
 - [x] is it acceptable to [move `process` capture from the top of the `cli/index` to the invocation](https://github.com/ember-cli/ember-cli/pull/7086/files#diff-3386b9e4d97ad97b803f4678beb66c5aR70)? and why not. explanation - https://github.com/ember-cli/ember-cli/pull/7086#discussion_r121495275
 - [ ] ~~merge #7019~~ now i think if this pr lands first it can improve test setup in #7019.
 - [ ] ~~merge #7041~~ now i think if this pr lands first it can improve test setup in #7041.
 - [x] rebase
 - [ ] optional. expand build task integration test to cover builder interruption/cleanup

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+require('../utilities/will-interrupt-process').capture(process);
+
 const instrumentation = require('../utilities/instrumentation');
 
 let initInstrumentation;
@@ -12,10 +14,6 @@ if (instrumentation.instrumentationEnabled()) {
   };
 }
 
-// work around misbehaving libraries, so we can correctly cleanup before
-// actually exiting.
-require('capture-exit').captureExit();
-
 // Main entry point
 const requireAsHash = require('../utilities/require-as-hash');
 const packageConfig = require('../../package.json');
@@ -23,15 +21,6 @@ const logger = require('heimdalljs-logger')('ember-cli:cli/index');
 const merge = require('ember-cli-lodash-subset').merge;
 const path = require('path');
 const heimdall = require('heimdalljs');
-
-// ember-cli and user apps have many dependencies, many of which require
-// process.addListener('exit', ....) for cleanup, by default this limit for
-// such listeners is 10, recently users have been increasing this and not to
-// their fault, rather they are including large and more diverse sets of
-// node_modules.
-//
-// https://github.com/babel/ember-cli-babel/issues/76
-process.setMaxListeners(1000);
 
 let version = packageConfig.version;
 let name = packageConfig.name;
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -63,10 +63,10 @@ function configureLogger(env) {
   }
 }
 
-// Options: Array cliArgs, Stream inputStream, Stream outputStream
+// Options: Array cliArgs, Stream inputStream, Stream outputStream, EventEmitter process
 module.exports = function(options) {
-  // work around misbehaving libraries,
-  // so we can correctly cleanup before actually exiting.
+  // `process` should be captured before we require any libraries which
+  // may use `process.exit` work arounds for async cleanup.
   willInterruptProcess.capture(options.process || process);
 
   let UI = options.UI || require('console-ui');
@@ -142,5 +142,5 @@ module.exports = function(options) {
     settings: merge(defaultUpdateCheckerOptions, config.getAll()),
   };
 
-  return cli.run(environment).finally(() => willInterruptProcess.free());
+  return cli.run(environment).finally(() => willInterruptProcess.release());
 };
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,7 +1,6 @@
 'use strict';
 
-require('../utilities/will-interrupt-process').capture(process);
-
+const willInterruptProcess = require('../utilities/will-interrupt-process');
 const instrumentation = require('../utilities/instrumentation');
 
 let initInstrumentation;
@@ -66,6 +65,10 @@ function configureLogger(env) {
 
 // Options: Array cliArgs, Stream inputStream, Stream outputStream
 module.exports = function(options) {
+  // work around misbehaving libraries,
+  // so we can correctly cleanup before actually exiting.
+  willInterruptProcess.capture(options.process || process);
+
   let UI = options.UI || require('console-ui');
   let Yam = options.Yam || require('yam');
   const CLI = require('./cli');
@@ -139,5 +142,5 @@ module.exports = function(options) {
     settings: merge(defaultUpdateCheckerOptions, config.getAll()),
   };
 
-  return cli.run(environment);
+  return cli.run(environment).finally(() => willInterruptProcess.free());
 };
--- a/lib/utilities/will-interrupt-process.js
+++ b/lib/utilities/will-interrupt-process.js
@@ -5,11 +5,12 @@
 // When the process interrupted("CTR+C") all the registered handlers are being invoked via `capture-exit`.
 // The process finishes after the last interruption handler is settled.
 
-let captureExit = require('capture-exit');
+const captureExit = require('capture-exit');
+const handlers = [];
+const EventEmitter = require('events');
 
 let _process,
     windowsCtrlCTrap,
-    handlers = [],
     originalIsRaw;
 
 module.exports = {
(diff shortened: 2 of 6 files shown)


lib/cli/index.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+require('../utilities/will-interrupt-process').capture(process);
```
stefanpenner: lets leave a comment saying why this needs to be at the top of the file (similar to the one we have for `setMaxListeners`.

lib/utilities/will-interrupt-process.js
```diff
@@ -5,13 +5,39 @@
...
-let exit = require('capture-exit');
-exit.captureExit();
+let captureExit = require('capture-exit');
 
-let windowsCtrlCTrap;
-let handlers = [];
+let _process,
+    windowsCtrlCTrap,
+    originalIsRaw,
+    handlers = [];
```
stefanpenner: `handles` should likely be `const`

lib/utilities/will-interrupt-process.js
```diff
@@ -5,13 +5,39 @@
 // When the process interrupted("CTR+C") all the registered handlers are being invoked via `capture-exit`.
 // The process finishes after the last interruption handler is settled.
 
-let exit = require('capture-exit');
-exit.captureExit();
+let captureExit = require('capture-exit');
```
stefanpenner: `const`

lib/utilities/will-interrupt-process.js
```diff
@@ -5,13 +5,39 @@
...
+    originalIsRaw,
+    handlers = [];
 
 module.exports = {
+  capture(outerProcess) {
+    if (_process) {
+      throw new Error('process already captured');
+    }
+
+    _process = outerProcess || process;
```
stefanpenner: should we fallback here or throw if no process is provided?

ro0gr: Originally I've added this fallback for convenience in unit tests [where actual `process` doesn't matter](https://github.com/ember-cli/ember-cli/pull/7086/files#diff-b206c498c137f28dedad7f205bfe68e6R60). But agree it's better to simplify here and throw if anything strange happens.

ro0gr:
@stefanpenner thank you for review. I'll address your items soon. 

Currently I'm blocked with the smoke tests on my local machine. Some of them are failing even in the master with a strange message:
>  1) Acceptance: addon-smoke-test "before all" hook
Directory 'C:\\Users\\ro0gr\\workspace\\ember-cli\\tmp\\some_cool_addon-8kyFSwDn.tmp' already exists

While debugging I can see that [`createTestTargets` creates a temp dir](https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/acceptance.js#L54) which then is used in the [`ember addon`](https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/acceptance.js#L59) command. This `ember addon` command throws an exception mentioned above which seems reasonable because we try to create an `addon` in existing directory. At this moment I can't figure out what exactly trolls me so hard. Investigating...

While I'm investigating this I'm also making some [suspicious things](https://github.com/ember-cli/ember-cli/pull/7086/files#diff-3386b9e4d97ad97b803f4678beb66c5aR68) in this pr in order to fix smoke slow test failures on CI.

lib/cli/index.js
```diff
@@ -77,6 +65,8 @@ function configureLogger(env) {
 
 // Options: Array cliArgs, Stream inputStream, Stream outputStream
 module.exports = function(options) {
+  willInterruptProcess.capture(options.process || process);
```
ro0gr: `tests/helpers/ember` [requires ](https://github.com/ember-cli/ember-cli/pull/7086/files#diff-046dd933d1aae8e9131812190820a4c6R5) `cli/index`. If we capture the process in the top of the `cli/index` we will make it once for all tests. 

I think we need to capture the `process` before `cli` command started and [free it](https://github.com/ember-cli/ember-cli/pull/7086/files#diff-3386b9e4d97ad97b803f4678beb66c5aR143) after it's finished.

ro0gr: The only thing I'm not aware of is what can happen if we move `capture`/`caprure-exit` from the top of the file inside this function invocation.

lib/cli/index.js
```diff
@@ -77,6 +65,10 @@ function configureLogger(env) {
 
 // Options: Array cliArgs, Stream inputStream, Stream outputStream
 module.exports = function(options) {
+  // work around misbehaving libraries,
+  // so we can correctly cleanup before actually exiting.
+  willInterruptProcess.capture(options.process || process);
```
ro0gr: ember test helper requires that the `process` has an ability to be captured on each cli invocation.
If we leave capturing of the `process` outside of invocation in the top of the file like it was before we will loose an ability to configure(capture/free) the process between test runs.

tests/helpers/ember.js
```diff
@@ -65,6 +66,7 @@ module.exports = function ember(args, options) {
   }
 
   cliInstance = cli({
+    process: new MockProcess(),
     inputStream,
```
ro0gr: Seems like we could pass `inputStream` and `outputStream` from the `MockProcess` instance.
Not sure right now which benefits does it offer but it seems more consistent.

tests/helpers/mock-process.js
```diff
@@ -0,0 +1,32 @@
...
+      isRaw: false,
+      setRawMode: td.function(),
+    }, options.stdin || {});
+
+    const topLevelProps = Object.assign({
+      platform: 'MockOS',
+      exit: td.function(),
+    }, options);
+
+    Object.assign(this, topLevelProps, { stdin });
```
ro0gr: This is in incomplete form right now. For example there is no `stdout`. Currently we have only what we actually use.

stefanpenner:
re-running appveyor

lib/utilities/will-interrupt-process.js
```diff
@@ -1,17 +1,68 @@
...
+
+    // work around misbehaving libraries, so we can correctly cleanup before actually exiting.
+    captureExit.captureExit();
+  },
+
+  /**
+   * Drops all the interruption handlers and disables an ability to add new one
+   *
+   * Note: We don't call `captureExit.releaseExit() here.
+   * In some rare scenarios it can lead to the hard to debug issues.
```
ro0gr: here I'm referring to the point 9) from [this comment](https://github.com/ember-cli/ember-cli/issues/6779#issuecomment-280940358). Not sure how to better document this

lib/cli/index.js
```diff
@@ -75,8 +63,12 @@ function configureLogger(env) {
   }
 }
 
-// Options: Array cliArgs, Stream inputStream, Stream outputStream
+// Options: Array cliArgs, Stream inputStream, Stream outputStream, EventEmitter process
 module.exports = function(options) {
+  // `process` should be captured before we require any libraries which
+  // may use `process.exit` work arounds for async cleanup.
+  willInterruptProcess.capture(options.process || process);
```
ro0gr: https://github.com/ember-cli/ember-cli/pull/6437 says that the actual problem was with `ora` which used by `console-ui`. So we need to capture `process` and `process.exit` before we require `console-ui`.

ro0gr:
I believe this pr is ready for review and good to go in its current state.

ro0gr:
I've tried to improve docs but it's still not perfect. Would like any suggestion for improvement

stefanpenner:
trying appveyor again ;)

lib/utilities/will-interrupt-process.js
hjdivad: @ro0gr maybe just add a "see {link-to-comment} as an example"

ro0gr: added

ro0gr:
rebased

stefanpenner:
@ro0gr !! 🎉  !!

PR #7087, opened 2017-06-03
eslint fixes for bin/ember

ro0gr:
I know it's better not to touch `bin/ember` at all but not in the cases when it doesn't pass `eslint` 🙀 
I've tried to address code style issues here and re-enable `eslint` for `bin/ember`.

### DIFF
--- a/bin/ember
+++ b/bin/ember
@@ -5,13 +5,13 @@
 process.title = 'ember';
 
 // require('time-require');
-var resolve = require('resolve');
-var exit = require('exit');
+const resolve = require('resolve');
+const exit = require('exit');
 
 resolve('ember-cli', {
-  basedir: process.cwd()
-}, function(error, projectLocalCli) {
-  var cli;
+  basedir: process.cwd(),
+}, (error, projectLocalCli) => {
+  let cli;
   if (error) {
     // If there is an error, resolve could not find the ember-cli
     // library from a package.json. Instead, include it from a relative
@@ -28,9 +28,9 @@ resolve('ember-cli', {
     cliArgs: process.argv.slice(2),
     inputStream: process.stdin,
     outputStream: process.stdout,
-    errorStream: process.stderr
-  }).then(function(result) {
-    var exitCode = typeof result === 'object' ? result.exitCode : result;
+    errorStream: process.stderr,
+  }).then(result => {
+    let exitCode = typeof result === 'object' ? result.exitCode : result;
     exit(exitCode);
   });
 });
--- a/tests/unit/lint-test.js
+++ b/tests/unit/lint-test.js
@@ -6,7 +6,7 @@ let paths = glob('tests/*');
 
 paths = paths.concat([
   'lib',
-  'bin',
+  'bin/ember',
   'blueprints',
 ]);
 


bin/ember
```diff
@@ -5,13 +5,13 @@
 process.title = 'ember';
 
 // require('time-require');
-var resolve = require('resolve');
-var exit = require('exit');
+const resolve = require('resolve');
```
ro0gr: `const` for `require` assignments is my personal preference. 
According to missing style guide requirements and `eslint` silence I make a conclusion that it's acceptable.

ro0gr:
_{{#offtop}}_
    I think it could be nice to have `eslint`ing for docs as well.
_{{/offtop}}_

stefanpenner:
I would prefer to keep the executable as backwards compatible as possible. As when installed it will "pollute" the users system. Users should still be able to use old ember's with the old nodes they supported without issue.

ro0gr:
Then should we disable eslint for this file or update rules to make it pass?

rwjblue:
@stefanpenner - I share your worry about changing this file (pretty much at all). However, I think this part doesn't make sense to me:

> IMHO Users should still be able to use old ember's with the old nodes they supported without issue.

For this to be an issue, the user would have to have installed a newer (Node 4+ only) version of ember-cli globally under Node `<= 0.12`, and I don't think that would actually work?

stefanpenner:
@ro0ger ya we should disable

kellyselden:
> For this to be an issue, the user would have to have installed a newer (Node 4+ only) version of ember-cli globally under Node <= 0.12, and I don't think that would actually work?

@rwjblue I thought this was the exact scenario we are catering to. Why wouldn't that work?

rwjblue:
Because there is no guarantee that newer `ember-cli`'s even install on Node 0.12 (and we definitely do not support that).

ro0gr:
I don't have a strong opinion here. But @rwjblue's argument seems reasonable to me.
What's the reason to keep compatibility with node@0.12 in `bin/ember` since we dropped node@0.12 support and the rest of `ember-cli` files are already incompatible?

ro0gr:
> IMHO we should keep it as it is unless we have very strong reasons to change this file
There are no any logic changes except of dropping of node@0.12 which is now officially dropped afaik.

The only potentially problematic scenario I can imagine is the following:
- user has node@0.12 installed
- ember-cli@latest is installed globally 
- local installation of ember-cli@2.12 in some project

With this preconditions user most probably could successfully use project local `ember-cli` instance. But it will fail on any command outside of `ember-cli` project(even `ember -v`). 

Is this the only scenario or am I missing something? This scenario seems a super edge case to me which might be acceptable not to support.

At the same time my editor complains about `eslint` errors when I occasionally end up in the `bin/ember` 👁

kellyselden:
Out of curiousity, I tried running the scenario:

node 0.12
local ember-cli 2.8.0
global ember-cli latest

```
C:\Users\kelly\Desktop\my-app>ember v
ember-cli: 2.8.0
node: 0.12.18
os: win32 x64

C:\Users\kelly\Desktop\my-app>cd ..

C:\Users\kelly\Desktop>ember v
C:\Users\kelly\AppData\Roaming\npm\node_modules\ember-cli\lib\cli\index.js:3
const instrumentation = require('../utilities/instrumentation');
^^^^^
SyntaxError: Use of const in strict mode.
```

And then I tried installing ember-cli latest globally with node 0.12 (npm 2): https://pastebin.com/Rb37fDEA

Everything is currently working as expected, don't know how long before it breaks though.

stefanpenner:
meeting outcome:

* lets leave this for now
* add eslint ignores if this causes grief.

kellyselden:
Also a comment as to why we keep this file so compatible would be nice for future travelers.

ro0gr:
Thanks for considering this change

PR #7088, opened 2017-06-04
Add custom transform when using `app.import`

kratiahuja:
# Motivation

Often we require addons to transform the contents of the library being imported using `app.import`. Addon authors need to have broccoli knowledge and end up writing boiler plate code. If the transformation needs to be done in multiple addons, it often causes grievances for addon authors that don't have broccoli knowledge. A classic example is the fastboot migration where addon authors need to wrap the content of their browser libraries with a fastboot check since the `process.env.EMBER_CLI_FASTBOOT` isn't available. 

`ember-cli` already provides to run AMD shim transform. This PR extends it for addon authors to provide their own transform.

# How it works:

A vendor library can be amd shim'd using the below API:

```js
app.import(<path>, { using: [{transformation: 'amd', as: 'hello-world'}]});
```

We extend the above without any API change to be:

```js
app.import(<path> {
   using: [
      {
         transformation: function(tree) {
              // do your transform
              return tree;
        }
     }, ...]
});
```

This PR extends the `transformation` to take in a custom function or `amd` and invoked the function if it is not `amd`. It also makes sure to run the transformation in the order they are provided in `using` array.

# TODO

- [ ]  Update guides [here](https://ember-cli.com/user-guide/#standard-anonymous-amd-asset) to talk about the extension.

cc: @stefanpenner @rwjblue

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -128,7 +128,7 @@ class EmberApp {
 
 
     this._scriptOutputFiles = {};
-    this.amdModuleNames = null;
+    this.entries = null;
 
     this.legacyFilesToAppend = [];
     this.vendorStaticStyles = [];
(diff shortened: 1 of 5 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1024,13 +1024,26 @@ class EmberApp {
...
-          files: Object.keys(this.amdModuleNames),
-          annotation: 'Funnel (named AMD)',
-        });
-        externalTree = this._mergeTrees([externalTree, shimAmd(anonymousAmd, this.amdModuleNames)], {
-          annotation: 'TreeMerger (named AMD)',
+      if (this.entries) {
+        let transformedTrees = [];
+        for (let filePath in this.entries) {
+          let entry = this.entries[filePath];
+          let treeToBeTransformed = new Funnel(externalTree, {
```
stefanpenner: my concern with this approach, is we will now be creating a funnel per input. I worry this will result in later performance issues.

The existing approach for the `amd` shim, is to do 1 funnel for all `amd` transforms. I suspect we should continue to follow this pattern.

kratiahuja: I agree with you. The primary reason I went with this approach is to respect the ordering. Example:

An addon imported a file and ran a bunch of transforms: a, b, and c. I thought we should make sure to run those in order. 

If ordering is not of importance and we run the AMD transform always last, then I can group the funnel operation to be per transform (which would do 1 funnel for all amd transforms) rather than per file.

@stefanpenner thoughts?

Turbo87:
superseded by https://github.com/ember-cli/ember-cli/pull/7245

PR #7094, opened 2017-06-05
concat instead of `unshift each` in `_processedExternalTree`

bekzod:


bekzod:
I think just needs test restart

kellyselden:
restarted

stefanpenner:
@homu r+

PR #7097, opened 2017-06-05
avoid extra `.map` in `addBowerPackagesToProject`

bekzod:


lib/models/blueprint.js
```diff
@@ -1031,8 +1031,8 @@ let Blueprint = CoreObject.extend({
     let packageNamesAndVersions = packages.map(pkg => {
```
kellyselden: This map seems out of place to me. Perhaps it should be a forEach that pushes to both `packageNames` and `packageNamesAndVersions` arrays.

lib/models/blueprint.js
```diff
@@ -1027,12 +1027,14 @@ let Blueprint = CoreObject.extend({
...
+    let packageNames = new Array(packages.length);
+    let packageNamesAndVersions = new Array(packages.length);
+    packages.forEach((pkg, i) => {
       pkg.source = pkg.source || pkg.name;
-      packageNames.push(pkg.name);
-      return pkg;
-    }).map(bowEpParser.compose);
+      packageNames[i] = pkg.name;
+      packageNamesAndVersions[i] = bowEpParser.compose(pkg);
+    });
```
Turbo87: @bekzod I'm not sure what the goal of this PR is. I'd personally prefer it like this:

```js
packages.forEach(pkg => (pkg.source = pkg.source || pkg.name));

let packageNames = packages.map(pkg => pkg.name);
let packageNamesAndVersions = packages.map(bowEpParser.compose);
```

It might not be as fast, but is a lot clearer IMHO

bekzod: goal was just cleanup, it was using `[].push` inside map (which looks weird) also just tried avoid extra iterations,  ^ what you suggested and what I did looks equal in clarity IMHO, should redo it like you suggested still ?

Turbo87: yeah, would be nice. I don't like the extra `i` parameter most of the time 😉

PR #7098, opened 2017-06-05
correct code formatting

bekzod:
makes it clearer confusing formatting

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -407,13 +407,15 @@ let Blueprint = CoreObject.extend({
   */
   _process(options, beforeHook, process, afterHook) {
     let intoDir = options.target;
-
     return this._locals(options)
-      .then(locals => Promise.resolve()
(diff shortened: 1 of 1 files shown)


stefanpenner:
```
    413:23  error  Unexpected block statement surrounding arrow body  arrow-body-style
    418:47  error  Missing semicolon      
```

Please make sure to run lint tests before :)

Turbo87:
I don't think this change makes the code any more readable then what we had before. What we would like to use are async functions but due to our Node support policy we unfortunately can't use those yet...

bekzod:
ok, but I haven't seen such promise formatting before :P , free to close

stefanpenner:
> I don't think this change makes the code any more readable then what we had before. What we would like to use are async functions but due to our Node support policy we unfortunately can't use those yet...

we could use `co`

Turbo87:
> we could use co

unfortunately we can't use co-wrapped functions as ES6 class methods 😞

PR #7104, opened 2017-06-06
cleanup `addon/dependencies`

bekzod:


Turbo87:
@bekzod what issue is this fixing? why is this change needed?

bekzod:
@Turbo87 just looked like it needed cleanup, just avoid calling `assign` when pkg is empty

PR #7105, opened 2017-06-06
use `const` where appropriate

bekzod:


Turbo87:
thanks @bekzod 

@homu r+

PR #7109, opened 2017-06-06
Fix `ember install` for scoped packages

ef4:
It's not correct to try to strip the scope out of package names. I don't know why the test mock was set up this way, but you can verify this change against a real scoped package by doing:

 * `ember new my-app`
 * `cd my-app`
 * `ember install @cardstack/git`

Without this change it fails, after this change it succeeds.

Stripping off scopes fundamentally can't be correct because it's legal for `@foo/x` and `@bar/x` to want to coexist in a single app. NPM consistently treats the scope as part of the name.

### DIFF
--- a/tests/unit/utilities/get-package-base-name-test.js
+++ b/tests/unit/utilities/get-package-base-name-test.js
@@ -8,8 +8,8 @@ describe('getPackageBaseName', function() {
     expect(getPackageBaseName('my-addon')).to.equal('my-addon');
   });
 
-  it('should return the package name without its scope', function() {
-    expect(getPackageBaseName('@scope/my-addon')).to.equal('my-addon');
+  it('should return the full name when scoped', function() {
+    expect(getPackageBaseName('@scope/my-addon')).to.equal('@scope/my-addon');
   });
 
   it('should strip away version numbers', function() {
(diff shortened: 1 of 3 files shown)


kellyselden:
cc @dfreeman, not insinuating you broke anything, just know you were dealing with scoped packages recently.

tests/unit/utilities/get-package-base-name-test.js
```diff
@@ -8,8 +8,8 @@ describe('getPackageBaseName', function() {
     expect(getPackageBaseName('my-addon')).to.equal('my-addon');
   });
 
-  it('should return the package name without its scope', function() {
-    expect(getPackageBaseName('@scope/my-addon')).to.equal('my-addon');
+  it('should return the full name when scoped', function() {
+    expect(getPackageBaseName('@scope/my-addon')).to.equal('@scope/my-addon');
```
stefanpenner: ya, this seems good to me.

stefanpenner:
bugfix release or bugfix beta?

raido:
Yay! I was just debugging why my addon does not generate correct import paths in files. Hopefully 2.15 ships soon.

PR #7110, opened 2017-06-07
convert to RSVP promise inside `utilities/execa`

bekzod:


stefanpenner:
Looks like a lunging error

stefanpenner:
@homu r+

kellyselden:
@bekzod Thank you for all the good refactors lately!

PR #7113, opened 2017-06-08
simplified promise chain in `git-init`

bekzod:


lib/tasks/git-init.js
```diff
@@ -19,20 +20,15 @@ module.exports = class GitInitTask extends Task {
...
-      return this._gitInit()
-        .then(() => this._gitAdd())
-        .then(() => this._gitCommit())
-        .then(() => ui.writeLine(chalk.green('Successfully initialized git.')));
-    })
-      .catch(error => {
-        if (gitVersionWorked) {
-          throw error;
+    return this._gitVersion()
+      .then(() => true, () => false)
```
kellyselden: It looks a little stranger now, but I like that you don't have to re-throw anymore.

bekzod: I wanted to move `.then(() => true, () => false)` to `this._gitVersion ` and rename `_gitVersion` => `_hasGit`, but then had to update test :P

PR #7117, opened 2017-06-09
[WIP] Add fastboot-disable import transformation

vlascik:
This adds a new transformation to `app.import` statement, similar to `transformation: 'amd'`, enabling you to avoid the boilerplate associated with wrapping libraries that break fastboot in `if (typeof FastBoot === 'undefined')`
Usage: 
```js
app.import('bower_components/datatables/media/js/jquery.dataTables.js', { using: [{ transformation: 'fastboot-disable' }] });
```
Just an initial riff to open a discussion.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -29,6 +29,7 @@ const ConfigReplace = require('broccoli-config-replace');
 const ConfigLoader = require('broccoli-config-loader');
 const mergeTrees = require('./merge-trees');
 const shimAmd = require('./amd-shim');
+const fastbootDisable = require('./fastboot-disable');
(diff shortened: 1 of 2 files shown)


vlascik:
Hm after submitting this I noticed a similar PR, https://github.com/ember-cli/ember-cli/pull/7088 . Not sure if I should close this or not.

kratiahuja:
I don't think we should be leaking fastboot semantics into `ember-cli`. We already discussed this in `ember-cli` meeting and we have a way to go forward. I will be putting together an RFC this weekend. 

My suggestion is close this since there is another PR that supersedes this.

vlascik:
Ok, I'll close this.

PR #7120, opened 2017-06-09
avoid calling extra `.filter` in `addon/_fileSystemInfo`

bekzod:


lib/models/addon.js
```diff
@@ -932,10 +932,8 @@ let addonProto = {
       files = files.concat(this._getAddonTemplatesTreeFiles());
     }
 
-    files = files.filter(Boolean);
-
     let extensionMatcher = new RegExp(`(${templateExtensions.join('|')})$`);
-    let hasTemplates = files.some(file => extensionMatcher.test(file));
+    let hasTemplates = files.some(file => !!file && extensionMatcher.test(file));
```
Turbo87: @bekzod can we use `Boolean(file)` instead of the double negation?

bekzod: actually we can remove it entirely :P I think

PR #7136, opened 2017-06-21
[BUGFIX] ember test work with both —server and —path

stefanpenner:
This allows `ember test --server` to work with already built assets.

```js
ember test --server --path ./path-to-dist
```

tasks:

- [x] make work
- [x] tests
- [x] confirm
- [x] decide which channel -> beta (as this has mega low risk, but is quite handy)


rwjblue:
Without watching what is the point of using `ember test --server --path ./somewhere-else`?  The reason we `throw` today was because it is not at all obvious that changes / updates to `--path` will never be reflected within the `ember test --server --path ./whatever` invocation.

stefanpenner:
> Without watching what is the point of using ember test --server --path ./somewhere-else

being able to connect a `test serve` to an existing build is quite handy, especially when wanting to quickly test something in an interactive way (but still via testem infrastructure) but without having to do a full build.

It is true that we will lose automatic reloads, but the developer can manually reload <cmd-r> etc.

Now that being said, I can add reloading to this as well...

hjdivad:
@rwjblue one reason is to kick off a set of tests without waiting for a new build (or doing another build) which can matter for slow/large mem builds

rwjblue:
I'm not opposed, but we either need to print a message on startup saying that changes to the files in `--path` will not be reflected or just make it work...

stefanpenner:
failing test (expected):

```
not ok 565 test command --server option throws an error if using a build path
  AssertionError: expected promise to be rejected but it was fulfilled with undefined
```

rwjblue:
Just discussed a bit with Stef, and any logging we do will be immediately replaced by the `testem` ncurses display. I am 👍 on this...

PR #7137, opened 2017-06-21
add in-app testing page reference

kellyselden:
Feel free to veto, but it seemed to me that a reference to this section should be called out in the readme.


kellyselden:
Rebased and green. Any opposition?

Turbo87:
@rwjblue opinions?

rwjblue:
In general, I am opposed to testing in any way other than the `ember test` command.  I do believe that Stef and I disagree here, so it would be good to hash it out on the weekly call.

PR #7142, opened 2017-06-25
[BUGFIX release] node 8

stefanpenner:


kellyselden:
This appears to no longer be an issue

PR #7152, opened 2017-06-26
Remove redundant chrome installation since appveyor has latest chrome

sivakumar-kailasam:
AppVeyor seems to be flaky at installing chrome via choco.  But [appveyor does bundle the latest chrome ](https://www.appveyor.com/docs/build-environment/#web-browsers) in the default image, so removing the redundant install step.

//cc @twokul


twokul:
very nice!

PR #7161, opened 2017-06-29
Use headless chrome in addon build config

### LINKED ISSUE #7155: Missing Chrome runner for addon blueprint.
The app blueprint was updated to use `chrome --headless`, but we missed the addon blueprint...

sivakumar-kailasam:
//cc @rwjblue.  Fixes #7155

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -6,6 +6,11 @@ node_js:
   - "4"
 
 sudo: false
+dist: trusty
+
+addons:
+  chrome: stable
+
 <% if (yarn) { %>
 cache:
   yarn: true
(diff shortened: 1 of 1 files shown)


blueprints/addon/files/.travis.yml
```diff
@@ -3,9 +3,14 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "4"
+  - "6"
```
rwjblue: We had intentionally left this at 4 so that folks would be testing against their lowest supported node. If we do update this, we need to change/remove the comment directly above.

sivakumar-kailasam: @rwjblue  So which one of the two would we like to do?

rwjblue: For now, lets stick with 4

rwjblue:
@homu r+

rwjblue:
😭  @homu

PR #7162, opened 2017-06-29
Headless chrome only in CI

sivakumar-kailasam:
Testem recently added the ability to support browser args specific to a testem mode. So we should've probably limited headless chrome only to CI mode. This PR addresses that.
//cc @Turbo87 & @rwjblue

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -9,11 +9,14 @@ module.exports = {
     'Chrome'
   ],
   'browser_args': {
-      'Chrome': [
+    'Chrome': {
+      'mode': 'ci',
+      'args': [
         '--disable-gpu',
         '--headless',
         '--remote-debugging-port=9222',
         '--window-size=1440,900'
-      ],
+      ]
+    }
   }
 };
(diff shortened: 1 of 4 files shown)


Turbo87:
I'm not a fan of this since in dev mode it will now be impossible to run headless Chrome. Without this change you can just run `ember test --serve --no-launch` (I aliased that to `ets`) instead and click on the link at the top to run in non-headless mode. If you're using `ember-exam` it will also open up several Chrome windows with this change, which can be a bit annoying and doesn't actually help with debugging at all.

tl;dr I vote NO

sivakumar-kailasam:
In case others feel the same, I'll close this one in favor of a PR just updating testem.

stefanpenner:
>  PR just updating testem.

In general upstreaming things to simplify stuff is a good idea.

sivakumar-kailasam:
@stefanpenner didn't quite get that.

Turbo87:
@sivakumar-kailasam I think he wanted to say that a PR that bumps testem will be valueable regardless of the status of this PR

sivakumar-kailasam:
Got it. Raised #7167 for that.

sivakumar-kailasam:
Rewritten commits on my branch to include only the testem config changes.

Issue #7163, opened 2017-06-29
Generates --classic type structure when both usePods is true and --pod is used 


emrekutlu:
```javascript
// project-root/.ember-cli
{
  "usePods": true
}
```
When <code>usePods</code> is <code>true</code>;

```
$ ember generate controller foo --pod

installing controller
  create app/controllers/foo.js
installing controller-test
  create tests/unit/controllers/foo-test.js
```
Generates in <code>--classic</code> structure not in <code>--pod</code>.

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 2.13.3
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
3.10.8
0.24.6
```

emrekutlu:
I submitted a PR https://github.com/ember-cli/ember-cli/pull/7164.

But I believe using mutually exclusive values in different arguments is problematic. It is not obvious what will happen if I use both <code>--classic</code> and <code>--pod</code> together. Better be something like <code>structure=pod</code> but I suppose you don't use this kind of command arguments.

webhive:
Confirm it. Got this issue too.

PR #7164, opened 2017-06-29
Fix generate command when both usePods option and --pod argument is used

emrekutlu:
Resolves https://github.com/ember-cli/ember-cli/issues/7163

### DIFF
--- a/lib/commands/generate.js
+++ b/lib/commands/generate.js
@@ -43,8 +43,16 @@ module.exports = Command.extend({
       args: rawArgs,
     };
 
-    if (this.settings && this.settings.usePods && !commandOptions.classic) {
-      commandOptions.pod = !commandOptions.pod;
+    if (this.settings && this.settings.usePods) {
+      if (commandOptions.pod) {
+        let warning = 'Using both .ember-cli usePods settings and --pod flag ';
+        warning += 'together has been deprecated.';
+        this.ui.writeDeprecateLine(warning);
+      }
+
+      if (!commandOptions.classic) {
+        commandOptions.pod = !commandOptions.pod;
+      }
     }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
I agree that this is confusing, but the intent here was fhT --pod flag negates the value of usePods. So if you had usePods as true you would use --pod to get non-pods.

emrekutlu:
Hmm, so what does <code>--classic</code> do?

rwjblue:
@Turbo87 - This patch has nothing to do with using both `--pod` and `--classic` at the same time, what am I missing?

hjdivad:
> @Turbo87 - This patch has nothing to do with using both --pod and --classic at the same time, what am I missing?

I think the point is that the existence of `--classic` implies the semantics of `--pod` is `--use-pods` and not `--toggle-pods` and not (only) that they're confusing when used together

rwjblue:
I totally agree with you there, but this change would be a breaking change for anyone leveraging the existing behavior. 

I'd be totally fine with adding a deprecation / warning for anyone that has both `usePods` and specifies either `--pods` or `--classic` specified, but not with arbitrarily changing the meaning of the flags (at least not within the same release, but it could be feathered out)...

Turbo87:
@emrekutlu can you prepare a PR that prints a warning to the console when `--pod` is used in combination with `usePods`?

rwjblue:
@emrekutlu - Ping?

emrekutlu:
@rwjblue sorry for the wait, I am going to update the pull request this week hopefully. Thanks for pinging.

emrekutlu:
@rwjblue I updated the pull request.

rwjblue:
Thank you!

PR #7169, opened 2017-06-29
Travis multiple blank line cleanup and if block code consolidation

kellyselden:
I had Travis file cleanup in #7066 that was closed. This takes the good cleanup without the firefox stuff.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -10,7 +10,6 @@ dist: trusty
 
 addons:
   chrome: stable
-
 <% if (yarn) { %>
 cache:
   yarn: true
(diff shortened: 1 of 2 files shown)


Turbo87:
hmm... are those files not checked by any tests?

stefanpenner:
Likely not a bad idea to load them via yml parser and ensure the internal state is as expected

Turbo87:
> Likely not a bad idea to load them via yml parser and ensure the internal state is as expected

sounds nice in theory, but might be overkill for what we're doing here 🤔

stefanpenner:
@Turbo87 future PR is fine

kellyselden:
I though about fixtures to test these files too. I think I will add them. It's so easy to regress the formatting in these files that I think it will be worth it.

PR #7172, opened 2017-06-30
bump lodash

bekzod:
this version drops `assign`, `some`, `any`, `keys` in favor of natives


bekzod:
forgot that there is greenkeeper

PR #7177, opened 2017-06-30
verify npm/yarn logic in travis files

kellyselden:
From https://github.com/ember-cli/ember-cli/pull/7169#issuecomment-312379050

Already found one bug in the files!

If this seems like too much overhead, please note I plan on building upon these tests with README checks in the next PR.

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -363,4 +363,76 @@ describe('Acceptance: ember new', function() {
     expect(file('app/templates/application.hbs'))
       .to.contain("Welcome to Ember");
   }));
+
+  describe('verify fictures', function() {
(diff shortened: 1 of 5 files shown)


tests/acceptance/new-test.js
```diff
@@ -363,4 +363,76 @@ describe('Acceptance: ember new', function() {
...
+        'new',
+        'foo',
+        '--skip-npm',
+        '--skip-bower',
+        '--skip-git',
+      ]);
+
+      [
+        '.travis.yml',
+      ].forEach(filePath => {
```
Turbo87: do we need this loop?

kellyselden: Not in this PR, but the next one that has README checks will make use of the loop. Future planning.

PR #7180, opened 2017-07-04
install npm 4 in addon travis using npm

kellyselden:
I was using node 4 in some addons, and bundled npm 2 makes it unbearably slow. npm 4 makes it a lot better. npm 5 was not stable enough in my testing.


stefanpenner:
@kellyselden this looks like a good change. Did you happen to try npm 5.0.4?

kellyselden:
@stefanpenner yep. https://travis-ci.org/ember-vr/ember-aframe-camera-extras/jobs/249847007

It looks like it skipped "dependencies" and only did "devDependencies".

PR #7191, opened 2017-07-05
blueprints/app: Update "ember-source" and "ember-data" to v2.14.0

Turbo87:


blueprints/app/files/package.json
```diff
@@ -29,11 +29,11 @@
     "ember-cli-shims": "^1.1.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^1.2.0",
-    "ember-data": "^2.13.2",
+    "ember-data": "~2.14.0",
```
locks: I think this should be `2.14.3` just for safesies? The previous 2.14 releases have some problems.

rwjblue: @locks - There is only 2.14.0 (you were thinking of 2.13).

Turbo87: no, he is right. updated to `ember-data@~2.14.3`.

rwjblue: WHOOPS! I was thinking this was for `ember-source` 🤦‍♂️

PR #7192, opened 2017-07-05
tests/acceptance: Delete broken "ember generate http-proxy" test

Turbo87:
as discussed on Slack since this is failing us for no good reason on AppVeyor


Turbo87:
@homu r+

maybe...

PR #7195, opened 2017-07-05
fix  --non-interactive test regression

kellyselden:
From #7130

Is homu still alive? It would have prevented this.

cc @Turbo87


rwjblue:
Homu has not been responding to us in the last couple weeks.

Turbo87:
actually home wouldn't have prevented it since it was caused by the upward merge of the stable release 😞

kellyselden:
@Turbo87 Ok my bad.

PR #7196, opened 2017-07-05
fix a bad merge conflict resolution

kellyselden:
I think that's what happened.


Turbo87:
Yep, sorry about that 😞

PR #7197, opened 2017-07-06
use newer yarn on travis

kellyselden:
https://travis-ci.org/ember-cli/ember-cli/jobs/250484161#L1848-L1851

```
not ok 91 Acceptance: ember new ember new uses yarn when blueprint has yarn.lock
  Error: Command failed: yarn install --ignore-optional --non-interactive
  
    error: unknown option `--non-interactive'
```


rwjblue:
@kellyselden since #7130 targeted release branch, I think this will need to as well...

kellyselden:
@rwjblue not needed. as it turns out, only the `dist: trusty` builds have old yarn. `release` branch isn't using that build yet.

kellyselden:
In other words, this change should have been included with the chrome headless stuff. Since release doesn't use chrome yet, this isn't needed.

PR #7198, opened 2017-07-06
Fix a test in blueprint addon

villander:


kellyselden:
This was fixed in #7196. Not sure why github doesn't say conflict. I think github's ui is broken in that regard.

villander:
@kellyselden Very strange, I found this in the master branch of ember-cli, and I say more have other tests failing.

PR #7202, opened 2017-07-06
Space.

johnunclesam:
Added space.


kellyselden:
Our anonymous function style is no space, as seen here https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/environment.js#L4. Thanks for the contribution though. You can ignore this file locally in your project if it is failing your linter.

johnunclesam:
@kellyselden ok, sorry I read now.

PR #7203, opened 2017-07-06
Space.

johnunclesam:
Added space.


kellyselden:
Same as #7202

kellyselden:
But the same global linting style applies here.

johnunclesam:
@kellyselden sorry i read it now.

PR #7206, opened 2017-07-06
[bugfix release] remove MODEL_FACTORY_INJECTIONS

kellyselden:
This is still giving deprecations in new 2.14 apps. We should promote this to release hotfix.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -3,11 +3,7 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
-let App;
-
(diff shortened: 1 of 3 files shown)


rtablada:
@kellyselden good idea on releasing a hotfix release.

From an optics and training standpoint it is confusing to new developers to start new applications with a deprecation warning.

rtablada:
@kellyselden or @Turbo87 any update on cutting a new version now that this blueprint update has been made?

kellyselden:
No blockers that I know of, @ember-cli/core ?

Turbo87:
would like to solve a few other issues too, but we can always do another release :)

PR #7207, opened 2017-07-07
fix: Add some defaults for index.html

knownasilya:
Basically, these are defaults that can be updated or left as is.
The use case for these is to improve accessibility. These metrics were brought up by the Lighthouse test.

Feel free to close, if these don't make sense. I know that `noscript` doesn't make sense if using fastboot, but I think most people aren't using it.

### DIFF
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -15,10 +15,6 @@
     {{content-for "head-footer"}}
   </head>
   <body>
-    <noscript>
-      JavaScript must be enabled to run this application.
-    </noscript>
-    
     {{content-for "body"}}
 
     <script src="{{rootURL}}assets/vendor.js"></script>
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
@@ -15,6 +15,10 @@
     {{content-for "head-footer"}}
   </head>
   <body>
+    <noscript>
+      JavaScript must be enabled to run this application.
+    </noscript>
+    
     {{content-for "body"}}
 
     <script src="{{rootURL}}assets/vendor.js"></script>


blueprints/app/files/app/index.html
```diff
@@ -15,6 +15,10 @@
     {{content-for "head-footer"}}
   </head>
   <body>
+    <noscript>
+      JavaScript must be enabled to run this application.
```
kellyselden: Do you know how this will affect fastboot apps that want to be able to run without js?

blueprints/app/files/app/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
kellyselden: Is english the standard default for this kind of thing? We won't hurt any feelings?

knownasilya: I think it's the most common, and having it there will prompt users to set it for themselves if it's different for their app.

blueprints/app/files/app/index.html
knownasilya: I am unsure about this. It might just render on the screen, but everything else should work. If there was a way to add a check if fastboot is enabled, I could add that.

kellyselden: cc @kratiahuja

kratiahuja: in general, I strongly disagree that we should add this no script block. It would imply we have no server side rendering solution available when there is available. Moreover, we plan to add the fastboot addon in the default blueprint when creating apps. So this message is surely going to be confusing.

Also there is no way to check if fastboot is used since it will be a runtime thing rather than static. Fastboot could remove the noscript block before sending the index.html but it seems wrong to me.

knownasilya: I think instead of not adding this, we need to have a way to know that Fastboot is running and remove it if it is. Because, Fastboot takes extra work, and even if it's added by default, some people would remove it, because it's extra work or doesn't fit into the budget of the app.

kratiahuja: I am not sure I understand what extra work FastBoot takes? It is supposed to help in your initial page renders rather than the other way round. If people would remove it, I would like to first understand why they would do so?

knownasilya: Because their code has to be written a certain way that guards for fastboot usage, which requires extra work, as well as the addons they use have to be fastboot compatible.

kratiahuja: Well I am talking about new apps that are created will have fastboot addon added by default. Regarding addons, most addons should be compatible and if they are not it should be straightforward to make them compatible. 

In general, I am -1 to adding this `noscript` by default in the blueprint for the primary purpose that it is going to be confusing for the community.

knownasilya: I must be misunderstanding how Fastboot works. Does it not require putting checks before any DOM operations in hooks like `didInsertElement`? This can't be avoided in new or old apps, so I'm not seeing Fastboot as being "free", thus I was saying that it requires work on the part of the developer to keep the app Fastboot "compliant".

> it is going to be confusing for the community.

Sounds like an education problem. But at the same time, if it was wrapped in something like:

```hbs
{{#unless isFastboot}}
  <noscript>...</noscript>
{{/unless}}
```

The confusion is minimized since it's obvious what happens in the Fastboot scenario.

---

I'm fine to remove this if it really is a bad idea. Just want to make sure we are
talking about the same things first.

kratiahuja: > Does it not require putting checks before any DOM operations in hooks like didInsertElement?

Hooks like `didInsertElement`, `willInsertElement` which are meant for DOM manipulation don't run in FastBoot. Therefore you don't need checks in those hooks. See: http://ember-fastboot.com/docs/user-guide#designing-components

> But at the same time, if it was wrapped in something like:

That's exactly what I am saying: it is not _easily_ possible since FastBoot is not a static check. It is a run-time check. You could have your app running in FastBoot and you will still end up with this `noscript` check because there is no way to have `{{unless isFastBoot}}` check. The only way to avoid the `noscript` check when running in FastBoot is:

1. Have a middleware that runs after fastboot-express-middleware and strips out the `noscript` using some regex
2. Have `fastboot` itself remove it before appending the serialized template in the index.html.

Both of the above means we are magically removing the `noscript` tag which is going to cause confusion IMO.

knownasilya: I see now. Thanks for your patience with me 😄  I'll remove this change.

Turbo87:
@knownasilya the cli team decided that having English as the default here can be misleading in some cases. We would rather prefer it if ember-intl for example would set that attribute based on the chosen locale.

knownasilya:
No problem 👍 just wanted to get the discussion going.

Turbo87:
see also https://github.com/ember-intl/ember-intl/issues/533

PR #7208, opened 2017-07-07
bugfix(legacy-addons): Fixes typo in babel transpilation options

pzuraq:
Fixes a small typo that causes the AMD interop to be added to legacy addons which don't specify their own transpiler. This probably only affects a very small set of users, we ran into it because we have a couple really old addons still and we're using the pre-npm ember-cli-shims which don't specify `__esModule`.

Unsure if this needs tests/how to test it, if it does I'm happy to add them with some guidance.


rwjblue:
👍

PR #7211, opened 2017-07-08
[bugfix release] Ember try remove node modules

kellyselden:
Travis is able to resolve the `ember` bin in node_modules no problem. Is this a legacy thing that I'm not aware of, or can it be safely removed?

cc @kategengler


rwjblue:
Somewhat recently travis removed node_modules/.bin from $PATH (and we had to add this prefix). They subsequently reverted that change because many things broke. 

IMHO, using node_modules/.bin is still the more correct way of doing this (without requiring $PATH shenanigans or installing ember-cli globally) so we should probably stick with that...

PR #7227, opened 2017-07-14
[BUGFIX beta] Don't merge `emberCLITree` twice

### LINKED ISSUE #7222: Including `emberCLITree` twice?
It seems like we include `emberCLITree` twice, [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1206) and [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1344).

I haven't run the tests but removing [this merge](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1342-L1348) seemed to "just work" (tested with the app locally).

needs a proper bugfix though, created to start tracking.

twokul:
Fixes #7222

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1339,14 +1339,6 @@ class EmberApp {
       appJs = processModulesOnly(appJs, 'Babel: Modules for app/');
     }
 
-    appJs = this._mergeTrees([
-      appJs,
-      this._processedEmberCLITree(),
-    ], {
-      annotation: 'TreeMerger (appJS  & processedEmberCLITree)',
-      overwrite: true,
-    });
-
     appJs = this._concatFiles(appJs, {
       inputFiles: [`${this.name}/**/*.js`],
       headerFiles: [


twokul:
`node@4` only fails on Appveyor; failure seems unrelated to the code change.

rwjblue:
Can you link where `appJs` gets the processed ember-cli tree the other time?

twokul:
we push `emberCLITree` to `sourceTrees` inside `appAndDependecies()` [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1206). And then push `emberCLITree` again inside `javascript()` call [here](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1344).

@rwjblue does that answer your question?

stefanpenner:
@twokul thanks for digging into this, and providing the links. I reviewed, and I agree this is redundant step (and likely a pretty costly one).

Here is my analysis, let me know if this is also what you saw:

1. `appAndDependencies` includes `emberCLITree` https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1206-L1218
2. `appAndDependencies` is only invoked from `javascript()`: https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1333
3. the `CLITree` is then merged with `appJs` https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1344
4. this merge is redundant since, the CLITree is already included (last tree to be merged) in the `appjs` tree

We are essentially doing:

```js
merge([merge([..., cliTree]), cliTree]);
```

Which is equivalent to:

```js
merge([..., cliTree]);
```

And interestingly, I suspect this may actually be buggy for users who don't have `ember-cli-babel` installed (super rare, and for old setups.. but possible). As they will clobber a module compiled cliTree with one that is not... https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1338

stefanpenner:
I feel good about this one (would like @rwjblue to also sign-off). We should now decide if this should ride the canary train, or land as a `[BUGFIX beta]`. I would vote `[BUGFIX beta]` if @twokul can try it out in some non-trivial apps, and make sure everything seems ok.

twokul:
@stefanpenner your analysis is correct, this is exactly what's happening :P thanks much for looking at it!

I tried it with one of the internal LinkedIn apps and it did work fine. I can tag this commit as `[BUGFIX beta]` and we could ship unless good Sir @rwjblue has concerns/other suggestions.

stefanpenner:
Chatted with @rwjblue, he's good with this.

stefanpenner:
Ah this also needs to be submitted to beta branch, @twokul mind doing that?

twokul:
@stefanpenner absolutely

PR #7229, opened 2017-07-14
[BUGFIX release] fix shutdown/interrupt code (rebased)

ro0gr:
rebased #7019 on top of master with #7086 included. 

There is one notable change in this PR: I've removed [`teardown()` from `cli/cli`](https://github.com/ember-cli/ember-cli/pull/7019/files#diff-16a41bfbe414387d773df099f941217eR197).
`teardown()` have changed the name to `release()` in master and it's now called from the [`cli/index`](https://github.com/ember-cli/ember-cli/blob/7f6246d63ef820a3eb07fbbb021969dbdcb9a262/lib/cli/index.js#L145)

Feel free to close it. Just wanted to push this fix/refactoring somehow in order to unlock #7069 

todo:
 - [ ] Simplify dealing with `capture-exit` vs `process.exit()` in the `cli/cli-test`. Possible solution: https://github.com/ember-cli/capture-exit/pull/23 
 - [ ] catch "ember new foo, test, SIGINT exits with error and clears tmp/" with unit tests and fix

### DIFF
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const captureExit = require('capture-exit');
-captureExit.captureExit();
-
 const glob = require('glob');
 const Mocha = require('mocha');
 const RSVP = require('rsvp');
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const captureExit = require('capture-exit');
+captureExit.captureExit();
+
 const glob = require('glob');
 const Mocha = require('mocha');
 const RSVP = require('rsvp');
--- a/tests/unit/cli/cli-test.js
+++ b/tests/unit/cli/cli-test.js
@@ -6,6 +6,7 @@ const MockAnalytics = require('../../helpers/mock-analytics');
 const td = require('testdouble');
 const Command = require('../../../lib/models/command');
 const Promise = require('rsvp').Promise;
+const SilentError = require('silent-error');
 
 let ui;
 let analytics;
@@ -65,10 +66,16 @@ function stubRun(name) {
 }
 
 describe('Unit: CLI', function() {
+  let teardownHandler;
   beforeEach(function() {
+    teardownHandler = td.function();
+    let teardown = td.function();
+    let addHandler = td.function();
+    td.when(addHandler(td.matchers.isA(Function))).thenReturn(teardownHandler);
+
     willInterruptProcess = td.replace('../../../lib/utilities/will-interrupt-process', {
-      addHandler: td.function(),
-      removeHandler: td.function(),
+      addHandler,
+      teardown,
     });
 
     CLI = require('../../../lib/cli/cli');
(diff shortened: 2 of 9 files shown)


stefanpenner:
hmm, looks like windows has a legitimate failure.

(also thanks for continuing to work on this, much appreciated!)

ro0gr:
Yes, I've noticed that. I'll take a look there

tests/runner.js
```diff
@@ -1,8 +1,5 @@
 'use strict';
 
-const captureExit = require('capture-exit');
-captureExit.captureExit();
```
ro0gr: I hope we don't need this after we abstracted `process` in `MockProcess`(I may be wrong)

tests/unit/cli/cli-test.js
```diff
@@ -66,11 +79,21 @@ function stubRun(name) {
...
+        // force `process-exit` to handle exit
+        process.exit();
+      },
     });
 
+    // capture-exit doesn't support `process` injection
+    // instead it always uses global `process`
+    origExit = process.exit;
+    // if we leave original exit then test run will be stopped once we send SIGINT
+    process.exit = function() {};
```
ro0gr: I've tried to explain the intention for this juggling with `exit` in comments. Basically the problem is that currently we can't path `process` instance to the `capture-exit.captureExit()`.

Btw I prefer to avoid changes in the `capture-exit` in the scope of this pr.

tests/unit/cli/cli-test.js
```diff
@@ -228,43 +254,78 @@ describe('Unit: CLI', function() {
...
-      return ember(['serve']).finally(function() {
-        td.verify(willInterruptProcess.removeHandler(onCommandInterrupt));
+      registerCommand(FakeCommand.extend({
+        run() {
+          throw error;
+        },
+      }));
+
+      return expect(ember(['fake'])).to.be.rejectedWith(error).then(() => {
+        expect(interruptionHandeled).to.equal(false);
```
ro0gr: FYI the [original](https://github.com/ember-cli/ember-cli/pull/7019/files#diff-aa94f362c4d6ae97d4d491b3a8cc4887R256) test tried to check that cleanup happens on error.  Anyway there is no such logic in cli.
Do we really want to cleanup if every time command throws an exception?

ro0gr:
status update: trying to catch 'ember new foo, test, SIGINT exits with error and clears tmp/' with unit tests

ro0gr:
I've played a little with `MockProcess` injection into the `capture-exit` and I don't really like this approach.

`MockProcess.exit()` isn't able to simulate a real `exit` behavior.  
The real `exit` allows a current nodejs event loop tick to be completed. but the next one tick will never happen, `MockProcess.exit()` can't achieve that:

```js
.then(() => mockProcess.exit())
.then(() => {
  // this will be executed
})
```
This makes unit testing of `cli` shutdown much less reliable(predictable). 

Currently I think it may worth an exploration to make a Mock for `capture-exit`. 
It will allow us to simulate integration with `capture-exit` as close as we can without [any changes in `capture-exit`](https://github.com/ember-cli/capture-exit/pull/23).

ro0gr:
close in favor of #7285

PR #7232, opened 2017-07-16
ENHANCEMENT - don't compress responses with the x-no-compression response header

akatov:
I'm using an http-proxy to proxy to an endpoint intended for long polling. Now this endpoint writes a newline to the HTTP connection in regular intervals to make sure the connection stays alive.

The compression middleware however buffers the output produced by the endpoint (and the http-proxy) waiting for it to complete so it can compress it. This leads to the initiator closing the connection after not receiving back any data.

Having the proposed change in place lets me circumvent this problem by setting a header on the response object like this:
```js
// server/proxies/long-poll.js
module.exports = function(app) {
  app.use('/long-poll', function(req, res) {
    res.setHeader('x-no-compression', 'true');
    require('http-proxy').createProxyServer().web(req, res, {
      target: 'http://long-polling.service.lvh.me:1234'
    });
  });
};
```

### DIFF
--- a/lib/tasks/server/express-server.js
+++ b/lib/tasks/server/express-server.js
@@ -192,7 +192,17 @@ class ExpressServerTask extends Task {
 
   startHttpServer() {
     this.app = this.express();
-    this.app.use(require('compression')());
+    const compression = require('compression');
+    this.app.use(compression({
+      filter(req, res) {
+        if (res.getHeader('x-no-compression')) {
+          // don't compress responses with this response header
+          return false;
+        } else {
+          return compression.filter(req, res);
+        }
+      },
+    }));
 
     this.setupHttpServer();
 
--- a/tests/unit/tasks/server/express-server-test.js
+++ b/tests/unit/tasks/server/express-server-test.js
@@ -210,6 +210,83 @@ describe('express-server', function() {
         });
     });
 
+    describe('compression', function() {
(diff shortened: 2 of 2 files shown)


stefanpenner:
I was going to say, maybe expressjs/compression supports this already, but apparently there readme suggests your approach https://github.com/expressjs/compression#filter-1!

I am fine with this change, but it likely needs a unit test. Or it may regress overtime.

stefanpenner:
Also looks like your tests are failing due to an eslint issue:

```
not ok 10 eslint should have no errors in lib
  lib/tasks/server/express-server.js
    204:8  error  Missing trailing comma  comma-dangle
  
  ✖ 1 problem (1 error, 0 warnings)
  
      at Context.<anonymous> (node_modules/mocha-eslint/index.js:36:15)
```

locally, you can run `npm run lint` to check linting rules without having to run the full tests.

akatov:
I fixed the eslint issue.
I'd also be grateful if you could point me to most fitting existing unit test to use as a blueprint for this feature?

akatov:
OK, added a unit test. Not really sure how to check whether the actual body is compressed, so I'm just checking for the correct response headers...

akatov:
rebased on v2.15.1

rwjblue:
Thanks for working on this!

PR #7233, opened 2017-07-16
Convert blueprints to use modules and bump ember-cli-babel

rwwagner90:
I converted the blueprints used in `ember init` over to use the new modules syntax, and bumped the `ember-cli-babel` version to ensure backwards compatibility with the polyfill.

We will also need to convert all the blueprints people run for the generators themselves, which I can do in a followup PR.

### DIFF
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -1,9 +1,9 @@
-import Ember from 'ember';
+import Application from '@ember/application';
 import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
-const App = Ember.Application.extend({
+const App = Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
   Resolver
--- a/blueprints/app/files/app/router.js
+++ b/blueprints/app/files/app/router.js
@@ -1,7 +1,7 @@
 import EmberRouter from '@ember/routing/router';
 import config from './config/environment';
 
-const Router = Router.extend({
+const Router = EmberRouter.extend({
   location: config.locationType,
   rootURL: config.rootURL
 });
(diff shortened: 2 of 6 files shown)


stefanpenner:
This looks great!

@rwjblue whats our release plan for the new module API?

rwwagner90:
@stefanpenner do we need to update all the blueprints people can run to generate files at the same time or can this be done as a first step? Happy to start on another PR for the other blueprints if we need that at the same time.

stefanpenner:
I suspect for updating peoples app a code-mode, maybe via ember-watson would be most appropriate?

Turbo87:
> do we need to update all the blueprints people can run to generate files at the same time

this will be hard anyway since they are inside of Ember.js, not Ember CLI, so it will depend on the Ember version if people will be able to take advantage of them. I'm also wondering if we should check the `ember-cli-babel` version of the host project before generating things and branch out depending on if that version supports the new module imports or not.

> I suspect for updating peoples app a code-mode, maybe via ember-watson would be most appropriate?

you mean like https://github.com/ember-cli/ember-modules-codemod? 😉

rwwagner90:
@Turbo87 I can make the conversion on the Ember.js repo as well, if you'd like.

rwjblue:
In general we are shooting for using new shims in blueprints as of 2.16. As mentioned by a few folks above, this will require changing things here (seems handled by this PR), in emberjs/ember.js, and the guides. All of the changes should be done in master branch (which means it targets 2.16) for each project (though the guides will have to hold off on merging due to CD). 

Tldr; we should land these changes whenever we are comfortable with them (they look good to me).

rwjblue:
>  I'm also wondering if we should check the ember-cli-babel version of the host project before generating things and branch out depending on if that version supports the new module imports or not.

Ya, I agree. The imports either will need to be made locals or we need to enable ember-cli-babel@5 to support the new imports (which I'd prefer not to do)...

PR #7234, opened 2017-07-16
Revert "Revert "Remove arbitrary *.js filtering for addon tree.""

stefanpenner:
This reverts commit dee542cdf9aad28df93b093ccdea21b0d9e103ab.

We should get this back in

- [ ] reproduction of the original issue
- [ ] ensure it is already addressed, or address the issue.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1137,11 +1137,7 @@ let addonProto = {
   addonJsFiles(tree) {
     this._requireBuildPackages();
 
-    let includePatterns = this.registry.extensionsForType('js')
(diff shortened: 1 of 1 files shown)


stefanpenner:
I could also pretty easily make plugin (or potentially adapt funnel) to fast-path this existing requirement. Rather then linkRoots, its essentially link all toplevel, except for `styles`...

stefanpenner:
the area of concern: https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L745-L750

Turbo87:

PR #7239, opened 2017-07-19
Removes private `_mergeTrees` function

twokul:
`_mergeTrees` served as a proxy around `mergeTrees` module; it doesn't seem to provide a lot of value.

I had to refactor out tests a little (as some of them were spying `app._mergeTrees`) and replaced it with similar tests + a new utility.

Update:

per @rwjblue request -- it seems like addons don't rely on `app._mergeTrees`. [Search on EmberObserver](https://emberobserver.com/code-search?codeQuery=_mergeTrees) returned 3 results but after double-checking the source code, it doesn't look like any of them are still using it.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -40,6 +40,7 @@ const Funnel = require('broccoli-funnel');
 const funnelReducer = require('broccoli-funnel-reducer');
 const logger = require('heimdalljs-logger')('ember-cli:ember-app');
 const addonProcessTree = require('../utilities/addon-process-tree');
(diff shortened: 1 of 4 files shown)


rwjblue:
In general this seems fine. I believe this was mostly only around to aid in testing, and IMHO the mock style tests end up being pretty brittle and provide little validation of the actual fundamental functionality that things rely on. 

Also, Please double check with an emberobserver.com/code-search to ensure this is not being used in addons. At one point I believe it was used by some addons that we would prefer to avoid breaking (though hopefully they have since updated away).

PR #7245, opened 2017-07-22
Add API to allow addons to define and use custom transform with app.import

kratiahuja:
Implementation of RFC: https://github.com/ember-cli/rfcs/pull/108

## TODO:
- [ ] Update ember-cli website with new API

cc: @stefanpenner @hjdivad @rwjblue @kellyselden

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -28,7 +28,6 @@ const concat = require('broccoli-concat');
 const ConfigReplace = require('broccoli-config-replace');
 const ConfigLoader = require('broccoli-config-loader');
 const mergeTrees = require('./merge-trees');
-const shimAmd = require('./amd-shim');
 const WatchedDir = require('broccoli-source').WatchedDir;
 const UnwatchedDir = require('broccoli-source').UnwatchedDir;
 
@@ -129,7 +128,8 @@ class EmberApp {
 
 
     this._scriptOutputFiles = {};
-    this.amdModuleNames = null;
+    this._customTransformOrder = [];
+    this._customTransforms = {};
 
     this.legacyFilesToAppend = [];
     this.vendorStaticStyles = [];
@@ -145,6 +145,7 @@ class EmberApp {
     this.project.addons.forEach(addon => addon.app = this);
     p.setupRegistry(this);
     this._notifyAddonIncluded();
+    this._importAddonTransforms();
 
     if (!this._addonInstalled('loader.js') && !this.options._ignoreMissingLoader) {
       throw new SilentError('The loader.js addon is missing from your project, please add it to `package.json`.');
@@ -505,9 +506,9 @@ class EmberApp {
     return !this._addonDisabledByBlacklist(addon) && !this._addonDisabledByWhitelist(addon);
   }
 
-
-
   /**
+    Calls the included hook on addons.
+
     @private
     @method _notifyAddonIncluded
   */
@@ -530,7 +531,7 @@ class EmberApp {
       });
     }
 
-    this.project.addons = this.project.addons.filter(function(addon) {
+    this.project.addons = this.project.addons.filter(addon => {
       if (this.shouldIncludeAddon(addon)) {
         if (addon.included) {
           addon.included(this);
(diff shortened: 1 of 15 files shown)


kratiahuja:
Travis is failing only on Node 4 on unrelated changes.

Turbo87:
@kratiahuja can we close https://github.com/ember-cli/ember-cli/pull/7088 or is that unrelated?

kratiahuja:
Yes we can close that. This supersedes that.

lib/broccoli/ember-app.js
```diff
@@ -128,7 +127,8 @@ class EmberApp {
 
 
     this._scriptOutputFiles = {};
-    this.amdModuleNames = null;
+    this.customTransformOrder = null;
```
kratiahuja: yes agreed. I forgot to do that. Will update.

lib/broccoli/ember-app.js
```diff
@@ -529,15 +530,55 @@ class EmberApp {
...
+    @private
+    @method _importAddonTransforms
+  */
+  _importAddonTransforms() {
+
+    this.project.addons.forEach(addon => {
+      if (this.shouldIncludeAddon(addon)) {
+        if (addon.importTransforms) {
+          if (!this.customTransformOrder) {
+            this.customTransformOrder = [];
```
kratiahuja: The goal is not laziness. I was just following the format of `_amdModuleNames`

lib/broccoli/ember-app.js
```diff
@@ -529,15 +530,55 @@ class EmberApp {
...
+          }
+
+          if (!this.customTransforms) {
+            this.customTransforms = {};
```
kratiahuja: Ditto as above

lib/broccoli/ember-app.js
```diff
@@ -529,15 +530,55 @@ class EmberApp {
...
+
+          if (!transforms) {
+            throw new Error(`Addon "${addon.name}" did not return a transform map from importTransforms function`);
+          }
+
+          Object.keys(transforms).forEach(transformName => {
+            // store the transform order to reterieve later
+            this.customTransformOrder.push(transformName);
+            // store the transform info
+            this.customTransforms[transformName] = {
```
kratiahuja: Yes I totally forgot about that. I'll fix that.

lib/broccoli/ember-app.js
```diff
@@ -1097,14 +1138,19 @@ class EmberApp {
...
+      if (this.customTransformOrder) {
+        this.customTransformOrder.forEach(transformName => {
+          let transformConfig = this.customTransforms[transformName];
+
+          let transformTree = new Funnel(externalTree, {
+            files: transformConfig.files,
+            annotation: `Funnel (custom transform: ${transformName})`,
+          });
+
+          externalTree = this._mergeTrees([externalTree, transformConfig.callbackFn(transformTree, transformConfig.options)], {
```
kratiahuja: Yes, `transformConfig.options` is a map of asset path to the options provided by `app.import`. We pass the whole map and read off the assetPath

lib/broccoli/ember-app.js
```diff
@@ -1663,20 +1710,32 @@ class EmberApp {
...
+            let availableTransformNames = this.customTransformOrder.join(',');
+            throw new Error(`while import ${assetPath}: found an unknown transformation name ${transformName}. Available transformNames are: ${availableTransformNames}`);
           }
-          // If the import is specified to be a different name we must break because of the broccoli rewrite behavior.
-          if (this.amdModuleNames[assetPath] && this.amdModuleNames[assetPath] !== entry.as) {
-            throw new Error(`Highlander error while importing ${assetPath}. You may not import an AMD transformed asset at different module names.`);
+
+          if (transformName === 'amd') {
+            // some custom amd checks
+            if (!entry.as) {
```
kratiahuja: Sure I see what you are saying. Basically expose a public hook for addons defining custom transform to also validate/process any additional options right?

rwjblue: Yep, exactly

lib/broccoli/ember-app.js
rwjblue: Hehe, it confuses me.

lib/broccoli/ember-app.js
kratiahuja: Sure sounds reasonable.

lib/broccoli/ember-app.js
rwjblue: Gotcha. I see it now, thank you for explaining.

lib/broccoli/ember-app.js
kratiahuja: Agreed. I'll fix it :)

kratiahuja:
@rwjblue updated per feedback. I changed the `importTransforms` API to accomodate `processOptions` as well. `importTranforms` can either define only the callback or callback and processOptions. 

Therefore it can have a structure as when defining `processOptions`:

```js
importTransforms() {
  return {
     'amd': { 
        transform: function(tree, options) {},
        processOptions: function(assetPath, entry, options) {
            // do the processing
            return options;
       }
     }
  }
}
```

When the `processOptions` does not need to be defined, the API can be used as:

```js
importTransforms() {
  return {
     'amd': function(tree, options) {},
  }
}
```

I rarely see addons requiring to processOptions, therefore wanted to make sure we support both.

lib/broccoli/ember-app.js
```diff
@@ -530,15 +531,63 @@ class EmberApp {
...
+            let transformConfig = {
+              files: [],
+              options: {},
+            };
+
+            // store the transform info
+            if (typeof transforms[transformName] === 'object') {
+              transformConfig['callback'] = transforms[transformName].transform;
+              transformConfig['processOptions'] = transforms[transformName].processOptions;
+            } else {
```
rwjblue: Should we have this be an `} else if (typeof transforms[transformName] === 'function') {}` then add a "real" `else` clause that throws?

lib/broccoli/ember-app.js
```diff
@@ -1658,20 +1711,22 @@ class EmberApp {
...
+          if (this._customTransformOrder.indexOf(transformName) === -1) {
+            let availableTransformNames = this._customTransformOrder.join(',');
+            throw new Error(`while import ${assetPath}: found an unknown transformation name ${transformName}. Available transformNames are: ${availableTransformNames}`);
           }
-          // If the import is specified to be a different name we must break because of the broccoli rewrite behavior.
-          if (this.amdModuleNames[assetPath] && this.amdModuleNames[assetPath] !== entry.as) {
-            throw new Error(`Highlander error while importing ${assetPath}. You may not import an AMD transformed asset at different module names.`);
+
+
+          if (this._customTransforms[transformName].processOptions) {
```
rwjblue: I'd prefer to just default `processOptions` to `(options) => options;`, which allows us to avoid the extra branching and simplifies this a tiny bit.

lib/broccoli/ember-app.js
kratiahuja: I just felt it was an unrequired check :)

lib/broccoli/ember-app.js
kratiahuja: Sure I can do that.

kratiahuja:
> I believe that we still need to add API docs to lib/models/addon.js for the new hook.

Yes I forgot about that too.

kratiahuja:
@rwjblue updated per our discussion. I wanted to add a link to the AMD addon as an example usage, I'll update that once this lands (so I have a link as well :) )

kratiahuja:
@rwjblue is this good to go?

PR #7246, opened 2017-07-22
[BUGFIX release] cherry-pick "Bumps `broccoli-builder` version to include stack traces fix

twokul:


stefanpenner:
Shipit once green!

PR #7247, opened 2017-07-23
Moves documentation files under `docs/`

twokul:
we had docs in both places. I figured it'd be good to consolidate them all in one place


kellyselden:
There is ongoing work in #7079 for this, with significant discussion about what files can and can't be moved.

PR #7249, opened 2017-07-24
[FIX] Allow preprocessors to use module unification dir structure

rtablada:
Currently using the MODULE_UNIFICATION flag works for traditional CSS files, but there was a hardcoded reference to `/app/styles` when instantiating preprocessors.

I have tested this with a new module unification app with sass stylesheets in:

* app/styles/app.scss
* src/ui/styles/app.scss

This may not solve all issues as there was an existing TODO in the file noting a needed fix for CSS preprocessing with module unification, but this is a first stab and gets `ember-cli-sass` to compile correctly.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -904,7 +904,11 @@ class EmberApp {
 
     // TODO: This isn't quite correct (but it does function properly in most cases),
     // and should be re-evaluated before enabling the `MODULE_UNIFICATION` feature
-    this._srcAfterStylePreprocessing = preprocessCss(srcAfterPreprocessTreeHook, '/src/ui/styles', '/assets', options);
+    let srcStylesPath = this._resolveLocal('src/ui/styles');
+
+    if (existsSync(srcStylesPath)) {
+      this._srcAfterStylePreprocessing = preprocessCss(srcAfterPreprocessTreeHook, '/src/ui/styles', '/assets', options);
+    }
 
     let srcAfterTemplatePreprocessing = preprocessTemplates(srcAfterPreprocessTreeHook, {
       registry: this.registry,
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1397,11 +1397,21 @@ class EmberApp {
         throw new SilentError(`Style file cannot have the name of the application - ${this.name}`);
       }
 
+      let directory = '/app/styles';
+
+      if (experiments.MODULE_UNIFICATION) {
+        let srcStylesPath = this._resolveLocal('src/ui/styles');
+
+        if (existsSync(srcStylesPath)) {
+          directory = '/src/ui/styles';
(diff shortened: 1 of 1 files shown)


lib/broccoli/ember-app.js
```diff
@@ -904,7 +904,11 @@ class EmberApp {
 
     // TODO: This isn't quite correct (but it does function properly in most cases),
     // and should be re-evaluated before enabling the `MODULE_UNIFICATION` feature
-    this._srcAfterStylePreprocessing = preprocessCss(srcAfterPreprocessTreeHook, '/src/ui/styles', '/assets', options);
+    let srcStylesPath = this._resolveLocal('src/ui/styles');
+
+    if (existsSync(srcStylesPath)) {
+      this._srcAfterStylePreprocessing = preprocessCss(srcAfterPreprocessTreeHook, '/src/ui/styles', '/assets', options);
```
rwjblue: I'm unsure about this exists check. The check prevents some currently supported use cases (e.g. having src/ui/styles/app.scss auto-generated by an earlier preprocessor). 

Can you explain what this change accomplishes (independent of the other change)?

lib/broccoli/ember-app.js
```diff
@@ -1397,11 +1401,21 @@ class EmberApp {
         throw new SilentError(`Style file cannot have the name of the application - ${this.name}`);
       }
 
+      let directory = '/app/styles';
+
+      if (experiments.MODULE_UNIFICATION) {
```
rwjblue: This doesn't seem quite correct to me. We don't want to confuse the existing `this.trees.styles` with module unification at all. As written, when ember-cli itself supports module unification, we would no longer support app/styles, which seems bad...

What are you shooting for here?

lib/broccoli/ember-app.js
rtablada: With an app that is partially still using some folders in the app directory, a build with preprocessing in the `app/styles` directory (possibly limited by addon trees support at the time) fails without this check.

I'd definitely be interested in a better way to check this rather exists. I brought this in looking at the logic in the existing `_initOptions` method.

lib/broccoli/ember-app.js
rtablada: This is the main piece of code required to support preprocessor support within the `src/ui/styles` directory.

Without this check passing in the styles directory tree path the preprocessors (at least when tested with sass) would get `app/styles` hardcoded.

The `if` statements here are to check that module unification is enabled and that the current project has styles located in the `/src/ui/styles` directory (adopted logic from the `_initOptions` method).

rtablada:
@rwjblue left feedback to your two comments, I'm working up a quick repo to show directory use cases and failures of each edgecase that I accounted for (you will likely spot more).

rtablada:
:wat: This morning this exact example was breaking. I'm closing until I am able to reproduce.

PR #7251, opened 2017-07-24
Private `_preMergeJavascript` hook

twokul:
# What's happening here?

Adding a new private hook that exposes all javascript trees _before_ they are merged into one application tree. This is the first step forward to implementing a more flexible build pipeline (tree-shaking & code splitting).

# Why?

This will enable us to start experiments with tree shaking.

On a high level (w/o going into specifics just yet), to implement tree shaking in Ember CLI we want two things:
  + an array of broccoli trees that make up the app
  + data structure that would tell us about the structure of the application & its dependencies

Assuming we have both of them, we could figure out what is needed and what is not & then proceed to invoking Broccoli to build the application.

[`toTree`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1706-L1713) method will:

+ invoke [`toArray`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1682-L1696) to get an array of trees (`styles`, `javascript`, `public`, `index`, test trees if needed)
+ merge all of them into one, run `postprocess` and return a single tree

The first issue with this approach is that methods that produce specific trees make assumptions about how the resulting tree should be composed.

Let's take `javascript` method for example. It returns a single tree. We merge `config`, `templates`, `app` (pre-processed app, post-processed app), `vendor`, `bower`, `addons`, `node_modules`, `amd_modules`, `ember-cli` and also `concat` trees.

A more flexible way (keeping in mind that we want to enforce separation of concerns) would be exposing an array of trees (w/o merges or concatinations) and let the next step in the build pipeline put them all into a `javascript` tree.

The second issue is that we don't have access to original `app` or `addons` trees, for example. That makes it hard to implement "tree-shaking".

Internally, we should operate on an array of trees as long as possible and merge into a tree at the very end.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -856,12 +856,14 @@ class EmberApp {
   /**
     @private
     @method _processedAppTree
+    @param {Array} addonAppTrees Array of broccoli trees from add-ons that contain `app/` directories.
+    @param {Object} appTree Broccoli tree that represents `app/` directory of the Ember application.
     @return
   */
-  _processedAppTree() {
+  _processedAppTree(addonAppTrees, appTree) {
     let appTrees = [].concat(
-      this.addonTreesFor('app'),
-      this._filterAppTree()
+      addonAppTrees,
+      appTree
     ).filter(Boolean);
 
     let mergedApp = mergeTrees(appTrees, {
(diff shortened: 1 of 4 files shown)


twokul:
@stefanpenner @rwjblue

rwjblue:
@twokul - This was replaced, right?

twokul:
@rwjblue yeah. we might need to reuse some of the work here but should be fine to close.

PR #7257, opened 2017-07-26
fix(package): update chalk to version 2.0.1

locks:
Closes #7170


locks:
I'm unsure if I proceeded correctly with the yarn.lock, reviewer please double check.

package.json
```diff
@@ -111,6 +112,7 @@
     "semver": "^5.1.1",
     "silent-error": "^1.0.0",
     "sort-package-json": "^1.4.0",
+    "strip-ansi": "^4.0.0",
```
kellyselden: Should this be a dev dependency?

lib/utilities/markdown-color.js
```diff
@@ -142,7 +142,9 @@ function renderStylesFactory(renderer, styleNames) {
   @param {String} [name]
 */
 function checkStyleName(renderer, name) {
-  if (Object.keys(renderer.styles).indexOf(name) > -1) {
+  const styles = renderer.styles || require('ansi-styles');
```
kellyselden: We tend to use `let` in function scope.

rwjblue: Ya, except when it's the result of `require`.

package.json
twokul: it seems like the right place (since `strip-ansi` was just extracted from `chalk` and we depended on `chalk`); we use styles from `strip-ansi` to print help out

rwjblue: 👍

kellyselden: I only see it used in tests. Am I missing something?

Turbo87: @kellyselden you're right, and that is why @locks apparently changed it to be a dev dep 😉

PR #7258, opened 2017-07-26
Make blueprint.supportsAddon set this.options

jsturgis:
This change fixes an issue where `supportsAddon` calls files
without first setting `this.options`.

If `this.options` is not set then blueprints which implement the `filesPath`
hook will not be passed an `options` object.

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -23,7 +23,7 @@ class GenerateTask extends Task {
     let addonBlueprint = this.lookupBlueprint(`${name}-addon`, true);
     // otherwise, use default addon-import
 
-    if (noAddonBlueprint.indexOf(name) < 0 && !addonBlueprint && (mainBlueprint && mainBlueprint.supportsAddon()) && options.args[1]) {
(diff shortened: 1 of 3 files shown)


lib/tasks/generate-from-blueprint.js
```diff
@@ -23,7 +23,7 @@ class GenerateTask extends Task {
     let addonBlueprint = this.lookupBlueprint(`${name}-addon`, true);
     // otherwise, use default addon-import
 
-    if (noAddonBlueprint.indexOf(name) < 0 && !addonBlueprint && (mainBlueprint && mainBlueprint.supportsAddon()) && options.args[1]) {
+    if (noAddonBlueprint.indexOf(name) < 0 && !addonBlueprint && (mainBlueprint && mainBlueprint.supportsAddon(options)) && options.args[1]) {
```
tastycode: This is a pretty complex statement. For clarity, consider extracting all or parts of this to intention revealing variables.
```javascript
const addonBlueprintNotFound = noAddonBlueprint.indexOf(name) < 0;
const addonSupported = mainBlueprint && mainBlueprint.supportsAddon(options);
const blueprintName = options.args[1]; // just a guess
if (!addonBlueprint && addonBlueprintNotFound && addonSupported  && blueprintName) {
```

PR #7259, opened 2017-07-26
Convert MarkdownColor to class syntax

locks:


lib/utilities/markdown-color.js
```diff
@@ -33,109 +31,117 @@ module.exports = MarkdownColor;
...
-  let colorTokens = unshiftValue(colors.concat(backgroundColors).map(getToken), {}).reduce(setToken);
-  return merge(colorTokens, defaultTokens);
-};
+  /*
+    Generate default style tokens
+
+    @method generateTokens
+    @return {Object}
+  */
+  generateTokens(renderer) {
```
twokul: it looks like this method is internal to the class; at the very least we should add `@private` to the docs and change the name to `_ generateTokens` imo. I'd probably go as far as extracting this as a util function and not putting it on the class.

lib/utilities/markdown-color.js
```diff
@@ -33,109 +31,117 @@ module.exports = MarkdownColor;
...
+  }
 
+  /*
+    Looks up multiple styles to apply to the rendered output
+    @method renderStylesFactory
+    @param {Object} [renderer] should match chalk.js format
+    @param {String, Array} [styleNames]
+    @return {Function} Function that will wrap the input string with styling
+  */
+  renderStylesFactory(renderer, styleNames) {
```
twokul: same here

lib/utilities/markdown-color.js
```diff
@@ -33,109 +31,117 @@ module.exports = MarkdownColor;
...
-  };
+  /*
+    Parse markdown and output as string
+
+    @method render
+    @param {String} [string]
+    @param {Object} [options]
+    @return {String}
+  */
+  render(string, options) {
```
twokul: it doesn't seem like `render` is used anywhere?

locks: it's being specifically tested for

lib/utilities/markdown-color.js
```diff
@@ -33,109 +31,117 @@ module.exports = MarkdownColor;
...
+    this.markdown = markdown().use(terminal, markdownOptions);
+    this.tokens = merge(tokens, optionTokens);
   }
 
-  return this.render(file, options);
-};
+  /*
+    Lookup markdown file and render contents
+
+    @private
```
Turbo87: why is this private now? seems like an unrelated change

lib/utilities/markdown-color.js
```diff
@@ -173,26 +180,30 @@ function getColorTokenRegex(name, options) {
...
 
 function escapeToken(match, pattern) {
   let output = pattern.replace(/\~/g, '');
+
   return `<${output}>`;
 }
 
 function unshiftValue(array, value) {
   array.unshift(value);
+
```
Turbo87: why the extra whitespace?

lib/utilities/markdown-color.js
Turbo87: @twokul the point of the PR is converting to class syntax. IMHO it shouldn't change any visibility levels, behaviors, or rename anything.

lib/utilities/markdown-color.js
locks: I find it tidier when returns are separated.

lib/utilities/markdown-color.js
```diff
@@ -177,7 +177,7 @@ function getToken(name, options) {
     name,
     token: `<${name}></${name}>`,
     pattern: getColorTokenRegex(name),
-    render: renderStylesFactory(renderer, name),
+    render: this.renderStylesFactory(renderer, name),
```
Turbo87: how does this function get access to `this`?

locks: Right, this was a tricky one. I'm binding `this`: https://github.com/ember-cli/ember-cli/pull/7259/files#diff-f6db1f1718f3ab9018f41d583b46f03dR111. Should I pull it into the class? I didn't want to change the surface API.

locks:
🏓

PR #7253, opened 2017-07-24
[do not merge] Investigate "process already captured" test failure

ro0gr:
Currently there are strange test failures on CI:
 - win + node@4 - https://ci.appveyor.com/project/embercli/ember-cli/build/13850/job/mnst7jeu3vg4isdw/tests
 - linux + node@8 - https://travis-ci.org/ember-cli/ember-cli/jobs/257052096#L1896

I've tried to reproduce it locally with different combinations of `node`/`npm`/`os` but no luck.

In this pr I've added stack traces for each process capture/teardown. I hope it can give more details on this issue.

PR #7261, opened 2017-07-26
double test timeout for install-test-slow

ro0gr:
this installs 2 npm packages:
 - ember-cli-fastclick
 - ember-cli-photoswipe

Then it installs `bower` in order to install bower deps.

In some environments(#7253) 60 sec is not enough.


ro0gr:
another option might be to install `bower` globally on CI

PR #7262, opened 2017-07-26
Convert to classes

twokul:


tests/helpers/proxy-server.js
```diff
@@ -2,17 +2,19 @@
...
-    _this.called = true;
-    _this.lastReq = req;
-    res.writeHead(200, { 'Content-Type': 'text/plain' });
-    res.end('okay');
-  });
-  this.httpServer.listen(3001);
-};
+class ProxyServer {
+  constructor() {
+    let _this = this;
```
Turbo87: seems unneeded if you convert to arrow functions too here

Issue #7243, opened 2017-07-20
Empty trees & `addon.treeFor`


twokul:
![screen shot 2017-07-20 at 9 40 14 am](https://user-images.githubusercontent.com/1131196/28428837-815a650c-6d2f-11e7-9a0a-5d60409dbb85.png)

Above is the output of `addonTreesFor(type);` (this particular one is for `'app'`). It's a newly created app (I only added `ember-cli-sass` and `ember-wormhole`).

We iterate over add-ons and call `treeFor` only to get back an empty tree. It seems wasteful to merge empty trees. I wonder if we could somehow hint to ember-cli that the add-on doesn't actually return a particular type of the tree so there's no need to call `treeFor`.

twokul:
turns out we have a function to [spot empty trees](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/merge-trees.js#L12-L33), we should use it.

rwjblue:
Those trees are almost certainly not empty.  They most likely contain a single `.gitkeep` file (please `ls -lah` one of the folders to confirm).  As you know, `Finder` does not show dot files by default.

That being said, I do agree that we should probably return `null` from `treeFor` when we know that the tree in question contains only a single `.gitkeep` file, and we should also probably consider removing `blueprints/addon/files/addon/.gitkeep` and `blueprints/addon/files/app/.gitkeep`.

twokul:
@rwjblue

![screen shot 2017-07-21 at 5 19 05 am](https://user-images.githubusercontent.com/1131196/28463206-2c899458-6dd4-11e7-911c-73d52901d272.png)

I like the idea of removing `.gitkeep`. I'll submit a PR and we can 👍 or 👎

locks:
Closing due to inactivity and not being relevant in a v2 addon format world.

PR #7240, opened 2017-07-19
[WIP] Splitting app & introducing assembler

twokul:

Issue #7228, opened 2017-07-14
Dump sub-trees of ember application for debugging/better visibility into pipeline to fs


twokul:
I wonder if it would be useful to have a flag that we pass to `ember build` that will allow to dump sub-trees of application to be written into "debug" folder; this mostly for debugging purposes. It also will give users better visibility into how `ember-cli` build pipeline works (which trees it operates on).

I found it super useful while working on tree shaking. I currently use `broccoli-debug` to generate the following output:

<img width="750" alt="screen shot 2017-07-14 at 2 51 47 pm" src="https://user-images.githubusercontent.com/1131196/28232140-00de003a-68a4-11e7-80b8-5b3198134204.png">

Wonder if it's something we want to make available to everyone.

stefanpenner:
Seems good as an ENV flag, as long as it's clearly documented not to be relied upon.

rwjblue:
Should use broccoli-debug. It's made for exactly this purpose.

rwjblue:
We should definitely do this (just like we do debug output for the console).

twokul:
once https://github.com/ember-cli/ember-cli/pull/7240 lands, we could do it in assembler

PR #7264, opened 2017-07-27
Introduces Bundler

twokul:
After talking to @stefanpenner and @rwjblue about how we should approach tree shaking, I believe this is the minimal change we should make to proceed forward and start experimenting.

`Bundler` does not introduce any new functionality (it's basically an extraction of our `concat` step from `ember-app`) but it make it one place where we can access a single tree with application code, addons and vendor files (`bower` and `node_modules`). It makes it a perfect place to start experimenting with tree shaking (say, introduce a new Broccoli Plugin that would do the heavy lifting).

There are couple of other PRs that I opened that helped me to understand the problem space better and ultimately could be closed (for now):

+ https://github.com/ember-cli/ember-cli/pull/7251/ is a good summary of the changes we need to do to land functionality described in https://github.com/ember-cli/ember-cli/issues/7228 and help us track down problems described in https://github.com/ember-cli/ember-cli/issues/7243
+ https://github.com/ember-cli/ember-cli/pull/7240/ is a good summary of the changes we need to make to split out "tree generation" and "config" (general thinking there is that we should be able to specify how to build trees in `ember-app` and let something else build it for us; this will make it way easier to write unit tests and we can finally trim down slow acceptance tests).

### DIFF
--- a/lib/broccoli/bundler.js
+++ b/lib/broccoli/bundler.js
@@ -1,0 +1,89 @@
+'use strict';
+
+const concat = require('broccoli-concat');
+const processModulesOnly = require('./babel-process-modules-only');
+
+module.exports = class Bundler {
+  /*
+    Responsibility of this class is to intelligently produce
+    the final output given a list of trees that were passed in.
+
+    @class Bundler
+    @constructor
+    @param {Object} Configuration options
+   */
+  constructor(options) {
+    this.name = options.name;
+    this.sourcemaps = options.sourcemaps;
+    this.appOutputPath = options.appOutputPath;
+    this.isBabelAvailable = options.isBabelAvailable;
+  }
+
+  /*
+    Concatenates all javascript Broccoli trees into one, as follows:
+
+    Given an input tree that looks like:
+
+    ```
+    addon-tree-output/
+      ember-ajax/
+      ember-data/
+      ember-engines/
+      ember-resolver/
+      ...
+    bower_components/
+      usertiming/
+      sinonjs/
+      ...
+    the-best-app-ever/
+      components/
+      config/
+      helpers/
+      routes/
+      ...
+    vendor/
+      ...
+      babel-core/
+      ...
+      broccoli-concat/
+      ...
+      ember-cli-template-lint/
+      ...
+    ```
+
+    Produces a tree that looks like:
+
+    ```
+    assets/
+      the-best-app-ever.js
+      the-best-app-ever.map (if sourcemaps are enabled)
+    ```
+
+    @method bundleAppJs
+    @return {Tree} Concatenated tree (application and dependencies).
+   */
+  bundleJs(applicationAndDepsTree, options) {
+    options = options || {};
+
+    let appJs = concat(applicationAndDepsTree, {
+      inputFiles: [`${this.name}/**/*.js`],
+      headerFiles: [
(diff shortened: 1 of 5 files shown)


kratiahuja:
Doesn't this need a RFC as well?

Turbo87:
> Doesn't this need a RFC as well?

sounds to me like it's only an internal change for now that does not add any public APIs so it doesn't seem that an RFC is needed yet for this step

kratiahuja:
I think it would worth writing this down IMO to help brainstorm :). But I am fine either ways.

lib/broccoli/bundler.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const concatFiles = require('./concat-files');
+const processModulesOnly = require('./babel-process-modules-only');
+
+class Bundler {
```
stefanpenner: `module.exports =` at the `class` keyword please, that way its easy to see (at the definition) if something is exported or not

lib/broccoli/bundler.js
```diff
@@ -0,0 +1,92 @@
...
+    @class Bundler
+    @constructor
+    @param {Object} Configuration options
+   */
+  constructor(options) {
+    this.name = options.name;
+    this.sourcemaps = options.sourcemaps;
+    this.appOutputPath = options.appOutputPath;
+    this.isBabelAvailable = options.isBabelAvailable;
+    this.appTreeAnnotation = options.appTreeAnnotation;
```
rwjblue: it seems odd to have the annotation here, but expect the returned tree to come from the `bundleAppJs`

rwjblue:
> Doesn't this need a RFC as well?

I think the larger concept around how to tree shake definitely deserves some sort of RFC. Unfortunately, at the moment we can't really start on that doc without a way to test/play around with implementations.

My personal preference would be to to extract all of the concat things we do into the bundler (vendor styles, app styles, vendor javascript, app javascript) without changing the fundamental logic. At that point we can begin adding much better test coverage and experimenting with custom bundling strategies. 

Once we can do that, we will then be able to author an RFC...

rwjblue:
I'm in favor of landing this as is, and continuing to extract the "concat things" to the bundler (without changing _what_ is done or the dist output in any way) in follow-up PR's.  When that phase is completed the contents of the `javascript` and `style` methods will have been moved into the new bundler class.

I believe that @twokul is already working on the next transition (creation of `vendor.js`), and can start on styles once that is completed.

twokul:
> I think the larger concept around how to tree shake definitely deserves some sort of RFC. Unfortunately, at the moment we can't really start on that doc without a way to test/play around with implementations.

I started to gather my thoughts around it [here](https://gist.github.com/twokul/959a6c684b4da5851474669542d73586). But there are a lot of unknowns. After talking to @stefanpenner and @rwjblue about general direction, they suggested we move slowly and did incremental steps w/ constant re-evaluation. I think it makes a lot of sense and builds our confidence in our solution.

> I believe that @twokul is already working on the next transition (creation of vendor.js), and can start on styles once that is completed.

going to submit PR's today

kratiahuja:
> Once we can do that, we will then be able to author an RFC... 

Sounds like a perfect plan!

stefanpenner:
@kratiahuja ya, we will totally get some good RFC material out :). It'll be an important one.

PR #7266, opened 2017-07-28
Fix JSON format of asset sizes report

marcoow:
This fixes the JSON variant of the asset sizes report by

* allowing the `--json` option (which actually was supported already: https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/show-asset-sizes.js#L13)
* preventing the `showGzipped` flag from being printed to the JSON report

### DIFF
--- a/lib/models/asset-size-printer.js
+++ b/lib/models/asset-size-printer.js
@@ -35,7 +35,12 @@ class AssetPrinterSize {
     let ui = this.ui;
     return this.makeAssetSizesObject().then(files => {
       if (files.length !== 0) {
-        ui.writeLine(JSON.stringify({ files }));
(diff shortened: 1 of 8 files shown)


lib/models/asset-size-printer.js
```diff
@@ -35,7 +35,13 @@ class AssetPrinterSize {
     let ui = this.ui;
     return this.makeAssetSizesObject().then(files => {
       if (files.length !== 0) {
-        ui.writeLine(JSON.stringify({ files }));
+        console.log(files);
```
Turbo87: 🤔

marcoow: 🙀 - fixed

PR #7270, opened 2017-07-28
Cache Project model config.

krisselden:
Add-ons have the expectation that project.config is "cheap" to call and this is not true.  The config for a build should be stable.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -171,7 +171,8 @@ class Builder extends CoreObject {
       .catch(error => {
         this.processAddonBuildSteps('buildError', error);
         throw error;
-      });
+      })
+      .finally(this.finalizeBuild.bind(this));
   }
 
   /**
@@ -221,6 +222,14 @@ class Builder extends CoreObject {
     this.project.ui.writeDeprecateLine('Heimdalljs < 0.1.4 found.  Please remove old versions of heimdalljs and reinstall (you can find them with `npm ls heimdalljs` as long as you have nothing `npm link`d).  Performance instrumentation data will be incomplete until then.', !process._heimdall);
 
     return value;
+  }
+
+  /**
+   * @private
+   * @method finalizeBuild
+   */
+  finalizeBuild() {
+    this.project.configCache.clear();
   }
 }
 
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -52,6 +52,7 @@ class Project {
     this.setupNodeModulesPath();
     this.addonDiscovery = new AddonDiscovery(this.ui);
     this.addonsFactory = new AddonsFactory(this, this);
+    this.configCache = new Map();
 
     /**
       Set when the `Watcher.detectWatchman` helper method finishes running,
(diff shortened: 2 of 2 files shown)


lib/models/builder.js
```diff
@@ -171,7 +171,7 @@ class Builder extends CoreObject {
       .catch(error => {
         this.processAddonBuildSteps('buildError', error);
         throw error;
-      });
+      }).finally(() => this.project.configCache.clear());
```
rwjblue: Minor nit-pick, the rest of this promise chain is dot aligned, I would expect this `.finally` to be on the next line.

lib/models/project.js
```diff
@@ -52,6 +52,7 @@ class Project {
     this.setupNodeModulesPath();
     this.addonDiscovery = new AddonDiscovery(this.ui);
     this.addonsFactory = new AddonsFactory(this, this);
+    this.configCache = new Map();
```
rwjblue: Can you make this `_configCache`?  We don't want it to be considered public API on `Project`.

lib/models/project.js
```diff
@@ -203,6 +204,21 @@ class Project {
...
+    return c;
+  }
+
+  /**
+   * @private
+   * @method config
+   * @param  {String} env Environment name
+   * @return {Object}     Merged confiration object
+   */
+  configWithoutCache(env) {
```
rwjblue: Can you underscore this?

krisselden: I doc'd it as private is that not enough?

krisselden: there is a lot of @private methods without underscore.

rwjblue:
@krisselden and I tracked down the failure to an issue with an addon mutating the provided `EmberENV`

rwjblue:
Updates:

* I fixed the mutation issue in the PR's that cross-linked here (mostly from ember-cli-htmlbars / ember-cli-htmlbars-inline-precompile). Once those were fixed and released, the previously hanging tests pass.
* I updated the target branch of this to be `beta`, and rebased.  This will go out in 2.15.

rwjblue:
Ugh, CI is now failing due to a bad merge conflict resolution I did when merging `release` into `beta` branch.  I'll need to fix that before we can land this (cross linking https://github.com/ember-cli/ember-cli/issues/7276).

rwjblue:
#7276 was fixed, restarting Travis now...

vitch:
This change has had an undesired effect on our code which I just noticed while trying to update `ember-cli` from `2.12` to `2.16`... 

We may have been abusing the `config` hook previously so here is a bit more about our use-case:

We want to change the contents of our addon's merged config dependent on a setting from the consuming app. We were achieving this by setting a variable in the addons `included` hook and accessing it from the `config` hook.

The first time `config` fired was before `included` had fired. But it fired (many) more times and some of those were after the `included` hook had fired. Which meant that the config as rendered to the apps `meta` tag was correct.

Should a workflow like this be supported? Is there any way for the `config` hook to access a variable from the consuming app (we were using an option in `ember-cli-build` but we could change this) when it first fires?

The reason that we want to do this at compile-time rather than run-time is to redact certain pieces of config on publicly visible apps, otherwise we could just filter at runtime when all of the variables are available...

PR #7274, opened 2017-07-31
Update yarn.lock.

rwjblue:


Turbo87:
Unfortunately this seems outdated by now 😞

PR #7275, opened 2017-07-31
Allow server middleware to answer non-get (POST/PATCH...) requests

cibernox:
This PR goes hand by hand with [a recently merged PR in ember-cli-fastboot](https://github.com/ember-fastboot/ember-cli-fastboot/pull/457).

This PR just lifts the check that the request is a GET request so fastboot can answer other kind of requests unconditionally.

On the original thread [some concerns about security were raised](https://github.com/ember-fastboot/ember-cli-fastboot/pull/354#issuecomment-292804064). I'm unsure about what concerns are those since this is just allowing the app to access data provided by the user, and besides ember-cli is a build tool and not something to be used in production.

However, if those security concerns really seem real, which do not to me, I'd be willing to add some sort of configuration option based on which ember-cli will ignore non-GET requests like today or answer them, perhaps in `.ember-cli`

**APPROACH CHANGE**

Now this PR only contains a refactor to extract the logic of what kind of requests should the middleware answer to into a method so the users can swap that logic by something that suits them.

### DIFF
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -46,16 +46,10 @@ class HistorySupportAddon {
     let watcher = options.watcher;
 
     let baseURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL || options.baseURL);
-    let baseURLRegexp = new RegExp(`^${baseURL}`);
 
     app.use((req, res, next) => {
       watcher.then(results => {
-
-        let acceptHeaders = req.headers.accept || [];
-        let hasHTMLHeader = acceptHeaders.indexOf('text/html') !== -1;
-        let isForBaseURL = baseURLRegexp.test(req.path);
-
-        if (hasHTMLHeader && isForBaseURL && req.method === 'GET') {
+        if (this.shouldHandleRequest(req, options)) {
           let assetPath = req.path.slice(baseURL.length);
           let isFile = false;
           try { isFile = fs.statSync(path.join(results.directory, assetPath)).isFile(); } catch (err) { /* ignore */ }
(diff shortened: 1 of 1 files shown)


rwjblue:
CI failures likely related to https://github.com/ember-cli/ember-cli/issues/7276.

kratiahuja:
I prefer this is done as an opt-in and keeping the default behavior as before.

rwjblue:
I'm 👍 on the updated approach.  It seems like a nice refactoring to me...

cibernox:
@rwjblue is there any change that this can make it into beta? At the end there was no change in behaviour.

kellyselden:
@cibernox Any day now we will be releasing the next beta cycle. Would 2.16.0-beta.1 be good enough, or do you really need a new 2.15 beta?

cibernox:
Actually I was thinking in 2.15 stable, but 2.16.beta.1, other than I'd prefer to use a stable version, should be fine.

devinus:
It's unclear how to opt-in to this approach.

Right now I'm forced to set `req.method = 'GET'` in a middleware before `history-support-middle` in an in-repo addon to allow POSTs to pass through.

cibernox:
@devinus In your `ember-cli-build.js` you can override the `shouldHandleRequest` method of the `history-support-middleware`.

This is the original: https://github.com/ember-cli/ember-cli/blob/19e44befff0124a13e6c36526ba017bb462f7fea/lib/tasks/server/middleware/history-support/index.js#L64-L76

This is how I overwrite it: 

```js
  let middlewareAddon = defaults.project.addons.find((a) => a.name === 'history-support-middleware');
  middlewareAddon.shouldHandleRequest = function(req, options) {
    let acceptHeaders = req.headers.accept || [];
    let acceptsHTML = acceptHeaders.indexOf('text/html') !== -1 || acceptHeaders.indexOf('*/*') !== -1;
    if (!acceptsHTML) {
      return false;
    }
    let baseURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL || options.baseURL);
    let baseURLRegexp = new RegExp(`^${baseURL}`);
    return baseURLRegexp.test(req.path);
  };
  //...
}
```

As you can see, the only diffeerence is that I've removed 
```js
    if (req.method !== 'GET') {
      return false;
    }

```

PR #7282, opened 2017-08-03
{BACKPORT to release} [BUGFIX beta] ember test work with both —server and —path

stefanpenner:
This has turned out to be extremely useful/productivity boost. And poses low risk, so I would like to get another release out that includes this.


stefanpenner:
We may not backport this all the way back to `release` if `beta` is all I end up needing. I want to make sure the tests are green, and this is tee'd up just in-case.

rwjblue:
This landed in a release already, closing...

PR #7285, opened 2017-08-06
improve removal of interruption handler

ro0gr:
This pr attempts to push forward refactoring from #7019 

### Notable Changes:
 - `will-interrupt-process.addHandler(cb)` now returns a function which is resposible for handler cleanup. It simplifies `will-interrupt-process` api.
- Previously we removed interruption handler only if the command finished successfully. Now [we cleanup](https://github.com/ember-cli/ember-cli/pull/7285/files#diff-16a41bfbe414387d773df099f941217eR165) even on a [command crash](https://github.com/ember-cli/ember-cli/pull/7285/files#diff-aa94f362c4d6ae97d4d491b3a8cc4887R312). 

### Testing
The main issue is that `MockProcess.exit()` isn't able to simulate a real `process.exit()` behavior.:
```js
.then(() => mockProcess.exit())
.then(() => {
  // this will be executed
})
```
At the same time we can't unit test `cli/cli` with a real `process.exit`. If we allow `process.exit()` to be executed in the test environment then test run will be immediately stopped.

In order to avoid dealing with a real `process.exit` and work-around[ inability to pass process to the `capture-exit`](https://github.com/ember-cli/capture-exit/pull/23), [MockCapureExit](https://github.com/ember-cli/ember-cli/pull/7285/files#diff-aa94f362c4d6ae97d4d491b3a8cc4887R57) has been introduced. It doesn't 100% replicate all of the `capture-exit` functionality. Anyway I believe it should be enough to unit test what we can.

Unfortunately it's still not enough to properly unit test process interruption in the cli. We still rely on "test, SIGINT exits with error and clears tmp/" acceptance test.

### DIFF
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const captureExit = require('capture-exit');
-captureExit.captureExit();
-
 const glob = require('glob');
 const Mocha = require('mocha');
 const RSVP = require('rsvp');
--- a/tests/unit/models/builder-test.js
+++ b/tests/unit/models/builder-test.js
@@ -11,6 +11,8 @@ const td = require('testdouble');
 const chai = require('../../chai');
 let expect = chai.expect;
 let file = chai.file;
+const willInterruptProcess = require('../../../lib/utilities/will-interrupt-process');
+const MockProcess = require('../../helpers/mock-process');
 
 let root = process.cwd();
 let tmproot = path.join(root, 'tmp');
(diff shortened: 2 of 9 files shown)


tests/runner.js
```diff
@@ -1,8 +1,5 @@
 'use strict';
 
-const captureExit = require('capture-exit');
```
ro0gr: I believe we can remove this as far as now we don't use real `process.exit()` in unit tests.

tests/unit/models/builder-test.js
```diff
@@ -36,27 +38,36 @@ describe('models/builder.js', function() {
...
+      command = new BuildCommand(commandOptions());
+    });
+
     beforeEach(function() {
       return mkTmpDirIn(tmproot).then(function(dir) {
         tmpdir = dir;
         builder = new Builder({
           setupBroccoliBuilder,
           project: new MockProject(),
+          onProcessInterrupt: willInterruptProcess,
```
ro0gr: Seems like `onProcessInterrupt` should be renamed to `willInterruptProcess`.
Also If somebody has a better name we can give it a better name.

ro0gr:
I've checked ["adapter throws when --base-class is same as name"](https://github.com/emberjs/data/blob/d35feb17fd67f4ff2016f86b45d9e81b0f53a5d4/node-tests/blueprints/adapter-test.js#L61) from ember-data. It works with this pr.

I think we can try to write some similar integration style tests in `ember-cli` in order to replace `SIGINT` acceptance test.

tests/unit/models/builder-test.js
ro0gr: I've seen @stefanpenner used `shutdown` term few times. Is `onShutdown` better term?

ro0gr:
Looking at this again I think it's better to make another try. `MockCapureExit` is a red line which should be not crossed I think

hjdivad:
> Looking at this again I think it's better to make another try. `MockCapureExite is a red line which should be not crossed I think

Is there something we can add to the capture exit API to make this easier to test?

ro0gr:
@hjdivad initially I've tried to make process injectable into capture-exit https://github.com/ember-cli/capture-exit/pull/23. However it seems to me that the cost of capture-exit re-write(and making it more complex) just for better ember-cli unit testing is to high.

Also I think it's close to impossible to properly unit test process.exit() outside of will-interrupt-process because there is no way to truly mimic process.exit() behavior because real exit stops the event loop.

Currently I believe the most approachable way to test cli/cli interruption edges is integration tests.  I'm going to make a new pass on this soon.

PR #7286, opened 2017-08-07
Update `amd-name-resolver` version to enable parallel babel transpile

mikrostew:


rwjblue:
Can you confirm that the yarn.lock doesn't need updating also?

mikrostew:
You're right, it should be updated.

PR #7298, opened 2017-08-19
Only warn when unknown transform name is used and skip transformation.

kratiahuja:
# Motivation
Currently in Ember CLI if a transformation name with `app.import` isn't found, we throw an error. Ember Addons that are fastboot compatible would also want to use `app.import` with the extended transformation API. These addons don't have control if the host app is fastboot compatible or not. Hence every addon would need to check if host app has fastboot addon installed and only then use the transformation. 

In order to avoid these `if else` blocks across addons, if the transform name isn't found we will warn the user and not add the file to the list of transforms. This addons can be used in apps that don't have fastboot enabled.

# Tests

Added relevant tests

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1686,16 +1686,18 @@ class EmberApp {
 
           let transformName = entry.transformation;
 
-          if (this._customTransformOrder.indexOf(transformName) === -1) {
-            let availableTransformNames = this._customTransformOrder.join(',');
-            throw new Error(`while import ${assetPath}: found an unknown transformation name ${transformName}. Available transformNames are: ${availableTransformNames}`);
-          }
-
           // process options for the transform and update the options
           this._customTransforms[transformName].options = this._customTransforms[transformName].processOptions(assetPath,
             entry, this._customTransforms[transformName].options);
-
-          this._customTransforms[transformName].files.push(assetPath);
+          
+          if (this._customTransformOrder.indexOf(transformName) === -1) {
+            // if transformation not found we ignore doing the transformation with warning
+            let availableTransformNames = this._customTransformOrder.join(',');
+            this.project.ui.writeWarnLine(`while import ${assetPath}: found an unknown transformation name ${transformName}. Available transformNames are: ${availableTransformNames}`);
+          } else {
+            // transformation is valid, so add it to the list
+            this._customTransforms[transformName].files.push(assetPath);
+          }
         });
       }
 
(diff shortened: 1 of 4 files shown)


rwjblue:
In general I'm in favor of the motivation, but I don't think this goes far enough.  Specifically, I don't like that the addons doing "the right thing" would emit any console warning output. 

Would it be possible instead to extract a smaller addon from ember-cli-fastboot that is something like `ember-cli-fastboot-import-transform` which can be depended on by all addons wanting to use `this.import` with the specific transform? This way ember-cli-fastboot could bring it, but also the addons that need/want fastboot wrapping can also bring it. This would prevent both the issue that motivates this PR and my concern about noisy consoles for "good" addons...

Thoughts?

kratiahuja:
> Would it be possible instead to extract a smaller addon from ember-cli-fastboot that is something like ember-cli-fastboot-import-transform which can be depended on by all addons wanting to use this.import with the specific transform?

The problem in this case is addons want to support a world of with and without fastboot itself. Which world are they in at build time depends on the host app (which may or may not include the fastboot addon) which makes this problem a clone for all addons. 

We could provide a utility addon/lib like `ember-cli-fastboot-transform` which checks if the host app has fastboot addon and then uses the transform, otherwise it simply uses `app.import` without the transform. The reason I am worried about that is every other addon would end up with that check which could cause performance issues for the app at build time. Of course, we could cache if the fastboot addon is included by host app.

rwjblue:
@kratiahuja I'm suggesting that that addon would be the one to actually add the transform. There would not be extra checks per addon: they simply depend on the newly extracted addon and then just use the fast boot transform all the time..

kellyselden:
Something about every addon that wants fastboot support needing a new dependency feels not ideal to me. Would it be possible to do Krati's original suggestion but without any warning if the transform isn't found? Just import the file/module like normal in that case.

rwjblue:
> Something about every addon that wants fastboot support needing a new dependency feels not ideal to me.

FWIW, it is only addons that need the app.import transform (which is far from all addons needing fastboot support). To me it seems _very_ reasonable to have a dependency that adds the support your addon requires...

> Would it be possible to do Krati's original suggestion but without any warning if the transform isn't found?

It's too easy to swallow typos / errors in this case. Good feedback when you have made a typo or other simple mistake is very valuable.

kratiahuja:
@rwjblue a transform needs to be only registered once. Let's say if the other addon `ember-cli-fastboot-transform` adds/registers the transform. Even if the other addons depend on this addon, it would mean that addon essentially is re-registered n times and you will see a warning since we pick the last. That seems a false positive to me. 

I totally understand and agree of throwing the un required warning when not using fastboot and it is confusing to the app developer.

Here's what I recommend (which essentially means this PR can be closed):
1. `ember-cli-fastboot` will register/add the transform
2. We expose another utility function in [`fastboot-transform`](https://github.com/kratiahuja/fastboot-transform) which basically checks if the host app has fastboot addon installed (of course we can cache that) and if so it calls `app.import` using the registered fastboot transform other wise calls `app.import` without any transform. 

Addons are already depending on `fastboot-transform` for wrapping their libraries with `app.import`, which essentially means for them it becomes an easier migration for them. 

Thoughts?

kratiahuja:
Going to close this.

PR #7300, opened 2017-08-20
Refactor Custom Transformation logic

sangm:
Adding tests for `option.using`

Previously, we were keeping state in two data structures `customTransformsOrder`
and `customTransforms`.

`customTransformsOrder` was used to maintain the order. We can normalize both of
these data structures by using a Map.

### DIFF
--- a/tests/unit/broccoli/ember-app-test.js
+++ b/tests/unit/broccoli/ember-app-test.js
@@ -171,7 +171,7 @@ describe('EmberApp', function() {
 
     describe('loader.js missing', function() {
       it('does not error when loader.js is present in registry.availablePlugins', function() {
-        expect(function() {
+        expect(() => {
           new EmberApp({
             project,
           });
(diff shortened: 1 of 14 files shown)


sangm:
@kratiahuja Fixed tests :)

tests/unit/broccoli/ember-app-test.js
```diff
@@ -884,6 +884,25 @@ describe('EmberApp', function() {
       expect(app._scriptOutputFiles['/assets/vendor.js']).to.contain('vendor/path/to/lib.js');
       expect(app._scriptOutputFiles['/assets/vendor.js']).to.contain('vendor/path/to/lib2.js');
     });
+
+    it('option.using throws exception given invalid inputs', function() {
+      // `using` is looped over if given, we should ensure this throws an exception with proper error message
+      expect(function() {
```
twokul: could we use arrow functions here?

kratiahuja:
@rwjblue I don't have write access yet. Can you please merge? Thank you! :)

kratiahuja:
Thanks @sangm !

rwjblue:
🎉🎊👌

PR #7303, opened 2017-08-21
Introduces a way to debug application/add-on trees

twokul:
fixes https://github.com/ember-cli/ember-cli/issues/7228.

Adds a simple way of dumping all of application trees via:

```bash
BROCCOLI_DEBUG="bundler:*" ember build
```

Mostly for debugging purposes. It also will give users better visibility into how ember-cli build pipeline works (which trees it operates on).

screen shot 2017-08-21 at 3 30 52 pm

### DIFF
--- a/docs/BUILD_PIPELINE_DEBUGGING.md
+++ b/docs/BUILD_PIPELINE_DEBUGGING.md
@@ -1,0 +1,33 @@
+# Build Pipeline Debugging
+
+**Note**: This is _only_ for debugging purposes and should not be relied upon.
+
+To gain more visibility into what broccoli trees Ember CLI operates on, you can run the following command:
+
+```sh
+BROCCOLI_DEBUG="bundler:*" ember build
+```
+
+Fish equivalent is:
+
+```sh
+env BROCCOLI_DEBUG="bundler:*" ember build
+```
+
+For a brand new ember application (`ember new your-app-name`), `DEBUG/` folder is going to be created with the following contents:
+
+```sh
+DEBUG/
+└── bundler:application-and-dependencies
+    ├── addon-tree-output
+    ├── tree-shake-test
+    └── vendor
+```
+
+`addon-tree-output` is the folder that contains all trees from Ember CLI add-ons;
+`your-app-name` is an application tree and `vendor` contains imported external assets.
+
+Ember CLI uses [`broccoli-debug`](https://github.com/broccolijs/broccoli-debug/) to generate debug output mentioned above.
+
+More tips can be found in our [PERF_GUIDE](https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md)
+under [DEBUG logging section](https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md#debug-logging).
--- a/lib/broccoli/bundler.js
+++ b/lib/broccoli/bundler.js
@@ -2,6 +2,7 @@
 
 const concat = require('broccoli-concat');
 const mergeTrees = require('./merge-trees');
+const BroccoliDebug = require('broccoli-debug');
 const processModulesOnly = require('./babel-process-modules-only');
 
 module.exports = class Bundler {
@@ -19,6 +20,7 @@ module.exports = class Bundler {
     this.appOutputPath = options.appOutputPath;
     this.vendorFilePath = options.vendorFilePath;
     this.isBabelAvailable = options.isBabelAvailable;
+    this._debugTree = BroccoliDebug.buildDebugCallback('bundler');
   }
 
   /*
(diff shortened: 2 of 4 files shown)


docs/BUILD_PIPILINE_DEBUGGING.md
```diff
@@ -0,0 +1,91 @@
+# Build Pipeline Debugging
```
rwjblue: typo in the title here (should be `PIPELINE` I think)

docs/BUILD_PIPILINE_DEBUGGING.md
```diff
@@ -0,0 +1,91 @@
...
+```sh
+env BROCCOLI_DEBUG="bundler:*" ember build
+```
+
+For a brand new ember application (`ember new your-app-name`), `DEBUG/` folder is going to be created with the following contents:
+
+```sh
+DEBUG/
+└── bundler:application-and-dependencies
+    ├── addon-tree-output
```
rwjblue: Not sure that we need to list all the contents here, perhaps just a `tree -L 2 DEBUG/` (2 levels) or something?

lib/broccoli/bundler.js
```diff
@@ -148,8 +149,9 @@ module.exports = class Bundler {
     options = options || {};
 
     let scriptOutputFiles = options.scriptOutputFiles;
+    let applicationAndDepsDebugTree = new BroccoliDebug(applicationAndDepsTree, 'bundler:application-and-dependencies');
```
rwjblue: We should avoid this style of `BroccoliDebug` instantiation as it cannot be "zero cost".  Instead, we should do:

```js
constructor() {
  this._debugTree = BroccoliDebug.buildDebugCallback('bundler');
}

bundleJs(_applicationAndDepsTree, options) {
  let applicationAndDepsTree = this._debugTree(_applicationAndDepsTree, 'application-and-dependencies');
  // ...snip...
}
```

If written in this way, when no environment variable is set there are no additional trees created (the callback is simply `(tree) => tree`), however if the environment variable is set it will return the wrapped tree.

rwjblue: Also, I think we should call this `bundle-js:input` (not `'application-and-dependencies'`)

lib/broccoli/bundler.js
```diff
@@ -168,7 +170,7 @@ module.exports = class Bundler {
       appJs = processModulesOnly(appJs, 'Babel: Modules for app/');
     }
 
-    let vendorFiles = this.getVendorFiles(applicationAndDepsTree, scriptOutputFiles);
+    let vendorFiles = this.getVendorFiles(applicationAndDepsDebugTree, scriptOutputFiles);
 
     return mergeTrees(vendorFiles.concat(appJs), {
```
rwjblue: This should be wrapped in a debug also:

```js
let combinedOutput = mergeTrees(....current thing....);
return this._debugTree(combinedOutput, 'bundle-js:output');
```

lib/broccoli/bundler.js
```diff
@@ -148,8 +150,9 @@ module.exports = class Bundler {
     options = options || {};
 
     let scriptOutputFiles = options.scriptOutputFiles;
+    let applicationAndDepsDebugTree = this._debugTree(applicationAndDepsTree, 'bundler:js:input');
```
rwjblue: The string passed to `BroccoliDebug.buildDebugCallback` is prefixed for all trees debugged by that particular `debugTree` callback. This will cause the debug dir to be called `DEBUG/bundler:bundler:js:input`.

rwjblue: Just need to make the second arg: `js:input`

lib/broccoli/bundler.js
```diff
@@ -168,10 +171,12 @@ module.exports = class Bundler {
...
 
-    let vendorFiles = this.getVendorFiles(applicationAndDepsTree, scriptOutputFiles);
+    let vendorFiles = this.getVendorFiles(applicationAndDepsDebugTree, scriptOutputFiles);
 
-    return mergeTrees(vendorFiles.concat(appJs), {
+    let combinedTree = mergeTrees(vendorFiles.concat(appJs), {
       annotation: 'TreeMerger (vendor & appJS)',
     });
+
+    return this._debugTree(combinedTree, 'bundler:js:output');
```
rwjblue: Should be `js:output`

sangm:
This is great for someone new to ember-cli to get introduced to the build process. Thanks @twokul

docs/BUILD_PIPELINE_DEBUGGING.md
```diff
@@ -0,0 +1,30 @@
+# Build Pipeline Debugging
```
kratiahuja: I think this should be cross linked in https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md

twokul: was there something in particular you wanted to see in the description? It's not exactly related to performance 🙂

kratiahuja: Oho there is a various tips and tricks section which contains some debugging info. In that section, adding this might be helpful.

twokul: done!

PR #7309, opened 2017-08-29
model/project: Freeze app config before caching it

Turbo87:
This should give at least slightly more meaningful errors and stacktraces when addons are trying to mutate the app config.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -209,6 +209,10 @@ class Project {
       let c = this.configCache.get(env);
       if (c === undefined) {
         c = this.configWithoutCache(env);
+
+        const deepFreeze = require('deep-freeze');
+        deepFreeze(c);
+
         this.configCache.set(env, c);
       }
       return c;
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "console-ui": "^1.0.2",
     "core-object": "^3.1.3",
     "dag-map": "^2.0.2",
+    "deep-freeze": "^0.0.1",
     "diff": "^3.2.0",
     "ember-cli-broccoli-sane-watcher": "^2.0.4",
     "ember-cli-is-package-missing": "^1.0.0",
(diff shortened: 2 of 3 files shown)


rwjblue:
OK, lets do it. Do you want to update this PR to do the first bit (disable caching) since it targets `beta` alread?

Turbo87:
nah, I'll just keep this open and rebase once `beta === 2.16` and open a new one to disable caching. should I disable or remove/revert it?

rwjblue:
I think either (whichever is easier) is fine...

Turbo87:
@rwjblue I've rebased this. As the code is protected by the feature flag it can already be merged and released in 2.15.

krisselden:
There was a better alternative. Still cache but clone the cached value on get, producing the value involves lots of I/O and a lodash deep copy preserves the existing semantics

Turbo87:
feel free to implement and test this proposal, since it's protected by a feature flag now we can still change it

krisselden:
There is no need to feature flag if it matches the existing observable behavior.  If the config is computed once, but cloned on this.project.config(env) there is no breaking change.  If we want to continue with the existing plan go for it.  I'm not going to bother with a PR behind a feature flag if we want to deprecate mutations and freeze, that is ultimately more efficient but the issue this was solving is computing it involves a lot of I/O if your repo has a lot of git refs, and addons were calling this method other than the one that just writes the config to meta.

Turbo87:
@krisselden FWIW feature flag and freezing are separate things. the feature flag is used for the whole config caching thing for now until we've made sure that nothing breaks this time. I'm not opposed to the idea of deep cloning instead of deep freezing, in fact I would prefer it. since the config is ultimately serialized as JSON I suppose we don't have to worry about having functions or other hard-to-clone data structures in there?

krisselden:
It is already required to be serializable and I don't prefer cloning unless the feature flag is on by default.  The cloning in my opinion makes the caching a non breaking change.

PR #7310, opened 2017-08-29
models/project: Hide app config caching behind CONFIG_CACHING feature flag

Turbo87:
as suggested in https://github.com/ember-cli/ember-cli/pull/7309#pullrequestreview-59242932


Turbo87:
@rwjblue the Node 8 test on Travis seems to fail for unrelated reasons 🤔

PR #7317, opened 2017-09-01
[INTERNAL] Introduce `broccoli-assembler`

twokul:
This is a non-breaking, backwards compatible change that gets us one step closer to a more composable, flexible build pipeline for Ember applications.

The main motivation behind this change is to make our build pipeline less naive (as of now it merely merges and concatenates input trees) as well as unlock that path forward for JS modules tree shaking and css/fonts optimizers.

This change separates two concerns: "how to build" and "build the final output" .

By introducing this separation, building the final output becomes a straightforward step: given all the inputs, merge them into one. It also allows us to introduce changes to the pipeline w/o changing too much code.

Separation of concerns is achieved via introducing the idea of "Assembler" and "Strategy". A strategy will describe a custom transform for input tree. An assembler will build the final output, given a list of transforms. For example, you could potentially create a `PerRouteAssetBundlingStrategy` to introduce per route splitting for Ember.js applications.

Changes made:
+ introduce a new dependency [`broccoli-assembler`](https://github.com/twokul/ember-cli-assembler)
+ introduce a concatenation strategy
+ refactor `bundler.js` code to use new concatenation strategy


package.json
```diff
@@ -41,6 +41,7 @@
     "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
     "bower-config": "^1.3.0",
     "bower-endpoint-parser": "0.2.2",
+    "broccoli-assembler": "twokul/broccoli-assembler",
```
rwjblue: Will need a release before we can merge here

stefanpenner:
I'd like to thorough review before this lands. I'll make time this weekend I promise <3.

package.json
twokul: absolutely! Maybe it would make sense to move it under ember-cli umbrella? Let me know.

twokul:
@stefanpenner I’d really appreciate that!

rwjblue:
Thanks for refactoring this @twokul. I'm happy to land this and iterate in repo while we figure out the best patterns and prepare for an RFC on the topic.  I feel that the assembler + strategy concept is superior to the prior revision (the bundler) and has a better and more flexible API.

👏

rwjblue:
@stefanpenner - This introduces no external API changes, results in exactly the same `dist/` output as we have had for many versions, and the trees being generated are very nearly identical.  I'm going to land this refactoring for now, but we can always address any specific concerns you have when you have time to discuss...

PR #7322, opened 2017-09-05
[INTERNAL] Update NPM to npm

kimroen:
The company and project is spelled `npm`, not `NPM`. These are mixed throughout this project. This updates the tests, readme wording, changelog, comments, blueprints and similar to reflect this.

I intentionally did not touch function names.

https://www.npmjs.com/policies/trademark

### DIFF
--- a/lib/tasks/npm-install.js
+++ b/lib/tasks/npm-install.js
@@ -27,11 +27,11 @@ class NpmInstallTask extends NpmTask {
   }
 
   formatStartMessage(packages) {
-    return `${this.useYarn ? 'Yarn' : 'NPM'}: Installing ${formatPackageList(packages)} ...`;
+    return `${this.useYarn ? 'Yarn' : 'npm'}: Installing ${formatPackageList(packages)} ...`;
   }
 
   formatCompleteMessage(packages) {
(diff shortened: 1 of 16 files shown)


lib/tasks/npm-install.js
```diff
@@ -27,11 +27,11 @@ class NpmInstallTask extends NpmTask {
   }
 
   formatStartMessage(packages) {
-    return `${this.useYarn ? 'Yarn' : 'NPM'}: Installing ${formatPackageList(packages)} ...`;
+    return `${this.useYarn ? 'Yarn' : 'npm'}: Installing ${formatPackageList(packages)} ...`;
```
kellyselden: This seems a little out of place, like someone might come along in the future and rename to "Npm" because it matches the rest.

kimroen: As far as I understand, `Npm` is not correct either.

This particular instance is actually the one I noticed and prompted me to fix it - here:

npm

kimroen: I guess it's kind of like saying that someone might rename "iPhone" to "Iphone" because it's next to "Android".

kellyselden: Good point. We will just have to be our own eslinters when looking out for this in the future!

rwjblue:
Thank you!

PR #7325, opened 2017-09-06
update ember-data version in beta

kellyselden:
since ember-data doesn't have a 2.16 beta yet, should we at least update to the latest version of the addon?


Turbo87:
sorry, I hadn't seen this PR. This has been fixed during the last beta release.

PR #7331, opened 2017-09-14
[EXPERIMENT] Standard ESM

twokul:
This is experimental commit based on the on of the conversations we had during Ember CLI core meeting. This change allows us to start using `import`/`export` across the codebase.

This change enables a smooth transition between Node and ES module formats and compensates for CJS and ESM incompatibility.

The tool used to enable this transition is [@std/esm](https://github.com/standard-things/esm).

`@std/esm` uses [reify](https://github.com/benjamn/reify). The `@std/esm` loader is intended for runtime use, so no build step or pre-publish step is required of the user. There is a 1 time transform cost per ESM module, after that the result is cached and loading is on par with CJS. The loader itself when published will be a small, ~40kb, zero dependency package.

Changes:

+ add `@std/esm` as a dependency
+ add `@std/esm` config to `package.json`
  + `"cjs": true` means we get to use `require` in files until we’re ready to convert to `import`
  + `"esm": "js”` means we don’t have to convert file extensions from `*.js` to `*.mjs`
+ repurpose `lib/cli/index.js` to use `@std/esm` loader
+ move `lib/cli/index.js` code to `lib/cli/main.js` and replaced `require`s with `import`s
+ remove `use strict;` from files as `@std/esm` adds it by default
+ add `overrides` to `.eslintrc.js` to hint different parse option
+ add a note on ESM support

Relevant links:

+ [ES Modules in Node Today!](https://medium.com/web-on-the-edge/es-modules-in-node-today-32cff914e4b)

Caveats:

_Usage of ‘use strict’;_. If we decide to go down the road of conversion, we should drop the usage of `use strict` not only in Ember CLI internally but in the blueprints as well (`ember-cli-build.js`, `config/environment.js`, `targets.js`).


bin/ember
```diff
@@ -1,42 +1,6 @@
 #!/usr/bin/env node
 'use strict';
 
-// This file is intentionally left compatible with older versions of
-// node for the following scenario:
-// * user runs node 0.12
-// * user is still running an older node 0.12 compatible ember-cli locally
-// * user unintentionally installs ember-cli latest globally
+const loader = require('@std/esm')(module);
```
rwjblue: We should move this out of `bin/ember` into `lib/cli/index.js`, and reset `bin/ember` to the prior state.  I don't see a reason to force this `bin/ember` incompatibility...

bin/main.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
+
+// This file is intentionally left compatible with older versions of
```
rwjblue: This comment isn't true any more...

.gitignore
```diff
@@ -7,3 +7,4 @@ npm-debug.log
 .deps-tmp
 /coverage
 *.pem
+.esm-cache
```
rwjblue: Should this also be added to `.npmignore`?

kanongil: No need. Npm already ignores files listed in `.gitignore`.

rwjblue: Hmm.  I don't think so? I just re-read through https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package and it seems that `.gitignore` is only used if you don't already have a `.npmignore`.  In our case, we do have an `.npmignore` so we will also need to update it.

kanongil: Wow, you are right. This `.ignore` stuff is quite awful. I'm sure there are tons of packages on the npm repo with fun secrets that have accentually been included.

kellyselden: You are very correct...

rwjblue:
@ember-cli/core - Any objections?

kellyselden:
LGTM. One final thought. Should we wait for the dust to settle just a little bit more? Is the rush to get this in worth it right now? or has all the debate out there settled at this point?

twokul:
@rwjblue r?

Turbo87:
@twokul have you experimented with using `require()` on ES6 modules from CJS modules?

twokul:
@Turbo87 the only caveat there would be to add `.default` to required module

```
```

rwjblue:
@twokul - Do you know if you can intermingle ES modules and CJS?  E.g. could we use imports and still use `module.exports = `?

twokul:
@rwjblue unfortunately, no. You would get `SyntaxError` like so

```bash
{ SyntaxError: Module '/Users/alexnavasardyan/code/oss/cli/ember-cli/lib/utilities/require-as-hash.js' does not provide an export named 'default'
    at null.<anonymous> (/Users/alexnavasardyan/code/oss/cli/ember-cli/lib/cli/main.js)
    at Object.<anonymous> (/Users/alexnavasardyan/code/oss/cli/ember-cli/lib/cli/main.js)
```

rwjblue:
So (sadly) I think our only option is to use ES modules only in files that are not imported by users (which may be difficult to determine). 

A middle step would be to start adding a `default` property to `module.exports` and deprecate consuming without `.default`...

jdalton:
Hi All, `@std/esm` has been updated a lot since this PR opened. You might try again with the `"@std/esm":"cjs"` option.

twokul:
@jdalton @rwjblue unfortunately, it still doesn't work with an updated version of `@std/esm`.

jdalton:
Please file an issue on `@std/esm` if you think it's a bug on it.

PR #7333, opened 2017-09-14
models/project: Use deep cloning instead of freezing the config

Turbo87:
addresses https://github.com/ember-cli/ember-cli/pull/7309#issuecomment-326519564

/cc @krisselden

### DIFF
--- a/lib/experiments/index.js
+++ b/lib/experiments/index.js
@@ -4,6 +4,7 @@ const getCallerFile = require('get-caller-file');
 const deprecate = require('../utilities/deprecate');
 
 let experiments = {
+  CONFIG_CACHING: symbol('config-caching'),
   MODULE_UNIFICATION: symbol('module-unification'),
   get INSTRUMENTATION() {
     deprecate(`\`experiments.INSTRUMENTATION\` is deprecated now that the feature has landed, use \`instrumentation\` method on your addon instead. Required from: \n${getCallerFile()}`, true);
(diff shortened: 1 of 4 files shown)


lolmaus:
@Turbo87 This PR breaks `ember-cli-dotenv`: https://github.com/fivetanley/ember-cli-dotenv/issues/30.

Specifically, the problem is that previously the app's `config/environment.js` function and the addon's `index.js` [config hook](https://github.com/fivetanley/ember-cli-dotenv/blob/master/index.js#L6-L54) were executed multiple times alternately.

Now both are executed only once, and `environment.js` goes first, so the `config` hook injects env vars too late.

Is this a regression or should `ember-cli-dotenv` use a different hook that fires earlier than the `environment.js` function? Which one?

lolmaus:
Hmm, even the `included` addon hook fires too late, after `config/environment.js` has already been established.

jamesdixon:
@Turbo87 any thoughts on this? This basically breaks the ability to load environment variables using dotenv. Using dotenv manually (see [comment](https://github.com/fivetanley/ember-cli-dotenv/issues/30#issuecomment-335412863)) works, but it then breaks the ability to use environment-specific dotenv files through `ember-cli-deploy` because it's configuration is overridden by the manual calls to dotenv.

rwjblue:
1) commenting on a closed issue/pr is very unlikely to get a response
2) please open a new issue
3) I think that addon is likely going to need to be restructured to work in this system
4) it is unlikely that we will revert this change given its impact and relatively small (so far) negative results

jamesdixon:
@rwjblue sorry about that. didn't want to create additional threads about the same thing, but i understand. thanks for the feedback

PR #7345, opened 2017-10-03
correct `rsvp` version in yarn.lock

bekzod:
https://github.com/ember-cli/ember-cli/pull/7340 in yarn.lock rsvp locked to fetch `4.6.1` when `4.0.2` needs to be fetched

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "promise-map-series": "^0.2.1",
     "quick-temp": "^0.1.8",
     "resolve": "^1.3.0",
-    "rsvp": "^4.0.2",
+    "rsvp": "^4.7.0",
     "sane": "^1.6.0",
     "semver": "^5.1.1",
     "silent-error": "^1.0.0",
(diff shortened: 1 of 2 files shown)


Turbo87:
@bekzod I don't understand. what's wrong with `4.6.1` being used?

bekzod:
@Turbo87 `4.6.1` is older version than `4.0.2`,  `4.6.1` was accidentally released :P

bekzod:
so if you use `yarn install` it is installing older version

Turbo87:
@bekzod keep in mind that all of our users will also install `4.6.1` by default so I think we should keep it that way for now until the situation is fixed by the rsvp.js maintainers

rwjblue:
I think we should just release 4.0.2 as 4.6.2 and keep incrementing upwards. I’ve fallen into the trap of getting 4.6.1 too (yarn seems to pick the highest non-prerelease not the `latest` tag). 

@stefanpenner - ^

rwjblue:
Latest and highest semver version now agree, lets update to using 4.7.0 as a minimum...

bekzod:
@Turbo87 updated to `4.7.0` now

Turbo87:
@bekzod @rwjblue thanks!

PR #7349, opened 2017-10-03
Limit allowed concurrency in CI environments.

rwjblue:
Ideally we could determine the actually allowed concurrency count, but I am not aware of a mechanism that works consistently here. If we could detect this, we should update [this line](https://github.com/stefanpenner/broccoli-persistent-filter/blob/57d8f901e955b5d08de72417263fd14fb1d53803/index.js#L92) to use the "right" value by default.

Happy to revert this if docker/kubernetes/xen/kvm experts can point us to a better solution than `require('os').cpus().length - 1` for default concurrency...

Fixes https://github.com/emberjs/ember.js/issues/15641

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -19,6 +19,9 @@ cache:
     - $HOME/.npm
 <% } %>
 env:
+  global:
+    # See https://git.io/vdao3 for details.
+    - JOBS=1
   # we recommend new addons test the current and previous LTS
   # as well as latest stable release (bonus points to beta/canary)
   - EMBER_TRY_SCENARIO=ember-lts-2.8
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -16,6 +16,11 @@ cache:
   directories:
     - $HOME/.npm
 <% } %>
+env:
+  global:
+    # See https://git.io/vdao3 for details.
+    - JOBS=1
+
 before_install:<% if (yarn) { %>
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH=$HOME/.yarn/bin:$PATH
--- a/docs/build-concurrency.md
+++ b/docs/build-concurrency.md
@@ -1,0 +1,15 @@
+# Build Concurrency
+
+In order to speed up your project's build time, Ember CLI has added a bit of
+concurrency throughout the build system. The exact number of parallel
+transpilation jobs that will be used can be customized via the `JOBS` process
+environment variable.
+
+The default value for `process.env.JOBS` is (max concurrency) - 1 (via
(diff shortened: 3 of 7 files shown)


Turbo87:
@rwjblue shouldn't we fix this in `broccoli-babel-transpiler` itself?

rwjblue:
After a *full 5 minutes* of googling @hjdivad pointed me to https://github.com/moby/moby/issues/20688#issuecomment-188923858, which we might be able to leverage in `broccoli-babel-transpiler`.

rwjblue:
@Turbo87 - Ya, probably, but we can only choose a default in `broccoli-persistent-filter` and I'm not sure that we can actually detect this perfectly yet.  The changes in this PR are safe and work right now, I'm happy to tweak as needed but I feel that this PR is roughly the safest path at the moment...

Turbo87:
> I feel that this PR is roughly the safest path at the moment...

can we set `JOBS=1` for the whole environment instead of patching it into each `ember test` call? is `ember-try` correctly passing that env var to `ember test`?

hjdivad:
> After a full 5 minutes of googling @hjdivad pointed me to moby/moby#20688 (comment), which we might be able to leverage in broccoli-persistent-filter.

The tl;dr here seems to be to parse `/sys/fs/cgroup/cpuset/cpuset.cpus` (see `man cpuset`)

Turbo87:
@hjdivad if I understand correctly the issue here is not incorrect CPU readings, but the container running out of memory. imagine a container that has 64 CPUs but only 512MB of memory, it would still fail even though we have lots of CPUs, so reading the correct amount of CPUs might not help that much either 🤷‍♂️

hjdivad:
> @hjdivad if I understand correctly the issue here is not incorrect CPU readings, but the container running out of memory. imagine a container that has 64 CPUs but only 512MB of memory, it would still fail even though we have lots of CPUs, so reading the correct amount of CPUs might not help that much either 🤷‍♂️

Yes although I expect these values to be related: 64 CPU available to a container but only 512MB memory is a pretty weird configuration.

rwjblue:
> can we set JOBS=1 for the whole environment instead of patching it into each ember test call? is ember-try correctly passing that env var to ember test?

Good point.  Updated.

Turbo87:
I'm wondering if we should add a comment to this to explain why this is necessary on TravisCI 🤔

rwjblue:
How about this:

```
    # the normal mechanism for detecting maximum number of concurrent jobs does not work
    # properly on CI (where the concurrency is limited to 2 but available CPU's is a much
    # larger number). this prevents the build from exhausting resources
```

rwjblue:
Updated with that comment inline.

rwjblue:
Updated instead to add a `docs/build-concurrency.md` file and link to that from the `.travis.yml`.  This allows us to tweak the messaging over time as needed, and avoids a lengthy comment in the configuration file...

PR #7351, opened 2017-10-04
Set entity name to current project name if none is provided

jeffrey008:
Example: 
`cd my-project`
`ember g whatever-blueprint` 

Then the entity name will be `my-project`.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -438,7 +438,7 @@ let Blueprint = CoreObject.extend({
         ' changes will be written.'));
     }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm.  I'm not sure we want to default this.  Can you explain why you want this behavior?

jeffrey008:
Because I had a blueprint in addon that will setup the new application. I want to blueprint to be as simple as possible so I want to call it with `ember g init-new-project` instead of `ember g init-new-project <new-name>`.

Issue #7327, opened 2017-09-06
You have not included `ember-cli-shims` in your project's `bower.json` or `package.json`.


johnunclesam:
I'm following this: https://emberjs.com/blog/2017/09/01/ember-2-15-released.html.

I upgraded my `package.json` with this:
```

    "ember-cli": "^2.16.0-beta.1",
    "ember-source": "^2.16.0-beta.1",
    "ember-cli-babel": "^6.8.2"
    --- removed "ember-cli-shims": "^1.1.0"
```

I receive this warning in `ember s`:

WARNING: You have not included `ember-cli-shims` in your project's `bower.json` or `package.json`. This only works if you provide an alternative yourself and unset `app.vendorFiles['app-shims.js']`.

kellyselden:
I believe you are missing the latest ember-cli-babel

johnunclesam:
Nope, I updated the first post. I have `"ember-cli-babel": "^6.8.2",`.

rwjblue:
We need to remove that warning.

rwjblue:
Was fixed by https://github.com/ember-cli/ember-cli/pull/7353.

PR #7353, opened 2017-10-04
remove ember-cli-shim warning

NullVoxPopuli:
ref: https://github.com/ember-cli/ember-cli/issues/7327


rwjblue:
We discussed this at the latest ember-cli meeting, and instead of removing directly we need to modify the conditional to only warn if the app does not depend on ember-cli-shims **and** it has at least one addon that depends on ember-cli-babel@5. 

This will remove the warning for folks that do not need it (ember-cli-babel@6.6+ does inline shim transpilation), and continue to be helpful for folks that do rely on the shims...

NullVoxPopuli:
@rwjblue thoughts on that?

rwjblue:
Ya, that's a great start! We now need to modify this to recursively traverse addons and check the same thing on each of them.

NullVoxPopuli:
ah, so that method doesn't already do that. gotchya

NullVoxPopuli:
is there someplace that already does a recursive dep lookup?

rwjblue:
I think something like this should work:

```js
function needsShims(subject) {
  let emberCLIBabelInstance = subject.findAddonByName('ember-cli-babel');
  if (emberCLIBabelInstance && semver.lt(emberCLIBabelInstance.pkg.version, '6')) {
    return true;
  }
  return subject.addons.some(needsShims);
}

//  ...snip...
let needsEmberCLIShims = needsShims(this.project);
if (!emberShims && !addonEmberCliShims && !bowerEmberCliShims && needsEmberCLIShims) {
  // current warning
}
```

NullVoxPopuli:
@rwjblue updated. :-)

rwjblue:
Looks like there are some CI failures (I haven’t reviewed the logs just yet), mind taking a look?

NullVoxPopuli:
CI passes!

package.json
```diff
@@ -39,6 +39,7 @@
   "dependencies": {
     "amd-name-resolver": "1.0.0",
     "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
+    "bower": "^1.8.2",
```
rwjblue: I think this must have been a left over from another change? We definitely don't want to add `bower` back as a dependency...

lib/broccoli/ember-app.js
```diff
@@ -328,6 +328,25 @@ class EmberApp {
 
   /**
     @private
+    @method _needsShims
+  */
+  _needsShims(subject) {
```
rwjblue: The extracted function is great, I think it would work a bit better as a separate stand alone function (e.g. not as a method on `EmberApp` itself), since we are using it unbound in the `addons.some(...)` bit just below.  As it stands, I'm not sure that more than two levels would work?

package.json
NullVoxPopuli: interesting.
must have been a bad rebase

lib/broccoli/ember-app.js
NullVoxPopuli: aight, I have an idea for a different abstraction that I think is more flexible.

NullVoxPopuli: @rwjblue updated!

NullVoxPopuli:
I'm not sure the error on node 7 is related to anything I did :-\

rwjblue:
OK, I updated this to target `beta`, rebased, and cleaned up and decided to inline the utility function into the `lib/broccoli/ember-app.js` file directly. 

I think this ready to go, @NullVoxPopuli mind reviewing one more time?

PR #7354, opened 2017-10-04
Pass given yarn cli option to NpmInstall.run

PieterJanVdb:
Fixes: #7355 

In `NpmTask.run` `this.useYarn` is overwritten by `options.useYarn` passed to the method when it's being called from `NpmInstallTask.run`.

However, when `NpmInstallTask.run` is being called from `AddonInstallTask.run`, `useYarn` is not passed as one of the options, meaning that in `NpmTask.findPackageManager`, `useYarn` will always be undefined and will always use Yarn if a lockfile is present.

This PR makes it so the option `useYarn` is passed to `NpmInstallTask.run`, with the value of either true (--yarn), false (--no-yarn) or undefined (not specified after which it will check for a yarn.lock file).

### DIFF
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -42,6 +42,7 @@ class AddonInstallTask extends Task {
       'save': commandOptions.save,
       'save-dev': !commandOptions.save,
       'save-exact': commandOptions.saveExact,
+      'useYarn': commandOptions.yarn,
     }).then(() => this.project.reloadAddons())
       .then(() => this.installBlueprint(blueprintInstall, packageNames, blueprintOptions))
       .finally(() => ui.stopProgress())
--- a/tests/unit/commands/install-test.js
+++ b/tests/unit/commands/install-test.js
@@ -106,6 +106,7 @@ describe('install command', function() {
           'save': false,
           'save-dev': true,
           'save-exact': false,
+          useYarn: undefined,
         }), { times: 1 });
       });
     });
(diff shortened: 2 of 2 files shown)


PieterJanVdb:
#7355

PieterJanVdb:
Hmm, the AppVeyor build seems to have failed on 1 single test on Node 7 which succeeds on the Travis CI build.

rwjblue:
I commented on the related issue. I'm unsure that we want to support `--yarn` and `--no-yarn` for all commands...

lennyburdette:
I opened a PR with the same change (and closed it when @Turbo87 pointed me here). Here's my motivation:

Currently, the NpmInstall task chooses Yarn over NPM based on the presence of the yarn.lock file in the Ember CLI app's project root. However, if the Ember CLI app is part of a Yarn workspace, the lockfile will be in a parent directory. In that case, [`hasYarnLock()`](https://github.com/lennyburdette/ember-cli/blob/master/lib/tasks/npm-task.js#L44) returns false and NpmInstall uses NPM.

I believe the most complete fix would be to change `hasYarnLock()` to traverse parent directories to find the lockfile, but that's a larger change. I would be happy with this bug fix as it enables me to add `"yarn": true` to our .ember-cli file and force yarn regardless of lockfile location.

lib/tasks/addon-install.js
```diff
@@ -42,6 +42,7 @@ class AddonInstallTask extends Task {
       'save': commandOptions.save,
       'save-dev': !commandOptions.save,
       'save-exact': commandOptions.saveExact,
+      'useYarn': commandOptions.yarn,
```
Turbo87: @PieterJanVdb we're explicitly passing `commandOptions.yarn` to the `NpmInstallTask` constructor above. I'm not quite sure why we need to pass it here too 🤔

lennyburdette: It's because `#run` is checking `options.useYarn` here: https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/npm-task.js#L139

Looks like we can just remove that line instead.

lennyburdette: Just tested it locally and it seems to work. Opened a quick PR here: https://github.com/ember-cli/ember-cli/pull/7481

Turbo87:
replaced by https://github.com/ember-cli/ember-cli/pull/7481

PR #7361, opened 2017-10-07
Update "ember-data" to v2.16.2

Turbo87:
/cc @ember-cli/core


rwjblue:
Looks like ember-data has errors in Node 4 (likely due to missing `'use strict'`).

Turbo87:
for future reference: v2.16.2 fixed the Node 4 issue that existed in the previous v2.16.x releases

PR #7372, opened 2017-10-10
[BUGFIX] Fix travis.yml in addon blueprint

simonihmig:
After running `ember-cli-update` in an addon and pushing, the Travis build was broken with a yaml parsing error: https://travis-ci.org/kaliber5/ember-window-mock/builds/286204403

The changes were introduced in #7349, but this does indeed not seem like valid yaml (`env` being used as object as well as array). Travis docs show this: https://docs.travis-ci.com/user/environment-variables/#Global-Variables

After adding the additional `matrix` key, this fixed my build...

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -22,14 +22,15 @@ env:
   global:
     # See https://git.io/vdao3 for details.
     - JOBS=1
-  # we recommend new addons test the current and previous LTS
-  # as well as latest stable release (bonus points to beta/canary)
-  - EMBER_TRY_SCENARIO=ember-lts-2.8
-  - EMBER_TRY_SCENARIO=ember-lts-2.12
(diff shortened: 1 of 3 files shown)


blueprints/addon/files/.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
   global:
     # See https://git.io/vdao3 for details.
     - JOBS=1
+  matrix:
   # we recommend new addons test the current and previous LTS
   # as well as latest stable release (bonus points to beta/canary)
   - EMBER_TRY_SCENARIO=ember-lts-2.8
```
rwjblue: I think this comment and the list item need to be indented an extra 2 spaces.

rwjblue:
Thank you! 

* Can you create a new issue around adding tests for validating the emitted `yml` files? 
* Can you target this PR against `release` branch (and rebase) so that we can get this out in a point release?

PR #7376, opened 2017-10-11
Bump ember-cli-babel and ember-cli-htmlbars-inline-precompile

rwwagner90:
Not technically necessary, but just bumping these to latest versions, to make my diffs more happy 😄


kellyselden:
We tend to not want the churn of updating to in range versions when we don't need to. How would this help you? It seems like it would have the opposite effect of adding more noise to your diffs.

rwwagner90:
@kellyselden it just puts it on the latest versions, and I generally try to keep mine on the latest, so keeping, for example `ember-cli-babel` at `^6.6.0` produces a diff for all my apps and addons because they are on `^6.8.2`. Certainly not required, since it does pull in the latest versions because of `^` but I feel we should still update it periodically when several version bumps have occurred, so it is closer to the actual version pulled in.

btecu:
Wasn't `^6.8.2` needed be able to drop `ember-cli-shims`?

kellyselden:
@rwwagner90 I assume you're using `ember init` which is resetting all your files? Try using [ember-cli-update](https://github.com/kellyselden/ember-cli-update), which will probably give you the diffs you want.

rwwagner90:
@kellyselden yes, I was using `ember init`. I tried ember-cli-update once and was confused on how to do diffs and such. I suppose I need to get used to it, but I am used to running `ember init`.

rwwagner90:
If @btecu is correct, and updating to `6.8.2` allows us to drop `ember-cli-shims` I can also remove that from the blueprint. Please advise on what change I should make to get this in or if we want to wait, we can close this.

rwwagner90:
Closing this since it doesn't seem to be a desired change and it's outdated on versions anyway.

Issue #7370, opened 2017-10-10
"found an unknown transformation name amd" when trying to import asset from addon


hakilebara:
I can import a standard anonymous AMD asset in `ember-cli-build.js`

```javascript
  app.import('node_modules/cropperjs/dist/cropper.js', {
    using: [
      { transformation: 'amd', as: 'cropperjs' }
    ]
  });
```
however the same import fails in the `included` hook of an addon or in-repo-addon
```javascript
/* eslint-env node */
'use strict';

module.exports = {
  name: 'beta-bug',

  isDevelopingAddon() {
    return true;
  },
  included(app) {
    this._super.included.apply(this, arguments);

    const srcFile = 'node_modules/cropperjs/dist/cropper.js';
    app.import(srcFile, {
      using: [
        { transformation: 'amd', as: 'cropperjs' }
      ]
    });
```
This is the error message I get:
```javascript
while import node_modules/cropperjs/dist/cropper.js: found an unknown transformation name amd. Available transformNames are:
Error: while import node_modules/cropperjs/dist/cropper.js: found an unknown transformation name amd. Available transformNames are:
    at options.using.forEach.entry (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:1710:19)
    at Array.forEach (<anonymous>)
    at EmberApp._import (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:1701:23)
    at EmberApp.import (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:1674:10)
    at Class.included (/Users/kaos/Projects/ember-beta/lib/beta-bug/index.js:14:15)
    at Class.superWrapper [as included] (/Users/kaos/Projects/ember-beta/node_modules/core-object/lib/assign-properties.js:34:20)
    at project.addons.project.addons.filter.addon (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:541:17)
    at Array.filter (<anonymous>)
    at EmberApp._notifyAddonIncluded (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:538:47)
    at new EmberApp (/Users/kaos/Projects/ember-beta/node_modules/ember-cli/lib/broccoli/ember-app.js:151:10)
```

[...]

PR #7377, opened 2017-10-11
[BUGFIX] Invoke transform registeration before included hook is called.

kratiahuja:
Fixes https://github.com/ember-cli/ember-cli/issues/7370 . Added tests too.

I think we will need to do another release of ember-cli 2.16.x. This is blocking a few folks to upgrade to ember-cli 2.16.

cc: @rwjblue


rwjblue:
I just retargeted and rebased.

kratiahuja:
Thank you @rwjblue !

Issue #7368, opened 2017-10-10
Error: ENOENT: no such file or directory, lstat '.../ember-test-selectors/app'


Turbo87:
I'm getting the following error when [upgrading `ember-test-selectors` to Ember CLI `v2.16.0`](https://github.com/simplabs/ember-test-selectors/pull/146):

```
Build failed.
The Broccoli Plugin: [broccoli-persistent-filter:EslintValidationFilter] failed with:
Error: ENOENT: no such file or directory, lstat '/Users/tbieniek/Code/ember-test-selectors/app'
    at Object.fs.lstatSync (fs.js:946:11)
    at symlink (/Users/tbieniek/Code/ember-test-selectors/node_modules/symlink-or-copy/index.js:79:26)
    at symlinkOrCopySync (/Users/tbieniek/Code/ember-test-selectors/node_modules/symlink-or-copy/index.js:65:5)
    at /Users/tbieniek/Code/ember-test-selectors/node_modules/broccoli-plugin/read_compat.js:86:11
    at tryCatch (/Users/tbieniek/Code/ember-test-selectors/node_modules/rsvp/dist/rsvp.js:525:12)
    at invokeCallback (/Users/tbieniek/Code/ember-test-selectors/node_modules/rsvp/dist/rsvp.js:538:13)
    at publish (/Users/tbieniek/Code/ember-test-selectors/node_modules/rsvp/dist/rsvp.js:508:7)
    at flush (/Users/tbieniek/Code/ember-test-selectors/node_modules/rsvp/dist/rsvp.js:2415:5)
    at _combinedTickCallback (internal/process/next_tick.js:95:7)
    at process._tickCallback (internal/process/next_tick.js:161:9)

The broccoli plugin was instantiated at:
    at EslintValidationFilter.Plugin (/Users/tbieniek/Code/ember-test-selectors/node_modules/broccoli-plugin/index.js:7:31)
    at EslintValidationFilter.Filter [as constructor] (/Users/tbieniek/Code/ember-test-selectors/node_modules/broccoli-persistent-filter/index.js:64:10)
    at new EslintValidationFilter (/Users/tbieniek/Code/ember-test-selectors/node_modules/broccoli-lint-eslint/lib/index.js:58:10)
    at Function.EslintValidationFilter.create (/Users/tbieniek/Code/ember-test-selectors/node_modules/broccoli-lint-eslint/lib/index.js:215:16)
[...]

PR #7378, opened 2017-10-11
Ensure test-support and addon-test-support are linted.

### LINKED ISSUE #7035: test-support and addon-test-support don't lint files

rwjblue:
* Rebase + cleanup of #7037 (add linting for test-support and addon-test-support)
* Add basic harness for unit testing broccoli build "things" for an addon
* Add tests for regression reported in #7368
* Ensure that the module paths when linting match actual paths on disk (instead of being in "fake" paths)

Fixes #7035
Closes #7037

/cc @trentmwillis

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1109,23 +1109,43 @@ let addonProto = {
 
     let addonPath = this._treePathFor('addon');
     if (existsSync(addonPath)) {
-      let addonJs = this.addonJsFiles(addonPath);
+      let addonJs = new Funnel(this.addonJsFiles(addonPath), {
+        srcDir: this.moduleName(),
+        destDir: 'addon',
+        allowEmpty: true,
+      });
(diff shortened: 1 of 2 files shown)


trentmwillis:
Thanks for finishing this for me! 😄

Issue #3664, opened 2015-03-26
blueprint generation should take into account .editorconfig


stefanpenner:

stefanpenner:
@trabus is this structure related, or ...

trabus:
No, I'd say it's more blueprint related. AST transform after generating, most likely.

stefanpenner:
> AST transform after generating, most likely.

ah, yes. Good call.

trabus:
cc @abuiles

efx:
+1
 Right now, generate blueprints seems to default to space based indentation while I am working in a project that requires tab based indentation. Using the .editorconfig would fix this problem quite cleanly.

stefanpenner:
> Using the .editorconfig would fix this problem quite cleanly.

yes correct, it may also be easily supported if we added post generation transforms.

efx:
how do 'post generation transforms' differ from the 'afterInstall' hook ?
I'm guessing the transforms would be easier to implement.

stefanpenner:
because we want this to run during the templating phase of any blueprint based generation.

vforge:
+1 for this

huntie:
+1 Would love to see this

ctcpip:
+1

stefanpenner:
I'll try to work on an RFC for this

ro0gr:
Just an idea. 
Also it would be great also to take into account eslint fix.
Looks like eslint has a [programmatic api](https://github.com/eslint/eslint/blob/master/lib/util/source-code-fixer.js) for that.

nathanhammond:
Note: this is incredibly low priority for us on the CLI team. In order for anything to come of this it _must_ have a champion. Further, since we originally wrote the config and blueprints they represent the style we use. 😜

We're in no way opposed to make this happen, but we'll likely only partner with somebody in the community who cares deeply. (Tabs vs. Spaces debate can begin anew.)

oskarrough:
@nathanhammond thanks for clarifying. I'd love to help out with this but unfortunately I'm not sure where to start. If someone has time to team up on this, let me know.

j127:
Specifying the indentation via a `.editorconfig` file is a great idea for a standard feature in code generators in general. Code styles are personal things that should be left up to the programmers, not frameworks. Not having this feature in a framework that uses a code generator as part of the standard workflow adds a lot of hassle, because the generated code may not be compatible with the rest of the project out of the box.

[...]

PR #7380, opened 2017-10-12
 BREAKING: always append EOL in insertIntoFile

kellyselden:
Right now you may or may not get an EOL insterted based on whether you are inserting in the middle or appending to the end. This was confusing me. This change unifies so that you always get the EOL.

This is one solution, the other is we could take EOL out completely and always leave it up to the consumer.

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -69,6 +69,7 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
   if (options.force) { insert = true; }
 
   if (insert) {
+    contentsToInsert += EOL;
     if (insertBehavior === 'end') {
       contentsToWrite += contentsToInsert;
     } else {
@@ -80,7 +81,7 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
         if (insertBehavior === 'after') { insertIndex += contentMarker.length; }
 
         contentsToWrite = contentsToWrite.slice(0, insertIndex) +
-          contentsToInsert + EOL +
+          contentsToInsert +
           contentsToWrite.slice(insertIndex);
       }
     }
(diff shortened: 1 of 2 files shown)


ro0gr:
@kellyselden what do you think about integrating `editorconfig --fix` option like [here](https://github.com/ember-cli/blprnt/pull/29) into blueprints?

I think it can solve such kind of issues forever.
Also people who have different `editorconfig` settings could immediately take benefit from this feature.

original idea #3664

ro0gr:
sorry, I misread your description.

Now I have the question regarding this pr. What's the reason to always insert a new line regardless of the insert position? 
In my opinion it's quite opinionated.

kellyselden:
@ro0gr https://github.com/ember-cli/ember-cli-blueprint-test-helpers/blob/master/blueprints/ember-cli-blueprint-test-helpers/index.js#L13

Your file goes from a file with a final newline to one without. But if you changed it to use `{ after: `ember-cli-build.js${EOL}` }`, you would get the free newline you expected.

ro0gr:
Now I see, thanks for explanation.

So my original question becomes valid to me again. Isn't it better to handle this kind of issues on post insert phase? 

I suspect there are still some code formatting edge cases in blueprints which we can handle ad-hoc ofcourse. However it seems it's better if it's handled automatically.

ro0gr:
Also what if some blueprints from the ecosystem already append an `EOL`? Will we add an additional one?

kellyselden:
> Also what if some blueprints from the ecosystem already append an EOL? Will we add an additional one?

That's an unfortunate side effect of this change.

kellyselden:
Now that I think of it, maybe we should take out all EOL appending. Trying to assume is just leading to hassle in my ember-cli-blueprint-test-helpers PRs.

PR #7381, opened 2017-10-12
[BUGFIX] Changes default Chrome remote debugging port.

### LINKED ISSUE #7366: testem blueprint should use --remote-debugging-port=0 for headless chrome
The `--remote-debugging-port` is required by headless chrome or it will exit even though testem does not use it, it still must be set.  If it is set to `--remote-debugging-port=0` it will open an ephemeral port.  With it hardcoded to 9222 it will conflict if tests are run in parallel.

Oreoz:
Fixes #7366.

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -14,7 +14,7 @@ module.exports = {
       args: [
         '--disable-gpu',
         '--headless',
-        '--remote-debugging-port=9222',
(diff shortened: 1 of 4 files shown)


rwjblue:
Awesome, thank you @Oreoz!

Caltor:
How do we know which port to connect to for remote debugging if using ephemeral port?

PR #7382, opened 2017-10-12
add option to not create file

kellyselden:
Useful when you are dealing with app vs addon inconsistencies, for example the .npmignore file. You want to append if exists, otherwise ignore and don't create.


kellyselden:
@rwjblue beta worthy?

PR #7384, opened 2017-10-12
support regex in insertIntoFile

kellyselden:
Right now if you use `{ before: '/tests' + EOL }`, you are missing those Windows users that choose to check in Unix line endings. Now you can use regex to catch both line endings.


kellyselden:
I got lazy and deferred to Travis for test results lol

kellyselden:
BTW, good thing we don't actually have a Travis on our team, he/she would be very confused.

rwjblue:
lol, indeed!

kellyselden:
@rwjblue Final approval? Also, should I target beta?

PR #7385, opened 2017-10-13
remove node 7 testing

kellyselden:
Do y'all feel like it's time?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ language: node_js
 node_js:
   - "4"
   - "6"
-  - "7"
   - "8"
 
 branches:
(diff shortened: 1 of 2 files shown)


rwjblue:
Seems good to me. Doing this on master would mean we drop Node 7 support in 2.18, which seems fine to me...

Turbo87:
Is there a Node 9 already that we could replace it with?

rwjblue:
Not yet, but based on the schedule [here](https://github.com/nodejs/Release#release-schedule1) there should be a Node 9 out sometime this month (essentially when Node 8 becomes LTS, Node 9 should be released).

PR #7387, opened 2017-10-13
move from implicit towards explicit testdouble dependency

kellyselden:
This came up in https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/138#issuecomment-336514382.

This hidden dependency is fishy. I think we should move towards the consumer passing it in, so the dependency is more clear.

### DIFF
--- a/tests/helpers/mock-process.js
+++ b/tests/helpers/mock-process.js
@@ -1,6 +1,5 @@
 'use strict';
 
-let td = require('testdouble');
 let EventEmitter = require('events');
 
 module.exports = class FakeProcess extends EventEmitter {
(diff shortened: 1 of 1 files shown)


Turbo87:
I'm not sure I agree with this. Technically the stuff in our tests folder shouldn't be used by anyone. The fact that we're using it for the blueprint tests is a little unfortunate. I wonder if it's reasonable to extract the test helper in a separate package.

rwjblue:
FWIW, a number of our test helpers should really be public API (MockProject, MockUI, etc). I definitely agree that they could be extracted as stand alone packages though.

ro0gr:
I've just noticed it [fails on npm@2](https://travis-ci.org/san650/ember-cli-page-object/jobs/289246356#L831) even with [testdouble listed as an ember-cli-blueprint-test-helpers dependency](https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/124).

Turbo87:
this seems obsolete now that https://github.com/ember-cli/ember-cli/pull/7395 is merged

PR #7388, opened 2017-10-14
BREAKING: remove EOL assumptions from insertIntoFile

kellyselden:
Closes #7380.

This is probably the better fix. It's too confusing guessing whether you get a free EOL or not when you are using `before`, `after`, or neither.

Not sure what we do about the breaking nature of this though.

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -2,7 +2,6 @@
 
 const fs = require('fs-extra');
 const existsSync = require('exists-sync');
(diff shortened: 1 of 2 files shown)


Turbo87:
> the breaking nature of this

that indeed seems to be an issue... we should probably discuss at the next call?

kellyselden:
The decision was to move all these helpers into it's own lib, so addon's can opt in to new breaking features instead of trying to support all versions of ember-cli.

Turbo87:
closing due to inactivity

PR #7395, opened 2017-10-19
Make "testdouble" dependency optional in MockProcess class

ro0gr:
Seems like `mock-process` caused some [inconveniences ](https://github.com/ember-cli/ember-cli/pull/7387) due to its `testdouble` dependency.

Actually we need `testdouble` only in `will-interrupt-process` unit tesst. We have nothing to do with `testdouble` in any other [scenarios](https://github.com/ember-cli/ember-cli/blob/master/tests/helpers/ember.js#L69).

Also previously I thought it was a good idea to implement `MockProcess.exit()` as a `testdouble.function()`. Unfortunately [it wasn't](https://github.com/ember-cli/ember-cli/pull/7285). This approach isn't able to replicate a real `process.exit()` behavior and leads to dishonest tests when we test cli/cli for example. 
That is the reason why I changed `MockProcess.exit()` to throw an exception.

Known packages who was forced to install `testdouble` due to ember-cli@2.16 upgrade:
 - [ember-cli-blueprint-test-helpers](https://github.com/ember-cli/ember-cli-blueprint-test-helpers)
 - [ember-cli-page-object](https://github.com/san650/ember-cli-page-object/pull/333/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R83)
 - ...

### DIFF
--- a/tests/helpers/mock-process.js
+++ b/tests/helpers/mock-process.js
@@ -1,29 +1,23 @@
 'use strict';
 
-let td = require('testdouble');
 let EventEmitter = require('events');
 
-module.exports = class FakeProcess extends EventEmitter {
+module.exports = class MockProcess extends EventEmitter {
   constructor(options) {
     super();
 
     options = options || {};
 
+    this.platform = 'MockOS';
+
     const stdin = Object.assign(new EventEmitter(), {
       isRaw: process.stdin.isRaw,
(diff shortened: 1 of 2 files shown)


Turbo87:
thanks @ro0gr!

Issue #7335, opened 2017-09-20
Yarn workspaces


bsvipas:
Hi, it doesn't work correctly with Yarn workspaces... Here is the issues: https://github.com/yarnpkg/yarn/issues/4503 (same as https://github.com/yarnpkg/yarn/issues/4081). All other apps works fine, they search for `node_modules` in root dir instead of current dir if they don't find it, but Ember just search it in current dir...

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
Ember: 2.15.1
NPM: 5.4.2
Yarn: 1.0.2
```

bsvipas:
This issue is related with `ember-cli-dependency-checker`. I fixed it and right now I need to polish it and I will try to make PR.

ef4:
Yes, ember-cli itself has a baked-in very bad assumption: the Project model has a `nodeModulesPath` property containing a single directory name. It's used both internally by ember-cli and by ember-cli-dependency-checker.

It's a bad assumption because node's module resolution algorithm doesn't work that way. There's no such thing as "the" node_modules directory. You can have many of them, and yarn workspaces uses this to great effect.

Fixing this will simplify ember-cli, because it has a bunch of code for trying to outguess node that can just be ripped out. The correct solution is to just use [resolve](https://www.npmjs.com/package/resolve) to look up modules relative to the project root.

lennyburdette:
@ef4 have you looked into yarn.lock file resolution as well? Some parts of Ember CLI assume that yarn.lock must exist in the Ember project root, but in a workspace the file will exist in the workspace root.

[Yarn's algorithm](https://github.com/yarnpkg/yarn/blob/master/src/config.js#L612) for finding the workspace root is somewhat straight-forward. Do you think it makes sense to port that into Ember CLI as a utility function for these three places?

[tasks/npm-task.js](https://github.com/ember-cli/ember-cli/blob/446f87bb63c8f9725920cfd9788396dcab14a911/lib/tasks/npm-task.js#L44)
[commands/init.js](https://github.com/ember-cli/ember-cli/blob/eef7317152576ee47d38fb6bbc3edd73b37c4ca4/lib/commands/init.js#L64)
[models/command.js](https://github.com/ember-cli/ember-cli/blob/d562bd36a653ae182ddc7fb1fa71a5fd83f26e0a/lib/models/command.js#L293)

cc @rwjblue

stefanpenner:
[...]

PR #7401, opened 2017-10-25
Resolve node modules correctly

ef4:
ember-cli attempts to provides its own module resolution semantics which are only a subset of node's. These resolution semantics are complex and we spend many lines of code implementing them, all for no benefit over node's native resolution algorithm.

As a side effect of this behavior, we cannot take advantage of nice things like [yarn workspaces](https://github.com/ember-cli/ember-cli/issues/7335).

This PR refactors ember-cli to use complete node modules resolution (as reimplemented in userspace by the [resolve](http://npmjs.com/package/resolve) package. As far as I can tell, I have preserved all existing behavior, including the (dubious and ripe for deprecation) EMBER_NODE_PATH environment variable.

The entire `node-modules-path` package can be dropped as a dependency once we decide to stop offering the (private-but-used-by-addons) `project.nodeModulesPath` property.

I think addons should be told to use `project.require` (already public) or `project.resolveSync` (would need to be marked public) wherever they were previously using `project.nodeModulesPath`. It's just fundamentally not correct to try to encode the modules location as a single path.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -12,8 +12,6 @@ const logger = heimdallLogger('ember-cli:addon');
 const treeCacheLogger = heimdallLogger('ember-cli:addon:tree-cache');
 const cacheKeyLogger = heimdallLogger('ember-cli:addon:cache-key-for-tree');
 
-const nodeModulesPath = require('node-modules-path');
-
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const preprocessJs = p.preprocessJs;
 const preprocessCss = p.preprocessCss;
@@ -207,7 +205,7 @@ let addonProto = {
     if (!this.root) {
       throw new Error('Addon classes must be instantiated with the `root` property');
     }
-    this.nodeModulesPath = nodeModulesPath(this.root);
+    this._nodeModulesPath = null;
 
     this.treePaths = {
       app: 'app',
@@ -1669,6 +1667,27 @@ Addon.prototype[BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS] = function() {
     return {};
   }
 };
+
+Object.defineProperty(Addon.prototype, 'nodeModulesPath', {
+  // This exists only for backward compat, because some addons use
+  // it. It was never marked public and we can totally nuke it, but we
+  // can be polite and give them a warning for a while.
+  get() {
+    if (!this._nodeModulesPath) {
+      let nodeModulesPath = require('node-modules-path');
+      let value = nodeModulesPath(this.root);
+      this._nodeModulesPath = { value };
+
+      let stack = new Error().stack;
+      stack = stack.split('\n')[2];
+      stack = stack.replace('    at ', '  ');
+
+      this.ui.writeDeprecateLine(`An addon is trying to access addon.nodeModulesPath. This is not a reliable way to discover npm modules. Instead, consider using require("resolve").sync(something, { basedir: addon.root }). Accessed from: ${stack.toString()}`);
+    }
+    return this._nodeModulesPath.value;
+  },
+});
+
 
 /**
   Returns the absolute path for a given addon
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -11,7 +11,6 @@ const fs = require('fs-extra');
 const existsSync = require('exists-sync');
 const _ = require('ember-cli-lodash-subset');
 const logger = require('heimdalljs-logger')('ember-cli:project');
-const nodeModulesPath = require('node-modules-path');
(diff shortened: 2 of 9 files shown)


lib/models/project.js
```diff
@@ -105,23 +104,44 @@ class Project {
...
+  // resolve the first dependency or devDependency we find.
+  hasDependencies() {
+    if (this.cli.disableDependencyChecker) {
+      // Blueprint tests set this flag.
+      return true;
+    }
+    for (let depName in this.dependencies()) {
+      try {
+        this.resolveSync(depName);
+        return true;
```
ro0gr: Looks like it will return true on the first found package. Is it intentional or should we check all the deps?

ro0gr: Ah sorry, missed you comment few lines above

lib/models/addon.js
```diff
@@ -1670,6 +1668,22 @@ Addon.prototype[BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS] = function() {
...
+Object.defineProperty(Addon.prototype, 'nodeModulesPath', {
+  // This exists only for backward compat, because some addons use
+  // it. It was never marked public and we can totally nuke it, but we
+  // can be polite and give them a warning for a while.
+  get() {
+    if (!this._nodeModulesPath) {
+      let nodeModulesPath = require('node-modules-path');
+      let value = nodeModulesPath(this.root);
+      this._nodeModulesPath = { value };
+      this.ui.writeDeprecateLine('An addon is trying to access addon.nodeModulesPath. This is not a reliable way to discover npm modules. Instead, consider using require("resolve").sync(something, { basedir: addon.root })');
```
rwjblue: We should include the stack here (or add an extra `debug` level) so folks have a chance at tracking this down

rwjblue: We could also open issues for the addons we can see:

https://emberobserver.com/code-search?codeQuery=nodeModulesPath

lib/models/project.js
```diff
@@ -300,20 +320,6 @@ class Project {
...
 
   /**
-    Resolves the absolute path to a file.
-
-    @private
-    @method resolve
-    @param  {String} file File to resolve
-    @return {String}      Absolute path to file
-   */
-  resolve(file) {
```
rwjblue: Thoughts on adding a private API deprecation here also?  Might not matter though because I didn't find any usages of `project.resolve` in emberobserver...

lib/models/project.js
```diff
@@ -323,7 +329,7 @@ class Project {
    */
   resolveSync(file) {
     return resolve.sync(file, {
-      basedir: this.root,
+      basedir: process.env.EMBER_NODE_PATH || this.root,
```
rwjblue: I don't grok what this environment variable is about, can you explain?

rwjblue: This is only used in one addon (per https://emberobserver.com/code-search?codeQuery=EMBER_NODE_PATH) and that looks like a copy-paste of ember-cli's own tests?

ef4: I would be totally happy to eliminate this feature, I just can't tell who's depending on it.  I think it's more of an app feature than an addon feature.

It was added here: https://github.com/ember-cli/ember-cli/pull/3771

I think it was a roundabout way of doing what this PR is also doing: allow people to put their node_modules in the other places that you would expect to work based on node behavior.

lib/models/project.js
ef4: I looked too, I couldn't find anything. I vote for leaving it out and you can assign the future issue to me when something blows up. :-)

lib/models/addon.js
ef4: I added a stack-based message to identify the culprit.

lib/models/project.js
rwjblue: Sounds good to me!

lib/models/project.js
rwjblue: Yeah, agreed.  Would you mind triggering a deprecation if we detect that `process.env.EMBER_NODE_PATH` is set?

rwjblue: Lets take this as a separate issue (and not block this PR on it)...

rwjblue:
Thank you @ef4!

ro0gr:
Today I've converted our solution to monorepo. It works great with current master. 
Thanks @ef4!

ef4:
That's great! I'm surprised it worked actually, because there are some standard addons that still need to be updated like ember-cli-dependency-checker.

MiguelMadero:
This is working great for us, with some minor tweaks. PR(s) coming

PR #7342, opened 2017-10-02
Updating testem.js for the app blueprint

scalvert:
Updates `app` blueprint's `testem.js` file to enable Chrome to run headless in CI, but non-headless otherwise.

PR #7412, opened 2017-11-02
Make dev tests run headless

morhook:
On a prior version, the tests run headless on dev machine (as also on CI). I liked that one better, not sure if it was the intention to take this functionality out

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -18,5 +18,14 @@ module.exports = {
         '--window-size=1440,900'
       ]
     },
+    Chrome: {
+      mode: 'dev',
+      args: [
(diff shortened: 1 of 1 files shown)


rwjblue:
Yep, the change was intentional (made in https://github.com/ember-cli/ember-cli/pull/7342). I think its totally reasonable for folks to have preferences here (please by all means keep your original setting in `testem.js`), but for an "out of the box" experience I think having a browser that you can interact with is better.

Thanks for taking the time to submit the PR, but I'm going to close as I believe we are happy with the existing setup.

morhook:
Thanks for responding so fast and explaining @rwjblue !! Sounds fair to me :)

PR #7413, opened 2017-11-02
Allows users to `disableDependencyChecker` from `.ember-cli` options.

MiguelMadero:
Right now we need to pass this option via the constructor, 
which is used for tests, but as a user I don't have a way to disable this. 
There are reasons to disable this outside of tests: 
* The DependencyChecker doesn't work today with yarn workspaces
* We may be using a different way to check dependencies like yarn --check-integrity

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -126,7 +126,7 @@ module.exports = function(options) {
     analytics: leek,
     testing: options.testing,
     name: options.cli ? options.cli.name : 'ember',
-    disableDependencyChecker: options.disableDependencyChecker,
+    disableDependencyChecker: config.options.disableDependencyChecker || options.disableDependencyChecker,
     root: options.cli ? options.cli.root : path.resolve(__dirname, '..', '..'),
     npmPackage: options.cli ? options.cli.npmPackage : 'ember-cli',
     initInstrumentation,


rwjblue:
You should either remove the addon or disable it via the SKIP_DEPENDENCY_CHECKER environment variable. Most likely the former if you are trying to always skip it...

rwjblue:
Both of the bullets mentioned here infer that you just don’t need the addon. Removing it (just like any other addon) is a perfectly viable option and seems much more inline with what your actual intentions are here...

MiguelMadero:
That sounds like a good option. But I'm not sure what addon you're referring to. 

I don't have ember-cli-dependency-checker installed. Not sure if that's the addon you're referring to. The problem seems to be on Project#hasDependency https://github.com/ember-cli/ember-cli/blob/mmadero/disableDependencyChecker-support/lib/models/project.js#130

MiguelMadero:
Looks like there is a separate bug introduced on #7401 I'll leave this open since it might make sense on its own

MiguelMadero:
I fixed it on https://github.com/ember-cli/ember-cli/pull/7414

MiguelMadero:
Hi Rob, 

My main use case is addressed now by #7414. There may still be a use case for this flag. What do you think? If there is I can document it and merge this if not, I'm happy to just close the PR.

MiguelMadero:
@rwjblue ping!

Turbo87:
sounds like this is mostly resolved now :)

MiguelMadero:
👍

PR #7415, opened 2017-11-03
Adds the ability to specify `.ember-cli` options at the root of the r…

MiguelMadero:
…epository.

Today we only check at the projectRoot and home (~/). 
With this change, the precedence is projectRoot, repositoryRoot and home. 
This is mostly important when we want to maintain consistent options in a monorepo.

cc: @rwjblue @ef4 this is what I mentioned on #ember-cli earlier today.

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -63,6 +63,24 @@ function configureLogger(env) {
   }
 }
 
+function getConfig (options, Project) {
+  const Yam = options.Yam || require('yam');
+  const projectRoot = Project.getProjectRoot();
+  const repositoryRoot = Project.getRepositoryRoot();
+  const useRepositoryRootAsSecondary = projectRoot !== repositoryRoot;
+  const projectConfig = new Yam('ember-cli', {
+    primary: projectRoot,
+    secondary: useRepositoryRootAsSecondary ? repositoryRoot : undefined
+  });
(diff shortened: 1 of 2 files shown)


ef4:
This is beyond the scope of what I intended by fixing node modules resolution. In that case it makes sense to look upward because that is the standard node behavior that all node devs must learn. In this case we are inventing wholly new API. 

I'm not strongly opposed, but I am skeptical.

MiguelMadero:
I understand this is a bit beyond. The work you did unlock the critical support for monorepos. When more people start using it, things like this might become an extension of that support. I agree this is new behavior and it might need some further discussion. The intention of opening the PR was to move the discussion off of slack and base it on a concrete implementation. 

There are many other tools that follow this behaviour like eslint (eslintrc and eslintignore), babelrc, yarnrc, prettier.config.js (just recently) and editorconfig are all common examples. tsconfig and tslint.json were exceptions last time I checked, but AFAIR they're both planning to support it as well. tslint.json supports using `extends` so at least there is a way to centralize the configuration.

MiguelMadero:
@rwjblue what do you think?

MiguelMadero:
@rwjblue ping!

MiguelMadero:
This is too old. Closing!

PR #7417, opened 2017-11-03
Issue warning for Node 7.

rwjblue:
Removing `7` from `.travis.yml` and `appveyor.yml` will trigger a warning for users still using Node 7.

We can drop support in `engines` once this deprecation has been absorbed throughout the community.


kellyselden:
Should be noted this is already in master, we are just pushing it out quicker.

PR #7432, opened 2017-11-10
Avoid publishing massive temp folder leaked by ember-try

ef4:
If anything interrupts an ember-try run this folder will be left behind, and it contains all your node_modules. It's easy to accidentally publish to npm. This happened to liquid-fire 0.29.0, which contains 153MB of uselessness.

I think this is appropriate in the base addon blueprint because ember-try is also in the blueprint. The alternative is to add a blueprint to ember-try that would insert this line, but that is more error prone and easier to drop during an ember-cli upgrade.

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -14,3 +14,4 @@
 bower.json
 ember-cli-build.js
 testem.js
+.node_modules.ember-try


kellyselden:
@rwjblue The reason we both thought it was in there already is because it was only in the gitignore, not npmignore. I will go audit that file for anything else missing.

kategengler:
👍  on adding to npmignore, but for documentation's sake: 

In the scenario when these are left behind, it is also likely that `node_modules`, `package.json` and/or `bower.json` are dirty from the ember-try run. Publishing in such a case is dubious. The command `ember try:reset` should set everything back to right.

PR #7438, opened 2017-11-11
fix(blueprint): skip uninstall if no matching package is installed

### LINKED ISSUE #7425: removePackagesFromProject in a yarn project throws if the packages aren't installed
When an addon calls `removePackagesFromProject` in a yarn project, ember-cli filters packages that aren't in the project from removal.

This means if an addon tries to remove a package that isn't installed, the projects list equals `[]`.
When passing it to yarn, it will generate `yarn remove --non-interactive` which throws an error:

`Not enough arguments, expected at least 1.`

One could probably fix this issue by not running the task if the list of packages is empty (somewhere around https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L960).

If the proposed fix is fine, I could open a PR to fix it.

p.s. I'm unsure if this also affects npm

---

```
ember-cli: 2.16.2
http_parser: 2.7.1
node: 8.8.1
v8: 6.1.534.42
uv: 1.15.0
zlib: 1.2.11
ares: 1.13.0
modules: 57
nghttp2: 1.25.0
openssl: 1.0.2l
icu: 59.1
unicode: 9.0
cldr: 31.0.1
tz: 2017b
os: linux x64
5.5.1
1.3.2
```

makepanic:
This PR changes the removePackagesFromProject behavior to resolve instead of running the uninstall task.

fixes #7425

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -957,6 +957,11 @@ let Blueprint = CoreObject.extend({
       }
     }
 
+    if (packageNames.length === 0) {
+      this._writeStatusToUI(chalk.yellow, 'remove', 'Skipping uninstall because no matching package is installed.');
+      return Promise.resolve();
+    }
+
     this._writeStatusToUI(chalk.green, installText, packageNames.join(', '));
 
     return task.run({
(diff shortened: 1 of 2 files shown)


rwjblue:
Looks good to me, thank you @makepanic!

PR #7440, opened 2017-11-11
add osx testing

kellyselden:
This is mainly for discussion. It is possible to test in OSX. The jobs tend to take a while to be allocated, so it would slow down our CI (as well as adding the additional job(s)).

Is this worth it? Do we ever regress in mac while Windows and Linux are fine?

cc @ember-cli/core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+
 sudo: false
 dist: trusty
 
@@ -38,12 +41,17 @@ matrix:
   - node_js: "6"
     env: TEST_COMMAND=test-all
 
-  # This is the reworking of the tests we run against the most-recent LTS.
   include:
+  # This is the reworking of the tests we run against the most-recent LTS.
   - node_js: "6"
     env: TEST_COMMAND=test-all:cover
   - node_js: "6"
     env: DISABLE_EMBER_CLI_FEATURES=true; TEST_COMMAND=test-all
+
+  - os: osx
+
+  allow_failures:
+  - os: osx
 
 before_install:
   # Once we no longer support Node.js 0.12 we can drop this entire section.


twokul:
> Is this worth it? Do we ever regress in mac while Windows and Linux are fine?

I would consider adding it because:

+ [majority](https://insights.stackoverflow.com/survey/2016#technology-desktop-operating-system) of developers use mac
+ new features like AFS (Apple File System) in [High Sierra](https://www.apple.com/macos/high-sierra/) might be useful to test before releasing to users (I've been running Ember CLI on High Sierra for a while now and haven't seen major regressions).

> The jobs tend to take a while to be allocated, so it would slow down our CI

do we know what the slow down would be?

kellyselden:
> do we know what the slow down would be?

Travis CI doesn't have that many divvied up for OSS project use I believe. You get placed in a queue during peak business hours.

Maybe they would donate priority status for us?

Turbo87:
I'd be okay with occasionally running CI on OSX, but doing it for every PR seems unnecessary if it increases the time until CI is green by such a large amount. The differences between OSX and Linux are rather small so I don't think it's currently worth it.

kellyselden:
We decided that blocking PRs until an OSX box is allocated is not worth it. We are going to try running this only on master instead.

kellyselden:
I couldn't figure out how to exclude PRs. It's just running allow failure now.

rwjblue:
Ya, I think this is the one configuration that I really dislike 😜 (allowed failure)....

kellyselden:
Yea, at this point we can keep or decline. I will leave it up to someone else to make the call.

Turbo87:
I don't see much practical value in this at this point. I'd vote for closing this for now and if OSX testing on Travis gets faster at some point we can revisit.

rwjblue:
Agreed. Thanks for working on this though @kellyselden!

PR #7441, opened 2017-11-13
Fix "travis_retry: command not found" error

twokul:
This issue prevents code coverage tools from working properly.

It turns out that `travis_retry` function is actually [not available on build machines](https://twitter.com/travisci/status/499196774712901632).

To work around that, we can copy `travis_retry` function from [travis-ci/travis-build repository](https://github.com/travis-ci/travis-build/blob/73f74a94957f73eb54dc821f80c0c85ad8f8aab7/lib/travis/build/script/templates/header.sh#L168-L187).

For reference, a [tweet by Travis CI](https://twitter.com/plexus/status/499194992632811520) saying that you should copy the `travis_retry` code from
[travis-ci/travis-build repository](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/templates/header.sh).

### DIFF
--- a/.travis/codecoverage.sh
+++ b/.travis/codecoverage.sh
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 
-# For now we will use `npm` for globals.
-# Can switch to `yarn` once we no longer support Node.js 0.12:
-# - travis_retry yarn global add coveralls codeclimate-test-reporter
-travis_retry npm install -g coveralls codeclimate-test-reporter
-cat coverage/lcov.info | codeclimate-test-reporter
-cat coverage/lcov.info | coveralls
+# fail script immediately on any errors in external commands
+set -e
+
+# Use `yarn` to install globals
+yarn global add coveralls codeclimate-test-reporter
+
+cat ./coverage/lcov.info | coveralls
+codeclimate-test-reporter < ./coverage/lcov.info
--- a/docs/CODE_COVERAGE.md
+++ b/docs/CODE_COVERAGE.md
@@ -1,0 +1,13 @@
+# Ember CLI Code Coverage & Analysis
+
+We are using [Coveralls](https://coveralls.io/) to track our code coverage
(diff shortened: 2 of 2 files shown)


.travis/codecoverage.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# fail script immediately on any errors in external commands
+set -e
+
+source travis_retry.sh
```
Turbo87: where does that script come from?

twokul: my understanding is that it should be available based on [this](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/templates/header.sh).

twokul: nvm, going to have to [copy the function](https://twitter.com/travisci/status/499196774712901632)

Turbo87: do we even need `travis_retry` here? it makes sense to use it for the initial `npm install` step, but if coverage reporting  deps fail it's not sooo bad, right?

.travis.yml
```diff
@@ -6,10 +6,7 @@ addons:
 
 language: node_js
 node_js:
-  - "4"
   - "6"
-  - "8"
```
rwjblue: Ummm, no?

.travis/codecoverage.sh
```diff
@@ -1,8 +1,32 @@
...
+    echo -e "\n${ANSI_RED}The command \"$@\" failed 3 times.${ANSI_RESET}\n" >&2
+  }
+
+  return $result
+}
+
 # For now we will use `npm` for globals.
 # Can switch to `yarn` once we no longer support Node.js 0.12:
 # - travis_retry yarn global add coveralls codeclimate-test-reporter
 travis_retry npm install -g coveralls codeclimate-test-reporter
```
rwjblue: I wonder if we can try without automatic retrying?  I don't think we have failed `npm install` / `yarn install` invocations very much lately...

PR #7443, opened 2017-11-15
Use `overrides` for a single `.eslintrc.js`.

rwjblue:
Update the `.eslintrc.js` used by apps and addons, to avoid needing separate "sprinkles" of eslint configuration throughout the projects codebase.

Compatible with `eslint@4.1.0`.

Additional reading:

* https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns
* https://github.com/eslint/eslint/pull/8081

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -9,5 +9,34 @@ module.exports = {
     browser: true
   },
   rules: {
-  }
+  },
+  overrides: [
+    // node files
+    {
+      files: [
+        'index.js',
+        'testem.js',
+        'ember-cli-build.js',
+        'config/**/*.js',
+        'tests/dummy/config/**/*.js'
+      ],
+      parserOptions: {
+        sourceType: 'script',
+        ecmaVersion: 2015
+      },
+      env: {
+        browser: false,
+        node: true
+      }
+    },
+
+    // test files
+    {
+      files: ['tests/**/*.js'],
+      excludedFiles: ['tests/dummy/**/*.js'],
+      env: {
+        embertest: true
+      }
+    }
+  ]
 };
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 module.exports = {
   <% if (yarn) { %>useYarn: true,
   <% } %>scenarios: [
--- a/blueprints/addon/files/addon-config/environment.js
+++ b/blueprints/addon/files/addon-config/environment.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 'use strict';
 
 module.exports = function(/* environment, appConfig */) {
--- a/blueprints/addon/files/ember-cli-build.js
+++ b/blueprints/addon/files/ember-cli-build.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 'use strict';
 
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
--- a/blueprints/addon/files/index.js
+++ b/blueprints/addon/files/index.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 'use strict';
 
 module.exports = {
--- a/blueprints/app/files/config/environment.js
+++ b/blueprints/app/files/config/environment.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 'use strict';
 
 module.exports = function(environment) {
--- a/blueprints/app/files/config/targets.js
+++ b/blueprints/app/files/config/targets.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 module.exports = {
   browsers: [
     'ie 9',
--- a/blueprints/app/files/ember-cli-build.js
+++ b/blueprints/app/files/ember-cli-build.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 'use strict';
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 module.exports = {
   test_page: 'tests/index.html?hidepassed',
   disable_watching: true,
--- a/blueprints/app/files/tests/.eslintrc.js
+++ b/blueprints/app/files/tests/.eslintrc.js
@@ -1,5 +1,0 @@
-module.exports = {
-  env: {
-    embertest: true
-  }
-};
--- a/tests/fixtures/addon/npm/config/ember-try.js
+++ b/tests/fixtures/addon/npm/config/ember-try.js
@@ -1,4 +1,3 @@
-/* eslint-env node */
 module.exports = {
   scenarios: [
     {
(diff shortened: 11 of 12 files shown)


rwjblue:
I will follow up and make sure that `ember-cli-eslint`'s blueprint is in sync once we settle on this...

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
   root: true,
   parserOptions: {
     ecmaVersion: 2017,
-    sourceType: 'module'
+    sourceType: 'module',
```
kellyselden: seems strange to only have trailing commas in this file. I vote for removal.

rwjblue: 😭 but ok 😄

Turbo87: I agree, we should add them to all files...

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
...
   rules: {
-  }
+  },
+  overrides: [
+    // node files
+    {
+      files: [
+        './ember-cli-build.js',
+        './testem.js',
+        './index.js',
```
Turbo87: should we use `./*.js` instead?

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
...
+        'config/**/*.js',
+      ],
+      parserOptions: {
+        sourceType: 'script',
+        ecmaVersion: 2015,
+      },
+      env: {
+        browser: false,
```
Turbo87: I think this has to be `browser: null` or something like that. iirc `true/false` only defines whether the global can be overwritten or something like that

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
...
+        node: true,
+      }
+    },
+
+    // test files
+    {
+      files: ['tests/**/*.js'],
+        env: {
```
Turbo87: indentation

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
...
-  }
+  },
+  overrides: [
+    // node files
+    {
+      files: [
+        './ember-cli-build.js',
+        './testem.js',
+        './index.js',
+        'config/**/*.js',
```
Turbo87: what about `tests/dummy/config/**/*.js`?

blueprints/app/files/.eslintrc.js
```diff
@@ -2,12 +2,39 @@ module.exports = {
...
+      },
+      env: {
+        browser: false,
+        node: true,
+      }
+    },
+
+    // test files
+    {
+      files: ['tests/**/*.js'],
```
Turbo87: can we exclude `tests/dummy/`?

it's okay if we can't since we haven't done it before either, but would be nice though

blueprints/app/files/.eslintrc.js
rwjblue: Ya, agreed. Though I’m happy to do this separately.

blueprints/app/files/.eslintrc.js
rwjblue: Interesting. I hadn’t thought of that. Should work nicely.

blueprints/app/files/.eslintrc.js
rwjblue: Really!?!?! Weird! I’ll look it up. I have used this on a few other projects, will test and update as needed.

blueprints/app/files/.eslintrc.js
rwjblue: 👍 - good idea

blueprints/app/files/.eslintrc.js
Turbo87: I just realized I might be confusing this with the stuff in the `globals` object, but we should verify just in case.

rwjblue: Ya, I was looking through the docs on configuring and I think you are thinking of the globals stuff.  For easy cross referencing:

* https://eslint.org/docs/user-guide/configuring#specifying-environments <- Docs on these `env` flags (no discussion of clobbering)
* https://eslint.org/docs/user-guide/configuring#specifying-globals <- Docs on `globals` (includes discussion on `true` / `false` flag to allow clobbering)

kellyselden:
I would like to hold off merging until our meeting. I'm not 100% in favor of this. I like getting rid of the first line sprinkles of node files, but I'm not convinced that removing our tests/.eslintrc.js is a good idea. Everybody's documentation and blueprint injection touch that file (globals and test helpers), and it seems to me that it is in it's correct place.

rwjblue:
> I would like to hold off merging until our meeting.

Seems good, happy to discuss...

> I like getting rid of the first line sprinkles of node files, but I'm not convinced that removing our tests/.eslintrc.js is a good idea. 

This is where we disagree. The current system of having many `.eslintrc` files all over makes it very difficult to understand what the current ruleset is at any given point (a number of times I've resorted to using `eslint --print-config some/file/path.js` just to figure out what the heck is running). Another issue is that (other than simple things like `env`) its not possible by adding a `.eslintrc` file to handle directories that contain _mixed_ rulesets .

> Everybody's documentation and blueprint injection touch that file (globals and test helpers), and it seems to me that it is in it's correct place.

I think this is actually a separate concern from the others. We should choose the right set of defaults, and recommend them while also providing time and assistance to addons that may need to make modifications. Modifying a single file (`.eslintrc.js`) in general, is easier than having to modify many potential locations...

kellyselden:
We decided in the core meeting that this is the way forward. With module unification, test files will be strewn about a project, so wildcard paths inside a single eslint config is the future (TM).

blueprints/app/files/.eslintrc.js
rwjblue: Yep, added and confirmed.  Works nicely!

rwjblue:
Rebased and addressed all outstanding comments...

blueprints/app/files/.eslintrc.js
rwjblue: Sadly, this doesn't work.  The main issue is that in the implementation (https://github.com/eslint/eslint/pull/8081) when the `files` and `excludedFiles` are passed in to `minimatch` ([here](https://github.com/eslint/eslint/blob/8e1a095f4cc3b371a0a424b66e13ec77734193a2/lib/config/config-ops.js#L368-L382)) they also pass `{ matchBase: true }` as options. 

From the `minimatch` README about `matchBase`:

> If set, then patterns without slashes will be matched against the basename of the path if it contains slashes. For example, a?b would match the path /xyz/123/acb, but not /xyz/acb/123.

**tl;dr** Due to the usage of `matchBase` it is not possible to use globs to match files in the project root. If we use `*.js` it matches _all_ files in the repo, if we use `./*.js` it matches no files, if we use `/*.js` it throws an error....

😭 😭 😭

Turbo87:
@rwjblue this shouldn't block, but can you verify the behavior for in-repo-addons?

rwjblue:
@Turbo87 - Good point, I will do some testing.  I _believe_ that we might need to add `lib/*/index.js` to the "node" config...

PR #7447, opened 2017-11-18
Remove ember-cli-legacy-blueprints.

rwjblue:
ember-cli has supported consuming `ember-source` for quite a number of versions now (since 2.12), the `ember-cli-legacy-blueprints` addon is only supporting usage of blueprints with versions of Ember prior to 2.12.

This commit removes `ember-cli-legacy-blueprints` from a default new project. If a given project still needs this addon, it should add it as its own dependency. Once added to the projects own dependencies the legacy blueprints will be available exactly the same as today.

---

Note: I do not consider this a breaking change. No functionality is lost for users following the normal blueprint updates...


kellyselden:
If we wanted to be extra nice, we could show a console warning when we don't detect a "route" blueprint. Then remove the warning with ember-cli 3.

Turbo87:
@rwjblue I would prefer to target this for the 3.0 release. If we release it in a minor we would need to make sure to have at least a deprecation period as Kelly suggested but since we're short on time until 3.0 we might as well do it then.

rwjblue:
Seems good to me. This just means the PR needs to wait a few days (master will target 3.0 after the upcoming release)...

Turbo87:
@rwjblue rebased this PR on top of `master` to restart CI

rwjblue:
Need to update a few test fixtures which are not including ember-source but are testing things like `ember g component`.

PR #7450, opened 2017-11-20
Add prettier and eslint-plugin-ember to linting config.

rwjblue:
Hot on the heels of updating the linting configuration to use a single file, this adds `prettier` and `eslint-plugin-ember` to a new app out of the box.

`eslint-plugin-ember@5` will be released soon, and includes a number of changes to the default recommended config. Specifically, the recommended set follows with `eslint`'s own strategy of only including "saftey related" rules in the recommended rule set. As such, all "style related" rules are no longer "on by default".  This definitely does not infer there is anything _wrong_ with those rules (and we may even add some of them into the default app config in the future), but we would like to bring `eslint-plugin-ember` into ember-cli gradually.

`prettier` has taken the JavaScript ecosystem by storm! It has quickly become the most common way to acheive "standard code style" without being nit-picky with linting rules. The configuration that is landed here integrates `prettier` into `eslint` so that `eslint` properly identifies things that are not "pretty" and a quick `eslint --fix` will auto-correct them. This also means that any editors with built-in `eslint --fix` support already support `prettier`. 🎉

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -37,6 +37,7 @@
     "ember-source": "~2.17.0-beta.2<% if (welcome) { %>",
     "ember-welcome-page": "^3.0.0<% } %>",
     "eslint-config-prettier": "^2.7.0",
+    "eslint-plugin-ember": "^5.0.0",
     "eslint-plugin-prettier": "^2.3.1",
     "loader.js": "^4.2.3",
     "prettier": "^1.8.2"
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -36,7 +36,10 @@
     "ember-resolver": "^4.0.0",
     "ember-source": "~2.17.0-beta.2<% if (welcome) { %>",
     "ember-welcome-page": "^3.0.0<% } %>",
-    "loader.js": "^4.2.3"
+    "eslint-config-prettier": "^2.7.0",
+    "eslint-plugin-prettier": "^2.3.1",
+    "loader.js": "^4.2.3",
+    "prettier": "^1.8.2"
   },
   "engines": {
     "node": "^4.5 || 6.* || >= 7.*"
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -5,71 +5,71 @@ module.exports = {
       name: 'ember-lts-2.12',
       npm: {
         devDependencies: {
-          'ember-source': '~2.12.0'
-        }
-      }
+          'ember-source': '~2.12.0',
+        },
+      },
     },
     {
       name: 'ember-lts-2.16',
       npm: {
         devDependencies: {
-          'ember-source': '~2.16.0'
-        }
-      }
+          'ember-source': '~2.16.0',
+        },
+      },
     },
     {
       name: 'ember-release',
       bower: {
         dependencies: {
-          'ember': 'components/ember#release'
+          ember: 'components/ember#release',
         },
         resolutions: {
-          'ember': 'release'
-        }
+          ember: 'release',
+        },
       },
       npm: {
         devDependencies: {
-          'ember-source': null
-        }
-      }
+          'ember-source': null,
+        },
+      },
     },
     {
       name: 'ember-beta',
       bower: {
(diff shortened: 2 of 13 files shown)


blueprints/app/files/package.json
```diff
@@ -36,7 +36,11 @@
     "ember-resolver": "^4.0.0",
     "ember-source": "~2.17.0-beta.2<% if (welcome) { %>",
     "ember-welcome-page": "^3.0.0<% } %>",
-    "loader.js": "^4.2.3"
+    "eslint-config-prettier": "^2.7.0",
+    "eslint-plugin-ember": "^5.0.0-beta.2",
```
rwjblue: 5.0.0 is out, I'll update to that shortly...

alexlafroscia:
A few thoughts: 

1. Prettier and Ember together make a ton of sense. Doubling down on community convention feels very “Ember” to me
2. Might want to think about updating the ESLint config in the `tests` directory that is generated by default. I’ve has issues with it not being marked as `root: true` when using the fixer on test files. The solve was to mark it as root and then explicitly extend from the config in the parent directory. 
3. Does it feel contradictory to turn off style-related things in the recommended config and then turn on Prettier, which is purely style-related? I love making it easier to get started with Prettier but am unsure about aggressively making decisions about code style for the community

rwjblue:
> Prettier and Ember together make a ton of sense. Doubling down on community convention feels very “Ember” to me

😄 🐹 ❤️ 

> Might want to think about updating the ESLint config in the tests directory that is generated by default. I’ve has issues with it not being marked as root: true when using the fixer on test files. The solve was to mark it as root and then explicitly extend from the config in the parent directory.

As of https://github.com/ember-cli/ember-cli/pull/7443 we only have a single eslint config (leveraging the newish `overrides` + globs), which should completely remove this issue. I also tested using both `neovim` + `ALE` (configured to run `eslint` as a fixer) and `emacs` + `flycheck`, before the changes in https://github.com/ember-cli/ember-cli/pull/7443 I had the same issue you mention here, after those changes things work well without modification.

> Does it feel contradictory to turn off style-related things in the recommended config and then turn on Prettier, which is purely style-related? I love making it easier to get started with Prettier but am unsure about aggressively making decisions about code style for the community

Thanks for bringing this up! I totally understand this position (and somewhat agree), however I still believe that the PR as it stands today is a good path forward. The primary difference here is the ability to `auto-fix`. With the `prettier` configuration that is added in this PR the rules do not seem "nit-picky" (even though they **totally are**), this is because a simple `eslint --fix` is all that is needed to clean things up. It essentially allows me to write JS how I "want" to (and TBH, I'm a tad sloppy 😉) and not worry about the exact formatting while editing.  This is _extremely_ empowering IMHO.

IMHO, the generally agreed upon stylistic rules (like `order-in-*`) in `eslint-plugin-ember` are *awesome*! However, to enable by default I think we will need more robust fixer support (this is being worked on in https://github.com/ember-cli/eslint-plugin-ember/pull/172).

kratiahuja:
Nvm I can see that is done using eslint —fix

Turbo87:
yep, although we could consider adding a `lint` (and/or `lint:fix`) script to `package.json` to make it easier to invoke ESLint. but that seems like a topic for a separate PR.

alexlafroscia:
Thanks for addressing my concerns, @rwjblue! I didn't know about the `overrides` thing -- I'm glad it makes that issue moot. I also see your point about auto-fixing being the difference between enabling Prettier by default and the other style-based rules. Hopefully those more Ember-specific rules can land fixer support too! I'll keep an eye on that PR.

BTW: I love the idea of adding `lint` and `lint:fix` by default. I always set up ESLint scripts that way, and it would be awesome to introduce a community standard around that too.

Standardize all the things!

rwjblue:
> BTW: I love the idea of adding lint and lint:fix by default. I always set up ESLint scripts that way, and it would be awesome to introduce a community standard around that too.

Agreed. Definitely a good follow up to this...

rwjblue:
Unfortunately, @kellyselden didn't have a chance to review this PR before I merged it and has some reservations.

@kellyselden - I am sorry for being over-eager and landing this too quickly.  I submitted #7453 to revert.

PR #7451, opened 2017-11-21
Add `eslint-plugin-node` for addons.

rwjblue:
Add `eslint-plugin-node` and its default rules to node-land files. This plugin allows us to fix the following common issues:

* Accidentially requiring files that are not included in `package.json`'s `files` array (or are specifically excluded via either `.npmignore` or `.gitignore`).
* Enforces all node land code to conform to the project `package.json`'s `engines` config. For example, when `engines` allows Node 4 linting would fail if someone used `let` or `const` without `'use strict';`.

These two gotchas are _super_ common for addons to hit, and this will save quite a few "quick-fix patch release" scenarios for folks going forward.

See [eslint-plugin-node's README](https://github.com/mysticatea/eslint-plugin-node#readme) for complete rule details.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -62,6 +62,9 @@ module.exports = {
     // add `ember-disable-prototype-extensions` to addons by default
     contents.devDependencies['ember-disable-prototype-extensions'] = '^1.1.2';
 
+    // add `eslint-plugin-node` to addons by default
+    contents.devDependencies['eslint-plugin-node'] = '^5.2.1';
+
     // use `ember-try` as test script in addons by default
     contents.scripts.test = 'ember try:each';
 
@@ -124,6 +127,7 @@ module.exports = {
       year: date.getFullYear(),
       yarn: options.yarn,
       welcome: options.welcome,
+      blueprint: 'addon',
     };
   },
 
--- a/blueprints/app/index.js
+++ b/blueprints/app/index.js
@@ -27,6 +27,7 @@ module.exports = {
       emberCLIVersion: require('../../package').version,
       yarn: options.yarn,
       welcome: options.welcome,
+      blueprint: 'app',
     };
   },
 };
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -406,6 +406,7 @@ describe('Acceptance: ember new', function() {
         'src/ui/routes/application/template.hbs',
         '.travis.yml',
         'README.md',
+        '.eslintrc.js',
       ].forEach(filePath => {
         expect(file(filePath))
           .to.equal(file(path.join(__dirname, '../fixtures', fixturePath, filePath)));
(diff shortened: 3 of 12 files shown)


blueprints/app/files/.eslintrc.js
```diff
@@ -33,13 +33,21 @@ module.exports = {
       env: {
         browser: false,
         node: true,
-      }
+      },<% if (isAddon) { %>
+      plugins: ['node'],
+      rules: Object.assign(
```
Turbo87: can we not use `extends` here?

rwjblue: Sadly no. You cannot use extends inside of overrides yet. 

This is being tracked in https://github.com/eslint/eslint/issues/8813.

rwjblue:
The latest round of changes are:

* Remove templating from `.eslintrc.js` and `package.json`. These issues definitely plague app developers also and this sort of linting is still useful.  We can bring templating back in a future PR if we want to do more customization around app vs addon `.eslintrc.js`. 
* Move unrelated test file changes into https://github.com/ember-cli/ember-cli/pull/7457 (to avoid confusion)

blueprints/app/files/.eslintrc.js
```diff
@@ -27,7 +27,15 @@ module.exports = {
       env: {
         browser: false,
         node: true
-      }
+      },
+      plugins: ['node'],
+      rules: Object.assign(
```
kellyselden: Is there a reason you didn't use `"extends": ["plugin:node/recommended"],`?

rwjblue: @Turbo87 asked the same question in https://github.com/ember-cli/ember-cli/pull/7451#discussion_r152213361 just above, answer is that sadly no. You cannot use extends inside of overrides yet.

This is being tracked in eslint/eslint#8813.

kellyselden: oh shoot. sorry I was on mobile and missed that.

kellyselden: I was gonna say add a comment to that issue to explain the code, but I read through the issue and doesn't even look like they want to do it. So a comment probably wouldn't make sense.

Turbo87:
@rwjblue I just realized that even if we added this plugin we wouldn't actually run ESLint on these files by default yet, so maybe we should get that sorted out first before adding a plugin that is only relevant for as of now unlinted files :thin

blueprints/app/files/.eslintrc.js
Turbo87:
```js
rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
  // add your custom rules and overrides for node files here
}),
```

can we reformat this to this?

rwjblue:
Ya, sounds good

rwjblue:
Just rebased, and updated things a bit.

* Added templating back to `.eslintrc.js` such that only addons get `eslint-plugin-node`
* Added explicit fixture style tests for the new `.eslintrc.js` templating in all configurations (`app`, `module-unification-app`, `addon`, with `yarn`, with `npm`, etc)

Should be good to go (assuming CI agrees 😝 )...

blueprints/app/files/.eslintrc.js
rwjblue: FWIW, looks like eslint _did_ add support for `extends` in overrides (landed in https://github.com/eslint/eslint/pull/11554 which is included in eslint@6.0.0+). We should work to remove this "bizarreness" now....

PR #7452, opened 2017-11-21
Add `lint:js` scripts.

rwjblue:
Add `lint` and `lint:fix` scripts, and add `lint` script to default test run for apps and addons.

This addresses an annoying hole in the default setup where the various "node land" files are not linted as part of CI.


twokul:
wondering if we should add a section about these new commands to `README.md` in blueprint

blueprints/app/files/package.json
```diff
@@ -12,6 +12,8 @@
   "repository": "",
   "scripts": {
     "build": "ember build",
+    "lint": "eslint *.js addon app config lib node-tests server tests",
+    "lint:fix": "eslint --fix *.js addon app config lib node-tests server tests",
```
kellyselden: Can this be `npm run lint -- --fix`? Or would yarn hate that? Maybe we can sub in yarn here too.

rwjblue: I generally find `npm run lint -- --fix` to be very hard to read (due to repeated `--` with different meaning)...

Are you mostly referring to removing the duplication between `lint` and `lint:fix`?

kellyselden: yep the duplication.

blueprints/addon/files/.travis.yml
```diff
@@ -52,3 +52,4 @@ script:
   # Usually, it's ok to finish the test scenario without reverting
   #  to the addon's original dependency state, skipping "cleanup".
   - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO --skip-cleanup
+  - <% if (yarn) { %>yarn<% } else { %>npm<% } %> run lint
```
Turbo87: yarn doesn't need `run`. `yarn lint` is sufficient.

blueprints/addon/files/README.md
```diff
@@ -19,6 +19,11 @@ This README outlines the details of collaborating on this Ember addon.
 * `ember test`
 * `ember test --server`
 
+### Linting
+
+* `<% if (yarn) { %>yarn<% } else { %>npm<% } %> run lint`
+* `<% if (yarn) { %>yarn<% } else { %>npm<% } %> run lint:fix`
```
Turbo87: see above

blueprints/app/files/.travis.yml
```diff
@@ -21,13 +21,14 @@ env:
...
 
 install:
-  - yarn install --non-interactive
+  - yarn install --non-interactive<% } %>
 
 script:
-  - yarn test<% } else { %>
-  - npm config set spin false<% } %>
+  - <% if (yarn) { %>yarn<% } else { %>npm<% } %> test
+  - <% if (yarn) { %>yarn<% } else { %>npm<% } %> run lint
```
Turbo87: maybe we should do linting before testing for faster feedback?

blueprints/app/files/package.json
```diff
@@ -12,6 +12,8 @@
   "repository": "",
   "scripts": {
     "build": "ember build",
+    "lint": "eslint *.js addon app config lib node-tests server tests",
+    "lint:fix": "<% if (yarn) { %>yarn run lint --fix<% } else { %>npm run lint -- --fix<% } %>",
```
Turbo87: for yarn we might not need `lint:fix` since `yarn lint --fix` will just work out of the box

given that template lint and stylelint also exist maybe we should call it `eslint` instead of `lint`? 🤔

rwjblue:
OK, finally got this updated (sorry for the delay). Addressed the following:

* Remove usage of `yarn run lint` (you can just `yarn lint`).
* Renamed `lint` script to `lint:js` to clarify that its only for JS files (and I suspect many apps would add their own `lint:css` script)
* Made the `lint:js:fix` variant only present on non-yarn addons.
* Move `yarn lint` / `npm run lint` earlier in list of CI scripts (to fail faster).

tests/fixtures/addon/npm/package.json
```diff
@@ -0,0 +1,52 @@
+{
```
kellyselden: Thank you for writing package.json fixture tests!

rwjblue:
> My remaining concern is, is it worth it to have the extra script in npm to save the extra "--"?

Not sure. I’d be happy to remove the second one and just document it in the README’s.  Think that is good enough?

kellyselden:
I like that idea.

rwjblue:
@kellyselden - 💨 POOF, DONE!

tests/fixtures/addon/npm/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "start": "ember serve",
+    "test": "ember try:each"
+  },
+  "dependencies": {
+    "ember-cli-babel": "^6.6.0"
+  },
+  "devDependencies": {
+    "broccoli-asset-rev": "^2.4.5",
+    "ember-ajax": "^3.0.0",
+    "ember-cli": "~<%= emberCLIVersion %>",
```
rwjblue: FWIW, I left this one value in the fixture (it is replaced in the test that does the assertion) specifically so that CI will not fail everytime we do a release...

ro0gr:
Recent [module-unification-app](https://github.com/ember-cli/ember-cli/pull/7488) blueprint is missing lint run script.

rwjblue:
Thanks @ro0gr (also grumble grumble 😜)

rwjblue:
Updated with changes to module-unification-app script:

* Synced up module-unification-app with main app blueprint
* Added module-unification-app fixture tests
* Updated module-unification-app to include `lint:js` script

PR #7453, opened 2017-11-21
Revert "Add prettier and eslint-plugin-ember to linting config."

rwjblue:
Reverts ember-cli/ember-cli#7450

Discussed with @kellyselden in slack, he would like more discussion before landing so this PR reverts for now.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -5,71 +5,71 @@ module.exports = {
       name: 'ember-lts-2.12',
       npm: {
         devDependencies: {
-          'ember-source': '~2.12.0',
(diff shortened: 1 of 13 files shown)


rwjblue:
FWIW, I still feel pretty strongly that the changes made in #7450 are the right path forward, and will continue to work towards convincing everyone else...

Turbo87:
@rwjblue it reads like Kelly is fine with the `ember` plugin rules though so maybe we can keep those in?

rwjblue:
TBH, I’m not actually sure what the exact concerns are. I’ll land this revert then bring things back one by one (in a fit of merge conflict horror) and ensure that @kellyselden (and other team members) have explicitly approved.

PR #7454, opened 2017-11-21
Remove assembler and strategies in favour of `packagerFor`

twokul:
More details could be found in [Packaging RFC](https://github.com/twokul/rfcs-1/blob/tree-shake-rfc/active/0051-packaging.md)

cc @kellyselden @rwjblue


rwjblue:
Generally looks good, though I haven’t looked into why CI is unhappy...

lib/broccoli/packager.js
```diff
@@ -0,0 +1,138 @@
...
+      the-best-app-ever.map (if sourcemaps are enabled)
+      vendor.js
+      vendor.map (if sourcemaps are enabled)
+    ```
+
+    @method packagerFor
+    @param {EmberApp} app
+    @return {BroccoliTree} Final Broccoli Tree.
+   */
+  packagerFor(app) {
```
rwjblue: If this is the only export, should we rename the file to be `packager-for` and modify this to `module.exports = function() {}`?

Specifically:

```js
const packagerFor = require('ember-cli/lib/broccoli/package').packagerFor;
```

vs:

```js
const packagerFor = require('ember-cli/lib/broccoli/packager-for');
```

lib/broccoli/packager.js
```diff
@@ -0,0 +1,138 @@
...
+      'vendor/loader/loader.js',
+      'vendor/ember-resolver/legacy-shims.js',
+      ...
+    ]
+  }
+  Produces an object that looks like:
+  {
+    headerFiles: [
+     'vendor/ember-cli/vendor-prefix.js',
+     'vendor/loader/loader.js',
```
ro0gr: It's a minor thing but seems like loader has to be placed right after the app-shims

lib/broccoli/packager.js
```diff
@@ -0,0 +1,138 @@
...
+      });
+
+      let importPaths = Object.keys(app._scriptOutputFiles);
+
+      // iterate over the keys and create N strategies
+      // to support scenarios like
+      // app.import('vendor/foobar.js', { outputFile: 'assets/baz.js' });
+      let packagedVendorTree = mergeTrees(importPaths.map(importPath => {
+        const files = app._scriptOutputFiles[importPath];
+        const isMainVendorFile = importPath === DEFAULT_CONFIG.outputPaths.vendor.js;
```
ro0gr: Should we take into account app.options.outputPaths.vendor.js here?

twokul: no, because we need to compare to default path

ro0gr: I'm not sure about this. If I specify custom path to my `vendor.js`:
```js
  let app = new EmberApp(defaults, {
    outputPaths: {
      vendor: {
        js: 'assets/funky-vendor.js'
      }
    }
  });
```
then `isMainVendorFile` is never equal to `true`, right?

kellyselden:
Updated title to reflect that we are actually just removing, not deprecating.

lib/broccoli/packager-for.js
```diff
@@ -0,0 +1,135 @@
...
+        'vendor/ember-cli/app-suffix.js',
+        'vendor/ember-cli/app-config.js',
+        'vendor/ember-cli/app-boot.js',
+      ],
+      outputFile: app.options.outputPaths.app.js,
+      annotation: 'Concat App',
+      sourceMapConfig: app.options.sourcemaps,
+    });
+
+    let importPaths = Object.keys(app._scriptOutputFiles);
```
kellyselden: Since this doesn't need inputTree, I would move it to the outside scope. Not necessary though.

lib/broccoli/ember-app.js
```diff
@@ -1404,38 +1405,15 @@ class EmberApp {
...
-        sourceMapConfig: this.options.sourcemaps,
-        annotation: 'Vendor JS',
-      }));
-    });
-
-    let assembler = new Assembler(applicationJs, {
-      annotation: 'Assembler (vendor & appJS)',
-      strategies,
-    });
+    const debugTree = BroccoliDebug.buildDebugCallback('assembler');
```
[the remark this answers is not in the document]
twokul: don't think so, could move it

rwjblue:
> Should we enable eslint-plugin-prefer-let for the project

Seems good to me

twokul:
> > Should we enable eslint-plugin-prefer-let for the project

> Seems good to me 

@kellyselden @rwjblue 

sure. the only downside is that our lazy `require`s are going to have to be converted to using `let`. For example, code in [lib/models/instrumentation.js#L12-L20](https://github.com/ember-cli/ember-cli/blob/master/lib/models/instrumentation.js#L12-L20) is going to be changed to:

```diff
function _enableFSMonitorIfInstrumentationEnabled() {
  let monitor;
  if (instrumentationEnabled()) {
-   const FSMonitor = require('heimdalljs-fs-monitor');
+   let FSMonitor = require('heimdalljs-fs-monitor');
    monitor = new FSMonitor();
    monitor.start();
  }
  return monitor;
}
```

we could theoretically right our own thing or just contribute back to support use case above (basically, always use `const` for `require`)

lib/broccoli/packager-for.js
```diff
@@ -0,0 +1,135 @@
...
+    vendor.js
+    vendor.map (if sourcemaps are enabled)
+  ```
+
+  @method packagerFor
+  @param {EmberApp} app
+  @return {BroccoliTree} Final Broccoli Tree.
+  */
+module.exports = function packagerFor(app) {
+  return function(inputTree) {
```
tsing80: packagerFor  or packageFor?

twokul: `packagerFor`

kellyselden:
@twokul @rwjblue tracked here https://github.com/cowboyd/eslint-plugin-prefer-let/pull/5

PR #7458, opened 2017-11-21
Add prettier to default linting configuration.

rwjblue:
Brings back part of #7450 which was reverted in #7453.

From the original description of #7450:

> `prettier` has taken the JavaScript ecosystem by storm! It has quickly become the most common way to acheive "standard code style" without being nit-picky with linting rules. The configuration that is landed here integrates `prettier` into `eslint` so that `eslint` properly identifies things that are not "pretty" and a quick `eslint --fix` will auto-correct them. This also means that any editors with built-in `eslint --fix` support already support `prettier`.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -5,71 +5,71 @@ module.exports = {
       name: 'ember-lts-2.12',
       npm: {
         devDependencies: {
-          'ember-source': '~2.12.0'
-        }
-      }
+          'ember-source': '~2.12.0',
+        },
+      },
     },
     {
       name: 'ember-lts-2.16',
       npm: {
         devDependencies: {
-          'ember-source': '~2.16.0'
-        }
-      }
+          'ember-source': '~2.16.0',
+        },
+      },
     },
     {
       name: 'ember-release',
       bower: {
         dependencies: {
-          'ember': 'components/ember#release'
+          ember: 'components/ember#release',
         },
         resolutions: {
-          'ember': 'release'
-        }
+          ember: 'release',
+        },
       },
       npm: {
         devDependencies: {
-          'ember-source': null
-        }
-      }
+          'ember-source': null,
+        },
+      },
     },
     {
       name: 'ember-beta',
       bower: {
         dependencies: {
-          'ember': 'components/ember#beta'
+          ember: 'components/ember#beta',
         },
         resolutions: {
-          'ember': 'beta'
-        }
+          ember: 'beta',
+        },
       },
       npm: {
         devDependencies: {
-          'ember-source': null
-        }
-      }
+          'ember-source': null,
+        },
+      },
     },
     {
       name: 'ember-canary',
       bower: {
         dependencies: {
-          'ember': 'components/ember#canary'
+          ember: 'components/ember#canary',
         },
         resolutions: {
-          'ember': 'canary'
-        }
+          ember: 'canary',
+        },
       },
       npm: {
         devDependencies: {
-          'ember-source': null
-        }
-      }
+          'ember-source': null,
+        },
+      },
     },
     {
       name: 'ember-default',
       npm: {
-        devDependencies: {}
-      }
-    }
-  ]
+        devDependencies: {},
+      },
+    },
+  ],
 };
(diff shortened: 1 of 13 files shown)


rwjblue:
@kellyselden - This was landed in #7450 then reverted in #7453 based on concerns you mentioned in Slack.  Can you please chime in here with those concerns, so they can be properly addressed?

kellyselden:
Unpopular opinion incoming:

Trailing commas

This style rule would become the new Ember style. Docs would eventually be updated. It still looks unfamiliar to me. I feel like a few power users of Ember have gotten used to the style, but why should we recommend everyone use this style. Is it worth the churn? Am I wrong and even non-power users prefer this style?

How we teach

I originally thought it would be hard for people not knowing what prettier was to work with it, but then I saw #7452 come through and this is no longer an issue

observations of prettier formatted code

Sometimes it cleans up code and does stuff you forgot to do (plus!), but I see it uglifying a lot of code. Particularly if statements and array collapsing. IMO code goes from being human readable to less so.

screen shot 2017-11-21 at 11 07 21 am

screen shot 2017-11-21 at 11 09 45 am

screen shot 2017-11-21 at 11 08 30 am

^ logic newlined to make it easier to grep is now back to being difficult (could be refactored, but that's another story)

screen shot 2017-11-21 at 11 23 38 am

^ seems strange to be selective like that, but not a big deal (don't worry, fake data)

Not sure about others, but I have always used copious amounts of newlines to make git history and blame easier to work with. This mantra kind goes out the window with array and boolean collapsing.

It feels like in a fair amount of cases we make the tradeoff of readable vs pretty. It may be worth it though.

Those are my thoughts. Feel free to poke holes in my observations.

twokul:
@kellyselden very much agree with you

kategengler:
@kellyselden In terms of the arrays getting uglier / compacted onto one line, that seems to be part of the reason prettier suggests keeping the line width to 80 (https://prettier.io/docs/en/options.html#print-width) - this PR sets it to 120.

Turbo87:
@kellyselden I have to admit that those are good points. Prettier can be disabled for certain lines but that would just add even more noise... Guess we should discuss this on Thursday before deciding anything

rwwagner90:
I have to agree that I don't like trailing commas. Makes the code look messier to me, although I do understand the benefits and why people would argue for them. As for array collapsing, I think that will just blow away any intended spacing people added on purpose. Why wouldn't we allow formatting either way?

alexlafroscia:
> As for array collapsing, I think that will just blow away any intended spacing people added on purpose. Why wouldn't we allow formatting either way?

I don't think you can do that with Prettier, can you? It'll try to put it all on one line if it fits, based on the configured line length. If it doesn't, each element gets their own line.

Maybe that's an argument for keeping the line length at the recommended 80 characters, or maybe 100 as a decent middle ground. I personally feel like their reasoning for maintaining the default 80 is valid.

rwwagner90:
@alexlafroscia if we only break when line length is exceeded, wouldn't we end up with arrays like:
```js
[
 'foo', 'bar', 'baz', ...
 'other strings', '123' ...
]
```

While that's not terribly unreadable, once we exceed the line length, a single item per line may be cleaner.

rwjblue:
We discussed this in today's ember-cli team meeting. While many team members are strongly in favor, we feel that this level of change warrants feedback from the broader community via an RFC.

I'll try to put the RFC together in the next week or so, and cross link back here...

alexlafroscia:
@rwwagner90 nope, that's not how Prettier formats arrays. If the _entire_ array fits on one line, then it's on one long. Otherwise, each element gets its own line. Never a handful on one line and a handful on the next.

NullVoxPopuli:
I think we should revisit prettier.
It forces a code style that is overall easier to read.

Wile the formatted array thing is an issue, I think it's more of an edge case, and not a reason to not use prettier.

pichfl:
I use prettier in many Ember based projects and as @kategengler mentions, trying to go to 120 chars per line will make some odd decisions. The recommended 80 chars are a bit narrow for my taste, but going half way to 100 per line did resolve many odd choices in my projects.

Would be great to bring this back.

NullVoxPopuli:
I think 100 chars is perfect in general

alexlafroscia:
If Prettier is added at all, I really think we should leave the configuration alone and use the default. It’s really easy for people to add their own `.prettierrc` if they want to, but I’m not sure there’s a good reason for Ember to take a different opinion on the defaults than Prettier itself.

Issue #7459, opened 2017-11-22
availableOptions not normalized after parent's init 


thoov:
```js
module.exports = Command.extend({
  name: 'foo',
  
  init() {
    this._super();

    this.availableOptions = [
      { name: 'bar', type: Boolean, default: false },
      { name: 'baz', type: Boolean, default: false }
    ]
  },

  run(commandOptions, rawArgs) {
    console.log(commandOptions); 
    
    // commandOptions instead looks like: { bar: false, undefined: false, ... } where 
    // a 'undefined' key has been added instead of 'baz'
  }
});
```

The issue is that the normalization of availableOptions is happening before the child's init is even called. We see that in the Command class's init method we call [registerOptions](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L159) which does all of the normalization of the availableOptions. However this will happen before the child's init method and definitely before the child's beforeRun method is invoked. Thus leading to a situation where the child's options never went through the normalization phase. Since it does not go through this phase it does not get the property 'key' added to it and when it tries to grab it later on it get 'undefined' which is what we see happening.

There seems to be 2 ways we can approach a fix for this:

1) Change how the command is invoked by moving registerOptions out of the init flow and into its own distinct path that can be executed by the [cli like so](https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L152):

```js
  // ...
  onProcessInterrupt.addHandler(onCommandInterrupt);

  return command.beforeRun(commandArgs);
}).then(() => {
  loggerTesting.info('cli: command.registerOptions');

  return command.registerOptions(); // <---- something like this
}).then(() => {
  loggerTesting.info('cli: command.validateAndRun');

  return command.validateAndRun(commandArgs);
}).then(result => {
// ...
``` 

2) Change the behavior of [assignOption](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L400) to be more proactive in case that the 'key' property is not there.

PR #7461, opened 2017-11-22
Change when availableOptions are normalized

thoov:
The normalization of the availableOptions was done during the init step on the parent Command class. However, this meant that if a child class was to extend from Command they could not set availableOptions within their own init or beforeRun steps as the normalization would have already been done. This change moves the normalization to just before it is needed and after init, beforeRun, etc have happened.

Fixes: #7459

### DIFF
--- a/tests/acceptance/help-test.js
+++ b/tests/acceptance/help-test.js
@@ -48,7 +48,7 @@ describe('Acceptance: ember help', function() {
   });
 
   it('works', function() {
-    command.validateAndRun([]);
+    command.run(options, []);
 
     let output = options.ui.output;
 
@@ -157,6 +157,7 @@ describe('Acceptance: ember help', function() {
 
       let json = convertToJson(options.ui.output);
       const expected = require('../fixtures/help/with-addon-commands.js');
+
       expect(json).to.deep.equal(expected);
     });
 
@@ -166,6 +167,7 @@ describe('Acceptance: ember help', function() {
       };
 
       command.run(options, []);
+
       let json = convertToJson(options.ui.output);
       const expected = require('../fixtures/help/with-addon-blueprints.js');
 
--- a/tests/acceptance/help-test.js
+++ b/tests/acceptance/help-test.js
@@ -48,7 +48,7 @@ describe('Acceptance: ember help', function() {
   });
 
   it('works', function() {
-    command.run(options, []);
+    command.validateAndRun([]);
 
     let output = options.ui.output;
 
@@ -140,12 +140,12 @@ describe('Acceptance: ember help', function() {
     });
 
     it('works', function() {
-      command.run(options, []);
+      command.validateAndRun([]).then(() => {
+        let json = convertToJson(options.ui.output);
+        const expected = require('../fixtures/help/help.js');
 
-      let json = convertToJson(options.ui.output);
-      const expected = require('../fixtures/help/help.js');
-
-      expect(json).to.deep.equal(expected);
+        expect(json).to.deep.equal(expected);
+      });
     });
 
     it('prints commands from addons', function() {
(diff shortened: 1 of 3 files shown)


tests/acceptance/help-test.js
```diff
@@ -140,38 +140,38 @@ describe('Acceptance: ember help', function() {
     });
 
     it('works', function() {
-      command.run(options, []);
-
-      let json = convertToJson(options.ui.output);
-      const expected = require('../fixtures/help/help.js');
+      command.validateAndRun([]).then(() => {
```
rwjblue: I think we need to return this promise, or the tests aren't going to properly wait for the async...

tests/acceptance/help-test.js
```diff
@@ -140,38 +140,38 @@ describe('Acceptance: ember help', function() {
...
+      });
     });
 
     it('prints commands from addons', function() {
       options.project.eachAddonCommand = function(cb) {
         cb('dummy-addon', { Foo: FooCommand });
       };
 
-      command.run(options, []);
+      command.validateAndRun([]).then(() => {
```
rwjblue: `return`

tests/acceptance/help-test.js
```diff
@@ -140,38 +140,38 @@ describe('Acceptance: ember help', function() {
...
     it('prints blueprints from addons', function() {
       options.project.blueprintLookupPaths = function() {
         return [path.join(__dirname, '..', 'fixtures', 'blueprints')];
       };
 
-      command.run(options, []);
-
-      let json = convertToJson(options.ui.output);
-      const expected = require('../fixtures/help/with-addon-blueprints.js');
+      command.validateAndRun([]).then(() => {
```
rwjblue: Needs `return`

thoov:
@rwjblue After going deeper on this I believe there are a few things that are contradictory about how the logic is constructed today.

The command and the surrounding logic is built with the assumption that availableOptions is a getter (as you brought up about moving towards). However, in command we have the `beforeRun`hook which states that we [allow extending availableOptions](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L256). This is currently not working as the normalizeOption method will have already run before this hook. Also, we run the availableOptions (and some other properties) through normalization steps which modify them in place. So to me these two things (move to a pure getter vs beforeRun extensibility & normalization mutability) contradict each other. As I see it here are the options we have:

1) Move to a "pure" getter of the these properties. This would mean not supporting one of the stated reasons for beforeRun or any kind of extensibility on these properties. I did a code [search on ember-observer for beforeRun](https://emberobserver.com/code-search?codeQuery=beforeRun) and this method does not seem to be in use but it is technically changing behavior of a public API.

2) This PR in its current form. All this does is make sure that these in place normalizations happen. This solution works fine however its more like putting a bandaid on (It is very easy in the future for someone to forget about manually normalizing these props).

3) Move away from in place modifications of these properties, [namely no longer call things like this](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L185). I have not gone completely down this path but it would mean putting checks in places [like here](https://github.com/ember-cli/ember-cli/blob/master/lib/models/command.js#L402) instead of relying on `.key`

I'm open to input on which path we feel is the best / acceptable.

stefanpenner:
@thoov is there a downside to merging this now, and making it better at a later day? Is there expected breakage/downside etc.?

thoov:
@stefanpenner 

> is there a downside to merging this now, and making it better at a later day?

Not that I can think of. We should whiteboard this the next time you are around so we can come up with a better longer term solution.

> Is there expected breakage/downside etc.?

No

stefanpenner:
>  We should whiteboard this the next time you are around so we can come up with a better longer term solution.

Let's doit!

Issue #7355, opened 2017-10-04
Passing --no-yarn to ember install has no effect


PieterJanVdb:
Using the --no-yarn option when using ember install has no effect, meaning it will still use yarn.
Same when using --yarn=false.

![reproduction](https://i.imgur.com/v5WxCh6.png)

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 2.15.1
http_parser: 2.7.0
node: 6.9.2
v8: 5.1.281.88
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
3.10.9
0.21.3
```

rwjblue:
In general, if you don't want to use `yarn` you would delete `yarn.lock`.  `ember-cli` will use `npm` if there is no `yarn.lock`.  IMHO, having a `yarn.lock` but intentionally not maintaining it when being asked to install things is a really odd position to be in and almost certainly will result in bugs...

PieterJanVdb:
@rwjblue  Ah my bad, I just thought it was a bug as --yarn is an option specified on the "ember install" command, as seen here (lib/commands/install):

```
module.exports = Command.extend({
  name: 'install',
  description: 'Installs an ember-cli addon from npm.',
  aliases: ['i'],
  works: 'insideProject',

  availableOptions: [
    { name: 'save',       type: Boolean, default: false, aliases: ['S'] },
    { name: 'save-dev',   type: Boolean, default: true,  aliases: ['D'] },
    { name: 'save-exact', type: Boolean, default: false, aliases: ['E', 'exact'] },
    { name: 'yarn',       type: Boolean, description: 'Use --yarn to enforce yarn usage, or --no-yarn to enforce npm' }, // no default means use yarn if the blueprint has a yarn.lock
  ],

```

But I completely follow your reasoning of having a yarn.lock and not maintaining it. We used to use yarn in the project I tried to use ember install --no-yarn but have since stopped using it and must not have removed the yarn.lock.

Shall I close the issue, or edit the command so that 'yarn' is not one of the CLI options?

rwjblue:
Hmm. Yes, if we have that option it should work. 

@turbo87 - Thoughts?

Turbo87:
fixed by https://github.com/ember-cli/ember-cli/issues/7355

PR #7463, opened 2017-11-28
[BUGFIX] pass --yarn through AddonInstall task to NpmInstall task

lennyburdette:
The `install` task documents a --yarn option which should change the behavior of the NpmInstall task. However, the AddonInstall task did not pass the option all the way through.

### DIFF
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -42,6 +42,7 @@ class AddonInstallTask extends Task {
       'save': commandOptions.save,
       'save-dev': !commandOptions.save,
       'save-exact': commandOptions.saveExact,
+      useYarn: Boolean(commandOptions.yarn),
     }).then(() => this.project.reloadAddons())
       .then(() => this.installBlueprint(blueprintInstall, packageNames, blueprintOptions))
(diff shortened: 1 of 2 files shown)


lennyburdette:
Addresses #7355 (found the issue after making the fix.)

lennyburdette:
@rwjblue @Turbo87 Is there any prior art for finding yarn.lock in a parent directory? The [existing check](https://github.com/lennyburdette/ember-cli/blob/master/lib/tasks/npm-task.js#L44) will fail if the Ember CLI app is part of a yarn workspace.

I decided to tackle this because with this fix we can add `"yarn": true` to `.ember-cli` and force yarn regardless of the lock file location.

Turbo87:
this is similar to https://github.com/ember-cli/ember-cli/pull/7354, right?

Turbo87:
> Is there any prior art for finding yarn.lock in a parent directory?

we are using `find-up` or something like that in a few places. maybe it would make sense in this case too 🤔

lennyburdette:
It's pretty much exactly the same as #7354 ... I'll close this and add my comments to that one. Thanks!

PR #7467, opened 2017-11-29
Revert "Disabled all experiments except for CONFIG_CACHING"

tchak:
I think experiments should be enabled on master cc @Turbo87


Turbo87:
merci! 👍

tchak:
It would have been too easy :)

PR #7468, opened 2017-11-29
use ember-source instead of bower component in ember try

kellyselden:
I had this branch laying around, not sure why I never submitted. Are these the right tags? cc @rwjblue

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -19,49 +19,25 @@ module.exports = {
     },
     {
       name: 'ember-release',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#release'
-        },
-        resolutions: {
-          'ember': 'release'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': 'github:emberjs/ember.js#release'
         }
       }
     },
     {
       name: 'ember-beta',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#beta'
-        },
-        resolutions: {
-          'ember': 'beta'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': 'github:emberjs/ember.js#beta'
         }
       }
     },
     {
       name: 'ember-canary',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#canary'
-        },
-        resolutions: {
-          'ember': 'canary'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': 'github:emberjs/ember.js#master'
         }
       }
     },
--- a/tests/fixtures/addon/npm/config/ember-try.js
+++ b/tests/fixtures/addon/npm/config/ember-try.js
@@ -18,49 +18,25 @@ module.exports = {
     },
     {
       name: 'ember-release',
-      bower: {
-        dependencies: {
-          'ember': 'components/ember#release'
-        },
-        resolutions: {
-          'ember': 'release'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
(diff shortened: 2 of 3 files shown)


blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -19,49 +19,25 @@ module.exports = {
...
-          'ember': 'components/ember#canary'
-        },
-        resolutions: {
-          'ember': 'canary'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': 'emberjs/ember.js'
```
rwjblue: I don't think this works exactly as we expect, can you test locally?  I believe you want `github:emberjs/ember.js#master` to explicitly state master.  It totally might work as is, I just am not sure...

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -19,49 +19,25 @@ module.exports = {
...
-          'ember': 'components/ember#release'
-        },
-        resolutions: {
-          'ember': 'release'
-        }
-      },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': 'latest'
```
rwjblue: This is slightly different from what we had before.  `latest` is going to be the most recent _tagged_ release on NPM, whereas `release` in bower was the build from the last commit on the `release` branch.

I believe that we want all three of these to be something like:

```js
npm: {
  devDependencies: {
    'ember-source': 'github:emberjs/ember.js#release',
  }
}
```

kellyselden: I will update all the branch hashes, as for github:, that can be left off and it will be implied. Would you prefer explicit or implicit github?

rwjblue: I like being explicit, mostly just because I've had the question asked by _a lot_ of teammates "what does this mean, where is this version number from" when using without `github:` prefix.  It _probably_ doesn't matter here (addon authors are almost certainly going to be able to figure it out), but if its all the same to you I like the prefix...

kellyselden: will do!

kellyselden:
All feedback addressed.

Turbo87:
wait, how does that work? does it build all of Ember on demand then?!

kellyselden:
@Turbo87 Your right, unless ember checks in dist, it won't work...

rwjblue:
npm@5 and yarn both support this as of the changes I made in in https://github.com/emberjs/ember.js/pull/15789

rwjblue:
They install devDeps then run prepare script. It’s documented in the npm install docs for git repo references.

Turbo87:
@rwjblue correct me if I'm wrong, but that would still mean that it needs to build Ember from scratch which will increase CI times, right?

rwjblue:
> correct me if I'm wrong, but that would still mean that it needs to build Ember from scratch which will increase CI times, right?

Yep, but 1) embers build is less than a minute, 2) bower support in Ember is gone as of 3.0, 3) this is _exactly_ the same as **every other addon** 😜...

rwjblue:
Oh, and 4) with targets, svelte, etc the host app will almost certainly be building Ember all the time soon.

kellyselden:
With addons we pin the npm version to 4. We will have to bump that to 5 before merging. I think 5 is stable enough that that shouldn't be an issue.

rwjblue:
> With addons we pin the npm version to 4. We will have to bump that to 5 before merging.

Well, or default to using yarn 😈

rwjblue:
I do believe there is a way to make it work with older npm versions, but someone needs to put in the time. 🤔

Turbo87:
> Well, or default to using yarn 😈

we should discuss this for 3.0...

rwjblue:
Was replaced by https://github.com/ember-cli/ember-cli/pull/7536

PR #7469, opened 2017-11-29
Add minimum supported version badge to default addon readme

kellyselden:
This is our recommended supported ember version for new addons (2 LTS's back). I use these a lot, and it seems to me, this is a good starting point for addon authors to consider their own support.

We can discuss at our meeting further.

### DIFF
--- a/blueprints/addon/files/README.md
+++ b/blueprints/addon/files/README.md
@@ -1,6 +1,8 @@
 <%= addonName %>
 ==============================================================================
 
+[![Ember Version](https://img.shields.io/badge/ember-2.18%2B-brightgreen.svg)](https://www.emberjs.com/)
+
 [Short description of the addon.]
 
 
--- a/blueprints/module-unification-addon/files/README.md
+++ b/blueprints/module-unification-addon/files/README.md
@@ -1,6 +1,8 @@
 <%= addonName %>
 ==============================================================================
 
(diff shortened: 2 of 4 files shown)


Turbo87:
why not use a service like http://shields.io/ for that? I'm a little worried about how long embadge.io will stick around 🤔

kellyselden:
I'm cool with a different provider.

kellyselden:
It looks like emberobserver might come through for us with a badge! cc @kategengler

Turbo87:
I'm starting to wonder if we should instead include a "Compatibility" section in the default generated README with bullet points for "Ember" and "Ember CLI" 🤔

kellyselden:
I changed it to shields.io, and I think @kennethlarsen expressed interest in this.

Turbo87:
@kennethlarsen opened a PR to add a dedicated README section for this: https://github.com/ember-cli/ember-cli/pull/8286

kellyselden:
I think a badge looks better, but that's the only reason for my preference of this...

Turbo87:
if it was only a single badge that might be true, but addons might have other dependencies too. they could depend on a minimum Ember version, Ember Data version, Ember CLI version, or even a minimum `ember-intl` version for example.

kellyselden:
Sure.

PR #7470, opened 2017-11-30
[BUGFIX] Restore `separator: '\n;'` to vendor JS concat

lennyburdette:
Prior to 2.17, Ember CLI used a semicolon to separate files concatenated into vendor.js: https://github.com/ember-cli/ember-cli/blob/v2.16.2/lib/broccoli/bundler.js#L92

Vendor shims are still generated with a leading paren:
```
(function() {
  function vendorModule() {
    'use strict';

    return {
      'default': self['test'],
      __esModule: true,
    };
  }

  define('test', [], vendorModule);
})();
```

This can cause `TypeError: (intermediate value)(...) is not a function` if the code immediately prior does not end in a semicolon.


rwjblue:
Seems good to fix, thank you for working on it!

Can you update the PR to target the `release` branch so we can get a patch release out for it?

Turbo87:
replaced by https://github.com/ember-cli/ember-cli/pull/7477

kellyselden:
We merged your commit in that one. Thanks for the contribution!

PR #7474, opened 2017-11-30
Give plugins and extends their own lines

rwwagner90:
I had several extra `plugins` and `extends` entries defined in some of my `.eslintrc.js` files. The idea here is, if we put each one on a separate line, and you use the default suggestions, plus add your own, you won't get a diff where the whole plugins and extends sections are completely rewritten, but the diff will just show the extra ones you added. Please let me know if you guys think this is a good idea or not!

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -4,8 +4,13 @@ module.exports = {
     ecmaVersion: 2017,
     sourceType: 'module'
   },
-  plugins: ['ember'],
-  extends: ['eslint:recommended', 'plugin:ember/recommended'],
+  plugins: [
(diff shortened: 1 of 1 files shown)


kellyselden:
Seems good to me. This would probably get reverted by prettier if we used it, just something for us to keep in mind.

rwwagner90:
@kellyselden what are the plans for prettier? Could we ignore this file?

kellyselden:
https://github.com/ember-cli/ember-cli/pull/7458

rwwagner90:
@kellyselden I see. I commented on that issue. I would personally be against forcing array collapsing, but not up to me, so unsure of whether or not we should treat this as a special case or not.

PR #7475, opened 2017-11-30
don't treat strings as regex in insertIntoFile

### LINKED ISSUE #7473: options.before/after in insertIntoFile only supports valid regex
The `insertIntoFile` example in the docs (https://ember-cli.com/api/classes/Blueprint.html#method_insertIntoFile) doesn't work in ember-cli@2.17+ because of the regex feature added to `options.before/after` in #7384

```js
''.match('Router.map(function() {')
// SyntaxError: Invalid regular expression: /Router.map(function() {/: Unterminated group
```

I noticed it when trying to install ember-cli-mirage, because of https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/ember-cli-mirage/index.js#L48

I'll probably raise a PR to workaround it in `ember-cli-mirage` unless you think it's something that should be addressed in `ember-cli`?

kellyselden:
Closes #7473

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -87,9 +87,11 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
         contentsToWrite += contentsToInsert;
       } else {
         let contentMarker = options[insertBehavior];
-        let matches = contentsToWrite.match(contentMarker);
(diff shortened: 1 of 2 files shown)


kellyselden:
CI is gradually going green. Anyone want to give it a once over for merge?

PR #7476, opened 2017-11-30
attempt to fix ci by adding sudo: required to .travis.ymll

lennyburdette:
as suggested here: https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-348235933

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 dist: trusty
 
 addons:


kellyselden:
It seems that travis is in the process of rolling out a fix https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-348332353

So a change is probably not going to be necessary.

lennyburdette:
:+1: I was mostly just curious to see if it would work. Thanks for the update!

kellyselden:
Thanks for being on top of this!

PR #7481, opened 2017-12-01
NpmInstallTask: `useYarn` from constructor args

lennyburdette:
Instead of from the args passed to `#run`.

This replaces https://github.com/ember-cli/ember-cli/pull/7354

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -136,8 +136,6 @@ class NpmTask extends Task {
   }
 
   run(options) {
-    this.useYarn = options.useYarn;
-
     return this.findPackageManager().then(() => {
       let ui = this.ui;
       let startMessage = this.formatStartMessage(options.packages);


lennyburdette:
@Turbo87 addressed your comment [here](https://github.com/ember-cli/ember-cli/pull/7354#discussion_r154315812) and I believe this will finally pass CI.

Turbo87:
@lennyburdette did you check that there are no calls to `#run` that use that option?

lennyburdette:
Here are the only places I can find that call `run` and none of them pass `useYarn`.

https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L909
https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L967
https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/addon-install.js#L40

Turbo87:
Thanks! 👍

knownasilya:
Anyone else use `--no-yarn` because you need to use `--ignore-engines`, maybe we should do `ember install my-addon -- --ignore-engine` format.

PR #7482, opened 2017-12-01
bump to npm 5 in addon CI

kellyselden:
This is the precursor necessary for https://github.com/ember-cli/ember-cli/pull/7468#issuecomment-348041400

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -46,7 +46,7 @@ install:
   - yarn install --no-lockfile --non-interactive
 <% } else { %>
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@5
   - npm --version
 <% } %>
 script:
--- a/tests/fixtures/addon/npm/.travis.yml
+++ b/tests/fixtures/addon/npm/.travis.yml
@@ -37,7 +37,7 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@5
   - npm --version
 
 script:


kellyselden:
any opposition? @ember-cli/core

Turbo87:
I have a hard time commenting on this since I don't use npm anymore. Whenever I'm forced to use it npm was much less reliable than yarn :(

FWIW we still can't use npm5 to publish Ember CLI itself since the tarball publishing feature is still broken 😭

kellyselden:
We should also look into how well ember-try is equipped to handle this. cc @kategengler 

I had a heck of a time getting npm 5 + ember try working https://github.com/kellyselden/ember-data-fetch-support/pull/107, not sure if I was doing something wrong or ember-try.

kellyselden:
I think it's time to YOLO merge this. Anyone have any issue with retargeting release and merging?

kellyselden:
@Turbo87 Yeah, once node 4 (npm 2) falls, then we could probably take this out.

kellyselden:
I take it back. The test times on travis between npm 4 and 5 are substantial. Couple minutes shaved off. I'd like to merge this. That ok @Turbo87?

Turbo87:
should we use `npm@6` now that it's officially released? I'm fine with either.

kellyselden:
Officially released had never meant stable for npm... I'd prefer not to jump to 6 yet for our user's sake.

mansona:
Personally, I think that we should be bumping to npm@6 (hence why I opened #8426 😂) and I agree with @rwjblue about testing in npm which is why I opened #8408 

I would agree with @kellyselden that the build time differences are **substantial** and in my very small sample set npm@6 is actually a minute or 2 faster than yarn 😱 and that's without a package-lock.json so I'm very surprised: https://travis-ci.org/ember-cli/ember-cli/builds/489432875?utm_source=github_status&utm_medium=notification

rwjblue:
The underlying issue was fixed by @kellyselden in https://github.com/ember-cli/ember-cli/pull/8645 and was released in ember-cli@3.10.1.

PR #7483, opened 2017-12-02
Modified some of the app blueprints with more explicit assignments.

davidrhyswhite:
Really simple example but I wondered if there would be some interest in a slightly more explicit ES6 syntax with regards to some of the generated files here as I often modify these after generating an application.

Specifically if any of the code mutates an object, even locally and scoped within a method it just feels slightly safer to avoid this as per: https://github.com/ember-cli/ember-cli/compare/master...davidrhyswhite:updating-app-bluprints-with-explicit-assignments?expand=1#diff-0fafaf304e20987aac38b22d10930bc0L8

I thought it would be best to see if there's any appetite for such changes on a small part before looking at all the generators? Or alternatively find out if these are set this way for a specific purpose?

### DIFF
--- a/blueprints/app/files/config/environment.js
+++ b/blueprints/app/files/config/environment.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = function(environment) {
-  let ENV = {
+  const ENV = {
     modulePrefix: '<%= modulePrefix %>',
     environment,
     rootURL: '/',
--- a/blueprints/app/files/ember-cli-build.js
+++ b/blueprints/app/files/ember-cli-build.js
@@ -3,7 +3,7 @@
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function(defaults) {
-  let app = new EmberApp(defaults, {
+  const app = new EmberApp(defaults, {
     // Add options here
   });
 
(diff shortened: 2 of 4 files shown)


davidrhyswhite:
Doh, I should have checked this against older versions of Node. I'll check Node v4 as I'm sure the `let` / `const` areas as stiff a small benefit.

Turbo87:
hey @davidrhyswhite, sorry for the (very) late response 😞 

https://github.com/DockYard/eslint-plugin-ember-suave/blob/master/docs/rules/no-const-outside-module-scope.md explains why in the Ember and Ember CLI projects we prefer to use `let` instead of `const` for most things (note that the "Rule Details" section of that link is currently inverted, see https://github.com/DockYard/eslint-plugin-ember-suave/pull/58 😉)

I do prefer using const in languages where the structure can actually be considered constant (e.g. Rust and C++), but imagine the following code:

```js
const obj = { foo: 1 };
obj.foo = 2;
```

The `const` should have indicated that `obj` is immutable, but it is actually only the reference that is constant, not the object itself. In other words even though `const` is used, in JS you can't actually trust the code that the object is indeed never modified.

The above plus the fact when you do want to modify a variable that was constant you have to change it back to `let` makes us think that it's mostly not worth the extra effort. We do prefer using `const` for module-scoped constants and top level CommonJS imports, but not for anything else.

PR #7487, opened 2017-12-06
[BUGFIX] give `ember new` error messages consistent color

GavinJoyce:
Running `ember new` currently results in a yellow message which is different from the other types of errors that `ember new *` can produce:

screen shot 2017-12-06 at 12 26 00

This PR makes all error messages red text


rwjblue:
Thanks @GavinJoyce !

PR #7488, opened 2017-12-06
Add `module-unification-app` blueprint

GavinJoyce:
This is the first step in a series of PRs which will bring support for Module Unification blueprints and generators.

An app using the module unification layout can now be created as follows:

`MODULE_UNIFICATION=true ember new my-app`
`MODULE_UNIFICATION=true ember new my-app --yarn`

Once this lands, https://github.com/emberjs/ember-module-unification-blueprint can be removed

TODO:

 * [x] add module unification `app` blueprint (extracted from https://github.com/emberjs/ember-module-unification-blueprint)
 * [x] tests
 * [x] address feedback

TODO in future PRs:

 * support `init` command
 * support `generate` commands
 * support addons
 * reduce the duplication between classic and MU blueprints (see https://github.com/ember-cli/ember-cli/pull/7488#issuecomment-349715930)
 * remove bower and user ember canary in package.json
 * replace `{{welcome-page}}` addon with component example

### DIFF
--- a/blueprints/module-unification-app/files/.eslintrc.js
+++ b/blueprints/module-unification-app/files/.eslintrc.js
@@ -1,0 +1,48 @@
+module.exports = {
+  root: true,
+  parserOptions: {
+    ecmaVersion: 2017,
+    sourceType: 'module'
+  },
+  plugins: [
+    'ember'
+  ],
+  extends: [
+    'eslint:recommended',
+    'plugin:ember/recommended'
+  ],
+  env: {
+    browser: true
+  },
+  rules: {
+  },
+  overrides: [
+    // node files
+    {
+      files: [
+        'index.js',
+        'testem.js',
+        'ember-cli-build.js',
+        'config/**/*.js',
+        'tests/dummy/config/**/*.js'
+      ],
+      parserOptions: {
+        sourceType: 'script',
+        ecmaVersion: 2015
+      },
+      env: {
+        browser: false,
+        node: true
+      }
+    },
+
+    // test files
+    {
+      files: ['tests/**/*.js'],
+      excludedFiles: ['tests/dummy/**/*.js'],
+      env: {
+        embertest: true
+      }
+    }
+  ]
+};
--- a/blueprints/module-unification-app/files/bower.json
+++ b/blueprints/module-unification-app/files/bower.json
@@ -1,0 +1,6 @@
+{
+  "name": "<%= name %>",
+  "dependencies": {
+    "ember": "components/ember#canary"
+  }
+}
--- a/blueprints/module-unification-app/files/config/environment.js
+++ b/blueprints/module-unification-app/files/config/environment.js
@@ -1,0 +1,56 @@
+'use strict';
+
+module.exports = function(environment) {
+  let ENV = {
+    'ember-resolver': {
+      features: {
+        EMBER_RESOLVER_MODULE_UNIFICATION: true
+      }
+    },
+    modulePrefix: '<%= modulePrefix %>',
+    environment,
+    rootURL: '/',
+    locationType: 'auto',
+    EmberENV: {
+      FEATURES: {
+        // Here you can enable experimental features on an ember canary build
+        // e.g. 'with-controller': true
+        'ember-module-unification': true
+      },
+      EXTEND_PROTOTYPES: {
+        // Prevent Ember Data from overriding Date.parse.
+        Date: false
+      }
+    },
+
+    APP: {
+      // Here you can pass flags/options to your application instance
+      // when it is created
+    }
+  };
+
+  if (environment === 'development') {
+    // ENV.APP.LOG_RESOLVER = true;
+    // ENV.APP.LOG_ACTIVE_GENERATION = true;
+    // ENV.APP.LOG_TRANSITIONS = true;
+    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
+    // ENV.APP.LOG_VIEW_LOOKUPS = true;
+  }
+
+  if (environment === 'test') {
+    // Testem prefers this...
+    ENV.locationType = 'none';
+
+    // keep test console output quieter
+    ENV.APP.LOG_ACTIVE_GENERATION = false;
+    ENV.APP.LOG_VIEW_LOOKUPS = false;
+
+    ENV.APP.rootElement = '#ember-testing';
+  }
+
+  if (environment === 'production') {
+
+  }
+
+  return ENV;
+};
--- a/blueprints/module-unification-app/files/config/targets.js
+++ b/blueprints/module-unification-app/files/config/targets.js
@@ -1,0 +1,8 @@
+module.exports = {
+  browsers: [
+    'ie 9',
+    'last 1 Chrome versions',
+    'last 1 Firefox versions',
+    'last 1 Safari versions'
+  ]
+};
--- a/blueprints/module-unification-app/files/ember-cli-build.js
+++ b/blueprints/module-unification-app/files/ember-cli-build.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, {
+    // Add options here
+  });
+
+  // Use `app.import` to add additional libraries to the generated
+  // output files.
+  //
+  // If you need to use different assets in different
+  // environments, specify an object as the first parameter. That
+  // object's keys should be the environment name and the values
+  // should be the asset to use in that environment.
+  //
+  // If the library that you are including contains AMD or ES6
+  // modules that you would like to import into your application
+  // please specify an object with the list of modules as keys
+  // along with the exports of each module as its value.
+
+  return app.toTree();
+};
--- a/blueprints/module-unification-app/files/package.json
+++ b/blueprints/module-unification-app/files/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "<%= name %>",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Small description for <%= name %> goes here",
+  "license": "MIT",
+  "author": "",
+  "directories": {
+    "doc": "doc",
+    "test": "tests"
+  },
+  "repository": "",
+  "scripts": {
+    "build": "ember build",
+    "start": "ember server",
+    "test": "ember test"
+  },
+  "devDependencies": {
+    "broccoli-asset-rev": "^2.4.5",
+    "ember-ajax": "^3.0.0",
+    "ember-cli": "github:ember-cli/ember-cli",
+    "ember-cli-app-version": "^3.0.0",
+    "ember-cli-babel": "^6.6.0",
+    "ember-cli-dependency-checker": "^2.0.0",
+    "ember-cli-eslint": "^4.2.1",
+    "ember-cli-htmlbars": "^2.0.1",
+    "ember-cli-htmlbars-inline-precompile": "^1.0.0",
+    "ember-cli-inject-live-reload": "^1.4.1",
+    "ember-cli-qunit": "^4.1.1",
+    "ember-cli-shims": "^1.2.0",
+    "ember-cli-sri": "^2.1.0",
+    "ember-cli-uglify": "^2.0.0",
+    "ember-data": "~2.18.0-beta.1",
+    "ember-export-application-global": "^2.0.0",
+    "ember-load-initializers": "^1.0.0",
+    "ember-resolver": "^4.0.0<% if (welcome) { %>",
+    "ember-welcome-page": "^3.0.0<% } %>",
+    "eslint-plugin-ember": "^5.0.0",
+    "loader.js": "^4.2.3"
+  },
+  "engines": {
+    "node": "^4.5 || 6.* || >= 7.*"
+  }
+}
--- a/blueprints/module-unification-app/files/src/resolver.js
+++ b/blueprints/module-unification-app/files/src/resolver.js
@@ -1,0 +1,13 @@
+import Resolver from 'ember-resolver/resolvers/fallback';
+import buildResolverConfig from 'ember-resolver/ember-config';
+import config from '../config/environment';
+
+let moduleConfig = buildResolverConfig(config.modulePrefix);
+/*
+ * If your application has custom types and collections, modify moduleConfig here
+ * to add support for them.
+ */
+
+export default Resolver.extend({
+  config: moduleConfig
+});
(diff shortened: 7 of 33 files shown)


rwjblue:
> support addons

@mansona and I have a spike of what the output for addons should look like in https://github.com/stonecircle/ember-addon-output/tree/feature/module-unification-initial-work, hopefully that will help with adding addon support with the env flag...

blueprints/module-unification-app/files/bower.json
```diff
@@ -0,0 +1,6 @@
+{
```
rwjblue: We don't emit `bower.json` any longer

blueprints/module-unification-app/files/.eslintrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
```
rwjblue: Should be sync'ed with app's `.eslintrc.js` file (we use `overrides` and such now)...

blueprints/module-unification-app/files/bower.json
rwjblue: Hmm, maybe we still need one to get ember-canary build? 🤔

blueprints/module-unification-app/files/public/crossdomain.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
```
rwjblue: I think this file is removed for the normal app blueprint?

blueprints/module-unification-app/files/tests/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
rwjblue: The usage of `overrides` in `.eslintrc.js` should remove the need for these files

blueprints/module-unification-app/files/src/ui/components/my-component/component.js
```diff
@@ -0,0 +1,5 @@
+import Component from "@ember/component";
```
rwjblue: We probably don't want to emit a component...

blueprints/module-unification-app/files/src/ui/components/my-component/template.hbs
```diff
@@ -0,0 +1,5 @@
+<code>src/ui/components/my-component/template.hbs</code>
```
rwjblue: 🔪

blueprints/module-unification-app/files/src/ui/routes/index/template.hbs
```diff
@@ -0,0 +1,3 @@
+<code>src/ui/routes/index/template.hbs</code>
+
+{{my-component}}
```
rwjblue: Should invoke the `{{welcome-page}}` addon...

tests/acceptance/new-test.js
```diff
@@ -65,6 +65,19 @@ describe('Acceptance: ember new', function() {
...
+    process.env.MODULE_UNIFICATION = 'true';
+    yield ember([
+      'new',
+      'foo',
+      '--skip-npm',
+      '--skip-bower',
+    ]);
+
+    confirmBlueprintedForDir('blueprints/module-unification-app');
+    process.env.MODULE_UNIFICATION = undefined;
```
rwjblue: Should probably done in an `afterEach` in case the generation blows up and leaves us in a bad state...

GavinJoyce:
Thanks for the feedback @rwjblue. I copied over the content of https://github.com/emberjs/ember-module-unification-blueprint directly and didn't consider making any changes to the content of those files. I'll address your feedback this evening, thanks

GavinJoyce:
> I am getting concerned about maintenance of both blueprints over time with the structure here in the PR. Do you think we could refactor things a bit such that we only have to maintain the differences from the main app blueprint? I think that would dramatically reduce the total maintenance burden if we can swing it.

That's interesting, I'll have a think about how we might solve this. Do we know how long do we expect to have to maintain the two blueprints for?

GavinJoyce:
A possible solution would be to define the shared files in one path and the specific files in another, something like:

```
blueprints/shared/app
blueprints/app
blueprints/module-unification-app
```

The `app` blueprint would be the merge of `blueprints/shared/app` & `blueprints/app`. The `module-unification-app` blueprint would be the merge of `blueprints/shared/app` & `blueprints/module-unification-app`.

We'd need to make some changes to how we test too.

@rwjblue what do you think of this approach?

rwjblue:
In general, it seems good to me.

FWIW,  I do agree with your earlier point too (around adding complexity for a temp situation), I’m just not sure how to guess how long we will need to support both. I suspect it will be at least a couple of releases...

 I’m mostly just trying to set us up for success and I worry that during the time where we maintain both we might accidentally forget to keep them in sync properly and therefore make the end-user tester experience much worse...

GavinJoyce:
How about we live with the duplication for the short term and build out support for module unification `app`, `addon` and generators first. Once we have that, we'll have a better picture of the overlap and we can then make a decision to either to begin work on removing the duplication or to defer the work until a little later.

This will allow me to initially focus on landing mu support in blueprints

rwjblue:
Ok, sounds good to me.

blueprints/module-unification-app/files/src/ui/routes/index/template.hbs
GavinJoyce: To mirror the `app` blueprint, I'm going to zap the `index` route and add `{{welcome-page}}` to `application.hbs`

blueprints/module-unification-app/files/bower.json
GavinJoyce: LMK if it's possible to opt in to canary in `package.json`. Otherwise, I'll leave this as is for now

blueprints/module-unification-app/files/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "ember-cli-sri": "^2.1.0",
+    "ember-cli-uglify": "^2.0.0",
+    "ember-data": "~2.18.0-beta.1",
+    "ember-export-application-global": "^2.0.0",
+    "ember-load-initializers": "^1.0.0",
+    "ember-resolver": "^4.0.0<% if (welcome) { %>",
+    "ember-welcome-page": "^3.0.0<% } %>",
+    "eslint-plugin-ember": "^5.0.0",
+    "loader.js": "^4.2.3"
+  },
```
GavinJoyce: I've updated these dependencies to more closely mirror https://github.com/ember-cli/ember-cli/blob/d9c015974436ab0b80f552ed7689e968e3d50f91/blueprints/app/files/package.json

GavinJoyce:
@rwjblue I've addressed your feedback. 

Here's a repo with the current output of `MODULE_UNIFICATION=true ember new module-unification-blueprint-app-output`:

https://github.com/GavinJoyce/module-unification-blueprint-app-output

screen shot 2017-12-07 at 13 54 03
screen shot 2017-12-07 at 13 54 09

(NOTE: `{{welcome-message}}` component doesn't seem to render)

blueprints/module-unification-app/files/bower.json
rwjblue: We can (via `github:emberjs/ember.js`) but it requires npm@5 or yarn@1 I believe.  Seems good to test if you have time?

blueprints/module-unification-app/files/src/resolver.js
```diff
@@ -0,0 +1,13 @@
+import Resolver from 'ember-resolver/resolvers/glimmer-wrapper';
```
rwjblue: Changing this to `ember-resolver/resolvers/fallback` _should_ allow "classic" addons to continue to work within module unification apps (and therefore fix `{{welcome-page}}` without modification). Can you test to confirm?

GavinJoyce: Thanks, I've made the change and confirmed that this works

blueprints/module-unification-app/files/tests/helpers/resolver.js
```diff
@@ -0,0 +1,11 @@
+import Resolver from '../../src/resolver';
```
rwjblue: This file should be removed

blueprints/module-unification-app/files/tests/test-helper.js
```diff
@@ -0,0 +1,8 @@
+import resolver from './helpers/resolver';
+import {
+  setResolver
+} from 'ember-qunit';
+import { start } from 'ember-cli-qunit';
+
+setResolver(resolver);
```
rwjblue: This should be `setApplication` instead (see app blueprint for example)

blueprints/module-unification-app/files/config/environment.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-env node */
```
rwjblue: this comment should be removed

blueprints/module-unification-app/files/config/targets.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env node */
```
rwjblue: this comment is no longer needed with `overrides` config

blueprints/module-unification-app/files/ember-cli-build.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-env node */
```
rwjblue: 🔪

blueprints/module-unification-app/files/testem.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-env node */
```
rwjblue: ✂️

blueprints/module-unification-app/files/tests/test-helper.js
```diff
@@ -0,0 +1,8 @@
+import resolver from './helpers/resolver';
+import {
+  setResolver
```
rwjblue: `setApplication`

blueprints/module-unification-app/files/bower.json
GavinJoyce: I couldn't seem to get this to work. I'll add to this my list and come back to it in a future PR

GavinJoyce:
@rwjblue I've addressed your feedback in https://github.com/ember-cli/ember-cli/pull/7488/commits/f8157a5786a0fc3778d36465c7eb3c55fd61d255 (I'll squash when we're happy).

I've tested the output with and without yarn and all looks good:

`MODULE_UNIFICATION=true .ember new module-unification-blueprint-app-output --yarn`

https://travis-ci.org/GavinJoyce/module-unification-blueprint-app-output (🍏 )
https://github.com/GavinJoyce/module-unification-blueprint-app-output

`MODULE_UNIFICATION=true .ember new module-unification-blueprint-app-yarn-output --yarn`

https://travis-ci.org/GavinJoyce/module-unification-blueprint-app-yarn-output (🍏 )
https://github.com/GavinJoyce/module-unification-blueprint-app-yarn-output

ro0gr:
@GavinJoyce looks like "[blueprints/module-unification-app/files/yarn.lock](https://github.com/GavinJoyce/ember-cli/blob/f8157a5786a0fc3778d36465c7eb3c55fd61d255/blueprints/module-unification-app/files/yarn.lock)" added by mistake(couldn't find a way to inline comment in empty file)

ro0gr:
> Do you think we could refactor things a bit such that we only have to maintain the differences from the main app blueprint? I think that would dramatically reduce the total maintenance burden if we can swing it.

We already have a [nice example](https://github.com/ember-cli/ember-cli/blob/bfb153d703e28b608a8252c702e8d6a011a76393/blueprints/addon/index.js#L128) for sharing files between `app` and `addon` blueprints. 
I think if we share at least files root files(.eslintrc, .editorconfig, testem.js, etc) it would be a win.

I'll take a look in this direction and report back.

ro0gr:
I've done some experimentation to reduce amount files to maintain #7500
Implementation still can be improved but it looks like `addon` blueprint's approach works for MU blueprint as well.

blueprints/module-unification-app/index.js
```diff
@@ -0,0 +1,31 @@
...
+      namespace,
+      emberCLIVersion: require('../../package').version,
+      yarn: options.yarn,
+      welcome: options.welcome,
+    };
+  },
+
+  fileMapTokens(options) {
+    return {
+      __component__() { return options.locals.component; },
```
ro0gr: I can't see `component` defined in a `locals`. Do we need this?

GavinJoyce:
> I've done some experimentation to reduce amount of files to maintain #7500
Implementation still can be improved but it looks like addon blueprint's approach works for MU blueprint as well.

@ro0gr nice one, this will make sharing files across blueprints easier (we'll still need to change how we test the output, as you noted in your PR). I'm still not clear which is better, 1) maintaining two separate trees and living with duplication for a relatively short amount of time or 2) using your technique with `this.lookupBlueprint('app').files();` to share a common base. I think it would be good to continue to use 1) in the short term and wait for some data before making an informed decision to jump to 2).

ro0gr:
>  I'm still not clear which is better, 1) maintaining two separate trees and living with duplication for a relatively short amount of time or 2) using your technique with this.lookupBlueprint('app').files(); to share a common base

@GavinJoyce I'm not sure how short is the "short term" you refer to. But it's obvious that duplication of large amount of files rarely leads to a good consequences. It's quite complicated to sync changes in +24 files(that's an amount of files we drop if share app blueprint files).

I think even now we might have syncing issues. Here is the list of files that are different between `app` and `module-unification-app` blueprints:

- https://github.com/GavinJoyce/ember-cli/blob/9202add03f1fd398abae3cf54e73b225991ab583/blueprints/module-unification-app/files/.travis.yml
- https://github.com/GavinJoyce/ember-cli/blob/9202add03f1fd398abae3cf54e73b225991ab583/blueprints/module-unification-app/files/testem.js

I'm not sure if that is intentional but it's much more convenient to have only those parts which are actually different even in a short term.

So I would vote to consider sharing files at least because technically it's the same approach we already use for `addon` blueprint, it seems to work and I can't see any obvious drawbacks with this approach.

blueprints/module-unification-app/files/src/resolver.js
```diff
@@ -0,0 +1,13 @@
+import Resolver from 'ember-resolver/resolvers/fallback';
```
mixonic: I'd very much like to have the default generated app *not* use the fallback resolver. IMO new apps should default to the future/optimal mode, and if you want to use a legacy addon or `app/` directory you can opt-in to the fallback.

mixonic:
Landing this and let's continue to improve it. Both @GavinJoyce and @ro0gr are interested in keeping the iteration moving.

cibernox:
Probably this is a known issue, but once inside an app created with the new blueprint, running `ember g route foo` fails. 
It creates a couple files in `/app/routes` and then fails with `ENOENT: no such file or directory, open '<app-root>/app/router.js'`

rwjblue:
I believe we will need to update those blueprints in ember-source itself.

PR #7489, opened 2017-12-06
Fix regression with scoped package name mismatches

rwwagner90:
https://github.com/ember-cli/ember-cli/commit/69750c9b8a078900e163598542c295a518d1b49f introduced a regression where, if your package name is scoped in package.json and not scoped in index.js, `isDevelopingAddon` would now evaluate to `false`, when it used to evaluate to `true`. This caused the `addon` folder to not be linted anymore.

This fixes the regression and allows the old behavior to work again, but also issues a warning, saying you should sync the names up, per the suggestion of @rwjblue

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -35,6 +35,7 @@ const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 const addonProcessTree = require('../utilities/addon-process-tree');
 const semver = require('semver');
 const processModulesOnly = require('../broccoli/babel-process-modules-only');
+const npa = require('npm-package-arg');
 
 const BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS = Symbol('BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS');
 
@@ -333,7 +334,29 @@ let addonProto = {
   */
   isDevelopingAddon() {
     if (process.env.EMBER_ADDON_ENV === 'development' && (this.parent instanceof Project)) {
-      return this.parent.name() === this.name;
+      const parentName = this.parent.name();
+      // If the name in package.json and index.js match, we're definitely developing
+      if (parentName === this.name) {
+        return true;
+      }
+
+      // If package.json and index.js names do not match, we should check for scoped vs non-scoped
+      const parsed = npa(parentName);
+      let parsedName = parsed.name;
+      if (parsed.scope) {
+        parsedName = parsedName.replace(parsed.scope, '').slice(1);
+      }
+      if (parsedName === this.name) {
+        if (!this._hasWarnedForMismatchedNames) {
+          this.ui.writeWarnLine(
+            'Your names in package.json and index.js should match. ' +
+            `You currently have ${parsed.name} in package.json and ${this.name} in index.js.`);
+
+          this._hasWarnedForMismatchedNames = true;
+        }
+
+        return true;
+      }
     }
     return false;
   },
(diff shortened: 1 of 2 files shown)


lib/models/addon.js
```diff
@@ -331,7 +334,29 @@ let addonProto = {
...
+      }
+
+      // If package.json and index.js names do not match, we should check for scoped vs non-scoped
+      const parsed = npa(parentName);
+      let parsedName = parsed.name;
+      if (parsed.scope) {
+        parsedName = parsedName.replace(parsed.scope, '').slice(1);
+      }
+      if (parsedName === this.name) {
+        if (!alreadyWarnedAddonNames.includes(this.name)) {
```
rwjblue: I'm wondering if we need this check. The warning should only be emitted for addon authors, and only for their specific addon.  In most cases I would assume it would only be emitted once (because its somewhat odd to have multiple versions of the same addon in the addons own build), and even in the less likely chance that it does get emitted twice it seems not _too_ bad...

My concern with this sort of module scoped guard is that we have no way to reset it during the lifetime of the running node process. This will make testing difficult, and in some cases will hide the warning when it _should_ be emitted...

rwwagner90:
@rwjblue added a test. Please let me know if that is sufficient.

lib/models/addon.js
rwwagner90: So do you think we should remove the `includes` check here? I was getting it logged 8 times running in my addon, which could get noisy.

rwjblue: Ah, we probably check `isDevelopingAddon` many times for the same instance.

Lets remove the module scope array and keep a local `this._hasWarnedForMismatchedNames = true` that we use instead.  What do you think?

rwwagner90: That sounds fine. Will push up a change that does that.

rwwagner90: Changed to use your suggestion. Let me know if anything else needs tweaking.

lib/models/addon.js
```diff
@@ -331,7 +332,29 @@ let addonProto = {
...
+
+      // If package.json and index.js names do not match, we should check for scoped vs non-scoped
+      const parsed = npa(parentName);
+      let parsedName = parsed.name;
+      if (parsed.scope) {
+        parsedName = parsedName.replace(parsed.scope, '').slice(1);
+      }
+      if (parsedName === this.name) {
+        if (!this._hasWarnedForMismatchedNames) {
+          console.warn(
```
rwjblue: Can you swap this for `this.ui.writeWarnLine`? The UI model honors our command line flags like `--silent`..

rwjblue:
Also, can you confirm which version introduced the regression? I believe it was 2.18.0-beta.1, right? If that is the case, can you update the PR target to be `beta` and rebase (to drop out the additional commits other than the one you are adding here)?

lib/models/addon.js
rwwagner90: Sure thing, swapping now.

rwwagner90:
This was introduced in 2.16, so I suppose we need to get it in the 2.16 LTS, beta, and master? Not sure.

rwwagner90:
@rwjblue any idea why travis is failing on not having `this.ui.writeWarnLine`? https://travis-ci.org/ember-cli/ember-cli/jobs/312999293#L1318

rwwagner90:
Fixed the tests 👍

rwwagner90:
@rwjblue sure thing, I just squashed. I tend to use the squash button on GitHub, as it saves me the steps. Would we be able to enable that option for the future here perhaps?

rwjblue:
In theory yes, but it’s a bit heavy handed. I don’t think we want “only one commit” (which is what the GH “squash and merge” feature does), we want commits for each specific “thing”. It’s a tricky one to balance, thanks for your patience and work on this!

PR #7490, opened 2017-12-06
Module Unification Addons 

rwjblue:
## Todo: 
- [x] Dry up lib/broccoli/ember-addon.js (copied from lib/broccoli/ember-app.js)
- [x] Process styles (in some way) from src/ in addon (models/addon.js)
- [x] fix bad error message for `Missing template processor` when missing ember-cli-htmlbars in dependecies i.e. it exists in dev-dependencies
- [ ] ~throw an error (or do something) when an addon can't decide if it's mu or classic (i.e. has src/ **and** addon/ or app/ folders)~ @rwjblue and @mixonic think having all three dirs is a valid case. We're going to leave the current ambiguous behavior.
- [x] ensure `src` processing is lazy (per https://github.com/ember-cli/ember-cli/pull/7501)

## Tests:
- [x] addons with src/ folder in classic apps

## Future TODOS:

- [ ] broccoli-module-unification-reexporter [edge cases?](https://github.com/stonecircle/broccoli-module-unification-reexporter/blob/538d923e7e8a8f134369d8b0b8b9bf1473e99e49/test/index.js#L83)


lib/models/addon.js
```diff
@@ -699,6 +702,24 @@ let addonProto = {
     @return {Tree} App file tree
   */
   treeForApp(tree) {
+    if (!experiments.MODULE_UNIFICATION || this.project.isModuleUnification()) {
```
iezer: pretty sure this should be `|| !this.project.isModuleUnification()) ` with a not (!)

rwjblue: discussed in slack, I think we actually need the following guard:

```js
if (!experiments.MODULE_UNIFICATION) {
  return tree;
} else if (project.isModuleUnification() && this.isModuleUnification()) {
  return null;
}
```

And add a new `isModuleUnification` method to the addon base class...

tests/fixtures/addon/kitchen-sink-mu/index.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  name: 'basic-thing',
```
iezer: @mixonic I wonder if these tests are failing because the name doesn't match the folder name. In the other examples they do. 
PS how do I run just these tests locally, and even better is it possible to run them in the browser?

tests/fixtures/addon/kitchen-sink-mu/src/styles/app.css
```diff
@@ -0,0 +1 @@
+/* addon/styles/app.css is present */
```
rwjblue: This is not the correct location for styles in MU

bartocc:
@rwjblue is this PR supposed to let addon authors write addons with the MU project structure ?

rwjblue:
@bartocc - Yes, it is the first baby step towards that goal.

Issue #7429, opened 2017-11-09
Turning on Instrumentation using ember-cli-dotenv


patocallaghan:
I'd like to enable the ember-cli instrumentation using the `.env` file from ember-cli-dotenv but it appears doing it that way sets the environment variable too late in the process. The `init` is set up here https://github.com/ember-cli/ember-cli/blob/a4734540b3889c7f6ef304fe6a81dd356f77bfe4/lib/cli/index.js#L6-L14 before the variable has been set. If I run `ember server` I get the following error

```
➜  instrumentation-dotenv git:(master) ember server
No such instrumentation "init"
Error: No such instrumentation "init"
    at Instrumentation._instrumentationFor (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/ember-cli/lib/models/instrumentation.js:171:13)
    at Instrumentation.stopAndReport (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/ember-cli/lib/models/instrumentation.js:227:22)
    at Promise.resolve.then (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/ember-cli/lib/cli/cli.js:153:25)
    at tryCatch (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/rsvp/dist/rsvp.js:525:12)
    at invokeCallback (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/rsvp/dist/rsvp.js:538:13)
    at /Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/rsvp/dist/rsvp.js:606:14
    at flush (/Users/patocallaghan/src/me/instrumentation-dotenv/node_modules/rsvp/dist/rsvp.js:2415:5)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
```

My use case is I'd like to collect metrics, i.e. `initial` and `rebuild` times, from my team but would rather they didn't have to:

1. set the `EMBER_CLI_INSTRUMENTATION` var themselves
2. provide a script which they'd use instead of `ember server` which sets the var and calls `ember server`.

I'd rather it configured by default i.e. turned on by some sort of config file.

I have a replication repo of the issue here https://github.com/patocallaghan/instrumentation-dotenv

[...]

PR #7491, opened 2017-12-06
Allow instrumentation to be turned on using .ember-cli

patocallaghan:
Addresses https://github.com/ember-cli/ember-cli/issues/7429

This is a WIP so hoping to get some feedback. I'm open to changing the namings + approach if required. What I initially thought to be straightforward has turned into somewhat of a rabbit-hole 😄 

* abstracted retrieving the `Yam` config for the project out to `utilities/get-config`.
* Introducing the config into `utiltities/instrumentation` meant needing to move some things around as there are circular dependencies between Instrumentation => Project => Instrumentation. Otherwise it would blow up on init.
* Testing is pretty tricky and not fully 🍏 yet. Due to:
  1. Finding an issue with `yam` whereby a falsey value returns `undefined`. I have a fix here https://github.com/twokul/yam/pull/16
  2. The `_enableFSMonitorIfInstrumentationEnabled ` tests in `models/instrumentation-test.js` are not atomic. Using `heimdall-fs-monitor` to wrap around `fs.statSync` doesn't clean up after itself so affects later tests. https://github.com/ember-cli/ember-cli/blob/d9c015974436ab0b80f552ed7689e968e3d50f91/tests/unit/models/instrumentation-test.js#L36-L83. In particular it's because 1) we don't reassign `fs.statSync` back to it's initial state after each test and 2) calling `FSMonitor()` sets some module-level vars in `heimdall-fs-monitor` (`isFirstInstance`, `isMonitorRegistrant`)  which means trying to set up `FSMonitor()` in a following test doesn't fully work. It doesn't re-"attach" the monitor functions.

Any feedback on the approach in general is appreciated. Also any ideas how to workaround around the `heimdall-fs-monitor` testing issue. I could submit a PR to fix that repo too if useful?

### DIFF
--- a/lib/models/instrumentation.js
+++ b/lib/models/instrumentation.js
@@ -9,9 +9,9 @@ const logger = require('heimdalljs-logger')('ember-cli:instrumentation');
 let vizEnabled = utilsInstrumentation.vizEnabled;
 let instrumentationEnabled = utilsInstrumentation.instrumentationEnabled;
 
-function _enableFSMonitorIfInstrumentationEnabled() {
+function _enableFSMonitorIfInstrumentationEnabled(config) {
   let monitor;
-  if (instrumentationEnabled()) {
+  if (instrumentationEnabled(config)) {
     const FSMonitor = require('heimdalljs-fs-monitor');
     monitor = new FSMonitor();
     monitor.start();
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -14,7 +14,6 @@ const logger = require('heimdalljs-logger')('ember-cli:project');
 const versionUtils = require('../utilities/version-utils');
 const emberCLIVersion = versionUtils.emberCLIVersion;
 const findAddonByName = require('../utilities/find-addon-by-name');
-const Instrumentation = require('./instrumentation');
 const heimdall = require('heimdalljs');
 
 let processCwd = process.cwd();
@@ -717,6 +716,10 @@ function ensureInstrumentation(cli, ui) {
     return cli.instrumentation;
   }
 
+  // Instrumentation `require` needs to occur inline due to circular dependencies between
+  // Instrumentation => getConfig => Project => Instrumentation. getConfig is used in Project to
+  // get the project root.
+  let Instrumentation = require('./instrumentation');
   // created without a `cli` object (possibly from deprecated `Brocfile.js`)
   return new Instrumentation({ ui, initInstrumentation: null });
 }
--- a/lib/utilities/get-config.js
+++ b/lib/utilities/get-config.js
@@ -1,0 +1,23 @@
+'use strict';
+
+let config;
+let Yam = require('yam');
+let Project = require('../models/project');
+
+function generateConfig() {
+  return new Yam('ember-cli', {
+    primary: Project.getProjectRoot(),
+  });
+}
+
+module.exports = function getConfig(configOverride) {
+  if (configOverride) {
+    return configOverride;
+  }
+
+  if (config === undefined) {
+    config = generateConfig();
+  }
+
+  return config;
+};
--- a/lib/utilities/instrumentation.js
+++ b/lib/utilities/instrumentation.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const getConfig = require('./get-config');
+
 let _hasWarnedLegacyViz = false;
 
 function vizEnabled() {
@@ -15,8 +17,13 @@ function vizEnabled() {
   return isEnabled || isLegacyEnabled;
 }
 
-function instrumentationEnabled() {
-  return vizEnabled() || process.env.EMBER_CLI_INSTRUMENTATION === '1';
+function isInstrumentationConfigEnabled(configOverride) {
+  let config = getConfig(configOverride);
+  return !!config.get('enableInstrumentation');
+}
+
+function instrumentationEnabled(config) {
+  return vizEnabled() || process.env.EMBER_CLI_INSTRUMENTATION === '1' || isInstrumentationConfigEnabled(config);
 }
 
 
--- a/tests/unit/models/instrumentation-test.js
+++ b/tests/unit/models/instrumentation-test.js
@@ -9,6 +9,7 @@ const path = require('path');
 const fse = require('fs-extra');
 const MockUI = require('console-ui/mock');
 const RSVP = require('rsvp');
+const Yam = require('yam');
 
 const MockProject = require('../../helpers/mock-project');
 const mkTmpDirIn = require('../../../lib/utilities/mk-tmp-dir-in');
@@ -39,9 +40,11 @@ describe('models/instrumentation.js', function() {
     beforeEach(function() {
       expect(!!process.env.BROCCOLI_VIZ).to.eql(false);
       expect(!!process.env.EMBER_CLI_INSTRUMENTATION).to.eql(false);
+      expect(fs.statSync).to.equal(originalStatSync);
     });
 
     afterEach(function() {
+      fs.statSync = originalStatSync;
       td.reset();
     });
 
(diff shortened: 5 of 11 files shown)


lib/models/instrumentation.js
```diff
@@ -9,9 +9,9 @@ const logger = require('heimdalljs-logger')('ember-cli:instrumentation');
 let vizEnabled = utilsInstrumentation.vizEnabled;
 let instrumentationEnabled = utilsInstrumentation.instrumentationEnabled;
 
-function _enableFSMonitorIfInstrumentationEnabled() {
+function _enableFSMonitorIfInstrumentationEnabled(config) {
```
patocallaghan: Need to be able to pass the `config` in here so we can "mock" it in tests? Is this a smell or okay to do?

lib/models/project.js
```diff
@@ -750,6 +749,7 @@ function ensureInstrumentation(cli, ui) {
     return cli.instrumentation;
   }
 
+  let Instrumentation = require('./instrumentation');
```
patocallaghan: I needed to move the Instrumentation require inline otherwise it blew up on init due to circular dependencies between Instrumentation => getConfig => Project => Instrumentation

lib/utilities/get-config.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+let Yam = require('yam');
+let Project = require('../models/project');
+let config = new Yam('ember-cli', {
+  primary: Project.getProjectRoot(),
+});
+
+module.exports = function getConfig(mockedConfig) {
```
patocallaghan: allow passing in a mocked config to support current behavior in some tests, e.g. `options.Yam` in `cli/index-test.js`

tests/unit/models/instrumentation-test.js
```diff
@@ -39,9 +40,11 @@ describe('models/instrumentation.js', function() {
     beforeEach(function() {
       expect(!!process.env.BROCCOLI_VIZ).to.eql(false);
       expect(!!process.env.EMBER_CLI_INSTRUMENTATION).to.eql(false);
+      expect(fs.statSync).to.equal(originalStatSync);
```
patocallaghan: Assert that we always start with clean slate on `fs.statSync`

tests/unit/models/instrumentation-test.js
```diff
@@ -39,9 +40,11 @@ describe('models/instrumentation.js', function() {
...
     });
 
     afterEach(function() {
+      fs.statSync = originalStatSync;
```
patocallaghan: Reset `fs.statSync` to its  original state

tests/unit/models/instrumentation-test.js
```diff
@@ -72,6 +75,35 @@ describe('models/instrumentation.js', function() {
     it('if instrumentation is enabled, monitor', function() {
       process.env.EMBER_CLI_INSTRUMENTATION = '1';
       let monitor = Instrumentation._enableFSMonitorIfInstrumentationEnabled();
+      try {
+        expect(fs.statSync).to.not.equal(originalStatSync);
```
patocallaghan: This test fails due to `heimdall-fs-monitor` not fully cleaning up after itself. Some module level variables are set in a previous test which cause it be `isEnabled` already. This means when calling `FSMonitor()` a second time, it doesn't re-"attach" the monitor functions around the `fs` functions.

hjdivad: okay it looks like this problem in `heimdall-fs-monitor` was actually fixed a while ago.  I've released 0.2.0 to deal with `stop` not cleaning up properly.

hjdivad:
@patocallaghan updated heimdall-fs-monitor on your branch; `stop` should clean up better now, LMK if you're still seeing issues here.

patocallaghan:
Appreciate the feedback @hjdivad. Will fix up per your comments.

tests/unit/settings-file/leek-options-test.js
```diff
@@ -6,22 +6,17 @@ const Yam = require('yam');
 const cliEntry = require('../../../lib/cli');
 
 describe('.ember-cli leek options', function() {
-  let settings;
   let passedOptions;
 
   before(function() {
-    settings = new Yam('ember-cli', {
+    let mockedYam = new Yam('ember-cli', {
```
patocallaghan: Needed to slightly refactor this test to get it to pass.

lib/utilities/get-config.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+let Yam = require('yam');
+let Project = require('../models/project');
+
+module.exports = function getConfig(configOverride) {
+  if (configOverride) {
+    return configOverride;
+  }
+  return new Yam('ember-cli', {
```
patocallaghan: I needed to bring the `new Yam()` into the `getConfig` function because it was failing the `tests/acceptance/pods-destroy-test.js` tests otherwise. Preferably the `new Yam` would happen at the module scope but in the pod-destroy tests the config wasn't being re-calculated after modifying the `.ember-cli` file.

patocallaghan:
@hjdivad made some changes per your feedback. Also tests should be 🍏 now.

patocallaghan:
@hjdivad I just did rebase against current master and the branch is still 🍏

lib/models/project.js
```diff
@@ -720,6 +719,8 @@ function ensureInstrumentation(cli, ui) {
     return cli.instrumentation;
   }
 
+  // Instrumentation `require` needs to occur inline due to circular dependencies between Instrumentation => getConfig => Project => Instrumentation. getConfig is used in Project to get the project root.
```
twokul: sorry for nitpicking but could you wrap break this comment into separate lines, around 80 characters long?

lib/utilities/instrumentation.js
```diff
@@ -15,8 +17,13 @@ function vizEnabled() {
   return isEnabled || isLegacyEnabled;
 }
 
-function instrumentationEnabled() {
-  return vizEnabled() || process.env.EMBER_CLI_INSTRUMENTATION === '1';
+function isConfigEnabled(configOverride) {
```
twokul: maybe better to name it `isInstrumetationConfigEnabled`?

patocallaghan:
No problem @twokul, I've addressed yours and @hjdivad's comments. I've also resolved some merge conflicts with the `yarn.lock` on master.

lib/utilities/get-config.js
stefanpenner: should we load a new Yam on each invocation of getConfig, or memoize?

patocallaghan: @stefanpenner I'd initially tried have the `new Yam` line at the module level but had to refactor because tests which relied on mocking the config failed (pod tests I believe). I can't remember the specific reasoning but I'll have another go and see.

hjdivad:
thanks @patocallaghan!

patocallaghan:
@stefanpenner @hjdivad I've just pushed a little spike of memoizing the `Yam` config in `getConfig` by moving it back to the module level. By doing that it fails a bunch of pod tests which rely on having the .ember-cli file having `usePods: true`. For example this test here: https://github.com/ember-cli/ember-cli/blob/6fe4412debab048641719e6403abc6e984237073/tests/acceptance/pods-destroy-test.js#L133-L142

When the test is initially run the `get-config` module is loaded and the `config` is generated but then when we get to the `assertDestroyAfterGenerateWithUsePods` we augment the `config` with the `usePods: true` setting:

https://github.com/ember-cli/ember-cli/blob/6fe4412debab048641719e6403abc6e984237073/tests/acceptance/pods-destroy-test.js#L94

But as we've memoized the config on module load we don't have a Yam with `usePods` set.

I've added some sort of solution but to be honest it feels really wrong. When in test mode it generates a new Yam but when not it uses the memoized one. There's a precedent for this elsewhere in the tests but it still doesn't make it right 😄 

If anyone has any suggestions for a better approach I'm all ears :)

patocallaghan:
@stefanpenner @hjdivad @twokul any thoughts re my last comment here? Happy to rethink the testing if it's not acceptable.

FYI I rebased against master and now the branch is failing. It appears master is also failing with the same error.

lib/utilities/get-config.js
```diff
@@ -3,11 +3,22 @@
...
-  }
+function generateConfig() {
   return new Yam('ember-cli', {
     primary: Project.getProjectRoot(),
   });
+}
+
+let config = generateConfig();
+
+module.exports = function getConfig(configOverride, shouldRegenerateConfig) {
```
hjdivad: instead of adding another parameter for this, what do you think about exporting a separate reset function for tests?

in es6 this would be a non default export, but in this case it should be fine to hang it off the default export.  eg


```js

let config;

module.exports = function getConfig(configOverride) {
  if (configOverride) {
    return configOverride;
  }

  if (config === undefined) {
    config = generateConfig();
  }

  return config;
}

// private; used in testing
module.exports._reset = function() {
  config = undefined;
}
```

patocallaghan: I'll give that a shot. Thanks 👍

patocallaghan: @hjdivad it turns out the problematic tests were actually removed in a recent commit 😄 So there's no longer a need to jump through hoops to get the suite to pass. Tests are 🍏 locally and the config memoization works.

hjdivad: oh nice; so if you rebase and push we should expect the PR to be green then

stefanpenner:
is there a downside to not memoizing?

lib/utilities/get-config.js
patocallaghan: it looks like master is still 🔴

patocallaghan:
@stefanpenner I guess the downside would be that we're creating new `Yam` objects every time we call `getConfig` even though it doesn't change over the lifecycle of the command (or can it?)

patocallaghan:
@hjdivad and @stefanpenner I've rebased again and the PR is now 🍏

hjdivad:
thanks @patocallaghan!

GavinJoyce:
This appears to break some generators which use `"usePods": true`: https://github.com/emberjs/ember.js/pull/16465/files#r179132445

lib/utilities/get-config.js
GavinJoyce: I think these tests we just moved into ember-source so that is why we're seeing the failures over there now when upgrading ember-cli.

@patocallaghan if it suits, let's have a chat about this tomorrow

PR #7500, opened 2017-12-12
Put MU blueprint on a diet

ro0gr:
This pr tries to reduce amount of MU blueprint files introduced in #7488 . 
Here used the same approach as in the [addon blueprint](https://github.com/ember-cli/ember-cli/blob/bfb153d703e28b608a8252c702e8d6a011a76393/blueprints/addon/index.js#L128).

Potentially we could get rid of "blueprints/module-unification-app/files/package.json" as well but not for now.

Tests are failing due to the usage of [confirmBlueprintForDir](https://github.com/ember-cli/ember-cli/compare/master...ro0gr:mu-blueprint-on-a-diet?expand=1#diff-976dc93f203f7e6cdd04ae6008f7d8b6R78) which doesn't seem as a suitable tool for checking composed blueprints. 

I think the test has to be changed to check generator output against MU fixture. Does it sound reasonable? Would appreciate any pointers for the test improvements.


blueprints/module-unification-app/index.js
```diff
@@ -0,0 +1,105 @@
...
+const stringUtil = require('ember-cli-string-utils');
+
+const FileInfo = require('../../lib/models/file-info');
+
+module.exports = {
+  description: 'Generates an Ember application with a module unification layout.',
+
+  filesToRemove: [],
+
+  ignoredFilePaths: [
```
ro0gr: I feel like `ignoredFilePaths` wants to be a part of `fileMap`, each key with a `null` value. Unfortunately  `Blueprint` model exects `fileMap` values to be string only.

mixonic:
@ro0gr I'm closing this since at the moment it contains https://github.com/ember-cli/ember-cli/pull/7488 in addition to your changes to improve re-use. Please feel free to re-open or open a new PR.

PR #7501, opened 2017-12-13
add delayed transpilation

kellyselden:
I would like some feedback. Things to note:

* we have to detect files that are already AMD because certain addons do their own transpilation.
* we have to detect addons in legacy 'modules/' dir and undo to transpile correctly.
* removed `treeForAddon`/`compileModules` warning, because it is now the desired behavior.
* removed `compileTemplates` no template processor warning, because it is now the desired behavior.

Sample console output:

```
DEPRECATION: Addon "@html-next/vertical-collection" (found at "C:\Users\kelly\Desktop\my-app\node_modules\@html-next\vertical-collection") is manually placing files in the legacy "modules" folder. Support for this will be removed in a future version.
DEPRECATION: Addon "ember-data" (found at "C:\Users\kelly\Desktop\my-app\node_modules\ember-data") is manually placing files in the legacy "modules" folder. Support for this will be removed in a future version.
```

Left to do:
* get fork of ember-cli-babel merged https://github.com/babel/ember-cli-babel/pull/204
* get #7634 merged
* disable for babel 5
* put behind feature flag


rtablada:
@kellyselden should the `AMD Excluder` be extracted into `broccoli-funnel-amd-excluder`? I'm not sure if that could potentially be helpful as a separate package rather than as part of Ember CLI?

Though I know that adds a maintenance cost too.

lib/broccoli/amd-excluder.js
```diff
@@ -0,0 +1,40 @@
...
+      directories: false,
+      globs: [`**/*.js`],
+    });
+
+    for (let file of files) {
+      let inputFilePath = path.join(inputPath, file);
+      let source = fs.readFileSync(inputFilePath, 'utf8');
+
+      // ember-data and others compile their own source
+      if (source.indexOf('define(') === 0) {
```
rtablada: Should be fine, but I'm not 100 about potential edgecases. It would likely not be compatible JS anyway so this should be good enough to figure out if something is AMD.

On the other hand, what if the module is exported for AMD but doesn't `define` immediately.

kellyselden:
@rtablada I am in favor of extracting it.

rtablada:
I'd like to see what documentation (if any) is needed for addon authors to make sure their addons work with this.

Bonus points for docs on how to utilize these changes in addons.

kellyselden:
@rtablada The only thing addons have to worry about is the first bullet point

> addons no longer receive preprocessTree and postprocessTree hooks.

lib/broccoli/amd-excluder.js
kellyselden: There are a lot of potential edge cases, but in reality we only have to worry about how babel formats the AMD code, which this should cover completely?

lib/broccoli/amd-excluder.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const Funnel = require('broccoli-funnel');
+const walkSync = require('walk-sync');
+
+class AmdExcluder extends Funnel {
```
Turbo87: can you use https://github.com/broccolijs/broccoli-test-helper to write a few tests for this class?

lib/broccoli/amd-excluder.js
```diff
@@ -0,0 +1,40 @@
...
+    this.options = options;
+  }
+
+  build() {
+    let inputPath = this.inputPaths[0];
+
+    let files = walkSync(inputPath, {
+      directories: false,
+      globs: [`**/*.js`],
+    });
```
Turbo87: does `build()` have to be sync? otherwise it might make sense to read and process several files in parallel 🤔

lib/broccoli/amd-excluder.js
```diff
@@ -0,0 +1,40 @@
...
+
+    for (let file of files) {
+      let inputFilePath = path.join(inputPath, file);
+      let source = fs.readFileSync(inputFilePath, 'utf8');
+
+      // ember-data and others compile their own source
+      if (source.indexOf('define(') === 0) {
+        this.exclude.push(file);
```
Turbo87: will this fill up `this.exclude` with duplicate entries on rebuilds? we might have to reset `this.exclude = [];` at the beginning of `build()` 🤔

lib/broccoli/amd-excluder.js
Turbo87: a little bit of documentation on this class might also be nice. e.g. answering the question that @rtablada raised about this being specialized to detect AMD modules compiled by Babel (5? 6? 7?)

lib/broccoli/ember-app.js
```diff
@@ -634,6 +635,39 @@ class EmberApp {
...
+
+  _compileAddonJs(tree) {
+    let precompiledSource = new AmdExcluder(tree, {
+      annotation: `_compileAddonJs`,
+    });
+
+    precompiledSource = preprocessJs(precompiledSource, '/', '/', {
+      annotation: `_compileAddonJs`,
+      registry: this.registry,
+    });
```
Turbo87: I'm wondering if in terms of performance it would be better to teach `preprocessJs()` about the AMD exclusion rule 🤔    but then again that is something that can probably be refactored later...

kellyselden:
wooo tests finally green.

lib/broccoli/amd-excluder.js
kellyselden: Doesn't have to be. Which package should I use for async walking?

lib/broccoli/amd-excluder.js
kellyselden: added docs

lib/broccoli/amd-excluder.js
Turbo87: I guess `walkSync` is okay, but the loop afterwards could possibly be concurrent

kellyselden: Changed to async readFile. Was this what you were thinking?

tests/unit/models/addon-test.js
```diff
@@ -631,28 +631,28 @@ describe('models/addon.js', function() {
...
-
-      expect(() => {
-        addon.compileTemplates();
-      }).to.throw(
-        `Addon templates were detected, but there are no template compilers registered for \`${addon.name}\`. ` +
-        `Please make sure your template precompiler (commonly \`ember-cli-htmlbars\`) is listed in \`dependencies\` ` +
-        `(NOT \`devDependencies\`) in \`${addon.name}\`'s \`package.json\`.`
-      );
-    });
+    // it('should throw a useful error if a template compiler is not present -- non-pods', function() {
```
twokul: needs to be deleted

lib/models/addon.js
```diff
@@ -1009,18 +1011,23 @@ let addonProto = {
     this._requireBuildPackages();
 
     if (this.shouldCompileTemplates()) {
-      if (!registryHasPreprocessor(this.registry, 'template')) {
-        throw new SilentError(`Addon templates were detected, but there are no template compilers registered for \`${this.name}\`. ` +
-          `Please make sure your template precompiler (commonly \`ember-cli-htmlbars\`) is listed in \`dependencies\` ` +
-          `(NOT \`devDependencies\`) in \`${this.name}\`'s \`package.json\`.`);
-      }
+      // if (!registryHasPreprocessor(this.registry, 'template')) {
```
twokul: needs to be deleted

lib/broccoli/ember-app.js
```diff
@@ -1079,7 +1116,9 @@ class EmberApp {
         annotation: 'TreeMerger: `addon/` trees',
       });
 
-      this._cachedAddonTree = new Funnel(combinedAddonTree, {
+      let compiledAddonTree = this._compileAddonTree(combinedAddonTree);
```
twokul: Can we add some broccoli-debug trees for before and after?  I'm mostly thinking this will be super useful for seeing where some transpilation issues _might_ be coming from...

lib/broccoli/ember-app.js
```diff
@@ -617,6 +619,41 @@ class EmberApp {
     }, []);
   }
 
+  _compileAddonTemplates(tree) {
+    if (registryHasPreprocessor(this.registry, 'template')) {
```
twokul: Can we actually get here?

lib/broccoli/amd-excluder.js
```diff
@@ -0,0 +1,45 @@
...
+    let files = walkSync(inputPath, {
+      directories: false,
+      globs: [`**/*.js`],
+    });
+
+    return Promise.all(files.map(file => {
+      let inputFilePath = path.join(inputPath, file);
+      return fs.readFile(inputFilePath, 'utf8').then(source => {
+        if (source.indexOf('define(') === 0) {
+          this.exclude.push(file);
```
twokul: Lets add a warning here...

twokul: `this.exclude` needs to be reset here (and we probably need to clear whatever funnel caches have been build up)

lib/broccoli/ember-app.js
```diff
@@ -617,6 +619,41 @@ class EmberApp {
...
+        annotation: `_compileAddonTemplates`,
+        registry: this.registry,
+      });
+    }
+
+    return tree;
+  }
+
+  _compileAddonJs(tree) {
+    let precompiledSource = new AmdExcluder(tree, {
```
twokul: lets add a flag to opt-in or out of this tree as an option to `new EmberApp`, the default value should be `true` (e.g. use the amd excluder) but ultimately we should try to change it to `false` once a good number of popular addons have been updated...

lib/models/addon.js
```diff
@@ -1178,7 +1173,28 @@ let addonProto = {
...
+
+    let original = this.options[emberCLIBabelConfigKey];
+
+    // the app will handle transpilation after it tree-shakes
+    // do it here instead of the constructor because
+    // ember-data and others do their own compilation in their
+    // treeForAddon without calling super
+    // they need the original params preserved because they call
+    // babel themselves and expect compilation the old way
+    this.options[emberCLIBabelConfigKey] = Object.assign({}, original, {
```
twokul: This really should be moved back to the constructor if at all possible, setting it here means that others that depend on it earlier just don't have access to the values when they expect...

lib/broccoli/ember-app.js
kellyselden: removed

lib/broccoli/ember-app.js
kellyselden: I emulated the style from the assembler. @twokul you may want to verify.

tests/unit/models/addon-test.js
kellyselden: deleted

lib/models/addon.js
kellyselden: deleted

lib/models/addon.js
kellyselden: I remember why I placed it here instead of the constructor. Having it in the constructor means I need to handle the legacy /modules using broccoli-module-normalizer. Making that change now.

lib/broccoli/ember-app.js
kellyselden: I remember why I added it. Without it we get errors in the styles tests https://travis-ci.org/ember-cli/ember-cli/jobs/337422087#L805-L813 missing template preprocessor. I will add more mocking to the styles tests.

kellyselden: added

lib/broccoli/amd-excluder.js
kellyselden: exclude is reset and it prints to console. the code has been moved to https://github.com/ember-cli/broccoli-amd-funnel. I also added tests.

twokul:
@kellyselden needs a rebase

lib/models/addon.js
```diff
@@ -223,8 +223,11 @@ let addonProto = {
 
     let emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
     this.__originalOptions[emberCLIBabelConfigKey] = this.options[emberCLIBabelConfigKey] = defaultsDeep(this.options[emberCLIBabelConfigKey], {
-      compileModules: true,
+      compileModules: false,
+      disablePresetEnv: true,
```
rwjblue: Can you also add `disableDebugTooling: true` to this list of options?

lib/models/addon.js
```diff
@@ -1061,18 +1063,6 @@ let addonProto = {
       this.options.babel = this.__originalOptions.babel;
     }
 
-    let emberCLIBabelConfigKey = this._emberCLIBabelConfigKey();
-    if (!this.options[emberCLIBabelConfigKey] || !this.options[emberCLIBabelConfigKey].compileModules) {
```
rwjblue: Lets bring this back, and change it around a bit to reset the internal `this.options` state to match what we setup in the constructor.

package.json
```diff
@@ -51,6 +52,7 @@
     "broccoli-funnel-reducer": "^1.0.0",
     "broccoli-merge-trees": "^2.0.0",
     "broccoli-middleware": "^1.0.0",
+    "broccoli-module-normalizer": "kellyselden/broccoli-module-normalizer#both-types",
```
rwjblue: Just noting, we need to land and release this....

tests/.eslintrc.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
 
     /*** mocha ***/
 
-    'mocha/no-exclusive-tests': 'error',
+    'mocha/no-exclusive-tests': 'off',
```
rwjblue: O_o

lib/models/addon.js
kellyselden: done

lib/models/addon.js
kellyselden: done

tests/.eslintrc.js
kellyselden: reverted

lib/models/addon.js
kellyselden: done

lib/broccoli/ember-app.js
kellyselden: done

lib/broccoli/ember-app.js
```diff
@@ -1081,7 +1119,54 @@ class EmberApp {
...
+
+      tree = this._debugTree(tree, `precompiledAddonTree:${addon.name}`);
+
+      if (!this.options.moduleNormalizerDisabled) {
+        // move legacy /modules/addon to /addon
+        let hasAlreadyPrintedModuleDeprecation;
+        tree = new ModuleNormalizer(tree, {
+          callback: () => {
+            if (!hasAlreadyPrintedModuleDeprecation) {
+              this.project.ui.writeDeprecateLine(`Addon "${addon.name}" is manually placing files in the legacy "modules" folder. Support for this will be removed in a future version.`);
```
kellyselden: @rwjblue Are you sure you want addon.root in this message? I don't see what benefit that gives you over addon.name alone, since the node_modules path will be flattened, I don't think you can tell either way which addon is bringing it in if not your app.

rwjblue: Without addon root you cannot detect if the message is caused by a nested addon. 

Imagine you have multiple versions of an addon. Some do not trigger the warning, but some do. This hypothetical addon is depended on heavily by many other addons. Based on the addon name alone, how can someone determine where to file an issue / PR? In this scenario the addon who’s name is printed is _already_ fixed! The actual offenders are the other addons that are including the old and deprecated version. AFAICT, this is only possible to detect via addon.root....

kellyselden:
I'm seeing a 30% slowdown in my testing. Not sure why.

twokul:
so far, both @kellyselden and I, are seeing anywhere between 15-30% slowdown (depending on the application we run benchmarks against).

_I ran tests against [`travis-web`](https://github.com/travis-ci/travis-web)_.

here's what a flame chart w/o delayed transpilation looks like:

screen shot 2018-02-17 at 4 20 42 pm

and here's a chart w/ delayed transpilation:

screen shot 2018-02-17 at 4 10 24 pm

Note, that we started running Babel twice (essentially unnecessary work). also, order in which we process add-ons have changed (from `active-model-adapter`, `ember-cli-mirage`, `ember-data`, `ember-modal-dialog`, `ember-moment` to `ember-data`, `ember-cli-mirage`, `ember-moment`). A thing to note is that some of the add-ons "disappeared" from flame graph and a wide gap on "unaccounted" work appeared. It's either [`ember-cli-inline-images`](https://github.com/drogus/ember-cli-inline-images) or GC kicks in.

We can optimize and don't run Babel twice but there's a gap we're still seeing:

screen shot 2018-02-17 at 4 10 34 pm

kellyselden:
All this new code is guarded around the feature flag DELAYED_TRANSPILATION. Because of this, and because @rwjblue and @twokul have reviewed this multiple times with no remaining problems, I think it is safe to merge.

stefanpenner:
I believe this PR is breaking things, specifically we seem to trying to transpile handlebars templates that exist `<addon>/src/<somefolder>/templates/model.hbs`

Issue #7511, opened 2017-12-19
ember new ignores --yarn


gossi:
Running `ember new foobar --yarn` ignores the `--yarn` argument. Turns out this if-block is false:

https://github.com/ember-cli/ember-cli/blob/80b4a7a5525355f405266d9660cf41915efbaa49/lib/tasks/npm-task.js#L104-L114

Also see this PR: https://github.com/ember-cli/ember-cli/pull/7481

My System:

```
$ ember version --verbose && npm --version && yarn --version
ember-cli: 2.17.1
http_parser: 2.7.0
node: 8.9.3
v8: 6.1.534.48
uv: 1.15.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 57
nghttp2: 1.25.0
openssl: 1.0.2n
icu: 60.2
unicode: 10.0
cldr: 32.0.1
tz: 2017c
os: darwin x64
5.6.0
1.3.2
```

lennyburdette:
I see my mistake: I checked that `useYarn` is never passed to `NpmInstall#run` — and it’s never passed directly, only through the `Command#runTask` abstraction. We should probably pick one or the other: either `useYarn` is passed to the task’s constructor or it’s passed to `run`, but not both. I’m guessing we’ll need to go with the latter since the constructor isn’t available when using `runTask`. Alternatively, we could add `useYarn` to `Command#_env` so that it’s passed to every task’s constructor. 

I’m computer-less through the holidays—sorry for not being able to address this sooner!

Turbo87:
This got fixed #7516

PR #7516, opened 2017-12-22
Fix `ember new --yarn` not using yarn

Turbo87:
Resolves https://github.com/ember-cli/ember-cli/issues/7511

/cc @PieterJanVdb @gossi @lennyburdette

### DIFF
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -25,7 +25,6 @@ class AddonInstallTask extends Task {
       ui: this.ui,
       analytics: this.analytics,
       project: this.project,
-      useYarn: commandOptions.yarn,
     });
 
     let blueprintInstall = new this.BlueprintTask({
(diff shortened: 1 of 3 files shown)


rwjblue:
Can you add a smoke test that confirms yarn.lock is created? One possible place for that would be the fixture tests, but you could do it elsewhere also...

rwjblue:
The simplest thing would be to add another assertion to these tests https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/new-test.js#L468-L490 that just confirms yarn.lock is _present_...

Turbo87:
@rwjblue the tests you linked run using `--skip-npm`, but that means that `yarn` will actually never run and thus not create a `yarn.lock` file either 🤔

rwjblue:
Yeah, I see that now.  I guess I assumed that even if we didn't run `yarn install` we would emit a `yarn.lock`, but clearly that is not true.

Turbo87:
> Assuming you have smoke tested this locally (npm link + ember new foo --yarn) and confirmed it is working "for real", I'm 👍 .

yep, seemed to work correctly for all cases now (`ember new --yarn`, `ember install --no-yarn`, ...)

PR #7527, opened 2017-12-28
EmberApp: Remove deprecated `contentFor()` hooks

Turbo87:
The way I understand the deprecation warning is that the `app-prefix`, `app-suffix`, `vendor-prefix` and `vendor-suffix` hooks are deprecated and can be removed completely for v3.0. Is that understanding correct?

### DIFF
--- a/lib/broccoli/app-prefix.js
+++ b/lib/broccoli/app-prefix.js
@@ -1,3 +1,1 @@
 "use strict";
-
-{{content-for 'app-prefix'}}
--- a/lib/broccoli/app-suffix.js
+++ b/lib/broccoli/app-suffix.js
@@ -1,1 +1,0 @@
-{{content-for 'app-suffix'}}
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1827,8 +1827,6 @@ class EmberApp {
    */
   contentFor(config, match, type) {
     let content = [];
-    let deprecatedHooks = ['app-prefix', 'app-suffix', 'vendor-prefix', 'vendor-suffix'];
-    let deprecate = this.project.ui.writeDeprecateLine.bind(this.project.ui);
 
     // This normalizes `rootURL` to the value which we use everywhere inside of Ember CLI.
     // This makes sure that the user doesn't have to account for it in application code.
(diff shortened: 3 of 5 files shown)


stefanpenner:
I believe this may need to be brought back. Specifically, there may not be a reasonable way to wrap `app.js` and `vendor.js` in timing metrics.

A concrete example would be to wrap `app.js` and `vendor.js` in `performance.mark` via the following content-for hooks:

* `vendor-prefix`
* `vendor-suffix`
* `app-prefix`
* `app-suffix`

```js
(typeof window !== 'undefined' && window && window.performance && window.performance.mark) {
  window.performance.mark('mark_app_start');
}

// ... <app.js or vendor.js> content...

if (typeof window !== 'undefined' && window && window.performance && window.performance.mark) {
  window.performance.mark('mark_app_end');
  if (window.performance.getEntriesByName('mark_app_start').length > 0) {
    window.performance.measure('mark_app_eval', 'mark_app_start', 'mark_app_end');
  }
}
```

---

I've tried a-few alternatives, the only one I can come up with would either break or risk breaking sourcemaps further, or require a second (both costly/error prone) sourcemap concat.

stefanpenner:
I believe this may need to be brought back. Specifically, there may not be a reasonable way to wrap `app.js` and `vendor.js` in timing metrics.

A concrete example would be to wrap `app.js` and `vendor.js` in `performance.mark` via the following content-for hooks:

* `vendor-prefix`
* `vendor-suffix`
* `app-prefix`
* `app-suffix`

```js
(typeof window !== 'undefined' && window && window.performance && window.performance.mark) {
  window.performance.mark('mark_app_start');
}

// ... <app.js or vendor.js> content...

if (typeof window !== 'undefined' && window && window.performance && window.performance.mark) {
  window.performance.mark('mark_app_end');
  if (window.performance.getEntriesByName('mark_app_start').length > 0) {
    window.performance.measure('mark_app_eval', 'mark_app_start', 'mark_app_end');
  }
}
```

---

I've tried a-few alternatives, the only one I can come up with would either break or risk breaking sourcemaps further, or require a second (both costly/error prone) sourcemap concat.

PR #7528, opened 2017-12-28
EmberApp: Overwrite `app/config/environment` in `tests.js`

### LINKED ISSUE #4535: Incorrect test environment when using storeConfigInMeta: false
When using `storeConfigInMeta: false` the testing environment is `"development"` instead of `"test"`. Steps to reproduce:
- Set `storeConfigInMeta: false`
- `ember serve`
- Visit http://localhost:4200/tests/index.html
- In the console, `require('my-app/config/environment').default.environment` is `development` instead of `test`.

In acceptance tests, this means that the router's `locationType` is not set to `"none"` which is really annoying to work with.

Turbo87:
This PR fixes the content of the `my-app/config/environment` module when `storeConfigInMeta: false` is used and http://localhost:4200/tests/ is visited. A similarly named module will be generated in the `tests.js` file which overwrites the one contained in `my-app.js`.

Overwriting in `test-support.js` unfortunately does not work because `test-support.js` is imported before `my-app.js` which would overwrite the config in the wrong way.

This is needed to make `setApplication()` and http://localhost:4200/tests/ work well together when `storeConfigInMeta: false` is used.

Resolves #4535

/cc @stefanpenner @kellyselden

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1222,6 +1222,33 @@ class EmberApp {
   }
 
   /**
+    @private
+    @method _testAppConfigTree
+    @return
+  */
+  _testAppConfigTree() {
+    if (!this._cachedTestAppConfigTree) {
+      let files = ['app-config.js'];
+
+      let emberCLITree = new ConfigReplace(new UnwatchedDir(__dirname), this._configTree(), {
+        configPath: path.join(this.name, 'config', 'environments', `test.json`),
+        files,
+
+        patterns: this._configReplacePatterns(),
+      });
+
+      this._cachedTestAppConfigTree = new Funnel(emberCLITree, {
+        files,
+        srcDir: '/',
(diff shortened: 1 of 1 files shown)


Turbo87:
I guess technically this is a bugfix and could be backported to the `release` branch, but I'm somewhat concerned about breaking stuff so maybe it's better to include this in the v3.0 release instead 🤔

rwjblue:
> Overwriting in test-support.js unfortunately does not work because test-support.js is imported before my-app.js which would overwrite the config in the wrong way.

test-support.js is not the correct place for this anyways, it belongs (as you have it) in tests.js...

rwjblue:
Changes seem good to me, though it would be nice to get a test in to ensure we don't regress either side...

rwjblue:
Going to land this (into `beta` branch), I'll followup with tests later this week...

PR #7532, opened 2018-01-02
testing ci

GavinJoyce:
testing CI, it seems like it may be broken


GavinJoyce:
yep, master is broken:

screen shot 2018-01-02 at 12 53 21

rwjblue:
Should be fixed now...

PR #7536, opened 2018-01-02
Avoid bower usage in config/ember-try.js.

rwjblue:


rwjblue:
You can see this passing in https://github.com/emberjs/ember-test-helpers/pull/298.  

For example, in that PR when running `ember-release` you can see that it used the correct Ember version:

```
Scenario ember-release: SUCCESS
Command run: ember test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember-source       │ http://builds.emb… │ 2.18.0-release+dee84b06      │ yarn     │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘
```

alexander-alvarez:
Not sure if this is worth noting, but I tried this config in my app, and what was previously passing when pulling from bower now fails with weird errors (canary, beta, release)

My canary job passed on `3.0.0-alpha.1-null+213d4366` when using bower
```
ember-cli: 2.16.2
node: 8.9.3
os: linux x64
```
note I'm also using ember-concurrency.

some stacktraces
```
TypeError: _args.slice is not a function
at taskHelperClosure (http://localhost:7357/assets/vendor.js:183764:27)
at performHelper (http://localhost:7357/assets/iris-web.js:7186:43)
at SimpleHelperReference.compute (http://localhost:7357/assets/vendor.js:36562:20)
at SimpleHelperReference.value (http://localhost:7357/assets/vendor.js:36252:53)
at CapturedNamedArguments.value (http://localhost:7357/assets/vendor.js:16768:43)
at processComponentArgs (http://localhost:7357/assets/vendor.js:36159:31)
at CurlyComponentManager.create (http://localhost:7357/assets/vendor.js:29978:63)
at http://localhost:7357/assets/vendor.js:17751:33
at AppendOpcodes.evaluate (http://localhost:7357/assets/vendor.js:16150:13)

```
```
at Object.actions.onChange (http://localhost:7357/assets/tests.js:1259:16)
at Backburner._join (http://localhost:7357/assets/vendor.js:24953:31)
at Backburner.join (http://localhost:7357/assets/vendor.js:24702:25)
at Function.run.join (http://localhost:7357/assets/vendor.js:41466:30)
```
```
>
at Object.<anonymous> (http://localhost:7357/assets/tests.js:2888:14)
at <anonymous>
```

Let me know if there's any other context I can give

kategengler:
I did some testing of this and found that the package is cached by npm by url, so this approach will not work unless the cache is cleared.

For example, Build #1 runs yesterday with `.../canary.tgz`, Build #2 runs today, without a cleared cache, will use the exact package from yesterday, making this not a successful way to test the latest canary. 

I observed the same behavior with yarn.

rwjblue:
Yep, my current plan is to update ember-try to resolve promises  (still not certain of the exact API) so that we can fetch the current SHA (which is an immutable tarball for that SHA specific URL)...

kellyselden:
Similar to #7468

rwjblue:
Updated to match the concerns mentioned by @kategengler above, should be ready for another round of review...

rwjblue:
I tested this setup in https://github.com/emberjs/ember-test-helpers/pull/299 (in case folks want to see a passing CI run)...

rwjblue:
> I think we can probably come up with a better API, but this seems okay to me for now until that API exists

Ya, definitely open to suggestions.

Turbo87:
@rwjblue why did you put this into `beta` instead of `release`? it seems that currently a lot of addon CI runs are failing because `beta` isn't running properly 🤔

rwjblue:
Example?

Turbo87:
nevermind, it seems I was confusing this with other issues. namely that those addons were still using PhantomJS which does not work anymore with Ember 3.0 due to `WeakMap` usage and other things.

Issue #7535, opened 2018-01-02
The config caching breaks the configPath option


kanongil:
After https://github.com/ember-cli/ember-cli/pull/7270, the `EmberApp()` `configPath` option no longer works (since 2.16-beta.2).

This is because the cache is not invalidated when the path is changed on project init here: https://github.com/ember-cli/ember-cli/blob/daf7fc93af7eef82ea29b4e950b00c397a7e7c93/lib/broccoli/ember-app.js#L201.

/cc @krisselden who implemented the feature.

rwjblue:
TBH, I doubt many of us knew the feature existed.  Can you share your usage scenario? 

It seems straightforward to fix (e.g. to purge the cache just after setting the update `configPath` function), mind sending in a PR?

kanongil:
I setup my project with it long ago, as a way to support multiple variant configurations. I can probably rework it, if you want to deprecate it.

PR in #7538.

locks:
Closed by #7535!

PR #7538, opened 2018-01-03
Fix `configPath` caching

kanongil:
See #7535 for details.

### DIFF
--- a/tests/unit/broccoli/ember-app-test.js
+++ b/tests/unit/broccoli/ember-app-test.js
@@ -673,6 +673,21 @@ describe('EmberApp', function () {
       expect(project.configPath().slice(-expected.length)).to.equal(expected);
     });
 
+    it('should update project.config() if configPath option is specified', function () {
+      project.require = function (path) {
+        return () => ({ path });
+      };
+
+      expect(project.config('development')).to.deep.equal({});
+
+      new EmberApp({
+        project,
+        configPath: path.join('..', '..', 'app-import', 'config', 'environment'),
+      });
+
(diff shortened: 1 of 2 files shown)


rwjblue:
@kanongil - Mind rebasing? There was an issue with CI that was recently resolved and I believe this will be passing after a rebase...

kanongil:
Rebased and pushed.

It seems my test fails on windows because the expect relies on unix path names:
```
expected 'C:\\projects\\ember-cli\\tests\\fixtures\\app-import\\config\\environment' to include 'app-import/config/environment'
```

tests/unit/broccoli/ember-app-test.js
```diff
@@ -49,6 +49,21 @@ describe('EmberApp', function() {
...
+      };
+
+      expect(project.config('development')).to.deep.equal({});
+
+      new EmberApp({
+        project,
+        configPath: '../../app-import/config/environment',
+      });
+
+      expect(project.config('development').path).to.contain('app-import/config/environment');
```
rwjblue: Maybe use path.join or normalize for the expected value?

rwjblue: @kanongil - ping?  I believe fixing this will fix the current failure...

Turbo87:
@kanongil friendly reminder

tests/unit/broccoli/ember-app-test.js
locks: I rebased and addressed this comment.

Issue #7530, opened 2018-01-02
Module Unification: Blueprints


GavinJoyce:
🔼 Part of [[QUEST] Module Unification: Final Cut](https://github.com/emberjs/ember.js/issues/16373)

---

This issue is part of the [Module Unification Quest](https://github.com/emberjs/ember.js/issues/16373) and will serve as a list of remaining tasks to land blueprint support for module unification apps.

## Interested in helping?

Great! Comment on the issue to take on one of these blueprints. They are mostly pretty straightforward, see the [service blueprint](https://github.com/emberjs/ember.js/pull/16397) for example. Anything without a 🔒 is up for grabs.

If you have any questions, you can ask them in the `#st-module-unification` channel on the [Ember community slack](https://emberjs.com/community/) or you can ask me directly `@gavinjoyce`.

## Creating a new module unification app

Install Ember CLI master:

```
npm install -g https://github.com/ember-cli/ember-cli.git
```

Generate a new app with the module unification env variables:

```
MODULE_UNIFICATION=true EMBER_CLI_MODULE_UNIFICATION=true ember new my-mu-app
```

Generate a component:

```
EMBER_CLI_MODULE_UNIFICATION=true ember g component x-button
```

Run the app:

```
EMBER_CLI_MODULE_UNIFICATION=true ember serve
```

### `ember-cli/ember-cli` ([blueprints](https://github.com/ember-cli/ember-cli/tree/master/blueprints)):

 * ~~`addon-import`~~ (not applicable)
 * [x] [`MODULE_UNIFICATION=true ember addon my-addon`](https://github.com/ember-cli/ember-cli/pull/7658)
 * [x] [`MODULE_UNIFICATION=true ember new my-app`](https://github.com/ember-cli/ember-cli/pull/7488)
 * ~~`ember generate blueprint`~~ (not applicable)
 * ~~`ember generate http-mock`~~ (not applicable)
 * ~~`ember generate http-proxy`~~ (not applicable)
 * [x] [`ember generate in-repo-addon`](https://github.com/ember-cli/ember-cli/pull/7783)
   * [ ] remove `.jshintrc` and `.eslintrc` (see [discussion](https://github.com/ember-cli/ember-cli/pull/7783#discussion_r188061985))
 *  ~~`ember generate lib`~~
 * ~~`ember generate server`~~ (not applicable)
 * ~~`ember generate vendor-shim`~~ (not applicable)

[...]

PR #7541, opened 2018-01-04
Add `project.isModuleUnification()`

GavinJoyce:
part of https://github.com/emberjs/ember.js/pull/16043
(which is part of https://github.com/ember-cli/ember-cli/issues/7530)

This PR adds initial support for `ember g *` commands within a module unification project. The existence of a `src` directory in the ember project determines if that project is a module unification project. After this is merged, the next step will be to update the [Ember blueprints](https://github.com/emberjs/ember.js/tree/master/blueprints) to support MU output by checking `project.isModuleUnification()`

/cc @rwjblue

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -141,6 +141,10 @@ class Project {
       return false;
     };
 
+    NULL_PROJECT.isModuleUnification = function() {
+      return false;
+    };
+
     NULL_PROJECT.name = function() {
       return path.basename(process.cwd());
     };
@@ -183,6 +187,16 @@ class Project {
    */
   isEmberCLIAddon() {
     return !!this.pkg.keywords && this.pkg.keywords.indexOf('ember-addon') > -1;
+  }
+
+  /**
+     Returns whether this is using a module unification format.
+     @private
+     @method isModuleUnification
+     @return {Boolean} Whether this is using a module unification format.
+    */
+  isModuleUnification() {
+    return this.has('src');
   }
 
   /**
--- a/tests/helpers/mock-project.js
+++ b/tests/helpers/mock-project.js
@@ -63,6 +63,10 @@ class MockProject extends Project {
   isEmberCLIAddon() {
     return false;
   }
+
+  isModuleUnification() {
+    return false;
+  }
 }
 
 module.exports = MockProject;
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -526,6 +526,25 @@ describe('models/project.js', function() {
     });
   });
 
+  describe('isModuleUnification', function() {
+    beforeEach(function() {
+      projectPath = `${process.cwd()}/tmp/test-app`;
+
+      makeProject();
+    });
+
+    it('returns false when `./src` does not exist', function() {
+      expect(project.isModuleUnification()).to.equal(false);
+    });
+
+    it('returns true when `./src` exists', function() {
+      let srcPath = path.join(projectPath, 'src');
+      fs.ensureDirSync(srcPath);
+
+      expect(project.isModuleUnification()).to.equal(true);
+    });
+  });
+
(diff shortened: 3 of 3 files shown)


tests/unit/tasks/generate-from-blueprint-test.js
```diff
@@ -0,0 +1,44 @@
...
+    describe('in a module unification app', function() {
+      beforeEach(function() {
+        task.project.isModuleUnification = function() {
+          return true;
+        };
+      });
+
+      it('prefixes `component`', function() {
+        let name = task.prefixBlueprintName('component');
+        expect(name).to.equal('module-unification-component');
```
alexander-alvarez: Should this not still be `component`?
We could probably hook into `isModuleUnification` that you added  and override `__path__` and `__templatepath__` in [ember.js component blueprint](https://github.com/emberjs/ember.js/blob/master/blueprints/component/index.js#L26..L43)

GavinJoyce: It could be, I considered that approach. I think it's cleaner to keep these blueprints separate though, especially given that there isn't much reuse between the contents of the files

GavinJoyce: A downside of my approach is that there is a coupling between ember-cli and ember (you would need to update both to use the new MU blueprints). Your suggested approach doesn't have this issue.

I'll try your approach as see how it sits, thanks!

alexander-alvarez: > would need to update both to use the new MU blueprints and again when we rename ..

I wouldn't say that's a downside to your approach, as we could pollyfill the MU blueprints, and issue a deprecation warning when the hosts' app version doesn't need the polyfill anymore

GavinJoyce: true

alexander-alvarez: I think the big difference is that we can experiment and build blueprints with `isModuleUnification` in ember (or even addons) without having to enable each different type of MU-blueprint here https://github.com/ember-cli/ember-cli/pull/7541/files#diff-c3643df1b2f6f7f28a101393e4fe0c12R99 

Already we can support many permutations of the same file type like we do here https://github.com/emberjs/ember.js/tree/master/blueprints/controller-test (e.g we could have a `module-unification-files` prefix there)

GavinJoyce: 👍

I don't believe that we've got access to the project in the blueprints, I'll add a `options.moduleUnification` property and refactor the `component` blueprint

alexander-alvarez: here it is: https://github.com/emberjs/ember.js/blob/master/blueprints/component/index.js#L58

GavinJoyce: great, thanks

GavinJoyce: thanks for your feedback, updated

lib/models/project.js
```diff
@@ -185,6 +189,10 @@ class Project {
     return !!this.pkg.keywords && this.pkg.keywords.indexOf('ember-addon') > -1;
   }
 
+  isModuleUnification() {
```
alexander-alvarez: documentation

ro0gr: is it supposed to be a public API? Maybe it worth to mark this method as private or protected in order to avoid following deprecation when `isModuleUnification` becomes default?

GavinJoyce: Good point. I think it should be private as it's just a temporary internal API to allow us to branch the blueprints.

GavinJoyce: updated

GavinJoyce: thanks, updated

PR #7542, opened 2018-01-04
trivial error-message fix for bad watcher type

dcombslinkedin:
I just wanted the error message to reflect the values being checked against.


twokul:
thanks!

PR #7546, opened 2018-01-11
Remove unused testing helper files.

rwjblue:
As of ember-source@3.0.0 (and ember-data@3.0.0) the testing blueprints
are automatically emitting emberjs/rfcs#232 or emberjs/rfcs#268
compatible output. With those, these helpers are no longer used for new
apps.

Existing apps should only delete these files once they have migrated to
the new testing system...


kellyselden:
We noticed on the call that some addons are putting things in these files. We will need to be aware of this.

0xadada:
@rwjblue i'm using the new QUnit API, but for an instance initializer test, i'm still using `startApp` and `destroyApp`, what would the replacement be in the new testing world?

example: https://github.com/mirai-audio/mir/blob/16/library/tests/unit/instance-initializers/i18n-test.js#L36

PR #7548, opened 2018-01-11
Make async/await work nicely by default.

rwjblue:
This change does a few things:

* Add `ember-maybe-import-regenerator` to the default blueprint.
* Drop IE9 from `config/targets.js`, and replace with IE11 (to match
  Ember 3.0.0's minimum browser requirements).
* Modify `config/targets.js` to make development builds target modern
  browsers.

This essentially brings over the configuration discussed in [this blog post](http://rwjblue.com/2017/10/30/async-await-configuration-adventure/):

* [Section 30](http://rwjblue.com/2017/10/30/async-await-configuration-adventure/#install-maybe-install-regenerator)
* [Section 10](http://rwjblue.com/2017/10/30/async-await-configuration-adventure/#limit-targets-locally)

### DIFF
--- a/blueprints/app/files/config/targets.js
+++ b/blueprints/app/files/config/targets.js
@@ -1,8 +1,16 @@
+const browsers = [
+  'last 1 Chrome versions',
+  'last 1 Firefox versions',
+  'last 1 Safari versions'
+];
+
+const isCI = !!process.env.CI;
+const isProduction = process.env.EMBER_ENV === 'production';
+
+if (isCI || isProduction) {
+  browsers.push('ie 11');
+}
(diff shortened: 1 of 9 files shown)


Turbo87:
FWIW I'm a little worried about dev being built differently from prod now... 🤔

rwjblue:
CI also checks the transpiled version too.

Turbo87:
yeah, I've seen that... still got a somewhat bad feeling about this

rwjblue:
Hmm. We should discuss in higher bandwidth. That specific part was not the high order bit here...

PR #7549, opened 2018-01-11
Use `sudo: required` to work around issue in CI.

rwjblue:
Without this change all tests that run `ember test` (which leverages headless chrome) will end up failing:

```
not ok 1 Chrome - error
    ---
        message: >
            Non-zero exit code: null
            Stderr: 
             [0111/201538.220867:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox is owned by root and has mode 4755.
            Failed to generate minidump.
            
        Log: |
            { type: 'error', text: 'Non-zero exit code: null' }
            { type: 'error',
              text: '[0111/201538.220867:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I\'m aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox is owned by root and has mode 4755.\nFailed to generate minidump.' }
    ...
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 dist: trusty
 
 addons:


Turbo87:
can we use `--no-sandbox` instead?

rwjblue:
Yep, definitely can.

jacobq:
This looks like it prevents use of the (faster) container-based VMs. Is that right? If so, is the reason that this was chosen as the default (vs. `--no-sandbox`) for some kind of security reason? It seems to me that the security risk is very low since one is presumably just running one's own tests (and on a VM at that).
https://docs.travis-ci.com/user/reference/overview/

Turbo87:
@jacobq see https://github.com/ember-cli/ember-cli/pull/7566

jacobq:
@Turbo87 Yes, I saw that, thanks, but my point/question was that shouldn't that make this `sudo: required` change unnecessary? (i.e. only one of those changes is needed so why do both?)

I suppose the VM boot time performance gain (50s / 10x) only really matters if it's on the order of the rest of the time (installing packages, running tests, etc.) and users can always tweak their own `.travis.yml` file as desired, so maybe it's no big deal. I mostly just want to understand the trade-offs so I can "get it right" for my projects.

Turbo87:
1) the change you're seeing here is only for Ember CLI itself, not for any projects using Ember CLI 
2) the change here has already been reverted and replaced by the `--no-sandbox` approach

PR #7553, opened 2018-01-12
Remove embertest from ESLint configuration.

rwjblue:
As of Ember 3.0 the new testing APIs introduced in emberjs/rfcs#232 and emberjs/rfcs#268 are enabled and used by default. In these APIs no globals are used, therefore this extra `tests/` override is removed.

This specifically removes the following globals from being allowed:

* `andThen`
* `click`
* `currentPath`
* `currentRouteName`
* `currentURL`
* `fillIn`
* `find`
* `findWithAssert`
* `keyEvent`
* `pauseTest`
* `resumeTest`
* `triggerEvent`
* `visit`
* `wait`

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -43,15 +43,6 @@ module.exports = {
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
         // add your custom rules and overrides for node files here
       })<% } %>
-    },
-
-    // test files
-    {
-      files: ['tests/**/*.js'],
-      excludedFiles: ['tests/dummy/**/*.js'],
-      env: {
-        embertest: true
-      }
     }
   ]
 };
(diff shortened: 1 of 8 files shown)


0xadada:
i love to see additional simplicity

Turbo87:
@rwjblue FYI `find` still seems to be included somehow. I have `embertest` disabled in one of our projects, but using global `find` is still not producing a warning somehow :-/

rwjblue:
@Turbo87 what is that `find`? I do not think it’s Ember’s version...

Turbo87:
I don't know, the only thing I know is that global `find` does not seem to trigger ESLint for some unknown reason

rwjblue:
If you log it to the console, then hover in the devtools it will show you the function source location and you can see where it’s coming from...

Turbo87:
I don't think `find` is actually available, but ESLint seems to think it would be an allowed global

rwjblue:
Maybe it’s this: https://developer.mozilla.org/en-US/docs/Web/API/Window/find ?

Turbo87:
could well be. I guess it comes from https://github.com/sindresorhus/globals/blob/master/globals.json#L353

PR #7560, opened 2018-01-17
Migrate to new format for mode specific arguments in testem config.

rwjblue:
Testem is deprecated the format currently in use for `browser_args` in https://github.com/testem/testem/pull/1182.

Specifically, this format is deprecated:

```js
browser_args: {
  Chrome: {
    mode: 'ci',
    args: [
      // args go here....
    ]
  }
}
```

In favor of this format:

```js
browser_args: {
  Chrome: {
    ci: [
      // args go here....
    ]
  }
}
```

This change is being made so that Testem can support using different sets of arguments for different modes. The change also adds an `all` key that can be used for shared arguments between modes (they are merged with the mode specific args).

---

The change in Testem is merged but not yet released, we will need to land this PR when the Testem release is done (likely into the release branch at that time to silence the deprecations as quickly as possible).

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -9,8 +9,7 @@ module.exports = {
   ],
   browser_args: {
     Chrome: {
-      mode: 'ci',
-      args: [
+      ci: [
         // --no-sandbox is needed when running Chrome inside a container
         process.env.CI ? '--no-sandbox' : null,
 


johanneswuerbach:
This should now be unblocked by https://github.com/testem/testem/releases/tag/v2.0.0

Turbo87:
@rwjblue can you rebase this? we use `testem@2` now, so this should be unblocked.

rwjblue:
Updated to point at release branch (and fix conflicts).

PR #7563, opened 2018-01-20
Fix 'const' declarations in non-strict mode are not supported

rwwagner90:
Without `'use strict';` I get lint errors of `'const' declarations in non-strict mode are not supported yet on Node 4.5.0` due to the new node linting rules. Adding this fixes the issue. Alternatively, we could disable that rule for this file or altogether.

### DIFF
--- a/blueprints/app/files/config/targets.js
+++ b/blueprints/app/files/config/targets.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const browsers = [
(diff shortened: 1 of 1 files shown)


rwwagner90:
It looks like a smoke test or something is failing. Any ideas why?

Turbo87:
@rwwagner90 this appears to be an unrelated failure

PR #7565, opened 2018-01-21
tests/acceptance: Update test number expectation

Turbo87:
`ember-qunit` is now generating a test of its own which we did not expect before. Updating the expected number of tests fixes our CI runs.


rwjblue:
This coupling is annoying. I wonder if we can change the assertion around a bit?

Turbo87:
merged upwards into `master` now

PR #7566, opened 2018-01-21
testem: Use `--no-sandbox` on TravisCI

Turbo87:
Similar to https://github.com/ember-cli/ember-cli-chai/pull/45 this updates the `testem.js` file of the `app` blueprint to run Chrome with `--no-sandbox` when used on TravisCI to avoid https://github.com/travis-ci/travis-ci/issues/8836

/cc @backspace

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -11,11 +11,14 @@ module.exports = {
     Chrome: {
       mode: 'ci',
       args: [
+        // --no-sandbox is needed when running Chrome inside a container
+        process.env.TRAVIS ? '--no-sandbox' : null,
+
         '--disable-gpu',
         '--headless',
         '--remote-debugging-port=0',
         '--window-size=1440,900'
-      ]
+      ].filter(Boolean)
     }
   }
 };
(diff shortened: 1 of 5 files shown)


blueprints/app/files/testem.js
```diff
@@ -11,11 +11,14 @@ module.exports = {
     Chrome: {
       mode: 'ci',
       args: [
+        // --no-sandbox is needed when running Chrome inside a container
+        process.env.TRAVIS ? '--no-sandbox' : null,
```
rwjblue: Should we use `process.env.CI` instead? As I understand it a number of CI providers have this same basic issues (e.g. GitLab CI), and the CI environment flag is pretty universal...

Turbo87: I'm not sure... GitLab CI can and often does use Docker containers, but you can also run it on real hosts, AppVeyor is likely VM-based, so there are also some CI systems that do not run in containers. The condition is easy enough to adjust if needed 🤔

backspace:
Nice, thanks for looking into this. I figured the sandbox was desirable for some reason? haha oops

Turbo87:
FYI I had to include two more commits to make CI pass on the `release` branch

PR #7569, opened 2018-01-22
mark "lib" folder as node style in eslint for apps

kellyselden:
in-repo addons will fail `npm run lint:js` without this.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -23,7 +23,8 @@ module.exports = {
         'index.js',<% } %>
         'testem.js',
         'ember-cli-build.js',
-        'config/**/*.js'<% if (blueprint !== 'app') { %>,
+        'config/**/*.js'<% if (blueprint === 'app') { %>,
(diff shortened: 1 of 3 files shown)


blueprints/app/files/.eslintrc.js
```diff
@@ -23,7 +23,8 @@ module.exports = {
         'index.js',<% } %>
         'testem.js',
         'ember-cli-build.js',
-        'config/**/*.js'<% if (blueprint !== 'app') { %>,
+        'config/**/*.js'<% if (blueprint === 'app') { %>,
+        'lib/**/*.js'<% } %><% if (blueprint !== 'app') { %>,
```
rwjblue: It isn't all of `lib/**`, its only `lib/**/index.js` and `lib/**/config/*.js`.

kellyselden: oh you're right

kellyselden: I went "lib/**/*.js" => "lib/*/index.js". We don't want to match an ember file named "index.js". Also our in-repo addon blueprint doesn't have config, so I left that out.

PR #7585, opened 2018-01-28
lint js for node@4 compatibility

ro0gr:
It should help contributors to detect issues locally in the editor instead of [CI](https://travis-ci.org/ember-cli/ember-cli/jobs/334482664#L521)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,8 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
   },
-  extends: 'eslint:recommended',
+  plugins: ["node"],
+  extends: ['eslint:recommended', 'plugin:node/recommended'],
   env: {
     browser: false,
     node: true,
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,9 @@
 /assets
 /dev
 /docs
-/tests/!(helpers)
+/tests/*
+/tests/*/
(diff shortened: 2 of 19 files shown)


kellyselden:
Looks like it needs more configuration.

ro0gr:
I believe all `node/no-unpublished-require` errors detected in the `tests/**/*/test.js`
> tests/unit/tasks/test-test.js
>    3:24  error  "chai" is not published  node/no-unpublished-require 

is a failure of `node/no-unpublished-require` rule. 
According to the [rule documentation](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md#rule-details)
> If a source code file satisfies all of the following conditions, the file is *published*.
>
> - "files" field of package.json includes the file or "files" field of package.json does not exist.
> - .npmignore does not include the file.

npm-ignored files should not be linted against this rule. I'll try to figure it out.

By the way we also have few potentially useful errors for code located in the `lib/` and `tests/helpers/` directories.

ro0gr:
I believe all the "no-extraneous-require" errors are false negative (https://github.com/mysticatea/eslint-plugin-node/issues/107)

ro0gr:
@kellyselden @rwjblue I think it's ready to go!

rwjblue:
Thank you @ro0gr!

PR #7592, opened 2018-01-30
Allow .npmrc to be respected during blueprint install

thoov:
NPM has 4 places it uses for .npmrc (https://docs.npmjs.com/files/npmrc#files). 3 of the 4 are supported as their locations are external to the project root but if there is an .npmrc file located at the project root it will not be respected. This is because we create a temp directory (createTempFolder) in which we run the npm install within. This PR add support for `--userconfig=/path/to/project-root/.npmrc` which will be used during the install of the blueprint only if a local .npmrc is present.  This is useful for installing private blueprints such as `ember init -b @company/company-specific-blueprint`.


thoov:
@rwjblue @stefanpenner and you guys review this

stefanpenner:
I believe if we copy `.npmrc` (if present) into the tempdir, everything should "just work" right?

lib/tasks/install-blueprint.js
```diff
@@ -135,6 +140,33 @@ class InstallBlueprintTask extends Task {
...
+   * registry.
+   */
+  _copyNpmrc(tmpDir) {
+    return new Promise((resolve, reject) => {
+      let rcPath = path.join(this.project.root, '.npmrc');
+      let tempLocation = path.join(tmpDir, '.npmrc');
+
+      if (fs.existsSync(rcPath)) {
+        try {
+          fs.writeFileSync(tempLocation, fs.readFileSync(rcPath));
```
twokul: I wonder if we can use [`fs.copy`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md) from [`fs-extra`](https://github.com/jprichardson/node-fs-extra)? We have it as [a dependency](https://github.com/ember-cli/ember-cli/blob/master/package.json#L80).

lib/tasks/install-blueprint.js
```diff
@@ -135,6 +140,33 @@ class InstallBlueprintTask extends Task {
...
+          return resolve();
```
twokul: do we want to `resolve` with some value maybe? `true`?

twokul:
@thoov could you squash commits as well?

lib/tasks/install-blueprint.js
thoov: Added true

lib/tasks/install-blueprint.js
thoov: Moved over to copySync. Thanks

lib/tasks/install-blueprint.js
```diff
@@ -135,6 +140,32 @@ class InstallBlueprintTask extends Task {
...
+  /*
+   * NPM has 4 places it uses for .npmrc. 3 of the 4 are supported as their locations
+   * are external to the project root but if there is an .npmrc file located at the project root it will
+   * not be respected. This is because we create a tmp directory (createTempFolder) where we run
+   * the npm install. This function simply copies over the .npmrc file to the tmp directory so that it
+   * will be used during the install of the blueprint. Useful for installing private blueprints
+   * such as `ember init -b @company/company-specific-blueprint` where the module is in a different
+   * registry.
+   */
+  _copyNpmrc(tmpDir) {
```
stefanpenner: why is this a promise, if all the IO comands within are sync?

thoov: Done

thoov:
@stefanpenner Can you take another look?

lib/tasks/install-blueprint.js
```diff
@@ -135,6 +140,28 @@ class InstallBlueprintTask extends Task {
...
+    let rcPath = path.join(this.project.root, '.npmrc');
+    let tempLocation = path.join(tmpDir, '.npmrc');
+
+    if (fs.existsSync(rcPath)) {
+      try {
```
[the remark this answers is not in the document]
thoov: removed

tests/unit/tasks/install-blueprint-test.js
```diff
@@ -1,9 +1,18 @@
...
+const path = require('path');
+const fs = require('fs-extra');
 const td = require('testdouble');
 const SilentError = require('silent-error');
+
 const expect = require('../../chai').expect;
+const mkTmpDirIn = require('../../../lib/utilities/mk-tmp-dir-in');
+const InstallBlueprintTask = require('../../../lib/tasks/install-blueprint');
+
+let root = process.cwd();
```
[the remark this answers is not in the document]
thoov: @stefanpenner What is wrong with process.cwd()?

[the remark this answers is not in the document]
twokul: `__dirname` will always return a path to the file whereas `process.cwd` will return a path where `node` is executed. By using `__dirname` you will make sure to have a stable path at all times (as you start with a file and going up). If you use use `process.cwd`, you will get different paths.

```sh
mocha tests/unit/tasks/install-blueprint-test.js
# process.cwd() => /full/path/to/ember/cli
# __dirname => /full/path/to/ember/cli/tests/unit/tasks/
```

vs.

```sh
cd tests/unit/tasks/ && mocha install-blueprint-test.js
# process.cwd() => /full/path/to/ember/cli/tests/unit/tasks/
# __dirname => /full/path/to/ember/cli/tests/unit/tasks/
```

`tmp` directory will be created in different places and we want to avoid that. @stefanpenner correct me if I'm wrong.

PR #7557, opened 2018-01-14
[wip] Change confirmBlueprinted to test against fixtures

ro0gr:
When worked on [MU blueprints improvements](https://github.com/ember-cli/ember-cli/pull/7500) I've noticed that the current [`confirmBlueprinted`](https://github.com/ember-cli/ember-cli/blob/fe6d773f94461d048c6b15d5d99b967abb917cb6/tests/acceptance/new-test.js#L37) is not a very suitable tool for testing of blueprints because:
 - it does relies on blueprint internals
 - it doesn't test file contents
 - it doesn't allow to test custom logic in blueprint's index.js

This pr tries to address these issues by testing of `new` and `addon` commands output against the fixtures. I've chosen  [`fixturify`](https://www.npmjs.com/package/fixturify) to deal with fixtures. It seems quite a convenient tool cause it efficiently serializes fs layout including file contents into a POJOs which allows us to make deep equals with ease.

I would appreciate a feedback before I've gone too far. Does it look like a good direction?

TODO:
 - [x] split welcome tests from yarn tests 
 - [x] unify `expectProject` tests and "verify fixtures" tests.
 - [ ] ~add `lodash.{get,set}` to `ember-cli-lodash-subset`~
worked around by flattening `fixturify` results
 - [x] cut number of `module-unification-app` fixture files by re-using app's project files
 - [x] fix tests

PR #7595, opened 2018-02-02
Test "new" and "addon" commands output against fixtures

ro0gr:
Supersedes #7557 
It's needed to safely experiment with blueprints specifically [mu-app](https://github.com/ember-cli/ember-cli/pull/7500) blueprint.

There are few issues with the current [`confirmBlueprinted`](https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/new-test.js#L38) approach:
 - it relies on blueprint internals:
   - it uses Blueprint `files/` directory as a single source of truth. So if we have some custom logic which filters/adds some files `confirmBlueprinted` doesn't allow us to test this. 
   - it relies on `Blueprint.renamedFiles` which is bad cause I believe that tests should not rely on the tested units.
 - it doesn't test the file contents. Only file file paths.

I think the main issue with this approach is that if we remove half of the blueprint files then test will continue to pass.

Also there are few more concerns:
 - we want to test app, addon, and mu-app w/ and w/o `ember-welcome`. But we don't want to have the whole project fixture for each of these cases. We need a way to use a default fixture and just apply small patches to it right in the test(w/o additional fixture folder). 
 - mu-app and app blueprint output have lots of common parts. It would be great not to duplicate this files. If we re-use app fixture for mu-app we can be more confident that mu-app blueprint is in sync with the app.

In order to address these issues I've introduced 2 new helpers:
### `loadProjectFixture(fixturePaths, options)` 
Loads the fixture from a single or multiple paths:
```js
// load npm app fixture
const npmApp = options => projectFixture('app/npm', options);
// load fixtures sequentially and merge them
const yarnAddon = options => projectFixture([npmAddon(), 'addon/yarn'], options);
```
Under the hood it uses `fixturify` to load the files but the result is flattened for the sake of a better diff results.

`loadProjectFixture` also allows to filter result with `files` option which accepts `minimatch` patterns:
```js
const muApp = options =>projectFixture([
  npmApp({ files: ['!app/**'] }), // load regular app but "app/" directory
  'module-unification-app/npm', // it contains only mu specific parts
], options);
```
And finally we can patch resulting files with a `patches` option:
```js
muApp({
  patches: {
    'vendor/script.js': '// test comment',
    'package.json': function(stringContent) {
        return stringContent;
     }
  }
})
```

### `expectProject(expectedProjectName, fixture)`

It's supposed to check that the current dir has an expected name and all the file content are equal to the fixture content. It prints 
 - a diff for each conflicting file 
 - list of missing and extra files if any.


package.json
```diff
@@ -127,10 +127,12 @@
     "chai-as-promised": "^7.0.0",
     "chai-files": "^1.2.0",
     "co": "^4.6.0",
+    "detect-indent": "^5.0.0",
     "ember-cli-blueprint-test-helpers": "^0.18.0",
     "ember-cli-internal-test-helpers": "^0.9.0",
     "eslint-plugin-chai-expect": "^1.1.1",
     "eslint-plugin-mocha": "^4.8.0",
+    "fixturify": "^0.3.4",
```
twokul: we should likely take advantage of `broccoli-test-helpers` which relies on [`fixturify`](https://github.com/broccolijs/broccoli-test-helper/blob/master/package.json#L39)

ro0gr: I'm not actually sure which benefits does it offer here. Looks like `broccoli-test-helper` is intended to be used for testing of the build output diffs which is not related for testing blueprints output I think. 

Of course I may miss something because I've never used `broccoli-test-helper` before... I'd be glad to switch to `broccoli-test-helpers` if there is some reason.

ro0gr:
Closing due to current removal of MU blueprints

PR #7598, opened 2018-02-04
Ignore all *.log files

twokul:
If you run into an issue with `yarn` locally, `yarn-error.log` won't be ignored.


rwjblue:
Seems good, probably also want to make sure its added to the `.npmignore` too...

twokul:
@rwjblue updated

PR #7599, opened 2018-02-04
Code cleanup and additional tests for `validProjectName`

twokul:


tests/unit/utilities/valid-project-name-test.js
```diff
@@ -5,17 +5,15 @@ const expect = require('chai').expect;
...
-    let nonConformantName = 'app';
-    let validated = validProjectName(nonConformantName);
-
-    expect(validated).to.not.be.ok;
+    ['app', '.', 'ember', 'so-cool.', 'vendor', '1234test'].forEach(name => {
+      expect(
+        validProjectName(name),
+        `'${name}' is an invalid name`
+      ).to.not.be.ok;
+    });
```
Turbo87: I usually prefer to generate complete tests instead of assertions, that way you can use `'${name}' is an invalid name` as the test name and don't have to add it to the assertions

lib/utilities/valid-project-name.js
```diff
@@ -1,11 +1,41 @@
 'use strict';
 
+/**
+ * Checks if the string starts with a number.
+ *
+ * @method startsWithNumber
+ * @return {Boolean}
+*/
+function startsWithNumber(name) {
+  let startsWithNumberRegex = new RegExp(/^\d.*$/);
```
Turbo87: if you use `/^\d.*$/` you don't need to wrap it in `new RegExp()`

twokul:
@Turbo87 friendly ping

lib/utilities/valid-project-name.js
```diff
@@ -1,11 +1,39 @@
 'use strict';
 
+/**
+ * Checks if the string starts with a number.
+ *
+ * @method startsWithNumber
+ * @return {Boolean}
+*/
+function startsWithNumber(name) {
+  return (/^\d.*$/).test(name);
```
Turbo87: I just noticed that the regex could actually be simplified to `/^\d/` since we don't care about the stuff that's after the first character

PR #7600, opened 2018-02-04
Remove "style file" silent error

twokul:
This check was originally added as a part of [this PR](https://github.com/ember-cli/ember-cli/pull/3642). Motivation behind the change was that build crashed w/ not-so-obvious error if stylesheet file had the same name as the project. I have doubts this check needs to exist anymore.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1448,10 +1448,6 @@ class EmberApp {
   */
   styles() {
     if (!this._cachedStylesTree) {
-      if (existsSync(`${this._stylesPath}/${this.name}.css`)) {
-        throw new SilentError(`Style file cannot have the name of the application - ${this.name}`);
-      }
-
       let addonTrees = this.addonTreesFor('styles');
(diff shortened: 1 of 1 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1448,9 +1448,9 @@ class EmberApp {
   */
   styles() {
     if (!this._cachedStylesTree) {
-      if (existsSync(`${this._stylesPath}/${this.name}.css`)) {
-        throw new SilentError(`Style file cannot have the name of the application - ${this.name}`);
-      }
+      //if (existsSync(`${this._stylesPath}/${this.name}.css`)) {
```
rwjblue: If we are going to remove it, let’s remove it. Commenting seems bad....

twokul: ofc, Good Sir. I opened this PR just to start a conversation.

twokul:
@rwjblue I did test it locally and it works just fine, both `ember serve` & `ember build`

PR #7605, opened 2018-02-05
blueprints/app: Add `qunit-dom` dependency by default

Turbo87:
Resolves https://github.com/emberjs/ember-qunit/issues/301

**Reasons:**

- because it makes DOM assertions so much more readable

/cc @ember-cli/core @trentmwillis @cibernox


rwjblue:
I'm definitely in favor, but I think we likely need an RFC first...

Turbo87:
https://github.com/ember-cli/rfcs/pull/116 just got merged, so this PR is good to go. I just rebased one more time.

PR #7606, opened 2018-02-05
Introduce `emberAppUtils`

twokul:
There are several motivations behind this change

+ reduct the API surface by passing parameters explicitly (as opposed to passing full `config` object)
+ make `ember-app` file smaller (this is going to happen regardless as a part of Default Packager work)
+ add documentation and tests to utility functions.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -12,7 +12,6 @@ const chalk = require('chalk');
 const resolve = require('resolve');
 
 const Project = require('../models/project');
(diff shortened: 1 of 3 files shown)


Turbo87:
any reason for putting them all in one file instead of one util per file?

twokul:
since they are all related to `ember-app` and are not used anywhere else, I decided to put them into one file @Turbo87

twokul:
@rwjblue I don't have a preference, what sounds better to you?

PR #7608, opened 2018-02-06
Allow --yarn to be respected when installing a blueprint module

thoov:
Currently if `--yarn` is present and you include a module name as a blueprint (ie: `ember init -b foobar --yarn`) ember-cli will install the module via npm and not yarn. This PR conditionally allows it to be installed via yarn or npm depending on the flags passed.

### DIFF
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -25,11 +25,12 @@ class InstallBlueprintTask extends Task {
   run(options) {
     let cwd = process.cwd();
     let name = options.rawName;
-    let blueprintOption = options.blueprint;
     // If we're in a dry run, pretend we changed directories.
     // Pretending we cd'd avoids prompts in the actual current directory.
     let fakeCwd = path.join(cwd, name);
     let target = options.dryRun ? fakeCwd : cwd;
+
+    this.execa = execa;
 
     let installOptions = {
       target,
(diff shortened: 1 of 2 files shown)


thoov:
@rwjblue @Turbo87 friendly ping

lib/tasks/install-blueprint.js
```diff
@@ -25,12 +25,13 @@ class InstallBlueprintTask extends Task {
   run(options) {
     let cwd = process.cwd();
     let name = options.rawName;
-    let blueprintOption = options.blueprint;
     // If we're in a dry run, pretend we changed directories.
     // Pretending we cd'd avoids prompts in the actual current directory.
     let fakeCwd = path.join(cwd, name);
     let target = options.dryRun ? fakeCwd : cwd;
 
+    this.execa = execa;
```
Turbo87: I think we should probably apply this default in the constructor/init 🤔

lib/tasks/install-blueprint.js
```diff
@@ -44,24 +45,24 @@ class InstallBlueprintTask extends Task {
 
     installOptions = merge(installOptions, options || {});
 
-    return this._resolveBlueprint(blueprintOption).then(blueprint => {
+    return this._resolveBlueprint(installOptions).then(blueprint => {
```
Turbo87: I think I would prefer it to not pass all of the options along. Either use a dedicated `yarn = true/false` parameter or an `options` parameter which only accepts/cares about `{ yarn: true/false }` for now. In other words, I think we should keep having `name` as a dedicated first parameter.

lib/tasks/install-blueprint.js
```diff
@@ -100,21 +101,37 @@ class InstallBlueprintTask extends Task {
...
 
-  _npmInstallModule(module, cwd) {
-    logger.info(`Running "npm install ${module}" in "${cwd}" ...`);
+  _npmInstallModule(module, cwd, options) {
+    let command = 'npm';
+    let args = ['install', module];
+
+    if (options && options.yarn === true) {
+      command = 'yarn';
+      args = ['add', module];
```
Turbo87: the problem with this might be that it will add the module to the `package.json`, which might not be intended. npm@5 does the same though, so this might be a problem in any case now 🤔

Turbo87:
closing due to inactivity

PR #7616, opened 2018-02-09
Remove extra parameter

twokul:


lib/broccoli/ember-app.js
```diff
@@ -745,7 +745,7 @@ class EmberApp {
     if (!experiments.MODULE_UNIFICATION || !this.trees.src) {
       index = this._rawAppIndex(htmlName);
```
Turbo87: the method is being used here with `optional=undefined`

Turbo87:
@twokul can we close this?

PR #7629, opened 2018-02-16
Use bower from project root instead of relative to file path

thoov:
During testing we change the current directory via `process.chdir(tmpdir)`. During
tests that run `ember new` this tmp directory becomes the project.root. The
bower executable should be installed into tmp/X/node_modules and not
ember-cli's node_modules. The error this was causing is that if bower
was not found (and it would not be as ember-cli does not depend on bower) bower would
install itself into ember-clis package.json and corrupt that file. This
now correctly uses the project.root as to where it should look for
bower.

### DIFF
--- a/lib/tasks/bower-install.js
+++ b/lib/tasks/bower-install.js
@@ -1,6 +1,5 @@
 'use strict';
 
-// Runs `bower install` in cwd
 const fs = require('fs-extra');
 const path = require('path');
 const existsSync = require('exists-sync');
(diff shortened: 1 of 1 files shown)


thoov:
You can verify this by running `yarn test-all` locally on master

Turbo87:
The difference here from a user perspective is that if an addon requests to use bower it was previously invisible to the user where bower was installed, while now it is explicitly added to his `package.json` file. I think I'm okay with this change as I don't see any real disadvantages from this, but we should be aware of this change 🤔

PR #7632, opened 2018-02-16
Pull script tags to head, add defer

NullVoxPopuli:
This has some neat pictures explaining the timeline of the different non-blocking loading options.
https://bitsofco.de/async-vs-defer/

With the script tags in the body, I wasn't able to get a steady app-loading animation going, while my scripts downloaded and got parsed. (this was in local dev, and my css loader (which was defined inline in index.html) was stuttering / "dropping frames")

### DIFF
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -12,14 +12,13 @@
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
 
+    <script defer src="{{rootURL}}assets/vendor.js"></script>
+    <script defer src="{{rootURL}}assets/<%= name %>.js"></script>
+
     {{content-for "head-footer"}}
   </head>
   <body>
     {{content-for "body"}}
-
-    <script src="{{rootURL}}assets/vendor.js"></script>
-    <script src="{{rootURL}}assets/<%= name %>.js"></script>
-
     {{content-for "body-footer"}}
   </body>
 </html>
--- a/blueprints/app/files/tests/index.html
+++ b/blueprints/app/files/tests/index.html
@@ -14,6 +14,12 @@
     <link rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
     <link rel="stylesheet" href="{{rootURL}}assets/test-support.css">
 
+    <script defer src="/testem.js" integrity="" data-embroider-ignore></script>
+    <script defer src="{{rootURL}}assets/vendor.js"></script>
+    <script defer src="{{rootURL}}assets/test-support.js"></script>
+    <script defer src="{{rootURL}}assets/<%= name %>.js"></script>
+    <script defer src="{{rootURL}}assets/tests.js"></script>
+
     {{content-for "head-footer"}}
     {{content-for "test-head-footer"}}
   </head>
@@ -27,12 +33,6 @@
         <div id="ember-testing"></div>
       </div>
     </div>
-
-    <script src="/testem.js" integrity="" data-embroider-ignore></script>
-    <script src="{{rootURL}}assets/vendor.js"></script>
-    <script src="{{rootURL}}assets/test-support.js"></script>
-    <script src="{{rootURL}}assets/<%= name %>.js"></script>
-    <script src="{{rootURL}}assets/tests.js"></script>
 
     {{content-for "body-footer"}}
     {{content-for "test-body-footer"}}


Turbo87:
@martndemus I think you played around with this already too. Any comments?

lifeart:
@NullVoxPopuli it can be solved with in-app addon and `content-for('body')` hook ( you may add additional js + css ) https://www.npmjs.com/package/ember-cli-content-for-config

kellyselden:
I'm also curious if your use case can be solved via content-for 'body'.

Turbo87:
@kellyselden it can't since the `<script>` tags are part of `index.html` and `content-for` can't remove stuff from that file

kellyselden:
I was curious if the animation logic getting added via content-for 'body' makes any of this necessary. Also if keeping the scripts where they are with the defer tag would be any different.

NullVoxPopuli:
From me to @Turbo87 on discord:
```
I think it's related to trying to get a better time-to-first-paint time:
my app has an "app shell" thing where just the critical html/css is on the index.html page. that critical html/css includes a loading css-animation.
I don't remember if that's what I was talking about specifically, I don't use script defer atm. I'd need to pick that PR back up and do some videos / benchmarking of what happens when
```

jrjohnson:
We also have an app shell loading screen and have been using `defer` on our script tags for ~6months. Without defer we were prone to some loading stutter in the SVG CSS animation when the scripts were reached and started downloading.

The only problem we've ever had is configuring `ember-auto-import` for async imports. We had to set `publicAssetUrl` which is documented at https://github.com/ef4/ember-auto-import#customizing-build-behavior. If this became the ember-cli default then it would probably be a good idea to update that addon as well.

Turbo87:
> The only problem we've ever had is configuring `ember-auto-import` for async imports.

I don't understand how that is related to `defer` for these static `<script>` tags. Can you elaborate?

lifeart:
https://addyosmani.com/blog/script-priorities/

stefanpenner:
If memory serves users often had issues with subtle hard to debug timing differences, this resulted in a more conservative default being choosen.

If we plan to change, we should ensure this is either no longer relevant or sufficiently low risk.

jrjohnson:
@Turbo87 

> > The only problem we've ever had is configuring `ember-auto-import` for async imports.
> 
> I don't understand how that is related to `defer` for these static `<script>` tags. Can you elaborate?

`ember-auto-import` uses the location it is loaded from to [auto detect where](https://github.com/ef4/ember-auto-import/blob/d034c9ff12f2770a0434a28589f5e6fa13233b4b/packages/ember-auto-import/ts/webpack.ts#L19-L25) async modules will be loaded from.

I don't know exactly why, but `defer` breaks this. Probably because load order can no longer be determined.

Turbo87:
that is a very surprising hack and definitely using private "API" 😅

NullVoxPopuli:
Has the team discussed this?

Looks like browser support even includes IE10!

fwiw, my specific use-case no longer requires this change, but now the change proposed in this PR feels more like a semantics-update than anything.
Relying on scripts to execute after DOM has been parsed by putting scripts at the end of the body has always felt weird to me. 

If anyone is curious how I'm doing app shell now, it's like this: https://github.com/NullVoxPopuli/emberclear/blob/5cb232c3030b9111f3999a9465ebc9127c0811f2/packages/frontend/app/index.html

I no longer remember if there is still a stutter, or not because my loading animation is super light and CSS-only

NullVoxPopuli:
https://javascript.info/script-async-defer this was a pretty good refresher on the behaviors.

I have an example app using `defer`, `embroider`, and chunk'd scripts here: https://github.com/NullVoxPopuli/ember-embroider-deferred-test

NullVoxPopuli:
@jrjohnson can you describe the need for publicAssetURL a bit more? how does using `defer` force the need for it?
do you have a repro repo up somewhere?

jrjohnson:
This has changed a bit with ember-auto-import v2 and is [documented here](https://github.com/ef4/ember-auto-import#customizing-html-insertion). As far as I can tell EAI detects where to place scripts based on where `vendor.js` gets evaluated. When that evaluation is deferred it isn't possible to detect this and as a result it's necessary to both configure EAI and add an additional attribute to the app's `index.html`. I don't see any reason this couldn't become default behavior if defer was also the new default.

NullVoxPopuli:
atm, the tests need fixed, which means the current load order needs fixed:
image

fixes coming soon (hopefully?, we'll see!)

NullVoxPopuli:
ok, I think _at this time_, we _can't use `defer`.

_something_ is changing what order the scripts are executed in.
image

The output `tests/index.html` scripts say:
```html
    <script defer src="/testem.js" integrity="" data-embroider-ignore></script>
    <script defer src="/assets/vendor.js"></script>
    <script defer src="/assets/test-support.js"></script>
<script src="/assets/chunk.vendors-node_modules_qunit_qunit_qunit_js.46e1ff8612789bc74f9f.js"></script>
<script src="/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js"></script>
    <script defer src="/assets/dummy.js"></script>
    <script defer src="/assets/tests.js"></script>
```

and I added `console.log` to each file so I could verify that the files are executed in the same order they are specified in the html, but the output was:

```
chunk.vendors
chunk.tests 
// ^ these execute first, but they *shouldn't* until test-support.js is executed
vendor.js
test-support.js
dummy.js 
tests.js 
```

Resulting in this error:
```

Uncaught ReferenceError: require is not defined
    <anonymous> Ember
    js http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:20
    __webpack_require__ http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:54
    <anonymous> http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:171
    O http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:91
    <anonymous> http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:173
    <anonymous> http://localhost:7357/assets/chunk.tests.7bd5c31ffc22ed9ebf98.js:176
```

because the chunks are not deferred.

So, a prereq for this PR would be to have embroider/auto-import add defer to their inserted scripts as well.

NullVoxPopuli:
I think this is maybe what needs updated: https://github.com/embroider-build/embroider/blob/631eb5d1b273ac3dde54902fe9c074493f36bc1a/packages/core/src/ember-html.ts#L100

Buuut is it worth the effort? idk 🙃

locks:
It seems like maybe it's not the correct time to affect this change, so I'll close the PR.

PR #7634, opened 2018-02-18
Addon._treeFor optimization

kellyselden:
Without this, an empty array goes through mergeTrees and becomes a valid tree. With this change, it remains a falsy value and fails this check https://github.com/ember-cli/ember-cli/blob/203164d6cc5be60918b39ce7e25530bdd22b739f/lib/broccoli/ember-app.js#L618 meaning less trees to deal with.

This change is only a minor optimization now, but is more noticeable when paired with #7501.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -611,7 +611,7 @@ class EmberApp {
     return this.project.addons.reduce((sum, addon) => {
       if (addon.treeFor) {
         let tree = addon.treeFor(type);
-        if (tree) {
+        if (tree && !mergeTrees.isEmptyTree(tree)) {
           sum.push({
             name: addon.name,
             tree,
--- a/lib/broccoli/merge-trees.js
+++ b/lib/broccoli/merge-trees.js
@@ -94,5 +94,9 @@ module.exports = function mergeTrees(_inputTrees, options) {
   }
 };
 
+module.exports.isEmptyTree = function isEmptyTree(tree) {
+  return tree === EMPTY_MERGE_TREE;
+};
+
 module.exports._overrideEmptyTree = overrideEmptyTree;
 module.exports._upstreamMergeTrees = upstreamMergeTrees;


kellyselden:
That would probably break lots of code.

kellyselden:
Oh our internal mergeTrees helper. Sorry I misread. That might work.

kellyselden:
@rwjblue @stefanpenner Does it make sense that our internal mergeTrees helper might return null? or is that a concern best left to the caller and always return a tree?

lib/models/addon.js
```diff
@@ -515,10 +515,15 @@ let addonProto = {
       trees.push(this.jshintAddonTree());
     }
 
-    let mergedTreesForType = mergeTrees(trees, {
-      overwrite: true,
-      annotation: `Addon#treeFor (${this.name} - ${treeType})`,
-    });
+    let mergedTreesForType;
+    if (trees.length) {
```
Turbo87: can we be explicit ans use `!== 0` here?

Turbo87: actually thinking about it, should we use the fastpath for `length === 1` too?

kellyselden: our `mergeTrees` helper already has the `length === 1` passthrough.

kellyselden: updated to `!== 0`

twokul:
@kellyselden have you tried moving that check into internal mergeTrees function?

kellyselden:
Yes. A lot of code wasn't expecting it, so I would like to make sure it is the right direction from others before fixing up all the code and tests.

stefanpenner:
I think @rwjblue's utility aims to fix this: https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/merge-trees.js

kellyselden:
I could export `function getEmptyTree` and use that to compare against instead of a falsy check. Do you think that would be better @stefanpenner?

lib/models/addon.js
```diff
@@ -515,10 +515,15 @@ let addonProto = {
       trees.push(this.jshintAddonTree());
     }
 
-    let mergedTreesForType = mergeTrees(trees, {
-      overwrite: true,
-      annotation: `Addon#treeFor (${this.name} - ${treeType})`,
-    });
+    let mergedTreesForType;
+    if (trees.length !== 0) {
```
rwjblue: The custom merge trees that we use (lib/broccoli/merge-trees.js) should already be taking care of this, can you check into why it isn't?

tests/unit/models/addon-test.js
```diff
@@ -278,10 +273,9 @@ describe('models/addon.js', function() {
             },
 
             extensionsForType() {
-              return ['js'];
+              return ['template'];
```
rwjblue: This seems odd, why are the test changes required?

lib/models/addon.js
kellyselden: But `EMPTY_MERGE_TREE` is truthy, is it not?

tests/unit/models/addon-test.js
kellyselden: because `Addon.treeFor` can now return a falsy value where is didn't before. `mergeTrees` helper always returns a truthy tree.

kellyselden:
@rwjblue updated with `isEmptyTree` like you suggested.

PR #7562, opened 2018-01-18
[WIP] Default Packager

twokul:

PR #7635, opened 2018-02-19
Packaged Bower

twokul:


lib/broccoli/ember-app.js
```diff
@@ -1040,34 +1044,6 @@ class EmberApp {
...
   }
 
-  /**
-    @private
-    @method _processedBowerTree
-    @return
-  */
-  _processedBowerTree() {
-    if (!this._cachedBowerTree) {
-      // do not attempt to merge bower and vendor together
```
kellyselden: these comments might make sense to come along with the change.

twokul: good point.

PR #7637, opened 2018-02-20
More comprehensive detect if ember-cli is being run within CI or not.

stefanpenner:
For this, we are using: https://github.com/watson/ci-info/

Which has a pretty comprehensive support list: https://github.com/watson/ci-info/#supported-ci-tools

This aims to ensure that console-ui’s new “write errors to disk” doesn’t occur in CI: https://github.com/ember-cli/console-ui/pull/19

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -3,6 +3,7 @@
 const willInterruptProcess = require('../utilities/will-interrupt-process');
 const instrumentation = require('../utilities/instrumentation');
 const getConfig = require('../utilities/get-config');
+const ciInfo = require('ci-info');
 
 let initInstrumentation;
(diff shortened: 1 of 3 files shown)


thoov:
Would it make sense to still allow `process.env.CI` to drive this as a fallback if ciInfo returns false?

stefanpenner:
> Would it make sense to still allow process.env.CI to drive this as a fallback if ciInfo returns false?

Luckily, `ci-info` also checks `process.env.CI` https://github.com/watson/ci-info/blob/master/index.js#L43. So I believe we "get it for free".

thoov:
Ahh, I missed that. Looks good

PR #7646, opened 2018-02-24
Consider all the root js files as a node files

ro0gr:
I think it's safe to say that there shouldn't be frontend files in the root of ember package. So we can probably simplify eslint rule?

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -19,10 +19,8 @@ module.exports = {
   overrides: [
     // node files
     {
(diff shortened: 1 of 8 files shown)


ro0gr:
hm.. eslint started complaining at "tests/test-helper.js" with `Parsing error: 'import' and 'export' may appear only with 'sourceType: module'`

ro0gr:
I've also tried to use relative pattern './*.js' but it doesn't work as well. Probably a related issue https://github.com/eslint/eslint/issues/9953

rwjblue:
I reported https://github.com/eslint/eslint/issues/9740 for this

PR #7649, opened 2018-02-25
only use the standard compilers to compile addon code

kellyselden:
instead of any custom one's added by the app. This will be an optimization for apps that have transpilers like coffeescript, typescript, stylus, emblem, etc. With the merge of #7501, now all these app transpilers run on addon code, which isn't necessary.

The question is, how safe is this? Are these addons guaranteed to be there? Is the registry ordered, could I take the last off the list instead?

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -667,20 +667,35 @@ class EmberApp {
     return this._addonTreesFor(type).map(addonBundle => addonBundle.tree);
   }
 
+  _getDefaultCompiler(type) {
+    // since we are now using the app's registry to compile addon code,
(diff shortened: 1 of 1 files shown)


twokul:
> Is the registry ordered, could I take the last off the list instead?

it seems we just [`push`](https://github.com/ember-cli/ember-cli-preprocessor-registry/blob/master/index.js#L47) onto an array; that might work.

> The question is, how safe is this?

I think it's safe to assume `ember-cli-babel` and `ember-cli-htmlbars`. Am I wrong in thinking that?

kellyselden:
Added comments and changed to getting the last in the array.

PR #7654, opened 2018-02-27
Package Vendor

twokul:

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -3,6 +3,7 @@
 const Funnel = require('broccoli-funnel');
 
 const DEFAULT_BOWER_PATH = 'bower_components';
+const DEFAULT_VENDOR_PATH = 'vendor';
 
 /**
  * Responsible for packaging Ember.js application.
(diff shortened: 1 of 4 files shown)


lib/broccoli/default-packager.js
```diff
@@ -44,12 +46,55 @@ module.exports = class DefaultPackager {
   packageBower(tree, bowerDirectory) {
     if (this._cachedBower === null) {
       this._cachedBower = new Funnel(tree, {
-        srcDir: '/',
```
Turbo87: seems unrelated?

tests/unit/broccoli/default-packager/vendor-test.js
```diff
@@ -0,0 +1,77 @@
...
+  after(co.wrap(function *() {
+    yield input.dispose();
+  }));
+
+  it('caches packaged vendor tree', co.wrap(function *() {
+    let defaultPackager = new DefaultPackager();
+
+    expect(defaultPackager._cachedVendor).to.equal(null);
+
+    yield buildOutput(defaultPackager.packageVendor(input.path()));
```
Turbo87: I think we might have to `output.dispose()`, right?

lib/broccoli/default-packager.js
twokul: it should have been removed in the first place, and I forgot to do it; unnecessary parameter

PR #7655, opened 2018-02-27
Package tests

twokul:

Should be merged after https://github.com/ember-cli/ember-cli/pull/7654

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -152,7 +152,6 @@ class EmberApp {
     // ensure addon.css always gets concated
     this._styleOutputFiles[this.options.outputPaths.vendor.css] = [];
 
-
     this._scriptOutputFiles = {};
     this._customTransformsMap = new Map();
 
@@ -160,7 +159,6 @@ class EmberApp {
     this.legacyTestFilesToAppend = [];
     this.vendorTestStaticStyles = [];
     this._nodeModules = new Map();
-    this._defaultPackager = new DefaultPackager();
 
     this.trees = this.options.trees;
 
@@ -176,6 +174,12 @@ class EmberApp {
     }
 
     this._debugTree = BroccoliDebug.buildDebugCallback('ember-app');
+
+    this._defaultPackager = new DefaultPackager({
+      name: this.name,
+      project: this.project,
+      registry: this.registry,
+    });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rwjblue:
looks like this needs a rebase

lib/broccoli/ember-app.js
```diff
@@ -1189,40 +1170,22 @@ class EmberApp {
...
+    if (!this._cachedExternalTree) {
+      let vendorTrees = this.addonTreesFor('vendor');
 
       if (this.trees.vendor) {
-        trees.push(this.trees.vendor);
+        vendorTrees.push(this.trees.vendor);
       }
 
-      let mergedVendor = mergeTrees(trees, {
+      let vendor = this._defaultPackager.packageVendor(mergeTrees(vendorTrees, {
```
rwjblue: Can you unroll this a bit? Its much easier to grok later...

lib/broccoli/ember-app.js
```diff
@@ -1392,7 +1355,13 @@ class EmberApp {
...
 
   test() {
-    let tests = this.addonPreprocessTree('test', this._processedTestsTree());
+    let addonTrees = this.addonTreesFor('test-support');
+    let mergedTests = mergeTrees(addonTrees.concat(this.trees.tests), {
+      overwrite: true,
+      annotation: 'TreeMerger (tests)',
+    });
+
+    let tests = this.addonPreprocessTree('test', this._defaultPackager.packageTests(mergedTests, this.name));
```
rwjblue: I'd conceptually prefer to have the `packageTests` handle the full thing (e.g. moving `addonPreprocessTree`, `preprocessJs`, and `addonPostprocessTree`)...

tests/unit/broccoli/default-packager/tests-test.js
```diff
@@ -0,0 +1,105 @@
...
+const p = require('ember-cli-preprocess-registry/preprocessors');
+const co = require('co');
+const expect = require('chai').expect;
+const DefaultPackager = require('../../../../lib/broccoli/default-packager');
+const broccoliTestHelper = require('broccoli-test-helper');
+
+const buildOutput = broccoliTestHelper.buildOutput;
+const createTempDir = broccoliTestHelper.createTempDir;
+
+const DEFAULT_DEPENDENCIES = {
```
kellyselden: Would it make sense to read this from something like https://github.com/ember-cli/ember-cli/blob/master/tests/fixtures/app/npm/package.json to prevent more duplication?

twokul: I thought about that but I think we should 🔥 those fixture files so decided to create an object instead. Also, packager will be extracted into a separate package.

kellyselden: I would use it until you extract. So someone doing a PR to update a package in the blueprint, not knowing about this test, will find all/replace the fixtures and wonder if they should touch this test or not.

Also these fixtures are needed to prevent regressing template conditional content (we regressed often without them).

twokul: To be clear, I added this object _only_ to be able to use `p.defaultRegistry()` b/c I didn't want to set up a fake one; it will not break if someone will do find all/replace and has nothing to do with blueprint's dependencies. Having said that, if you feel strongly, I can change.

kellyselden: What dependencies does defaultRegistry require? Could we give it just the deps it needs to succeed?

twokul:
@kellyselden updated

PR #7656, opened 2018-03-01
Extract `contentFor` into `ember-app-utils`

twokul:
Main motivation behind this change is on-going extraction into default
packager. Because, we previously bound `contentFor` to application's
context, it was harder to extract the way we assemble certain trees
(config tree, for example).

This change does the following:

+ removes that binding and replaces it with explicit contract (we pass
now pass in additional `options` object) which makes it easier to test
(no need to set up `EmberApp`) and easier to reason about
+ adds missing tests for both "app-boot" and "test-body-footer"
types
+ removes `contentFor` method from `EmberApp`: not a breaking change
since it is a private method


lib/broccoli/ember-app.js
```diff
@@ -895,6 +894,11 @@ class EmberApp {
     @return
   */
   _configReplacePatterns() {
+    let addons = this.project.addons;
```
rwjblue: To clarify that these variables are only used for passing options to `contentFor` maybe you could tweak this to be an `options` POJO in the format passed below.  It _should_ reduce overall changes, and make it clearer that these are _only_ used for passing to `contentFor` util. 

What do you think?

lib/utilities/ember-app-utils.js
```diff
@@ -50,4 +52,89 @@ function calculateBaseTag(baseURL, locationType) {
...
+ * - 'head-footer'
+ * - 'test-header-footer'
+ * - 'body-footer'
+ * - 'test-body-footer'
+ *
+ * @method contentFor
+ * @param {Object} config Ember.js application configuration
+ * @param {RegExp} mathch Regular expression to match against
+ * @param {String} type Type of content
+ * @param {Object} options Settings that control the default content
```
rwjblue: Can you splat out the options here?

```
* @param {boolean} options.autoRun controls...
```

PR #7657, opened 2018-03-01
Add blueprint for setup-application-test

ryanto:
This gives ember applications a single place to put testing configuration related to application tests.


ryanto:
I tested this out locally and everything works, but I'm wondering what the best way to write a test for this is? I poked test fixtures and looked for other files like the test-helper blueprint, but couldn't find it.

blueprints/app/files/tests/helpers/setup-application-test.js
```diff
@@ -0,0 +1,5 @@
+import { setupApplicationTest } from 'ember-qunit';
+
+export default function(hooks) {
+  setupApplicationTest(hooks);
```
bendemboski: This seems like a good place to put a comment telling users why this file exists and what to do here, similar to [this](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/ember-cli-build.js#L10). Good opportunity to educate, and also avoid confusion over this apparently extraneous file.

blueprints/app/files/tests/helpers/setup-application-test.js
```diff
@@ -0,0 +1,5 @@
+import { setupApplicationTest } from 'ember-qunit';
+
+export default function(hooks) {
```
Turbo87: I thought you wanted to make it a named export? 🤔

ryanto: Yup, I'll change to 1 file (`test-setup-functions.js`) with 3 exports.

ryanto:
Ok based on some discussion today...

- The file that exports the setup functions will be `tests/helpers/index.js`
- There will be comments in the file explaining why it exists and how you can use it
- It will export functions for each type of test, but the rendering and test setup functions will be re-exports, application will create a new function.
- Alias the application setup as `upstreamSetupApplicationTest`

Turbo87:
FWIW I would have appreciated it if we had figured this out before updating all of the blueprints in `ember-source` because we will now have to do it all over again... :-/

rondale-sc:
In ember-mocha you need to call `upstreamApplicationTest()` before you define any lifecycle hook (ie `afterEach` and `beforeEach`).  If you don't `this.owner` (and more) are undefined.

Turbo87:
yeah, agreed, that `upstreamApplicationTest()` call should definitely be first

ryanto:
Awesome, thanks for bringing this to my attention. I'll get that change in.

locks:
It was brought to my attention that there is an unimplemented RFC that goes in this direction: https://github.com/emberjs/rfcs/blob/master/text/0637-customizable-test-setups.md. I have opened an issue on this repository to restart the implementation, since this PR is quite outdated by now. Thank you for the help!

PR #7658, opened 2018-03-02
Module Unification Addon blueprint

cibernox:
This adds the ability of generating addons using the module unification layout.

MODULE_UNIFICATION=true ember addon my-addon
MODULE_UNIFICATION=true ember addon my-addon --yarn

@GavinJoyce @mixonic

### DIFF
--- a/blueprints/module-unification-addon/files/.travis.yml
+++ b/blueprints/module-unification-addon/files/.travis.yml
@@ -1,0 +1,47 @@
+---
+language: node_js
+node_js:
+  # we recommend testing addons with the same minimum supported node version as Ember CLI
+  # so that your addon works for all apps
+  - "4"
+
+sudo: false
+dist: trusty
+
+addons:
+  chrome: stable
+
+cache:
+  directories:
+    - $HOME/.npm
+
+env:
+  global:
+    # See https://git.io/vdao3 for details.
+    - JOBS=1
+  matrix:
+    # we recommend new addons test the current and previous LTS
+    # as well as latest stable release (bonus points to beta/canary)
+    - EMBER_TRY_SCENARIO=ember-lts-2.12
+    - EMBER_TRY_SCENARIO=ember-lts-2.16
+    - EMBER_TRY_SCENARIO=ember-lts-2.18
+    - EMBER_TRY_SCENARIO=ember-release
+    - EMBER_TRY_SCENARIO=ember-beta
+    - EMBER_TRY_SCENARIO=ember-canary
+    - EMBER_TRY_SCENARIO=ember-default
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - env: EMBER_TRY_SCENARIO=ember-canary
+
+before_install:
+  - npm config set spin false
+  - npm install -g npm@4
+  - npm --version
+
+script:
+  - npm run lint:js
+  # Usually, it's ok to finish the test scenario without reverting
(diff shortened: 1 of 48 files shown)


blueprints/module-unification-addon/index.js
```diff
@@ -0,0 +1,29 @@
...
+    let name = stringUtil.dasherize(rawName);
+    let namespace = stringUtil.classify(rawName);
+    let addonName = stringUtil.dasherize(rawName);
+
+    return {
+      name,
+      modulePrefix: name,
+      namespace,
+      addonName,
+      // addonModulePrefix: addonName,
```
cibernox: It's not clear to me the difference between `addonModulePrefix` and `addonName` and which one should I use.

cibernox:
Can someone rebuild this? I think it's a random failure.

blueprints/module-unification-addon/index.js
mixonic: I don't see `addonModulePrefix` used anywhere in these files.

cibernox: It is not, I decided to use addon name. But I wanted to understand if there was a good reason for that weird naming

tests/fixtures/module-unification-addon/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+The MIT License (MIT)
+
+Copyright (c) 2018
```
ro0gr: it will fail each year :)

tests/fixtures/module-unification-addon/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "test": "ember test",
+    "test:all": "ember try:each"
+  },
+  "dependencies": {
+    "ember-cli-babel": "^6.6.0"
+  },
+  "devDependencies": {
+    "broccoli-asset-rev": "^2.4.5",
+    "ember-ajax": "^3.0.0",
+    "ember-cli": "~3.1.0-beta.1",
```
ro0gr: it will fail each release

cibernox: The test is testing that the same files are being created, not that the contents of those files is the same. Afaik, all other tests are similar.

ro0gr: oh, right. there is no verification of result files contents

blueprints/module-unification-addon/index.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+const addonBlueprint = require('../addon');
+
+module.exports = Object.assign({}, addonBlueprint, {
```
cibernox: This is where it all happens. The only difference with the generate of standard addons is the description and the files. I couldn't find a better way of "inheriting" from an existing blueprint.

blueprints/module-unification-addon/files/.travis.yml
```diff
@@ -0,0 +1,47 @@
+---
```
rwjblue: Seems weird to have _both_ `.travis.yml` and `travis.yml` in the diff here

rwjblue: seems like all of the `.foo` files have this issue?

rwjblue:
Chatted with @mixonic in slack, we agree that this can land and changing the default dummy app setup to be module unification should happen in a follow up PR.

PR #7660, opened 2018-03-02
improve logic for if addon is module-unification

iezer:
- required so that a module-unification app can handle a mix of addons
   which are module-unification (in which case they have a src folder
   but no app folder), or classic (in which case they have an app folder).


rwjblue:
Thank you @iezer!

PR #7663, opened 2018-03-06
lint vendor js files as AMD

kellyselden:


kiwiupover:
I agree with @rwjblue 
if a dev wants to add vendor to their linking path then they can to hat themselves

kellyselden:
Should we add an exclusion to our .eslintrc to not lint vendor? The reason I ask is, even though our "lint:js" script doesn't include vendor, any editor with an eslint plugin will have red lines everywhere when looking at a vendor file.

Turbo87:
https://github.com/ember-cli/ember-cli/pull/7769 added an `.eslintignore` where we could/should add `vendor`. @kellyselden can you update the PR?

sandstrom:
I'm doing some issue gardening 🌱🌿 🌷 and came upon this issue. Since it's quite old I just wanted to ask if this is still relevant? If it isn't, maybe we can close this issue?

By closing some old issues we reduce the list of open issues to a more manageable set.

PR #7664, opened 2018-03-07
Support serving wasm with application/wasm

krisselden:
The fastest way to load wasm is WebAssembly.instantiateStreaming
but it requires the mime-type to be correct.

I can't update the mime module because it has engines >=6.

I updated broccoli-middleware but have to manually define it for testem.


stefanpenner:
@krisselden LGTM, is there a test we can add do ensure we don't regress?

krisselden:
I'll look for a spot somewhere to add a test, do you know why YUI doc would be failing?

twokul:
@krisselden is there a reason not to bump `broccoli-middleware` in `package.json`?

just looked into errors - message is misleading, it has something to do with the way we check the content of the response in tests (`response.body.toString()` is `[object Object]`); changing it to `response.text.trim()` works fixes it (not sure why, maybe compatibility issues with the version of `express`?)

rwjblue:
The failures should been fixed by broccoli-middleware@1.2.1 (they were legit failures). 

Agree with @twokul, we should bump the minimum broccoli-middleware version to 1.2.1 as part of this PR...

krisselden:
Once tests pass I think this is good now. Added tests for both the express side and the testem side and the smoke test ensures it works with WebAssembly.instantiateStreaming

lib/commands/test.js
```diff
@@ -7,6 +7,9 @@ const SilentError = require('silent-error');
 const path = require('path');
 const Win = require('../utilities/windows-admin');
 const existsSync = require('exists-sync');
+
+require('express').static.mime.define({ 'application/wasm': ['wasm'] });
```
Turbo87: is there a reason why we do this in `lib/commands/test.js`? seems unrelated to me 🤔

rwjblue: The file serving that we do is already WASM aware (due to changes in broccoli-middleware), this specific line is ensuring that _testem_'s server also uses the correct mime type.  

tldr; its here because its specifically to support testem...

Turbo87: could use a comment though...

PR #7667, opened 2018-03-07
MU addons must generate a MU dummy app

cibernox:
/cc @rwjblue @mixonic


tests/fixtures/module-unification-addon/tests/dummy/config/targets.js
```diff
@@ -1,18 +1,8 @@
...
-const isCI = !!process.env.CI;
-const isProduction = process.env.EMBER_ENV === 'production';
-
-if (isCI || isProduction) {
-  browsers.push('ie 11');
-}
-
 module.exports = {
-  browsers
+  browsers: [
```
cibernox: Looking at this diff I've realized that the module-unification-app blueprint is outdate compared with the regular `app` blueprint. We need to come up with a system that reuses as much as possible from the standard blueprint and only swaps the indispensable files, so the chances of this happening again is lower.

ro0gr: > module-unification-app blueprint is outdate compared with the regular app blueprint

https://github.com/ember-cli/ember-cli/pull/7595 aims to address this by re-using the same `config/targets.js` fixture for both `app` and `module-unification-app` blueprints testing. 

Also I've been working on reducing duplication between these blueprints(https://github.com/ember-cli/ember-cli/pull/7500) but it was blocked due to bad testing story(I believe).

tests/fixtures/module-unification-addon/package.json
```diff
@@ -40,13 +40,12 @@
     "ember-load-initializers": "^1.0.0",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-resolver": "^4.0.0",
-    "ember-source": "~3.1.0-beta.1",
+    "ember-source": "http://builds.emberjs.com/canary.tgz",
```
mixonic: This URL is out of date. I don't think we should ever be using it, /cc @rwjblue yes?

tests/fixtures/module-unification-addon/tests/dummy/src/resolver.js
```diff
@@ -0,0 +1,13 @@
+import Resolver from 'ember-resolver/resolvers/fallback';
```
mixonic: I suggest we should have this use the glimmer wrapper instead of the fallback resolver:

```js
import Resolver from 'ember-resolver/resolvers/glimmer-wrapper';
```

The glimmer wrapper only implements module unification rules. The fallback is only needed if you have classic addons.

tests/fixtures/module-unification-addon/package.json
mixonic: Also at https://github.com/ember-cli/ember-cli/blob/e9a7b587a1ebd063985816330fd1a193652cec54/blueprints/module-unification-app/files/package.json#L39 for MU apps.

rwjblue: Confirm. We need to fetch the current URL each time (via `ember-source-channel-url` which returns a promise)

tests/fixtures/module-unification-addon/tests/dummy/src/resolver.js
rwjblue: I think all apps have classic addons (literally all addons are “classic” right now).

mixonic: The dummy app of an brand new module unification addon should not have an classic addons.

rwjblue: Definitely seems ideal, but I highly doubt it is the current state.

tests/fixtures/module-unification-addon/package.json
cibernox: @rwjblue Any guidance how to get the `ember-source-channel-url` at build time? Specially about how to handle it in testing, where we shouldn't rely on the network.

tests/fixtures/module-unification-addon/tests/dummy/src/resolver.js
cibernox: What is the consensus here?

mixonic: Ok! We had some more discussion. Lets stick with the fallback resolver. We should use the fallback setup as described at https://github.com/rwjblue/ember-module-migrator#running-module-unification-with-fallback-to-classic-app-layout. We want to use this setup for MU apps since there are still classic addons in the default `package.json` and it is reasonable to expect they work out of the box.

I suggest that we should remove the fallback setup before we remove any feature flags here. I've added that step to our blocker list for removing the feature flags.

tests/fixtures/module-unification-addon/tests/dummy/src/main.js
```diff
@@ -0,0 +1,14 @@
...
+import loadInitializers from "ember-load-initializers";
+import config from "../config/environment";
+
+const App = Application.extend({
+  modulePrefix: config.modulePrefix,
+  podModulePrefix: config.podModulePrefix,
+  Resolver
+});
+
+loadInitializers(App, config.modulePrefix + "/src/init");
```
mixonic: Per the fallback setup instructions at https://github.com/rwjblue/ember-module-migrator#running-module-unification-with-fallback-to-classic-app-layout another `loadInitializers` line is needed here that loads addon initializers in the `app` folder.

mixonic: This same addition is needed at https://github.com/ember-cli/ember-cli/blob/e9a7b587a1ebd063985816330fd1a193652cec54/blueprints/module-unification-app/files/src/main.js#L12

mixonic:
In https://github.com/ember-cli/ember-cli/blob/master/blueprints/module-unification-addon/files/tests/test-helper.js#L1 the import must be from `../src/main`

tests/fixtures/module-unification-addon/tests/dummy/src/resolver.js
cibernox: Added!

tests/fixtures/module-unification-addon/tests/dummy/src/main.js
cibernox: Added

mixonic:
@cibernox yesterday @rwjblue was eyeing the `ember-source-channel-url` stuff. He acknowledged it was non-trivial ;-)

I'm going to merge this as a good incremental improvement but we do need the `ember-source-channel-url` stuff for it to really work. The `ember-canary.tgz` file is hopelessly out of date.

mixonic:
Thank you @cibernox!!

PR #7668, opened 2018-03-07
Only watch test/dummy/app on addons if it exist

cibernox:
Related with #7667 

Without this change, the default generated addon fails on `ember serve`


cibernox:
Hold on. Don't merge this, I believe we need more changes and I might be able to squeeze them in this same PR.

cibernox:
@rwjblue can you merge this?

Issue #7642, opened 2018-02-22
Deprecate ember-cli-babel 5


kellyselden:
The new work we are doing to support packaging and tree-shaking won't work with babel 5. We are also probably going to remove babel 5 support in a minor version of ember-cli, as we determined we don't have a babel version support matrix as part of our semver policy.

Dhaulagiri:
Clarifying for interested spectators like myself, this would mean any addons using Babel 5 would no longer work in the consuming app once this support is removed?

Turbo87:
Correct, but in reality updating an addon to Babel 6 is actually not that hard since most of the hard work is abstracted away in `ember-cli-babel`.

Dhaulagiri:
definitey. I’ve been able to update most of our addons at this point so now it’s a matter of the less maintained ones where getting any PRs merged and released is hard to do 🌵

kellyselden:
Completed on master.

PR #7676, opened 2018-03-12
Deprecate ember-cli-babel 5.x

raytiley:
This renames the `checkEmberCliShims` to `_checkEmberCliBabel`. The return value still represents if shims are required, but while recursing the addons a warning is logged for any `ember-cli-babel` versions less then `6.x`.

Implements #7642


kellyselden:
Thanks @raytiley

raytiley:
My pleasure.

PR #7679, opened 2018-03-13
Update init to be src/ friendly

mixonic:
I need some testing solution for this, but I'd like to try and get `init` working with module unification today.

Without `init` working, there isn't a good canonical way for someone who uses the ember-module-migrator to get the remaining setup files in place.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -32,8 +32,14 @@ module.exports = Command.extend({
 
   _defaultBlueprint() {
     if (this.project.isEmberCLIAddon()) {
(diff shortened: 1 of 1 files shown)


lib/commands/init.js
```diff
@@ -32,9 +32,17 @@ module.exports = Command.extend({
 
   _defaultBlueprint() {
     if (this.project.isEmberCLIAddon()) {
-      return 'addon';
+      if (this.project.isModuleUnification()) {
+        return 'module-unification-addon';
+      } else {
```
rwjblue: These else + returns are failing linting

PR #7684, opened 2018-03-16
Remove ember-data from default blueprint

runspired:
At our most recent face-to-face, the `ember-data` team discussed the learning story for `ember-data` and how it fits into the larger picture of the `ember` ecosystem learning story.

We determined that we should remove `ember-data` from the default blueprint for new applications in order to reduce the mental overhead of creating a new application, to improve the ability for the learning team to introduce concepts incrementally, to reduce the default file-size of a new Ember application, and to clarify that using `ember-data` is not a requirement of using `ember`.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -31,7 +31,6 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
-    "ember-data": "~3.1.0-beta.1",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.0.0",
     "ember-maybe-import-regenerator": "^0.1.6",
--- a/tests/fixtures/app/npm/package.json
+++ b/tests/fixtures/app/npm/package.json
@@ -31,7 +31,6 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
-    "ember-data": "~3.1.0-beta.1",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.0.0",
     "ember-maybe-import-regenerator": "^0.1.6",
--- a/tests/fixtures/app/yarn/package.json
+++ b/tests/fixtures/app/yarn/package.json
@@ -31,7 +31,6 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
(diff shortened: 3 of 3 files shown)


jenweber:
Can this be RFC'd? We have a lot of work to do on the learning resources side, and it seems there are some enterprise impacts that I'll let other people chime in on.

runspired:
@jenweber I believe we have resolved the impacts you refer to. In terms of the learning side, could you give us a list of what needs to be done? We're happy to help and happy to hold off until that work is complete.

twokul:
Before merging, I think we need to make sure we communicate the reasoning clearly: blog post, changes to the website if needed, potential impact, and release of CLI with the change.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kellyselden:
I would like to debate this in RFC land before merging as well. As I wasn't in the ember-data meeting, I have no idea the thought process that lead to this. On one hand, we have RFCs like [adding service worker addons](https://github.com/ember-cli/rfcs/pull/117) to the default blueprint to make ember a more complete experience from the get go, then on the other extreme we are removing ember-data which is vital to most to make ember a complete solution.

The tldr; is we should have a determining factor of what makes a default addon, and apply it to all addons, whether it be ember-data, service-workers, fastboot, etc.

tomdale:
This is a large, fundamental change and should definitely be opened to the RFC process. I would love to see the discussion from the in-person meeting seed that RFC, so that the community and other teams can weigh in.

stefanpenner:
This aligns with the "install your way to a complete app" ideas. I suspect the messaging of this should be made clear though, we wouldn't want people to run to the hills when it comes to ED. Ad-hoc attempts at ED often lead people down an sad path.

wycats:
I very strongly agree with @kellyselden.

runspired:
> The tldr; is we should have a determining factor of what makes a default addon, and apply it to all addons, whether it be ember-data, service-workers, fastboot, etc.

While I agree with this sentiment, this seems like a separate RFC that the ember-cli and learning folks should tackle.

> On one hand, we have RFCs like adding service worker addons to the default blueprint to make ember a more complete experience from the get go, then on the other extreme we are removing ember-data which is vital to most to make ember a complete solution.

Best left to RFC discussion but a casual TL;DR would be does an addon answer `yes` to "Is this complete experience zero-config?".  Unlike many of the addons recommended by default, `ember-data` requires substantial configuration and customization for many teams to use.

runspired:
After discussing this further, we've decided to shelve this desire until such time as:

- `ember-cli` has established guidance on what belongs in the default blueprint
- `ember-data` has published packages allowing for more incremental adoption of its concepts.
- The learning team has established a learning story that incrementally introduces ember-data

PR #7685, opened 2018-03-17
fix the shims messaging

kellyselden:
It was missing the termination condition when `addons` is empty causing it to always be false.


stefanpenner:
do we need a test?

rwjblue:
Yeah, probably

kellyselden:
Added a test.

PR #7686, opened 2018-03-17
WIP: Promote DELAYED_TRANSPILATION experiment

kellyselden:
I tested babel 5 in both an app and addon with DELAYED_TRANSPILATION on, and it works. I thought it would crash and burn, which is why I added the experiment, but it didn't. So we can promote it out of experiment now.

cc @rwjblue @twokul


kiwiupover:
Good work @kellyselden

Turbo87:
@kellyselden @rwjblue I've updated the PR to enable the experiment by default

rwjblue:
❤️

PR #7695, opened 2018-03-21
Guide new addons away from jQuery dependency

ef4:
We already have consensus in a [merged RFC](https://github.com/emberjs/rfcs/blob/master/text/0294-optional-jquery.md) that we want addons to not depend on jQuery whenever possible.

This is a change to the default addon blueprint that will make jQuery opt-in, rather than opt-out.

I realize there is a longer-term plan that would clean this up even further by creating an API clearer than `vendorFiles` for controlling jQuery dependency, but this seems like a stepwise improvement.

### DIFF
--- a/blueprints/addon/files/ember-cli-build.js
+++ b/blueprints/addon/files/ember-cli-build.js
@@ -4,6 +4,13 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function(defaults) {
   let app = new EmberAddon(defaults, {
+    /* 
+      Leave jQuery out of this addon's own test suite & dummy app by default, 
+      so that the addon can be used in apps without jQuery. If you really need 
+      jQuery, it's safe to remove this line.
+    */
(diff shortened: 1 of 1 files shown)


cibernox:
Perhaps a new entry for "@ember/optional-features" will be more user friendly than this config going forward.

kellyselden:
Will the default testing blueprints be jQuery-less by the time this hits stable?

cibernox:
@kellyseldel I thought they already were. Isn't the new testing API the default?

kellyselden:
On most recent stable (3.0.2), generating a component test gives me `
  assert.equal(this.$().text().trim(), '');`

cibernox:
@rwjblue ^ Is this an oversight in the blueprints or new tests are not enabled because any legit reason?

rwjblue:
Hmm, wat!?!?! Can you push up that repo? I'm pretty surprised...

kellyselden:
My mistake. The ember and ember-cli versions were mismatched. New blueprints are being generated.

PR #7698, opened 2018-03-23
BUGFIX upgrading Ember 2.18 apps doesn't add find() test helper

CodingItWrong:
When upgrading an Ember 2.18 app to Ember 3.0, the version of ember-cli-qunit is not updated from `^4.1.1`. If there is already a package lockfile, then the version of it won't be updated, and this means the find() helper won't be available in the Ember 3.0 style of testing.

This PR updates ember-cli-qunit in the blueprint package.json to `^4.3.2`, to ensure that find() will be available in upgraded projects.

This problem doesn't occur for newly-created projects or for projects that don't have an NPM or Yarn package lockfile, because the version constraint `^4.1.1` _will_ pull down `4.3.2` or newer.


rwjblue:
Seems good to me!

PR #7700, opened 2018-03-23
Passing defaultOptions to testem to prevent the cwd and config_dir set in testem.js from being overridden by ember-cli

arthirm:

https://github.com/testem/testem/pull/1219

### DIFF
--- a/lib/tasks/test-server.js
+++ b/lib/tasks/test-server.js
@@ -9,13 +9,7 @@ class TestServerTask extends TestTask {
   invokeTestem(options) {
     let task = this;
     return new Promise((resolve, reject) => {
-      if (task.testem.setDefaultOptions) {
-        task.testem.setDefaultOptions(task.testemOptions(options));
-      } else {
-        // To make it backward compatibile with older versions of testem (2.0.0 and lower)
-        options = task.testemOptions(options);
-      }
-
+      task.testem.setDefaultOptions(task.testemOptions(options));
       task.testem.startDev(options, (exitCode, error) => {
         if (error) {
           reject(error);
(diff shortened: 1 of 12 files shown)


tests/unit/tasks/test-server-test.js
```diff
@@ -45,6 +45,46 @@ describe('test server', function() {
     watcher.emit('change');
   });
 
+  it('transforms and sets defaultOptions in testem and invokes testem properly', function(done) {
```
stefanpenner: i would prefer if we just use promises here, rather then a mix of `done` and promises. The promise mode for mocha will nicely handle errors for us.

lib/tasks/test-server.js
```diff
@@ -8,9 +8,15 @@ const SilentError = require('silent-error');
 class TestServerTask extends TestTask {
   invokeTestem(options) {
     let task = this;
-
     return new Promise((resolve, reject) => {
-      task.testem.startDev(task.testemOptions(options), (exitCode, error) => {
+      if (task.testem.setDefaultOptions) {
+        task.testem.setDefaultOptions(task.testemOptions(options));
```
stefanpenner: this looks good

lib/tasks/test-server.js
```diff
@@ -8,9 +8,15 @@ const SilentError = require('silent-error');
...
+      } else {
+        // To make it backward compatibile with older versions of testem (2.0.0 and lower)
```
stefanpenner: I don't think we need this branch, as ember-cli will need to bump its minimum testem version to one that supports this feature

johanneswuerbach:
I’m currently on vacation, but I’ll cut a new version once I’m back in 2 days

arthirm:
Fixing the issue in https://github.com/ember-cli/ember-cli/pull/7728 and closing this PR.

PR #7707, opened 2018-03-29
Added package-info-cache (fast package caching)

dcombslinkedin:
This is near completion, but I wanted to get a WIP version in before the final attempt so that @stefanpenner and @rwjblue could take a look and suggest improvements. (NOTE: this is being entered against master, but original development was against 3.0.1.) Here's the feature description:

Doing profiling of build time processing for a large project at LinkedIn found that the addon discovery process was taking over 50 seconds just to determine all where all the addon packages were on disk and read and parse their package.json files. I found that much of this was due to repeated validation of the path to the addon package.json and index.js files (fs.realpathSync turns out to be fairly expensive when done a lot, and the project was calling it something like 80K times). The implementation was already re-reading the package information and then throwing it away many times.

This PR creates a new PackageInfoCache class, with a singleton instance added to either CLI (if the build is being done through the CLI or an individual project (if the build is done on a single project but not through the CLI). If a build, like ours here at LinkedIn, includes multiple projects, the fact the instance is in CLI means they all take advantage of the same cache (I'll call it PIC from here out) and don't need to rediscover any packages.

When a Project (ember-cli/lib/models/project) is constructed, a reference to a local or CLI version of PIC is made and PIC.loadProject(projectPath) is called. loadProject then creates an internal instance of a PackageInfo class for the project. It parses the package.json file, then creates separate ProjectInfo instances for all internal addons, all CLI in-repo addons and all 'dependency' and 'devDependency' entries. In the process of that it does the same thing that the node resolution algorithm does to find package.json files, but tries as much as possible to not use 'fs.realpathSync' more than once per file path, caching the result of each path resolution for speed. 

Because of the cache, reading all the dependencies, parsing their package.json files, determining the addons and collecting information about them (prior to addon-instantiation), current performance on my Mac Pro machine (new 2017 model) has gone from 54 seconds down to 900 milliseconds. Obviously the precise performance will differ based on machine size, but overall it's about a 60x increase.

The result of all the above processing is that the PackageInfoCache contains a number of entries, one per package for each real file path (i.e. its cache entries are keyed on the real path for a package's root directory--all symlinks and .. entries are resolved). A PackageInfo object contains the following fields of interest:

* pkg - the object resulting from the package.json file
* realPath - (string) the resolved path for the package's root directory
* cache - a reference to the PIC object that contains the packageInfo instance
* errors - an ErrorList object containing ErrorEntry objects, one per error discovered during the processing of the package and all its various dependencies. When a packageInfo object is being created, if the package itself is bad (e.g. no or malformed package.json file) the error is created and put in the parent. If the package is generally constructed correctly, anything wrong with the contents of the object itself (e.g. one of its dependencies cannot be found), the error is registered with the object. The PIC has a method 'showErrors' to dump a list of all the errors in all the PackageInfo objects. The caller could then decide if any of those errors is enough to fail the build.

Various types of PackageInfo objects can have other fields:
* addonConstructor - filled out but not yet used, this would replace part of addons-factory so that constructing an instance of an addon would be done through the PIC/PackageInfo, simplifying the code
* addonMainPath - for an addon package, the real path to the package's 'main' file
* inRepoAddons - an array of PackageInfo objects, one per in-repo package definition (note that the in-repo addon definition is included in the PIC-level cache by its file path already, this is just a reference to that same set of objects)
* internalAddons - same as above, but for internal addons (project only)
* cliInfo - same as above, but for the package that corresponds to the path <pkg>.cli.root
* dependencyPackages - a JS object keyed by dependency name of references to PackageInfo objects, one per entry in the package's 'dependency' attribute list)
* devDependencyPackages - similar, but for the 'devDependency' attribute list
* nodeModules - a NodeModulesList object (basically contains an array of PackageInfo and other NodeModulesList objects) of all packages and subdirectories in a node_modules directory.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -75,7 +75,7 @@ class CLI {
       initInstrumentation: options.initInstrumentation,
     });
 
-    this._packageInfoCache = new PackageInfoCache();
+    this._packageInfoCache = new PackageInfoCache(this.ui);
 
     logger.info('testing %o', !!this.testing);
   }
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -1,17 +1,17 @@
-'use strict';
+"use strict";
 
-const RSVP = require('rsvp');
+const RSVP = require("rsvp");
 
-const lookupCommand = require('./lookup-command');
-const getOptionArgs = require('../utilities/get-option-args');
-const logger = require('heimdalljs-logger')('ember-cli:cli');
-const loggerTesting = require('heimdalljs-logger')('ember-cli:testing');
-const Instrumentation = require('../models/instrumentation');
+const lookupCommand = require("./lookup-command");
+const getOptionArgs = require("../utilities/get-option-args");
+const logger = require("heimdalljs-logger")("ember-cli:cli");
+const loggerTesting = require("heimdalljs-logger")("ember-cli:testing");
+const Instrumentation = require("../models/instrumentation");
 const PackageInfoCache = require("../models/package-info-cache");
-const heimdall = require('heimdalljs');
+const heimdall = require("heimdalljs");
 
 const Promise = RSVP.Promise;
-const onProcessInterrupt = require('../utilities/will-interrupt-process');
+const onProcessInterrupt = require("../utilities/will-interrupt-process");
 
 class CLI {
   /**
@@ -70,14 +70,16 @@ class CLI {
      * @private
      * @property instrumentation
      */
-    this.instrumentation = options.instrumentation || new Instrumentation({
-      ui: options.ui,
-      initInstrumentation: options.initInstrumentation,
-    });
+    this.instrumentation =
+      options.instrumentation ||
+      new Instrumentation({
+        ui: options.ui,
+        initInstrumentation: options.initInstrumentation
+      });
 
     this._packageInfoCache = new PackageInfoCache(this.ui);
 
-    logger.info('testing %o', !!this.testing);
+    logger.info("testing %o", !!this.testing);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


stefanpenner:
Thanks @dcombslinkedin, i'll review tonight/tomorrow.

As discussed, we will need to add appropriate unit tests.

lib/models/addons-factory.js
```diff
@@ -35,9 +35,15 @@ class AddonsFactory {
     logger.info('     addon names are:', Object.keys(addonPackages));
 
     for (let name in addonPackages) {
+      console.log(`### factory.initAddons for '${name}'`);
       addonInfo = addonPackages[name];
       emberAddonConfig = addonInfo.pkg['ember-addon'];
 
+      if (!emberAddonConfig) {
```
stefanpenner: this seem suspicious, how would other addons positions themselves before/after an existing addon, if its ember-addon field was empty/missing?

lib/models/addons-factory.js
```diff
@@ -35,9 +35,15 @@ class AddonsFactory {
     logger.info('     addon names are:', Object.keys(addonPackages));
 
     for (let name in addonPackages) {
+      console.log(`### factory.initAddons for '${name}'`);
```
stefanpenner: before merging, these logs will have to be removed, or switched to use heimdall-logger/debug

lib/models/project.js
```diff
@@ -416,9 +446,54 @@ class Project {
     @method discoverAddons
    */
   discoverAddons() {
-    let addonsList = this.addonDiscovery.discoverProjectAddons(this);
+    console.log(" ");
+    console.log(`discoverAddons for root ${this.root}, pkg ${this.pkg.name}.`);
```
stefanpenner: before merging, these logs will have to be removed, or switched to use heimdall-logger/debug

lib/models/project.js
```diff
@@ -416,9 +446,54 @@ class Project {
...
+    if (projectPackageInfo.inRepoAddons) {
+      addonPackageList.push(...projectPackageInfo.inRepoAddons);
+    }
+
+    this.addonPackages = {};
 
-    this.addonPackages = this.addonDiscovery.addonPackages(addonsList);
+    addonPackageList.forEach(packageInfo => {
+      let relpath = path.relative(this.root, packageInfo.realPath);
+      console.log(`  ${packageInfo.name}: relpath: ${relpath}`);
```
stefanpenner: before merging, these logs will have to be removed, or switched to use heimdall-logger/debug

lib/models/project.js
```diff
@@ -416,9 +446,54 @@ class Project {
...
+      let addonPkgs = Object.values(projectPackageInfo.devDependencyPackages)
+          .filter(packageInfo => packageInfo.isAddon());
+      addonPackageList.push(...addonPkgs);
+    }
+
+    if (projectPackageInfo.inRepoAddons) {
+      addonPackageList.push(...projectPackageInfo.inRepoAddons);
+    }
+
+    this.addonPackages = {};
```
stefanpenner: `this.addonPackages = Object.create(null)``

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
+/**
```
stefanpenner: `'use strict'`

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+ * Performance cache for information about packages (projects/addons/"apps"/modules)
+ * under an initial root directory and resolving addon/dependency links to other packages.
+ */
+var fs = require("fs-extra");
```
stefanpenner: `const`

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+/**
+ * Resolve the real path for a file, return null if does not
+ * exist or is not a file or FIFO, return the real path otherwise.
+ *
+ * @private
+ * @method getRealFilePath
+ * @param  {String} file file path
+ * @return {String} real path or null
+ */
+function getRealFilePath(file) {
```
stefanpenner: this function needs unit test coverage

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+/**
+ * Resolve the real path for a directory, return null if does not
+ * exist or is not a directory, return the real path otherwise.
+ *
+ * @private
+ * @method getRealDirectoryPath
+ * @param  {String} file file path
+ * @return {String} real path or null
+ */
+function getRealDirectoryPath(dir) {
```
stefanpenner: this function needs unit test coverage

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+const PACKAGE_JSON = "package.json";
+
+const ERROR_PACKAGE_DIR_MISSING = "packageDirectoryMissing";
+const ERROR_PACKAGE_JSON_MISSING = "packageJsonMissing";
+const ERROR_PACKAGE_JSON_PARSE = "packageJsonParse";
+const ERROR_EMBER_ADDON_MAIN_MISSING = "emberAddonMainMissing";
+const ERROR_DEPENDENCIES_MISSING = "dependenciesMissing";
+const ERROR_NODEMODULES_ENTRY_MISSING = "modulesEntryMissing";
+
+class PackageInfo {
```
stefanpenner: this class needs test coverage, likely unit

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+      };
+
+      this.addonConstructor = ctor;
+    }
+
+    return this.addonConstructor;
+  }
+}
+
+class NodeModulesList {
```
stefanpenner: this class needs test coverage

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+
+    return val;
+  }
+}
+
+/**
+ * Class that stores entries that are either PackageInfo or NodeModulesList objects.
+ * The entries are stored in a map keyed by real directory path.
+ */
+class PackageInfoCache {
```
stefanpenner: this class needs test coverage

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+      console.warn(obj.realPath + ":");
+      rootPath = obj.realPath;
+    }
+
+    let errorList = obj.errors;
+
+    errorList.getErrors().forEach(errorEntry => {
+      switch(errorEntry.type) {
+      case ERROR_PACKAGE_DIR_MISSING:
+        console.warn(`   Missing package directory at relative path '${path.relative(rootPath, errorEntry.data)}'`);
```
stefanpenner: currently we use console-ui to interact with the console. I suspect this will also need to.

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+class PackageInfo {
+  constructor(pkgObj, realPath, cache) {
+    this.pkg = pkgObj;
+    this.pkg['ember-addon'] = this.pkg['ember-addon'] || {};
+    this.realPath = realPath;
+    this.cache = cache;
+    this.errors = new ErrorList();
+
+    // other fields that will be set as needed:
+    //  this.addonConstructor (ctor function or object - addons only)
```
stefanpenner: we should likely set then as `undefined` or `null` in the constructor, so the JIT can shape these objects all the same.

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,810 @@
...
+  }
+
+  // XXX Should this move to Addon.js (and change appropriately to use the cache here?)
+  getAddonConstructor() {
+    if (this.isAddon() && !this.addonConstructor) {
+      if (!this.addonMainPath) {
+        return null;
+      }
+
+      let module = require(this.addonMainPath);
```
stefanpenner: not for this PR, but future work. We should likely instrument the time it takes to require an addon, and set a budget. This will help ensure loading commands isn't too slow.

stefanpenner:
> 54 seconds down to 900 milliseconds. Obviously the precise performance will differ based on machine size, but overall it's about a 60x increase.

I like these numbers!

lib/models/addons-factory.js
dcombslinkedin: Re this one, I was going to remove it. Will do so for the next checkin (it was just there for debugging).

lib/models/addons-factory.js
dcombslinkedin: Oops - I forgot to remove that. Good catch! 

At one point I'd missed a step in the previous version of addon-discovery where it sets the 'ember-addon' field to {} if it wasn't present. This resulted in an NPE here. I added the check to see what was going on. Then I fixed PackageInfo to do the relevant setting during the ctor so the check isn't needed anymore. I've removed it and will check in the fix.

lib/models/project.js
dcombslinkedin: Yep, I've removed all console.XXX calls that I added (there are a couple in comments somebody else did earlier that I left alone).

lib/models/project.js
dcombslinkedin: Good catch - missed that one.

lib/models/package-info-cache.js
dcombslinkedin: Yep, that comes next.

lib/models/package-info-cache.js
dcombslinkedin: Yep.

lib/models/package-info-cache.js
dcombslinkedin: Okay, I'll go figure out how to do that. I was wondering what was needed.

lib/models/package-info-cache.js
dcombslinkedin: Nice idea! Doing that now.

lib/models/package-info-cache.js
dcombslinkedin: Added a comment to that effect.

lib/cli/cli.js
```diff
@@ -1,16 +1,17 @@
-'use strict';
+"use strict";
```
kellyselden: I would prefer to keep styling refactors out of this PR.

dcombslinkedin: I'm fine with that. For some weird reason I was getting errors in the job because of the formatting, so I ran 'prettier' and ended up with this. I'll see about backing that out and starting over.

PR #7709, opened 2018-03-29
rename jshintAddonTree to linterAddonTree

billybonks:
While investigating something else i noticed that this method name didn't really describe what the function was doing any more :)


Turbo87:
looks like CI is failing 🤔

kellyselden:
@rwjblue Correct me if I'm wrong, but even though this is labelled as private, this is probably going to break people using these functions. We should probably have a deprecation period.

rwjblue:
>   Correct me if I'm wrong, but even though this is labelled as private, this is probably going to break people using these functions. We should probably have a deprecation period.

Correct, but it would have to be someone directly calling `jshintAddonTree` which seems pretty unusual (there are only two usages on [emberobserver.com code search](https://emberobserver.com/code-search?codeQuery=jshintAddonTree)).  Most folks would rely on the linting behavior implicitly, so they wouldn't notice the rename. I'm still 👍 on a deprecation though...

lib/models/addon.js
```diff
@@ -1179,13 +1179,13 @@ let addonProto = {
...
+    Invokes lintTree for addon that implements lintTree function
 
     @private
-    @method jshintAddonTree
-    @return {Tree} Tree with JShint output (tests)
+    @method linterAddonTree
+    @return {Tree} Tree with tests generated by included linters
   */
-  jshintAddonTree() {
+  linterAddonTree() {
```
rwjblue: Lets call it `lintAddonTree` instead

billybonks:
deprecated original function / renamed rename 😄 looking into cli, let me know about the deprecation copy.

billybonks:
it seems like master is failing in general on same spec https://travis-ci.org/ember-cli/ember-cli/jobs/360034648#L1969

lib/models/addon.js
```diff
@@ -1239,7 +1243,7 @@ let addonProto = {
     return new Funnel(lintedAddon, {
       srcDir: '/',
       destDir: `${this.name}/tests/`,
-      annotation: `Funnel: Addon#jshintAddonTree(${this.name})`,
+      annotation: `Funnel: Addon#lintAddonTree(${this.name})`,
```
billybonks: I am not sure if stats are depending on annotations, so the question is that do we need a long lead time to change the annotation.

rwjblue:
Sorry for the delay here, but I think its probably best to close this PR. I think that we should _probably_ consider deprecating the tree based linting hooks (we've migrated default apps off of them as of 3.17.0-beta.1).

gabrielcsapo:
is there an open issue to remove jshintAddonTree?

rwjblue:
Dunno, I don't see an obvious one...

PR #7712, opened 2018-03-30
compact code a bit so it's less effort to read

billybonks:
Makes reading code easier, since it's more compact. uses existing trees, instead of creating new ones and shares code.

would like to know if maintainers are happy with this idea :) (tests are currently failing but will get them fixed)

I started here 1f1ea37a075699748fe657ff5805870bca476b5d then thought to myself maybe i can combine addons into this fbbca19ea74f9a7d27f2b1ab5a78fdb3596daf26, but that lead me to find out that some deeper refactoring would be required to make that happen. so i compromised with 15ad7c3fdffdf0eb39bc93d875253f1e8d22d929, but i feel like this should be easily possible ffbf04bd4dc3d75621235c4d10705ef3df3f8369

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1228,29 +1228,14 @@ let addonProto = {
       trees = trees.concat(lintAddonJsTrees, lintTemplateTrees);
     }
 
-    let addonTestSupportPath = this._treePathFor('addon-test-support');
-    if (fs.existsSync(addonTestSupportPath)) {
-      let addonTestSupportTree = new Funnel(addonTestSupportPath, { destDir: 'addon-test-support' });
-      let lintAddonTestSupportJsTrees = this._eachProjectAddonInvoke('lintTree', [
-        'addon-test-support',
-        addonTestSupportTree,
-      ]);
-      trees = trees.concat(lintAddonTestSupportJsTrees);
-    }
-
-    let appPath = this._treePathFor('app');
-    if (fs.existsSync(appPath)) {
-      let appTree = new Funnel(appPath, { destDir: 'app' });
-      let lintAppJsTrees = this._eachProjectAddonInvoke('lintTree', ['app', appTree]);
-      trees = trees.concat(lintAppJsTrees);
-    }
-
-    let testSupportPath = this._treePathFor('test-support');
-    if (fs.existsSync(testSupportPath)) {
-      let testSupportTree = new Funnel(testSupportPath, { destDir: 'test-support' });
-      let lintTestSupportJsTrees = this._eachProjectAddonInvoke('lintTree', ['test-support', testSupportTree]);
-      trees = trees.concat(lintTestSupportJsTrees);
-    }
+    ['test-support', 'app', 'addon-test-support'].forEach(treeName => {
+      let treePath = this._treePathFor(treeName);
+      if (fs.existsSync(treePath)) {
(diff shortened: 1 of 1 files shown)


lib/models/addon.js
```diff
@@ -1206,31 +1205,21 @@ let addonProto = {
...
 
-    let addonTestSupportPath = this._treePathFor('addon-test-support');
-    if (fs.existsSync(addonTestSupportPath)) {
-      let addonTestSupportTree = new Funnel(addonTestSupportPath, { destDir: 'addon-test-support' });
-      let lintAddonTestSupportJsTrees = this._eachProjectAddonInvoke('lintTree', ['addon-test-support', addonTestSupportTree]);
-      trees = trees.concat(lintAddonTestSupportJsTrees);
     }
 
-    let appPath = this._treePathFor('app');
-    if (fs.existsSync(appPath)) {
```
billybonks: this will cause error

lib/models/addon.js
```diff
@@ -1243,6 +1232,26 @@ let addonProto = {
...
+    @private
+    @param  {String} type will be fowarded to addon as type paramter, and is used to find path
+    @param  {Tree} treeName optional if this paramter is passed it will not lookup tree using treeName, and pass this tree instead
+    @method _lintTree
+    @return {Tree} Tree with linter tests
+  */
+  _lintTree(type, treeName, destDir) {
+    let tree = null;
+    treeName = treeName || type;
+    destDir = destDir || treeName;
```
Turbo87: we don't seem to use the `treeName` or `destDir` parameters anywhere when calling the method. IMHO we shouldn't add them unless we actually need them.

lib/models/addon.js
```diff
@@ -1243,6 +1232,26 @@ let addonProto = {
...
+    let treePath = this._treePathFor(treeName);
+    if (fs.existsSync(treePath)) {
+      tree = new Funnel(treePath, { destDir });
+    } else {
+      return;
+    }
+    return this._eachProjectAddonInvoke('lintTree', [treeName, tree]);
+  },
```
Turbo87: could likely be simplified to:

```js
  _lintTree(type) {
    let treePath = this._treePathFor(type);
    if (!fs.existsSync(treePath)) {
      return;
    }

    let tree = new Funnel(treePath, { type });
    return this._eachProjectAddonInvoke('lintTree', [type, tree]);
  }
```

lib/models/addon.js
```diff
@@ -1206,31 +1205,21 @@ let addonProto = {
...
+      this._lintTree('app'),
+      this._lintTree('addon-test-support')
+    ).filter(Boolean);
+
 
-    let testSupportPath = this._treePathFor('test-support');
-    if (fs.existsSync(testSupportPath)) {
-      let testSupportTree = new Funnel(testSupportPath, { destDir: 'test-support' });
-      let lintTestSupportJsTrees = this._eachProjectAddonInvoke('lintTree', ['test-support', testSupportTree]);
-      trees = trees.concat(lintTestSupportJsTrees);
```
Turbo87: note that this was `concat()`, not `push()`. this might have to be flattened before we can `concat()` the `_lintTree()` results. 🤔

locks: How does this look?

locks:
I was going to close/open to kick CI, oops 😬

PR #7713, opened 2018-03-30
Make styles first class lint citizens

billybonks:
Currently styles are not first class lint citizens, and this causes a bunch of problems that i will try to describe.

type=='app' only contains javascript
image

so i have been doing the following

```
if(type == 'app'){
  return StyleLinter.create(this.app.trees.app, this.styleLintOptions)
}
```

for addons that value maps to `tests/dummy` directory
for apps that value maps to `app` directory

but some addon's want to lint `app` directory as well.

so stylelinters use a config called `included paths`

```
    let toBeLinted = [ this.app.trees.app ];
    if (this.styleLintOptions.includePaths) {
      toBeLinted.push.apply(toBeLinted, this.styleLintOptions.includePaths);
    }
    
    let linted = toBeLinted.map(function(tree) {
      let filteredTreeToBeLinted = new Funnel(tree, {
        exclude: ['**/*.js']
      });
      return StyleLinter.create(tree, this.styleLintOptions)
```

but the issue is if you have multiple addons in your app or you are developing an addon type=='app' gets called multiple times.

image

In order to rectify this we need each lint method the auxillary one and the main one to use `type=='styles'`

that way each time type==`styles` we can be sure that it is a unique branch of the code so unique addon/engine/ or the main app (it seems like the main `app` directory is considered an addon for addon repositories. and stylelinters can be stable 😍 

examples of work arounds
- https://github.com/billybonks/ember-cli-stylelint/blob/master/index.js#L64
- https://github.com/tomasbasham/ember-cli-scss-lint/blob/master/index.js#L54

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1475,7 +1475,7 @@ class EmberApp {
   }
 
   /**
-    Runs the `app`, `tests` and `templates` trees through the chain of addons that produces lint trees.
+    Runs the `app`, `tests` and `templates`, `styles` trees through the chain of addons that produces lint trees.
 
     Those lint trees are afterwards funneled into the `tests` folder, babel-ified and returned as an array.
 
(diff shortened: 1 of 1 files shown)


billybonks:
waiting on the result of these pr's
https://github.com/ember-cli/ember-cli/pull/7709

locks:
Closing due to inactivity.

PR #7717, opened 2018-03-31
(fix ci) assert no filters matched

billybonks:
From what i understand, the config of this test would cause jshint tests to not be generated, and so if we filter by jshints we expect to see no failing tests.

recently qunit started throwing errors if nothing matched the filter 
https://github.com/qunitjs/qunit/blame/fc1c3ce4f9a403e3cfbd987a29aefb076560bf2b/src/core/processing-queue.js#L157

I catch the error assert it happened ( the assertions are a bit fragile). if that assertion did not run the qunit matched tests filtered as jshint


kellyselden:
Thanks!

PR #7721, opened 2018-04-02
update eslint-plugin-node for addons

kellyselden:
nothing too extreme in the breaking changes/rules


kellyselden:
Appveyor has a bad cache issue on one job, and I think merging is easier than trying to purge that bad cache (tied to this PR).

PR #7724, opened 2018-04-02
github package was renamed @octokit/rest

kellyselden:
https://github.com/octokit/rest.js/releases/tag/v14.0.0


kellyselden:
Works for me (on node 8 mac)

PR #7728, opened 2018-04-02
Passing defaultOptions to testem to prevent the cwd and config_dir set in testem.js from being overridden by ember-cli

arthirm:

testem/testem#1219


Turbo87:
It appears that this patch has broken `ember-exam` ... 😭

PR #7729, opened 2018-04-03
package-info-cache (fast package caching)

dcombslinkedin:
This is now at the point where all previous ember-cli tests pass (I will shortly add tests for package-info-cache). I would like @stefanpenner and @rwjblue to take a look and suggest improvements before the final PR with new tests goes in. Here's the feature description:

Doing profiling of build time processing for a large project at LinkedIn found that the addon discovery process was taking over 50 seconds just to determine all where all the addon packages were on disk and read and parse their package.json files. I found that much of this was due to repeated validation of the path to the addon package.json and index.js files (fs.realpathSync turns out to be fairly expensive when done a lot, and the project was calling it something like 80K times). The implementation was already re-reading the package information and then throwing it away many times.

This PR creates a new PackageInfoCache class, with a singleton instance added to either CLI (if the build is being done through the CLI or an individual project (if the build is done on a single project but not through the CLI). If a build, like ours here at LinkedIn, includes multiple projects, the fact the instance is in CLI means they all take advantage of the same cache (I'll call it PIC from here out) and don't need to rediscover any packages.

When a Project (ember-cli/lib/models/project) is constructed, a reference to a local or CLI version of PIC is made and PIC.loadProject(projectPath) is called. loadProject then creates an internal instance of a PackageInfo class for the project. It parses the package.json file, then creates separate ProjectInfo instances for all internal addons, all CLI in-repo addons and all 'dependency' and 'devDependency' entries. In the process of that it does the same thing that the node resolution algorithm does to find package.json files, but tries as much as possible to not use 'fs.realpathSync' more than once per file path, caching the result of each path resolution for speed.

Because of the cache, reading all the dependencies, parsing their package.json files, determining the addons and collecting information about them (prior to addon-instantiation), current performance on my Mac Pro machine (new 2017 model) has gone from 54 seconds down to 900 milliseconds. Obviously the precise performance will differ based on machine size, but overall it's about a 60x increase.

The result of all the above processing is that the PackageInfoCache contains a number of entries, one per package for each real file path (i.e. its cache entries are keyed on the real path for a package's root directory--all symlinks and .. entries are resolved). A PackageInfo object contains the following fields of interest:

pkg - the object resulting from the package.json file
realPath - (string) the resolved path for the package's root directory
cache - a reference to the PIC object that contains the packageInfo instance
errors - an ErrorList object containing ErrorEntry objects, one per error discovered during the processing of the package and all its various dependencies. When a packageInfo object is being created, if the package itself is bad (e.g. no or malformed package.json file) the error is created and put in the parent. If the package is generally constructed correctly, anything wrong with the contents of the object itself (e.g. one of its dependencies cannot be found), the error is registered with the object. The PIC has a method 'showErrors' to dump a list of all the errors in all the PackageInfo objects. The caller could then decide if any of those errors is enough to fail the build.
Various types of PackageInfo objects can have other fields:

addonConstructor - filled out but not yet used, this would replace part of addons-factory so that constructing an instance of an addon would be done through the PIC/PackageInfo, simplifying the code
addonMainPath - for an addon package, the real path to the package's 'main' file
inRepoAddons - an array of PackageInfo objects, one per in-repo package definition (note that the in-repo addon definition is included in the PIC-level cache by its file path already, this is just a reference to that same set of objects)
internalAddons - same as above, but for internal addons (project only)
cliInfo - same as above, but for the package that corresponds to the path .cli.root
dependencyPackages - a JS object keyed by dependency name of references to PackageInfo objects, one per entry in the package's 'dependency' attribute list)
devDependencyPackages - similar, but for the 'devDependency' attribute list
nodeModules - a NodeModulesList object (basically contains an array of PackageInfo and other NodeModulesList objects) of all packages and subdirectories in a node_modules directory.


dcombslinkedin:
Yes, I think those are all that's left, now that eslint is passing. Working on them today.

dcombslinkedin:
The latest changes fix issues with running the ember-cli tests (specifically things like supporting 'project.reloadPkg'). At this point I believe no more feature work is needed, and will be adding tests shortly.

lib/models/project.js
```diff
@@ -75,14 +76,38 @@ class Project {
     instrumentation.project = this;
 
     this.emberCLIVersion = emberCLIVersion;
+
     this._nodeModulesPath = null;
+
+    if (this.cli && this.cli._packageInfoCache) {
```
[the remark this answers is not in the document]
dcombslinkedin: * If somebody directly creates a project, they could pass in null for cli. I also saw it in a piece of code checking for isEmberCliProject. It is admittedly not consistent across the file.
* Originally MockCLI didn't have a _packageInfoCache entry. I changed the code the other day to add one. It was tripping over that issue in tests a long time ago, so the second test really isn't needed any more. First one, not so sure.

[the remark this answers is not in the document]
dcombslinkedin: Removed the second part of the test.

lib/models/project.js
```diff
@@ -75,14 +76,38 @@ class Project {
...
+    } else {
+      this._packageInfoCache = new PackageInfoCache(this.ui);
+    }
+
+    // we're not dealing with NULL_PROJECT (note that it has not
+    // be set yet, so we can't compare to that var.)
+    this._packageInfo = this._packageInfoCache.loadProject(this);
+
+    // force us to use the real path as the root.
+    if (this.root !== this._packageInfo.realPath) {
```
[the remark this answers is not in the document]
dcombslinkedin: Yes, on the 'would it make more sense' part. At one point I was finding some weird instability, so I added an if and a log message. I'd rather that we assign root from the packageInfo instance, since that's already gone through the fs.realPathSync() tests itself, for speed. Are you okay with just .     this.root = this._packageInfo.realPath .   instead?  That's what I had before.

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+/**
+ * Resolve the real path for a directory, return null if does not
+ * exist or is not a directory, return the real path otherwise.
+ *
+ * @private
+ * @method getRealDirectoryPath
+ * @param  {String} directoryPath the path to the directory of interest
+ * @return {String} real path or null
+ */
+function getRealDirectoryPath(directoryPath) {
```
stefanpenner: needs unit test

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+/**
+ * Resolve the real path for a file, return null if does not
+ * exist or is not a file or FIFO, return the real path otherwise.
+ *
+ * @private
+ * @method getRealFilePath
+ * @param  {String} filePath the path to the file of interest
+ * @return {String} real path or null
+ */
+function getRealFilePath(filePath) {
```
stefanpenner: needs unit test

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+
+/*
+ * Small utility class to store a list of errors during processing.
+ * Instances of this exist in the PackageInfo and PackageInfoCache
+ * objects.
+ *
+ * @private
+ * @class ErrorEntry
+ */
+class ErrorList {
```
stefanpenner: needs unit test

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+  }
+
+  isAddon() {
+    let val =
+      Array.isArray(this.pkg.keywords) &&
+      this.pkg.keywords.indexOf('ember-addon') >= 0;
+    return val;
+  }
+
+  // XXX Should this move to Addon.js (and change appropriately to use the cache here?)
```
stefanpenner: should this comment be addressed before merging?

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+   * Return a PackageInfo object for a given package name (which may include
+   * a scope)
+   *
+   * @private
+   * @method _findPackage
+   * @param {String} packageName the name (possibly including a scope) of
+   *    the PackageInfo the caller wants returned.
+   * @return the desired PackageInfo if one exists for the given name, else null.
+   */
+  _findPackage(packageName) {
```
stefanpenner: unit test

tests/unit/models/package-info-cache-test.js
```diff
@@ -0,0 +1,219 @@
...
+    cli = new MockCLI({ ui });
+  });
+
+  describe('packageInfo contents tests on valid project', function() {
+    beforeEach(function() {
+      projectPath = path.resolve(addonFixturePath, 'simple');
+      packageJsonPath = path.join(projectPath, 'package.json');
+      packageContents = require(packageJsonPath);
+      project = new Project(projectPath, packageContents, ui, cli);
+      pic = project._packageInfoCache;
```
stefanpenner: given that this is used in tests, should we make `project.packageInfoCache` instead? And annotate its API doc to be `@protected`? It feels a tad funky to reach into private stuff, to test.

package.json
```diff
@@ -39,6 +39,7 @@
   "dependencies": {
     "amd-name-resolver": "^1.2.0",
     "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
+    "bower": "^1.8.4",
```
stefanpenner: By default, we don't include bower. It is something that is on-demand installed when required. The reason was, `bower` is quite a large dependency.

tests/unit/models/package-info-cache-test.js
```diff
@@ -0,0 +1,219 @@
...
+const path = require('path');
+const expect = require('chai').expect;
+const Project = require('../../../lib/models/project');
+let addonFixturePath = path.resolve(__dirname, '../../fixtures/addon');
+const MockUI = require('console-ui/mock');
+const MockCLI = require('../../helpers/mock-cli');
+
+describe('models/package-info-cache.js', function() {
+  let project, projectPath, packageJsonPath, packageContents, projectPackageInfo, ui, cli, pic;
+  this.timeout(100000);
```
stefanpenner: this timeout seems rather high, the tests bellow all seem super fast. I suspect this was only added to aide local debugging, but I believe it can be removed

lib/models/addon.js
```diff
@@ -215,12 +221,26 @@ let addonProto = {
...
+      // for it so we get the PackageInfo for it and its children.
+      this._packageInfoCache = new PackageInfoCache(this.ui);
+    }
 
-    if (!this.name) {
-      throw new SilentError(`An addon must define a \`name\` property (found at ${this.root}).`);
+    this._packageInfo = this._packageInfoCache.loadAddon(this);
+
+    // force us to use the real path as the root.
+    if (this.root !== this._packageInfo.realPath) {
```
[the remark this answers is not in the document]
dcombslinkedin: I removed the test, but would like to keep the determining of real paths in PIC, instead of here. It's already set up there to do it (using getRealDirectoryPath), and does set the value back here in Addon. Same as project.

lib/models/package-info-cache.js
dcombslinkedin: Probably. At this point I'm going to leave it, so will consider just removing the function entirely, since things are working and addon-factory already does what this is doing (since I more or less copied from there). We could revisit later on a new effort.

lib/models/addon.js
```diff
@@ -167,36 +168,41 @@ function warn(message) {
...
-    this.addonDiscovery = new AddonDiscovery(this.ui);
     this.addonsFactory = new AddonsFactory(this, this.project);
     this.registry = p.defaultRegistry(this);
     this._didRequiredBuildPackages = false;
 
     if (!this.root) {
       throw new Error('Addon classes must be instantiated with the `root` property');
     }
+
+    if (!this.name) {
```
twokul: seems like a new behaviour? should we unit test this as well? also, what's the reasoning behind throwing an error?

lib/models/package-info-cache.js
dcombslinkedin: Several of the tests for PIC itself contain references to error lists that will fail if they aren't right. I guess I could add more specifically for ErrorList itself, but not sure if that's just overkill.

lib/models/package-info-cache.js
dcombslinkedin: How does one create a unit test for a private function, without creating an API that shouldn't be public?

lib/models/package-info-cache.js
dcombslinkedin: Removed the function, and the reference to 'this.addonConstructor' in the members of PackageInfo.

lib/models/addon.js
```diff
@@ -371,16 +391,76 @@ let addonProto = {
...
+    // There are cases where an addon can be created in memory and not actually
+    // have a root entry (or have one that is not actually pointing to a directory,
+    // like 'foo' in some of the tests. We don't want to crash, but want to let
+    // things continue even though they're empty.
+    if (pkgInfo) {
+
+      // The 'dependencyPackages' object contains ALL dependencies, not just those
+      // that are addons. We want only the addons.
+      if (pkgInfo.dependencyPackages) {
+        Object.keys(pkgInfo.dependencyPackages).forEach(name => {
```
twokul: can we use `filter` here?

lib/models/addon.js
```diff
@@ -371,16 +391,76 @@ let addonProto = {
...
+            if (dependencyInfo.isAddon()) {
+              addonPackageList.push(dependencyInfo);
+            }
+          }
+        });
+      }
 
-    this.addonPackages = this.addonDiscovery.addonPackages(addonsList);
+      if (pkgInfo.inRepoAddons) {
+        pkgInfo.inRepoAddons.forEach(addon => addonPackageList.push(addon));
```
twokul: maybe we could just `addonPackageList.concat(pkgInfo.inRepoAddons)`?

tests/unit/models/package-info-cache-test.js
dcombslinkedin: The original timeout was 500000! I'll reset down to 20000, which is still about 4x what's really necessary.

lib/models/addon.js
dcombslinkedin: Actually that was already in the class when I started working on it. It might have been farther down in the constructor.

lib/models/addon.js
```diff
@@ -371,16 +391,76 @@ let addonProto = {
...
+          this.addonPackages[packageInfo.name] = addonInfo;
+        }
+      } else {
+        invalidAddonPackages.push(packageInfo);
+      }
+    });
+
+    if (invalidAddonPackages.length > 0) {
+      this.ui.writeWarnLine(`For addon at path ${this._packageInfo.realPath}:`);
+      invalidAddonPackages.forEach(packageInfo => {
```
twokul: is additional iteration necessary here? can we move `this.ui.writeWarnLine` to the `else` branch of the `if` above?

lib/models/addon.js
dcombslinkedin: Not really, because the list contains names, and the result is additions of their corresponding values to the list--we're not just filtering the names. If we could have used Object.values() it would have been okay, but that's not in ES5.

lib/models/project.js
```diff
@@ -408,16 +438,83 @@ class Project {
...
+          packageInfo.pkg
+        );
+      } else {
+        invalidAddonPackages.push(packageInfo);
+      }
+    });
+
+    if (invalidAddonPackages.length > 0) {
+      this.ui.writeWarnLine(`For project at path ${this._packageInfo.realPath}:`);
+      invalidAddonPackages.forEach(packageInfo => {
```
twokul: same comment as above

lib/models/project.js
```diff
@@ -408,16 +438,83 @@ class Project {
...
+
+    if (this.isEmberCLIAddon()) {
+      addonPackageList.push(pkgInfo);
+    }
+
+    let cliInRepoAddons = pkgInfo.cliEntry
+      ? pkgInfo.cliEntry.inRepoAddons
+      : null;
+    if (cliInRepoAddons) {
+      cliInRepoAddons.forEach(addon => addonPackageList.push(addon));
```
twokul: could probably use `concat`

lib/models/project.js
```diff
@@ -408,16 +438,83 @@ class Project {
...
+    }
+
+    if (pkgInfo.internalAddons) {
+      pkgInfo.internalAddons.forEach(addon => addonPackageList.push(addon));
+    }
+
+    // The 'dependencyPackages' object contains ALL dependencies, not just those
+    // that are addons. We want only the addons
+    if (pkgInfo.dependencyPackages) {
+      Object.keys(pkgInfo.dependencyPackages).forEach(name => {
```
twokul: could probably use `filter`

lib/models/addon.js
dcombslinkedin: Definitely! Good catch. Done.

lib/models/project.js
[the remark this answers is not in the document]
twokul: > If somebody directly creates a project, they could pass in null for cli.

should this be allowed? is it because of our mocks in tests?

lib/models/addon.js
dcombslinkedin: I was doing it separately so that we get a single list like the following:

For addon at path XXX:
    Excluding invalid/malformed/missing addon at relative path YYY
    Excluding invalid/malformed/missing addon at relative path ZZZ

I think what you're suggesting results in either no header line about the XXX addon, or duplications of that message. Or am I mistaken?

dcombslinkedin: In the above, the 'excluding' lines are supposed to be indented 2 spaces.

twokul: I see what you mean, makes sense!

lib/models/project.js
```diff
@@ -315,7 +342,10 @@ class Project {
     @return {Boolean}      Whether or not the file is present
    */
   has(file) {
-    return fs.existsSync(path.join(this.root, file)) || fs.existsSync(path.join(this.root, `${file}.js`));
+    return (
```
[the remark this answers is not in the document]
dcombslinkedin: Thanks! I'd used 'prettier' at one point and forgot to revert this one. Done.

tests/unit/models/package-info-cache-test.js
```diff
@@ -0,0 +1,219 @@
...
+    ui = new MockUI();
+    cli = new MockCLI({ ui });
+  });
+
+  describe('packageInfo contents tests on valid project', function() {
+    beforeEach(function() {
+      projectPath = path.resolve(addonFixturePath, 'simple');
+      packageJsonPath = path.join(projectPath, 'package.json');
+      packageContents = require(packageJsonPath);
+      project = new Project(projectPath, packageContents, ui, cli);
```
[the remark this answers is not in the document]
dcombslinkedin: The PackageInfoCache is generally used by creating it inside a Project or Addon, and then calling 'loadProject' or 'loadAddon' on that object. That's how it starts the discovery process. The PackageInfo and other internal objects there are really just supposed to be created in the PIC itself. Make sense?

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
+'use strict';
```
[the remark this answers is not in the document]
dcombslinkedin: I guess we could, though it would then sort of pollute the directory with what really are internal items in the cache (the PackageInfo, the NodeModuleList, the ErrorList and ErrorEntry objects). The cache is sort of irrelevant without them. I guess we could make a subdirectory of 'lib/models', but didn't know what you guys preferred.

lib/models/package-info-cache.js
dcombslinkedin: Asked in various places. ;-) - How do you test a private function?

lib/models/package-info-cache.js
```diff
@@ -0,0 +1,1049 @@
...
+    // other packages that have errors. We will try to process
+    // things anyway.
+    if (!val.processed) {
+      this.projects.push(projectObj);
+
+      // projects are a bit different than standard addons, in that they have
+      // possibly a CLI addon and internal addons. Add those now.
+      let pkgInfo = val;
+
+      pkgInfo.project = projectObj;
```
twokul: why are we assigning `project` property only in `loadProject`?

lib/models/project.js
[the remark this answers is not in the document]
dcombslinkedin: Probably not, but I found that without the second part of the test (I have re-added it), blueprint tests fail. Apparently they are somehow creating a project without a CLI, or an empty one. I fixed MockCLI a while back so it does create a _packageInfoCache, but it may not always be being used.

tests/unit/models/package-info-cache-test.js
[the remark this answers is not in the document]
twokul: yeah! since it's `package-info` unit test, my assumption was that we would bypass `Project` or `Addon` altogether and will be testing `PackageInfoCache` directly. Looking at the current implementation, we pass `this` to both `loadProject` and `loadAddon` so my first thought was to avoid passing `this` and pass _only_ the arguments we need to minimize api surface and things the method relies on

tests/unit/models/package-info-cache-test.js
dcombslinkedin: That sounds like a good idea. Same for addon.

package.json
dcombslinkedin: Not sure how that ended up there - it probably got installed and I just assumed it was okay. Will remove it.

lib/models/addon.js
twokul: wonder if we could use [`entries`](https://node.green/#ES2015-built-in-extensions-Array-prototype-methods-Array-prototype-entries)

lib/models/package-info-cache.js
dcombslinkedin: Because the PackageInfo itself is general - it is used for any package.json. The only place where we actually know we have a project is in loadProject(). That's also why project-specific stuff is done in loadProject (like cli.root and supportedInternalAddonPaths) rather than _readPackage. Make sense?

lib/models/addon.js
dcombslinkedin: Actually, it doesn't work. pgkInfo.dependencyPackages is a plain JS object (i.e., a map), so Array.entries won't work directly. If you use Object.keys() on it first, then you have an array of names, which means you still need to get the value from the dependencyPackages property. We really need Object.values(), which does exactly what we want.

tests/unit/models/package-info-cache-test.js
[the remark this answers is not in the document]
dcombslinkedin: Well, we could create a new PackageInfoCache(), then call loadProject on it to load a project, since that's the public API, but creating the project itself already creates a PackageInfoCache as part of its constructor, so it's sort of pointless. We don't really want to call _readPackageInfo directly, since it's internal.

lib/models/package-info-cache.js
dcombslinkedin: A question: are we supposed to be able to unit test internal functions like this one directly, or via side-effect, or what? _findPackage (in NodeModulesList) is only called as part of _addDependencies, and that would

lib/models/package-info-cache.js
dcombslinkedin: Exposed the function as part of module.exports, and added tests for getRealDirectoryPath and getRealFilePath.

lib/models/package-info-cache.js
dcombslinkedin: Exposed the function as part of module.exports, and added tests for getRealDirectoryPath and getRealFilePath.

lib/models/package-info-cache.js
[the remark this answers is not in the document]
dcombslinkedin: @twokul, any thoughts on my comment? I've tried to add more documentation comments (header for each class, for example), but I agree that it's not a small file.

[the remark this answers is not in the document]
stefanpenner: My recommendation would be a `package-info` folder, containing:

* `index.js` the package-info itself
* each atomic unit as a sibling file

lib/models/addon.js
stefanpenner: the following may work, although I'm not sure its worth changing.

```
Object.entries(pkgInfo.dependencyPackages).forEach(([ name, dependencyInfo ]) => console.log('name', name, 'depInfo', dependencyInfo))
```

lib/models/project.js
stefanpenner: Ya, until our lowest node LTS support version is 8. I think we are stuck without.

I don't think we need to add a polyfil, what is written seems fine.

lib/models/package-info-cache.js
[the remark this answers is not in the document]
dcombslinkedin: Done and checked in. The directory is called package-info-cache, and each class is its own file.

lib/models/project.js
```diff
@@ -408,16 +435,83 @@ class Project {
...
+      : null;
+    if (cliInRepoAddons) {
+      addonPackageList = addonPackageList.concat(cliInRepoAddons);
+    }
+
+    if (pkgInfo.internalAddons) {
+      addonPackageList = addonPackageList.concat(pkgInfo.internalAddons);
+    }
+
+    // The 'dependencyPackages' object contains ALL dependencies, not just those
```
mikrostew: The next 40-ish lines are very similar to the same code in discoverAddons() in lib/models/addon - this should be refactored to remove the duplication

kellyselden:
> remove typo-fixup stuff, will commit in later PR

I think you should submit that first, so if we were to revert this, it wouldn't be conflicted with the typo stuff on top of it.

dcombslinkedin:
Okay, then I'll need to revert the revert (will do after I submit a separate PR - we're only talking about a few lines in about 5 files.) Coming up.

dcombslinkedin:
Created the PR https://github.com/ember-cli/ember-cli/pull/7794 for the typo fixes.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+    let emberAddonInfo = pkg['ember-addon'];
+
+    // Set up packageInfos for any in-repo addons
+    if (emberAddonInfo) {
+      let paths = emberAddonInfo.paths;
+
+      if (paths) {
+        paths.forEach(p => {
+          let addonPath = path.join(realPath, p); // real path, though may not exist.
+          let val = this._readPackage(addonPath); // may have errors in the addon package.
```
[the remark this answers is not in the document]
dcombslinkedin: Can't do what you want exactly - pkgInfo is already defined. I'll do something, though.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+   * Once all the addon processing is done, go back through all cache entries
+   * to create references between the packageInfo objects.
+   *
+   * @public
+   * @method loadAddon
+   * @return {PackageInfo} the PackageInfo object for the given Addon object.
+   * Note that if the addon path is already in the cache, that will be returned.
+   * No copy is made.
+   */
+  loadAddon(addonObj) {
```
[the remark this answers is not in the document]
dcombslinkedin: An actual Addon instance (as opposed to an addonInfo, which is the {path, name, pkg}, or an addon package info. I'll change to addonInstance

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+    let val = this._readPackage(addonObj.root, addonObj.pkg);
```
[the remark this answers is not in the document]
dcombslinkedin: Yes. Changed to pkgInfo, and added header @param comment. Same for loadProject.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+   */
+  loadProject(projectObj) {
+
+    let val = this._readPackage(projectObj.root, projectObj.pkg);
+
+    // NOTE: the returned val may contain errors, or may contain
+    // other packages that have errors. We will try to process
+    // things anyway.
+    if (!val.processed) {
+      this.projects.push(projectObj);
```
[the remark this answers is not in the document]
dcombslinkedin: In the voyager-web case we have multiple projects right now. We really should probably change them to engines, but that work hasn't been done yet, and I don't know if/when it is scheduled. In the meantime, I need this to support reloadProjects(), which is used to support project.reloadPkg().

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+   *
+   * @public
+   * @method loadProject
+   * @return {PackageInfo} the PackageInfo object for the given Project object.
+   * Note that if the project path is already in the cache, that will be returned.
+   * No copy is made.
+   */
+  loadProject(projectObj) {
+
+    let val = this._readPackage(projectObj.root, projectObj.pkg);
```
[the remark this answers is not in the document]
dcombslinkedin: Yep - did as you suggested.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+        case Errors.ERROR_NODEMODULES_ENTRY_MISSING:
+          this.ui.writeWarnLine(`   Missing node_modules entry '[${errorEntry.data}'`);
+          break;
+      }
+    });
+  }
+
+  /**
+   * Process the root directory of a project, given a
+   * Project object (we need the object in order to find the internal addons).
```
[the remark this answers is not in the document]
dcombslinkedin: I changed the name to projectInstance, and explain in a header comment:
@param projectInstance the instance of the Project object to load package data about into the cache.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+/**
+ * Resolve the real path for a directory, return null if does not
+ * exist or is not a directory, return the real path otherwise.
+ *
+ * @private
+ * @method getRealDirectoryPath
+ * @param  {String} directoryPath the path to the directory of interest
+ * @return {String} real path or null
+ */
+function getRealDirectoryPath(directoryPath) {
```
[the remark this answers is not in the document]
dcombslinkedin: One is used to test that a given path exists and is a directory. The other is used to test that a given path exists, and is a file. Expecting a directory when the path exists and is a file is not a valid thing to do, so getRealDirectoryPath will return null in that case. Similarly for getRealFilePath. These are offshoots of similar methods that were in the 'resolve' package (if I remember the name right), the one that traveled up the FS tree looking for a file or directory. I just optimized it to minimize calling fs.xxx methods.

lib/models/package-info-cache/index.js
```diff
@@ -0,0 +1,751 @@
...
+ */
+
+const fs = require('fs-extra');
+const path = require('path');
+const ErrorList = require('./error-list');
+const Errors = require('./errors');
+const PackageInfo = require('./package-info');
+const NodeModulesList = require('./node-modules-list');
+
+let realFilePathCache;
```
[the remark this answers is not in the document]
dcombslinkedin: You should have one package-info-cache each time you run 'cli'. That was an idea Stef and I discussed at the beginning, that you want to have a new cache each time you create a new CLI, so it works with the existing tests and test mechanism. The file and dir package cache needs to be cleared out each time you change CLI, so I just put them in module scope. They definitely do not want to be in global scope.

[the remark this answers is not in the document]
dcombslinkedin: Regarding the base comment, the problem with addon-discovery is that is was moving a lot of non-common code into a common location (e.g., the project-specific code) that really should have been in project.js to adhere to OO encapsulation principles. The generality was also slowing things down. I moved such things into project or addon where reasonable. discoverAddons() was actually in project.js and addon.js to start with, so I'd left them there for this work. Also, the set of things gathered in Addon is fairly different than that in Project, except for the handling of dependencies (which I take care of in PackageInfoCache) so I was trying to keep them separate rather than having a single method with various parameters located outside each - that's just more work for no particularly good reason, IMHO. For example, the test for excluding 'ember-cli' in dependencies is Addon-specific, not in Project. The test for including CLI is project only, not addon. Given the differences, I'd really prefer to leave things as they are for this part.

dcombslinkedin:
Added changes based on @rwjblue comments. Squashed everything into one commit.

lib/models/project.js
```diff
@@ -409,16 +431,83 @@ class Project {
...
+   * package.json contents in this.addonPackages as key-value pairs.
+   *
+   * Any packageInfos that we find that are marked as not valid are excluded.
+   *
+   * @private
+   * @method discoverAddons
    */
   discoverAddons() {
-    let addonsList = this.addonDiscovery.discoverProjectAddons(this);
+    let pkgInfo = this.packageInfoCache.getEntry(this.root);
```
stefanpenner: `Project.prototype.discoverAddons` and `Addon.prototype.discoverAddons` are both complex functions and seem similar. Could we unify them as a single (or a handful) standalone functions? The duplication between comes at a high risk/cost currently.

This should mitigate the last technical concern @rwjblue has.

dcombslinkedin: Guys, I think you're basically asking me to unify two things that are not the same, and in a way that doesn't really make sense. Because they aren't the same, and because they take advantage of data that is specific to the particular objects, the idea of encapsulation dictates that they really should belong where they are. Here's a pseudo-code version of each:

```javascript
Project.discoverAddons() {
    get package entry
    create empty addonPackageList
    if this is an emberCLIAddon, add it to the package list
    if the project has a CLI, add those in-repo addons to the list
    if the project has internal addons, add those to the list
    if the package has 'dependencies' packages, add any that are addons to the list
    if the package has 'devDependencies' packages, add any that are addons to the list
    if the package itself has in-repo addons, add those to the list
    create an object with an AddonInfo for each valid addon package
    write out error messages for any invalid addon packages.
}
```
```javascript
Addon.discoverAddons() {
    get package entry
    create empty addon package list
    if we actually have package information (the addon is in the cache) {
        add any 'dependencies' packages that are NOT 'ember-cli' to the package list
     }
    add in-repo addons to the lsit

    create the same object of AddonInfo instances, except for each addon package,
        make sure that the calling addon allows it.

    write to the console for any addons that are considered invalid (
}
```

Given that project has many more places to check, and Addon has various checks for things like shouldIncludeChildAddon, the actual number of duplicate lines in the two routines is maybe 10 or 12. Having a combined routine externally that has to reach into the Project object or Addon object, or take several parameters to distinguish between them, is just bad code, in my opinion. It violates encapsulation and requires extra processing where none is needed, slowing things down (however slightly). 

I really feel strongly that these two should stay where they are for those reasons. The AddonDiscovery had the same sorts of problems, which is part of why it should have been refactored in the first place.

dcombslinkedin:
Added support for moving the messy details of discoverAddons into PackageInfo, per discussion with @stefanpenner and @rwjblue. Squashed the result.

lib/models/project.js
dcombslinkedin: Changes are done.

lib/models/addon.js
```diff
@@ -371,16 +387,36 @@ let addonProto = {
...
+   * and in-repo addons
+   *
+   * Any packageInfos that we find that are marked as not valid are excluded.
+   *
+   * @private
+   * @method discoverAddons
    */
   discoverAddons() {
-    let addonsList = this.addonDiscovery.discoverChildAddons(this);
+    let pkgInfo = this.packageInfoCache.getEntry(this.root);
```
stefanpenner: this refactoring looks great!

kiwiupover:
Thanks @dcombslinkedin for your work on this. You have saved many of us from waisting time on twitter. Cheers

lib/models/addon.js
```diff
@@ -371,16 +387,36 @@ let addonProto = {
...
+
+    if (pkgInfo) {
+      let self = this;
+      let addonPackageList = pkgInfo.discoverAddonAddons();
+      this.addonPackages =
+          pkgInfo.generateAddonPackages(addonPackageList,
+            addonInfo => (self.shouldIncludeChildAddon && !self.shouldIncludeChildAddon(addonInfo)));
```
stefanpenner, line 409: This appears to have introduced a regression, see originaly:

https://github.com/ember-cli/ember-cli/pull/7729/files#diff-6a115371fcdc66c67f9710569fa7c15eL75

stefanpenner: Actually, the double negation is occurring, this split over two different files. This may be fine.

Issue #7726, opened 2018-04-02
Module-unification component blueprint not working


xg-wang:
reproduce:
```
npm i -g github:ember-cli/ember-cli
MODULE_UNIFICATION=true ember new foo
cd foo
ember g component foo-bar
```
error would be:
> Cannot read property 'isModuleUnification' of undefined

This is due to the component blueprint in https://github.com/emberjs/ember.js/pull/16043 depends on `this.project` which doesn't exists [link](https://github.com/emberjs/ember.js/pull/16043/files?diff=split#diff-017cbe40fcdfe1804dd384796ec8f809R28).

The tests are correct, but they passed because the ember-cli was pinned to an [earlier version](https://github.com/emberjs/ember.js/blob/400ec6ee69265ce45538e07ace9523926569f805/package.json#L110) and some code wasn't covered.

cc @GavinJoyce 

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.1.0-beta.1
http_parser: 2.7.0
node: 8.9.4
v8: 6.1.534.50
uv: 1.15.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 57
nghttp2: 1.25.0
openssl: 1.0.2n
icu: 59.1
unicode: 9.0
cldr: 31.0.1
tz: 2017b
os: darwin x64
5.8.0
1.5.1
```

GavinJoyce:
Here's what I think is happening.

I made this change recently: https://github.com/ember-cli/ember-cli/commit/7b6be06caaf52e08e9d5d46d5ac79cda482fd1c7. I extracted out a variable as the logic was getting complicated, this had the side effect of calling [`blueprint.supportsAddon()`](https://github.com/ember-cli/ember-cli/blob/e23c2ef653cffa4a95db8dad78214cef20222b79/lib/models/blueprint.js#L775) in more cases that before. It turns out the `blueprint.supportsAddon()` calls `this.files()` which must evaluate something in the blueprint without the `this.project` context.

I'll work on a fix now

GavinJoyce:
@xg-wang I believe that this will fix it: https://github.com/ember-cli/ember-cli/pull/7731

xg-wang:
[...]

PR #7731, opened 2018-04-03
restore `addon-import` logic

GavinJoyce:
Fixes https://github.com/ember-cli/ember-cli/issues/7726

In my attempt to simplify a quite complex piece of code while adding an additional condition, I inadvertently modified the logic such that  `mainBlueprint.supportsAddon()` was called in more cases that before my change. It seems that calling [`blueprint.supportsAddon()`](https://github.com/ember-cli/ember-cli/blob/e23c2ef653cffa4a95db8dad78214cef20222b79/lib/models/blueprint.js#L775) has some side-effects (as it calls `blueprint.files()`) which resulted in some blueprint code executing without having a reference to `this.project`: https://github.com/ember-cli/ember-cli/issues/7726

Here is the original logic:

```js
if (noAddonBlueprint.indexOf(name) < 0 && !addonBlueprint && (mainBlueprint && mainBlueprint.supportsAddon()) && options.args[1]) {
  addonBlueprint = this.lookupBlueprint('addon-import', true);
}
```

This is all quite brittle, but as this code will be removed once MU takes over the world, I think it's ok to leave as is.

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -23,9 +23,12 @@ class GenerateTask extends Task {
     let addonBlueprint = this.lookupBlueprint(`${name}-addon`, true);
 
     // otherwise, use default addon-import
-    let mainBlueprintSupportsAddon = mainBlueprint && mainBlueprint.supportsAddon() && !this.project.isModuleUnification();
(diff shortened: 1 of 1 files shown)


GavinJoyce:
CI seems to be an unrelated timeout:

screen shot 2018-04-03 at 16 53 48
screen shot 2018-04-03 at 16 54 16

rwjblue:
Restarted that CI job

PR #7736, opened 2018-04-04
add addon-test-support/index.js to eslint glob bug mitigation

kellyselden:


kellyselden:
linking https://github.com/eslint/eslint/issues/9740 and https://github.com/eslint/eslint/pull/9748

kellyselden:
appveyor node 8 yarn cache is still corrupt, we might need them to kick it.

PR #7737, opened 2018-04-05
Fix issues identified with "delayed transpilation" system.

rwjblue:
This PR includes two specific fixes (discovered during pairing session with @twokul this morning).

---

Ensure `treeForSrc` does nothing when `isModuleUnification` is falsey.

An earlier change made `isModuleUnification` a manual opt-in (due to the breaking change of detecting `src/` and assuming it was module unification), however the `treeForSrc` implementation was not properly leveraging `this.isModuleUnification()` to determine if it should even do any work.

This adds an additional conditional to the short curcuit, and ensures that `treeForSrc` does _nothing_ when a given addon is not opting in to module unification.

---

Cleanup the mechanism to determine the default plugin for a given type.

Instead of using the previous heuristics we require that a plugin declare itself as the default for type. This would look like:

```js
registry.add('js', {
 name: 'ember-cli-babel',
 ext: 'js',
 isDefaultForType: true,
 toTree: (tree) => this.transpileTree(tree)
});
```

If no plugins declare that they are the default for a given type, all plugins will be used (since choosing the default arbitrarily is not a valid option).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -688,7 +688,7 @@ class EmberApp {
 
     if (defaultsForType.length > 1) {
       throw new Error(
-        `There are multiple preprocessor plugins marked as default that are registered for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
+        `There are multiple preprocessor plugins marked as default for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
       );
     }
 
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -688,7 +688,7 @@ class EmberApp {
 
     if (defaultsForType.length > 1) {
       throw new Error(
-        `There are multiple preprocessor plugins that are registered for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
+        `There are multiple preprocessor plugins marked as default that are registered for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
       );
     }
 
@@ -696,7 +696,7 @@ class EmberApp {
   }
 
   _compileAddonTemplates(tree) {
-    let defaultPluginForType = this._getDefaultPluginForType('js');
+    let defaultPluginForType = this._getDefaultPluginForType('template');
     let options = {
       annotation: `_compileAddonTemplates`,
       registry: this.registry,
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -682,20 +682,47 @@ class EmberApp {
     return this._addonTreesFor(type).map(addonBundle => addonBundle.tree);
   }
 
+  _getDefaultPluginForType(type) {
+    let plugins = this.registry.load(type);
+    let defaultsForType = plugins.filter(plugin => plugin.isDefaultForType);
+
+    if (defaultsForType.length > 1) {
+      throw new Error(
+        `There are multiple preprocessor plugins that are registered for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
+      );
+    }
+
+    return defaultsForType[0];
+  }
+
   _compileAddonTemplates(tree) {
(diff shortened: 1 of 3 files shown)


rwjblue:
With these changes, travis-web builds successfully (previously many transpilation errors occurred).

stefanpenner:
can we get a regression test in place?

rwjblue:
Yep, was already working on it between meetings....

lib/broccoli/ember-app.js
```diff
@@ -682,34 +682,46 @@ class EmberApp {
...
-    // operating on addon code where there weren't previously.
-    // This prevents that by using the default app compiler only.
+  _getDefaultPluginForType(type) {
     let plugins = this.registry.load(type);
-    return plugins[plugins.length - 1];
+    let defaultsForType = plugins.filter(plugin => plugin.isDefaultForType);
+
+    if (defaultsForType.length > 1) {
+      throw new Error(
+        `There are multiple preprocessor plugins that are registered for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
```
kellyselden: We should probably mention `... marked as default ...`, otherwise it doesn't seem very actionable.

lib/broccoli/ember-app.js
```diff
@@ -682,34 +682,46 @@ class EmberApp {
...
   }
 
   _compileAddonTemplates(tree) {
-    let plugin = this._getDefaultCompiler('template');
-    if (plugin) {
-      tree = plugin.toTree(tree, {
-        annotation: `_compileAddonTemplates`,
-        registry: this.registry,
-      });
+    let defaultPluginForType = this._getDefaultPluginForType('js');
```
kellyselden: Typo? js -> template

rwjblue: yep, copy pasta error

lib/broccoli/ember-app.js
```diff
@@ -682,34 +682,46 @@ class EmberApp {
...
-    // operating on addon code where there weren't previously.
-    // This prevents that by using the default app compiler only.
+  _getDefaultPluginForType(type) {
     let plugins = this.registry.load(type);
-    return plugins[plugins.length - 1];
+    let defaultsForType = plugins.filter(plugin => plugin.isDefaultForType);
+
+    if (defaultsForType.length > 1) {
+      throw new Error(
+        `There are multiple preprocessor plugins marked as default for '${type}': ${defaultsForType.map(p => p.name).join(', ')}`
```
twokul: `p.name` actually _does not_ reflect reality, we don't require pushing `name` property when we register preprocessors:

```javascript
registry.add('template', {
  isDefaultForType: true,
  ext: 'hbs',
  toTree(tree) { return tree; }
});
```

twokul: we could change the way we put together `precompiledSource` from `addonBundles` and get the `name` that way, or could just ask for additional `name` property during registration (the first option is more work)

tests/unit/broccoli/ember-app/app-and-dependencies-test.js
```diff
@@ -264,4 +272,79 @@ describe('EmberApp#appAndDependencies', function() {
...
+    app.registry.add('template', {
+      ext: 'hbs',
+      name: 'faulty-addon',
+      isDefaultForType: true,
+      toTree(tree) {
+        return tree;
+      },
+    });
+
+    yield co(function *() {
```
twokul: since we can't just use `expect().to.throw()`, resorted to combo of `yield` and `co`

lib/broccoli/ember-app.js
rwjblue: Seems like most things use a name already (ember-cli-babel does for sure)...

twokul: We will need to patch `ember-cli-babel` and `ember-cli-htmlbars` either way (as well as other preprocessors)

rwjblue:
Thanks for cleaning this up @twokul!

PR #7740, opened 2018-04-06
remove EMBER_TRY_SCENARIO=ember-default

kellyselden:
I could have swore @rwjblue already submitted this, but I couldn't find it. The usefulness of this extra CI job does not warrant the additional CI time IMO.


rwjblue:
Hmm. I actually think it is warranted, but I’d prefer to do it as a gateway stage...

rwjblue:
@Turbo87 - SG, I'll try to get a PR up for stages so we can discuss at our next meeting...

kellyselden:
This doesn't apply anymore.

PR #7744, opened 2018-04-11
Update `ember` & `ember data`

btecu:

PR #7747, opened 2018-04-11
Update ember-data and ember-source versions

twokul:
supersedes https://github.com/ember-cli/ember-cli/pull/7744


btecu:
wow, gg

PR #7749, opened 2018-04-12
remove trailing comma

kellyselden:
Accidentally targeted master on the last one.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -6,7 +6,7 @@ module.exports = function() {
   return Promise.all([
     getChannelURL('release'),
     getChannelURL('beta'),
-    getChannelURL('canary'),
+    getChannelURL('canary')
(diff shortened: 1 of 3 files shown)


twokul:
https://github.com/ember-cli/ember-cli/pull/7747

kellyselden:
It's probably the same issue, but the release branch. beta branch might suffer as well.

Turbo87:
was this already published with the trailing comma? I'm not sure it's worth the churn to fix it unless we enable the trailing comma rule of ESLint by default.

kellyselden:
> was this already published with the trailing comma?

yes.

> I'm not sure it's worth the churn to fix

I don't think churn is as much of a concern now that we have ember-cli-update. Also, I think it is worth it because projects that have a trailing-comma rule (all of mine) now get eslint errors with this version. I don't think it's worth putting in an exclusion everywhere for a file that accidentally violated a convention.

PR #7758, opened 2018-04-13
Allowing cwd to be passed to testem without being overridden by ember-cli

arthirm:

testem/testem#1219

Previously this commit caused the tests to run in only 1 partition because when ember-cli was passing progOptions to testem, it didnt transform the option's property names like testPage to test_page which has partition information from ember exam. Hence by default it ran in 1 partition.
ember exam does not fail if it is asked to run in multiple partitions but it ran only in 1 partition.


stefanpenner:
I'm good with beta.

PR #7762, opened 2018-04-15
Remove default value for `ember addon --yarn`

lennyburdette:
Like `ember new`, `ember addon` should not have a default value for the `--yarn` option so that it will search for a yarn.lock or Yarn workspace root unless the option is specified exactly.


lennyburdette:
Thanks for the quick review @kellyselden ! I fixed the broken tests, however one Travis part failed with a timeout running eslint.

kellyselden:
I've been seeing that lately. Reran.

PR #7766, opened 2018-04-16
Beta backport testem fix

stefanpenner:


Turbo87:
@stefanpenner the `beta` branch is currently not in the correct state since we haven't released a 3.2 beta yet. Since we will promote `master` to `beta` with that release the PR is already part of the next 3.2 beta.

stefanpenner:
@rwjblue  will that block our engines related bugs or will the "being disabled by default" guard us?

@Turbo87 ok, ETA for beta being in a correct state?

Turbo87:
> ETA for beta being in a correct state?

once someone manages to find time to do the beta release according to the RELEASE.md

I won't have time today or tomorrow, but can probably get it done on Wednesday if I remember

stefanpenner:
@Turbo87 having beta be inconsistent state for a large window seems like a risk. Should we consider updating the `RELEASE.md` to ensure these windows are strongly discouraged?

Turbo87:
> Should we consider updating the RELEASE.md to ensure these windows are discouraged?

👍

stefanpenner:
@Turbo87 no worries. I think an idea would be, if such a process is interrupted, to just open an issue with the details. That way the repo maintainers are aware, and can potentially lend a hand. Currently, it isn't clear how one could resume where you left off, or what the current state is.

PR #7769, opened 2018-04-18
add blueprints to linting coverage

kellyselden:
We got burned by the "node/no-unpublished-require" in blueprint files, so lets add it to lint coverage. I also exclude the templated files from eslint.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -75,7 +75,7 @@ module.exports = {
     contents.scripts['test:all'] = 'ember try:each';
 
     // add addon specific directories to lint:js script
-    contents.scripts['lint:js'] = 'eslint ./*.js addon addon-test-support app config lib server test-support tests';
+    contents.scripts['lint:js'] = 'eslint ./*.js addon addon-test-support app blueprints config lib server test-support tests';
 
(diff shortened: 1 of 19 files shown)


dfreeman:
Is there a reason we explicitly enumerate items to lint instead of having the script be `eslint .` with a `.eslintignore` file?

kellyselden:
I'm not sure, @rwjblue? Maybe because '/your-custom-folder' would start getting linted when it wasn't before and cause undo pain?

dfreeman:
I can only speak for myself and what I've seen with my team, but I've generally found that if I have directories full of JS files, I probably want them to be linted, and it's easier to treat opting out as the exception instead of the other way around.

With the current setup, bugs like this continue to creep in where we have code we assume is getting linted only to realize later that it wasn't.

Using `.eslintignore` also plays more nicely with editor integrations, since the warnings I do or don't see when I open up a file to edit will match what actually happens when I run `lint:js`.

rwjblue:
Nope, no good reason. I've also recently started using `eslint .` + `.eslintignore` in a number of projects specifically for the reasons @dfreeman mentions...

kellyselden:
I'll make the change after this and see if there are any additional gotchas.

PR #7774, opened 2018-04-19
remove ember-cli-shims in blueprint

xg-wang:
Remove [ember-cli-shims](https://github.com/ember-cli/ember-cli-shims) when addons have upgraded to ember-cli-babel >= 6.6.0

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -28,7 +28,6 @@
     "ember-cli-htmlbars-inline-precompile": "^1.0.0",
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-qunit": "^4.1.1",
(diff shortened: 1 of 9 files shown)


thoov:
@xg-wang I believe there are some fixtures that have `ember-cli-shims` in it that will need to be removed

xg-wang:
@thoov I'll look into the mu failure test

xg-wang:
fixed by #7771

Turbo87:
given that `master` will be 3.3 which is still about 9 weeks away and we've deprecated Babel 5, this seems like a reasonable change 🤔 

@rwjblue @kellyselden @twokul thoughts?

twokul:
yeah, I think it should be fine

rwjblue:
:shipit:

PR #7783, opened 2018-04-21
Update `in-repo-addon` blueprint to support module unification

GavinJoyce:
Part of https://github.com/ember-cli/ember-cli/issues/7530

This PR updates the `in-repo-addon` blueprint to generate an in-repo addon in the `packages` directory when the app is MU. It also introduces a new `packages` blueprint which will be the successor to the classic `lib` blueprint.

TODO:

 * [x] wrap MU tests with MU experiment check
 * [x] fix remaining test failures


blueprints/packages/files/packages/.jshintrc
```diff
@@ -0,0 +1,4 @@
+{
```
rwjblue: I don't think we should have both a `.jshintrc` _and_ an `.eslintrc`, can you remove the `.jshintrc.js` file?

blueprints/packages/files/packages/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
rwjblue: Generally speaking, we should not be adding `.eslintrc.js` files, instead we should rely on `overrides` configuration in the top level eslintrc.

blueprints/packages/files/packages/.jshintrc
GavinJoyce: 👍 

I should remove this from the `lib` blueprint too. Happy for me to do this in a follow on PR so that this remains focused?

blueprints/packages/files/packages/.eslintrc.js
GavinJoyce: I'm following the pattern of the existing `lib` blueprint. I can follow this PR up with changes to both `lib` and `packages`

rwjblue: sounds good

PR #7784, opened 2018-04-22
Remove private `javascript` method

twokul:
+ Introduce new packager helpers
+ Convert `app.import` tests to use packager

This is a commit from https://github.com/ember-cli/ember-cli/pull/7562 as it's much easier to land things one-by-one rather than in one big PR.

### DIFF
--- a/tests/helpers/default-packager.js
+++ b/tests/helpers/default-packager.js
@@ -1,0 +1,254 @@
+'use strict';
+
+/*
+ * A list of "helper" functions to automate tedious tasks of generating and
+ * validating folder structures.
+*/
+
+const assert = require('assert');
+
+const DEFAULT_ADDON_MODULES = {
+  'ember-cli-foobar': '^1.0.0',
+};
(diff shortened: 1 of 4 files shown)


tests/helpers/default-packager.js
```diff
@@ -0,0 +1,275 @@
...
+    'app.css': 'html,body{height:100%}',
+  },
+  templates: {
+    'application.hbs': '{{outlet}}',
+  },
+  config: {
+    'environment.js': `module.exports = function() { return { modulePrefix: 'the-best-app-ever' }; };`,
+  },
+  'package.json': JSON.stringify({
+    name: 'the-best-app-ever',
```
kellyselden: taken from other PR, does this need to be a complete package.json, or can we get by by only filling out necessary keys?

kellyselden: Also, I just saw that this is copied code, so I'm not as concerned about you fixing.

PR #7787, opened 2018-04-23
Remove private `_filterAppTree` method

twokul:
`_filterAppTree` was used to exclude templates and styles from
application tree. This should only be done when linting files and could
be omitted altogether for constructing `dist` tree.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -891,36 +891,6 @@ class EmberApp {
     });
   }
 
-  /**
-    Filters styles and templates from the `app` tree.
-
-    @private
-    @method _filterAppTree
-    @return {Tree}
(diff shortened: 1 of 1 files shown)


twokul:
@rwjblue well, I don't think it's dumb necessarily: filtering once and passing filtered tree to all the linters is not that bad of an idea.

> Presumably, a linter should only look at its own file types...

I agree here. Two things though:

+ that would mean extra operations (if you have `ember-cli-eslint` + `ember-cli-template-lint` + `ember-cli-sass-lint`, they will all run filtering)
+ we're going to have to tweak linter add-ons themselves

do we want to do it now?

rwjblue:
AFAIK, they already do this and wouldn’t require changes anyways....

PR #7793, opened 2018-04-26
Node 10 beta

stefanpenner:


Turbo87:
we can/should merge upwards once #7792 is merged

stefanpenner:
> we can/should merge upwards once #7792 is merged

whats the difference?

Turbo87:
it is less duplication. if you refer to that change somewhere you only have one link, instead of three links. the changelog generator will also only show this change once instead of three times.

stefanpenner:
> it is less duplication. if you refer to that change somewhere you only have one link, instead of three links. the changelog generator will also only show this change once instead of three times.

Sounds reasonable.

PR #7798, opened 2018-04-30
Add Broccoli 2.0 support

oligriffiths:
# What
Once https://github.com/broccolijs/broccoli/pull/363 merges, this PR adds support for broccoli 2.0 in ember-cli via `BROCCOLI_2` environment flag. If there is a better way to do this, please LMK and I will change.

The requisite change is to require the `broccoli` package, and to reformat the response from the broccoli `build()` method, into:
```js
{
  directory: node.outputPath,
  graph: node,
}
```

# Test
In my testing, the output of both the 0-18x series and master is the same, I verified this with:
```
ember build dist1
EMBER_CLI_BROCCOLI_2=1 ember build dist2
diff -rq dist1 dist2
```
It should output nothing.

Updated tests to validate the build results are the same, and support for system tmp dir.


Turbo87:
nice work! 👏 

> via BROCCOLI_MASTER environment flag. If there is a better way to do this, please LMK and I will change

we usually use https://github.com/ember-cli/ember-cli/blob/master/lib/experiments/index.js for these things, though that does not use env vars (yet). I would recommend migrating the feature flag to that file but instead of using the `symbol()` code there keep using the env var that you already have.

oligriffiths:
@Turbo87 Done. What's the best way to CI test this? Is it possible to re-run all tests with and without the env var set?

Turbo87:
> What's the best way to CI test this? Is it possible to re-run all tests with and without the env var set?

What we currently do is: on `master` all experiments are enabled, on `beta` and `release` all experiments are supposed to be disabled. That led to `beta` currently being broken since the tests passed with the Module Unification experiment being enabled, but when I disabled the experiment for the beta release the test suite started breaking.

What I'm meaning to say: what we currently do is not great and I think running the test suite with experiments turned off by default would be better and then adding another CI job per experiment.

@ember-cli/core thoughts?

mansona:
@Turbo87 I don't know how crazy you want to get with the CI setup but I think that we should be considering running a full on/off matrix of experiments on each CI build, that way we would catch any potential interactions between experiments.

I don't think it needs to go so far as to test each node version with the full matrix of experiments but maybe we could pick one node version and then run the experiment matrix off it as a smoke test? 

The current issue with the Beta branch was introduced on 1st March and would have shown on any test that disabled the MU experiment.

I'm happy to help set this up 😄

Turbo87:
My approach would be:

- one job per node supported version running without experiments
- one job per experiment running the fastest node version (10)
- one job running with all experiments enabled on the fastest node version

Certainly can't catch all issues, but seems like a reasonable compromise. PR welcome, but we should discuss the details in the CLI call on Thursday.

oligriffiths:
This approach seems to make sense and provides a decent balance between every permutation of experiments thus taking longer to the suite to pass and coverage of conditions. 

In my case, there’s a bug somewhere as the test suite doesn’t pass with the experiment enabled.

rwjblue:
I submitted https://github.com/ember-cli/ember-cli/pull/7803 for refactoring the CI runs...

lib/experiments/index.js
```diff
@@ -6,6 +6,7 @@ const symbol = require('../utilities/symbol');
 let experiments = {
   MODULE_UNIFICATION: symbol('module-unification'),
   DELAYED_TRANSPILATION: symbol('delayed-transpilation'),
+  BROCCOLI_MASTER: process.env.BROCCOLI_MASTER === '1', // Note: EXPERIMENTAL, DO NOT USE!
```
rwjblue: Hehe, these are all experimental 😉

lib/models/builder.js
```diff
@@ -44,7 +44,12 @@ class Builder extends CoreObject {
     this.environment = this.environment || 'development';
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
 
-    const broccoli = require('broccoli-builder');
+    let broccoli;
+    if (this.broccoliMaster) {
```
rwjblue: I'd rather do the experiments check here, instead of in the build task

```js
if (experiments.BROCCOLI_MASTER) {

} else {

}
```

lib/models/builder.js
```diff
@@ -152,6 +157,17 @@ class Builder extends CoreObject {
 
     return this.processAddonBuildSteps('preBuild')
       .then(() => this.builder.build(willReadStringDir))
+      .then(node => {
+        // broccoli-builder reformats the response into {directory, graph}
+        if (this.broccoliMaster) {
```
rwjblue: Lets do the experiments guard here directly also

lib/tasks/build.js
```diff
@@ -17,6 +18,7 @@ class BuildTask extends Task {
       outputPath: options.outputPath,
       environment: options.environment,
       project: this.project,
+      broccoliMaster: experiments.BROCCOLI_MASTER,
```
rwjblue: Should be able to revert changes to this file once the builder itself does its own experiments check...

lib/tasks/serve.js
```diff
@@ -26,6 +27,7 @@ class ServeTask extends Task {
       outputPath: options.outputPath,
       project: this.project,
       environment: options.environment,
+      broccoliMaster: experiments.BROCCOLI_MASTER,
```
rwjblue: Should be able to revert changes to this file once the builder itself does its own experiments check...

lib/experiments/index.js
rwjblue: Lets make the flag: `process.env.EMBER_CLI_BROCCOLI_1` (I don't like using `_MASTER` conceptually here)...

rwjblue:
FYI - I just released broccoli@2.0.0-beta.2 (updates deps, drops RSVP, drops Node 4).

rwjblue:
@oligriffiths - I just rebased, added the bump commit to broccoli@2.0.0-beta.2, and force pushed to kick off a new CI run. I think this is _very close_ to being landable...

bin/ember
```diff
@@ -38,5 +38,10 @@ resolve('ember-cli', {
   }).then(function(result) {
     var exitCode = typeof result === 'object' ? result.exitCode : result;
     exit(exitCode);
+  }).catch((err) => {
+    // should not normally reach here
+    // this handles the case where console-ui is broken
+    console.error(err);
```
stefanpenner: In general we would prefer to not mutate this file. But this may be an exception, @rwjblue  c/d?

lib/models/builder.js
```diff
@@ -26,7 +27,11 @@ class Builder extends CoreObject {
   constructor(options) {
     super(options);
 
+    this.broccoli2 = Boolean(options.broccoli2 === undefined ? experiments.BROCCOLI_2 : options.broccoli2);
```
rwjblue: I think it would be better to remove `this.broccoli2` completely, and replace it with direct `experiments.BROCCOLI_2` checks...

lib/models/builder.js
```diff
@@ -26,7 +27,11 @@ class Builder extends CoreObject {
...
+    this.systemTemp = Boolean(options.systemTemp === undefined ? experiments.SYSTEM_TEMP : options.systemTemp);
```
rwjblue: Same RE: experiment checks. In general it should not be possible to force these experiments to true without `process.env.EMBER_CLI_*` is set...

lib/models/builder.js
```diff
@@ -226,6 +253,58 @@ class Builder extends CoreObject {
...
+
+  /**
+   * broccoli-builder reformats the response into {directory, graph}, this method is a backwards
+   * compatible shim for broccoli 1.x
+   * @private
+   * @method compatNode
+   * @param node The node returned from Broccoli builder
+   */
+  compatNode(node) {
+    if (this.broccoli2) {
```
rwjblue: I was hoping that we could structure the code such that the broccoli@2 case is the "normal" case, and we do any shimming/polyfilling for the older broccoli-builder APIs...

package.json
```diff
@@ -130,6 +131,7 @@
     "chai-as-promised": "^7.0.0",
     "chai-files": "^1.2.0",
     "co": "^4.6.0",
+    "dir-compare": "^1.4.0",
```
rwjblue: What are we using this for?

lib/models/builder.js
oligriffiths: I couldn’t see how to mock that property in tests. I want to compare with the property on and off in the same test.

package.json
oligriffiths: Comparing the directory structure and files for a Broccoli-builder vs broccoli 2.0 build to ensure they’re identical.

rwjblue: Hmm, I think in general I'd prefer to use something like:

```
expect(fixturify.readSync(oneDir)).to.deep.equal(fixturify.readSync(twoDir))
```

oligriffiths: the more you know. Will test that.

oligriffiths: Will that compare file contents/hash?

lib/models/builder.js
rwjblue: Why? I think having tests around the expected behaviors should be enough. In general, from the outside, I don't think which builder we use should actually matter much, right?

oligriffiths: Well, we want to ensure that the files are exactly the same between both versions.
I guess if we have a fixture directory and compare against that, that'd also work.

rwjblue: Totally agreed. I think the goal here should be that the migration is seemless and the output is identical. And we should be able to do that via fixtures (either inside the file system, or just via fixturify in the test itself)..

oligriffiths: Cool. Are we doing this anywhere already?

bin/ember
stefanpenner: thanks

package.json
oligriffiths: @rwjblue are there already tests that already assert the output of the build is correct?

lib/models/builder.js
oligriffiths: I think we should do that once we move out of experimental mode

lib/models/builder.js
```diff
@@ -66,6 +66,9 @@ class Builder extends CoreObject {
       };
     } else {
       broccoli = require('broccoli-builder');
+      if (this.systemTemp) {
+        console.warn('EMBER_CLI_SYSTEM_TEMP only works in combination with EMBER_CLI_BROCCOLI_2');
```
stefanpenner: we should simply error here

oligriffiths: Will need to isolate tests in that case, because we can't do only combo environment variables (e.g. BROCCOLI_2 and SYSTEM_TEMP). Right now, the SYSTEM_TEMP flag will cause this error when being run without setting BROCCOLI_2.
Will do a little refactor.

oligriffiths: Done

tests/unit/models/builder-test.js
```diff
@@ -202,92 +203,70 @@ describe('models/builder.js', function() {
...
-        broccoli2: true,
       });
 
       return builder.build().then(function(result) {
         expect(Object.keys(result)).to.eql(['directory', 'graph']);
-        expect(result.graph.__heimdall__).to.not.be.undefined;
+        if (experiments.BROCCOLI_2) {
+          expect(result.graph.__heimdall__).to.not.be.undefined;
+        } else {
+          expect(result.graph.constructor.name).to.equal('Node');
```
krisselden: why?

oligriffiths: Not sure I understand the question...

rwjblue:
Updated to ensure `SYSTEM_TEMP` experiment only runs with `BROCCOLI_2`.

rwjblue:
Just rebased against master to get some of the CI fixes that have landed...

oligriffiths:
@rwjblue tests all passing 🎉

oligriffiths:
@krisselden I had to remove your changes to get tests passing. Things have changed a bit since you implemented them, and im not 100% sure what they were changing. So if you wouldn't mind testing this branch now, and see if you get the same errors, I'd appreciate it.

rwjblue:
Rebased and updated broccoli to 2.0.0-beta.3...

oligriffiths:
Whoop 🎉

PR #7801, opened 2018-05-01
Update node support policy docs.

rwjblue:
* Drop 4
* Drop 5
* Drop 7
* Update EOL dates for 6 and 9
* Add 10


Turbo87:
should we target `release` with this?

rwjblue:
Sure I will update

rwjblue:
OK, updated.

PR #7803, opened 2018-05-01
Migrate to using Travis stages.

rwjblue:
* One stage (running first) for testing `yarn test` in Node 10 without experiments.
* _IIF_ that passes, run `yarn test-all` for the following:
  * Node 6
  * Node 8
  * Node 9
  * Node 10 with code coverage
  * Node 10 with all experiments enabled together
  * Node 10 with only MODULE_UNIFICATION experiment enabled
  * Node 10 with only DELAYED_TRANSPILATION experiment enabled
* _IIF_ all prior builds pass *AND* running on release branch, run deploy script


lib/experiments/index.js
```diff
@@ -1,11 +1,22 @@
...
+    return true;
+  }
+
+  return false;
+}
 
 let experiments = {
-  MODULE_UNIFICATION: symbol('module-unification'),
-  DELAYED_TRANSPILATION: symbol('delayed-transpilation'),
+  MODULE_UNIFICATION: isExperimentEnabled('MODULE_UNIFICATION'),
```
Turbo87: we will likely need to update the MU usage instructions, because now canary does not use these experiments by default anymore

Turbo87:
unfortunately this now seems to fail with the same issue as our `beta` branch due to the disabled MU experiment

.travis.yml
```diff
@@ -6,10 +6,6 @@ addons:
 
 language: node_js
 node_js:
-  - "4"
-  - "6"
-  - "8"
-  - "9"
```
Turbo87: We also need to check the code that parses this file and extracts the supported Node versions.

rwjblue:
> unfortunately this now seems to fail with the same issue as our beta branch due to the disabled MU experiment

Yep, but thats a good thing!  I paired with @real_ate and @mixonic a few minutes ago and got the main MU failures fixed (should be a PR open soon). There are still a couple failures from `DELAYED_TRANSPILATION` that I need to circle around with @kellyselden on...

.travis.yml
rwjblue: Great catch, I'll check that

docs/experiments.md
```diff
@@ -1,38 +1,53 @@
-# Experiments
+ Experiments
```
Turbo87: intentional change?

rwjblue: No, fixing...

rwjblue:
OK, fixed up issues in beta which _should_ also fix the failures we were getting here.

Turbo87:
looks like we're having some other issues still 🤔

mansona:
@Turbo87 looks like a node module issue, I've just run it locally and it seems to work fine. Are you able to just restart the build?

Turbo87:
build restarted, but still failing with the same issue. looks like this is a genuine issue with the changes in this PR

Turbo87:
hmm, one downside of the stages approach is that CI times are significantly worse since the next stage will only start when the previous one has finished, while previously all jobs were running in parallel. Until now the CI runs for a PR took around 22 min., now they are beyond 50 min. I don't see any immediate advantage from running in stages here aside from it looking a little more structured, so maybe we should reconsider running all jobs in parallel again. 🤔

Turbo87:
```
/home/travis/.travis/job_stages: line 57: ./travis/deploy.sh: No such file or directory
```

rwjblue:
I’ll squash it down to two stages instead of 4. That will reduce the total clock time. 

I do think that it is good to gate the majority of jobs on having a single passing test run though...

rwjblue:
Also, not sure what’s up with `./travis/deploy.sh` not being found. Will poke at that shortly...

Turbo87:
> not sure what’s up with ./travis/deploy.sh not being found

It's `./.travis` instead of `./travis`

> I do think that it is good to gate the majority of jobs on having a single passing test run though...

that drops all concurrency advantages though because jobs can not run in parallel until the first job passed

rwjblue:
> that drops all concurrency advantages though because jobs can not run in parallel until the first job passed

Not _all_ concurrency advantages. Just some ;). The benefit is not kicking off those jobs (and therefore not backing up CI) when even a basic test run doesn’t pass.

rwjblue:
Another possibility that still would eliminate a non-trivial number of extra jobs when basic things are broken would be to run linting and non-slow tests in the first build, and then kick off all subsequent jobs (test-all:slow) in parallel.

Turbo87:
sounds like a good compromise 👍

rwjblue:
Updated to do:

* One stage (running first) for testing `yarn test` in Node 10 without experiments.
* _IIF_ that passes, run `yarn test-all` for the following:
  * Node 6
  * Node 8
  * Node 9
  * Node 10 with code coverage
  * Node 10 with all experiments enabled together
  * Node 10 with only MODULE_UNIFICATION experiment enabled
  * Node 10 with only DELAYED_TRANSPILATION experiment enabled
* _IIF_ all prior builds pass *AND* running on release branch, run deploy script

Lets see how that works out in total clock time...

oligriffiths:
@rwjblue is it worth being able to exclude an experiment from testing. The reason I ask, is that in order to land https://github.com/ember-cli/ember-cli/pull/7798 we need to add instrumentation (heimdall) support, but that's in progress. So https://github.com/ember-cli/ember-cli/pull/7798 wouldn't be able to land without doing the heimdall support.

lib/experiments/index.js
rwjblue: Yep, I discussed this with @mixonic and @real_ate the other day. I _believe_ that @mixonic is on it...

rwjblue:
> Coverage decreased (-0.9%) to 91.543%

O_o - I didn't remove any tests!?!?!

mansona:
@rwjblue you removed the `tests/experiments.js` file so maybe it was being counted and the overall ratio has changed? ¯\_(ツ)_/¯

PR #7804, opened 2018-05-01
removing extra `app` folder check and error message

mansona:
This is reverting a small change that was introduced in this commit https://github.com/ember-cli/ember-cli/commit/0ab09675e476dd270968c31af8dd5b58126c4bc4#diff-392bb207fea8bcbeb4a084bb445f7cc7R253

After a discussion today with @rwjblue and @mixonic we decided that we would re-introduce this check on master because there are a few other things that will need to be improved/changed at the same time and we didn't want to introduce this on the beta branch.


rwjblue:
Brings us down to 4 failures which are related to delayed transpilation.

PR #7807, opened 2018-05-02
Modify development server compression behaviors.

rwjblue:
* Allow opting out of compression via `ember serve --no-compression`
* Avoid adding compression for server sent events

---

From the `compression` packages README:

> Because of the nature of compression this module does not work out of the box with server-sent events. To compress content, a window of the output needs to be buffered up in order to get good compression.  Typically when using server-sent events, there are certain block of data that need to reach the client.

There are work arounds to this (every middleware that handles SSE's would have to immediately call `res.flush()`), but they do not work when proxying (because it is impossible for the proxy itself to know when `.flush()` should be called.

This change updates the compression code to simply disable itself for `text/event-stream` (which is what SSE uses as its mime type).

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -35,6 +35,7 @@ module.exports = Command.extend({
     { name: 'ssl',                  type: Boolean, default: false },
     { name: 'ssl-key',              type: String,  default: 'ssl/server.key' },
     { name: 'ssl-cert',             type: String,  default: 'ssl/server.crt' },
+    { name: 'compression',          type: Boolean, default: true,                                                                              description: 'Set to false to disable compression' },
   ],
 
   run(commandOptions) {
--- a/lib/tasks/server/express-server.js
+++ b/lib/tasks/server/express-server.js
@@ -191,17 +191,19 @@ class ExpressServerTask extends Task {
 
   startHttpServer() {
     this.app = this.express();
-    const compression = require('compression');
-    this.app.use(compression({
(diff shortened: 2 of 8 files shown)


Turbo87:
do we need a dedicated CLI option for this when we know that `text/event-stream` is what we need to have a special case for?

rwjblue:
> do we need a dedicated CLI option for this when we know that text/event-stream is what we need to have a special case for?

Well, when I started I didn't actually know that. To me, it seems reasonable to be able to opt-out of compression completely.  I'm happy to split that off though and focus initially on fixing the bug with SSE (targeting release most likely).

Turbo87:
> To me, it seems reasonable to be able to opt-out of compression completely.

for what use-case? 🤔

rwjblue:
I split the SSE bugfix into a separate PR: https://github.com/ember-cli/ember-cli/pull/7809

rwjblue:
> for what use-case? 🤔

Well, one example is that compression is actually _slower_ when simply using localhost...

Turbo87:
Shouldn't we generally disable it then?

rwjblue:
Hehe, then it wouldn't be possible to run performance tests with simulated slow networks. 😺 

I totally agree that having better defaults is good, I'm just not sure that we can get this right _enough_ to avoid needing to be able to customize in project...

Turbo87:
@rwjblue do you want to rebase and open up for discussion again?

PR #7809, opened 2018-05-02
Do not attempt to compress server sent events.

rwjblue:
From the `compression` packages README:

> Because of the nature of compression this module does not work out of
> the box with server-sent events. To compress content, a window of the
> output needs to be buffered up in order to get good compression.
> Typically when using server-sent events, there are certain block of data
> that need to reach the client.

There are work arounds to this (every middleware that handles SSE's would have to immediately call `res.flush()`), but they do not work when proxying (because it is impossible for the proxy itself to know when `.flush()` should be called

This change updates the compression code to simply disable itself for `text/event-stream` (which is what SSE uses as its mime type).


tests/unit/tasks/server/express-server-test.js
```diff
@@ -284,8 +284,43 @@ describe('express-server', function() {
...
+        };
+
+        subject.start({
+          proxy: 'http://localhost:3001/',
+          host: undefined,
+          port: '1337',
+          rootURL: '/',
+          compression: true,
+        })
+          .then(function() {
```
Turbo87: Can we use co instead of the promise chain? and possibly promisify request?

rwjblue: Good idea, done.

PR #7811, opened 2018-05-04
replace TRAVIS with CI

kellyselden:
This affects our gitlab runner too, and since we already use `process.env.CI` in the "targets.js" file, it also makes sense to be consistent.


Turbo87:
I'm a little worried about this change since it disables the sandbox on all CI systems, not just the ones that use sandboxed containers anyway... 🤔

jacobq:
FWIW, having this as `process.env.TRAVIS` prevented my tests from running when building on AppVeyor. Changing to `process.env.CI` resolved it, so I am happy to see that this is now the default.

PR #7813, opened 2018-05-06
speed up addon initialization by using PackageInfoCache instead of Addon.lookup

dcombslinkedin:
With the new PackageInfoCache functionality, it is possible to use the information in the cache to create AddonConstructor instances for addons, rather than relying on Addon.lookup() to do it. That method relies on checking with the file system, which is exactly what the PackageInfoCache has already done, so there is no need to do so any again.

The new functionality is primarily a new 'getAddonConstructor' method in package_info.js, and removal of Addon.lookup() and Addon.resolvePath(), which were marked as @private methods. I also removed the tests for lookup() and resolvePath() and created several new similar ones in package-info-cache-tests.js. Finally, I was able to move addons-factory.js to a single static method, so I moved the file from addons-factory.js to instantiate-addons.js.


twokul:
looks good at the first glance

lib/models/instantiate-addons.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {Object} parent an Addon or Project that is the direct containing object of the list
+ *   of children defined in addonPackages.
+ * @param {Project} project the project that contains the parent (so either the addon's project
+ *   if parent is an addon, or the project itself if it is a project). It is possible when
+ *   constructing custom addon instances that the project will actually be undefined--various
+ *   addon tests do this, for example.
+ * @param {Object} a map of addon name (including scope) to an AddonInfo with the name, path and
+ *   'pkg' object for that addon's package.json). These are what is turned into addons.
+ */
+module.exports = (parent, project, addonPackages) => {
```
stefanpenner: nit-pick: As this doesn't intend to close over `this`, I would prefer if this is a named `function` statement

stefanpenner:
looked like maybe an intermittent test failure, restarting the job that failed.

dcombslinkedin:
Thanks, Stef! I *really* appreciate it.

PR #7817, opened 2018-05-06
Enable `prefer-destructuring` rule

twokul:


lib/commands/destroy.js
```diff
@@ -28,8 +28,8 @@ module.exports = Command.extend({
   beforeRun: mergeBlueprintOptions,
 
   run(commandOptions, rawArgs) {
-    let blueprintName = rawArgs[0];
-    let entityName = rawArgs[1];
+    let [blueprintName] = rawArgs;
+    let [, entityName] = rawArgs;
```
kellyselden: looks like the autofixer screwed up here

lib/commands/destroy.js
```diff
@@ -61,7 +61,7 @@ module.exports = Command.extend({
   },
 
   printDetailedHelp() {
-    let ui = this.ui;
+    const { ui } = this;
```
kellyselden: keep as let

lib/errors/silent.js
```diff
@@ -6,9 +6,11 @@ const deprecate = require('../utilities/deprecate');
 Object.defineProperty(module, 'exports', {
   get() {
     // Get the call stack so we can let the user know what module is using the deprecated function.
+    /* eslint-disable prefer-destructuring */
     let stack = new Error().stack;
```
kellyselden: `let { stack } = new Error();` won't work here?

lib/models/command.js
```diff
@@ -459,7 +459,7 @@ let Command = CoreObject.extend({
         key = alias;
         value = [`--${option.name}`];
       } else if (aliasType === 'object') {
-        key = Object.keys(alias)[0];
+        [key] = Object.keys(alias);
```
kellyselden: it's interesting this is allowed but

```
    // eslint-disable-next-line prefer-destructuring
     version = require(path.join(emberCLIBabelInstance.root, 'package')).version;
```

above isn't. hmm.

lib/tasks/server/livereload-server.js
```diff
@@ -21,12 +21,12 @@ function createServer(options) {
...
 
 function isNotRemoved(entryTuple) {
-  let operation = entryTuple[0];
+  let [operation] = entryTuple;
   return operation !== 'unlink' && operation !== 'rmdir';
 }
 
 function isNotDirectory(entryTuple) {
-  let entry = entryTuple[2];
+  let [, , entry] = entryTuple;
```
kellyselden: that's a little unfortunate.

Turbo87:
I'd prefer not to use this rule as it can lead to much less readable code in some instances as @kellyselden has already pointed out. I'm okay with using destructuring if it improves readability, but I'm against enforcing it.

twokul:
@Turbo87 I generally don't like inconsistency throughout (we have a fair share of it in our codebase) and I would prefer a mix of enforce + `eslint-disable` when needed.

That being said I can close if that's not something we want to do 👩‍💻

Turbo87:
I would personally vote for closing, but let's wait what the others have to say.

lib/tasks/server/middleware/broccoli-serve-files/package.json
```diff
@@ -3,6 +3,9 @@
   "keywords": [
     "ember-addon"
   ],
+  "engines": {
```
kellyselden: looks like these snuck in

twokul:
@ember-cli/core thoughts?

Turbo87:
closing due to inactivity

PR #7819, opened 2018-05-08
Move styles packaging

twokul:


lib/broccoli/ember-app.js
```diff
@@ -1725,21 +1675,23 @@ class EmberApp {
     let trees = [
       this.getAppJavascript(),
       this.getAddonTemplates(),
+      this._debugTree(this.getAddonStyles(), 'addon-styles'),
```
kellyselden: why this one debug and not the others?

twokul: removed

PR #7822, opened 2018-05-10
Moves tests packaging out of `ember-app`

twokul:


kellyselden:
Seems good at first glance.

Issue #7800, opened 2018-04-30
CSS Imports stop working with Ember-CLI 3.1


hinti:
CSS Imports stopped working with Ember-CLI 3.1 (Works with Ember-CLI 3.0)

**Steps to reproduce (Windows 10)**

```
ember new cssimport
cd cssimport
echo @import url("general.css"); >> app\styles\app.css
echo body { background: green; } > app\styles\general.css
yarn build -prod
```

**Outcome of Production build**

```
cleaning up...
Build failed.
Build Error (CleanCSSFilter) in assets/cssimport.css

Broken @import declaration of "general.css"

Stack Trace and Error Report: C:\.../error.dump.2f53f5cb9cc11605612cee8836a9fd88.log

```

**Stack Trace**

```
 ERROR Summary:
  - broccoliBuilderErrorStack: Error: Broken @import declaration of "general.css"
    at arrayToError (C:\...\cssimport\node_modules\array-to-error\index.js:45:15)
    at minifyCallback (C:\...\cssimport\node_modules\clean-css-promise\index.js:33:20)
    at whenSourceMapReady (C:\...\cssimport\node_modules\clean-css\lib\clean.js:139:16)
    at Object.whenDone (C:\...\cssimport\node_modules\clean-css\lib\clean.js:155:14)
    at processNext (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:105:13)
    at importFrom (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:79:10)
    at processNext (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:104:16)
    at inlineLocalResource (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:355:12)
    at inline (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:218:10)
    at importFrom (C:\...\cssimport\node_modules\clean-css\lib\imports\inliner.js:74:12)
  - codeFrame: Broken @import declaration of "general.css"
  - errorMessage: Build Canceled: Broccoli Builder ran into an error with `CleanCSSFilter` plugin. ðŸ’¥
Broken @import declaration of "general.css"

```

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.1.2
http_parser: 2.8.0
node: 8.11.1
v8: 6.2.414.50
uv: 1.19.1
zlib: 1.2.11
ares: 1.10.1-DEV
[...]

Issue #37, opened 2014-03-04
consider setting a lower bound on node versions we support


stefanpenner:
we should atleast warn people if they are using an old node...

PR #7823, opened 2018-05-10
BUGFIX #7800 propagate minifyCSS options to CleanCSS

sranka:
This PR contains a fix for https://github.com/ember-cli/ember-cli/issues/7800. The root cause is a regression defect introduced in Ember 3.1 by https://github.com/ember-cli/ember-cli/commit/b0727ec5ea953c57f83178bb48bbd654693cab0c#diff-392bb207fea8bcbeb4a084bb445f7cc7

The "minifyCSS' options were properly passed to CleanCSS module.  This is herein fixed in default-packer.js using the commit diff that introduces this regression defect. The style-tests.js was enhanced to test the documented (https://ember-cli.com/asset-compilation#css) **processImport** flag in "minifyCSS" options.


lib/broccoli/default-packager.js
```diff
@@ -601,8 +601,9 @@ module.exports = class DefaultPackager {
       });
 
       if (cssMinificationEnabled === true) {
-        preprocessedStyles = preprocessMinifyCss(preprocessedStyles, options);
-        vendorStyles = preprocessMinifyCss(vendorStyles, options);
+        options.minifyCSS.registry = options.registry;
```
twokul: I think we should tweak `minifyCSS` options in `ember-app`, when they are passed to the constructor

```javascript
let minifyCSS = this.options.minifyCSS;
minifyCSS.registry = this.options.registry;

new DefaultPackager({
  minifyCSS
});
```

tests/unit/broccoli/default-packager/styles-test.js
```diff
@@ -70,7 +71,9 @@ describe('Default Packager: Styles', function() {
   }));
 
   afterEach(co.wrap(function *() {
-    yield output.dispose();
+    if (output) {
```
twokul: typically that means that `output` wasn't assigned, and if tests work correctly, we don't need this `if`. why did you add it?

tests/unit/broccoli/default-packager/styles-test.js
```diff
@@ -93,7 +96,7 @@ describe('Default Packager: Styles', function() {
 
       minifyCSS: {
         enabled: true,
-        options: {},
+        options: { processImport: false },
```
twokul: don't think this is relevant to caching test

tests/unit/broccoli/default-packager/styles-test.js
```diff
@@ -151,7 +155,7 @@ describe('Default Packager: Styles', function() {
     ).to.equal('body { height: 100%; }');
     expect(
       outputFiles.assets['the-best-app-ever.css'].trim()
-    ).to.equal('html { height: 100%; }');
+    ).to.equal(MODULES['the-best-app-ever'].styles['app.css']);
```
twokul: could you revert this change and match against the actual content? makes it easier to understand w/o having to scroll up

tests/unit/broccoli/default-packager/styles-test.js
sranka: Failing tests hide the original error when output is undefined, that's why ... mean a cleanup can be performed only on smth that exists. Not an important change though.

lib/broccoli/default-packager.js
twokul: spread should work in node 6+

sranka:
I reverted changes in afterTest of styles-test.js (the 1st minor), the second minor thing (reordering of registry option setup for minifyCSS) is smth in which I don't understand the 'how' and 'why' part.

samselikoff:
I'm trying to understand the changes here & in v3.1, and I'm a little over my head.

3.1 seemed to break `@import 'tailwind.css'` over in Ember CLI Tailwind. [Check out issue #37](https://github.com/embermap/ember-cli-tailwind/issues/37) for more details.

I'm trying out everything on Ember CLI 3.4.0-beta.2 to see if this PR fixed things, but I'm hitting a new bug in `development` environment (which was working in 3.1):

image

This is what we see in Chrome, and in Safari we just see a 404 for assets/tailwind.css. I don't think the built file ends up in `/assets` at all.

I'm not sure if all of this is related to how we're building `tailwind.css` or the relevant changes in Ember CLI-land.

Any ideas? Could someone point me in the right direction?

PR #7828, opened 2018-05-14
Change tree passed to `package` hook

twokul:
+ `addon-tree-ouput` folder is now `node_modules`

screen shot 2018-05-13 at 8 26 42 pm

### DIFF
--- a/docs/BUILD_PIPELINE_DEBUGGING.md
+++ b/docs/BUILD_PIPELINE_DEBUGGING.md
@@ -19,12 +19,12 @@ For a brand new ember application (`ember new your-app-name`), `DEBUG/` folder i
 ```sh
 DEBUG/
 └── bundler:application-and-dependencies
-    ├── addon-tree-output
+    ├── node_modules
     ├── tree-shake-test
     └── vendor
 ```
 
-`addon-tree-output` is the folder that contains all trees from Ember CLI add-ons;
+`node_modules` is the folder that contains all trees from Ember CLI add-ons;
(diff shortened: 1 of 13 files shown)


kellyselden:
The two changes seem unrelated to me. Should they be 2 PRs?

kellyselden:
Does this affect any `app.import`ing from node_modules, or any future native importing from node_modules? In other words, that the tree/folder would be shared by ember addons and non ember addons?

twokul:
it shouldn't but you made me thinking of something. 

hold off on merging this one, I should add more test coverage 😄

rwjblue:
@twokul - Where are we on this?

twokul:
@rwjblue @kellyselden one tricky bit about renaming to `node_modules` is glob patterns like `node_modules/**/*.js` or `node_modules/**/*.css`. we obviously don't want _all_ of node modules in our vendor files. one idea is to reach into `project.addons` and include _only_ ember-cli add-ons

rwjblue:
Sorry, @twokul I missed this before and it now has merge conflicts. Would you mind rebasing?

twokul:
@rwjblue will do this weekend

kellyselden:
Once this merges, my custom packager will (obviously) start breaking, so I would like to test this branch out with modifications to my packager to see how it works. I don't anticipate any issues.

kellyselden:
Should "addon-test-support" change too?

rwjblue:
I think at this point, it probably shouldn't be changed.

PR #7829, opened 2018-05-14
Move styles from add-ons under `app-name`

twokul:
Before this change the structure of the tree that is passed in to `package` hook was

```
addon-modules/
vendor/
bower_components/
app-name/
app/
  styles/
    ...
```

This change will ensure we don't have an orphan `app/styles` folder since it's confusing and inconsistent.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1006,9 +1006,13 @@ class EmberApp {
    * @return {BroccoliTree}
   */
   getAddonStyles() {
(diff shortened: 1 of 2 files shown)


kellyselden:
Can we have a little rationale is to why this is needed?

twokul:
@kellyselden updated the description

twokul:
@kellyselden `addonStyles` returns files that are automatically re-exported as a part of app's namespace. 

added test coverage for `getAddonStyles`

PR #7834, opened 2018-05-15
Add yarn.lock to npmignore on addons blueprint

josemarluedke:
I saw @rwjblue and @Turbo87 talking about this in slack.


josemarluedke:
I have added to the MU addon blueprint as well. App blueprints don't have the `npmignore` file, so we are good there.

rwjblue:
Thank you!

PR #7835, opened 2018-05-17
docs/RELEASE: Remove obsolete instructions about experiments

Turbo87:
These instructions are obsolete now because of #7803

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -95,7 +95,6 @@ git checkout -B master --track origin/master
 # Merge the new beta release into the "master" branch
 git merge vX.Y.0-beta.1
 
-# Confirm that experiments are turned on before pushing it back to master.
 # Push back upstream.
 git push origin
 ```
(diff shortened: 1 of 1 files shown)


Turbo87:
> when an experiment is not enabled by default (e.g. we are still iterating on the feature so haven't decided to make it a default) we will not be backporting fixes to beta/release.

I'm not sure how that is related. 🤔 

> we should still change the lib/experiments/index.js to ensure that once an experiment rolls to beta it cannot be enabled.

this has always been a major source of confusion when promoting canary to beta and even worse when merging a new beta into `master`. I'd prefer if we didn't have to manually touch the file during releases anymore, which would also make automatic releasing via CI a little easier in the end.

rwjblue:
>  I'd prefer if we didn't have to manually touch the file during releases anymore, which would also make automatic releasing via CI a little easier in the end.

Totally agree.  How can we detect that we are using a non-tagged version programmatically?

Turbo87:
> How can we detect that we are using a non-tagged version programmatically?

I don't think that's necessary. IMHO we could just keep the experiments in the releases, since they will be disabled by default. If people reeeeally want to use them in the stable release they will find a way to enable them either way...

rwjblue:
>  If people reeeeally want to use them in the stable release they will find a way to enable them either way

🤔 maybe seems fine? As long as we agree that bugfixes for non-enabled experiments do not get "fast-tracked" into beta/release...

PR #7841, opened 2018-05-22
Only import jquery if `@ember/jquery` is a dependency

cibernox:
This will only works for ember-source >= 3.3.0

The logic is:
1. If `@ember/jquery` is present, exit. The addon will know better.
2. If the app is NOT using `ember-source` means it's an old version using bower, so import jquery from bower path.
3. If the app is using `ember-source` but it's not using `@ember/optional-features`, add jquery.
4. If the app is using `ember-source`, it's also using `@ember/optional-features` and the jquery flag is enabled, also add jquery.
5. In the cases where jquery is added by ember-cli, a deprecation message is shown asking users to install `@ember/jquery`

**TODO:** Add tests.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -465,19 +465,22 @@ class EmberApp {
   }
 
   _addJqueryInLegacyEmber() {
-    let emberJQuery = this.project.findAddonByName('@ember/jquery');
-    if (emberJQuery) {
-      this.vendorFiles['jquery.js'] = emberJQuery.paths.jquery;
-    } else {
-      let ember = this.project.findAddonByName('ember-source');
-      if (ember) {
-        if (semver.lt(this.project.addonPackages['ember-source'].pkg.version, '3.3.0')) {
-          this.vendorFiles['jquery.js'] = ember.paths.jquery;
-        }
-      } else {
-        this.vendorFiles['jquery.js'] = `${this.bowerDirectory}/jquery/dist/jquery.js`;
-      }
+    if (this.project.findAddonByName('@ember/jquery')) {
+      return;
     }
+    let ember = this.project.findAddonByName('ember-source');
+    let jqueryPath;
+    if (ember) {
+      let optionFeatures = this.project.findAddonByName('@ember/optional-features');
+      if (optionFeatures && !optionFeatures.isFeatureEnabled('jquery-integration')) {
+        return;
+      }
+      this.project.ui.writeDeprecateLine('Ember will stop including jQuery by default in an upcoming version. If you wish keep using jQuery in your application explicitly add `@ember/jquery` to your package.json');
+      jqueryPath = ember.paths.jquery;
+    } else {
+      jqueryPath = `${this.bowerDirectory}/jquery/dist/jquery.js`;
+    }
+    this.vendorFiles = merge({ 'jquery.js': jqueryPath }, this.vendorFiles);
   }
 
   /**
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -387,17 +387,13 @@ class EmberApp {
     let productionEmber;
     let emberTesting;
     let emberShims = null;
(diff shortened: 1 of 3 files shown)


tests/helpers/default-packager.js
```diff
@@ -54,7 +54,7 @@ const DEFAULT_NODE_MODULES = {
       `,
     'package.json': JSON.stringify({
       name: 'ember-source',
-      version: '~3.0.0-beta.1',
+      version: '3.0.0-beta.1',
```
cibernox: I corrected this because the version property on an addon's package.json should not have ~ or ^.

tests/unit/broccoli/ember-app-test.js
```diff
@@ -1088,9 +1088,9 @@ describe('EmberApp', function() {
 
   describe('vendorFiles', function() {
     let defaultVendorFiles = [
-      'jquery.js',
       'ember.js',
       'app-shims.js',
+      'jquery.js',
```
cibernox: Is this order important? I assume it's not.

lib/broccoli/ember-app.js
```diff
@@ -467,6 +464,22 @@ class EmberApp {
     }
   }
 
+  _addJqueryInLegacyEmber() {
+    let emberJQuery = this.project.findAddonByName('@ember/jquery');
+    if (emberJQuery) {
+      this.vendorFiles['jquery.js'] = emberJQuery.paths.jquery;
```
rwjblue: If the addon is found, we shouldn't do anything with `vendorFiles` (we should instead rely on the addon having done a normal `app.import`).

lib/broccoli/ember-app.js
```diff
@@ -467,6 +464,22 @@ class EmberApp {
...
+    } else {
+      let ember = this.project.findAddonByName('ember-source');
+      if (ember) {
+        if (semver.lt(this.project.addonPackages['ember-source'].pkg.version, '3.3.0')) {
```
rwjblue: We should add some additional logic here:

1. if `config/optional-features.json` is present **AND** the (still to be created) `jquery-integration` optional feature flag is set to `true` or is not defined: we should continue to use `ember.paths.jquery` and issue a console warning/deprecation
2. if `config/optional-features.json` is _not_ present we issue a console warning/deprecation suggesting opting out of the feature via `@ember/optional-features` **or** adding the `@ember/jquery` addon

tests/unit/broccoli/ember-app-test.js
rwjblue: The order matters _somewhat_ because `jquery.js` _must_ end up in `dist/assets/vendor.js` **before** Ember itself...

lib/broccoli/ember-app.js
ro0gr: @rwjblue I know it's an intimate API but some addons can [rely](https://github.com/san650/ember-cli-page-object/blob/ea4bd566089c4ee7f3e5c9a1e629279e3b4ea736/index.js#L23) on `this.vendorFiles['jquery.js']` value. Maybe we can keep it as a getter with a deprecation?

rwjblue: Seems fine to me, in that case the addon should check with `@ember/optional-features` first (which would be how they avoid the deprecation).

ro0gr: From the EmberObserver [code search](https://emberobserver.com/code-search?codeQuery=%5B%27jquery.js%27%5D) looks like currently "ember-cli-page-object" is the only public addon which needs to detect "jquery" existance at build time. So it potentially could be solved in this addon space. 

However `this.vendorFiles['jquery.js']` was a thing for ages so it might make a sense to leave it readable for some time..

cibernox:
@rwjblue Updated description in the PR.

cibernox:
Update to ensure jquery is imported before ember.

cibernox:
@rwjblue I've added tests, this should be ready.

PR #7844, opened 2018-05-23
Add @ember/jquery and @ember/optional-features by default.

rwjblue:
* Apps get `@ember/jquery` for backwards compatibility
* Addons do not include `@ember/jquery`
* All projects include `@ember/optional-features` and either opt-in or
  out as appropriate.
* Update to ember-try 1.0.0-beta.3
* Make default scenario run without jQuery
* Add `default-with-jquery` ember-try scenario

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -64,6 +64,19 @@ module.exports = function() {
           npm: {
             devDependencies: {}
           }
+        },
+        {
+          name: 'ember-default-with-jquery',
+          env: {
+            EMBER_OPTIONAL_FEATURES: JSON.stringify({
+              'jquery-integration': true,
+            }),
+          },
+          npm: {
+            devDependencies: {
+              '@ember/jquery': '^0.5.1'
+            }
+          }
         }
       ]
     };
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -69,8 +69,8 @@ module.exports = function() {
           name: 'ember-default-with-jquery',
           env: {
             EMBER_OPTIONAL_FEATURES: JSON.stringify({
-              'jquery-integration': true,
-            }),
+              'jquery-integration': true
+            })
           },
           npm: {
             devDependencies: {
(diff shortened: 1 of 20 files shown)


Turbo87:
The Readme of @ember/optional-features explicitly mentions that it is only meant to be used in apps, not addons :thinking:

Turbo87:
ah, apparently I missed that it would be added to the dummy app, not the addon itself... 🤦‍♂️

rwjblue:
Yeah, it’s only added as a dev dep in the ember-try scenario. The goal there is to help ensure addons can work in both scenarios as there are some minor differences with event dispatching that _mostly_ don’t matter but sometimes might (and it seems better to have it under test as suggested by the original RFC).

cibernox:
@rwjblue this looks good. Shall we merge?

tests/unit/blueprints/addon-test.js
```diff
@@ -77,7 +77,7 @@ describe('blueprint - addon', function() {
   "devDependencies": {\n\
     "ember-disable-prototype-extensions": "^1.1.2",\n\
     "ember-source-channel-url": "^1.0.1",\n\
-    "ember-try": "^0.2.23",\n\
+    "ember-try": "^1.0.0-beta.3",\n\
```
kiwiupover: @rwjblue How far way is ember-try 1.0?

rwjblue: We should plan to have 1.0.0 released before this version (ember-cli@3.3.0) is released...

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -64,6 +64,19 @@ module.exports = function() {
           npm: {
             devDependencies: {}
           }
+        },
+        {
+          name: 'ember-default-with-jquery',
+          env: {
+            EMBER_OPTIONAL_FEATURES: JSON.stringify({
+              'jquery-integration': true,
```
kellyselden: unneeded trailing comma

blueprints/addon/files/addon-config/ember-try.js
```diff
@@ -64,6 +64,19 @@ module.exports = function() {
...
+            }),
```
kellyselden: unneeded trailing comma

rwjblue:
OK, fixed up the linting issues...

PR #7780, opened 2018-04-19
run `eslint .` plus .eslintignore

kellyselden:
Closes #7770

node_modules appears to be auto ignored, so no override necessary.

PR #7849, opened 2018-05-25
continued ignore cleanup and syncing

kellyselden:
Continuing some work missed in #7780


tests/fixtures/module-unification-addon/.npmignore
```diff
@@ -1,21 +1,21 @@
...
-.gitignore
-.watchmanconfig
-.travis.yml
-bower.json
-ember-cli-build.js
-testem.js
+/dist/
+/tests/
+/tmp/
+.gitkeep
```
kellyselden: I tested this and it still does what we want.

Issue #7852, opened 2018-05-26
package hook is getting compiled addon code


kellyselden:
Even with DELAYED_TRANSPILATION on, the package hook is getting AMD addon code. It is because of this line https://github.com/ember-cli/ember-cli/blob/5e124442e7366ecaf93f22973819db12c03056e6/lib/broccoli/ember-app.js#L1164.

This block needs to move to after the package hook is called (for our fallback packager to handle).

cc @twokul

locks:
Should be fixed by referenced PR.

PR #7853, opened 2018-05-27
SAVEPOINT

twokul:
I'd like some 👀 on this before I start changing tests/etc.

fix https://github.com/ember-cli/ember-cli/issues/7852

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1161,14 +1161,10 @@ class EmberApp {
     if (experiments.DELAYED_TRANSPILATION) {
       precompiledSource = this._debugTree(precompiledSource, `precompiledAddonTree:${type}`);
 
-      let compiledSource = this._compileAddonTree(precompiledSource, options.skipTemplates);
-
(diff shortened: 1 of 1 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1621,6 +1617,43 @@ class EmberApp {
   }
 
   _legacyPackager(fullTree) {
+    if (experiments.DELAYED_TRANSPILATION) {
+      let compiledAddonSource;
+
+      let compiledAddons = this._compileAddonTree(new Funnel(this.addonTree(), {
+        srcDir: 'addon-tree-output',
```
kellyselden: We could probably cache the contents of `_addonTree` before the final funnel to avoid this funnel.

lib/broccoli/ember-app.js
```diff
@@ -1161,14 +1161,10 @@ class EmberApp {
     if (experiments.DELAYED_TRANSPILATION) {
       precompiledSource = this._debugTree(precompiledSource, `precompiledAddonTree:${type}`);
```
kellyselden: These debug's should be moved and preserved in the new section I think.

PR #7858, opened 2018-05-29
throw error on `window.find` in tests

kellyselden:
Solves https://github.com/emberjs/ember-test-helpers/pull/369#issuecomment-392975541 with a linter change rather than a code hack.

The upside of it being here instead of the eslint-plugin-ember recommended is it can target tests only.

Sister https://github.com/ember-cli/eslint-plugin-ember/pull/258


Turbo87:
I prefer this solution over including it in the plugin. The problem though is that quite a few people probably still use the old testing APIs where you had to use the global find function. Since that is not deprecated yet it's a bit difficult to introduce the warning for everyone. We would at least need to adjust the warning message to reflect that.

rwjblue:
Closing as I think we are going to have to fix this over in eslint-plugin-ember.

PR #7860, opened 2018-05-30
precompile the app tree (to remove stuff like decorators)

kellyselden:
needs a test

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1006,7 +1006,7 @@ class EmberApp {
    * @method getAppJavascript
    * @return {BroccoliTree}
   */
-  getAppJavascript() {
+  getAppJavascript(isPackageHookSupplied) {
     let appTrees = [].concat(
       this.addonTreesFor('app'),
       this.trees.app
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -760,6 +762,36 @@ class EmberApp {
     return tree;
   }
 
+  _precompileAppJsTree(tree) {
```
twokul: conceptually, we should move _all_ transpilation from `ember-app` and defer that to the packager. can you move this functionality to the packager?

kellyselden: We need to provide the package hook ES current, therefore this needs to be before the packager.

lib/broccoli/ember-app.js
```diff
@@ -974,7 +1006,7 @@ class EmberApp {
    * @method getAppJavascript
    * @return {BroccoliTree}
   */
-  getAppJavascript() {
+  getAppJavascript(isPackageHookSupplied) {
```
kellyselden: @twokul I added the optimization you asked for.

lib/broccoli/ember-app.js
```diff
@@ -985,11 +1017,17 @@ class EmberApp {
...
     });
 
-    return new Funnel(mergedApp, {
+    let appTree = new Funnel(mergedApp, {
       srcDir: '/',
       destDir: this.name,
       annotation: 'ProcessedAppTree',
     });
+
+    if (experiments.DELAYED_TRANSPILATION && isPackageHookSupplied) {
```
twokul: shouldn't we check if packager experiment is enabled or not rather than checking for `package` property?

PR #7861, opened 2018-05-31
Config hook called too soon during serve

thoov:
The config hook gets called eagerly during the serve flow and that
leads to issues with the context of the config hooks not being set
up properly. An example of this is if you have:

```js
config() {
  return this.app && this.app.options[this.name];
}
```

`this.app` will be undefined but that same hook during a build will not be undefined.

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -43,8 +43,6 @@ module.exports = Command.extend({
     return this._checkExpressPort(commandOptions)
       .then(this._autoFindLiveReloadPort.bind(this))
       .then(commandOptions => {
-        let config = this.project.config(commandOptions.environment);
-
         this.ui.writeDeprecateLine(
           'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
           !(!('rootURL' in config) && config.baseURL));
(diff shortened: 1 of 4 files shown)


thoov:
@stefanpenner @xg-wang Lets discuss this tomorrow

lib/commands/serve.js
```diff
@@ -43,8 +43,6 @@ module.exports = Command.extend({
     return this._checkExpressPort(commandOptions)
       .then(this._autoFindLiveReloadPort.bind(this))
       .then(commandOptions => {
-        let config = this.project.config(commandOptions.environment);
-
         this.ui.writeDeprecateLine(
           'Using the `baseURL` setting is deprecated, use `rootURL` instead.',
           !(!('rootURL' in config) && config.baseURL));
```
xg-wang: This should also be moved.

lib/models/watcher.js
```diff
@@ -48,7 +48,9 @@ class Watcher extends CoreObject {
 
     if (this.serving) {
       let options = this.options;
-      let baseURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL || options.baseURL);
+      let config = this.options.project.config();
```
xg-wang: Should be something like `let config = this.project.config(options.environment)`

PR #7863, opened 2018-06-02
Bump minimum Node version for addons.

rwjblue:
Node 4 is completely unsupported now, there is no reason to attempt to run it in CI still. Addons generated by ember-cli@3.2.0 should not try to use Node 4...

Note: targets beta branch...

TODO:

- [x] Fix `engines` support in `package.json` to drop Node 4 (and 7).
- [x] Update ember-cli's own `package.json` for supported Node engines.


Turbo87:
I'm wondering if we should include a note somewhere that people should bump majors when doing this. Especially with ember-cli-update this is an easy to miss change.

rwjblue:
OK, updated with the last TODO items.  Should be good to go (assuming everyone is still 👍 )...

Turbo87:
@rwjblue looks like we also need to adjust the AppVeyor config 😉

rwjblue:
Thank you @Turbo87!

PR #7864, opened 2018-06-03
Refactor packager

xg-wang:
This was missing in https://github.com/ember-cli/ember-cli/commit/1d243e6ef74cd37ab01d2c149c16b68618baa486

Should we consider adding 'testem' to `fingerprint.exclude`  in broccoli-asset-rev default instead of here?


lib/broccoli/ember-app.js
```diff
@@ -152,21 +152,21 @@ class EmberApp {
     this._defaultPackager = new DefaultPackager({
       env: this.env,
       name: this.name,
+      autoRun: this.options.autoRun,
```
twokul: is there a reason to change the order of arguments?

twokul:
`fingerprint.exclude` is something that `ember-app` should specify

lib/broccoli/ember-app.js
xg-wang: I changed the order so it's same to the order in the default-packager https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/default-packager.js#L125

tests/unit/broccoli/default-packager/tests-test.js
```diff
@@ -208,6 +208,8 @@ describe('Default Packager: Tests', function() {
     expect(outputFiles.assets['tests.js']).to.include(`define('the-best-app-ever/config/environment'`);
     expect(outputFiles.assets['tests.js']).to.include(`require('the-best-app-ever/tests/test-helper');`);
     expect(outputFiles.assets['tests.js']).to.include('EmberENV.TESTS_FILE_LOADED = true;');
+
+    expect(defaultPackager.options.fingerprint.exclude).to.include('testem');
```
twokul: it is a little confusing to have this assertion here; the code above clearly passes a fingerprint object with empty `exclude` array.

I think it would be better to add a separate test

```javascript
it('excludes testem files by default', co.wrap(function *() {
  let defaultPackager = new DefaultPackager({
    // ... code ...
    fingerprint: {
      exclude: []
    }
  });

  output = yield buildOutput(defaultPackager.packageTests(input.path()));

  expect(defaultPackager.fingerprint.exclude).to.include('testem');
});
```

But as I mentioned in the comment, I think `exclude` array should be passed in from the `ember-app`. That would mean moving [this code](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/default-packager.js#L1125-L1127) to `ember-app` and just using the argument within the packager itself:

```javascript
// ... code ...
if (this.fingerprint && this.fingerprint.exclude) {
  this.fingerprint.exclude.push('testem');
}

this._defaultPackager = new DefaultPackager({
  // ... code ...
  fingerprint: this.fingerprint,
});
```

twokul:
@xg-wang left one more comment; looks good overall, thanks for the patch!

xg-wang:
I think this is good for a refactor purpose now. But to trigger exclude testem, we need to set the option explicitly:
```js
let app = new EmberApp({
    fingerprint: {
        exclude: []
    },
});
```
This can be default in production test, per the original pr for this change https://github.com/ember-cli/ember-cli/pull/1230

PR #7875, opened 2018-06-13
Fix `ember serve` throwing .replace of undefined

thoov:
This issue was that the options param changed after: https://github.com/ember-cli/ember-cli/pull/7861. Specifically that rootURL and baseURL were taken off of commandOptions. This had a problem where the express-server when passing down the options into addons `serverMiddleware` hook would not contain those properties. An example https://github.com/ember-cli/ember-cli-inject-live-reload/blob/master/index.js#L37


thoov:
@rwjblue @stefanpenner Could you guys review?

PR #7879, opened 2018-06-20
testem: Improve Chrome command line flags

stefanpenner:
Some better testem config defaults for headless chrome.

Context for --disable-dev-shm-usage -> https://bugs.chromium.org/p/chromium/issues/detail?id=736452#c64

cc @step2yeung thoughts?

PR #7880, opened 2018-06-21
testem: Improve Chrome command line flags

stefanpenner:
Some better testem config defaults for headless chrome

Context for --disable-dev-shm-usage -> https://bugs.chromium.org/p/chromium/issues/detail?id=736452#c64

Backport of https://github.com/ember-cli/ember-cli/pull/7879

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -12,9 +12,11 @@ module.exports = {
       ci: [
         // --no-sandbox is needed when running Chrome inside a container
         process.env.CI ? '--no-sandbox' : null,
-
-        '--disable-gpu',
         '--headless',
+        '--disable-gpu',
+        '--disable-dev-shm-usage',
+        '--disable-software-rasterizer',
+        '--mute-audio',
         '--remote-debugging-port=0',
         '--window-size=1440,900'
       ].filter(Boolean)


blueprints/app/files/testem.js
```diff
@@ -11,10 +11,12 @@ module.exports = {
     Chrome: {
       ci: [
         // --no-sandbox is needed when running Chrome inside a container
-        process.env.CI ? '--no-sandbox' : null,
-
-        '--disable-gpu',
+        process.env.TRAVIS ? '--no-sandbox' : null,
```
Turbo87: was the change from `CI` to `TRAVIS` intended?

kellyselden: good catch, we want CI here

stefanpenner:
@Turbo87 i appreciate you scrutiny, let me share my motivations, maybe they are sufficient.

> Is this fixing a bug?

It's part of the solution for some users re: https://github.com/testem/testem/issues/1021

Specifically some hard to diagnose browser crashes are the result of a insufficiently sized `/dev/shm`, newer linux kernels allow Chrome to simply create ad-hoc in-memory descriptors (no need to rely on a memory mount such as `/dev/shm`). Chrome recently added support for these. But unfortunately many popular distributions (RHEL) don't have this yet, and many CI environments as stuck on old versions.

Some context: https://github.com/GoogleChrome/puppeteer/blob/v1.0.0/docs/troubleshooting.md#tips

Although the flag may have some downsides (in theory slightly slower IPC for shared blobs), the lack of mysterious browser crashes due to what appears to be `/dev/shm/` space limits is a life saver.

I would hate to have new users generating/upgrading a known to be bad and trolly testem config.

PR #7884, opened 2018-06-22
Fix for `yarn` globally unavailable, breaks `npm test`

### LINKED ISSUE #7484: Without `yarn` globally available, `npm test` fails
The test `'ember new uses yarn when blueprint has yarn.lock'` (in `tests/acceptance/new-test.js`) does not pass if a global `yarn` is not available.  

I identified this while running a `npm test` on a fresh machine without `yarn`.  In this particular case, I think that we should probably not fail CI (or if we do, it should be an earlier assertion that properly messages that yarn is required).

abhilashlr:
Fixes #7484

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -20,6 +20,8 @@ const forEach = require('ember-cli-lodash-subset').forEach;
 const assertVersionLock = require('../helpers/assert-version-lock');
 const experiments = require('../../lib/experiments');
 
+const Command = require('../../tests/helpers/command-generator');
+
 let tmpDir = './tmp/new-test';
 
 describe('Acceptance: ember new', function() {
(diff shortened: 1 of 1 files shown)


abhilashlr:
@rwjblue @Turbo87 @kellyselden Please review this and let me know if this is the same as what was stated in the issue comment.

abhilashlr:
Any updates on this? @kellyselden @rwjblue

tests/acceptance/new-test.js
```diff
@@ -285,6 +287,14 @@ describe('Acceptance: ember new', function() {
   }));
 
   it('ember new uses yarn when blueprint has yarn.lock', co.wrap(function *() {
+    try {
+      let yarn = new Command('yarn');
```
Turbo87: running `yarn` might be problematic, as it would actually run and install stuff if it actually is available. we might be able to use something like https://github.com/npm/node-which to figure out if `yarn` exists or not 🤔

tests/acceptance/new-test.js
```diff
@@ -285,6 +287,14 @@ describe('Acceptance: ember new', function() {
...
+      yarn.invoke();
+    } catch (err) {
+      console.log('No yarn present, skipping test');
+      return;
```
Turbo87: instead of returning from the function we should properly mark it as "skipped". if I remember correctly there are two ways of doing it:
a) calling `this.skip()` in the test code
b) using `it.skip('foo', function() {...})` instead of `it(...)`

abhilashlr:
@Turbo87 I closed this PR since it was having huge differences since it was left as is. I shall create a new PR against master with your suggestions, however, I face an issue with eslint on `var which = require('which')` where, it throws no-extraneous imports. I am assuming it has to be added as part of `package.json` ('which': '1.3.1')?

Turbo87:
> I am assuming it has to be added as part of `package.json` ('which': '1.3.1')?

probably, yeah

Issue #7699, opened 2018-03-23
Prevent generation of .js.js files


MrChriZ:
Small feature request.

If I type 
ember g controller post.js

Ember creates a post.js.js and a test post.js.js

Would it perhaps be useful instead to either strip the .js or display an error message or something similar?

Turbo87:
sounds like a reasonable suggestion. do you have time to work on a pull request for this?

abhilashlr:
@Turbo87 I have a solution for this, but I'm not sure if that is the perfect one thoug. Will share a PR, please let me know if that is fine.

romgere:
@Turbo87  is this have to be done in ember-source blueprint ? And it's concern only controller or others  blueprint (route, helper, component, ...) ?

Turbo87:
@romgere yes, I think doing it in the blueprints in the `ember-source` repo is the best way to do it. the issue only talks about controllers, but doing the same for routes, components, etc. is a good idea. probably also for templates and `.hbs` 🤔

romgere:
OK I'am going to try to do it...
Just one more question about pod. For commands  with "--prod" option like `ember g component foo.js --pod` what's the expected behavior :thinking:  ? `app/components/foo/component.js` or `app/components/foo.js/component.js` file ?

Turbo87:
> For commands with "--prod" option like `ember g component foo.js --pod` what's the expected behavior 🤔 ? 

`app/components/foo/component.js` AFAIK

Gaurav0:
Is this fixed?

romgere:
@Gaurav0 
![image](https://user-images.githubusercontent.com/13900970/66254133-867ae600-e772-11e9-828d-7abce9455dc1.png)
I think yes. This issue can be closed now.

PR #7885, opened 2018-06-23
Prevent generation of .js.js files

abhilashlr:
Fix for #7699


abhilashlr:
@kellyselden @Turbo87

Turbo87:
hmm, I'm not sure we should do this in the `generate` command itself, since there could be blueprints where a `.js` suffix is useful. I guess this should be part of the individual blueprints instead 🤔

PR #7891, opened 2018-06-29
Pass resultAnnotation object to build function

lifeart:
Pass resultAnnotation object to build function to allow custom caching strategies.
example: https://github.com/ember-cli/broccoli-builder/pull/30
This trick produce 4x pefrormance improvements on windows mashines

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -188,7 +188,7 @@ class Builder extends CoreObject {
 
       if (this.broccoliBuilderFallback) {
         try {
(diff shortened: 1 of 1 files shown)


locks:
Rebased and updated the changed to what I believe is the current equivalent.

locks:
I only just noticed that this change seems to be incorporated in another PR https://github.com/ember-cli/ember-cli/pull/8551/files#diff-53c8b5d6578c6f2fb2869c4d31bdc3a1R187.
Closing this so we can focus on the other one!

PR #7893, opened 2018-06-29
Fix `public` trees merge bug

twokul:
When we gather "public" trees and merge them w/o an "overwrite" option,
the build might fail because both application `public` folder and an
add-on's `public` folder have the same structure (or if multiple add-ons
have the same `public` folder structure.)

To ensure it doesn't happen and to restore previous behaviour, adding
`overwrite` option.

cc @kellyselden @kiwiupover


kellyselden:
is this testable?

twokul:
@kellyselden sure is -- our current setup of testing is less than ideal, trying to see if we can get a small, easy improvement in how we test (like a builder object)

Issue #7862, opened 2018-06-02
[MU] Funnel in Styles (no more app.trees.styles)


gossi:
I've been experimenting around with MU but might also be in combination with the new packager. So tried to port over some things from my classic app I want to use with MU. I wanted to continue using bootstrap scss in my app (without using `ember-bootstrap` addon). In my classic app I would do:

```
const bootstrapTree = new Funnel('node_modules/bootstrap/scss', {
	destDir: 'bootstrap'
});
app.trees.styles = mergeTrees([app.trees.styles, bootstrapTree]);
```

Now, that stopped working with MU, since there is no more `app.trees.styles`. Instead I had to funnel it into `app.trees.src`. Also `app.tree.src = mergeTrees(...);`  didn't work either, because I got this error:

```
EslintValidationFilter can only handle one:* broccoli nodes, but part of the given input pipeline is a many:* node. (broccoli-merge-trees is an example of a many:* node) Please perform many:* operations after linting.
```

I ended up overwriting `getSrc()` and funnel in bootstrap scss, this is how it looks:

```
app.getSrc = function () {
	let rawSrcTree = this.trees.src;

	if (!rawSrcTree) {
		return null;
	}

	const bootstrapTree = new Funnel('node_modules/bootstrap/scss', {
		destDir: 'src/ui/styles/bootstrap',
		annotation: 'Bootstrap'
	});

	const src = new Funnel(rawSrcTree, {
		destDir: 'src',
	});

	return mergeTrees([src, bootstrapTree], {
		overwrite: true
	});
};
```

I don't like that solution for various reasons:

1) If something upstream `getSrc()` changes, I don't get it (though can get the original src tree from `super`.
2) It is unhandy and error prone.

I don't know what the new Packager will bring and if there is a solution to this already planned. Especially if there is a `package` hook already in place. I'm pretty sure, people will come here and claim their `app.trees.styles` stopped working and are looking for a solution to this.

PS. I'm on ember-cli canary.

twokul:
@gossi sorry for taking so long to respond!

I've got couple of questions:

+ why don't you want to use `ember-bootstrap` add-on? totally fine not to use it, just curious
+ in MU world, `app.trees.src` contains all the styles now and you should be able to do 

```javascript
app.trees.src = mergeTrees[[app.trees.src, bootstrapTree], {
  annotation: 'Adding bootstrap',
  overwrite: true,
})
```

gossi:
[...]

PR #7895, opened 2018-07-03
Explicitly include `this.trees.styles`

twokul:
After packager work landed, there was no need to explicitly merge
`app.trees.styles` into a "full tree" as for classic model they would
be a part of "app" tree, for MU they would be a part of "src" tree.

As it turned out some developers relied on the fact that they could add
things to `app.trees.styles` so that packager changes turned out to be
breaking changes.

Since this behaviour manifested only when using MU *and* adding styles
to `app.trees.styles`, this is not critical.

Fix https://github.com/ember-cli/ember-cli/issues/7862

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1608,6 +1608,7 @@ class EmberApp {
     return [
       this.getAppJavascript(this._isPackageHookSupplied),
       this.getAddonTemplates(),
+      this.trees.styles,
       this.getAddonStyles(),
       this.getTests(),
       this.getExternalTree(),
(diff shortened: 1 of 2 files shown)


kellyselden:
Is there any way we can deprecate this sanely?

twokul:
@kellyselden According to @rwjblue `app.trees` is public API. The usage of them is not documented anywhere, but people [do rely on them](https://emberobserver.com/code-search?codeQuery=app.trees&sort=score&sortAscending=false): figuring out "proper" `projectRoot` or accessing file tree depending if the code runs within the app or add-on.

`package` hook is a step in the right direction of opening our build pipeline more, but ultimately I don't think we can deprecate `app.trees` now

locks:
Closing due to inactivity.

PR #7904, opened 2018-07-08
further unify all our *ignore files

kellyselden:
I removed code from tests/fixtures/module-unification-addon. No sense keeping in sync when they are untested. #7903


blueprints/addon/files/npmignore
```diff
@@ -1,23 +1,28 @@
...
+/yarn.lock
+.gitkeep
 
 # ember-try
-.node_modules.ember-try/
-bower.json.ember-try
-package.json.ember-try
+/.node_modules.ember-try/
+/bower.json.ember-try/
+/package.json.ember-try/
```
Turbo87: those two are not folders so they should not end with `/`

blueprints/app/files/.eslintignore
```diff
@@ -14,3 +14,5 @@
 
 # ember-try
 /.node_modules.ember-try/
+/bower.json.ember-try/
+/package.json.ember-try/
```
Turbo87: see above

blueprints/addon/files/npmignore
kellyselden: Good catch

Issue #7486, opened 2017-12-05
move ember-cli-build.js to inside dummy folder for addons


kellyselden:
Tracking this because I can't immediately jump on this, and would be great for new contributors!

Addons have two build files in their root dir, "index.js" and "ember-cli-build.js". "index.js" is run when an addon is built inside a host app, and "ember-cli.build.js" is the addon's "dummy" app's build file that is only used when running the addon's tests. It is for this reason, I think we should move the file to "tests/dummy/ember-cli-build.js" in the blueprints, and also make ember-cli recognize it at the new location.

villander:
@kellyselden I changed the `ember-cli-build` layout but the` ember addon <your-addon> `command continues to generate the same structure. Where is the recognize of this?

I'm very excited for my first contribution to ember-cli

villander:
@kellyselden I was able to move `ember-cli-build.js` to the` tests / dummy` pass, and I changed the recognize (setupBroccoliBuilder), but I get the error `ember-cli-build.js found.` and even putting a console.log in this function I could not debug, it is never fired.

Can you help me with that?

villander:
`findBuildFile` uses https://github.com/sindresorhus/find-up

So I do not need to change anything in the `builder`. But it still does not find `ember-cli-build.js` inside the` test / dummy` folder. I looked for the cause, but it did not make much sense to me.

villander:
@Turbo87 can you give me a hint how to solve this please?

Turbo87:
I don't know, sorry

rwjblue:
You need to update the `findBuildFile` utilty to leverage the nested path when the project is an addon.

The `findBuildFile` utiltity is used here: https://github.com/ember-cli/ember-cli/blob/87337ee328b4db84e087b8166d698d31ead24b02/lib/models/builder.js#L49

astronomersiva:
I gave a [shot](https://github.com/ember-cli/ember-cli/pull/7907/) at this. Please take a look.

locks:
Closing as the PR was also closed. See the discussion in https://github.com/ember-cli/ember-cli/pull/7907 for more details. Thank you @astronomersiva for your help!

kellyselden:
I don't know why that PR was closed, but I still think this is the right thing to do.

rwjblue:
Ya, I generally agree. We should move at least move `ember-cli-build.js` and `config/ember-try.js` into the dummy app...

Turbo87:
[...]

PR #7907, opened 2018-07-09
Move ember-cli-build to tests/dummy for addons

astronomersiva:
Addresses #7486

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -99,6 +99,10 @@ abbrev@1:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
 
+abbrev@1.0.x:
+  version "1.0.9"
+  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.0.9.tgz#91b4792588a7738c25f35dd6f63752a2f8776135"
+
 accepts@~1.3.4, accepts@~1.3.5:
   version "1.3.5"
   resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.3.5.tgz#eb777df6011723a3b14e8a72c0805c8e86746bd2"
@@ -337,7 +341,7 @@ async-promise-queue@^1.0.3:
     async "^2.4.1"
     debug "^2.6.8"
 
-async@^1.4.0, async@^1.5.2:
+async@1.x, async@^1.4.0, async@^1.5.2:
   version "1.5.2"
   resolved "https://registry.yarnpkg.com/async/-/async-1.5.2.tgz#ec6a61ae56480c0c3cb241c95618e20892f9672a"
 
@@ -1910,6 +1914,17 @@ escape-string-regexp@1.0.5, escape-string-regexp@^1.0.2, escape-string-regexp@^1
   version "1.0.5"
   resolved "https://registry.yarnpkg.com/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz#1b61c0562190a8dff6ae3bb2cf0200ca130b86d4"
 
+escodegen@1.8.x:
+  version "1.8.1"
+  resolved "https://registry.yarnpkg.com/escodegen/-/escodegen-1.8.1.tgz#5a5b53af4693110bebb0867aa3430dd3b70a1018"
+  dependencies:
+    esprima "^2.7.1"
+    estraverse "^1.9.1"
+    esutils "^2.0.2"
+    optionator "^0.8.1"
+  optionalDependencies:
+    source-map "~0.2.0"
+
 eslint-plugin-chai-expect@^1.1.1:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/eslint-plugin-chai-expect/-/eslint-plugin-chai-expect-1.1.1.tgz#cd640b8b38cf6c3abcc378673b7b173b99ddc70a"
@@ -2003,6 +2018,10 @@ espree@^3.5.4:
     acorn "^5.5.0"
     acorn-jsx "^3.0.0"
 
+esprima@2.7.x, esprima@^2.7.1:
+  version "2.7.3"
+  resolved "https://registry.yarnpkg.com/esprima/-/esprima-2.7.3.tgz#96e3b70d5779f6ad49cd032673d1c312767ba581"
+
 esprima@^4.0.0:
   version "4.0.1"
   resolved "https://registry.yarnpkg.com/esprima/-/esprima-4.0.1.tgz#13b04cdb3e6c5d19df91ab6987a8695619b0aa71"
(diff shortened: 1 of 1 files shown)


astronomersiva:
cc: @kellyselden

kellyselden:
We must add some tests for this. We have to make sure the old location still works (and the new location is in fact being used). But this great work.

astronomersiva:
Sure, I will try to get some tests written for this.

I also wanted your opinion on this - https://github.com/ember-cli/ember-cli/pull/7907/files#diff-d5db1dafa8d20a2880b7ce82881645d1R22.

I believe this `if` check is no longer needed as `ember-cli-build.js` is going to be present at the project root for both addons and apps. Will returning `projectRoot` be sufficient here?

astronomersiva:
Sure, I will try to get some tests written for this. I also wanted your opinion on [this](https://github.com/ember-cli/ember-cli/pull/7907/files#diff-d5db1dafa8d20a2880b7ce82881645d1R22).

I believe this if check is no longer needed as ember-cli-build.js is going to be present at the project root for both addons and apps. Will returning projectRoot be sufficient here?

lib/utilities/find-build-file.js
```diff
@@ -3,15 +3,28 @@
...
+    options.cwd = path.resolve(process.cwd(), `${projectRoot}/tests/dummy`);
+  }
+
+  let buildFilePath = findUp.sync(file, options);
   if (!buildFilePath) {
     return null;
   }
 
-  process.chdir(path.dirname(buildFilePath));
+  if (isEmberCLIAddon) {
```
astronomersiva: Can we just return the projectRoot here?

astronomersiva:
@kellyselden I gave a try at writing some tests...please take a look.

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -99,6 +100,68 @@ if (!experiments.MODULE_UNIFICATION) {
...
 
+    it('works when ember-cli-build is inside tests/dummy', co.wrap(function *() {
+      yield copyFixtureFiles('addon/ember-cli-build-at-tests');
+
+      let packageJsonPath = path.join(addonRoot, 'package.json');
+      let packageJson = fs.readJsonSync(packageJsonPath);
+
+      packageJson.dependencies = packageJson.dependencies || {};
+      // add HTMLBars for templates (generators do this automatically when components/templates are added)
+      packageJson.dependencies['ember-cli-htmlbars'] = 'latest';
```
kellyselden: Can't this be put into the fixtures?

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -99,6 +100,68 @@ if (!experiments.MODULE_UNIFICATION) {
...
+
+      fs.writeJsonSync(packageJsonPath, packageJson);
+
+      let result = yield runCommand('node_modules/ember-cli/bin/ember', 'build');
+
+      expect(result.code).to.eql(0);
+
+      let cssPath = path.join(addonRoot, 'addon', 'styles', 'app.css');
```
kellyselden: Why check css for an ember-cli-build test?

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -99,6 +100,68 @@ if (!experiments.MODULE_UNIFICATION) {
...
+      let builtCss = path.join(addonRoot, 'dist', 'assets', `vendor-${hex}.css`);
+      let builtCssExists = fs.existsSync(builtCss);
+
+      expect(builtCssExists).to.eql(true);
+    }));
+
+    it('works when ember-cli-build is at addon root', co.wrap(function *() {
+      yield copyFixtureFiles('addon/ember-cli-build-at-root');
+
+      // remove ember-cli-build.js created at addon root
```
kellyselden: Comment seems wrong

tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -99,6 +100,68 @@ if (!experiments.MODULE_UNIFICATION) {
...
+      let dummyPath = path.join(addonRoot, 'tests', 'dummy', 'ember-cli-build.js');
+      if (fs.existsSync(dummyPath)) {
+        fs.unlinkSync(dummyPath);
```
kellyselden: Why would this file be here? Doesn't this mean the fixtures are wrong?

kellyselden:
you also have a change to an experiments file that probably was an accident

tests/acceptance/addon-smoke-test-slow.js
astronomersiva: Tried it initially but started getting `node_modules is empty` errors. Since this is all handled by the acceptance test helpers, I looked around and found this in a few other places.

tests/acceptance/addon-smoke-test-slow.js
astronomersiva: I wanted to ensure that the options of ember-cli-build are being respected(in this case, fingerprinting) when the file is placed at both locations. No specific reason as to why I chose app.css and not some other file.

tests/acceptance/addon-smoke-test-slow.js
astronomersiva: This is because of us creating test targets in the [before hook](https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/addon-smoke-test-slow.js#L32). This results in an `ember-cli-build.js` being present in `tests/dummy` because of my changes and since we run find-up from that directory for an addon, that file's contents are taken for this test which isn't desirable.

astronomersiva:
@kellyselden I have made some changes and left a few comments..Please take a look when you are free.

astronomersiva:
A gentle reminder @kellyselden  😃

astronomersiva:
@kellyselden , I would really appreciate if you take a look at this 😄

blueprints/addon/index.js
```diff
@@ -142,8 +142,16 @@ module.exports = {
     let appFiles = this.lookupBlueprint(this.appBlueprintName).files();
     let addonFilesPath = this.filesPath(this.options);
     let addonFiles = walkSync(addonFilesPath);
+    let files = appFiles.concat(addonFiles);
 
-    return uniq(appFiles.concat(addonFiles));
+    // There will be two ember-cli-build.js present when app and addon
+    // files are concatenated - one at tests/dummy and project root.
+    let i = files.indexOf('ember-cli-build.js');
```
rwjblue: I think I would tweak the assignment above:

```js
let files = [].concat(
  appFiles.filter(f => f !== 'ember-cli-build.js'),
  addonFiles,
);
```

Instead of splicing manually...

lib/utilities/find-build-file.js
```diff
@@ -2,15 +2,23 @@
...
+module.exports = function(file, args = {}) {
+  let { isEmberCLIAddon, projectRoot } = args;
+  let options = { cwd: projectRoot };
 
-  // Note: In the future this should throw
+  if (isEmberCLIAddon) {
+    // for addons, the ember-cli-build.js file will be at tests/dummy.
+    // find-up will only find files upwards in the directory.
+    // To solve this, resolve the path for ember-cli for Ember Addons.
+    options.cwd = path.resolve(process.cwd(), `${projectRoot}/tests/dummy`);
```
rwjblue: I don't like this hard-coding, it makes (for example) having multiple dummy apps impossible...

knownasilya:
Why was this closed?

knownasilya:
Nvm, there is no code any more.

Issue #7896, opened 2018-07-05
Define ENV variables in .ember-cli


gossi:
It can become tedious to prefix every command with the correct environment variables over and over again. At the moment for example if you have an experiment running, you have to prefix the ember command with that experiment name, e.g. `EMBER_CLI_MODULE_UNIFICATION=true ember serve`.

As those ENV variables depend on their project, it makes most sense to define them once for that project and use ember commands as usual. `.ember-cli` looks like a good place for me to define them and could look like this:

```json
{
  "disableAnalytics": false,
  "EMBER_CLI_MODULE_UNIFICATION": true
};
```

or be more explicit:

```json
{
  "disableAnalytics": false,
  "ENV": {
    "EMBER_CLI_MODULE_UNIFICATION": true
  }
};
```

Alternatively read in a `.env` file(s) where variables can be defined. There is kind of a standard here I guess? Which is used by CI/CD services. I haven't run enough research on this one to make a clear statement other than "more research needed".

astronomersiva:
+1 to this. I have been running a module unification project myself and this often gets in the way.

I feel `.ember-cli` is the right place for this as `.env` files are generally ignored in source control. There can be an `experiments` field in the `.ember-cli` file that we can make use of [here](https://github.com/ember-cli/ember-cli/blob/master/lib/experiments/index.js).

I am willing to work on this if this seems reasonable.

cc: @rwjblue

stefanpenner:
> I feel .ember-cli is the right place for this as .env files are generally ignored in source control. There can be an experiments field in the .ember-cli file that we can make use of here.

Sounds reasonable to me.

gossi:
Solution is in #8157

locks:
Fix has been merged!

btecu:
It looks like it got reverted: https://github.com/ember-cli/ember-cli/pull/8008.

locks:
Hm, it looks like it was. Thanks for the link! I think I'll keep the issue closed for a couple of reasons. One is that module unification has been dropped for a long time, the other is that usage of dotenv is widespread and that seems like a better solution than putting more functionality into `.ember-cli`.

PR #7910, opened 2018-07-11
Allow specifying environment variables in .ember-cli

astronomersiva:
Addresses #7896.

Makes it possible to specify environment variables in `.ember-cli` like

```javascript
     {
          "environment": {
               "EMBER_CLI_MODULE_UNIFICATION": true
          }
     }
```

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -77,6 +77,14 @@ module.exports = function(options) {
   const Project = require('../models/project');
   let config = getConfig(options.Yam);
 
+  let envVariables = config.get('environment') || {};
+  for (let key in envVariables) {
+    // env variables in process.env should have more priority over the ones specified in .ember-cli.
+    if (!(key in process.env)) {
+      process.env[key] = envVariables[key];
+    }
+  }
+
   configureLogger(process.env);
 
   // TODO: one UI (lib/models/project.js also has one for now...)
--- a/tests/unit/settings-file/env-variables-test.js
+++ b/tests/unit/settings-file/env-variables-test.js
@@ -1,0 +1,54 @@
+'use strict';
+
+const co = require('co');
+const expect = require('chai').expect;
+const MockUI = require('console-ui/mock');
+const Yam = require('yam');
+const cliEntry = require('../../../lib/cli');
+const broccoliTestHelper = require('broccoli-test-helper');
+
+const buildOutput = broccoliTestHelper.buildOutput;
+const createTempDir = broccoliTestHelper.createTempDir;
+
+describe('.ember-cli environment options', function() {
+  let sampleApp;
+
+  beforeEach(co.wrap(function *() {
+    process.env.SHOULDNT_BE_THERE = 'Normal';
+
+    sampleApp = yield createTempDir();
+    sampleApp.write({
+      '.ember-cli': JSON.stringify({
+        environment: {
+          'MY_ENV_VARIABLE': 'Yayy',
+          'SHOULDNT_BE_THERE': 'Strange',
+        },
+      }),
+    });
+
+    let primaryPath = sampleApp.path();
+
+    yield buildOutput(primaryPath);
+
+    let mockedYam = new Yam('ember-cli', {
+      primary: primaryPath,
+    });
+
+    cliEntry({
+      UI: MockUI,
+      Yam: mockedYam,
+    });
+  }));
+
+  afterEach(co.wrap(function *() {
+    delete process.env.MY_ENV_VARIABLE;
+    delete process.env.SHOULDNT_BE_THERE;
+
+    yield sampleApp.dispose();
+  }));
+
+  it('should assign environment variables in .ember-cli to process.env', function() {
+    expect(process.env.MY_ENV_VARIABLE).to.equal('Yayy');
+    expect(process.env.SHOULDNT_BE_THERE).to.equal('Normal');
+  });
+});
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -17,7 +17,6 @@ const heimdall = require('heimdalljs');
 const PackageInfoCache = require('../models/package-info-cache');
 
 const instantiateAddons = require('../models/instantiate-addons');
-const experiments = require('../experiments');
 
 let processCwd = process.cwd();
 
@@ -216,6 +215,7 @@ class Project {
      @return {Boolean} Whether this is using a module unification format.
     */
   isModuleUnification() {
+    const experiments = require('../experiments');
     if (experiments.MODULE_UNIFICATION) {
       return this.has('src');
     } else {


lib/experiments/index.js
```diff
@@ -1,5 +1,16 @@
...
+const path = require('path');
+const findup = require('find-up');
+const Yam = require('yam');
+
+const cliPath = findup.sync('.ember-cli');
+const emberCli = new Yam('ember-cli', {
+  primary: path.dirname(cliPath),
+});
+const experimentsInDotCli = emberCli.get('experiments') || {};
+Object.assign(process.env, experimentsInDotCli);
```
rwjblue: Hmm, I'd prefer to do this where we do the rest of the yam stuff ([around here](https://github.com/ember-cli/ember-cli/blob/0ef3f1b9ce8450ccb87dc61cd87547a4ce280ad7/lib/cli/index.js#L78)). I also think that if we are "just" merging with `process.env` we shouldn't use `experiments` as the key in `.ember-cli` (we would use `environment` or something else)...

astronomersiva: That was my original intention as well but there was an issue with that.

We require models/project [here](https://github.com/astronomersiva/ember-cli/blob/master/lib/cli/index.js#L77) which requires lib/experiments internally. Due to this, the experiments are set even before the `.ember-cli` is loaded.

Placing the Project require below the getConfig also won't help as getConfig again uses lib/experiments internally.

I agree that the key name has to be changed to `environment` in `.ember-cli`.

astronomersiva:
@rwjblue , I have made the suggested changes. Please take a look :)

twokul:
might be a silly idea but why not teach `yam` to understand `process.env` as a fallback? that way all configuration is still handled in one place and we don't need to make changes to cli itself?

astronomersiva:
@twokul, I am not very convinced about having process.env as a fallback for the following reasons:

* I might have certain env variables in `.ember-cli`'s environment property. It will not be possible to temporarily and quickly try out some other value for one of those env variables by just doing `SOME_VAR=SOMETHING_ELSE ember s` as `.ember-cli` will be the primary source.
* env variables will be present in `process.env` and `.ember-cli`'s `environment` field. We get the config from `.ember-cli` itself in [get-config](https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/get-config.js#L7). We will have to make Yam use the environment field in `.ember-cli` as the primary source and `process.env` itself as the secondary source.

Apologies if I have misunderstood your comment :)

twokul:
@astronomersiva 

> We will have to make Yam use the environment field in .ember-cli as the primary source and process.env itself as the secondary source.

is there something wrong with that? 😄

astronomersiva:
@twokul , maybe I am not understanding you here properly but the whole point of this PR was to enable people to put their ENV variables in `.ember-cli`. Using Yam to have `process.env` as a secondary source for `.ember-cli` is not going to help with that.

> is there something wrong with that?

Consider an env variable `MY_ENV_VARIABLE` present in both `process.env` and inside the `environment` field in `.ember-cli`. AFAIK, you can use Yam to read a file only. So, even if you have process.env as the secondary source, the config object provided by Yam will be like this:

```javascript
let config = new Yam('ember-cli', {
    primary: project_root
    secondary: process.env // i believe this is what you are talking about
});

// config will be 
{
    MY_ENV_VARIABLE: 'some value'         // taken from process.env
    environment: {
        MY_ENV_VARIABLE: 'some_value'   // taken from .ember-cli
    }
}
```

We will now have to get the values present in the environment field, iterate over them and verify they are not present in `process.env` directly to ensure the behaviour I have written about in the first part of my previous comment. 

One solution I see for this is to make Yam automatically treat values inside the `environment` field as special and assign those values to `process.env` after doing the same checks that I have done [here](https://github.com/ember-cli/ember-cli/pull/7910/files#diff-3386b9e4d97ad97b803f4678beb66c5aR86).

astronomersiva:
@rwjblue would really appreciate if you take another look at this.

lib/cli/index.js
```diff
@@ -77,6 +77,17 @@ module.exports = function(options) {
   const Project = require('../models/project');
   let config = getConfig(options.Yam);
 
+  let envVariables = config.get('environment') || {};
+  for (let key in envVariables) {
+    let value = envVariables[key];
+    // * env variables passed directly should have more priority over .ember-cli.
+    // * process.env will have only string values. Therefore, check if the value
+    //   in .ember-cli is truthy and then assign to process.env.
+    if (!process.env.hasOwnProperty(key) && value) {
```
rwjblue: I think I would do `if (!(key in process.env)) {}`, also why does `value` need to be truthy? I'm concerned about `null` and `0` as values, where a normal `process.env.something = null` would coerce to `'null'`...

tests/unit/settings-file/env-variables-test.js
```diff
@@ -0,0 +1,52 @@
...
+  let sampleApp;
+
+  before(co.wrap(function *() {
+    sampleApp = yield createTempDir();
+
+    sampleApp.write({
+      '.ember-cli': JSON.stringify({
+        environment: {
+          "MY_ENV_VARIABLE": "Yayy",
+          "SHOULDNT_BE_THERE": false,
```
rwjblue: Not sure about this one (see question above)

tests/unit/settings-file/env-variables-test.js
```diff
@@ -0,0 +1,52 @@
...
+      primary: primaryPath,
+    });
+
+    cliEntry({
+      UI: MockUI,
+      Yam: mockedYam,
+    });
+  }));
+
+  after(co.wrap(function *() {
```
rwjblue: I think you want this to be `afterEach`

tests/unit/settings-file/env-variables-test.js
```diff
@@ -0,0 +1,52 @@
...
+const cliEntry = require('../../../lib/cli');
+const broccoliTestHelper = require('broccoli-test-helper');
+
+const buildOutput = broccoliTestHelper.buildOutput;
+const createTempDir = broccoliTestHelper.createTempDir;
+
+describe('.ember-cli environment options', function() {
+  let sampleApp;
+
+  before(co.wrap(function *() {
```
rwjblue: Seems like it should be `beforeEach`

lib/cli/index.js
astronomersiva: That makes sense, thanks 👍

astronomersiva:
@rwjblue , thanks for your time 😄 I have made the suggested changes.

knownasilya:
👏

rwjblue:
Thank you @astronomersiva!

kellyselden:
Seems to work well here https://github.com/kellyselden/ember-cli-rollup-packager/pull/30. Nice work!

astronomersiva:
@kellyselden , thanks :) I also noticed that it is possible to have the `.ember-cli` file as a JS file. Might help if you want to do some checks within the file or maybe even get rid of all the double quotes for the keys.

gossi:
Thank you so much for this ❤️

nightire:
@astronomersiva there is a problem, if you import `config/environment` in your app, `config.environment` will be replaced as `"environment"` in `.ember-cli`, but it used to be `process.env.EMBER_ENV`. It is important sometimes to determine the environment variable in applications.

before this change:

image

after this change:

image

tested this feature in ember-cli v3.3

ef4:
@nightire You're right and that is very surprising to me. I filed #7980 for that and an additional similar case.

lougreenwood:
What is the recommended way to access these environment vars in Ember? Looks like I could use the `environment` mentioned above, but I get the impression that is a bug and will likely be reverted?

lougreenwood:
I'm not sure I follow - are you saying that any values set in the new `environment` object will be merged into and made available in `process.env` in for usual consumption in `config/environment.js`?

ef4:
Yes. That (and only that) was the intended behavior of this feature.

lougreenwood:
Nevermind, seems it was an obvious solution for my use case where `process` doesn't exist for the final compiled app (e.g in your wonderful Prember add-on :wink:):

`const EmberCliConfig = require('../.ember-cli');`

If anyone's interested in my use case (or maybe Google is...) I'm trying to define config env vars for my app which has access to process.env for local dev (`ember serve`), but falls back to some string config value (injected by CI build), for deployment to a server with pre-rendered static HTML (no node for process.env).

This new feature which allows me to define them in `.ember-cli.js` allows me to:
* move all definitions and string replacement targets into one file (was previously replacing in every pre-rendered *.html)
* detect and set use fallback when in CI

My config definitions are now much cleaner and more robust as a result and also improves my Cordova build pipeline :)

PR #7912, opened 2018-07-11
Allowing cwd to be passed to testem without being overridden by ember-cli

arthirm:
Back-porting https://github.com/ember-cli/ember-cli/pull/7758 to ember-cli 3.1.4


rwjblue:
I'm not sure we should backport this to an older (and unsupported release). Does 3.2 already include the fix? Is there any reason not to advise folks to upgrade?

arthirm:
As discussed in slack I will close this PR since 3.2 version has this fix already.

PR #7914, opened 2018-07-12
updated count expected in package-info-cache test 

dcombslinkedin:
Because the test actually tries to look at files expected in ember-cli's package dependencies, the count expected changes from version to version of ember-cli. This is fixed for 3.2.0.


dcombslinkedin:
removing as it appears the right value was already present (though failed in my local version).

dcombslinkedin:
Weird! It's basically counting how many entries there are in the node_modules directory. I assumed that should be static, but apparently not for some weird reason. I'll investigate more, and may change the test to something a little less volatile that should pass consistently.

PR #7916, opened 2018-07-12
Support node-http-proxy timeout options

jboler:
My app has some file upload requests that take longer than the default proxy timeout of 2 minutes. It would be useful to be able to test long running requests in development which is possible if I can specify the ember server proxy timeout.

Options taken from https://github.com/nodejitsu/node-http-proxy#options


jboler:
How about just 1 new option, --proxy-in-timeout and I can rename it to --proxy-timeout ?

rwjblue:
I'm just thinking that there are already a number of proxy options (`--proxy` itself, `--secure-proxy`, `--transparent-proxy`), maybe we should allow `--proxy` to be an object (or JSON string) that groups all the proxy options 🤔

gnapse:
I too would benefit from this, and being a bit of an edge case there's no harm in accepting it as is to solve the pain of those who need it. We can always open a new ticket to discuss what to do about all the proxy options and how to refactor them, including this new one. Otherwise the acceptance of this feature would be delayed.

klobuczek:
I think there should be just one proxy option that allows funneling any option proxy supports. Like for example jruby allows passing any option to java with the `-J` option e.g. `-J-Xmx512m`.

PR #7929, opened 2018-07-20
Changed the condition to check if addon name is same in index.js and package.json to be generic

arthirm:
Earlier the warning was displayed for addon's with different names only if the addon's had scopes in their names.

### DIFF
--- a/tests/unit/models/addon-test.js
+++ b/tests/unit/models/addon-test.js
@@ -330,10 +330,19 @@ describe('models/addon.js', function() {
 
       it('returns true when the addon name is prefixed in package.json and not in index.js', function() {
         process.env.EMBER_ADDON_ENV = 'development';
-
+        project.root = 'foo';
         project.name = () => ('@foo/my-addon');
         addon.name = 'my-addon';
         expect(addon.isDevelopingAddon(), 'addon is being developed').to.eql(true);
+      });
+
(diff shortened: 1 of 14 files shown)


tests/unit/models/addon-test.js
```diff
@@ -330,12 +330,21 @@ describe('models/addon.js', function() {
...
         process.env.EMBER_ADDON_ENV = 'development';
-
+        project.root = 'foo';
         project.name = () => ('@foo/my-addon');
         addon.name = 'my-addon';
         expect(addon.isDevelopingAddon(), 'addon is being developed').to.eql(true);
       });
 
+      it('displays warning if addon name is different in package.json and index.js ', function() {
+        process.env.EMBER_ADDON_ENV = 'development';
```
rwjblue: This seems to never be cleaned up (you will need to add an `afterEach(function() { delete process.env.EMBER_ADDON_ENV; })` to ensure that it is cleared)...

arthirm:
There is already an afterEach which cleans up the env variable.

The tests are failing because earlier the `isDevelopingAddon` function will return false if the addon has different names in index.js and package.json. Now because I corrected the logic, lint is run on many files which were not run before. So correcting the fixtures files should fix the failures.

rwjblue:
Thank you!

PR #7934, opened 2018-07-21
Fix parallel

stefanpenner:
- [x] This will need a backport to release / beta
- [x] confirm this works 
- [x] see if I can easily make this fail a test


stefanpenner:
@rwjblue is back-porting this.

PR #7935, opened 2018-07-22
WIP: Scan all files for AMD instead of the first one

kellyselden:
Through my testing in real apps (DELAYED_TRANSPILATION), the simple approach to filtering out AMD is not good enough. We were trying to scan addon files, and on the first found file, determine whether the whole tree is AMD or not. In the real world, addons are doing any comination of the following:
1. calling `babel.transpileTree` with our options (ES6)
1. calling `babel.transpileTree` with their own options (AMD)
1. calling `_super` in `treeForAddon` (ES6)
1. running rollup and outputting AMD
1. running rollup and outputting ES6
1. bundle other addons in their own tree (mixed)

More specifically, in [Ghost-Admin](https://github.com/TryGhost/Ghost-Admin) plus [vertical-collection](https://github.com/html-next/vertical-collection), The addon is doing 1, 2, 3, 4 and 6 from the above, resulting in a mix of AMD and ES6. cc @kevinansfield @runspired
The solution is to remove the optimization in AmdFunnel that early exits on the first found file, and instead treat it like a real funnel that checks each file.


sandstrom:
I'm doing some issue gardening 🌱🌿 🌷 and came upon this issue. Since it's quite old I just wanted to ask if this is still relevant? 

Based on my (limited) view it seems that broccoli will eventually get replaced by embroider and webpack. If that's the case, maybe we can close this?

By closing some old issues we reduce the list of open issues to a more manageable set.

PR #7937, opened 2018-07-23
Bumping default target browser for ie to ie11

arthirm:


rwjblue:
Updated test looks great, restarted CI to see if recent fixes on master fix the CI failures...

tests/acceptance/new-test.js
```diff
@@ -31,6 +31,7 @@ describe('Acceptance: ember new', function() {
   }));
 
   afterEach(function() {
+    delete process.env.CI;
```
rwjblue: I believe this is causing the failures we are seeing in CI. Specifically because we rely on `process.env.CI` to be set so that we use the correct configuration for chrome (to avoid the setuid errors). See our default generated `testem.js` for how that works.

Can you tweak this to stash off the value and restore it instead? Something like:

```js
let ORIGINAL_PROCESS_ENV_CI;
beforeEach(function() {
  // ...snip...
  ORIGINAL_PROCESS_ENV_CI = process.env.CI;
});

afterEach(function() {
  if (ORIGINAL_PROCESS_ENV_CI === undefined) {
    delete process.env.CI;
  } else {
    process.env.CI = ORIGINAL_PROCESS_ENV_CI;
  }
});
```

rwjblue:
Thank you!

PR #7941, opened 2018-07-26
Fix bugs in package hook

twokul:
Notable changes:

+ aligned `package` function behaviour with the one in RFC: `package`
function can now only be passed through `options` object to the `EmberApp`
constructor
+ `toArray` method now actually returns an array, not a broccoli tree
+ `_legacyPackager` renamed to `_legacyPackage`; since it's private API
and it wasn't announced or broadcasted (more over, it is behind a
feature flag and only used internally), name change should be fine
+ `addonPostprocessTree('all')` is now getting properly called (this bug
was surfaced whilst integrating with `ember-engines`)

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -179,6 +179,7 @@ class EmberApp {
       },
     });
 
+    this._isPackageHookSupplied = typeof this.options.package === 'function';
     this._cachedAddonBundles = {};
   }
 
@@ -1604,11 +1605,8 @@ class EmberApp {
     @return {Array} An array of trees
    */
   toArray() {
-    // allows us to optimize if they aren't using the hook
-    let isPackageHookSupplied = typeof this.package === 'function';
-
-    let trees = [
-      this.getAppJavascript(isPackageHookSupplied),
+    return [
+      this.getAppJavascript(this._isPackageHookSupplied),
       this.getAddonTemplates(),
       this.getAddonStyles(),
       this.getTests(),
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1752,12 +1738,26 @@ class EmberApp {
     @return {Tree}                  Merged tree for this application
    */
   toTree(additionalTrees) {
-    let tree = mergeTrees(this.toArray().concat(additionalTrees || []), {
+    let trees = this.toArray().concat(additionalTrees || []);
+    let isPackageHookSupplied = typeof this.package === 'function' || typeof this.options.package === 'function';
```
kellyselden: We should either extract this or pass it in to toArray to avoid duplication.

lib/broccoli/ember-app.js
```diff
@@ -1752,12 +1738,26 @@ class EmberApp {
...
-      annotation: 'TreeMerger (allTrees)',
+      annotation: 'Full Application',
     });
 
-    return this.addonPostprocessTree('all', tree);
+    fullTree = this._debugTree(fullTree, 'prepackage');
+
+    if (experiments.PACKAGER) {
+      if (isPackageHookSupplied) {
+        return this.addonPostprocessTree('all', packageFn.call(this, fullTree));
```
kellyselden: Maybe we should do `packagedTree = packageFn.call(this, fullTree)` here, then we can do `return this.addonPostprocessTree('all', packagedTree || this.defaultPackage(fullTree))` below.

twokul:
@kellyselden comments addressed

kellyselden:
One last thing. I feel like we should take a quick informal vote on which one is preferred:

```js
  let app = new EmberApp(defaults, {
  });

  app.package = customPackager({
    // additional options
  });

  return app.toTree();
};
```

vs

```js
  let app = new EmberApp(defaults, {
    package: customPackager({
      // additional options
    })
  });

  return app.toTree();
};
```

less nesting vs less LOC.

PR #7942, opened 2018-07-27
Set default `__root__` to `src` (MU blueprints)

dcyriller:
I didn't find  a place to test it. Will update the PR with a test if needed!


dcyriller:
Closing it, as not ready yet.

PR #7946, opened 2018-07-31
upgrade ember-cli-htmlbars to 3.x

stefanpenner:
No functional changes, it only dropped node 4.x support


stefanpenner:
@rwjblue we should backport to beta as well. No risk, since it only added tests and dropped unsupported node support.

stefanpenner:
@rwjblue should I backport, or do you want to?

PR #7947, opened 2018-07-31
Adding support for in repo addons in non-conventional directories

scalvert:
Currently, ember-cli assumes all in-repo addons should live in `lib`. As such, blueprints that are run to generate entities in in-repo addons also assume `lib`. This PR allows for custom paths to be provided while defaulting to `lib`.


jackson-dean:
Do we also need to do something to make the `destroy` command work for custom paths as well?
Also seems a somewhat inconsistent use of `path.join(...)` and `string.join('/')`, but it was doing that before, so it probably just works everywhere or someone would have complained by now.

scalvert:
Good question, @jackson-dean. I'll test out the `destroy` to see if it works. 

Agree about the inconsistencies. Generally, I prefer not using string methods to manipulate paths. I may consider changing it, but don't want to pull the thread too much on this PR.

rwjblue:
Blueprints control both `generate` and `destroy` (`install` method is `ember generate` and `uninstall` is `ember destroy`), both should be addressed by this...

PR #7950, opened 2018-08-01
Better handle ambiguity between package.json and index.js name field.

rwjblue:
* Backports #7929 which added a warning if the `name` property in an addon's `index.js` doesn't match the package.json` value
* Changes the `addon` blueprint to ensure `name` property just references the `package.json`'s `name`
* Updates changes originally from #7929 to be a hard error when locally developing an addon whose `package.json` name differs from `index.js` name.
  * Ensure error message includes the offending addon's path (either `process.cwd()` or `path.relative(process.cwd(), addon.root)`)
* Updates `Project.prototype.findAddonByName` / `Addon.prototype.findAddonByName` to handle the changes to mentioned above. Specifically, `findAddonByName` will prefer exact matches to the `package.json` name over exact matches in `index.js` name.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -366,7 +366,7 @@ let addonProto = {
 
       // Addon names in index.js and package.json should be the same at all times whether they have scope or not.
       if (this.root === this.parent.root) {
-        if (parentName !== this.name) {
+        if (parentName !== this.name && !process.env.EMBER_CLI_IGNORE_ADDON_NAME_MISMATCH) {
           let pathToDisplay = process.cwd() === this.root ? process.cwd() : path.relative(process.cwd(), this.root);
 
           throw new SilentError(
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -366,12 +366,12 @@ let addonProto = {
 
       // Addon names in index.js and package.json should be the same at all times whether they have scope or not.
       if (this.root === this.parent.root) {
-        if (parentName !== this.name && !this._hasWarnedForMismatchedNames) {
-          this.ui.writeWarnLine(
-            'Your names in package.json and index.js should match. ' +
-            `You currently have ${parentName} in package.json and ${this.name} in index.js.`);
+        if (parentName !== this.name) {
+          let pathToDisplay = process.cwd() === this.root ? process.cwd() : path.relative(process.cwd(), this.root);
 
-          this._hasWarnedForMismatchedNames = true;
+          throw new SilentError(
+            'Your names in package.json and index.js should match. ' +
+            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.`);
         }
 
         return true;
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -36,7 +36,6 @@ const addonProcessTree = require('../utilities/addon-process-tree');
 const emberCLIBabelConfigKey = require('../utilities/ember-cli-babel-config-key');
 const semver = require('semver');
 const processModulesOnly = require('../broccoli/babel-process-modules-only');
-const npa = require('npm-package-arg');
 const registryHasPreprocessor = require('../utilities/registry-has-preprocessor');
 
 const BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS = Symbol('BUILD_BABEL_OPTIONS_FOR_PREPROCESSORS');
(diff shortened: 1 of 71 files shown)


lib/models/addon.js
```diff
@@ -365,19 +364,14 @@ let addonProto = {
...
-      }
-      if (parsedName === this.name) {
-        if (!this._hasWarnedForMismatchedNames) {
-          this.ui.writeWarnLine(
-            'Your names in package.json and index.js should match. ' +
-            `You currently have ${parsed.name} in package.json and ${this.name} in index.js.`);
+      // Addon names in index.js and package.json should be the same at all times whether they have scope or not.
+      if (this.root === this.parent.root) {
+        if (parentName !== this.name) {
+          let pathToDisplay = process.cwd() === this.root ? process.cwd() : path.relative(process.cwd(), this.root);
```
stefanpenner: To understand correctly, this code is to display the full root, unless contained within the cwd(), then only should the relative path.

Is that worth it, or should we simply just always show the full path?

lib/utilities/find-addon-by-name.js
```diff
@@ -1,11 +1,34 @@
...
 module.exports = function findAddonByName(addons, name) {
-  function matchAddon(name, addon) {
-    return name === addon.name || (addon.pkg && name === addon.pkg.name);
-  }
+  let exactMatchFromPkg = addons.find(addon => addon.pkg && addon.pkg.name === name);
+  let exactMatchFromIndex = addons.find(addon => addon.name === name);
+  let unscopedMatchFromIndex = addons.find(addon => addon.name && unscope(addon.name) === unscope(name));
 
-  return find(addons, addon => matchAddon(name, addon));
+  return [exactMatchFromPkg, exactMatchFromIndex, unscopedMatchFromIndex].find(addon => !!addon);
```
stefanpenner: we should likely consider warning if there is ambiguity (even if we choose to put that warning behind a flag)

lib/utilities/find-addon-by-name.js
```diff
@@ -1,11 +1,34 @@
...
+  - index.js name matches requested name exactly
+  - unscoped (leaf portion) index.js name matches unscoped requested name
+*/
 module.exports = function findAddonByName(addons, name) {
-  function matchAddon(name, addon) {
-    return name === addon.name || (addon.pkg && name === addon.pkg.name);
-  }
+  let exactMatchFromPkg = addons.find(addon => addon.pkg && addon.pkg.name === name);
+  let exactMatchFromIndex = addons.find(addon => addon.name === name);
+  let unscopedMatchFromIndex = addons.find(addon => addon.name && unscope(addon.name) === unscope(name));
```
stefanpenner: `unscoped(name)` should be done once, outside of the loop

lib/utilities/find-addon-by-name.js
```diff
@@ -1,11 +1,34 @@
...
+  - package.json name matches requested name exactly
+  - index.js name matches requested name exactly
+  - unscoped (leaf portion) index.js name matches unscoped requested name
+*/
 module.exports = function findAddonByName(addons, name) {
-  function matchAddon(name, addon) {
-    return name === addon.name || (addon.pkg && name === addon.pkg.name);
-  }
+  let exactMatchFromPkg = addons.find(addon => addon.pkg && addon.pkg.name === name);
+  let exactMatchFromIndex = addons.find(addon => addon.name === name);
```
stefanpenner: if exactMatchFromPkg return early

stefanpenner:
1. make tests green
2. try in some apps, and see how noisy it is
3. report back here.

stefanpenner:
So what I implemented looks good in a real non-trivial app, and more or less passes tests.

- [ ] still need to fix tests
- [ ] need to implement dag-map features (but requires unit tests first)

dfreeman:
If I'm reading this right, I think we (and presumably others) will run into trouble if this lands before a solution for https://github.com/ember-engines/ember-engines/issues/524 does — that issue is currently blocking us from having the name of our engine addons match their package name.

dfreeman:
But if `isDevelopingAddon` raises an error, I can't do any development in that engine, right?

rwjblue:
> But if isDevelopingAddon raises an error, I can't do any development in that engine, right?

You basically would have to implement `isDevelopingAddon`, but obviously that sucks (because you would then lose out on the local dev features). The real answer here is just that we need to fix the issue over in engines ASAP. In the meantime, I'll add an opt-out for the error via process environment variable...

dfreeman:
Thanks ❤️

PR #7951, opened 2018-08-02
Update testing instructions

kanongil:


kanongil:
Actually, you should probably also add a note that a local yarn installation is required for the suite to pass.

PR #7958, opened 2018-08-04
Gather doc files in docs directory

dcyriller:


dcyriller:
I double checked and found out I missed two inbound links. Thank you :)

dcyriller:
Just found out why the tests were failing, it should be good now.

docs/perf-guide.md
```diff
@@ -214,7 +214,7 @@ DEBUG_DEPTH=5 DEBUG_LEVEL=debug DEBUG=broccoli-merge-trees: ember build
 `[... ConcatWithMaps#782 -> BroccoliMergeTrees#783]` means that the log entry
 occurred in broccoli merge-trees node with id 783, whose parent was a concat
 with maps node with id 782.  These ids are shown in the visualization graph.
-See [Visualization](#Visualization) for details.
+See [Visualization](./perf-guide.md#Visualization) for details.
```
Turbo87: this change seems unnecessary 🤔

tests/unit/tasks/serve-test.js
```diff
@@ -70,7 +70,7 @@ describe('serve task', function() {
     });
 
     it(`Serves ember app from given path`, function() {
-      runServeTask('assets');
+      runServeTask('docs');
```
Turbo87: can you explain this change?

docs/perf-guide.md
dcyriller: I did this because the link wasn't working, I'm not sure why but you can try to click it here: https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md

tests/unit/tasks/serve-test.js
dcyriller: In this PR I've removed the `assets` directory.

The `serve` task may take `path` as an option. It is the build directory to serve.

This test is covering this specific option. As I deleted `assets`, I had to replace it with another existing directory.

Turbo87: ahhh, I see, thanks!

dcyriller: ahah this failing test wasn't clear to me as well :)

docs/perf-guide.md
dcyriller: Yes, it was, I've made the update.

dcyriller:
@Turbo87 do you think we could merge it before it's getting too old?

I did this while reading ember-cli for the first time. The aim of this PR was to simplify the project's code reading.

Turbo87:
yes, thanks for the reminder!

PR #7960, opened 2018-08-06
account for filePath overrides in blueprint linting

kellyselden:
I've discovered that /files/ is not good enough for linting ignore because a `filePath` override can be anything. This new rule covers that.


kellyselden:
> Noticed that it seems like we are also ignoring blueprints/foo/index.js (which should be linted)?

That terminating / should mean that blueprints/foo/index.js is covered.

rwjblue:
Closing for now (I think customizing `files/` _should_ require tweaking linting config)...

PR #7961, opened 2018-08-06
Turn experiments object into a function

dcyriller:
When testing an experiment in `ember-cli` scope, a new CI scenario is the go-to solution. This involves scopping tests like this:
```js
// in ember-cli
if (experiments.MY_EXPERIMENT) {
  describe('my experiment module', ...)
  // or
  module('my experiment module', ...)
} else {
  describe('regular module', ...)
  // or
  module('regular module', ...)
}
```

But when testing an experiment outside of `ember-cli` scope, things get different. And we may want to be able to test in sequence:

```js
// in ember-data or in another addon
describe('my addon blueprint', {
  it('works - in a classic app', ...);

  it('works - in a MU app', ...);
})
```

The issue is: as an object, [experiments](https://github.com/ember-cli/ember-cli/blob/v3.4.0-beta.2/lib/experiments/index.js#L25) is cached between the two tests (run on the same process). In these scenarios, we need not to cache the object, hence the function.

### DIFF
--- a/docs/experiments.md
+++ b/docs/experiments.md
@@ -5,20 +5,19 @@
 They are defined in `lib/experiments/index.js`. For example:
 
 ```javascript
-let experiments = {
-  CONFIG_CACHING: isExperimentEnabled('CONFIG_CACHING'),
-};
-
+const availableExperiments = [
+  'CONFIG_CACHING',
+];
 ```
 
 When a new feature is added, all supporting code and tests **must** be guarded
 to ensure all tests pass when the feature is enabled _or_ disabled:
 
 ```javascript
-const experiments = require('../experiments');
+const { isExperimentEnabled } = require('../experiments');
 
 // ...snip...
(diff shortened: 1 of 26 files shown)


dcyriller:
A use of this PR would be here https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/153

rwjblue:
I am not opposed to changing this, but I would prefer a slightly different API.

Seems like we can make this work:

```
const { isExperimentEnabled } = require('./lib/experiments');

if (isExperimentEnabled('MODULE_UNIFICATION')) {

}
```

What do you think?

dcyriller:
It sounds more consistent indeed. I just updated the PR.

dcyriller:
I manually relaunched the tests. One test suite is still failing, it's the one that generates coverage. But it's also failing on master.

PR #7965, opened 2018-08-07
Don’t produce duplicate warnings when invoking findAddonByName with a…

stefanpenner:
…n ambiguous add-on name


stefanpenner:
@rwjblue the tests seem to be failing for an unrelated issue. I'll try to look into this in the AM.

PR #7966, opened 2018-08-07
adding flag to stop repeated dumping of same package errors

dcombslinkedin:
When package-info-cache dumps errors, this will prevent the same addon package from dumping its errors to the console log more than once, regardless of how many different other addons/projects refer to that same addon package.

### DIFF
--- a/lib/models/package-info-cache/package-info.js
+++ b/lib/models/package-info-cache/package-info.js
@@ -40,6 +40,8 @@ class PackageInfo {
     // Missing dependencies will not be considered an error, since they may
     // not actually be used.
     this.valid = true;
+
+    this._hasDumpedInvalidAddonPackages = false;
   }
 
   // Make various fields of the pkg object available.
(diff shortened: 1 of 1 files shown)


lib/models/package-info-cache/package-info.js
```diff
@@ -291,17 +291,22 @@ class PackageInfo {
   }
 
   dumpInvalidAddonPackages(addonPackageList) {
-    let invalidPackages = this.getInvalidPackages(addonPackageList);
+    if (!this.dumpedInvalidPackages) {
```
stefanpenner: 2 nit picks:

* as this is private state, can we do `this._dumpedInvalidPackages`
* could we keep the guard isolated at the top of the method?

```
dumpInvalidAddonPackages(addonPackageList) {
  if (this._dumpedInvalidPackages) { return; }
  this._dumpedInvalidPackages = true;

  // rest of code
}
```

For two reasons:

* keeps the guard code close together
* meets our existing style guide

dcombslinkedin: Sure. I'll fix it. I'd forgotten that's how you like to do the guard code.

stefanpenner:
LGTM

stefanpenner:
1 job failed, seemed to be unrelated NPM issue. Restarting it... let's see.

stefanpenner:
re-targetted to beta

stefanpenner:
tests are actually passing, but they failed do to an unrelated issue @rwjblue is debugging.

PR #7974, opened 2018-08-07
Prevent double builds in CI for branches pushed by owner.

rwjblue:
Prior to this change, the addon author (unless the addon was in an org) would always end up with _two_ CI builds for every PR. One for the branch push and one for the PR update.

This change prevents branch builds for branches other than `master`, therefore avoiding the issue.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -23,6 +23,12 @@ env:
     # See https://git.io/vdao3 for details.
     - JOBS=1
 
+branches:
+  only:
+    - master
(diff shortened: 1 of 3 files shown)


kellyselden:
My only concern is this would be an extra hurdle for people trying to set up things like greenkeeper, remembering to include their branch regex as well. That said, I often do this in my projects as well.

rwjblue:
Ya, dependabot (and I think renovate) handle this nicely by just leveraging the PR status. 

We can always add an additional whitelist to allow the green keeper / renovate / dependabot style branch names...

PR #7976, opened 2018-08-14
Upgrade ember-source to master in `new` blueprint

dcyriller:
This fixes a bug some MU experimenters ran into.

**How to reproduce**
Create a new MU app with latest `ember-cli` and generate a route / service.

When doing `EMBER_CLI_MODULE_UNIFICATION=true ember new my-app`, users get a project with an outdated version of `ember-source` (from March). 

Then, `EMBER_CLI_MODULE_UNIFICATION=true ember generate route foo` do not have the expected behavior, it generates the files inside `app` directory.

### DIFF
--- a/blueprints/module-unification-app/files/package.json
+++ b/blueprints/module-unification-app/files/package.json
@@ -3,13 +3,13 @@
   "version": "0.0.0",
   "private": true,
   "description": "Small description for <%= name %> goes here",
+  "repository": "",
   "license": "MIT",
   "author": "",
   "directories": {
     "doc": "doc",
     "test": "tests"
   },
-  "repository": "",
   "scripts": {
     "build": "ember build",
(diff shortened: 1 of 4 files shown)


dcyriller:
If using master branch of `ember-source` is not an issue, we could do the same for `ember-data`.

dcyriller:
@CodingItWrong I updated the PR with `ember-source`'s canary build (instead of master branch). This also fixes the tests. I should have used it in the first place. Could you give it another try?

CodingItWrong:
@dcyriller yup, using that updated canary build works for me! i as a non-merge-access person approve this PR 👍 hehe. Much appreciated!!!!

CodingItWrong:
Ping to reviewers—can someone review and merge this PR?

kellyselden:
LGTM thanks.

PR #7977, opened 2018-08-14
Use existing build from specified path, --path option added

SparshithNR:
Added --path option using which now serve takes build from specified path and avoids rebuilding the project. 

- If the path specified doesn't exist it throws an error and quits serve. 
- If build exists, serves files from given location, which will disable livereload and rebuilding the project.
- watcher and builder are mocked when --path option is used.


lib/commands/serve.js
```diff
@@ -38,6 +38,7 @@ module.exports = Command.extend({
     { name: 'ssl',                  type: Boolean, default: false },
     { name: 'ssl-key',              type: String,  default: 'ssl/server.key' },
     { name: 'ssl-cert',             type: String,  default: 'ssl/server.crt' },
+    { name: 'path',                 type: 'Path',                                                                                              description: 'Reuse an existing build at given path.' },
```
stefanpenner: funny whitespace between `,` and `description`

lib/tasks/serve.js
```diff
@@ -57,9 +81,20 @@ class ServeTask extends Task {
     this._runDeferred = RSVP.defer();
 
     return expressServer.start(options)
-      .then(() => this._runDeferred.promise);
+      .then(() => {
+        if (hasBuild) {
+          this.serveUrl(options);
```
stefanpenner: This surprises me, don't we still want to share the serve url? The user will still visit this URL, but we merely wont do rebuilds for them automatically.

lib/tasks/serve.js
```diff
@@ -18,6 +32,16 @@ class ServeTask extends Task {
   }
 
   run(options) {
+    let hasBuild = !!options.path;
+
+    if (hasBuild) {
+      if (!fs.existsSync(options.path)) {
+        throw new SilentError(`The path ${options.path} does not exist. Please specify a valid build directory to test.`);
```
stefanpenner: `Please specify a valid build directory to test.` should likely be `Please specify a valid build directory to serve.`

tests/unit/tasks/serve-test.js
```diff
@@ -59,10 +59,24 @@ describe('serve task', function() {
...
   }
 
+  describe('run with path', function() {
+    it(`Throws error if path doesn't exist`, function() {
+      expect(runServeTask.bind(this, 'xyz')).to.throw('The path xyz does not exist. Please specify a valid build directory to test.');
+    });
+
+    it(`Serves ember app from given path`, function() {
+      runServeTask('assets');
+      RSVP.resolve().then(() => {
```
stefanpenner: `return`

PR #7984, opened 2018-08-17
Reject when command args validation fails

bobisjan:
When required argument is missing, then exit code should be non-zero value.


rwjblue:
Thank you!

Issue #7987, opened 2018-08-17
treeForStyles regression since 3.4.0-beta.2


simonihmig:
When using `ember-cli-sass` and `ember-bootstrap` in an app using at least 3.4.0-beta.2 (or -beta.3), this error shows up:

```
Sass Syntax Error (SassCompiler) in stylestest/styles/app.scss:3:1

Error: File to import not found or unreadable: ember-bootstrap/bootstrap.
        on line 3 of tmp/sass_compiler-input_base_path-RoP5qwYN.tmp/stylestest/styles/app.scss
>> @import "ember-bootstrap/bootstrap";
```

This does *not* happen with 3.4.0-beta.1 (or any other version before that AFAIK)!

The addon funnels Sass files from `node_modules` in `treeForStyles()`, for easy importing in `app.scss` (here: https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L193-L199).

FWIW, `ember-sass-bootstrap` does basically the same, and suffers from exactly the same problem! (cc @ef4)

### Investigation

I tried to investigate this, could not really find anything suspicious in between -beta.1 and -beta.2 (https://github.com/ember-cli/ember-cli/compare/v3.4.0-beta.1...v3.4.0-beta.2#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) though. I patched `ember-cli-sass` to wrap the input tree for the Sass plugin in a `debug()` (here: https://github.com/aexmachina/ember-cli-sass/blob/master/index.js#L26), and these are the results:

For 3.4.0-beta.1:
![image](https://user-images.githubusercontent.com/1325249/44292862-a0641b80-a287-11e8-8683-ec357f6ec551.png)

For 3.4.0-beta.2:
![image](https://user-images.githubusercontent.com/1325249/44292868-aa861a00-a287-11e8-9171-6aa45e1b3ab4.png)

As you can see the trees are quite different, for beta.2
* has a `stylestest` (name of app) folder instead of `app`
* that tree includes all the *.js files as well! 🤔
* the funneled `treeForStyles` tree (the highlighted `ember-bootstrap` line) is *within* this folder, instead of a sibling of it

### Reproduction

Find the steps to reproduce this here: https://github.com/kaliber5/ember-bootstrap/issues/645

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.4.0-beta.3
http_parser: 2.8.0
node: 10.5.0
[...]

PR #7991, opened 2018-08-22
Fix for style processing

twokul:
As Ember CLI gathers the trees from add-ons, it needs to properly place
them within its "working" directory for processing later (both JS and
SASS/LESS).

The root cause of the issue was that files returned from the add-ons
weren't properly moved, and will later cause transpilation failures as
preprocessing step couldn't find files it needed to resolve `@import`
statements within `app.scss`.

Related issue https://github.com/ember-cli/ember-cli/issues/7987

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1078,7 +1078,8 @@ class EmberApp {
     });
 
     return new Funnel(mergedAddonStyles, {
-      destDir: this.name,
+      allowEmpty: true,
+      destDir: `${this.name}/styles`,
     });
   }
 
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1078,7 +1078,8 @@ class EmberApp {
     });
 
     return new Funnel(mergedAddonStyles, {
-      destDir: this.name,
+      allowEmpty: true,
```
kellyselden: should we regression test this empty tree?

twokul: we should!

PR #7994, opened 2018-08-23
[PERF] Speed up package info cache

stefanpenner:
[PERF] PackageInfoCache should not traverse node_modules which will not be considered for add-on discovery

* reduces vanilla ember-app boot up time by 200-400ms
* reduces vanilla ember-app add-on discovery by 2/3’s
* apps with large node_module folders benefit much more.

How does this work?

only walk node_modules of:

* the root dependency (the app, add-on, etc)
* child dependencies marked as ember-addon

Skip over the rest and consider them as empty as our algorithm will not discover add-ons from these node_modules directories anyways.

This reduces unneeded node_module traversals dramatically

### DIFF
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -240,7 +240,7 @@ class PackageInfoCache {
    */
   loadProject(projectInstance) {
 
-    let pkgInfo = this._readPackage(projectInstance.root, projectInstance.pkg);
+    let pkgInfo = this._readPackage(projectInstance.root, projectInstance.pkg, true);
 
     // NOTE: the returned val may contain errors, or may contain
     // other packages that have errors. We will try to process
@@ -494,8 +494,10 @@ class PackageInfoCache {
    * @method _readPackage
    * @param {String} pkgDir the path of the directory to read the package.json from and
    *  process the contents and create a new cache entry or entries.
+   * @param {Boolean} isRoot, for when this is to be considered the root
+   * package, whose dependencies we must all consider for discovery
    */
-  _readPackage(packageDir, pkg) {
+  _readPackage(packageDir, pkg, isRoot) {
     let normalizedPackageDir = path.normalize(packageDir);
 
     // Most of the time, normalizedPackageDir is already a real path (i.e. fs.realpathSync
@@ -572,7 +574,7 @@ class PackageInfoCache {
 
     // Create a new PackageInfo and load any errors as needed.
     // Note that pkg may be an empty object here.
-    pkgInfo = new PackageInfo(pkg, realPath, this);
+    pkgInfo = new PackageInfo(pkg, realPath, this, isRoot);
 
     if (setupErrors.hasErrors()) {
       pkgInfo.errors = setupErrors;
@@ -582,7 +584,7 @@ class PackageInfoCache {
     // If we have an ember-addon, check that the main exists and points
     // to a valid file.
     if (Array.isArray(pkg.keywords) &&
-        pkg.keywords.indexOf('ember-addon') >= 0) {
+      pkg.keywords.indexOf('ember-addon') >= 0) {
 
       // Note: when we have both 'main' and ember-addon:main, the latter takes precedence
       let main = (pkg['ember-addon'] && pkg['ember-addon'].main) || pkg['main'];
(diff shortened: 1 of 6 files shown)


stefanpenner:
cc @dcombslinkedin

lib/models/package-info-cache/index.js
```diff
@@ -494,8 +494,10 @@ class PackageInfoCache {
    * @method _readPackage
    * @param {String} pkgDir the path of the directory to read the package.json from and
    *  process the contents and create a new cache entry or entries.
+   * @param {Boolean} isRoot, for when this is to be considered the root
+   * package, whos dependencies we must all consider for discovery
```
dcombslinkedin: whose, not 'whos'

lib/models/package-info-cache/index.js
```diff
@@ -626,15 +628,21 @@ class PackageInfoCache {
...
+      // if we find an issue, if nobody below is actually invoking the addon.
+      let nodeModules = this._readNodeModulesList(path.join(realPath, 'node_modules'));
 
-    if (nodeModules) {
-      pkgInfo.nodeModules = nodeModules;
+      if (nodeModules) {
+        pkgInfo.nodeModules = nodeModules;
+      }
+    } else {
+      // will not have addons, so even if their are node_modules here, we can
```
dcombslinkedin: 'there are', not 'their are'

lib/models/package-info-cache/index.js
```diff
@@ -626,15 +628,21 @@ class PackageInfoCache {
...
+      // simply pretend their are none.
```
dcombslinkedin: 'there', not 'their'

stefanpenner:
From an addon discovery standpoint the algorithm is no different. The exception being it traverses an order to two magnitudes less directories as it knows it cannot discover addons down those paths.

Although for some commands, like build a second or so isn’t much. But it is quite a lot for other commands like help or generate etc. especially for vanilla apps, where the package-info work almost doubles cli boot up time.

lib/models/package-info-cache/package-info.js
```diff
@@ -5,6 +5,15 @@ const ErrorList = require('./error-list');
...
 const AddonInfo = require('../addon-info');
 
+function isEmberAddon(pkg) {
+  if (Array.isArray(pkg.keywords) && pkg.keywords.indexOf('ember-addon') > -1) {
+    // is an addon, so it may have addons.
+    return true;
+  } else {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
stefanpenner: Ya I wasn’t sure either, this was slightly more complicated in an earlier revision. I should remove the pointless if statement as your recommend

stefanpenner:
The reason I noticed this, was I am working on a simpler and faster addon/cli test harness and I noticed I was wasting 600ms for each test iteration in this code. Before 800ms and after 200ms.

This change improves my test throughout but nearly 3x.

dcombslinkedin:
@stefanpenner , I don't see any problems with the speedup you made. And I'm definitely all for making things like help and generate go faster. I'd noticed that lag myself a couple times, so it's all good. I'm not sure I could prove something regressed at this point, since the original code was intended to just follow the idea of the node resolution algorithm, just remove the repeated file-system accesses. Since you know the algorithm results for addon-discovery are unchanged, that's good enough for me! Thanks. ShipIt!

lib/models/package-info-cache/node-modules-list.js
```diff
@@ -11,14 +12,34 @@ const ErrorList = require('./error-list');
  * @public
  * @class NodeModulesList
  */
-class NodeModulesList {
-  constructor(realPath, cache) {
+module.exports = class NodeModulesList {
+  constructor(realPath) {
     this.realPath = realPath;
-    this.cache = cache;
```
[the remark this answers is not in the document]
stefanpenner: Ya I wasn’t sure, but I can bring it back for consistency.

lib/models/package-info-cache/index.js
stefanpenner: english, how does it even work

dcombslinkedin: :-) My mom was an English major and writer, what can I say? It's inbred. ;-)

stefanpenner:
@dcombslinkedin thanks for the review!

dcombslinkedin:
No problem. I still have a small bit of 'it's my baby!', but that will go away soon. ;-) I love the fact this makes the performance even better.

stefanpenner:
@dcombslinkedin ya its just a small tweak, aligning the PackageInfo's node resolution algorithm to be more closely to that of AddonDiscovery, as we only utilize PackageInfoCache to service AddonDiscovery.

stefanpenner:
test failures appear to be unrelated NPM issues ...

PR #8000, opened 2018-08-28
Adding `--in` option for customizing path for in repo addon blueprints

scalvert:
Ember generate allows you to generate entities inside in-repo addons using the `--in-repo-addon` option (not to be confused with the `ember generate` `in-repo-addon` subcommand, which creates an in-repo addon). This option only partially accounts for custom directories (following [this PR](https://github.com/ember-cli/ember-cli/pull/7947)), which is a requirement in some larger applications. 

This PR proposes removing `--in-repo-addon` in favor of a new option: `--in`. The `--in` option differs from the existing `--in-repo-addon` option as it takes a path pointing to the root of the **addon** you would like your blueprint run against.

Valid use cases for the `--in` option would be:

### Case 1

Projects contained in a subdirectory of the root of the repository (monorepos):

```
ember generate component foo-bar --in ./subproject/baz
```

**Result**: 
Will create component files in:

```
// component blueprint 
subproject/baz/app/components/foo-bar.js
subproject/baz/app/templates/components/foo-bar.hbs
    
// component-test blueprint
subproject/tests/integration/components/foo-bar-test.js

// component-addon blueprint
subproject/baz/app/components/foo-bar.js
```

### Case 2

Standard in repo addons (contained in lib):

```
ember generate component foo-bar --in ./lib/baz
```

**Result**: 
Will create component files in:

```
// component blueprint
lib/baz/addon/components/foo-bar.js
lib/baz/addon/templates/components/foo-bar.hbs
    
// component-test blueprint
tests/integration/components/foo-bar-test.js

// component-addon blueprint
lib/baz/app/components/foo-bar.js
```

### Case 3

In repo addon in custom root level path:

```
ember generate component foo-bar --in ./custom/baz
```

**Result**: 
Will create component files in:

```
// component blueprint
custom/baz/addon/components/foo-bar.js
custom/baz/addon/templates/components/foo-bar.hbs
    
// component-test blueprint
tests/integration/components/foo-bar-test.js

// component-addon blueprint
custom/baz/app/components/foo-bar.js
```

### Case 4

In repo addon in custom nested path:

```
ember generate component foo-bar --in ./custom/sub/baz
```

**Result**: 
Will create component files in:

```
// component blueprint
custom/sub/baz/addon/components/foo-bar.js
custom/sub/baz/addon/templates/components/foo-bar.hbs
    
// component-test blueprint
tests/integration/components/foo-bar-test.js

// component-addon blueprint
custom/sub/baz/app/components/foo-bar.js
```


lib/models/blueprint.js
```diff
@@ -387,12 +388,15 @@ let Blueprint = CoreObject.extend({
...
-  _checkInRepoAddonExists(inRepoAddon) {
-    if (inRepoAddon) {
-      let addon = findAddonByName(this.project, inRepoAddon);
+  _checkInRepoAddonExists(options) {
+    let addon;
+
+    if (options.inRepoAddon || options.in) {
+      addon = options.inRepoAddon ? findAddonByName(this.project, options.inRepoAddon) : findAddonByPath(this.project, options.in);
+
       if (!addon) {
```
thoov: This error seems specific to the inRepoAdddon case. We should either change the message or guard against going down this path

scalvert: Already fixed that. Pushing now.

scalvert: Sorry, I thought you were referring to the error message. In terms of the error, we still want to check whether the in-repo addon exists. This is still a valid case regardless of the option used. Does that make sense?

lib/models/blueprint.js
```diff
@@ -385,14 +385,26 @@ let Blueprint = CoreObject.extend({
...
+    }
+
+    if (options.in) {
+      addon = findAddonByPath(this.project, options.in);
+
       if (!addon) {
-        throw new SilentError(`You specified the in-repo-addon flag, but the ` +
-          `in-repo-addon '${inRepoAddon}' does not exist. Please check the name and try again.`);
+        throw new SilentError(`You specified the 'in' flag, but the ` +
+          `in-repo-addon '${options.inRepoAddon}' does not exist. Please check the name and try again.`);
```
thoov: Won't options.inRepoAddon be undefined here? Should the error be more like: 

```js
throw new SilentError(`You specified the 'in' flag, but the ` +
          `in-repo-addon located at '${options.in}' does not exist. Please check the path and try again.`);
```

scalvert: That was a mistake. Good catch.

lib/models/blueprint.js
```diff
@@ -385,14 +385,26 @@ let Blueprint = CoreObject.extend({
...
+      }
+    }
+
+    if (options.in) {
+      addon = findAddonByPath(this.project, options.in);
+
+      if (!addon) {
+        throw new SilentError(`You specified the 'in' flag, but the ` +
+          `in-repo-addon '${options.in}' does not exist. Please check the name and try again.`);
```
rwjblue: Should probably not say `in-repo-addon` here

lib/models/blueprint.js
```diff
@@ -800,6 +820,7 @@ let Blueprint = CoreObject.extend({
       hasPathToken: this.hasPathToken,
       inAddon,
       inRepoAddon: options.inRepoAddon,
+      in: options.in,
```
rwjblue: My concern here is that `options.in` is not "sanitized" (transformed into an absolute path) and anyone that wants to use it to build their locals will need to do the `path.relative(project.root, options.in)` dance... Should we mutate `options.in` early in the argument parsing (e.g. up in the command itself)?

lib/commands/generate.js
```diff
@@ -22,6 +22,7 @@ module.exports = Command.extend({
     { name: 'classic',       type: Boolean, default: false, aliases: ['c'] },
     { name: 'dummy',         type: Boolean, default: false, aliases: ['dum', 'id'] },
     { name: 'in-repo-addon', type: String,  default: null,  aliases: ['in-repo', 'ir'] },
+    { name: 'in',            type: String,  default: null },
```
rwjblue: Can you add a description here?

lib/models/blueprint.js
scalvert: Yep, agree. I was thinking just 'in repo addon', not dashes. Would that be more clear?

lib/models/blueprint.js
scalvert: Yep we can. I know you and I discussed this via chat, and I'm coming around to this idea. I can't think of a reason why we'd need `options.in` in its bare form, eg. un-absolutized, can you?

lib/commands/generate.js
scalvert: For sure.

lib/models/blueprint.js
rwjblue: No, not really. Though if we _do_ I think we could always reparse from `process.argv` (e.g. its still "around", just harder to get to)...

scalvert: 👍

trabus:
I would like to request not removing the `--in-repo-addon` option and instead rewire it to use the `--in` option with default arguments pathing to the lib folder, fulfilling the current functionality. Many people I know use the `-ir` or `--in-repo-addon` flag as is currently, and I think removing the option could be a jarring experience.

trabus:
Your scenarios described in 2, 3, and 4 should also have generated an app re-export file too.
```
// component addon blueprint
lib/baz/app/components/foo-bar.js
```
I'm fairly certain this is required for components at minimum, since the template for components is included as the layout and not exported from addons like regular templates (please correct me if I've missed something). Any blueprint with an `-addon` tacked onto it will generate this re-export file to be picked up by the consuming application.

rwjblue:
@trabus - Thanks for reviewing! FWIW, this doesn't remove `--in-repo-addon` / `--ir` (it just adds `--in`), I totally agree that (at a later time) we should make `--in-repo-addon` leverage `--in` internally...

scalvert:
@trabus thanks for the review. I also thought of adding defaulting to `lib`. I think ultimately we could do that. 

As for the reexports, I may have omitted them but ultimately I'm not changing the blueprints at all. This PR merely changes how those blueprints are run. I'm happy to add the reexports for clarity.

jakebixbyavalara:
@rwjblue I was mainly responding to his mention up above where he'd discussed the plan to deprecate the `--in-repo-addon` flag in future PRs. I figured rewiring would probably be easier than deprecating anyhow.
> Deprecating the old --in-repo-addon option should be done via the following:
> * add a deprecation warning when using the --in-repo-addon option
> * deprecate at next LTS version of ember-cli

@scalvert I actually think what you've done here is a great primitive and shouldn't be limited to in repo addons, especially with the rise of monorepos and different project file structures. The `--in` flag makes a ton of sense in that regard (as "generate this blueprint *_in_* _blank_"), and reusing it for the internals of other commands actually is a pattern quite I'm fond of.

Regarding the reexports, I just wanted to make sure we were accounting for them. I wasn't sure if you'd left them off or if you were including actual output from your existing PR code. I just wanted to make sure it wasn't overlooked. 👍

scalvert:
I'm with you 100%, @jakebixbyavalara.

scalvert:
@jakebixbyavalara I've updated the PR description with the re-exports.

I think it's reasonable to move the `--in-repo-addon` internals to use the `--in` internal functionality. This would definitely clean the code up a bit. I'm open to taking out the deprecation warning in favor of moving to that approach. Does that sound good?

jakebixbyavalara:
@scalvert Yeah, that sounds great.

jakebixbyavalara:
@scalvert I meant to ask earlier, but is there a reason your PR is going against `ember-cli:beta` instead of `ember-cli:master`?

lib/commands/destroy.js
```diff
@@ -51,6 +54,10 @@ module.exports = Command.extend({
       commandOptions.pod = !commandOptions.pod;
     }
 
+    if (commandOptions.in) {
+      commandOptions.in = path.relative(this.project.root, commandOptions.in);
```
rwjblue: This is still a relative path, can you tweak this to match the logic in generate?

Issue #7998, opened 2018-08-28
Package Info tests are failing on windows


twokul:
https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/package-info-cache-test.js#L185

```sh
 1) models/package-info-cache.js
      packageInfo
        valid project
          validates projectPackageInfo:
    AssertionError: expected undefined to exist
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:214:38)

 2) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 0 errors:
    TypeError: Cannot read property 'hasErrors' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:220:35)

 3) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 1 dependencyPackage:
    TypeError: Cannot read property 'dependencyPackages' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:225:53)

 4) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 82devDependencyPackages:
    TypeError: Cannot read property 'devDependencyPackages' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:233:56)

 5) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 1 in-repo addon named "ember-super-button":
    TypeError: Cannot read property 'inRepoAddons' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:240:47)

 6) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 7 internal addon packages:
    TypeError: Cannot read property 'internalAddons' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:249:49)

 7) models/package-info-cache.js
      packageInfo
        valid project
          shows projectPackageInfo has 7 node-module entries:
    TypeError: Cannot read property 'nodeModules' of undefined
     at Context.<anonymous> (tests\unit\models\package-info-cache-test.js:256:46)
```

twokul:
likely realted to https://github.com/ember-cli/ember-cli/issues/7981

twokul:
cc @dcombslinkedin

dcombslinkedin:
[...]

PR #8002, opened 2018-08-29
fixed package-info-cache-test to use ${path.sep} instead of / to work on Windows

dcombslinkedin:


tests/unit/models/package-info-cache-test.js
```diff
@@ -207,7 +208,7 @@ describe('models/package-info-cache.js', function() {
         project = fixturifyProject.buildProjectModel(Project);
         project.discoverAddons();
         pic = project.packageInfoCache;
-        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}/simple-ember-app`);
+        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}${path.sep}simple-ember-app`);
```
Turbo87: we should probably use `path.join(fixturifyProject.root, 'simple-ember-app')` instead

stefanpenner:
getEntry should normalize it’s input, we should not leak platform specific specificier details to the caller.

Instead we should mimic what all fs functions do, and normalize

dcombslinkedin:
@stefanpenner , you're probably right about getEntry, but that also means that the cache keys (paths) need to be normalized on the way in, in several places. I can do it, but it's a bit of work. I'll file a ticket on myself to support that.

scalvert:
This is blocking other PRs. Do you have an ETA for this fix? Asking for a friend... :)

twokul:
@scalvert https://github.com/ember-cli/ember-cli/issues/7998#issuecomment-417422371 😝

PR #8003, opened 2018-08-29
Use `path.separator` instead of `/`

twokul:
TL;DR: paths are different on *nix and Windows systems. In order to
successfully run tests on both platforms, we should use
`path.separator`.

Original fix landed in beta via https://github.com/ember-cli/ember-cli/pull/8002

### DIFF
--- a/tests/unit/models/package-info-cache-test.js
+++ b/tests/unit/models/package-info-cache-test.js
@@ -203,7 +203,7 @@ describe('models/package-info-cache.js', function() {
         project = fixturifyProject.buildProjectModel(Project);
         project.discoverAddons();
         pic = project.packageInfoCache;
-        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}/simple-ember-app`);
+        projectPackageInfo = pic.getEntry(path.join(fixturifyProject.root, 'simple-ember-app'));
       });
 
       after(function() {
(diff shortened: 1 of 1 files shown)


stefanpenner:
We should not leak path.sep here, please instead update the functions you pass these file paths to to normalize their input. 

That way they work as other fs functions do

tests/unit/models/package-info-cache-test.js
```diff
@@ -203,7 +203,7 @@ describe('models/package-info-cache.js', function() {
         project = fixturifyProject.buildProjectModel(Project);
         project.discoverAddons();
         pic = project.packageInfoCache;
-        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}/simple-ember-app`);
+        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}${path.separator}simple-ember-app`);
```
Turbo87: we should probably use `path.join(fixturifyProject.root, 'simple-ember-app')` instead

stefanpenner:
@twokul / @dcombslinkedin hows the [proper fix](https://github.com/ember-cli/ember-cli/pull/8003#issuecomment-416961693) coming?

twokul:
@stefanpenner under water at work + cli release 3.4; afaik we would need to change the key for `entries` map (since we use paths as keys.) I won't be able to work on this this week

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/8015

stefanpenner:
don't worry about: `cli release 3.4` @rwjblue or myself can handle that.

PR #8005, opened 2018-08-31
[BUGFIX] Continue to use jQuery in Ember 2.x scenarios

cibernox:
It turns out the default blueprint for addons generates an addon whose 2.X scenarios in ember-try fail by default.

In https://github.com/ember-cli/ember-cli/commit/ca68ba3a5c5f8781a3650b553cf2050afc8b69f7#diff-d2dc84345b41f0668236a56410bcc7bc addons began to be tested without jQuery by default, which is great, but in Ember 2.X we **also** need to install the `ember-native-dom-event-dispatcher`.

I still suspect this may not help the tests for Ember 2.12, but let's start from here.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -12,25 +12,25 @@ module.exports = function() {
       <% if (yarn) { %>useYarn: true,
       <% } %>scenarios: [
         {
-          name: 'ember-lts-2.12',
-          npm: {
-            devDependencies: {
-              'ember-source': '~2.12.0'
-            }
-          }
-        },
-        {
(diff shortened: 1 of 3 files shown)


rwjblue:
This is looking good, I'd like to get this targetted at `beta` branch (which will require you to update the PR target *and* drop out any additional commits that are on master where you forked from and beta).

Also, can you confirm if we do or do not need to also add `@ember/jquery` to each of these 2.x scenarios? It seems to me that we _should_ have it in there, but I just wanting to make sure we don't emit warnings about it during these try scenarios...

cibernox:
@rwjblue I added the `@ember/jquery` addon to those scenarios to silence the warning (although I think it would work anyway).

Rebased over beta.

PR #8008, opened 2018-09-01
Revert "Allow specifying ENV variables in .ember-cli"

### LINKED ISSUE #7980: Breaking changes to meaning of `environment`
The [new feature that allows you to put environment variables in .ember-cli](https://github.com/ember-cli/ember-cli/pull/7910) has some very surprising side effects:

1. At runtime, `require('your-app/config/environment').default.environment` changes from a string like `"development"` to a POJO containing the settings from your `.ember-cli`.

2. Custom ember-cli commands that previously accepted their own option named `environment` are now getting stomped with the settings from `.ember-cli`. Here is [an example](https://github.com/cardstack/cardstack/blob/8ec12aa1f2bb2c88496d302336586d2a74dbac7d/packages/hub/commands/start-native.js#L11-L15) of a command that breaks if you try to use the new feature.

astronomersiva:
This reverts #7910.

Fixes #7980.

Apologies to everyone who got bit by this.


rwjblue:
> Apologies to everyone who got bit by this

No worries! And I think we still want to bring it back, we just need to revert while we figure out the right path forward...

NullVoxPopuli:
@rwjblue can we just use "environmentVariables" or "envVars" as the key in the .ember-cli file?

PR #8010, opened 2018-09-03
[FEATURE] Namespace(in subfolder) MU addon styles

patricklx:
This is to prevent style file name clashes between MU addons, since with MU addons we dont expect it to be merged with the main app. 

Therefore the style files will be moved into a subdir named after their addon.

Also Improved the tests to have more real behaviour. Which I extracted into a separate PR: #8146

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1073,19 +1073,20 @@ class EmberApp {
    * @return {BroccoliTree}
   */
   getStyles() {
-    let styles;
+    let styles = [];
     if (this.trees.styles) {
-      styles = new Funnel(this.trees.styles, {
-        srcDir: '/',
-        destDir: '/app/styles',
-        annotation: 'Funnel (styles)',
-      });
+      styles.push(this.trees.styles);
     }
     let addons = this.addonTreesFor('styles');
-
-    return mergeTrees(addons.concat(styles), {
+    styles = styles.concat(addons);
+    let mergedAddonStyles = mergeTrees(styles, {
       overwrite: true,
-      annotation: 'Styles',
+      annotation: 'Addon Styles',
+    });
+
+    return new Funnel(mergedAddonStyles, {
+      allowEmpty: true,
+      destDir: isExperimentEnabled('MODULE_UNIFICATION') ? 'src/ui/styles' : `app/styles`,
     });
   }
 
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -216,6 +216,10 @@ let addonProto = {
       public: 'public',
     };
 
+    if (this.isModuleUnification()) {
+      this.treePaths.styles = 'src/ui/styles';
+    }
+
     this.treeForMethods = defaultsDeep({}, DEFAULT_TREE_FOR_METHODS);
 
     if (this.parent) {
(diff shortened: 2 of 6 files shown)


patricklx:
requesting review @wycats  :)
seems similar to https://github.com/ember-cli/ember-cli/pull/8013

patricklx:
@Turbo87  I saw you added this to the Module Unification Project. 
The tests are passing in travis, but appvoyer has problems... can you have a look at appvoyer, or restart them? Seems to be a problem with npm...

patricklx:
Or @twokul , anyone?

Turbo87:
sorry, I have no idea about this feature 😞

patricklx:
mmm, maybe `enhancement` is more correct.  I guess I should split it into `bugfix` and another PR for `enahncement`
But since MU addons will be namespaces, the developer will not expect any nameclashes with other addons, right? 
But currently it can still happen with styles

lib/broccoli/ember-app.js
```diff
@@ -1073,19 +1073,20 @@ class EmberApp {
...
+    styles = styles.concat(addons);
+    let mergedAddonStyles = mergeTrees(styles, {
       overwrite: true,
-      annotation: 'Styles',
+      annotation: 'Addon Styles',
+    });
+
+    return new Funnel(mergedAddonStyles, {
+      allowEmpty: true,
+      destDir: isExperimentEnabled('MODULE_UNIFICATION') ? 'src/ui/styles' : `app/styles`,
```
ppcano: In other places is checking for the existence of the `src` folder like:

`isExperimentEnabled('MODULE_UNIFICATION') && this.trees.src`

lib/models/addon.js
```diff
@@ -873,15 +877,37 @@ let addonProto = {
     @return {Tree} The return tree has the same contents as the input tree, but is moved so that the `app/styles/` path is preserved.
   */
   treeForStyles(tree) {
+
+    if (!this.name || this.name === 'superFn') {
```
ppcano: I am not sure if this change belongs to the scope of this PR.

tests/unit/models/addon-test.js
```diff
@@ -735,17 +735,33 @@ describe('models/addon.js', function() {
 
   describe('treeForStyles', function() {
     let builder, addon;
+    const { isExperimentEnabled } = require('../../../lib/experiments');
```
ppcano: Can this `require` statement be moved to the import section at the top of file?

tests/unit/models/addon-test.js
```diff
@@ -735,17 +735,33 @@ describe('models/addon.js', function() {
...
+        if (isExperimentEnabled('MODULE_UNIFICATION')) {
+          return true;
+        }
+      };
 
       let BaseAddon = Addon.extend({
         name: 'test-project',
         root: projectPath,
+        isModuleUnification() {
+          const { isExperimentEnabled } = require('../../../lib/experiments');
```
ppcano: Can this `require` statement be moved to the import section at the top of file?

tests/unit/models/addon-test.js
```diff
@@ -757,19 +773,25 @@ describe('models/addon.js', function() {
...
 
       return builder.build()
         .then(function(results) {
           let outputPath = results.directory;
 
           let expected = [
-            'app/',
-            'app/styles/',
-            'app/styles/foo-bar.css',
+            'foo-bar.css',
```
ppcano: It looks that some of the existing tests (like this line) have changed and I found the same situation in other tests.

I don't know much about the internal of the ember-cli project, but I think the `MU addon style` feature should not affect Classic addons or apps, hence I am not sure if it is convenient to change the existing behaviour of these tests when running without enabling `MODULE_UNIFICATION`.

ppcano:
@patricklx 

I am looking at MU app/addon styles and I found this PR, so I have commented it a bit. 

>  I guess I should split it into bugfix and another PR for enahncement


I know this PR also fixes #8133 (fixed by #8134), what other cases does this PR fix?

Turbo87:
closing due to https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html

PR #8013, opened 2018-09-05
Fix SASS compilation issues

twokul:
There are several ways add-on can add a style to the application:

+ create `app/styles` folder
+ create `addon/styles` folder
+ return a tree from `treeForStyles` hook

Previously, some add-ons (when using `treeForStyles` hook) would return
content under `app/styles`. That used to work before packager changes.

Because there's no contract (interface) that Ember CLI provides when you return
a tree from `treeForStyles` hook, packager treats it as an opaque tree
and moves its contents later on the "right" place. It unfortunately
creates scenarios as such:

```
the-best-app-ever/
  styles/
    app/
      styles/
```

where nested `app/styles` is a tree from an add-on.

While I believe add-ons should return content under the "root" of the
tree, we do need to support current behaviour w/o breaking anything.

Useful links:

+ `app/styles` folder in
[`ember-basic-dropdown`](https://github.com/cibernox/ember-basic-dropdown/tree/v1.0.3/app/styles)
+ `treeForStyles` hook in
[`ember-cli-tailwind`](https://github.com/embermap/ember-cli-tailwind/blob/v0.6.1/index.js#L67)
+ `treeForStyles` hook in
[`ember-bootstrap`](https://github.com/kaliber5/ember-bootstrap/blob/v0.11.3/index.js#L54)

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -1072,12 +1072,26 @@ class EmberApp {
    * @return {BroccoliTree}
   */
   getAddonStyles() {
-    let mergedAddonStyles = mergeTrees(this.addonTreesFor('styles'), {
+    let addons = this.addonTreesFor('styles');
+
+    let flatten = addons.map(tree =>
+      new Funnel(tree, {
+        getDestinationPath(relativePath) {
+          let stylesDirExists = relativePath.startsWith('app/styles');
+
+          if (stylesDirExists) {
+            return relativePath.replace('app/styles/', '');
+          }
+
+          return relativePath;
+        },
+      })
+    );
+    let merged = mergeTrees(flatten, {
       overwrite: true,
       annotation: 'Addon Styles',
     });
-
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -1072,12 +1072,26 @@ class EmberApp {
    * @return {BroccoliTree}
   */
   getAddonStyles() {
-    let mergedAddonStyles = mergeTrees(this.addonTreesFor('styles'), {
+    let addons = this.addonTreesFor('styles');
+
+    let flatten = addons.map(tree =>
+      new Funnel(tree, {
```
kellyselden: Can you add comments in here about stuff to deprecate as people are doing things we don't want?

rwjblue:
Merging (so we can get the release out), @twokul would you mind following up with the comments that @kellyselden mentioned/

twokul:
@rwjblue absolutely

PR #8015, opened 2018-09-05
Align PackageInfoCache platform specific path usage with the rest of …

stefanpenner:
…the code-base.

In general, ember-cli avoids unnecessary path normalization.
Typically, passing user input un-normalized to fs methods, the exceptions
are when we perform string mutation or path comparisons.

The goal here, is to ensure all API’s can handle both posix and win32 paths,
and that internal data-structures are consistent regardless of which platform
they are run on.

Related #8006 #8002 #7998

### DIFF
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -665,7 +665,8 @@ class PackageInfoCache {
    *  to read the package.json from and process the contents and create a
    *  new cache entry or entries.
    */
-  _readNodeModulesList(nodeModulesDir) {
+  _readNodeModulesList(_nodeModulesDir) {
+    let nodeModulesDir = path.normalize(_nodeModulesDir);
     // Much of the time, normalizedNodeModulesDir is already a real path (i.e.
     // fs.realpathSync will return the same value as normalizedNodeModulesDir, if
     // the directory actually exists). Because of that, we'll assume
@@ -687,7 +688,7 @@ class PackageInfoCache {
 
     // realPath may be different than the original normalizedNodeModulesDir, so
     // we need to check the cache again.
-    if (realPath !== path.normalize(nodeModulesDir)) {
+    if (realPath !== nodeModulesDir) {
       nodeModulesList = this.getEntry(realPath);
       if (nodeModulesList) {
         return nodeModulesList;
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -67,9 +67,10 @@ function getRealFilePath(filePath) {
  */
 function getRealDirectoryPath(directoryPath) {
   let realPath;
+  let normalizedDirectoryPath = path.posix.normalize(directoryPath);
 
   try {
-    realPath = realDirectoryPathCache[directoryPath];
+    realPath = realDirectoryPathCache[normalizedDirectoryPath];
 
     if (realPath) {
       return realPath;
@@ -92,7 +93,7 @@ function getRealDirectoryPath(directoryPath) {
     }
   }
 
-  realDirectoryPathCache[directoryPath] = realPath;
+  realDirectoryPathCache[normalizedDirectoryPath] = realPath;
 
   return realPath;
 }
@@ -371,8 +372,8 @@ class PackageInfoCache {
    * @private
    * @method _addEntry
    */
-  _addEntry(path, entry) {
-    this.entries[path] = entry;
+  _addEntry(thePath, entry) {
+    this.entries[path.posix.resolve(thePath)] = entry;
   }
 
   /**
@@ -383,8 +384,8 @@ class PackageInfoCache {
    * @param {String} path the real path whose PackageInfo or NodeModulesList is desired.
    * @return {PackageInfo} or {NodeModulesList} the desired entry.
    */
-  getEntry(path) {
-    return this.entries[path];
+  getEntry(thePath) {
+    return this.entries[path.posix.resolve(thePath)];
   }
 
   /**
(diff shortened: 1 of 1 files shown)


stefanpenner:
likely some things to fix, will do so after lunch.

lib/models/package-info-cache/index.js
```diff
@@ -495,7 +496,7 @@ class PackageInfoCache {
    * @param {String} pkgDir the path of the directory to read the package.json from and
    *  process the contents and create a new cache entry or entries.
    * @param {Boolean} isRoot, for when this is to be considered the root
-   * package, whose dependencies we must all consider for discovery
+   * package, whos dependencies we must all consider for discovery
```
dcombslinkedin: Please fix this typo (again).

lib/models/package-info-cache/index.js
```diff
@@ -445,7 +446,7 @@ class PackageInfoCache {
 
       let nodeModulesPath = endsWithNodeModules
         ? currPath
-        : `${currPath}${path.sep}node_modules`;
+        : `${currPath}/node_modules`;
```
dcombslinkedin: Regardless of the other fixes, I *really* don't like this usage of / in the path. I think it was @twokul that suggested using path.join(currPath, 'node_modules) here instead. I think that's much more preferable. It doesn't mean the other changes to always convert to posix paths are not good, but explicitly relying on them by inserting / as a constant just seems wrong.

lib/models/package-info-cache/index.js
```diff
@@ -634,7 +635,7 @@ class PackageInfoCache {
       // directory from lower down in the project tree, and we want to get
       // the data into the cache ASAP. It may not necessarily be a 'real' error
       // if we find an issue, if nobody below is actually invoking the addon.
-      let nodeModules = this._readNodeModulesList(path.join(realPath, 'node_modules'));
+      let nodeModules = this._readNodeModulesList(`${realPath}/node_modules`);
```
dcombslinkedin: See my above comment - I think this *really* should use path.join().

tests/unit/models/package-info-cache-test.js
```diff
@@ -80,7 +80,7 @@ describe('models/package-info-cache.js', function() {
 
     beforeEach(function() {
       projectPath = path.resolve(addonFixturePath, 'simple');
-      packageJsonPath = path.join(projectPath, 'package.json');
+      packageJsonPath = path.normalize(`${projectPath}/package.json`);
```
dcombslinkedin: Third repetition of the same comment - I *really* believe that we should be using path.join() as a best practice.

tests/unit/models/package-info-cache-test.js
```diff
@@ -208,7 +208,7 @@ describe('models/package-info-cache.js', function() {
         project = fixturifyProject.buildProjectModel(Project);
         project.discoverAddons();
         pic = project.packageInfoCache;
-        projectPackageInfo = pic.getEntry(path.join(fixturifyProject.root, 'simple-ember-app'));
+        projectPackageInfo = pic.getEntry(`${fixturifyProject.root}/simple-ember-app`);
```
dcombslinkedin: Fourth repetition of the same comment from above about using path.join.

tests/unit/models/package-info-cache-test.js
```diff
@@ -255,7 +255,7 @@ describe('models/package-info-cache.js', function() {
 
         expect(inRepoAddons).to.exist;
         expect(inRepoAddons.length).to.equal(1);
-        expect(inRepoAddons[0].realPath).to.contain(path.join('simple-ember-app', 'lib', 'ember-super-button'));
+        expect(inRepoAddons[0].realPath).to.contain('simple-ember-app/lib/ember-super-button');
```
dcombslinkedin: Yet again on path.join().

stefanpenner:
@dcombslinkedin in general we author in posix and normalize explicitly and only when actually required.

This ensures we don't mutate input unnecessarily, but also ensures we explicitly normalize when we our algorithms must.

Can you help me understand what `path.join` as a best practice achieves?

dcombslinkedin:
@stefanpenner, when you say 'in general we author in posix', to me that says "I'm allowed to use something other than posix, since it's not a hard and fast requirement, it's "in general"".  So I could use \ if I'm coming from a Windows background and just providing test data for something. 

You say "normalize explicitly and only when actually required". I agree with that - if it is possible that somebody could construct a non-normalized path, it's incumbent on the code to handle that case. 

To me a best practice here would be "rather than confuse later readers of the test code by using / or \ explicitly and making them guess whether the code underneath will take care of normalizing (it will, but they may not know that), use path.join(), which gets you two things: (a) it's immediately obvious that the code is trying to allow for either style of path, but the user need not care and (b), the insertion of the separator is handled by a well-known method, path.join(), rather than by the somewhat less well-known method (the Ember-cli code inside the package-info-cache). 

Make sense? 

I don't feel strongly enough about this to jump up and down and throw a fit or block a merge (even if I could), but I do think the points above are reasonable.

stefanpenner:
> @stefanpenner, when you say 'in general we author in posix', to me that says "I'm allowed to use something other than posix, since it's not a hard and fast requirement, it's "in general"". So I could use \ if I'm coming from a Windows background and just providing test data for something.

Yes, consumers of ember-cli can write `\` or `/` but ember-cli's code-base is to be author by default with a posix style, in scenarios were we must test `\` or "current platform behavior" we do so explicitly.

tests/unit/models/package-info-cache-test.js
stefanpenner: this one I am fine with changing.

lib/models/package-info-cache/index.js
stefanpenner: In internal code, I prefer to leave input as-is, until the normalization is actually required. This avoids wack-a-mole, as only code which cares always normalizes before utilization.

lib/models/package-info-cache/index.js
stefanpenner: In internal code, I prefer to leave input as-is, until the normalization is actually required. This avoids wack-a-mole, as only code which cares always normalizes before utilization.

locks:
Closing as outdated.

PR #8024, opened 2018-09-06
[BUGFIX] Remove 2.12 scenario from travis.yml

cibernox:
In https://github.com/ember-cli/ember-cli/pull/8005 I removed 2.12 from `config/ember-try.js` but I forgot to remove it from the `.travis.yml` too.


cibernox:
Can someone re-run the scenario that timed out?

rwjblue:
This likely requires a 3.4.2, can you confirm?

Also, I updated to target `release` branch since I think #8005 landed in 3.4.x...

cibernox:
Yes, it does require a new patch release, otherwise running tests of a brand new addon will fail due to a missing scenario.

PR #8025, opened 2018-09-06
Update MU blueprint to latest canary

CodingItWrong:
Updates to the latest canary build as of today.

This fixes an error using angle bracket invocation syntax with a Module Unification app. Previously, trying to invoke a component `<MyComponent />` gives the error "Assertion Failed: You cannot use 'MyComponent' as a component name. Component names must contain a hyphen."


rwjblue:
Awesome, thank you!

PR #8032, opened 2018-09-10
Update minimum broccoli-viz version

gandalfar:
Update to current stable of broccoli-viz to ensure that it correctly parses output.


Turbo87:
thanks :)

PR #8044, opened 2018-09-12
Fix livereload failure with ssl enabled and Add a test for livereload with prefix

SparshithNR:
This commit has three tasks

- fix live reload issue with ssl enabled.

- Refactor livereload server creation.

- Add a test for livereload prefix


rwjblue:
I changed the base branch to release, and will rebase out the extraneous commits soon...

rwjblue:
😩 apparently soon took longer than I thought, finally got it going.

Should be able to land it once CI is happy...

Issue #8040, opened 2018-09-12
Build failure without `tests/` directory.


btecu:
`Windows 10` with `node 10` and `ember-cli 3.4.2` is causing the following error when trying to build:

```
ERROR Summary:

  - broccoliBuilderErrorStack: Error: ENOENT: no such file or directory, lstat 'D:\Ember\tmp\funnel-input_base_path-Y7gy2Rs4.tmp/tests/index.html'
    at Object.lstatSync (fs.js:805:3)
    at symlinkWindows (D:\Ember\node_modules\symlink-or-copy\index.js:126:25)
    at Function.symlinkOrCopySync [as sync] (D:\Ember\node_modules\symlink-or-copy\index.js:83:5)
    at Funnel._copy (D:\Ember\node_modules\broccoli-funnel\index.js:494:19)
    at Funnel.processFile (D:\Ember\node_modules\broccoli-funnel\index.js:477:8)
    at Funnel.applyPatch [as _applyPatch] (D:\Ember\node_modules\broccoli-funnel\index.js:390:12)
    at Funnel.<anonymous> (D:\Ember\node_modules\broccoli-funnel\index.js:344:10)
    at Array.forEach (<anonymous>)
    at Funnel.processFilters (D:\Ember\node_modules\broccoli-funnel\index.js:343:11)
    at Funnel.build (D:\Ember\node_modules\broccoli-funnel\index.js:259:10)
  - codeFrame: ENOENT: no such file or directory, lstat 'D:\Ember\tmp\funnel-input_base_path-Y7gy2Rs4.tmp/tests/index.html'
  - errorMessage: Build Canceled: Broccoli Builder ran into an error with `Funnel` plugin. 💥
ENOENT: no such file or directory, lstat 'D:\Ember\tmp\funnel-input_base_path-Y7gy2Rs4.tmp/tests/index.html'
  - errorType: Build Error
  - location:
    - column: [undefined]
    - file: [undefined]
    - line: [undefined]
    - treeDir: [undefined]
  - message: Build Canceled: Broccoli Builder ran into an error with `Funnel` plugin. 💥
ENOENT: no such file or directory, lstat 'D:\Ember\tmp\funnel-input_base_path-Y7gy2Rs4.tmp/tests/index.html'
  - name: Error
  - nodeAnnotation: Funnel (test index)
  - nodeName: Funnel
[...]

PR #8046, opened 2018-09-13
Do not to fail if `tests` folder is not present

twokul:
fixes https://github.com/ember-cli/ember-cli/issues/8040


runspired:
This issue appears to be back again in `3.9` (currently complicating my `ember-data` release efforts).

It complains about the lack of a `tests/` directory and if I add an empty one it complains that `tests/dummy/public` is empty. I will add open a new issue to report.

Issue #8039, opened 2018-09-12
[3.4] app.import type: 'test' + transformation is broken


lennyburdette:
Reproduction: https://github.com/lennyburdette/ember-3-4-import-bug

This import works fine, allowing `import sinon from 'sinon'`.

```js
app.import('node_modules/sinon/pkg/sinon.js', {
  using: [{ transformation: 'amd', as: 'sinon' }]
});
```

But if I add `type: "test"` so that the code is appended to test-support.js, it appears in the output without the transform.

It looks like `applyCustomTransforms()` is called only in [`processAppAndDependencies()`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/default-packager.js#L252), not in [`packageTestFiles()`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/default-packager.js#L1098).

cc @twokul @rwjblue

rwjblue:
Thanks for reporting @lennyburdette! I totally agree with your assessment.

rwjblue:
Fixed by #8047

lennyburdette:
Thanks!!

rwjblue:
fix released in v3.4.3

PR #8047, opened 2018-09-13
Fix transforms for tests

twokul:
fixes https://github.com/ember-cli/ember-cli/issues/8039


rwjblue:
I’m not sure I understand this, if we are calling apply custom transforms on multiple / various trees why would we memoize?

twokul:
@rwjblue at the point we call package, transforms map remains the same, that is we call it once, get processed trees and can memoize (no need to apply transforms again.) does that make sense?

Issue #8045, opened 2018-09-13
WebSockets stopped working in 3.4.2


boris-petrov:
I spent a few hours debugging this. Running `ember server --proxy http://something` on `3.4.2` when we have a CometD WebSocket connection leads to `'ws://127.0.0.1:4200/project/streaming' failed: Invalid frame header`. This is a bug in the Ember proxy and a pretty nasty regression. :)

rwjblue:
I would love to work on this and get it fixed in a 3.4.3 but I really need a reproduction to be able to dig in. Can _someone_ put together a demo that we can use?

boris-petrov:
I'll take a look tomorrow but I'm not sure how to create a simple reproduction without a server involved which might get complicated. Let me get back to you tomorrow if someone doesn't do it before me.

rwjblue:
Thank you!

boris-petrov:
@rwjblue - here is the repo: https://github.com/boris-petrov/ember-cli-websocket-bug

I'm using `cometd-nodejs-server` which actually doesn't support WebSockets (took me an hour of debugging and 3 seconds of reading the documentation to figure that out :) ) but the error on the FE side is still there.

1. Checkout the repo.
2. `npm install`
3. `cd websocket-server; node server.js`
4. In another terminal - `ember server`

You'll see the error (that's under Chrome. Firefox gives a different error message). If you checkout the next-to-last commit, you'll see it "working" (i.e. the server returning 400 for the websocket connection because it doesn't implement it). Hope this is enough. Please tell me if you need more information.

rwjblue:
Awesome, thank you for working on the reproduction. We’ll try to dig into it ASAP.

scottkidder:
Maybe this is useful for creating the reproduction?
https://websocket.org/echo.html

rwjblue:
Submitted #8057 to fix this regression.

westlywright:
@rwjblue Was attempting to upgrade this morning and I found the same issue. I applied your fix as a patch and WS started working again. Will keep an eye for the release with this fix before upgrading. Thanks!

rwjblue:
Fixed by #8057, will release shortly...

rwjblue:
fix released in v3.4.3

boris-petrov:
@rwjblue - thank you for the quick fix! Can I delete the reproduction repo? I guess you don't need it any more?

rwjblue:
yep, sounds good

PR #8057, opened 2018-09-19
Ensure livereload support does not break proxied websockets.

rwjblue:
Prior to this change the tiny-lr server would intercept and attempt to process **all** websocket upgrade attempts even though it should only process the requests for its own paths.

This fix ensures that we only defer to tiny-lr's websocket handling for livereload requests.

Fixes https://github.com/ember-cli/ember-cli/issues/8045.


stefanpenner:
@rwjblue since this is fixing a regression, we need to be doubly sure we don't regress. Can you add a test?

PR #8067, opened 2018-09-24
Less restrictive blueprints - change how addons are identified

scalvert:
The current logic used to evaluate whether a directory is an addon for the `--in` and `--in-repo-addon` options is fairly restrictive. It uses a project's discovered addons, a pre-identified list of direct addon depencencies, to evaluate whether one can run blueprints against a particular directory. This works well for most conventional cases, but in cases where a project uses an unconventional structure (I can hear you all saying "convention over configuration as I type this), such as a **monorepo**, this breaks down. 

In order to satisfy current uses cases while allowing more unconventional uses cases, this PR proposes slightly altering the logic of how an in-repo addon is evaluated. It merely uses the heuristic of whether a directory _resembles_ an addon, specifically if its package.json contains the addon keyword. This allows both current, _conventional_ use cases to work in addition to _unconventional_ use cases. Everyone is happy; we all celebrate 🎉 .

Existing tests should cover this change in internal behavior.

### DIFF
--- a/tests/acceptance/in-option-generate-test.js
+++ b/tests/acceptance/in-option-generate-test.js
@@ -39,6 +39,15 @@ describe('Acceptance: ember generate with --in option', function() {
     return remove(tmproot);
   });
 
+  function removeAddonPath() {
+    let packageJsonPath = path.join(process.cwd(), 'package.json');
+    let packageJson = fs.readJsonSync(packageJsonPath);
+
+    delete packageJson['ember-addon'].paths;
+
+    return fs.writeJsonSync(packageJsonPath, packageJson);
+  }
+
   it('generate blueprint foo using lib', function() {
     // build an app with an in-repo addon in a non-standard path
     return initApp()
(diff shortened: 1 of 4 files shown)


scalvert:
Done!

tests/acceptance/in-option-generate-test.js
```diff
@@ -94,4 +94,19 @@ describe('Acceptance: ember generate with --in option', function() {
         })
       );
   });
+
+  it('generate blueprint foo using sibling path', function() {
+    // build an app with an in-repo addon in a non-standard path
+    return initApp()
+      .then(() => fs.mkdirp('../sibling'))
+      .then(() => generateUtils.inRepoAddon('../sibling'))
```
rwjblue: Can you double check what this does? I want to make sure that it _does not_ add `../sibling` to the current projects `package.json`’s `ember-addon.paths`. In other words, the `../sibling` dir should have **no relationship** with the current project...

scalvert: Good call. Fixing.

PR #8068, opened 2018-09-25
Enable BROCCOLI_2 and SYSTEM_TEMP experiments by default.

rwjblue:
* Refactored `isExperimentEnabled` to account for "default enabled" experiments
* Add ability to turn off default enabled experiments (by setting the environment variable to `false`)
* Add tests for `isExperimentEnabled`
* Enable `BROCCOLI_2` by default
* Enable `SYSTEM_TEMP` by default
* Updated to latest broccoli@2
* Refactor usage of experiments in `lib/models/builder.js`

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -27,9 +27,6 @@ class Builder extends CoreObject {
   constructor(options) {
     super(options);
 
-    this.broccoli2 = isExperimentEnabled('BROCCOLI_2');
-    this.systemTemp = isExperimentEnabled('SYSTEM_TEMP');
-
     this.setupBroccoliBuilder();
 
     this._instantiationStack = (new Error()).stack.replace(/[^\n]*\n/, '');
@@ -50,12 +47,12 @@ class Builder extends CoreObject {
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
 
     let broccoli, options = {};
-    if (this.broccoli2) {
+    if (isExperimentEnabled('BROCCOLI_2')) {
       broccoli = require('broccoli');
       let tmpDir;
 
       // If not using system temp dir, compatability mode with broccoli-builder, tmp in root
-      if (!this.systemTemp) {
+      if (!isExperimentEnabled('SYSTEM_TEMP')) {
         tmpDir = `${this.project.root}/tmp`;
         if (!fs.existsSync(tmpDir)) {
           fs.mkdir(tmpDir);
@@ -66,7 +63,7 @@ class Builder extends CoreObject {
       };
     } else {
       broccoli = require('broccoli-builder');
-      if (this.systemTemp) {
+      if (isExperimentEnabled('SYSTEM_TEMP')) {
         console.warn('EMBER_CLI_SYSTEM_TEMP only works in combination with EMBER_CLI_BROCCOLI_2');
       }
     }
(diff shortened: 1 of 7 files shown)


oligriffiths:
Cool stuff. I wanna get broccoli to throw a specific error for the read/rebuild api usage that can be caught by ember-cli and fall back to broccoli-builder. This should provide a goodbye BC story

lib/models/builder.js
```diff
@@ -27,9 +27,6 @@ class Builder extends CoreObject {
   constructor(options) {
     super(options);
 
-    this.broccoli2 = isExperimentEnabled('BROCCOLI_2');
-    this.systemTemp = isExperimentEnabled('SYSTEM_TEMP');
```
Turbo87: out of curiosity, what's the reason for this change?

Turbo87:
looks like CI in unhappy:
```
Uncaught ReferenceError: heimdall is not defined at http://localhost:7357/assets/vendor.js, line 68178
```

lib/models/builder.js
rwjblue: It was something that I asked for in the original Broccoli 2 PR, but it didn't get done there and I didn't think it was blocking.

Basically, this makes removing the flag easier (just search for `isExperimentEnabled('BROCCOLI_2')` and remove all the conditionals), but you are totally correct that it isn't _required_...

Turbo87: okay, makes sense, thanks

rwjblue:
> looks like CI in unhappy:

ya, will be digging in to see why

package.json
```diff
@@ -40,7 +40,7 @@
     "babel-plugin-transform-es2015-modules-amd": "^6.24.1",
     "bower-config": "^1.3.0",
     "bower-endpoint-parser": "0.2.2",
-    "broccoli": "^2.0.0-beta.3",
+    "broccoli": "^2.0.0-beta.4",
```
stefanpenner: I think we spoke, we should likely cut stable so that we don't need to rely on beta's. Doing so is likely a pre-req to enabling this feature by default.

sdhull:
I'm not exactly _certain_ that this is related, but my team and I frequently get the following error after waking our computers up from sleep and trying to resume development:
```
Uncaught ReferenceError: module is not defined
   at index.js:1963
loader.js:247 Uncaught Error: Could not find module `ember-resolver` imported from `client/resolver`
   at missingModule (loader.js:247)
   at findModule (loader.js:258)
   at Module.findDeps (loader.js:168)
   at findModule (loader.js:262)
   at Module.findDeps (loader.js:168)
   at findModule (loader.js:262)
   at requireModule (loader.js:24)
   at app-boot.js:3
```

The only way we've found to fix this error is to restart our computers. We're on ember-cli `3.5.1`

oligriffiths:
You may want to open a new issue for this. I don’t think it’s related to Broccoli 2 afaict. Looks like a node error.

sdhull:
@oligriffiths thanks, I will. I posted here because it seemed to start after we updated to 3.5.0 (which I believe is the first version that SYSTEM_TEMP was enabled by default), and the fact that restarting is the only fix we've found also seems to point to SYSTEM_TEMP as the culprit.

If I could find a safe pattern for clearing broccoli artifacts from `/tmp`, I feel like that would be worth a shot, but it doesn't strike me as safe to clear out wholesale.

Do you think I should open a new issue on this repo or in broccoli directly? It's a tough issue to repro, it only happens once in awhile. But needing to restart the machine to fix it is fairly onerous, so we'd love to see it get fixed.

This seems to be related: https://github.com/broccolijs/broccoli/issues/144

oligriffiths:
@sdhull Have you tried `EMBER_CLI_BROCCOLI_2=false ember build`
This will revert back to the old broccoli 0.x series, if the issue persists, then it's not a Broccoli 2 issue.

sdhull:
@oligriffiths excellent suggestion, why didn't I think of that?!

We all just updated node from 8.x to 10.x however we are still occasionally getting this error after waking our computers up from sleep. Next time it happens to me I'll try setting that flag in combination with `ember serve` to see if that resolves it. Thanks!! 🤞

sdhull:
OK well @oligriffiths (and any intrepid googlers) wanted to give you an update:

I set `EMBER_CLI_SYSTEM_TEMP=false` in my zshrc last week. Just now I got the dreaded `ember-resolver` error. I tried `rm -rf tmp/ dist/` and restarting my server but it didn't help. I tried `rm -rf tmp dist` and then tried running `EMBER_CLI_BROCCOLI_2=false ember serve` and still no luck.

Soooo `ember-resolver` issue remains elusive and unexplained. Seems unrelated to node (as I've updated to node 10.x since I started getting this). Also it's apparently unrelated to `SYSTEM_TEMP` and `BROCCOLI_2`, so I'm at a loss as to how to proceed.

It only comes up maybe once or twice a week, and is impossible to repro on-demand so I guess it's likely to remain a mystery.

timiyay:
I've opened an issue about "build errors that go away after reboot" here: https://github.com/ember-cli/ember-cli/issues/8436

oligriffiths:
@sdhull thanks for the feedback, good to know it’s not a Broccoli 2 issue.

PR #8069, opened 2018-09-25
Adding back feature to customization of serveURL

SparshithNR:
Adding back the feature from #7936 which got removed by ccfa432


rwjblue:
Updated to target `release` and rebased.

PR #8080, opened 2018-09-29
update @ember/optional-features in blueprint

kellyselden:


rwjblue:
Thank you! We should double check that the rest of things are ok, but this seems like a good start...

rwjblue:
Needs a rebase to get CI passing again I believe.

kellyselden:
This is solved.

PR #8083, opened 2018-10-01
Catch InvalidNodeError for Broccoli 2.0 and fallback to broccoli-builder

oligriffiths:
As the title says. This should provide a decent level of BC for users with old plugins.

This can be tested with a new ember project, and yarn linking this branch of `ember-cli` into the new project.

Then `ember install ember-browserify` which uses the old read/rebuild API

Then `ember build` results in:

image

Now this might be *too much* output, perhaps we only need the `ERROR`, given that `broccoli-builder` also outputs the offending plugin.

Thoughts...

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -27,6 +27,8 @@ class Builder extends CoreObject {
   constructor(options) {
     super(options);
 
+    // Use Broccoli 2.0 by default, if this fails due to .read/.rebuild API, fallback to broccoli-builder
+    this.broccoliBuilderFallback = false;
     this.setupBroccoliBuilder();
 
     this._instantiationStack = (new Error()).stack.replace(/[^\n]*\n/, '');
@@ -40,11 +42,28 @@ class Builder extends CoreObject {
 
   /**
    * @private
+   * @method readBuildFile
+   * @param path The file path to read the build file from
+   */
+  readBuildFile(path) {
+    // Load the build file
+    let buildFile = findBuildFile('ember-cli-build.js', path);
+    if (buildFile) {
+      return buildFile({ project: this.project });
+    }
+
+    throw new SilentError('No ember-cli-build.js found.');
+  }
+
+  /**
+   * @private
    * @method setupBroccoliBuilder
    */
   setupBroccoliBuilder() {
     this.environment = this.environment || 'development';
     process.env.EMBER_ENV = process.env.EMBER_ENV || this.environment;
+
+    this.tree = this.readBuildFile(this.project.root);
 
     let broccoli, options = {};
     if (isExperimentEnabled('BROCCOLI_2')) {
@@ -61,20 +80,33 @@ class Builder extends CoreObject {
       options = {
         tmpdir: tmpDir,
       };
-    } else {
-      broccoli = require('broccoli-builder');
-      if (isExperimentEnabled('SYSTEM_TEMP')) {
-        console.warn('EMBER_CLI_SYSTEM_TEMP only works in combination with EMBER_CLI_BROCCOLI_2');
+
+      try {
+        this.builder = new broccoli.Builder(this.tree, options);
+        return;
+      } catch (e) {
+        // Catch here to trap InvalidNodeError. If this is thrown, it's because the node provided is not valid
+        // and likely uses the old .read/.rebuild API, so fallback to broccoli-builder that supports that API
+        if (
+          !(e instanceof broccoli.Builder.InvalidNodeError) ||
+          e.message.indexOf('The .read/.rebuild API is no longer supported as of Broccoli 1.0') === -1
+        ) {
+          throw e;
+        }
+
+        // Fallback to broccoli-builder
+        let error = `Invalid Broccoli2 node detected, falling back to broccoli-builder. Broccoli error:\n`
+        error += `---------------\n`;
+        error += e.message;
+        error += `---------------\n`;
+        this.ui.writeWarnLine(error);
       }
+    } else if (isExperimentEnabled('SYSTEM_TEMP')) {
+      this.ui.writeWarnLine('EMBER_CLI_SYSTEM_TEMP only works in combination with EMBER_CLI_BROCCOLI_2');
     }
 
-    let buildFile = findBuildFile('ember-cli-build.js', this.project.root);
-    if (buildFile) {
-      this.tree = buildFile({ project: this.project });
-    } else {
-      throw new SilentError('No ember-cli-build.js found.');
-    }
-
+    broccoli = require('broccoli-builder');
+    this.broccoliBuilderFallback = true;
     this.builder = new broccoli.Builder(this.tree, options);
   }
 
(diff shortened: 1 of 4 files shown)


lib/models/builder.js
```diff
@@ -64,18 +72,31 @@ class Builder extends CoreObject {
...
-      broccoli = require('broccoli-builder');
-      if (this.systemTemp) {
-        console.warn('EMBER_CLI_SYSTEM_TEMP only works in combination with EMBER_CLI_BROCCOLI_2');
+
+      try {
+        this.builder = new broccoli.Builder(this.tree, options);
+        return;
+      } catch (e) {
+        // Catch here to trap InvalidNodeError. If this is thrown, it's because the node provided is not valid
+        // and likely uses the old .read/.rebuild API, so fallback to broccoli-builder that supports that API
```
rwjblue: What other cases cause us to throw an `InvalidNodeError`?

lib/models/builder.js
```diff
@@ -64,18 +72,31 @@ class Builder extends CoreObject {
...
+        if (!e instanceof broccoli.Builder.InvalidNodeError) {
+          throw e;
+        }
+
+        // Fallback to broccoli-builder
+        console.log('');
```
rwjblue: We should use `this.ui` instead of `console` (comes from https://github.com/ember-cli/console-ui)...

lib/models/builder.js
```diff
@@ -64,18 +72,31 @@ class Builder extends CoreObject {
...
+        console.warn(chalk.red('ERROR: Invalid Broccoli2 node detected, falling back to broccoli-builder. Broccoli error:'));
+        console.warn(chalk.yellow('---------------'));
+        console.warn(chalk.yellow(e.message));
+        console.warn(chalk.yellow('---------------'));
+        this.broccoli2 = false;
```
rwjblue: I just removed these properties in beta branch, I'll merge upwards so that shows up on master in a few minutes

lib/models/builder.js
oligriffiths: https://github.com/broccolijs/broccoli-node-info/blob/master/index.js#L49

So I've updated the catch to check the message string, bit hacky but whatever

lib/models/builder.js
oligriffiths: Can you point me to an implementation? `this.ui` is undefined

oligriffiths:
I wonder if we should catch all exceptions as a result of setting up the broccoli 2 builder, for full fallback support. @stefanpenner @rwjblue thoughts?

oligriffiths:
As Stefan said, let's start with catching only the needed and dial it up from there.

lib/models/builder.js
```diff
@@ -46,13 +63,15 @@ class Builder extends CoreObject {
...
+    this.tree = this.readBuildFile(this.project.root);
+
+    let broccoli, options = {}, systemTemp = isExperimentEnabled('SYSTEM_TEMP');
     if (isExperimentEnabled('BROCCOLI_2')) {
       broccoli = require('broccoli');
       let tmpDir;
 
       // If not using system temp dir, compatability mode with broccoli-builder, tmp in root
-      if (!isExperimentEnabled('SYSTEM_TEMP')) {
+      if (!systemTemp) {
```
rwjblue: Please do not cache experiment values. Each branch point should call `isExperimentEnabled` directly.

lib/models/builder.js
```diff
@@ -61,20 +80,36 @@ class Builder extends CoreObject {
       options = {
         tmpdir: tmpDir,
       };
-    } else {
```
rwjblue: We still need this else case. We need to be able to disable the `BROCCOLI_2` experiment if there are significant issues.

lib/models/builder.js
```diff
@@ -61,20 +80,36 @@ class Builder extends CoreObject {
...
+        systemTemp = false;
       }
     }
 
-    let buildFile = findBuildFile('ember-cli-build.js', this.project.root);
-    if (buildFile) {
-      this.tree = buildFile({ project: this.project });
-    } else {
-      throw new SilentError('No ember-cli-build.js found.');
+    if (systemTemp) {
```
rwjblue: Please don’t cache the experiment values, always check `isExperimentEnabled`

lib/models/builder.js
rwjblue: Each of the tasks that instantiates this builder model (build, build-watch, serve) pass `ui`, it should_ be present. 

https://github.com/ember-cli/ember-cli/blob/2805962d1c9188ef36ca7f9abec360cbebbd47a6/lib/tasks/build.js#L15-L18

I’m not sure how it would end up undefined. Can you confirm that the environment that you were testing it in was setup correctly? I could imagine for example that some tests may not instantiate with the correct arguments...

tests/unit/models/builder-test.js
```diff
@@ -498,4 +502,42 @@ describe('models/builder.js', function() {
       });
     });
   });
+
+  if (isExperimentEnabled('BROCCOLI_2')) {
+    describe('fallback from broccoli 2 to broccoli-builder', function() {
+      it('falls back to broccoli-builder if an InvalidNode error is thrown for read/rebuild api', function() {
+        let spy = sinon.spy(console, 'warn');
```
rwjblue: This should be removed once the tests properly instantiate the builder with our MockUI object. You can reference other tests to see how to make assertions on the output when using it.

tests/unit/models/builder-test.js
```diff
@@ -498,4 +502,42 @@ describe('models/builder.js', function() {
...
+
+        const builder = new Builder({
```
rwjblue: This instance of the builder won’t have `this.ui` because it is not being constructed with it.

lib/models/builder.js
oligriffiths: I disagree. The `return` here https://github.com/ember-cli/ember-cli/pull/8083/files#diff-53c8b5d6578c6f2fb2869c4d31bdc3a1R86 will ensure this code path is not un, unless BROCCOLI_2 experiment is disable OR `.read/.rebuild` error is thrown during brocccoli 2 builder construction.

rwjblue: Awesome, thank you for pointing that out.

rwjblue:
I pushed a commit that replaces `console` usage with `console-ui` (and fixed up the tests that were not properly passing in the `ui` object).

Only thing left here is a "real world test".

rwjblue:
The only tests that exist for the fallback behavior are very mock heavy. We should be able to author a test that uses the builder model with a pipeline that _actually_ includes a node with the `.read` API, and confirm that the final build output is actually correct (along with the console output).

oligriffiths:
@rwjblue something like this? https://github.com/ember-cli/ember-cli/pull/8083/files#diff-32c532ce82c7b2f25aa8a1be7fa97672R259https://github.com/ember-cli/ember-cli/pull/8083/files#diff-32c532ce82c7b2f25aa8a1be7fa97672R259 https://github.com/ember-cli/ember-cli/pull/8083/files#diff-ffcb7262da504ca7263dec73783976b0

rwjblue:
Ya that test definitely ticks the box, but I was generally looking for a test that uses broccoli-test-helper directly. I migrated the test you added into that format and updated. 

This is ready to go once CI is happy...

oligriffiths:
@rwjblue nice, makes sense. Not sure what’s up with CI, 51 failures, not 100% sure what the issue is

rwjblue:
Ya, seems weird, I restarted CI...

rwjblue:
OK, I think I figured it out. `lib/utilities/find-build-file.js` does a `process.chdir` that I wasn't resetting after adding the new test (which left `process.cwd()` in `$TMPDIR`).

rwjblue:
Remaining CI failure is being tracked over in https://github.com/ember-cli/ember-cli/issues/8098. It does not seem related to this PR (the same test is failing in `beta` and `master` branches).

oligriffiths:
Ok cool. Do we need to wait until CI passes to merge then?

rwjblue:
Hopefully getting to passing today. Fell asleep last night doing a bisect (release is good, beta is bad), only 3 more steps to go though.

oligriffiths:
What a way to fall asleep 😂

PR #8059, opened 2018-09-19
Environment variables in ember cli file

NullVoxPopuli:
Iterates on https://github.com/ember-cli/ember-cli/pull/7910

Hopefully https://github.com/ember-cli/ember-cli/issues/7980 is still fixed

Somewhat reverts the revert: https://github.com/ember-cli/ember-cli/pull/8008 

:)

PR #8085, opened 2018-10-03
Temporarily update blueprint to reduce friction for contributors

NullVoxPopuli:
not having the environment flags set is a real "gotchya".
this does create a hard dependency on yarn -- but this is temporary.

These environment variables could move to .ember-cli when 
https://github.com/ember-cli/ember-cli/pull/8059 is merged


ppcano:
@NullVoxPopuli 

I wonder why the module unification app and addon blueprint has to use a particular ember-source and ember-cli@master.

https://github.com/ember-cli/ember-cli/blob/master/blueprints/module-unification-app/files/package.json


Shouldn't it mimic the behaviour of classic app blueprint?

https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json

I can create a PR for that

NullVoxPopuli:
> Shouldn't it mimic the behaviour of classic app blueprint?

I think it's just that sometimes things in MU break (since it's not even released). So the hashes specified are known to be working:tm:

blueprints/module-unification-app/files/package.json
```diff
@@ -11,18 +11,19 @@
     "test": "tests"
   },
   "scripts": {
-    "build": "ember build",
+    "ember": "EMBER_CLI_MODULE_UNIFICATION=true MODULE_UNIFICATION=true ./node_modules/.bin/ember",
```
kellyselden: this env setting style only works in bash. Windows will break.

kellyselden: cross-env might work for you here.

kellyselden:
You can use https://github.com/ember-cli/ember-cli/pull/8219 as a guide to support both yarn and npm.

NullVoxPopuli:
closed in favor of https://github.com/ember-cli/ember-cli/pull/8157 and the octane blueprint

PR #8086, opened 2018-10-03
Prefer walk-sync for AssetPrinterSize (speeds things up)

stefanpenner:


rwjblue:
Glob is attempting to match each file path against the “desired” matcher (tldr; regexp matching per “level”), but we really don’t need sophisticated matching at all. We just want all files, and a quick “endsWith” check.

Turbo87:
Hmm okay, that makes sense. I would have assumed that the regex time is small compared to the IO time, but maybe not. 

The other thing I'm wondering is why we need this class to be that fast. Is your app so huge that the asset size printing is starting to show perf problems?

Also feel free to merge, I'm just asking questions out of curiosity, not blocking anything

rwjblue:
> The other thing I'm wondering is why we need this class to be that fast. Is your app so huge that the asset size printing is starting to show perf problems?

tldr; this is mostly just a “drive by” cleanup that we noticed while debugging a gnarly production issue. 

We basically hit the node memory ceiling and were doing a bit of debugging on production build processes running in CI which caused us to step through this code again. Due to the memory pressure issues (which were not really related to the changes here) we noticed that the time we were spending in _just_ listing the files was quite a bit longer than we expected it to be (~ 15-20 seconds) so we briefly looked at what the code was doing and noticed that glob wasn’t really needed for us.

PR #8096, opened 2018-10-08
Use regex to parse /livereload urls

SparshithNR:
- Using strict `/livereload` matching is not very optimal.
- livereload takes `path` as an option which changes URL for upgrade requests. Ex `locahost:4200/path`. 
- LiveReload protocol insists `path` to end with `/livereload`. http://livereload.com/api/protocol/#basics
- The file `is-live-reload-request.js` can be an independent module like [clean-base-url](https://github.com/ember-cli/clean-base-url) which can be used by middlewares which listen for `upgrade` requests in ember-addons. This check is needed if we have listener for `upgrade` requests as we are now serving livereload and ember app on the same port.

### DIFF
--- a/lib/utilities/is-live-reload-request.js
+++ b/lib/utilities/is-live-reload-request.js
@@ -1,0 +1,6 @@
+'use strict';
+
+const cleanBaseUrl = require('clean-base-url');
+module.exports = function isLiveReloadRequest(url, liveReloadPrefix) {
+  return url === `${cleanBaseUrl(liveReloadPrefix)}livereload`;
+};
--- a/lib/utilities/is-live-reload-request.js
+++ b/lib/utilities/is-live-reload-request.js
@@ -1,6 +1,17 @@
 'use strict';
 
 const cleanBaseUrl = require('clean-base-url');
+const deprecate = require('./deprecate');
+
 module.exports = function isLiveReloadRequest(url, liveReloadPrefix) {
-  return url === `${cleanBaseUrl(liveReloadPrefix)}livereload`;
+  let regex = /\/livereload$/gi;
+  if (url === `${cleanBaseUrl(liveReloadPrefix)}livereload`) {
+    return true;
+  } else if (regex.test(url)) {
+    //version needs to be updated according to the this PR (https://github.com/ember-cli/ember-cli-inject-live-reload/pull/55)
+    //in master of ember-cli-inject-live-reload.
+    deprecate(`Upgrade ember-cli-inject-live-reload version to 1.10.0 or above`, true);
+    return true;
+  }
+  return false;
 };
--- a/lib/tasks/server/livereload-server.js
+++ b/lib/tasks/server/livereload-server.js
@@ -8,6 +8,7 @@ const logger = require('heimdalljs-logger')('ember-cli:live-reload:');
 const fs = require('fs');
 const Promise = require('rsvp').Promise;
 const cleanBaseUrl = require('clean-base-url');
+const isLiveReloadRequest = require('../../utilities/is-live-reload-request');
 
 function isNotRemoved(entryTuple) {
   let operation = entryTuple[0];
@@ -102,7 +103,7 @@ module.exports = class LiveReloadServer {
     // Reload on express server restarts
     this.app.on('restart', this.didRestart.bind(this));
     this.httpServer.on('upgrade', (req, socket, head) => {
-      if (req.url.indexOf('/livereload') === 0) {
+      if (isLiveReloadRequest(req.url, this.liveReloadPrefix)) {
         this.liveReloadServer.websocketify(req, socket, head);
       }
     });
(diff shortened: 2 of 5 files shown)


lib/utilities/is-live-reload-request.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+module.exports = function isLiveReloadRequest(url) {
+  let regex = /\/livereload$/gi;
```
rwjblue: This should receive the live reload path command line option, and use that to match absolute path. We should not need a regexp match here.

SparshithNR: There is no current command line option to take `livereload request path`, Are we going to provide a new one?

SparshithNR: Oh. Purpose of this option is different than what you pointed above. This path where `livereload server` is mounted with respect to `ember's express server`. This is the location from where `livereload.js` is served not the `upgrade` request. `livereload.js` running in the client(browser) is completely unaware of this option. If we are going to use this option as `path` to `livereload` requests, we have to set `host` and `port` in `ember-cli`  or in [ember-cli-inject-livereload](https://github.com/ember-cli/ember-cli-inject-live-reload/blob/master/index.js#L19) as well, because of the [this check](https://github.com/livereload/livereload-js/blob/master/dist/livereload.js#L314) in `livereload.js`.

SparshithNR:
- [ ]  Use the specified live reload host to avoid fuzzy matching with a regexp (just because the protocol requires it end with /livereload does not mean that every path in the appp ending in /livereload is for the client side live reload usage).
- [x] remove package-lock.json

lib/utilities/is-live-reload-request.js
rwjblue: > Purpose of this option is different than what you pointed above. This path where livereload server is mounted with respect to ember's express server. 

Yep, you are absolutely describing the current situation, I am saying that is wrong.

> livereload.js running in the client(browser) is completely unaware of this option.

Yep, I'm with you here. 

> If we are going to use this option as path to livereload requests, we have to set host and port in ember-cli or in ember-cli-inject-livereload as well, because of the this check in livereload.js.

We already do set the host and port in `ember-cli-inject-live-reload` [here](https://github.com/ember-cli/ember-cli-inject-live-reload/blob/68c237eca6cefc760a97ab30f37c215c84bd7fc7/index.js#L35). I agree that we need to properly set `path` in that query string as well (though it will need to use the current URL + the `liveReloadPrefix`).

---

The point of `liveReloadPrefix` is to scope **_all_** of the live reload related things including but not limited to:

* loading the `ember-cli-inject-live-reload.js` asset from the ember-cli express server
* loading the actual `livereload.js` asset from the tiny-lr server
* handling the upgrade requests from the live reload client

SparshithNR: Got it. Thank you for the detailed explanation. I will make the changes and update the PR as soon as possible.

SparshithNR:
This has a hard dependency over the version of `ember-cli-inject-live-reload` [PR](https://github.com/ember-cli/ember-cli-inject-live-reload/pull/55)

rwjblue:
> This has a hard dependency over the version of ember-cli-inject-live-reload PR

Can you help me understand this? What specifically makes it a hard dependency?

SparshithNR:
We are expecting upgrade requests to be **livereloadPrefix/livereload**. To achieve this livereload.js running in client must get **path** as an argument. [PR](https://github.com/ember-cli/ember-cli-inject-live-reload/pull/55) in `ember-cli-inject-live-reload` does that work and [PR#55](https://github.com/ember-cli/ember-cli-inject-live-reload/pull/55) needs this PR to be included in ember-cli so that **path** being set in livereload is properly interpreted in ember-cli `livereload-server.js`.

rwjblue:
Gotcha, thank you for the explanation! I think for now, we need to support both old and new worlds. If we receive a request for `/livereload` we should issue a deprecation instructing folks to upgrade to the new version of ember-cli-inject-live-reload. I _believe_ the new utility file is a great place for this, we could issue the deprecation and continue to allow upgrading...

SparshithNR:
Utility file is not accessible from `proxy-server` since it is an addon. We probably do something like `clean-base-url` module. Which imported and used in any project. So that if someone using the same logic to bypass `livereload` requests being consumed by other upgrade handlers, we can `is-live-reload-request` module. One place of truth for everyone so that in future if something changes we can just update this module.

rwjblue:
> Utility file is not accessible from proxy-server since it is an addon.

I don't fully understand what you mean. The utility **and** the proxy-server are both part of ember-cli, right? You should be able to require between the two without issue...

SparshithNR:
Node is able to resolve this. 
When I run yarn test, I get lint error
I added eslint-disable-next-line, Is it a good practice? 
1) eslint
       should have no errors in lib:
lib/tasks/server/middleware/proxy-server/index.js
  3:37  error  "../../../../utilities/is-live-reload-request" is not published  node/no-unpublished-require

rwjblue:
Ya, in this case I think disabling the rule inline is correct.  Those extra `package.json`'s make `node/no-unpublished-require` rule pretty confused (since in this case **all** of those files are actually published together, and these aren't separate packages).

PR #8100, opened 2018-10-09
Attempt to fix CI with npm@5 and npm@6.

rwjblue:
As of 2018-10-09 npm 5 and 6 _break_ the heirarchy of `node_modules` after a `npm install foo` (deletes files/folders other than `node_modules/foo` that were not created by the `npm` client itself)`

This ensures that we run a full `npm install` **after** any `npm install foo` funs to ensure that we have a fully functional `node_modules` heirarchy.

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -151,6 +151,20 @@ class NpmTask extends Task {
       } else {
         let args = this.toNpmArgs(this.command, options);
         promise = this.npm(args);
+
+        // as of 2018-10-09 npm 5 and 6 _break_ the heirarchy of `node_modules`
+        // after a `npm install foo` (deletes files/folders other than
+        // what was directly installed) in some circumstances, see:
+        //
+        // * https://github.com/npm/npm/issues/16853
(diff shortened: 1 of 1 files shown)


Turbo87:
> As of 2018-10-09 npm 5 and 6 _break_ the heirarchy of `node_modules` after a `npm install foo` (deletes files/folders other than `node_modules/foo` that were not created by the `npm` client itself)`

- `npm install`
- `npm install foo`
- `node_modules` only contains `foo` (+ subdeps) now?

do I understand that correctly?!

rwjblue:
See:

* https://github.com/npm/npm/issues/16853
* https://github.com/npm/npm/issues/17379

Turbo87:
can we add a link to this PR or to those issues to the code comment?

rwjblue:
Sure, sounds good. When I initially did this I guess I thought “everyone” was aware of these issues. Obviously, that’s still not a good reason not to explain in more detail in the comment...

rwjblue:
Updated with more detailed links in the inline comment.

PR #8111, opened 2018-10-12
Fix smoke test

dcombslinkedin:
The smoke test 'ember new foo, server, SIGINT clears tmp/' now fails locally for me because the 'tmp' directory does not exist inside the app root dir. After discussion with @rwjblue, changed the test to fail if tmp is present.


dcombslinkedin:
Closing as it contains part of another pending pull request. Will resubmit.

PR #8116, opened 2018-10-13
Only assert on fallback messaging when BROCCOLI_2 enabled.

rwjblue:
This fixes the current `beta` branch failure (the `EMBER_CLI_BROCCOLI_2=false` build).


rwjblue:
😐 apparently appveyor is not happy still

rwjblue:
Landing this for now, will continue to work on appveyor

PR #8120, opened 2018-10-15
Adjust release instructions

Turbo87:
some small things I noticed while doing the v3.5.0 release


Turbo87:
ignoring failing CI since this is only updating docs and CI is quite flaky unfortunately...

PR #8117, opened 2018-10-13
[BUGFIX RELEASE] [BUGFIX LTS] Do not ignore dotfiles in ESLint

Gaurav0:
Same as #8105 , but targeting release

PR #8124, opened 2018-10-15
Ignore node modules for eslint

anehx:
Since https://github.com/ember-cli/ember-cli/pull/8117 I had an issue where files in `/node_modules/.bin/` were linted - which is obviously incorrect. ESlint (https://eslint.org/docs/user-guide/configuring#eslintignore) states that node and bower modules are excluded by default (which would make the `bower_components` obsolete as well), however I think that this new ignore rule (`!.*`) overwrites that.

### DIFF
--- a/blueprints/app/files/.eslintignore
+++ b/blueprints/app/files/.eslintignore
@@ -8,6 +8,7 @@
 
 # dependencies
 /bower_components/
+/node_modules/
 
 # misc
 /coverage/
--- a/blueprints/module-unification-app/files/.eslintignore
+++ b/blueprints/module-unification-app/files/.eslintignore
@@ -8,6 +8,7 @@
 
 # dependencies
 /bower_components/
+/node_modules/
 
 # misc
 /coverage/


blueprints/app/files/.eslintignore
```diff
@@ -8,6 +8,7 @@
 
 # dependencies
 /bower_components/
+/node_modules/
```
kellyselden: or should we do `/!.*` below instead?

anehx: I'm not sure if that maybe breaks some things which were fixed in #8117. E.g if you had a `/tests/.foo.js` file..

Gaurav0: I think we need to add !/node_modules ?

anehx: @Gaurav0 Why would we want to **not** ignore `node_modules`?

Gaurav0: Sorry meant specifically add node_modules.

Gaurav0: I've included this in my PR and working on fixing all tests.

Turbo87:
closing in favor of #8127

hope that's okay for you @anehx 🙏

PR #8134, opened 2018-10-16
use module unification layout for styles

### LINKED ISSUE #8133: Fix addon styles concatenation
In 'MODULE_UNIFICATION' project, styles are added under `src/ui/styles`.

However, addons add their styles under `app/styles/`, which are not being copied to `src/ui/styles` and this is resulting in no stylesheet files in the final build.

raghuvaran:
Fixes #8133


ppcano:
I have created a test for this case at #8289

PR #8138, opened 2018-10-17
Remove ember-ajax from addon blueprint

### LINKED ISSUE #7760: Obligatory ember-ajax in add-on generator templates/blueprints?
Not sure what is the background of such a decision, but do we really need `ember-ajax` as a "starter" dependency for a new add-on setup?

Which part of it are we using for instance for UI components? If it's redundant, then I can help with removal of it  👍

initram:
Fixes #7760

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -59,6 +59,9 @@ module.exports = {
     // addons should test _without_ jquery by default
     delete contents.devDependencies['@ember/jquery'];
 
+    // ember-ajax depends on jquery, make it opt-in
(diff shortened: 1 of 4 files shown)


initram:
As far as I could see module-unification-addon inherits from addon in someway, and I could not find similar code to that of addon to remove ember-data and the like. So what I ended up doing was just removing ember-ajax from module-unification-addon test fixture file. I hope this works.

initram:
The builds failed, but I don't think it is caused by my changes. Do I have to do anything to make sure my changes can be merged, or will someone look through the build logs and decide if it is okay to merge?

PR #8139, opened 2018-10-17
Removes ember-welcome-page addon and adds it as part of app folder

abhilashlr:


abhilashlr:
Ah! looks like some test is failing. Will have a look at it

rwjblue:
I do not like the idea that a newly generated project has a bunch of stuff to delete in it...

rwjblue:
If anything, this should be a good topic for an RFC to decide. Perhaps I'm in the minority...

locks:
@rwjblue RFC seems good :) From a learning perspective it's much more explicit to see the actual component and be able to edit it than having a "magic" `{{welcome-page}}` that is introduced from somewhere. Regarding stuff to delete, it's a folder vs editing `application.hbs` 😅. Let's continue this discussion in the upcoming RFC.

rwjblue:
> Regarding stuff to delete, it's a folder vs editing application.hbs

This does not seem correct to me (even in MU land we don’t have any specific guidance on static public assets inside a component directory), but happy to have the detailed discussion in an RFC. 😺

PR #8142, opened 2018-10-17
Fix yarn test failures

### LINKED ISSUE #7484: Without `yarn` globally available, `npm test` fails

abhilashlr:
Fixes #7484

### DIFF
--- a/tests/helpers/acceptance.js
+++ b/tests/helpers/acceptance.js
@@ -4,6 +4,7 @@ const symlinkOrCopySync = require('symlink-or-copy').sync;
 const path = require('path');
 const fs = require('fs-extra');
 const runCommand = require('./run-command');
+const which = require('which');
 
 let root = path.resolve(__dirname, '..', '..');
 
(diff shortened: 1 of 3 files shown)


abhilashlr:
@Turbo87 I hope this is a better fix for solving yarn failures in test cases. Please review and let me know if there needs to be any changes done.

tests/helpers/acceptance.js
```diff
@@ -83,7 +84,9 @@ function linkDependencies(projectName) {
   let nodeManifest = fs.readFileSync(path.join(runFixture, 'package.json'));
 
   let packageCache = new PackageCache(root);
-  packageCache.create('node', 'yarn', nodeManifest, [{ name: 'ember-cli', path: root }]);
+  let packager = which.sync('yarn', { nothrow: true }) ? 'yarn' : 'npm';
```
Turbo87: I guess it would make sense to only run this once for the whole test suite. Can you extract the `which.sync('yarn', { nothrow: true })` code to e.g. `tests/helpers/has-yarn.js` and export the result from that file?

abhilashlr:
@Turbo87 done with the review comments.

tests/helpers/has-global-yarn.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const which = require('which');
+
+function hasGlobalYarn() {
+  return which.sync('yarn', { nothrow: true });
+}
+
+module.exports = hasGlobalYarn;
```
Turbo87: exporting a function here still means that we call `which` whenever we call the function. it would be best if we only did it once though.

I think something like 
```js
module.exports = which.sync('yarn', { nothrow: true });
```
should be sufficient here.

Turbo87: exporting a function here still means that we call `which` whenever we call the function. it would be best if we only did it once though.

I think something like 
```js
module.exports = which.sync('yarn', { nothrow: true });
```
should be sufficient here.

abhilashlr: That's a good one @Turbo87 👍 I've fixed it now.

abhilashlr:
Thanks @Turbo87 👍

PR #8143, opened 2018-10-18
adding fix to use NULL nodeModuleList to optimize when dealing with non-addons

dcombslinkedin:
courtesy of spenner, with some adjustments of my own.

### DIFF
--- a/lib/models/package-info-cache/error-list.js
+++ b/lib/models/package-info-cache/error-list.js
@@ -10,7 +10,7 @@ const ErrorEntry = require('./error-entry');
  * @protected
  * @class ErrorList
  */
-class ErrorList {
+module.exports = class ErrorList {
   constructor() {
     this.errors = [];
   }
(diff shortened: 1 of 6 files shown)


dcombslinkedin:
Fails in AppVeyor have nothing to do with the package-info-cache code. They're things like missing directories or files in other tests.

stefanpenner:
This lgtm. But I am just about to head on vacation so I’ll leave this to @rwjblue

dcombslinkedin:
Thanks, @stefanpenner. Have a good vacation!

rwjblue:
Updated to target `beta` and rebased.

dcombslinkedin:
Thanks, @rwjblue !

dcombslinkedin:
This is once again the same set of errors that have been showing up in many builds lately. None of the errors has anything to do with the changes to the code being checked in. @rwjblue, could you do the merge since it passes Travis?

PR #8146, opened 2018-10-19
Fix MU addon styles

### LINKED ISSUE #8133: Fix addon styles concatenation

patricklx:
Before , addon stlyes where taken from /app/styles and merged into <appname>/app/styles, instead of src/ui/styles in case of module unification.
Changed Funnel to change destination to src/ui/styles for MU.

Closes #8133

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -943,26 +943,7 @@ let addonProto = {
     @return {Tree} The return tree has the same contents as the input tree, but is moved so that the `app/styles/` path is preserved.
   */
   treeForStyles(tree) {
-    if (!tree) {
-      return tree;
-    }
-
-    const name = this.moduleName();
-
-    const addonStyles = new Funnel(tree, {
-      allowEmpty: true,
-      srcDir: this.treePaths.styles,
-      destDir: '.',
-    });
-
-    const customStyles = new Funnel(tree, {
-      exclude: [this.treePaths.styles],
-      destDir: '.',
-    });
-
-    return mergeTrees([addonStyles, customStyles], {
-      annotation: `Addon#treeForStyles (${name})`,
-    });
+    return tree;
   },
 
   /**
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -298,6 +298,10 @@ let addonProto = {
       public: 'public',
     };
 
+    if (this.isModuleUnification()) {
+      this.treePaths.styles = 'src/ui/styles';
+    }
+
     this.treeForMethods = defaultsDeep({}, DEFAULT_TREE_FOR_METHODS);
 
     if (this.parent) {
@@ -943,9 +947,21 @@ let addonProto = {
       return tree;
     }
 
-    return new Funnel(tree, {
-      destDir: 'app/styles',
-      annotation: `Addon#treeForStyles (${this.name})`,
+    const name = this.moduleName();
+
+    const addonStyles = new Funnel(tree, {
+      allowEmpty: true,
+      srcDir: this.treePaths.styles,
+      destDir: '.',
+    });
+
+    const customStyles = new Funnel(tree, {
+      exclude: [this.treePaths.styles],
+      destDir: '.',
+    });
+
+    return mergeTrees([addonStyles, customStyles], {
+      annotation: `Addon#treeForStyles (${name})`,
     });
   },
 
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -700,6 +700,7 @@ class EmberApp {
         let tree = addon.treeFor(type);
         if (tree && !mergeTrees.isEmptyTree(tree)) {
           sum.push({
+            addon,
             name: addon.name,
             tree,
             root: addon.root,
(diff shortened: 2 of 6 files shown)


simonihmig:
I tried this, and can confirm this is fixing all the current issues I had with importing (Sass) styles from a classic addon in a MU app:
* `ember-bootstrap`(https://github.com/kaliber5/ember-bootstrap/issues/648)
* `ember-component-css` (based on https://github.com/ebryn/ember-component-css/pull/294)

lib/models/addon.js
```diff
@@ -873,15 +877,32 @@ let addonProto = {
...
-    return new Funnel(tree, {
-      destDir: 'app/styles',
-      annotation: `Addon#treeForStyles (${this.name})`,
+    const name = this.moduleName();
+
+    const addonStyles = new Funnel(tree, {
+      allowEmpty: true,
+      srcDir: this.treePaths.styles,
+      destDir: '.',
+    });
```
simonihmig: It seems this is to support addons exposing the styles tree with a `app/styles` (or `src/ui/styles`) folder structure, while previously the Funnel was only taking styles exposed at the root level and moved them to the `destDir` (`only `app/styles` before). So basically you could expose `foo.scss` as `/foo.scss` or as `app/styles/foo.scss`, is that right?

Is so, why is that needed? As this seems to expand the public API contract of the `treeForStyles()` hook (though the public API of those hooks was always kinda fuzzy, AFAICT). Or is it just to support new MU addons?

Also does this rely on addons calling `super` in their `treeForStyles()`? Which some (many?) didn't so far:
* https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L193-L199
* https://github.com/embermap/ember-cli-tailwind/blob/master/index.js#L69-L77

lib/models/addon.js
```diff
@@ -873,15 +877,32 @@ let addonProto = {
     @return {Tree} The return tree has the same contents as the input tree, but is moved so that the `app/styles/` path is preserved.
   */
   treeForStyles(tree) {
+
+    if (!this.name || this.name === 'superFn') {
+      throw new Error('needs to call _super.treeForStyles.call(...)');
```
simonihmig: I guess the point is to correctly pass the `this` context, so maybe include that in your exception message? 

```suggestion
      throw new Error('needs to call _super.treeForStyles.call(this, tree)');
```

lib/models/addon.js
patricklx: Ah, what I wanted to achieve here was some way to support ember-component-css with addon components out of the box. Therefore I also wanted to merge in all styles, which are not in app/styles...

mmmm, I wasnt aware that many do not call super here... okay, I guess its not so a good idea to add important logic here.
Then i move the logic somewhere else.

patricklx: So, what about moving the logic to `getStyles?
https://github.com/ember-cli/ember-cli/pull/8146/files#diff-392bb207fea8bcbeb4a084bb445f7cc7R1077

nikdojo:
Hi,

is it possible to push this PR a little, because it's a really critical fix for MU implementation?
Thx.

lib/models/addon.js
patricklx: What i also saw is that root level was already merged into app/styles, so both ways would have worked before?
What i saw is that then there is an app/styles in app/styles folder. But will recheck

patricklx: Okay, so after reviewing.
the styles tree is build here 
https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L614
which takes the path from
https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L208

So, all files in <addon>/app/styles will be in the root, when in arrives at `treeForStyles`, but addons could overwrite it and merge in styles, either at root or to `app/styles`, then after calling `this._super.treeForStyles`, root level styles would be moved to the `app/styles`, and `app/styles`, would also still work, but will be duplicated. So to make sure that we dont get into that duplicate state, I dont move app/styles anymore, only root styles.

lib/models/addon.js
patricklx: I removed this part, to not require changes on addons

ppcano:
#8289 could be a duplicated of this one.

lolmaus:
@ppcano Hey Pepe! Have you tried this branch? Is it working for you?

We're stuck with this branch and seem to gradually fall behind `master`. I'm thinking of mentioning it in the MU epic issue.

patricklx:
I can rebase it again..., but should really be considered..

ppcano:
@lolmaus Hej!!

I have worked on this at #8289, and it looks ready, I'd appreciate if someone can test it as well. 

Both PRs look to solve the same issue. I found some inconsistencies of the broccoli styles tree of MU and Classic apps, and I ended up with a different implementation.

I am gonna ping the #st-module-unification channel to see if someone could review it.

ppcano:
#8289 has been finally merged and it should fix this issue.

Please, it would help if you test your apps using the latest `ember-cli` and report if your issue has been fixed.

cc @lolmaus @nikdojo @patricklx @simonihmig

simonihmig:
@ppcano just gave it a try, and works just perfectly! 🎉

Thanks a bunch for finally getting this fixed and merged!

rwjblue:
Thank you all, sorry for the long delays!

patricklx:
@rwjblue I tested the with the latest ember cli master. but unfortunately it's still not working for me.... :(
I have an MU addon in an MU project, and I do not get the styles from the addon.

Also my test here does not seem pass with the current maste branch:
https://github.com/ember-cli/ember-cli/compare/master...patricklx:fix-mu-styles#diff-1cef190fe22e6c889bb23fda5a8296a3R473

So I would like to reopen this.

NullVoxPopuli:
@ppcano I wonder if it's related to this: https://github.com/miguelcobain/ember-paper/issues/883#issuecomment-460339264

ppcano:
Re-opened at #8411

PR #8155, opened 2018-10-26
lint dummy node files slightly differently

kellyselden:
A single override for node files seems like not enough. We really have two groups of node files, those that are published on npm, and those that are not. For those that are not, it is OK to require modules that are only dev dependencies. This creates a second override which turns that rule off.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -48,6 +48,17 @@ module.exports = {
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
         // add your custom rules and overrides for node files here
       })<% } %>
-    }
+    }<% if (blueprint !== 'app') { %>,
+
+    // dummy node files
+    {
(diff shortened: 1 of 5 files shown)


blueprints/app/files/.eslintrc.js
```diff
@@ -48,6 +48,17 @@ module.exports = {
...
+    }<% if (blueprint !== 'app') { %>,
+
+    // dummy node files
+    {
+      files: [
+        'ember-cli-build.js',
+        'tests/dummy/config/**/*.js'
+      ],
+      rules: {
+        'node/no-unpublished-require': 'off'
```
rwjblue: This rule already does not run on files that are not published to `npm`, disabling the rule is probably not the right answer. We should instead fix our `.npmignore`s (and the rule is doing its job)...

kellyselden: You're right. Correcting the .npmignore made everything better.

PR #8156, opened 2018-10-26
fix: clean up readme noise

knownasilya:
Basically a new README is very noisy and distracting. I left the addon name line long since it could be a varied length, and programmatically making that line is more effort for sure.

Also added syntax highlighting to the install command


Turbo87:
I disagree with this change. If you look at the rendered result it will also have headings with a bottom border spanning the whole page, so this brings the textual representation closer to the rendered result. It also has the benefit of more clearly separating the sections.

PR #8157, opened 2018-10-26
change .ember-cli to .ember-cli.js

NullVoxPopuli:
now we don't need package.json shenanigans for MU

### DIFF
--- a/blueprints/module-unification-app/files/.ember-cli.js
+++ b/blueprints/module-unification-app/files/.ember-cli.js
@@ -1,4 +1,6 @@
-{
+process.env.EMBER_CLI_MODULE_UNIFICATION = true;
+
+module.exports = {
   /**
     Ember CLI sends analytics information by default. The data is completely
     anonymous, but there are times when you might want to disable this behavior.
--- a/blueprints/module-unification-app/files/.ember-cli.js
+++ b/blueprints/module-unification-app/files/.ember-cli.js
@@ -1,3 +1,5 @@
+'use strict';
+
 process.env.EMBER_CLI_MODULE_UNIFICATION = true;
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


blueprints/module-unification-app/files/.ember-cli.js
```diff
@@ -1,4 +1,6 @@
-{
+process.env.EMBER_CLI_MODULE_UNIFICATION = true;
```
rwjblue: Lets add `'use strict';` up here too....

ppcano:
@NullVoxPopuli This change could also be needed in the the MU addon blueprint.

https://github.com/ember-cli/ember-cli/blob/master/blueprints/module-unification-addon/files/.ember-cli

NullVoxPopuli:
@ppcano @rwjblue rebased and copied the .ember-cli.js to mu-addon.

questions re: octane 
- will there be a separate ocntane addon blueprint repo like there is for the octane app blueprint?
- should all the other .ember-cli files change over to .js files?

NullVoxPopuli:
Unsure. Afaik, remote blueprint packages have a rigid structure, where there is a single files directory at the repo root. I haven't looked at the blueprint resolution code yet, though... So.. idk if it'd be possible. It'd be fantastic if it were though. 
I'd also like to use variables and symlinks for a bunch of the duplicate stuff

Turbo87:
FWIW I'm not a fan of this change. The file currently can be statically parsed by e.g. IDE, which will no longer be the case once we allow JS. IMHO we need stronger reasons to change this than just making a temporary workaround work a little nicer.

rwjblue:
> The file currently can be statically parsed by e.g. IDE, which will no longer be the case once we allow JS

FWIW, we already allow `.ember-cli.js`.

locks:
Long live MU! Thanks for the contributions.

PR #7485, opened 2017-12-04
[WIP] Addons using module unification

mansona:
## Todo: 
- [ ] Dry up lib/broccoli/ember-addon.js (copied from lib/broccoli/ember-app.js)
- [ ] Process styles (in some way) from src/ in addon (models/addon.js)
- [ ] fix bad error message for `Missing template processor` when missing ember-cli-htmlbars in dependecies i.e. it exists in dev-dependencies
- [ ] broccoli-module-unification-reexporter [edge cases?](https://github.com/stonecircle/broccoli-module-unification-reexporter/blob/538d923e7e8a8f134369d8b0b8b9bf1473e99e49/test/index.js#L83)
- [ ] throw an error (or do something) when an addon can't decide if it's mu or classic (i.e. has src/ **and** addon/ or app/ folders)

## Tests:
- [ ]  _addonSrcTree - ember-app.js
- [ ] _processedVendorTree - test that it includes stuff from the addon src/ folder
- [ ] treeForSrc - models/addon.js
- [ ] addons with src/ folder in classic apps

PR #8158, opened 2018-10-27
Changes dummy app's detection and `project.isModuleUnification()`

NullVoxPopuli:
from the quest issue: https://github.com/emberjs/ember.js/issues/16373

specifically, this partially covers:
 -  Fix lib/models/project.js's isModuleUnification method to not only rely on presence of top level src as "the project is module unification format". Specifically, an addon may be in module unification format, but have its dummy app be in classic format.
 - Ember CLI uses the presence of src/ to configure the return value of isModuleUnification(). This works well for apps and addons. The detection fails for a classic dummy app in a module unification addon, however. Although a dummy app may be in tests/dummy/app/ with no tests/dummy/src/ directory, the addon's root level src/ confuses our logic and the dummy app is incorrectly treated as a module unification app.

proposed scheme for multiple dummy apps:

 - \<root\>/
   - tests/
     - dummy/
        - {app-name1}
           - a classic app
        - {app-name2}
           - a mu app

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -224,7 +224,7 @@ class Project {
   }
 
   isDummyApp() {
-    return this.has('dummy');
+    return this.root.includes('/dummy');
   }
 
   /**
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -223,6 +223,10 @@ class Project {
     }
   }
 
+  isDummyApp() {
+    return this.has('dummy');
+  }
+
   /**
     Returns the path to the configuration.
 
--- a/tests/unit/models/project-test.js
+++ b/tests/unit/models/project-test.js
@@ -557,24 +557,105 @@ describe('models/project.js', function() {
 
   if (isExperimentEnabled('MODULE_UNIFICATION')) {
     describe('isModuleUnification', function() {
+      describe('is app/addon', function() {
+        beforeEach(function() {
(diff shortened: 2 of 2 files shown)


NullVoxPopuli:
@ppcano @rwjblue @tomdale (@mixonic ?) Do we have a plan for this behavior? I was kinda just poking around with this, and don't know if there is a specific direction that we _should_ go in?

ppcano:
Added a test at #8322 -> `MU addon can run a Classic dummy app`.

The implementation was already done for this case (@mansona  at #7485 ?)

Will multiple dummy apps be automatically supported? At first glance, it does not look like to be a common case.

NullVoxPopuli:
@ppcano maybe not  super long term, but I can see it being incredibly beneficial while we have a bunch of octane and classic apps out in the wild simultaneously. As an addon author, I'd want to be able to verify that if my addon is a classic addon, it'll work with both octane and classic apps, and if my addon is an octane addon, that'll it'll work with octane and classic apps.

The matrix of assurance that I want addon authors to be able to complete:
<table>
<tr>
<th>
Addon ⇨<br>
App ⇩
</th>
<th>Octane</th>
<th>Classic</th>
</tr>
<tr><th>Octane</th><td>✔</td><td>✔</td></tr>
<tr><th>Classic</th><td>✘</td><td>✔</td></tr>
</table>

NullVoxPopuli:
RIP MU

PR #8159, opened 2018-10-29
cherry pick commit to put package-info-cache msgs behind DEBUG flag (…

dcombslinkedin:
…3ea02c9b73)

(from Dave Combs, not Stefan Penner, on 10/27/18)

### DIFF
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -11,6 +11,7 @@ const ErrorList = require('./error-list');
 const Errors = require('./errors');
 const PackageInfo = require('./package-info');
 const NodeModulesList = require('./node-modules-list');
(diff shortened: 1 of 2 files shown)


stefanpenner:
restarting node 6 failure.

stefanpenner:
@rwjblue node failure seems to be strange, any ideaS? Also should we backport this, if the fix is already in 3.5?

stefanpenner:
As 3.5 is already available, and this is not a show stopper, I think we don't need to back port this.

Thanks for the backport, but I this one I think we will have to pass on.

dcombslinkedin:
Thanks, @stefanpenner -- I appreciate the help from both you and @rwjblue on this, and agree with the result.

PR #8165, opened 2018-11-05
remove unused "ember-default" scenario

kellyselden:
This scenario was replaced with just `npm test` once we adopted Travis stages. It no longer shows up in the .travis.yml.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -60,12 +60,6 @@ module.exports = function() {
           }
         },
         {
-          name: 'ember-default',
-          npm: {
-            devDependencies: {}
-          }
-        },
-        {
           name: 'ember-default-with-jquery',
           env: {
             EMBER_OPTIONAL_FEATURES: JSON.stringify({
--- a/tests/fixtures/addon/npm/config/ember-try.js
+++ b/tests/fixtures/addon/npm/config/ember-try.js
@@ -59,12 +59,6 @@ module.exports = function() {
           }
         },
         {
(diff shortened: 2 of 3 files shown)


bendemboski:
@kellyselden @rwjblue I'm concerned about this change. Now if you create an addon with the default blueprint and run `yarn test:all`, you get

```
Scenario ember-lts-2.16: SUCCESS
Command run: ember test
with env: {
  "EMBER_OPTIONAL_FEATURES": "{\"jquery-integration\":true}"
}
Scenario ember-lts-2.18: SUCCESS
Command run: ember test
with env: {
  "EMBER_OPTIONAL_FEATURES": "{\"jquery-integration\":true}"
}
Scenario ember-release: SUCCESS
Command run: ember test
Scenario ember-beta: SUCCESS
Command run: ember test
Scenario ember-canary: SUCCESS
Command run: ember test
Scenario ember-default-with-jquery: SUCCESS
Command run: ember test
with env: {
  "EMBER_OPTIONAL_FEATURES": "{\"jquery-integration\":true}"
}
```

but no test of the default scenario without jQuery integration. This seems wrong, and also a dangerous change for folks using CI systems other than travis -- suddenly `yarn test:all` doesn't test all and had I not noticed this change in the release notes, I'd suddenly have a hole in my test coverage.

I think we should back this out so `yarn test:all` actually tests all again.

rwjblue:
Thank you for the well reasoned reply here @bendemboski. Would gladly accept a PR reverting (target `release` branch so we can fix in a patch release).

bendemboski:
#8288

PR #8167, opened 2018-11-05
variables `test` and `hinting` are expected to be boolean not string.

SparshithNR:
variables `test` and `hinting` are read from `env` variable, which is of type `string`. 
This PR will check if those variable exist and then convert it to boolean (using JSON.prase).
We can use Boolean(x) for conversion. Please suggest if you have a better method.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -185,7 +185,7 @@ class EmberApp {
     @param {Object} options
   */
   _initTestsAndHinting(options) {
-    let testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND || !this.isProduction;
+    let testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND === 'true' || !this.isProduction;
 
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -194,7 +194,7 @@ class EmberApp {
     @param {Object} options
   */
   _initTestsAndHinting(options) {
-    let testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND || !this.isProduction;
+    let testsEnabledDefault = (process.env.EMBER_CLI_TEST_COMMAND && JSON.parse(process.env.EMBER_CLI_TEST_COMMAND)) || !this.isProduction;
```
rwjblue: We don't need to parse the string (its only ever set to `true`), we should probably just coerce to boolean. 

Mind updating to:

```js
let testsEnabledDefault = process.env.EMBER_CLI_TEST_COMMAND === 'true' || !this.isProduction;
```

PR #8180, opened 2018-11-10
Clarify error message around package/addon name mismatch

mike-north:
This error message was originally introduced at #7950, but it provided no information (or links to information) about why this new constraint was introduced.

This is particularly important for handling common cases for ember addon packages that provide a different module namespace (i.e, the 'ember-lodash' or 'ember-fetch' cases)

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -375,9 +375,9 @@ let addonProto = {
 
           throw new SilentError(
             'Your names in package.json and index.js should match. ' +
-            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.` +
-            'For more information about this workaround, see: https://github.com/ember-cli/ember-cli/pull/7950. ' +
-            'Until ember-cli v3.9, this error can be disabled by setting env variable EMBER_CLI_IGNORE_ADDON_NAME_MISMATCH to "true"');
+            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js. ` +
+            'Until ember-cli v3.9, this error can be disabled by setting env variable EMBER_CLI_IGNORE_ADDON_NAME_MISMATCH to "true". ' +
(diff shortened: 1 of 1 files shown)


lib/models/addon.js
```diff
@@ -375,7 +375,9 @@ let addonProto = {
 
           throw new SilentError(
             'Your names in package.json and index.js should match. ' +
-            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.`);
+            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.` +
+            'For more information, see: https://github.com/ember-cli/ember-cli/pull/7950. '+
+            'This error can be disabled by setting env variable EMBER_CLI_IGNORE_ADDON_NAME_MISMATCH to "true"');
```
rwjblue: We should say that this work around will be removed in 3.9

rwjblue:
not ok 9 eslint should have no errors in lib
  lib/models/addon.js
    379:114  error  Infix operators must be spaced  space-infix-ops

lib/models/addon.js
```diff
@@ -375,7 +375,9 @@ let addonProto = {
 
           throw new SilentError(
             'Your names in package.json and index.js should match. ' +
-            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.`);
+            `The addon in ${pathToDisplay} currently have '${parentName}' in package.json and '${this.name}' in index.js.` +
+            'For more information about this workaround, see: https://github.com/ember-cli/ember-cli/pull/7950. ' +
+            'Until ember-cli v3.9, this error can be disabled by setting env variable EMBER_CLI_IGNORE_ADDON_NAME_MISMATCH to "true"');
```
Turbo87: @mike-north can you swap the last two lines?

mike-north: done

PR #8197, opened 2018-11-15
fix(default-packager): [MU] process templates only once

### LINKED ISSUE #8196: default-packager runs `processTemplates` twice in Module Unification mode
I'm using the latest ember-cli master e02ec00543195ecddf475319273519f40f3d8bef.

If Module Unification is enabled, templates are being processed twice.

First in L240:

https://github.com/ember-cli/ember-cli/blob/e02ec00543195ecddf475319273519f40f3d8bef/lib/broccoli/default-packager.js#L236-L246

Which calls `preprocessTemplates` in L467 in `this.processTemplates`:

https://github.com/ember-cli/ember-cli/blob/e02ec00543195ecddf475319273519f40f3d8bef/lib/broccoli/default-packager.js#L439-L474

And then in `this.processSrc` in L693:

https://github.com/ember-cli/ember-cli/blob/e02ec00543195ecddf475319273519f40f3d8bef/lib/broccoli/default-packager.js#L674-L696

Is this intentional / correct?

I noticed this, because I tried getting `ember-cli-addon-docs` to run with a MU addon. However the [`HBSContentExtractor`](https://github.com/ember-learn/ember-cli-addon-docs/blob/e13a69010d95185ceb5672982a13cfedbd92763a/lib/preprocessors/hbs-content-extractor.js#L12-L16) is consequently called twice, which causes an error in [`broccoli-bridge`](https://github.com/dfreeman/broccoli-bridge).

buschtoens:
Fixes #8196.

I am not sure, whether this is the correct solution, but it makes sense to me.

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -237,7 +237,7 @@ module.exports = class DefaultPackager {
     if (this._cachedProcessedAppAndDependencies === null) {
       let config = this.packageConfig();
       let internal = this.packageEmberCliInternalFiles();
-      let templates = this.processTemplates(allTrees);
+      let templates = this.isModuleUnificationEnabled ? null : this.processTemplates(allTrees);
 
       let trees = [
         allTrees,


lib/broccoli/default-packager.js
```diff
@@ -237,7 +237,7 @@ module.exports = class DefaultPackager {
     if (this._cachedProcessedAppAndDependencies === null) {
       let config = this.packageConfig();
       let internal = this.packageEmberCliInternalFiles();
-      let templates = this.processTemplates(allTrees);
+      let templates = this.isModuleUnificationEnabled ? null : this.processTemplates(allTrees);
```
NullVoxPopuli: does this need tests?

what is the intention here?

buschtoens: > does this need tests?

Probably? :D

> what is the intention here?

You can find a detailed description in #8196.

rwjblue:
Replaced by https://github.com/ember-cli/ember-cli/pull/8414

PR #8203, opened 2018-11-22
Move contribution info away from README to CONTRIBUTING

kennethlarsen:
This PR moves information about how to contribute to an addon away from README.md to a new CONTRIBUTING.md. 

The README is a bit cluttered by default when generating a new addon and I think we would benefit a lot by encouraging addon developers to use a CONTRIBUTING.md by default. 

Let me know what you think.

### DIFF
--- a/blueprints/addon/files/README.md
+++ b/blueprints/addon/files/README.md
@@ -17,34 +17,6 @@ Usage
 [Longer description of how to use the addon in apps.]
 
 
-Contributing
-------------------------------------------------------------------------------
-
-### Installation
-
-* `git clone <repository-url>`
-* `cd <%= addonName %>`
-* `<% if (yarn) { %>yarn<% } else { %>npm<% } %> install`
-
-### Linting
-
-* `<% if (yarn) { %>yarn lint:hbs<% } else { %>npm run lint:hbs<% } %>`
-* `<% if (yarn) { %>yarn lint:js<% } else { %>npm run lint:js<% } %>`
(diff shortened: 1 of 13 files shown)


blueprints/addon/files/README.md
```diff
@@ -17,34 +17,6 @@ Usage
 [Longer description of how to use the addon in apps.]
 
 
-Contributing
```
kellyselden: Should we link to Contributing? or is that an implied convention?

Turbo87: > is that an implied convention?

yep, see https://help.github.com/articles/setting-guidelines-for-repository-contributors/

kellyselden: Thanks for the link. That link reminded me that we do have a docs folder https://github.com/ember-cli/ember-cli/blob/e02ec00543/blueprints/app/files/package.json#L10

Perhaps we should move it there?

Turbo87: hmm, looks like GitHub supports picking it up from there, but for visibility reason I think I'd prefer to keep it at top-level. `docs` for me is for second level documentation, where files like README or CONTRIBUTING can like to.

blueprints/addon/files/npmignore
```diff
@@ -22,6 +22,7 @@
 /tests/
 /yarn.lock
 .gitkeep
+/CONTRIBUTING.md
```
kellyselden: Please preserve the ordering we have here.

kennethlarsen:
@rwjblue I've updated with MU-stuff. As far as I could see there are no MU tests for the content of the generated addon but only the file count - correct?

PR #8205, opened 2018-11-22
run eslint-plugin-node on apps

kellyselden:
I'm finding that it is pretty useful to run eslint-plugin-node on apps, even though the code is unpublished. You get the nice rules that prevent running newer node code and relying on transient deps. Especially if your code is ember-cli-deploy code that doesn't always run with your test suite.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -44,11 +44,15 @@ module.exports = {
       env: {
         browser: false,
         node: true
-      }<% if (blueprint !== 'app') { %>,
+      },
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
-        // add your custom rules and overrides for node files here
-      })<% } %>
+        // add your custom rules and overrides for node files here<% if (blueprint === 'app') { %>
+
+        // this can be removed once the following is fixed
+        // https://github.com/mysticatea/eslint-plugin-node/issues/77
+        'node/no-unpublished-require': 'off'<% } %>
+      })
     }
   ]
 };
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -42,6 +42,7 @@
     "ember-source": "~3.10.0-beta.1<% if (welcome) { %>",
     "ember-welcome-page": "^4.0.0<% } %>",
     "eslint-plugin-ember": "^6.2.0",
+    "eslint-plugin-node": "^8.0.1",
     "loader.js": "^4.7.0",
     "qunit-dom": "^0.8.4"
   },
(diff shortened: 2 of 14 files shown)


blueprints/app/files/.eslintrc.js
```diff
@@ -43,11 +43,15 @@ module.exports = {
...
+      },
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
-        // add your custom rules and overrides for node files here
-      })<% } %>
+        // add your custom rules and overrides for node files here<% if (blueprint !== 'app') { %>
+
+        // this can be removed once the following is fixed
+        // https://github.com/mysticatea/eslint-plugin-node/issues/77
+        'node/no-unpublished-require': 'off'<% } %>
```
rwjblue: This is because "everything" is a devDependencies?

Couldn't we fix by adding `files: ['dist']` or something similar to `package.json`?

blueprints/app/files/.eslintrc.js
```diff
@@ -43,11 +43,15 @@ module.exports = {
       env: {
         browser: false,
         node: true
-      }<% if (blueprint !== 'app') { %>,
+      },
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
-        // add your custom rules and overrides for node files here
-      })<% } %>
+        // add your custom rules and overrides for node files here<% if (blueprint !== 'app') { %>
```
rwjblue: Still has a trailing conditional here, is that intended?

blueprints/app/files/.eslintrc.js
kellyselden: `files: ['dist']` might work, but would be a confusion addition to imply your app's dist should but published. Plus we can't put a comment to explain it in the json file.

blueprints/app/files/.eslintrc.js
kellyselden: Yes. That is the closing `if` to the one above. The fixtures should confirm this is working (once I rebase and get them passing).

blueprints/app/files/package.json
```diff
@@ -42,6 +42,7 @@
     "ember-source": "~3.5.1<% if (welcome) { %>",
     "ember-welcome-page": "^3.2.0<% } %>",
     "eslint-plugin-ember": "^5.2.0",
+    "eslint-plugin-node": "^7.0.1",
```
rwjblue: should be updated to latest version (currently 8.0.1)

blueprints/app/files/.eslintrc.js
rwjblue: I was mistakenly thinking of no-extraneous-require's functionality, sorry for the confusion

blueprints/app/files/package.json
kellyselden: v7 is already brought in via the addon. I wouldn't want to touch the version in this PR. Created https://github.com/ember-cli/ember-cli/pull/8572.

rwjblue: > I wouldn't want to touch the version in this PR

Why?

rwjblue:
BTW, it seems like this should target beta not release...

PR #8206, opened 2018-11-22
remove duplicate eslint fixtures

kellyselden:
These were set up as if there was npm vs yarn uniqueness, but that's not the case. We can remove these duplicates.


rwjblue:
Probably needs a rebase to fix CI?

PR #8209, opened 2018-11-23
Skip broken smoke tests

Turbo87:
It looks like this test is breaking our CI runs for some reason. Let's hope this PR fixes CI again...

### DIFF
--- a/tests/acceptance/smoke-test-slow.js
+++ b/tests/acceptance/smoke-test-slow.js
@@ -235,7 +235,7 @@ describe('Acceptance: smoke-test', function() {
     });
   }));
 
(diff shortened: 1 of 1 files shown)


kellyselden:
It looks like node 10 + no experiments is the only thing that didn't regress https://travis-ci.org/ember-cli/ember-cli/builds/458393972. Do you think you could do something like "if node 10 && !Object.keys(experiments).length ? it : it.skip"?

Turbo87:
> Do you think you could do something like "if node 10 && !Object.keys(experiments).length ? it : it.skip"?

tried it in https://github.com/ember-cli/ember-cli/commit/e3dc5c531b07ec1a98dd9196438e0d5711a92685, but CI then also failed on that job unfortunately. reverted that experiment and as discussed on Discord this is 👍 on merging.

PR #8219, opened 2018-11-28
WIP: Swap `ember-cli-eslint` for ESLint

kellyselden:
Implementation of https://github.com/ember-cli/rfcs/pull/121.

I already started doing this at work, so I want to get it upstream.

### DIFF
--- a/blueprints/app/files/README.md
+++ b/blueprints/app/files/README.md
@@ -31,8 +31,8 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Running Tests
 
-* `ember test`
-* `ember test --server`
+* `<% if (yarn) { %>yarn test<% } else { %>npm test<% } %>`
+* `<% if (yarn) { %>yarn test --server<% } else { %>npm test -- --server<% } %>`
 
 ### Linting
 
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -15,7 +15,7 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
-    "test": "ember test"
+    "test": "<% if (yarn) { %>yarn lint:js<% } else { %>npm run lint:js<% } %> && ember test"
   },
   "devDependencies": {
     "@ember/jquery": "^0.5.2",
@@ -26,7 +26,6 @@
     "ember-cli-app-version": "^3.2.0",
     "ember-cli-babel": "^7.1.2",
     "ember-cli-dependency-checker": "^3.1.0",
-    "ember-cli-eslint": "^5.1.0",
     "ember-cli-htmlbars": "^3.0.0",
     "ember-cli-htmlbars-inline-precompile": "^1.0.3",
     "ember-cli-inject-live-reload": "^1.8.2",
@@ -41,6 +40,7 @@
     "ember-resolver": "^5.0.1",
     "ember-source": "~3.9.0-beta.2<% if (welcome) { %>",
     "ember-welcome-page": "^4.0.0<% } %>",
+    "eslint": "^5.15.1",
     "eslint-plugin-ember": "^5.2.0",
     "loader.js": "^4.7.0",
     "qunit-dom": "^0.8.0"
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -42,8 +42,6 @@ jobs:
       install:
         - yarn install --non-interactive<% } %>
       script:
-        - <% if (yarn) { %>yarn lint:hbs<% } else { %>npm run lint:hbs<% } %>
-        - <% if (yarn) { %>yarn lint:js<% } else { %>npm run lint:js<% } %>
         - <% if (yarn) { %>yarn test<% } else { %>npm test<% } %><% if (yarn) { %>
 
     - name: "Floating Dependencies"
--- a/blueprints/app/files/.travis.yml
+++ b/blueprints/app/files/.travis.yml
@@ -30,6 +30,4 @@ install:
   - yarn install --non-interactive<% } %>
 
 script:
-  - <% if (yarn) { %>yarn lint:hbs<% } else { %>npm run lint:hbs<% } %>
-  - <% if (yarn) { %>yarn lint:js<% } else { %>npm run lint:js<% } %>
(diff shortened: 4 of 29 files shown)


blueprints/app/files/README.md
```diff
@@ -31,8 +31,8 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Running Tests
 
-* `ember test`
-* `ember test --server`
+* `<% if (yarn) { %>yarn test<% } else { %>npm test<% } %>`
```
kellyselden: We already have linting mentioned below, so it's a bit redundant to mention the command that runs linting as well. On the other hand, how are people going to learn that `ember test` no longer lints, and that you should be running `npm test` to get the old behavior?

Also, we now have a weird mix of npm-scripts wrapped commands and native ember-cli commands. We may need learning team help on resolving this confusion. cc @jenweber @NullVoxPopuli @kennethlarsen

blueprints/app/files/package.json
```diff
@@ -15,7 +15,7 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
```
kellyselden: Should we include this in the start and build commands (and test:all) as well?

blueprints/app/files/package.json
```diff
@@ -15,7 +15,7 @@
...
-    "test": "ember test"
+    "test": "<% if (yarn) { %>yarn lint:hbs<% } else { %>npm run lint:hbs<% } %> && ember test"
```
kellyselden: We could also utilize "pretest" as well. I'm not sure of the advantages of each. This is essentially equivalent:

```
"pretest": "npm run lint:js",
"test": "ember test"
```

blueprints/app/files/package.json
```diff
@@ -15,7 +15,7 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
-    "test": "ember test"
+    "test": "<% if (yarn) { %>yarn lint:js<% } else { %>npm run lint:js<% } %> && ember test"
```
kellyselden: I left off `npm run lint:hbs` because there is currently no way that I know to take template linting out of `ember test`. Once that is possible, we can add it here. @rwjblue correct me if I'm wrong.

blueprints/app/files/README.md
NullVoxPopuli: > You mean yarn lint:js && ember test?

idk, I prefix everything with yarn, so I _know_ I'm using the local version

> It is already split in travis (or should be).

 #biwinning

blueprints/app/files/README.md
```diff
@@ -31,8 +31,8 @@ Make use of the many generators for code, try `ember help generate` for more det
...
+* `<% if (yarn) { %>yarn test --server<% } else { %>npm test -- --server<% } %>`
```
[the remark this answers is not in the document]
john-kurkowski: Right, I do like 1 SPOT to look for all errors in all files, continuously. Today, that's the launched Testem UI. It reminds me of things my editor doesn't, because I switched away from the file with errors.

Running another watch process is an option, sure.

Maybe a commit hook would cover my edge case.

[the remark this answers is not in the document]
kellyselden: I don't think a commit hook would solve your live updating needs, that's like a "final check once you're all done", right? not a "while you work" solution.

blueprints/app/files/README.md
jenweber: Perhaps the test output could indicate that linting is not included. For example, it could say afterwards "to run linting, use `npm test`." I will have to think about this some. 

Beginners aside, this is a tough question because I imagine that many people just do ember test in their CI. I don't use Travis and I'm pretty sure that I just run `ember test` in CI.

onlymejosh:
From reading the RFC it looks like there was a discussion about adding the documentation for setting up `testem.js` with eslint.

> As part of the implementation work we would like to clearly document _how_ to setup `testem.js` to properly run linting (both eslint and template linting) during `ember test --server` and _at least_ link to it from an inline comment in `testem.js` (possibly even include it automatically?). (@rwjblue)

This should be a requirement of this PR as without it, it will completely change the workflow of some developers. The testem ui is basically my one-stop-shop for errors, I could add another step but it feels like going backwards rather than forwards. I don't want to have to rely on CI to tell me about linting issues and precommit hooks are a drag.

kellyselden:
@onlymejosh I added the testem hook `before_tests` to run linting to keep the conversation going.

@rwjblue Was this what you were thinking?

onlymejosh:
If there is a linting issue would this stop the rest of the tests running? If so this is still a backwards change in my eyes. It should be the same as it currently is, linting issues should appear in the list of failures.

rwjblue:
>  If so this is still a backwards change in my eyes

I'm sorry you feel this way.

> It should be the same as it currently is, linting issues should appear in the list of failures.

Simply put, thats not going to happen. The RFC (and comment thread) specifically explain the rational, and when implementation is completed a newly generated app will not have the setup you are talking about. However, it is **absolutely** still a thing you can choose to opt in to (by installing ember-cli-eslint in your particular application).

onlymejosh:
Ah ok so the solution is to use ember-cli-eslint

> As part of the implementation work we would like to clearly document how to setup testem.js to properly run linting (both eslint and template linting) during ember test --server and at least link to it from an inline comment in testem.js (possibly even include it automatically?).

Is this what you meant by documenting? I guess I was confused by that comment. As I took it as it would act the same / similarly  but it looks like that is not the case?

sangm:
Any update on this?

blueprints/app/files/package.json
happycollision: You can remove `ember-cli-template-lint` and replace with `ember-template-lint` directly. This will keep Test'em from linting the has files, but you can still `yarn lint:hbs` with no issues.

blueprints/app/files/package.json
happycollision: Compatibility with Windows is the advantage of `pretest`.

rwjblue:
The basic parts of this were by https://github.com/ember-cli/ember-cli/pull/9009, 

I'm going to close this specific PR as rebasing it would be a nightmare (removal of MU, landing of an alternate approach in #9009, etc), but I'd love to see the other conversations that were ongoing here in smaller PRs focused on each topic.

PR #8230, opened 2018-11-29
[Bugfix release] Addon name mismatch should be a warning

chadhietala:
This PR changes the way we handle a mismatch in an addon name and the name field in the package.json. To help addon authors provide a deprecation path for their users I created [ember-cli/canonicalize-addon-name](https://github.com/ember-cli/canonicalize-addon-name).

### canonicalize-addon-name
This utility library provides functionality to addon authors to emit runtime deprecations for the mismatched import paths. It also provides the an ember-cli command that addon authors can offer to applications to codemod their application away from the old imports.

### DIFF
--- a/ADDON-NAME-MIGRATION.md
+++ b/ADDON-NAME-MIGRATION.md
@@ -1,0 +1,11 @@
+# Addon Name Migration
+
+Prior to 3.5.1 of ember-cli we would allow the name of the addon; specified in the index.js, and the name in the package.json to differ. We would like to canonicalize the name so they are the same. While we feel that this divergence is uncommon, addons that did not have a canonicalized name would produce code that did not align to the package name. This will make efforts like tree-shaking much easier to do as there doesn't need to be multiple resolution rules for `import`s in application code. This also will make debugging code much easier because the runtime module name will reflect the NPM package name.
+
+### For Addon Authors
+
+To help migrate this category of addons we have created a utility for providing a migration path for your consumers. [ember-cli/canonicalize-addon-name](https://github.com/ember-cli/canonicalize-addon-name) provides functionality to add runtime deprecations. It also provides a codemod that you can provide to your consumers to help them migrate.
+
+### For Addon Consumers
+
+If you have encountered this warning please open an issue up against the offending addon, so they can provide you with a migration path.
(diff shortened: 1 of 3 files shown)


rwjblue:
Should we updated the message to point folks towards https://github.com/ember-cli/canonicalize-addon-name or perhaps add a MD file in docs talking about a migration path?

chadhietala:
@rwjblue I added a markdown file and updated the message to point to that file.

rwjblue:
Rebased and targetted at `release` branch.

stefanpenner:
It isn't clear to me, this is actually the right path forward.

stefanpenner:
@chadhietala have you looked at `modulePrefix` property? It is used by `moduleName()` (public API) and I believe is the public API to have the es6 module differ from the name of the add-on.

https://github.com/ember-cli/ember-cli/blob/104fb6077ef6beb225b60288b693739e9200b814/lib/models/addon.js#L1436-L1439

stefanpenner:
Although I do think we should transition the community to avoid mismatched modules to asset with future work, that is public API right now via `moduleName()`. 

I think we should likely split this into two parts:

* (Now) ensure upgrades the 3.5.1 go better
* (Later) RFC/design to transition folks away from mismatched app.name and moduleName, but I think that should be part of more thorough effort. For example @ef4's work etc.

ef4:
In addition to changing `pkg.name` or `modulePrefix`, there are yet other ways that addons escape their own package namespace. A popular one is using `treeForAddonTestSupport` without calling `super`, which allows you to emit modules into any package namespace.

FWIW, embroider already handles all of these with no runtime cost, mostly by rewriting the consuming side. So while I'm in favor of deprecating the old behaviors, there's no hurry from my perspective.

stefanpenner:
@ef4 you likely are aware, but `moduleName() { }` is a public API intended explicitly to bypass allow for this (just in-case you didn't know, although I suspect that is unlikely).

ef4:
Yeah, I consider all of these patterns public API.

In general I don't want to hit addon authors with a steady stream of small changes. I'd rather be 100% compatible with all existing addons, and have people excited about porting addons to [a much stricter format](https://github.com/embroider-build/embroider/blob/master/SPEC.md) because they'll get an immediate benefit (build perf and sometimes app size) for each one.

(And that spec I linked is not an RFC yet because I don't want people to start porting to a moving target.)

allthesignals:
Hmm, this could be really helpful even for regular apps.

I've seen situations in which a developer changed the name of the package.json `name` key (to something clearer), which led to a lot of wheel-spinning because the app's index.html would then be pointing to the wrong javascript/css build output due to the mismatch.

The only evidence one sees of something _going wrong_ is a blank page with some console messages about _MimeTypes_ (because the server provides HTML 404 pages).

locks:
Since `index.js` now uses the [name off of `package.json`](https://github.com/ember-cli/ember-addon-output/blob/v4.3.0/index.js#L4), I believe this is alleviated.

PR #8194, opened 2018-11-15
`new` mu-app blueprint does not add `tests/{unit,integration}` folders

ppcano:
Relates to #7530 

Assuming that `unit` and `integration` tests should be placed where its "tested" class is located, there is no need to auto-generate the `tests/{unit,integration}` folders anymore.

The PR will make the `blueprint` to stop adding these folders, but it will generate a `tests/acceptance` folder instead.

PR #8231, opened 2018-11-30
Add tests for the MU new blueprint to validate the content

ppcano:
The PR tries to achieve test parity between MU and classic command. 

Related PRs to homogenize MU and classic tests  #8269, #8194, #8312


package-lock.json
```diff
@@ -0,0 +1,10955 @@
+{
```
Turbo87: @ppcano we use yarn instead of npm in this project. please remove this file.

tests/fixtures/module-unification-addon/.eslintrc.js
```diff
@@ -20,7 +20,6 @@ module.exports = {
     // node files
     {
       files: [
-        '.eslintrc.js',
```
Turbo87: why did this change?

Turbo87: and if we also have ` tests/fixtures/module-unification-addon/npm/.eslintrc.js` then what is this file here (` tests/fixtures/module-unification-addon/.eslintrc.js`) being used for?

ppcano: @Turbo87 

> why did this change?

The existing addon blueprint does not have this line. 

Do you want me to add it to both places ( blueprint file and the fixture)?


---------------------------------------------------------

`tests/fixtures/module-unification-addon/.eslintrc.js` is needed because how `checkEslintConfig` works, this function checks the `estlintrc.js` on the parent folder.


`tests/fixtures/module-unification-addon/npm/.eslintrc.js` is needed for the test `EMBER_CLI_MODULE_UNIFICATION: ember addon foo works`. 

````
        expect(actualFiles).to.deep.equal(expectedFiles);
````


Now, I think this `EMBER_CLI_MODULE_UNIFICATION: ember addon foo works` test is not needed anymore because with this PR the existing tests for Non-MU mode will also run for the MU mode.

What should I do here? Preserve both `eslintrc.js`, remove the mentioned tests or something else?

---------------------------------------------------------

I think, the end goal could be to have tests for same behaviour `MU` and `Non-MU`, so there are not existing tests that are skipped on the MU mode.

This is what I am trying to achieve with this and coming PRs. Please, correct me if I am wrong with this assumption.

package-lock.json
ppcano: Oops!, sorry, there was so many files with the feature that I didn't realize about this one. 

Removed at https://github.com/ember-cli/ember-cli/pull/8231/commits/a0d59f1a70b7925f570bb07862a2f0ea362e7499

ppcano:
@Turbo87 

> due to the size of the PR it is somewhat hard to review. I've left a few comments.

Yes too many files, but most of them are for adding a new fixture case for MU addon folder:

https://github.com/ember-cli/ember-cli/tree/master/tests/fixtures/module-unification-addon

which it will have two subfolders like:

`/module-unification-addon/npm`:  test MU addon with npm
`/module-unification-addon/yarn`:  test MU addon with yarn

So we can run the same tests that we do for the Non-MU mode.

In summary:

- the blueprint has NOT been changed.
- more fixtures have been added.
- tests have been added but not removed (I have asked in another comment about that).

So I think the PR is safe for merging.

tests/fixtures/module-unification-addon/.eslintrc.js
ppcano: Added `.eslintrc.js` in blueprint and fixture at https://github.com/ember-cli/ember-cli/pull/8231/commits/6af2c66b5ed71dcd3d1be5f6401cbd63ec440eaf

ppcano:
I did some changes based on the review, rebased and now CI is green.

PR #8240, opened 2018-11-30
Reset blueprint's ignoredFiles

dcyriller:
When running `ember new` or `ember addon` in sequence, on the same node
process, `ignoredFiles` gets properly incremented. But it was never reset.

The change this PR introduce is especially usefull in tests.

This PR aims at reseting the key so that the following test setup
works properly:
```javascript
beforeEach(() => {
  emberNew({ isModuleUnification: true, target: 'addon' })
})
```

Indeed, a bug occured while testing the `addon` blueprint for the Module
Unification Quest. Because in the tests, we want to make sure the `src`
directory exists. Previously, `addon` directory was disapearing but
the issue remained silent as we were not asserting its existence.

I set the release branch as a target. Because this would unblock `ember-cli` upgrade in emberjs package (https://github.com/emberjs/ember.js/pull/17212). Not sure if it's appropriate though.

/cc @ppcano

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -30,6 +30,10 @@ const Promise = RSVP.Promise;
 const stat = RSVP.denodeify(fs.stat);
 const writeFile = RSVP.denodeify(fs.outputFile);
 
+const initialIgnoredFiles = [
+  '.DS_Store',
+];
+
 /**
   A blueprint is a bundle of template files with optional install
   logic.
@@ -693,13 +697,15 @@ let Blueprint = CoreObject.extend({
   },
 
   /**
-    Add update files to ignored files
+    Add update files to ignored files or reset them
     @private
     @method _ignoreUpdateFiles
   */
   _ignoreUpdateFiles() {
     if (this.isUpdate()) {
       Blueprint.ignoredFiles = Blueprint.ignoredFiles.concat(Blueprint.ignoredUpdateFiles);
+    } else {
+      Blueprint.ignoredFiles = initialIgnoredFiles;
     }
   },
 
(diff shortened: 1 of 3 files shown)


lib/models/blueprint.js
```diff
@@ -1379,12 +1381,18 @@ Blueprint.renamedFiles = {
 
 /**
   @static
-  @property ignoredFiles
+  @property initialIgnoredFiles
 */
-Blueprint.ignoredFiles = [
+Blueprint.initialIgnoredFiles = [
```
kellyselden: We can store this off the prototype at the top of the file. No need to expose this internal detail.

dcyriller: Thanks @kellyselden, I updated it

dcyriller:
I've inserted a commit to highlight the bug. The second commit is the fix. I'm going to add a third commit to actually prevent regression with a test.

dcyriller:
@kellyselden I added a test. I ended up with: a first commit for the failing test, a second commit for the fix.

ppcano:
@dcyriller Some tests are failing, try running the tests with the MU flag:

`EMBER_CLI_MODULE_UNIFICATION=true node tests/runner.js tests/acceptance/addon-generate-test.js`

dcyriller:
Thank you @ppcano, it's fixed now

kellyselden:
Can you rebase one more time? We fixed the remaining AppVeyor issues so it should go green.

dcyriller:
Great. I rebased, CI is running.

rwjblue:
Thanks for working on this!

ppcano:
emberjs/ember.js#17253 finally upgraded Ember to `ember-cli@3.5.1` without this fix. 


But this fix is needed if we want to remove unnecessary call to `fs.ensureDirSync('src')` on MU blueprint tests. This PR will fix the changes of  emberjs/ember.js#17267

kellyselden:
Restarted AppVeyor, all green now.

PR #8261, opened 2018-12-04
add server/**/*.js to eslint node files for app

kellyselden:
`ember g server` puts node files in here, so we should probably have them in the default eslint list.


kellyselden:
That's a good idea.

rwjblue:
@kellyselden - Had a chance to work on that?

Turbo87:
seems good as a first step, we can still move it into the server blueprint later

PR #8264, opened 2018-12-05
Remove redundant _requireBuildPackages and format comments

xg-wang:
`_requireBuildPackages ` is a no-op since https://github.com/ember-cli/ember-cli/pull/7525


rwjblue:
Thank you @xg-wang!

PR #8268, opened 2018-12-06
suppress EOL from insertIntoFile

kellyselden:
Without this, it is impossible to do this (with our api):

`"test": "ember test",` => `"test": "yarn lint:scss && ember test",`

With this you could do:

```js
    await this.insertIntoFile(
      'package.json',
      `yarn lint:scss && `,
      {
        after: `"test": "`,
        appendEOL: false
      }
    );
```

Alternative solution is:

```js
    let pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
    if (pkg.scripts && pkg.scripts.test) {
      pkg.scripts.test = pkg.scripts.test.replace('ember test', 'yarn lint:scss && ember test');
      fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + require('os').EOL);
    }
```

### DIFF
--- a/lib/utilities/insert-into-file.js
+++ b/lib/utilities/insert-into-file.js
@@ -98,8 +98,12 @@ function insertIntoFile(fullPath, contentsToInsert, providedOptions) {
           let insertIndex = contentMarkerIndex;
           if (insertBehavior === 'after') { insertIndex += contentMarker.length; }
 
+          if (options.appendEOL !== false) {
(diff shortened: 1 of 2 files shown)


Turbo87:
I'd prefer to deprecate `insertIntoFile()`... 😅

kellyselden:
That's also an option. It didn't really stand the test of time.

rwjblue:
Where do we stand here?

PR #8286, opened 2018-12-12
Provide a compatibility section in addon READMEs

kennethlarsen:
I think it would be helpful to add a compatibility section to addon READMEs suggesting to use a version of Ember that is no lower than two LTS's back.

Let me know what you think.


blueprints/addon/files/README.md
```diff
@@ -10,12 +10,15 @@ Installation
 ember install <%= addonName %>
 ```
 
-
```
Turbo87: this extra newline was added to give the subsections a bit more margin/padding

blueprints/addon/files/README.md
```diff
@@ -10,12 +10,15 @@ Installation
...
 Usage
 ------------------------------------------------------------------------------
 
 [Longer description of how to use the addon in apps.]
 
+Compatibility
```
Turbo87: I would propose to put this section above "Usage" (possible even above "Install") because "Usage" could get quite long 🤔

blueprints/addon/files/README.md
```diff
@@ -10,12 +10,15 @@ Installation
...
+------------------------------------------------------------------------------
+
+It is recommended to use Ember 2.18 or anything higher than that for this addon to work properly.
```
Turbo87: I'm wondering if prose is the right form for that. I was thinking maybe a list of bullet points would be more suited here. Something like:

- Ember.js v2.18 or above
- Ember CLI v3.5 or above
- `ember-intl` v4.x

blueprints/addon/files/README.md
kennethlarsen: Yes, I think that's a good point.

blueprints/addon/files/README.md
```diff
@@ -3,6 +3,13 @@
 
 [Short description of the addon.]
 
+
+Compatibility
+------------------------------------------------------------------------------
+
+* Ember.js v2.18 or above
```
Turbo87: I just remembered that we now put addons on `ember-cli-babel@7`, which means that they need "Ember CLI 2.13 or above" too. We should probably add that too.

Turbo87:
I'd personally be happy with targetting this PR to be released with EmberCLI 3.7, which means that it would need to target the `beta` branch.

blueprints/addon/files/README.md
kellyselden: We also only test LTS by default, so while Ember v3.1 is technically above Ember.js v2.18, it isn't really compatible because it isn't tested. Not sure if we should indicate this or not though.

Turbo87: the big question here is if addons should state that they only support the latest two LTS releases. that would mean dropping support for older Ember releases could be done without bumping major versions, which I'm personally not a fan of.

blueprints/addon/files/README.md
```diff
@@ -3,6 +3,14 @@
...
+* Ember CLI 2.13 or above
```
Turbo87: @kennethlarsen sorry, I missed it too above, but could you change it to `v2.13` so that it matches the Ember.js line formatting?

PR #8288, opened 2018-12-12
Revert #8165

bendemboski:
Removing the default scenario from `ember-try.js` omits it from tests runs initiated via `yarn test:all` or `ember try:each`, and potentially creates holes in test coverage for users that use CI services other than travis.

Reverts #8165

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -59,6 +59,10 @@ module.exports = function() {
             }
           }
         },
+        // The default `.travis.yml` runs this scenario via `<% if (yarn) { %>yarn<% } else { %>npm<% } %> test`,
+        // not via `ember try`. It's still included here so that running
+        // `ember try:each` manually or from a customized CI config will run it
+        // along with all the other scenarios.
         {
           name: 'ember-default',
           npm: {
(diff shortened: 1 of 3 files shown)


rwjblue:
Since https://github.com/ember-cli/ember-cli/pull/8165 landed in release, we should target this at release. 

Can you:

* update target branch in GH
* rebase dropping all commits other than the one you are introducing
* force update your remote branch

I’m happy to do ^ when I’m back at a computer, just let me know if you’d like help there...

kellyselden:
Also please add a comment so this doesn't happen again.

bendemboski:
@rwjblue I think I did what you asked -- lemme know if I got it wrong.

@kellyselden I'm not sure of a great way to do that without the comment or some funky whitespace ending up in the generated file, or making it kinda awkward in the template. Is something like this what you were thinking?

```javascript
        {
          name: 'ember-canary',
          npm: {
            devDependencies: {
              'ember-source': urls[2]
            }
          }
        }, <% // even though .travis.yml doesn't use this scenario, we still want it so `yarn test:all` will include it, both for manual invocation and users using other CI systems/a customized travis config %>
        {
          name: 'ember-default',
          npm: {
            devDependencies: {}
          }
        },

bendemboski:
Also, not sure what's up with the checks (appveyor not re-running, travis and coverage not run at all) -- do I need to close and re-open the PR or something? If you don't know offhand I'm sure I can figure it out.

kellyselden:
@bendemboski I would leave the comment in for all people. I think it would be useful to explain to people why it's there (for people like me who didn't think it was used).

kellyselden:
+1 on the wording. Thanks.

PR #8297, opened 2018-12-14
Remove last usage of Babel 6.

rwjblue:
This last usage is _only_ used for supporting compiling templates in `addon/` and supporting `addon-test-support/` in addons that themselves do not have any `.js` processors (e.g. no dependency on `ember-cli-babel`). Note: those scenarios have been emitting a warning for quite some time...

This change is *not* a breaking change to our users (the module compilation between Babel 6 and Babel 7 is compatible).

### DIFF
--- a/lib/utilities/relative-module-paths.js
+++ b/lib/utilities/relative-module-paths.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const path = require('path');
+const ensurePosix = require('ensure-posix-path');
+const { moduleResolve } = require('amd-name-resolver');
+
+const BASE_DIR = path.resolve(`${__dirname}/..`);
+
+function getRelativeModulePath(modulePath) {
+  return ensurePosix(path.relative(process.cwd(), modulePath));
+}
+
+function resolveRelativeModulePath(name, child) {
+  return moduleResolve(name, getRelativeModulePath(child));
+}
+
+Object.keys(module.exports).forEach(key => {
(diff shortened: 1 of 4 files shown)


lib/utilities/relative-module-paths.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+Object.keys(module.exports).forEach(key => {
+  module.exports[key].baseDir = () => BASE_DIR;
+
+  module.exports[key]._parallelBabel = {
+    requireFile: __filename,
+    useMethod: key,
+  };
+});
```
Turbo87: I'm wondering if we should move this into `amd-name-resolver` to not copy it around everywhere 🤔

rwjblue: Yeah, seems reasonable to me

rwjblue: Should we land this and then extract, or extract then land?

Turbo87: I'm totally fine with landing and extracting later

rwjblue:
Hehe, I was ;)

rwjblue:
Totally fine if we want to revert from release and just land in beta though, up to you...

Turbo87:
guess it would be the right thing to do... sorry that I didn't notice it earlier 😞

rwjblue:
Nah, no worries!

PR #8299, opened 2018-12-14
remove ember-lts-2.16 ember-try scenario

kellyselden:
It's time to remove this from the blueprint now that 3.4 is in there.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -53,7 +53,6 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-lts-2.16
     - env: EMBER_TRY_SCENARIO=ember-lts-2.18
     - env: EMBER_TRY_SCENARIO=ember-lts-3.4
     - env: EMBER_TRY_SCENARIO=ember-release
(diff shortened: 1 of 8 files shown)


rwjblue:
@kellyselden - I'm 👍 on removing, but I think we should probably stick it in beta branch. Expanding coverage doesn't seem "breaking" but I think removing it might...

Turbo87:
> I think we should probably stick it in beta branch

agreed

Turbo87:
@kellyselden can you retarget to beta? we can discuss backporting on the next call, but I'd like to at least add it to the next beta

blueprints/addon/files/.travis.yml
```diff
@@ -53,7 +53,6 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-lts-2.16
     - env: EMBER_TRY_SCENARIO=ember-lts-2.18
```
buschtoens: This now incorrectly triggers a build job without environment variables: https://travis-ci.org/buschtoens/ember-lazy-mount/builds/477951861

image

We need to remove the `-` from `- env:` for `ember-lts-2.18`. Super unintuitive and easy to mess API by Travis. 😞

Turbo87: Oh, crap, sorry about that 😢

Could you open a PR against the release branch that fixes it?

buschtoens: No worries, I was dumbfounded as well. Wouldn't have caught that mistake myself. 😊

buschtoens: This is in fact _so hard_ to do right, that I _just_ fucked it up myself. 😂 

https://github.com/buschtoens/ember-lazy-mount/commit/c95124abfb79d7c45413e832dffbe3b0b44effb0
image

PR #8308, opened 2018-12-18
ignore .env* files

kellyselden:
Because lots of people have these, and since npm doesn't merge .npmignore files from system unlike .gitignore, we can be nice to people and prevent the mistake.


Turbo87:
@kellyselden this is not a bugfix though, so should not target LTS or `release`. I would suggest to aim this at `beta`.

kellyselden:
retargetted

PR #8309, opened 2018-12-19
fix(blueprints/addon): add contributing section

knownasilya:
Section links to the new contributing guide.
For discovery and easy access.


kennethlarsen:
@Turbo87 Yeah that's probably a good idea in case someone is not super familiar with the github structure.

Just needs a fixture for the MU addon though

Turbo87:
@knownasilya can you add that to the MU addon README too?

knownasilya:
@Turbo87 all set

PR #8314, opened 2018-12-21
In MU apps, exclude test files from the app JS file

ppcano:
The PR fixes that test files in the `src` folder are added to the `testJsFile` instead of the `appJsFile`.

The current implementation recognize tests with the pattern `**/*-test.js`, but this will fail if a non-test class is named like `xxx-test.js`. 

```
services/app-test.js         // the service
services/app-test-test.js // test of the `app-test` service
```

This case will fail because both files will be added into the `testJsFile`.

It should fix some of the issues reported at emberjs/ember.js#17224

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -676,6 +676,7 @@ module.exports = class DefaultPackager {
       let src = new Funnel(tree, {
         srcDir: 'src',
         destDir: 'src',
+        exclude: ['**/*-test.js'],
         annotation: 'Module Unification Src',
       });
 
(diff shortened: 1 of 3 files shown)


lib/broccoli/default-packager.js
```diff
@@ -676,6 +676,7 @@ module.exports = class DefaultPackager {
       let src = new Funnel(tree, {
         srcDir: 'src',
         destDir: 'src',
+        exclude: ['**/*-test.js'],
```
rwjblue: Can we make the exclude only if MU is enabled?

ppcano: @rwjblue It is already done because this funnel is only executed if MU is enabled:

```
processSrc(tree) {
    if (this._cachedProcessedSrc === null && this.isModuleUnificationEnabled) {
      let src = new Funnel(tree, {
        srcDir: 'src',
        destDir: 'src',
        exclude: ['**/*-test.js'],
        annotation: 'Module Unification Src',
      });
```

I don't think it is needed to change the `exclude` behaviour, let me know otherwise.

NullVoxPopuli:
I still have test files in my bundle :-\

ppcano:
@NullVoxPopuli which test files are included in the bundle app, do you have a repo I can reproduce it?

NullVoxPopuli:
This branch: https://gitlab.com/NullVoxPopuli/emberclear/merge_requests/155

PR #8315, opened 2018-12-22
fix(blueprints/addon): add angle-bracket invoocation polyfill to ember-try

knownasilya:
Since we know that Ember versions < 3.4 don't support AB, we can add this by default for addons so that their tests pass and they can start using AB at least in their tests without any worries.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -19,7 +19,8 @@ module.exports = function() {
           npm: {
             devDependencies: {
               '@ember/jquery': '^0.5.1',
-              'ember-source': '~2.16.0'
+              'ember-source': '~2.16.0',
+              'ember-angle-bracket-invocation-polyfill': '^1.2.3'
             }
           }
         },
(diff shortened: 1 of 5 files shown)


Turbo87:
agreed, if it uses angle brackets then either it should declare that it only supports Ember 3.4+, or it should have the polyfill as a dependency

knownasilya:
This is for writing tests using AB, not necessarily promoting AB for consumption.

Turbo87:
If you want to write tests using AB you should probably add that as a dev dependency

knownasilya:
Another thing for addon authors to have to think about. This is compounded even more if they are new to the ecosystem and don't have experience pre-AB. If they are starting out with AB, most likely their tests will be written in that too. Lets make sure that their addons are tested against older versions right from the start, without wasting time to track down the issue or ignoring older version support or even worse, tests all together.

I'm fine with another solution, as long as we can do it by default.

Turbo87:
Either they care about backwards compact, in which case they need the polyfill or they only care about AB in which case they can remove the older ember scenarios. I don't think we need to change anything here.

@ember-cli/ember-cli opinions?

PR #8317, opened 2018-12-23
Upgrade `ember-source` for MU new blueprints

### LINKED ISSUE #8316: Should the MU blueprints set `ember-source` to the latest Ember release?
When testing a MU app, I realised that a few blueprints (controllers, helpers, initializers...) were not working; the reason is that the MU app blueprint has configured an old `ember-source` version.

```
    "ember-source": "https://s3.amazonaws.com/builds.emberjs.com/canary/shas/caa2c70d40dbfeffb0e6c9ac15ff6d72c8e24bae.tgz",
```

The latest `ember-source` upgrade for MU blueprint was done at #7976.

The mentioned blueprints are already working on the latest Ember release, then changing the `ember-source` version fixes the blueprints. 

I wonder what is really preventing from using the latest ember version with MU apps, I posted previously the question at https://github.com/ember-cli/ember-cli/pull/8085#issuecomment-443087725

I think it may be better if MU apps uses the latest Ember release:
-  users testing MU could benefit of the latest changes (in my case, it was the Ember blueprints).
-  users could report any existing issue against the latest stable version.

I can work on the PR but I would like to know which strategy should follow at this moment: 
-  upgrade to a particular build 
-  use the latest release. 

If the second option is decided, the MU blueprint should also be updated with each new Ember version like at https://github.com/ember-cli/ember-cli/commit/a824605717b7239972b17dd2611f79b046fc48bf

ppcano:
Closes #8316 

> The blueprint should point to the most recent canary tarball (via ember-source-channel-url). We need to use canary specifically because MU requires a feature flag to be enabled which is only possible in canary builds.

The last upgrade was done by #7976


ppcano:
I thought about that too when reviewing https://github.com/ember-cli/ember-octane-blueprint/blob/master/index.js, but this change will likely require to change some ember-cli tests. 

As you see in the PR the test fixtures are also changed because the content of the `npm/yarn` file is compared and validated. 

I am not sure atm how to change the tests if the `ember-source`value is dynamically fetched.

We follow the same approach when upgrading the `ember-source` of the Classic app as show at https://github.com/ember-cli/ember-cli/commit/a824605717b7239972b17dd2611f79b046fc48bf

rwjblue:
> I am not sure atm how to change the tests if the ember-sourcevalue is dynamically fetched.

If you take a look here, we already do a custom matcher for `package.json` so taht we can account for dynamic `ember-cli` versions, I think this function can _fairly_ easily be extended to support custom `ember-source` version. What do you think?

https://github.com/ember-cli/ember-cli/blob/413b26073fb3b09d79dbd6adef5e4e894f4f5835/tests/acceptance/new-test.js#L428

PR #8323, opened 2018-12-28
Fetch dynamically the `ember-source` version for MU blueprints

ppcano:
Follows up #8317. 

With this change, we won't have to set manually the ember-source version in the blueprint and test fixtures, and it will always select the latest Canary build.

cc @rwjblue 

-----------------------------------

Additionally, it could be good if we improve the procedure to upgrade the Ember and Ember Data versions (see this [commit](https://github.com/ember-cli/ember-cli/commit/a824605717b7239972b17dd2611f79b046fc48bf)). 

We could define a top level variable to be reused in the blueprint and test fixtures, in a similar way than the `<%= emberCLIVersion %>`.

Thoughts?

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -13,6 +13,7 @@ let date = new Date();
 const normalizeEntityName = require('ember-cli-normalize-entity-name');
 const stringifyAndNormalize = require('../../lib/utilities/stringify-and-normalize');
 const FileInfo = require('../../lib/models/file-info');
+const getURLFor = require('ember-source-channel-url');
 
 const replacers = {
   'package.json'(content) {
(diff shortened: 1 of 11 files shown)


rwjblue:
> Additionally, it could be good if we improve the procedure to upgrade the Ember and Ember Data versions (see this commit).

Totally agree.

blueprints/addon/index.js
```diff
@@ -106,6 +109,12 @@ module.exports = {
     return new FileInfo(options);
   },
 
+  beforeInstall() {
```
rwjblue: Should this be in the module-unification-addon blueprint instead of the addon one?

ppcano: Done by https://github.com/ember-cli/ember-cli/pull/8323/commits/89e76fc94fe98f7b6db549ddea8d94da6118594e

Issue #8164, opened 2018-10-31
Yarn v2 / PnP Support


rwjblue:
Yarn 1.12 added support for a new "Plug 'n Play" feature by way of an `--pnp` option. This issue is meant to track adding support for `ember new foo --yarn` to support using `--pnp` without issue.

Further Reading:

* https://github.com/yarnpkg/rfcs/pull/101
* https://twitter.com/yarnpkg/status/1057632564217704453
* https://gist.github.com/arcanis/02b49752c385908479734d8027d7a6c7

Current known issues:

- [ ] Need to remove the `node_modules appears empty, you may need to run 'yarn install'` warning when `.pnp.js` is present
- [ ] PackageInfoCache has significant issues (currently emits tons of warnings, since all of the structures that it expects to exist are missing), and _may_ not be relevant at all in a PnP app. Needs further research...
- [ ] Update ember-try to handle this nicely https://github.com/ember-cli/ember-try/issues/214
- [ ] Updated version of `resolve` including the changes from ~~https://github.com/browserify/resolve/pull/170~~ https://github.com/browserify/resolve/pull/172, https://github.com/browserify/resolve/pull/173. and https://github.com/browserify/resolve/pull/174  (to ensure `resolve.sync` usage is aware of PnP)
- [x] Update `ember-cli-dependency-checker` https://github.com/quaertym/ember-cli-dependency-checker/issues/92
- [x] Update `ember-cli-version-checker` to address https://github.com/ember-cli/ember-cli-version-checker/issues/78

drummy2:
I need this so bad.  Still looks quite far away :(

musaffa:
@rwjblue Hi, it has been a while. Are you still pursuing this feature? What are the main blockers?

bartocc:
I am experimenting with the [PnP API](https://yarnpkg.com/advanced/pnpapi) to make ember-cli compatible with Yarn PnP.

I might open a draft PR at some point to show the progress and get ideas from the community

sesam:
How is Yarn PnP compatibility going?

bartocc:
@sesam for the time being, I use https://github.com/yarnpkg/berry/tree/master/packages/plugin-node-modules

@ef4, do you believe embroider will change things regarding yarn PNP ?

gilest:
Not sure if I should post this here or make another issue but I found an API breakage when using `ember install` also.

[...]

PR #8330, opened 2019-01-02
blueprints/app/gitignore: Ignore Yarn PnP files

Turbo87:
see https://github.com/yarnpkg/pnp-sample-app/blob/master/.gitignore

Related to https://github.com/ember-cli/ember-cli/issues/8164

### DIFF
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -9,6 +9,7 @@
 /node_modules/
 
 # misc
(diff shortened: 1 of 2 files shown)


kellyselden:
We should probably start respecting the yarn/npm options to make this conditional.

Turbo87:
I'm not sure if it's worth the extra complexity for these files 🤔

PR #8344, opened 2019-01-08
Add no-implicit-this template lint rule by default.

rwjblue:
As part of [emberjs/rfcs#308](https://emberjs.github.io/rfcs/0308-deprecate-property-lookup-fallback.html#transition-path)'s phase 2. The main guides/docs have been updated (will be deployed with 3.7 docs) in https://github.com/ember-learn/guides-source/pull/349.

### DIFF
--- a/blueprints/app/files/.template-lintrc.js
+++ b/blueprints/app/files/.template-lintrc.js
@@ -1,5 +1,9 @@
 'use strict';
 
 module.exports = {
-  extends: 'recommended'
+  extends: 'recommended',
+
+  rules: {
+    'no-implicit-this': true
+  }
 };


kellyselden:
Yeah it seems like this isn't a style issue and perhaps should be recommended?

rwjblue:
Please review the provided RFC’s transition path ([here](https://emberjs.github.io/rfcs/0308-deprecate-property-lookup-fallback.html#transition-path)). Later phases absolutely include this in the recommended set. 

Having the rule enabled directly in the application:

* Makes it clearer where these new errors are coming from as folks upgrade
* Makes it obvious that application authors can “simply” remove the rule from their config to avoid dealing with it

With that being said, if we want to skip directly to phase 3 (adding the rule to recommended set), I’m personally ok with it...

Turbo87:
if we add it here it will take another ~3 months to actually be released to the public, if we change it in template-lint we don't necessarily have to wait that long. (dependency update in our blueprint here will take the same time, but people can upgrade the dep independently)

I guess I'm just not a fan of having unnecessary churn in these files, especially if the change is only temporary. Adding it and then removing it again shortly after seems to have a bit of potential for confusion 🤔

rwjblue:
>  Adding it and then removing it again shortly after seems to have a bit of potential for confusion 🤔

Sure, I agree. I think the bigger issue here is that folks _commonly_ skip releases when they upgrade ember-cli blueprint content. In that scenario, they would jump right to the next stage _anyways_ (making this change a moot point).

rwjblue:
Haha, I'm convinced. Let me do the major bump in ember-template-lint....

Turbo87:
ember-template-lint v1.0.0 has been released including this rule, so I'll go ahead and close this PR as obsolete

rwjblue:
It’s fine to close, but 1.0.0 does not include this change (seemed too trolly to add a breaking change after an 8 month beta)...

rwjblue:
https://github.com/ember-template-lint/ember-template-lint/pull/601 is going to enable it by default and be a 2.0.0

Turbo87:
oh, whoops, I guess I should've actually read the changelog 🙈

PR #8347, opened 2019-01-08
Use cwd from testem if configured else use default value

arthirm:
If cwd is configured in testem.js file it should use it, else outputPath should be used as cwd.


stefanpenner:
Will review

arthirm:
There was a confusion because of internal monkey patching. This change is not needed. Am closing this PR.

PR #8358, opened 2019-01-11
In MU apps, exclude TS test files from the app JS file

ppcano:
It continues #8314 and fixes the issue reported at https://github.com/emberjs/ember.js/issues/17224#issuecomment-452294926


rwjblue:
Thank you!

PR #8365, opened 2019-01-16
[BUGFIX] Use moduleName() for templates

pzuraq:
We were working on an internal addon and trying to upgrade `ember-cli`, and were attempting to use `moduleName()` to squash the whole "name must match package name" errors temporarily, but some of the modules weren't respecting `moduleName`. It turned out that templates do not currently use it, and this PR fixes that.

There weren't any tests for `moduleName()` that I could find, happy to add some but would need some guidance on where the best place would be for that, and how to do it

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1150,7 +1150,7 @@ let addonProto = {
     if (addonTemplates) {
       standardTemplates = new Funnel(addonTemplates, {
         srcDir: '/',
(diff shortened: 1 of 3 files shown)


mactenm:
a pin  point that loooks shoulf be the logo is that all that musing is life just with 4 simple people behind it just easy draeing flat style with a lot of shade

pzuraq:
Sure thing 👍

PR #8370, opened 2019-01-17
[BUGFIX][lts] Use moduleName() for templates

pzuraq:
We were working on an internal addon and trying to upgrade ember-cli, and were attempting to use moduleName() to squash the whole "name must match package name" errors temporarily, but some of the modules weren't respecting moduleName. It turned out that templates do not currently use it, and this PR fixes that.

---

Note: This was migrated over from https://github.com/ember-cli/ember-cli/pull/8365 (to retarget and rebase).

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1150,7 +1150,7 @@ let addonProto = {
     if (addonTemplates) {
       standardTemplates = new Funnel(addonTemplates, {
         srcDir: '/',
-        destDir: `${this.name}/templates`,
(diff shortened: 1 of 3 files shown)


rwjblue:
Looks like the failure must be related (that retriggering passes...)

rwjblue:
Hmm, @pzuraq have you had a chance to poke at these failures? I don't really see how they could be related, but lts-3-4 branch itself does seem to pass so I guess it has to be related somehow...

rwjblue:
I just updated and retargeted at `release` branch.

PR #8374, opened 2019-01-18
WIP: package-info-cache: Add Yarn PnP support

Turbo87:
This is marked WIP because it seems to work, but I basically have no idea what I actually did there and have a hard time judging if the corner cases will still work correctly... 😞 

related to https://github.com/ember-cli/ember-cli/issues/8164

### Todo

- [ ] feature flag the new code

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -116,5 +116,9 @@ module.exports = {
     'symbol-description': 2,
     'template-curly-spacing': 2,
     'yield-star-spacing': 2,
+
+    'node/no-missing-require': ['error', {
+      'allowModules': ['pnpapi'],
+    }],
   },
 };
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -510,7 +510,7 @@ class PackageInfoCache {
    * @param {Boolean} isRoot, for when this is to be considered the root
    * package, whose dependencies we must all consider for discovery.
    */
-  _readPackage(packageDir, pkg, isRoot) {
+  _readPackage(packageDir, pkg, isRoot, skipReal = false) {
     let normalizedPackageDir = path.normalize(packageDir);
 
     // Most of the time, normalizedPackageDir is already a real path (i.e. fs.realpathSync
@@ -530,7 +530,7 @@ class PackageInfoCache {
     // we can actually find a real path (including resolving links if needed).
     let pathFailed = false;
 
-    let realPath = getRealDirectoryPath(normalizedPackageDir);
+    let realPath = skipReal ? normalizedPackageDir : getRealDirectoryPath(normalizedPackageDir);
 
     if (realPath) {
       if (realPath !== normalizedPackageDir) {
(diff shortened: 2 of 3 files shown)


Turbo87:
@stefanpenner @dcombslinkedin @rwjblue would be great to get your eyes and thoughts on this! I tried to look into it again, but as mentioned above I basically have no idea how the package info cache works. seems too complex for me to reason about it :-/

with just the change to `addDependencies()` and using `NODE_PRESERVE_SYMLINKS=true` (on CLI `master`) things seem to work. Actively disabling the real path resolution using the additional argument in this PR also appears to work.

I'm very scared though that this change is breaking other things that I'm not aware of so please let me know what you think about this.

davecombs:
@Turbo87, @spenner, @rwjblue, I thought that a version of this had already been done to use resolve-package-path in preference to the implementation of realFilePathCache and realDirectoryPathCache defined here. Is that right?

If not, I'd suggest we do that, even if we don't end up using all parts of the 4-part cache that's defined there, and let that part handle the PnP support and the addition of the extra argument (called skipReal here so it's all in one place.

One other thing: is 'skipReal' only intended to be used for the case where the node options are set to maintain symlinks? If so, I'd suggest a more specific name like keepSymlinks (or whatever it was in resolve-package-path). The flag AFAIK is not really general purpose - it's only for this one case.

Comments?

Turbo87:
> I thought that a version of this had already been done to use resolve-package-path in preference to the implementation of realFilePathCache and realDirectoryPathCache defined here. Is that right?

yes, the `master` branch is using the path functions of `resolve-package-path` for the Package Info Cache. but that does not seem to solve the PnP compatibility issue. when you run `ember build` with `DEBUG="ember-cli:package-info-cache"` it will show you that it was not able to find any of the addons and finally the build pipeline will show an error that it could not find a template compiler.

> is 'skipReal' only intended to be used for the case where the node options are set to maintain symlinks?

`skipReal` is currently more of a hack since I was looking for feedback from you guys on a better solution. the problem is that PnP uses the "virtual" paths for any lookups changing to the real paths creates problems during Babel plugin lookup.

to answer your question: no, `skipReal` is unrelated to the `--preserve-symlinks` option of Node, though if that option is used it appears that `skipReal` is now no longer necessary to fix that Babel plugin resolver issue.

---

FYI for all of these tests I've used a very basic side project of mine: https://github.com/Turbo87/lva-camo-status/

davecombs:
I’ll take a look tomorrow or Sunday.

stefanpenner:
Ya the current usage of resolve-package-path merely shares the real path cache.

The issue is the custom addon discovery code, in the package-info-cache stuff is more or less redundant with pnp as their is no hierarchy to traverse.

I have a unverified suspicion that the underlying caches we have makes the custom traversal code redundant.  Potentially just complicate matters, such as introducing the above branching.

Garibaldy:
> @stefanpenner @dcombslinkedin @rwjblue would be great to get your eyes and thoughts on this! I tried to look into it again, but as mentioned above I basically have no idea how the package info cache works. seems too complex for me to reason about it :-/
> 
> with just the change to `addDependencies()` and using `NODE_PRESERVE_SYMLINKS=true` (on CLI `master`) things seem to work. Actively disabling the real path resolution using the additional argument in this PR also appears to work.
> 
> I'm very scared though that this change is breaking other things that I'm not aware of so please let me know what you think about this.

what it does 'NODE_PRESERVE_SYMLINKS' is to make a hard copy to the node_modules folder of the packages it will behave at the same speed as npm or just Yarn , in some cases a lot slower

related #8611 

also PNPM  is running the same issue  and until it get fixed they got a flag that behave the same way 'NODE_PRESERVE_SYMLINKS'

LIKE 

`PNPM install --shamefully-flatten `

PR #8379, opened 2019-01-22
Fix the module path for MU non-acceptance tests

### LINKED ISSUE #8363: In MU apps, relative imports in a test file fails
Reported at https://github.com/ember-cli/ember-cli/commit/c90329c0fade6b249ada95dd37a897ff3bd97e32#commitcomment-31925271

Given a relative import in a test file:

```
//src/utils/dom/unit-test
import * as string from './utils';
```

```
Error: Could not find module `$packageName/tests/utils/dom/utils` 
               imported from `$packageName/tests/utils/dom/unit-test`
```

The test files use the `${packageName}/tests` namespace which is different than the app namespace `${packageName}/src`, therefore the relative import fails.

I will look at it.

ppcano:
Closes #8363

It set correctly the module path of the MU tests; given the following test file `/src/utils/unit-test`:

Before -> `$packageName/tests/utils/unit-test` 
Now ->  `$packageName/src/utils/unit-test`

The fix solves relative imports within the test file. The PR includes an unit test:

```
import { module, test } from 'qunit';
import stringFn from './string';

 module('Unit | Utility | string', function() {
  ...
 });
```

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -844,27 +844,30 @@ module.exports = class DefaultPackager {
         exclude: ['addon-test-support/**/*'],
       });
 
-      if (this.isModuleUnificationEnabled) {
-
-        let testSrcTree = new Funnel(tree, {
-          srcDir: 'src',
-          include: ['**/*-test.{js,ts}'],
-          annotation: 'Module Unification Tests',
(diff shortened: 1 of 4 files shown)


rwjblue:
Awesome, thank you!

PR #8384, opened 2019-01-25
Support octane blueprints

ppcano:
Previously discussed at ember-cli/ember-octane-blueprint#7. This PR allows running the octane blueprints like:

```
ember new my-app -b octane-app
ember addon my-addon -b octane-addon
```

- [ ] The dependency at the `package.json` file must be changed when ember-cli/ember-octane-blueprint#8 is merged.

The PRs tries to include initially the minimal possible changes, if they are merged, we can continue iterating the octane blueprints and test how the current `MU` tests suite could be run with the octane blueprints.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -553,14 +553,15 @@ class Project {
     @return {Array} List of paths
    */
   blueprintLookupPaths() {
-    if (this.isEmberCLIProject()) {
-      let lookupPaths = [this.localBlueprintLookupPath()];
-      let addonLookupPaths = this.addonBlueprintLookupPaths();
 
-      return lookupPaths.concat(addonLookupPaths);
-    } else {
-      return this.addonBlueprintLookupPaths();
+    let octaneBlueprintPath = path.join(__dirname, '..', '..', 'node_modules', 'ember-octane-blueprint', 'blueprints');
+    let lookupPaths = [octaneBlueprintPath];
+    if (this.isEmberCLIProject()) {
+      lookupPaths.push(this.localBlueprintLookupPath());
     }
+
+    let addonLookupPaths = this.addonBlueprintLookupPaths();
+    return lookupPaths.concat(addonLookupPaths);
   }
 
   /**
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-preprocess-registry": "^3.1.2",
     "ember-cli-string-utils": "^1.1.0",
-    "ember-octane-blueprint": "ppcano/ember-octane-blueprint#ea9175e4d0788b70ca778193cc7f8d54fd8d5ef0",
+    "ember-octane-blueprint": "ember-cli/ember-octane-blueprint#8/head",
     "ember-source-channel-url": "^1.1.0",
     "ensure-posix-path": "^1.0.2",
     "execa": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-preprocess-registry": "^3.1.2",
     "ember-cli-string-utils": "^1.1.0",
+    "ember-octane-blueprint": "ppcano/ember-octane-blueprint#ea9175e4d0788b70ca778193cc7f8d54fd8d5ef0",
     "ember-source-channel-url": "^1.1.0",
     "ensure-posix-path": "^1.0.2",
     "execa": "^1.0.0",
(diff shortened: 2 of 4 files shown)


ppcano:
The problem with the test looks to be that Travis is not installing the correct commit version https://github.com/ember-cli/ember-octane-blueprint/pull/8/commits/ea9175e4d0788b70ca778193cc7f8d54fd8d5ef0

We can wait for https://github.com/ember-cli/ember-octane-blueprint/pull/8 to become merged to change the dependency version and try again.

package.json
```diff
@@ -74,6 +74,7 @@
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-preprocess-registry": "^3.1.2",
     "ember-cli-string-utils": "^1.1.0",
+    "ember-octane-blueprint": "ember-cli/ember-octane-blueprint#8/head",
```
Turbo87: I don't think it's a good idea to have git dependencies here

lib/models/project.js
```diff
@@ -553,14 +553,15 @@ class Project {
...
    */
   blueprintLookupPaths() {
-    if (this.isEmberCLIProject()) {
-      let lookupPaths = [this.localBlueprintLookupPath()];
-      let addonLookupPaths = this.addonBlueprintLookupPaths();
 
-      return lookupPaths.concat(addonLookupPaths);
-    } else {
-      return this.addonBlueprintLookupPaths();
+    let octaneBlueprintPath = path.join(__dirname, '..', '..', 'node_modules', 'ember-octane-blueprint', 'blueprints');
```
Turbo87: this should use `require.resolve()` or something similar

package.json
ppcano: I think the plan is to use a version once the related PR is merged and published. I do not have permission to do that.

Turbo87: still seems strange to me to have it as a dependency tbh. can't we download/clone the blueprint on demand instead?

ppcano: @Turbo87 Not sure. I open the discussion at https://github.com/ember-cli/ember-octane-blueprint/issues/7, and @NullVoxPopuli agreed on this strategy.

As far as I talked with @rwjblue, the requirements were:

1) we wanted to have an independent repo for the `octane` blueprints that it includes both the  `octane-app` and `octane-addon` blueprints where we could iterate quickly these blueprints.

2) it should facilitate an easy path for users to try octane. With this approach, any user that install the latest `ember-cli` could run:

```
ember new my-app -b octane-app
ember addon my-addon -b octane-addon
```

3) we want to create tests for the octane blueprints, but we wanted to reuse when possible the existing MU blueprint tests to avoid the duplication of both test suites. I commented a possible approach at https://github.com/ember-cli/ember-octane-blueprint/issues/7#issuecomment-450040740

> can't we download/clone the blueprint on demand instead?

What is your suggestion?

NullVoxPopuli: @ppcano I may have misunderstood before, but idk if we want to include the octane blueprint by default quite yet. most people won't be using it, so it'd just be bloat on their system.

I think e-cli should clone the blueprint repo if it doesn't already exist on their system (this'll also help get away from only using published npm packages)

rwjblue:
Hmm, why does ember-cli need to be aware of this internally? Is there a specific issue with using two separate packages (in a monorepo setup)? e.g. `@ember/octane-blueprint` and `@ember/octane-addon-blueprint`?

NullVoxPopuli:
@rwjblue afaik, blueprint-only packages need to have their files in the root of the repo -- we could do some publishing config to get around that, but I guess the goal here was to reduce the friction for using an external blueprint.

At present:
 - need to publish npm package
   - npm package needs to have `files` directory at the root

This can be fine for official blueprints, but I just see this as too much friction for 3rd-party blueprint makers.
Maybe this needs to be an RFC.

For now, we could do some travis / auto-publishing from the master branch to publish the different npm packages. I'm totes fine with going that route for now. 
Automation is bae.

NullVoxPopuli:
@ppcano so, how about we go the route where we proceed with the mono-repoing of the blueprint repo, and I'll just do some travis config tomorrow for auto-publishing to npm (currently using my own npm creds).

rwjblue:
@NullVoxPopuli

>afaik, blueprint-only packages need to have their files in the root of the repo

Yes, exactly, this is why I proposed having two packages 😸.

Concretely, I'm suggesting a monorepo structure like:

```
.
└── packages
    ├── @ember
    │   ├── octane-addon-blueprint
    │   │   ├── files
    │   │   └── package.json
    │   └── octane-blueprint
    │       ├── files
    │       └── package.json
    └── internal-blueprint-tests
```

This allows us to:

* not require changes in ember-cli
* continue to test the blueprints in relationship to each other

ppcano:
Then, , I am closing this PR. Thank you everyone for participating on the discussion. 

@NullVoxPopuli I wonder how users will finally run the octane blueprints:

Will `ember new my-app -b octane-app` work?

rwjblue:
Basically yes, though I’d like the packages to be `@ember/octane-blueprint` and `@ember/octane-addon-blueprint` I think. Obviously we have time to bikeshed on the names a bit...

PR #8399, opened 2019-02-03
[INTERNAL] Improve jQuery deprecation message

simonihmig:
* As discussed with @rwjblue, this makes the existing deprecation message more specific, stating explicitly that  the jQuery integration per se is deprecated and will be removed from Ember itself
* added tests

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -489,7 +489,11 @@ class EmberApp {
       if (optionFeatures && !optionFeatures.isFeatureEnabled('jquery-integration')) {
         return;
       }
-      this.project.ui.writeDeprecateLine('Ember will stop including jQuery by default in an upcoming version. If you wish keep using jQuery in your application explicitly add `@ember/jquery` to your package.json');
+      this.project.ui.writeDeprecateLine(
+        'The integration of jQuery into Ember has been deprecated and will be removed with Ember 4.0. You can either' +
+        ' opt-out of using jQuery, or install the `@ember/jquery` addon to provide the jQuery integration. Please' +
+        ' consult the deprecation guide for further details: https://emberjs.com/deprecations/v3.x#toc_jquery-apis'
+      );
       jqueryPath = ember.paths.jquery;
     } else {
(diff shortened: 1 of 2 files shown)


tests/unit/broccoli/ember-app-test.js
```diff
@@ -511,6 +515,10 @@ describe('EmberApp', function() {
         'bar': 'bar',
       });
 
+      app = new EmberApp({
+        project,
+      });
+
```
simonihmig: This change was required as `app` would be `undefined` otherwise. It accidentally worked before, as `app` initialized and leaking from another test before.

lib/broccoli/ember-app.js
```diff
@@ -489,7 +489,11 @@ class EmberApp {
       if (optionFeatures && !optionFeatures.isFeatureEnabled('jquery-integration')) {
         return;
       }
-      this.project.ui.writeDeprecateLine('Ember will stop including jQuery by default in an upcoming version. If you wish keep using jQuery in your application explicitly add `@ember/jquery` to your package.json');
+      this.project.ui.writeDeprecateLine(
```
kellyselden: Forgive me for being out of the loop, but is it not possible to put this in the ember-source addon itself?

simonihmig: I think you are right, and given that it's ember-source which is actually deprecating things (i.e. causing breaking changes in the future), it would make more sense to put it there. Especially that ember-source and ember-cli might not be used in lock step.

On the other hand, IIRC this was already added a while ago when we started to not include jQuery anymore (which historically ember-cli did, not ember-source!) when `@ember/jquery` is detected. So when we want to move that message to ember-source, we would probably have to check the ember-source version here, to not issue a deprecation message twice!? 

A bit of shenanigans involved, but I can do it if we agree it's worth it!?

alechirsch: It is possible to print out where jquery is being included? We have a project with tons of dependencies and it would be a huge task to sift through all of those and figure out which one is using it.

PR #8404, opened 2019-02-04
Improve styles tests

patricklx:
@rwjblue trying the get at least some of my work from https://github.com/ember-cli/ember-cli/pull/8146 for contribution :)

This improves the tests to use real addon instances, so they are closer to real use. Previous test did not catch the issues.
Also the test `can handle empty styles folders`, does actually not test on an empty folder.


ppcano:
@patricklx can we close this PR in favor of https://github.com/ember-cli/ember-cli/pull/8424?

patricklx:
sure, I'll close it

PR #7810, opened 2018-05-03
WIP: Testing CI on npm@6 on travis

mansona:
I have something that I want to try out that is very specific to the Travis infrastructure. The first commit is to show that it's failing before the experiment

PR #8408, opened 2019-02-05
testing with npm

mansona:
This is essentially an extension of https://github.com/ember-cli/ember-cli/pull/7810

I think it's potentially worth considering having at least one test run with npm only (no yarn) as a kind of smoke test for any future issues that might crop up 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,11 @@ jobs:
     - env: EMBER_CLI_BROCCOLI_2=false
     - env: EMBER_CLI_SYSTEM_TEMP=false
 
+    - name: NODE_10_NPM
+      node_js: 10
+      npm: true
+      script: npm run test:all
+
     - stage: deploy
       script: ./.travis/deploy.sh
 


.travis.yml
```diff
@@ -59,6 +59,11 @@ jobs:
     - env: EMBER_CLI_BROCCOLI_2=false
     - env: EMBER_CLI_SYSTEM_TEMP=false
 
+    - env: NAME=NODE_10_NPM
```
Turbo87: we can use `name: npm` now instead of misusing `env`

.travis.yml
```diff
@@ -59,6 +59,11 @@ jobs:
...
+      node_js: 10
+      install: npm install
+      script: npm run test:all
```
Turbo87: this sounds like it would add significantly more time to our already long CI pipeline 😢

.travis.yml
mansona: 👍

.travis.yml
mansona: I guess it depends on what you mean by "significant" 😂 

In the case of individual jobs it seems like the npm run is consistently faster than yarn in the 2 out of 2 builds that I've seen now 😂 Not a great sample size but considering that's without a package-lock.json I'm pretty impressed.

Secondly, it seems like we are getting a 5 job concurrency on the big stage and adding this job brings the total number of jobs to 10. In reality I guess that means that as long as we keep the job count below 10 we should be fine when it comes to elapsed time 🤔

rwjblue:
Going to close this now, we no longer pin down npm version. Happy to review/help with another PR to master with GH actions if you still think we should move forward.

PR #8410, opened 2019-02-06
convert add-to-output-repos sh to js

kellyselden:
This will allow me to add more abilities easier (extraction), and makes it easier to debug.


dev/add-to-output-repos.js
```diff
@@ -0,0 +1,73 @@
...
+  return new Promise((resolve, reject) => {
+    _exec(command, (err, stdout) => {
+      if (err) {
+        return reject(err);
+      }
+
+      resolve(stdout.trim());
+    });
+  });
+}
```
Turbo87: any reasons for not using `execa` instead?

kellyselden: It wasn't working for the git clone for some reason.

PR #8289, opened 2018-12-13
Include addon styles for MU apps

ppcano:
1) The PR supersedes #8134.

This change fixes the [CSS stylesheets that require importing from addons](https://cli.emberjs.com/release/writing-addons/intro-tutorial/#cssstylesheetsthatrequireimportingfromaddons) feature for MU applications importing styles from classic addons.

2)  This PR also supersedes #8146 to allow MU applications importing SASS modules like the `ember-styleguide`.

MU applications will now execute  `preprocessCss` on `packageStyles` like Classic applications and it removes the existing `preprocessCss` call of the `processSrc` method. 

I am not sure this is the correct implementation but this implementation looks more natural and it enables the `specifying custom output paths works properly` test at https://github.com/ember-cli/ember-cli/pull/8289/commits/6c65235254f446b9dc2d4d78396c05cec97df5d5  (cc @twokul).

A similar issue was reported at https://github.com/ember-cli/ember-cli/pull/8197 where MU applications executed `preprocessTemplates` twice on `processSrc` and `processTemplates`.

Closes #8134, #8133,  #8146

PR #8414, opened 2019-02-07
`preprocessTemplates` is called only once in MU layout

### LINKED ISSUE #8196: default-packager runs `processTemplates` twice in Module Unification mode

ppcano:
Fixes #8196 and supersedes #8197. The PR avoid `preprocessTemplates` being called twice on the same templates.

The implementation follows the same type of changes that were done at #8289 where `defaultPackager.packageStyles` became the function to process styles (Classic and MU layout).

This PR changes `defaultPackager.processTemplates` to be the only function to process templates (Classic and MU layout).

`defaultPackager.processSrc` becomes `defaultPackager.processJavascriptSrc` and will only process JS/TS files for MU apps.

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -245,7 +245,7 @@ module.exports = class DefaultPackager {
       let trees = [
         allTrees,
         templates,
-        this.processSrc(allTrees),
+        this.processJavascriptSrc(allTrees),
       ].filter(Boolean);
 
       let mergedTree = mergeTrees(trees, {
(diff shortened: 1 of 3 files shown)


lib/broccoli/default-packager.js
```diff
@@ -444,19 +444,36 @@ module.exports = class DefaultPackager {
...
+
+      if (this.isModuleUnificationEnabled) {
+        let mu = new Funnel(templates, {
+          srcDir: 'src',
+          destDir: `${this.name}/src`,
+          include,
+          exclude: ['templates/**/*'],
+          annotation: 'MU Templates',
+        });
+        mu = this._debugTree(mu, 'mu-layout:tree:template');
```
rwjblue: 👍

PR #8419, opened 2019-02-09
WIP: add hook for addons to add to the resolver config

NullVoxPopuli:
some info here: https://github.com/emberjs/ember.js/issues/17234#issuecomment-461810595 (comment on the octane tracking issue).

This is important so addon authors can add to the resolver config at build time, and that the "install steps" for addons with custom types and collections are kept as they are today (small).

I got stuck, cause I couldn't find where src/resolver.js was being loaded.

The closest I got was actually in ember-resolver, where the default config path from ember-resolver is added to the build tree: 
https://github.com/ember-cli/ember-resolver/blob/master/index.js#L64-L67
and that lead me to here: https://github.com/ember-cli/ember-cli-preprocess-registry/blob/master/preprocessors.js#L141
and now I don't know how or why anything is relevant :(

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -639,6 +639,14 @@ class EmberApp {
       if (addon.included) {
         addon.included(this);
       }
+
+      // addons may provide their own types and collections
+      // within the resolver
+      if (addon.addToResolver) {
+        let addonResolverConfig = addon.addToResolver(this.project.resolver);
+
+        merge(addonResolverConfig, this.project.resolver);
+      }
     });
   }
 


ppcano:
This requirement was initially created at the [QUEST Module Unification - Final Cut](https://github.com/emberjs/ember.js/issues/16373)

> Ember addons need an API that allows them to alter the module unification config (for example https://github.com/ember-cli/ember-resolver/blob/master/mu-trees/addon/ember-config.js) and add their own collections and types. In theory Ember (or the resolver?) would use this same API to add its own types and collections.

ppcano:
The addon could provide its resolver config like: 

```js
module.exports = {
  name: 'my-addon',
  resolverConfig(app, parentAddon) {
    
    if (app.isModuleUnification && app.isModuleUnification()) {
      return {
        types: {
          'session': { definitiveCollection: 'session' },
          'session-store': { definitiveCollection: 'session-store' },
        },
        collections: {}
      }
    }
  }
};
```


In this case, the autogenerated resolver could be something like:

```js
// Octane/MU blueprint should auto-generate this `src/resolver.js`

import Resolver from 'ember-resolver/resolvers/fallback';
import buildResolverConfig from 'ember-resolver/ember-config';
import config from '../config/environment';

// `ember-resolver/addon-config` has to be auto-generated based on the outcome `addon.resolverConfig({})` call of all the addons
import addonResolverConfigs from 'ember-resolver/addon-config';
import mergeConfig from 'ember-resolver/merge-config';


let moduleConfig = buildResolverConfig(config.modulePrefix);

// merge addon custom types and collections provided by the `addon.resolverConfig()` API
moduleConfig = mergeConfig(moduleConfig, addonResolverConfigs);


/*
 * If your application has custom types and collections, modify moduleConfig here
 * to add support for them.
 */
export default Resolver.extend({
  config: moduleConfig
});
```

This approach requires more technical research but I wonder if this could be a valid solution.

ppcano:
This PR could be closed in favour of https://github.com/ember-cli/ember-resolver/pull/348

NullVoxPopuli:
Thanks @ppcano !

PR #8426, opened 2019-02-12
updating the version of npm that is installed in addon blueprint

mansona:
Considering this npm version that is being installed is from before the time of lock files I think it's about time to consider upgrading it 🙈 

This will significantly improve DX for quite a few people 👍

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -68,7 +68,7 @@ install:
   - yarn install --no-lockfile --non-interactive
 <% } else { %>
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@6
   - npm --version
 <% } %>
 script:
--- a/blueprints/module-unification-addon/files/.travis.yml
+++ b/blueprints/module-unification-addon/files/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@6
   - npm --version
 
 script:
--- a/tests/fixtures/addon/npm/.travis.yml
+++ b/tests/fixtures/addon/npm/.travis.yml
@@ -53,7 +53,7 @@ jobs:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@6
   - npm --version
 
 script:
--- a/tests/fixtures/module-unification-addon/npm/.travis.yml
+++ b/tests/fixtures/module-unification-addon/npm/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@6
   - npm --version
 
 script:
--- a/tests/fixtures/module-unification-addon/yarn/.travis.yml
+++ b/tests/fixtures/module-unification-addon/yarn/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@4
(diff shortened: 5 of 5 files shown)


kellyselden:
Relevant https://github.com/ember-cli/ember-cli/pull/7482

blueprints/addon/files/.travis.yml
```diff
@@ -68,7 +68,7 @@ install:
   - yarn install --no-lockfile --non-interactive
 <% } else { %>
   - npm config set spin false
-  - npm install -g npm@4
+  - npm install -g npm@6
```
rwjblue: Why do we pin this at all? Is there a valid reason to _not_ do:

```suggestion
  - npm install -g npm
```

kellyselden: npm bumps min node version in their majors.

mansona: @rwjblue I actually 100% agree with you on this one, currently all of our CI runs are manually installing the absolute latest yarn using a script (at least Travis does this automatically) so I think to put things on a level playing field we should install latest npm 👍 although I would change it to `npm@latest` to be explicit

@kellyselden npm supports all supported node versions just like we do so when they stop supporting node 6 we will be bumping from node 6 in this template too. If we're concerned by this should we be locking down yarn to a particular major too?

kellyselden: The problem is old addons that don't update their blueprints will drift into a new npm version that doesn't support their node version.

Turbo87: maybe we shouldn't update npm at all and just use the version that is distributed with the Node.js version?

mansona: @Turbo87 I'm not sure that's a great idea because it is always going to be out of date then. Most CI setups install the absolute latest Yarn so it would make sense to do the same thing here too for npm

Personally, I agree with @rwjblue and would prefer to install `npm@latest` but @kellyselden makes a good point.

I think choosing between `npm@latest` and `npm@6` is what we should be doing here and not sticking with the status quo

kellyselden: I vote against npm latest. They make breaking changes for a reason, there's no telling what effect it would have on older addons with old node versions.

rwjblue: Hmm, can we instead just use the default `npm`? Now that Node 6 is gone all supported Node versions use _at least_ `npm@5` (but most are 6+). If a given addon has issues that are fixed with a new `npm` version they can update it themselves....

At the moment we are _basically_ doing the worst possible thing: forcing an **old and unsupported npm on all addons**

rwjblue: > I vote against npm latest. They make breaking changes for a reason, there's no telling what effect it would have on older addons with old node versions.

Then we should do the same with `yarn`. Currently, we install latest `yarn`, and _they_ release breaking changes (e.g. dropping Node 4 and 6) within a major version (much like ember-cli itself).

kellyselden: @rwjblue This PR https://github.com/ember-cli/ember-cli/pull/8645 changed to use the npm version coming from node.

It looks like https://yarnpkg.com/lang/en/docs/install-ci/ yarn is already on Travis machines, so perhaps we could remove that chunk too?

rwjblue: We can test it again, but the `yarn` that comes on the Travis boxes (last I tested) was very old (1.3.2 IIRC) and had a number of errors.

rwjblue:
The underlying issue was fixed by @kellyselden in https://github.com/ember-cli/ember-cli/pull/8645 and was released in ember-cli@3.10.1.

Issue #7666, opened 2018-03-07
Test's dummy app doesn't reexport /src in Module Unification adons


cibernox:
The logic in #7490 has a hole.

When you generate an addon using the module unification layout but with the dummy app inside the test folder in "classic" mode, `isModuleUnification` check thinks the app is un MU layout, and doesn't reexport the contents of the addon's `/src` into `/app` automatically.

Reproduction: https://github.com/cibernox/my-mu-addon

gossi:
I think a more applicable approach for testing is what @mansona has done in https://github.com/mansona/testing-mu where you have in-repo apps that will serve as test apps. I think the blueprint for a mu addon should contain two apps by default: one classic, one mu-app to test the addon in both locations.

ppcano:
https://github.com/ember-cli/ember-cli/pull/8431 should fix this issue

Turbo87:
closing due to https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html

PR #8424, opened 2019-02-12
WIP - fix MU style merging

patricklx:
depends on 
https://github.com/ember-cli/ember-cli/pull/8431

will rebase on it

PR #8431, opened 2019-02-14
Fix MU addon component invocation

patricklx:
MU addons are currently not detected as such, and therefore are always merged into the APP namespace, even when its also an MU project.

This also causes MU addons to be namespaced in tests with MU dummy app, which required the change in the acceptance test

That means the this invocation cannot be used anymore (where basic-thing is an addon component)
```handlebars
{{basic-thing}}WOOT!!{{\basic-thing}}
```
and instead the following must be used
```html
<SomeCoolAddon::BasicThing>WOOT!!</SomeCoolAddon::BasicThing>
```


ppcano:
@patricklx 


> needs emberjs canary version for acceptance test

Should be fixed by https://github.com/ember-cli/ember-cli/pull/8431. Please, test it if you have time.

> * fix ember-module-unification flag detection

I will investigate it. The issue here was commented at https://github.com/ember-cli/ember-cli/pull/8424#discussion_r256795007

Please, could you rename the PR title as `WIP - xxxxx`?


I think it worths to mention that the goal of this PR is to detect an `MU addon`. MU addon components will not be automatically merged in the app namespace and they should be called with the MU syntax.

```
<SomeCoolAddon::BasicThing>WOOT!!</SomeCoolAddon::BasicThing>
```

ppcano:
@patricklx 

Now, this PR is working: an MU app must consume an MU addon component with the new MU syntax.

```
<SomeCoolAddon::BasicThing>WOOT!!</SomeCoolAddon::BasicThing>
```

I think we should also clarify and include a test for the following issue:

**How should a classic app consume an MU addon component?** 


I tried something similar on the `MU addon works with classic dummy app` test of https://github.com/ember-cli/ember-cli/pull/8322 but I had to enable the `EMBER_CLI_MODULE_UNIFICATION` in the dummy classic app but this does not look to be the correct way. 

Let me know your thoughts and if you want to check this case.

patricklx:
mmm, I tested it now, extending this branch, by changing the MU flags checks in addon.js here to this.isModuleUnification().
I found that there is a re-export here, which exports the MU addon components, but only if MU flag is ON
https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L817
`treeForApp` 
 * should return the tree if both are classic layout.
 * return null if both are MU
 * return the re-exported components when addon is MU, but project is classic

2d problem:
 * src not in external tree
so also had to change `treeForSrc` to check `this.isModuleUnification`

and then it passes! :) Is that what we want?

ppcano:
@patricklx 

Yes, I think so. The addon should build its tree based on:
-  `this.isModuleUnification()`: am I an MU or Classic addon? 
-  `this.project.isModuleUnification`: is the app MU or Classic?


Your changes looks correct, we have now more MU tests; they should fail if something is wrong.

> I created a similar test like yours and with this changes:

I would include the mentioned change and the test in this PR. 

This test should validate that a Classic app **without** `EMBER_CLI_MODULE_UNIFICATION` can consume an MU addon component and determines the syntax `<BasicThing>` vs `<ns:BasicThing>` or both (I don't know which one must be used). 


It may be very good if this PR also includes the "inverse" test case: 

**An MU app with `EMBER_CLI_MODULE_UNIFICATION` can consume an Classic addon.**

Including both tests will supercede #8322.

ppcano:
@bartocc This issue may be of your interest. You asked in [Discord](https://discordapp.com/channels/480462759797063690/484527343331704832/544513838217953301). 

> is it possible to use an MU addon in a classic app today ?

bartocc:
thx for the ping @ppcano

lib/broccoli/default-packager.js
```diff
@@ -844,7 +845,9 @@ module.exports = class DefaultPackager {
         annotation: 'Tests To Process',
       });
 
-      if (this.isModuleUnificationEnabled) {
+      // src dir not there when using classic addon with MU dummy app
+      const srcPath = typeof tree === 'string' ? `${tree}/src` : 'src';
```
ppcano: I commented this case at https://github.com/ember-cli/ember-cli/pull/8322/files#r254987783

I think my implementation for skipping this logic provides a little more information about what/when and why this is happening. This issue requires further investigation, but I don't want to be a blocker atm.

Please, I would like you to copy my same implementation using the `dirNotFound` variable, and extend the doc comment if possible. I will continue the investigation of this particular case with `rwjblue`.

ppcano:
This PR supersedes #8322

patricklx:
@ppcano its ready now,

lib/broccoli/default-packager.js
```diff
@@ -844,34 +845,46 @@ module.exports = class DefaultPackager {
...
         // ember-addon
         if (this.name === 'dummy') {
           testSrcTree = 'src';
           destDir = `${this.project.name()}/src`;
+          // Classic Addon with MU dummy app `kitchen-sink-with-mu-dummy-app` will throw: `Dir not found: src` exception
+          // `dirNotFound` will skip this error. Find better way to avoid this error.
+
+          // also, in tests we pass root path to this function
+          // in default-packager/tests-test.js
+          const srcPath = typeof tree === 'string' ? `${tree}/src` : 'src';
```
ppcano: @rwjblue This is the same issue we discussed at https://github.com/ember-cli/ember-cli/pull/8322#discussion_r255207949


**The error is triggered by `broccoli/lib/builder` before `BroccoliFunnel.build` is executed; the `BroccoliFunnel.allowEmpty` routine is never executed.**


I can investigate this but I need guidance on where to start.

ppcano:
@patricklx Great!

https://github.com/ember-cli/ember-cli/pull/8431/commits/a6a94194750515d480472cd2e931f3e73cf292df will be included at https://github.com/ember-cli/ember-cli/pull/8424 as a separate PR, right? LGTM

I think the title and description could be updated to describe better what the PR does: 

The title could be something like: `Fix MU addon component invocation` or `Fix invocation for MU addons` ?

The description could show an example how it was working before and how it works now, in order anyone could see easily what the PR changes instead of having to look the PR code.

------------------------------------------------------

@rwjblue 

- https://github.com/ember-cli/ember-cli/pull/8424 depends on this. We have to move forward this PR first.

- Please, read my comment at https://github.com/ember-cli/ember-cli/pull/8431#discussion_r258838593. I can investigate it now or after being merged. You decide it on your review.

lib/broccoli/default-packager.js
ppcano: @patricklx This should be fixed by https://gist.github.com/ppcano/78e14886c7a660d17eadab709802adb0

Would you mind to include it and check if all the tests pass.

ppcano:
@patricklx could you rebase commits into one?

@rwjblue this PR is a required change for https://github.com/ember-cli/ember-cli/pull/8424 and it is totally ready now (it does not include the [mentioned dirNotFound hack](https://github.com/ember-cli/ember-cli/pull/8322#discussion_r255207949)).

patricklx:
@ppcano @rwjblue rebased, should be ready now!

ppcano:
This issue should close https://github.com/ember-cli/ember-cli/issues/7666

patricklx:
hi @ppcano @rwjblue anything left to do on this?

rwjblue:
sorry @patricklx, will try to review soon (a bit swampped just before EmberConf)

stefanpenner:
As module unification is no more, I believe this can be closed. If I am closing this in error please let me know why, and we can reopen.

PR #8432, opened 2019-02-14
Fix how MU blueprints fetches `ember-source`

ppcano:
Fix the initial implementation of https://github.com/ember-cli/ember-cli/pull/8323.

In some cases, the `ember-source` version was assigned empty because the promise was not fulfilled yet.

I think the test suite did not fail because it may have been using a local `ember-source` version.

Making `locals` to return a Promise looks correct based on the [`locals` documentation](https://ember-cli.com/extending/#locals).

> This hook must return an object or a Promise which resolves to an object. The resolved object will be merged with the aforementioned default locals.

I added an assertion that failed previously.

Initially, I realized about this error when adding the feature to the `octane-blueprint`(the octane blueprints follow this structure), and this error has also been reported at https://github.com/ember-cli/ember-cli/pull/8424/files#r256792430 and https://github.com/ember-cli/ember-cli/pull/8431

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -37,7 +37,7 @@ module.exports = {
   updatePackageJson(content) {
     let contents = JSON.parse(content);
 
-    contents.name = this.locals(this.options).addonName;
+    contents.name = stringUtil.dasherize(this.options.entity.name);
     contents.description = this.description;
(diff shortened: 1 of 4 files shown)


ppcano:
cc @rwjblue Test passes.

PR #8459, opened 2019-02-27
Drop Node 6 support

Turbo87:
According to https://nodejs.org/en/about/releases/ support for Node.js 6 will end on 2019-04-01. Ember CLI v3.9.0 will be released right after that, but `master` is already targetting v3.10.0 so there should be 6+ weeks between Node.js 6 EOL and the release of this PR.

As stated in https://emberjs.com/blog/2016/09/07/ember-node-lts-support.html this will **not** cause/require a major version bump as our official support policy does not change. We recommend that all sub-dependencies, libraries and addons do bump their major version when dropping support for Node.js 6 though. The main reason for us not doing so is to keep the versions of Ember.js and Ember CLI in sync.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,6 @@ jobs:
         - yarn test
 
     - stage: additional tests
-      name: Node.js 6
-      node_js: 6
 
     - name: Node.js 8
       node_js: 8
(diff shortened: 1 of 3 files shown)


kellyselden:
Also our policy of including ember-cli with a tilde (`"ember-cli": "~3.8.0",`) further protects people.

Turbo87:
I've changed the target branch to `next` now and will merge this, to allow https://github.com/ember-cli/ember-cli/issues/8511 to move forward.

PR #8465, opened 2019-02-28
Remove obsolete `BROCCOLI_2` and `SYSTEM_TEMP` experiments

Turbo87:
These experiments have been around for a while and were enabled by default 3 minor releases ago. Since things appear to be working well, I guess it's time to remove those experiments now.

This also has the nice side-effect of making our CI faster again :)

/cc @oligriffiths


oligriffiths:
Works for me. The fallback will still works and output an error so I think we’re covered.

Turbo87:
yep, it's not changing any behavior, just removing the env vars for completely opting out of Broccoli 2 and the system temp folder

oligriffiths:
Dope. Ship it!

Turbo87:
closing in favor of https://github.com/ember-cli/ember-cli/pull/8470

PR #8469, opened 2019-03-03
Move all package related path resolution to resolve-package-path.

stefanpenner:
This ensures hash-for-dep, ember-cli-verison-checker, and ember-cli all have the same behavior and potentially the same cache.

This also should in theory make the resolution aspects yarn PNP aware.
https://github.com/stefanpenner/resolve-package-path/blob/e995d8c74bb94c21aa06d1e3fbb131ee4aa9557f/tests/index-tests.js#L69

### DIFF
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -12,91 +12,10 @@ const Errors = require('./errors');
 const PackageInfo = require('./package-info');
 const NodeModulesList = require('./node-modules-list');
 const logger = require('heimdalljs-logger')('ember-cli:package-info-cache');
+const resolvePackagePath = require('resolve-package-path');
 
-let realFilePathCache;
-let realDirectoryPathCache;
-
-/**
- * Resolve the real path for a file, return null if does not
- * exist or is not a file or FIFO, return the real path otherwise.
- *
- * @private
- * @method getRealFilePath
- * @param  {String} filePath the path to the file of interest
- * @return {String} real path or null
- */
-function getRealFilePath(filePath) {
-  let realPath;
-
-  try {
-    realPath = realFilePathCache[filePath];
-
-    if (realPath) {
-      return realPath;
-    }
-
-    let stat = fs.statSync(filePath);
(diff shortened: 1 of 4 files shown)


lib/models/package-info-cache/index.js
```diff
@@ -12,91 +12,10 @@ const Errors = require('./errors');
 const PackageInfo = require('./package-info');
 const NodeModulesList = require('./node-modules-list');
 const logger = require('heimdalljs-logger')('ember-cli:package-info-cache');
+const resolvePackageCache = require('resolve-package-path');
```
davecombs: Suggest calling this resolvePackagePath instead, just to match the package name.

lib/models/package-info-cache/index.js
```diff
@@ -561,7 +478,7 @@ class PackageInfoCache {
       if (!pathFailed) {
         // we have a valid realPath
         let packageJsonPath = path.join(realPath, PACKAGE_JSON);
-        let pkgfile = getRealFilePath(packageJsonPath);
+        let pkgfile = resolvePackageCache(realPath);
```
davecombs: Given what you have at lines 17 & 18, shouldn't this remain as 'getRealFilePath'?

stefanpenner: yes, oops moving to quick.

davecombs: whoaaaaaa!!! :-)

stefanpenner:
Thanks @turbo87

chriskrycho:
Going to be digging through *why*, but this [breaks] ember-cli-typescript's blueprint tests. (I need to see if it breaks actually ember-cli-typescript running in non-test scenarios; that's up next.)

[breaks]: https://github.com/typed-ember/ember-cli-typescript/issues/670

chriskrycho:
This appears to be a problem *only* with our tests—things work as expected in a real app; I'm evaluating the tests to see if I can figure out what they were depending on that is no longer working after this change.

PR #8504, opened 2019-03-16
Use `prettier` with ESLint integration to format the code

Turbo87:
I guess the title speaks for itself. For those that don't know that prettier is: http://prettier.io/

also, greetings from the plane to PDX for EmberConf 😂


tests/acceptance/addon-smoke-test-slow.js
```diff
@@ -57,120 +57,130 @@ describe('Acceptance: addon-smoke-test', function() {
...
 
-    expect(result.code).to.eql(0);
-  }));
+      expect(result.code).to.eql(0);
+    })
+  );
+
+  it(
+    'npm pack does not include unnecessary files',
+    co.wrap(function*() {
```
kellyselden: ouch

kellyselden:
2 thoughts

This makes the code look overall uglier, not prettier (subjective of course).

Is this actually solving a problem we have?

That said, this is not a hill I want to die on, and I will not oppose a merge.

tests/acceptance/addon-smoke-test-slow.js
Turbo87: we can set `printWidth: 120` if that helps?

rwjblue:
> This makes the code look overall uglier, not prettier

I couldn't disagree more 😸 

> Is this actually solving a problem we have?

IMHO, yes. The codebase currently is very inconsistent style wise, this leads to confusion as folks attempt to add new features / changes.

PR #8517, opened 2019-03-19
[ENHANCEMENT] Adding --output-path to test command

step2yeung:
Adding --output-path to `test` command to be on parity with `build` and `serve`

One difference is there will be no default for `test` command's `output-path` option.


stefanpenner:
@step2yeung LGTM, this fixes a symmetry issue between commands. Thanks for fixing!

stefanpenner:
@rwjblue should we backport to beta?

PR #8538, opened 2019-03-29
BUGFIX [Close #8536] remove chrome flag on Circle CI builds

### LINKED ISSUE #8536: Circle CI tests fail on Chrome with the testem `disable-dev-shm-usage` flag enabled
Circle CI tests fail on Chrome `>= 71` with the testem `disable-dev-shm-usage` flag enabled with the following error:

`FATAL:gpu_data_manager_impl_private.cc(892)] The display compositor is frequently crashing. Goodbye`

**Full error**
```
not ok 1 Chrome - [undefined ms] - error
    ---
        message: >
            Non-zero exit code: null
            Stderr: 
             Fontconfig warning: "/etc/fonts/fonts.conf", line 100: unknown element "blank"
            
            DevTools listening on ws://127.0.0.1:38833/devtools/browser/b48fcc54-a189-4e54-8c45-e2c46e81f798
            [0326/141652.369424:FATAL:gpu_data_manager_impl_private.cc(892)] The display compositor is frequently crashing. Goodbye.
            Failed to generate minidump.
            
        Log: |
            { type: 'error', text: 'Non-zero exit code: null' }
            { type: 'error',
              text:
               'Fontconfig warning: "/etc/fonts/fonts.conf", line 100: unknown element "blank"\n\nDevTools listening on ws://127.0.0.1:38833/devtools/browser/b48fcc54-a189-4e54-8c45-e2c46e81f798\n[0326/141652.369424:FATAL:gpu_data_manager_impl_private.cc(892)] The display compositor is frequently crashing. Goodbye.\nFailed to generate minidump.' }
    ...

1..1
# tests 1
# pass  0
# skip  0
# fail  1
Testem finished with non-zero exit code. Tests failed.
```

## Background

A new Google Chrome flag was introduced that will disable use of shm memory in favor of using the `tmp` directory as described in [this chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=736452#c64). That bug references [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=715363), which is pointing out that by default, docker sets the `/dev/shm` size to 64MB which often isn't enough for big web pages. 

So disabling shm in `testem.js` helps tests running in Docker not run out of shared memory as it'll use a different memory mapping scheme:
https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/testem.js#L17

While this flag is useful in some situations, for example on Travis CI, on Circle CI the tests run with a larger amount of memory, so disabling the shm system isn't required, and actually causes another problem. 

CircleCI apparently runs the docker images with a higher `/dev/shm` limit, as indicated in this forum post (https://discuss.circleci.com/t/shared-memory-warnings/11761/7). I've confirmed that on the free tier the size is to `30G`.

## Fix

The fix would be to check for the existence of a CircleCI envionment variable set in all CI runs, and disable it similar to how we enable the `--no-sandbox` flag today.

```javascript
process.env.CIRCLECI ? null : '--disable-dev-shm-usage',
```

Changes to testem.js to include this flag were done in the following commit: https://github.com/ember-cli/ember-cli/commit/9dbd15ef223889db6c49899cb6f8ed641cc9cafe#diff-f7e2d0d5944207cb98b51991f8da57b0

I'll implement this fix myself tomorrow 🍻

0xadada:
checks for the existence of a CircleCI envionment variable `CIRCLECI=true` which is set in all CI runs, and disable the flag similar to how we enable the `--no-sandbox` flag for CI runs.

Different flag, inverse logic.

Fixes https://github.com/ember-cli/ember-cli/issues/8536

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -12,9 +12,10 @@ module.exports = {
       ci: [
         // --no-sandbox is needed when running Chrome inside a container
         process.env.CI ? '--no-sandbox' : null,
+        // --disable-dev-shm-usage shouldn't be used in Circle CI builds
+        process.env.CIRCLECI ? null : '--disable-dev-shm-usage',
         '--headless',
         '--disable-gpu',
-        '--disable-dev-shm-usage',
         '--disable-software-rasterizer',
         '--mute-audio',
         '--remote-debugging-port=0',


kellyselden:
Is there not a case that will work for both to avoid this Circle specific thing? If not, could we remove the line from testem.js, and instead re-apply the var in .travis.yml somehow?

I recognize the problem, but I'd like to avoid sprinkling CI code across the blueprints.

bendemboski:
It very much looks to me like we can just remove the flag unconditionally. On CircleCI paid, I have verified that `/dev/shm` is huge -- plenty big to safely use. @0xadada has confirmed the same on CircleCI free (although @acorncom, you said you might have seen differently? We should confirm).

I just ran a debug travis build of a public/free repo (my fork of `ember-ajax`), and it looks like there's plenty of `dev/shm` there too (3GB):

```
travis@travis-job-2f58bfd9-6677-4bc0-acce-50e93987005e:~/build/bendemboski/ember-ajax$ df /dev/shm
Filesystem     1K-blocks  Used Available Use% Mounted on
none             3829684     0   3829684   0% /run/shm
```

I suppose it's possible that debug builds have more `/dev/shm` than non-debug free builds, and I can't find any documentation describing the memory limits. When I have time I can push a branch that logs out the `/dev/shm` size in a non-debug container just to double-check.

Of course there could be other CI systems (GitLab?)/custom infrastructure where `/dev/shm/` is the default tiny size, but people can always customize their `testem.js`.

@stefanpenner you originally added this flag -- did you observe problems anywhere, or was it just based on the general knowledge/various bugs about docker's default tiny size for `/dev/shm`? Now that it's causing failures with Chrome stable, do you see a reason not to remove it?

bendemboski:
Confirmed -- a free travis build without debug enabled has the same 4GB of `/dev/shm`.

0xadada:
> Is there not a case that will work for both to avoid this Circle specific thing? If not, could we remove the line from testem.js, and instead re-apply the var in .travis.yml somehow?
> 
> I recognize the problem, but I'd like to avoid sprinkling CI code across the blueprints.

agreed, if @stefanpenner concurs, i'll update the PR to simply remove it. It seems to me like we can remove it, but this initial approach was more conservative.

PR #8539, opened 2019-03-30
[ENHANCEMENT] Added `--pods` alias to generator's pod option

yohanmishkin:
When we talk about pods it seems most common to use or hear the word "pods". Rarely do I hear mention of a singular pod though clearly it exists conceptually and in reality. 

When using ember-cli's generator (and destroyer) in a project with pods file layout I often find myself trying to invoke a non-existent `--pods` command option.

This PR adds an alias to the `--pod` generator option.

_Notes_
- I did not add any tests because I couldn't find any precedent for testing option aliases.
- ~I did not generate a new help `help.txt` fixture because I wasn't sure how.~


kellyselden:
Seems fine to me.

PR #8545, opened 2019-04-02
WIP Azure pipelines

mike-north:
This is an experiment to see if Azure Dev Ops may be a better option for ember-cli's CI pipeline than what's currently being used.

My suggestion is to run it side-by-side with the existing tools, to see if it's more reliable, faster, and surfaces more actionable feedback in the github UI

### DIFF
--- a/blueprints/app/files/testem.js
+++ b/blueprints/app/files/testem.js
@@ -1,6 +1,7 @@
 module.exports = {
   test_page: 'tests/index.html?hidepassed',
   disable_watching: true,
+  browser_start_timeout: 60000,
   launch_in_ci: [
     'Chrome'
   ],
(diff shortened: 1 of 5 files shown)


blueprints/app/files/testem.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   test_page: 'tests/index.html?hidepassed',
   disable_watching: true,
+  browser_start_timeout: 60000,
```
mike-north: I don't understand the particulars yet, but there's a race condition between testem starting up and the browser connecting. The default timeout is occasionally not a sufficient wait, and the end result is having to kick the job again to get a successful test run

igorT: Ah nice, I was struggling with this, but it seems to only error on ubuntu and doesn't on OS X

rwjblue:
We've adopted GH actions so I'm going to go ahead and close this, but please take a look and let me know if you still think we should investigate this...

PR #8548, opened 2019-04-03
Use loadBrocfile from Broccoli & support ESM & Typescript

oligriffiths:
This PR implements the `loadBrocfile()` function from Broccoli which comes with ESM and TypeScript support.
I've also updated the default blueprints to ESM by default.

Tests added with fixtures.


tests/fixtures/build/esm/ember-cli-build.js
```diff
@@ -0,0 +1,7 @@
+import funnel from 'broccoli-funnel';
+
+export default function() {
+    return funnel('app', {
```
kellyselden: nitpick but can you use our convention two spaces in these fixture files?

oligriffiths:
@kellyselden thanks for the review, will work up some numbers and get back to you.

oligriffiths:
@kellyselden So ESM support comes from https://github.com/broccolijs/broccoli/blob/master/lib/load_brocfile.js#L36

Regarding performance, I added some heimdall metrics, here are the results:

New ember build:
* cjs: 300ms
* esm: 450ms
* delta: 150ms
* ts: 2000-3000ms
* delta: 1700-2700ms

Travis-web build:
* cjs: 950ms
* esm: 1600ms
* delta: +700ms
* ts: 3000-4000ms
* delta: +2050-3050ms

So I can confirm that on my machine it does take longer to load the build file as esm, so yeah not great, albeit a one off cost, thoughts?
The typescript builds are a little less predictable and vary more, however they don't have as big a delta as I thought they would.

Might be best to limit ESM loading to `ember-cli-build.mjs` files, as we already have a switch for typescript for `.ts` files.
The thing is, if we want people to start authoring plugins in at least ESM, then it needs to be the default.

Please let me know your thoughts.

locks:
Closing as stale. Thanks for all the hard work Oli!

PR #8551, opened 2019-04-03
Expose build annotation information to the Addon's preBuild method.

chriseppstein:
This lets the preBuild method prepare the addon for the next build iteration.

For instance, when a `rebuild` occurs, the `changedFiles` can be inspected. Based on the file paths or extensions, a persistent addon might be able to set a flag to make the next build a no-op because it can easily determine that the files changed aren't related to the addon's purpose.

I would like to use this in ember-cli-eyeglass to avoid doing any work during a rebuild unless one of the files that's changed is a Sass file. Similarly, I can imagine that if the only file that changed is a Sass file, there's probably no reason to run babel addons.

### DIFF
--- a/lib/models/addon.js
+++ b/lib/models/addon.js
@@ -1602,9 +1602,16 @@ let addonProto = {
 /**
   This hook is called before a build takes place.
 
+  It is passed information about what triggered the build.
+  For the initial build, the list of changed files is empty.
+
   @public
   @method preBuild
-  @param {Object} result Build object
+  @param {Object} annotation Build annotation
+    type: 'rebuild' | 'initial'
+    reason: 'watcher' | 'build'
+    primaryFile: string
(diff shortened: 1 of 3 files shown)


stefanpenner:
I like the idea. Let me kick the tires on it today.

tomdale:
Is the provided metadata consistent across all of the different watcher implementations we support?

rwjblue:
I rebased this against current `master` branch (incorporating a number of fixes), and added some tests (refactoring the builder tests to assert on the arguments to each hook).

rwjblue:
OK, now that CI is green and we have some tests, I'd like to see what folks think about this.

@stefanpenner - What do you think?

bertdeblock:
Is this still something we want to work on?

bertdeblock:
Going to close for now.

Issue #8511, opened 2019-03-18
Modernize codebase


Turbo87:
### Modernization

As NODE 6 will be dropped soon, we would like to take this opportunity to upgrade the code-base to modern features, such as native promises and async/await.

Porting the code base to Native Promises, and async-await syntax will provide us with:

* better promise based async callstacks (in Node 12) and overall better debugging.

To accomplish this, we should (when possible)

* utilize [async/await](https://javascript.info/async-await) 
* dropping `co` usage in tests for `async/await`
* dropping RSVP/custom Promise code helpers
* use `util.promisify` instead of `RSVP.denodify`
* for any async IO, while leaving sync IO as is, attempt to use native promise variant (such as `fs.promises`)

Note: Node 8 does not support `Promise.prototype.finally` for this we will need to use https://www.npmjs.com/package/promise.prototype.finally

In addition to improve async code, feel free to cleanup and refactor to any other missed ES2015+ features.

### How to Contribute:

* claim a file to upgrade
* submit that updated file to this branch <branch-name>
* claim another file
* etc

### Test Files:

|complete|owner|file|
|---------|-------|:------|
|<ul><li>[ ] assigned </li><li>[ ] complete</li><ul>||[tests/unit/tasks/install-blueprint-test.js](https://github.com/ember-cli/ember-cli/tree/master/tests/unit/tasks/install-blueprint-test.js)|
|<ul><li>[x] assigned </li><li>[x] complete</li><ul>||[tests/unit/tasks/bower-install-test.js](https://github.com/ember-cli/ember-cli/tree/master/tests/unit/tasks/bower-install-test.js)|
|<ul><li>[ ] assigned </li><li>[ ] complete</li><ul>||[tests/unit/tasks/test-server-test.js](https://github.com/ember-cli/ember-cli/tree/master/tests/unit/tasks/test-server-test.js)|
|<ul><li>[ ] assigned </li><li>[ ] complete</li><ul>||[tests/unit/tasks/serve-test.js](https://github.com/ember-cli/ember-cli/tree/master/tests/unit/tasks/serve-test.js)|
|<ul><li>[ ] assigned </li><li>[ ] complete</li><ul>||[tests/unit/tasks/server/middleware/tests-server-test.js](https://github.com/ember-cli/ember-cli/tree/master/tests/unit/tasks/server/middleware/tests-server-test.js)|
[...]

PR #8552, opened 2019-04-03
analytics-test: Use async/await

RichardOtvos:
This is one of the files that requested to be modernized with async await in a #8511 . 
Since this is my first PR just wanted to make sure that, this is what you need @Turbo87 , and after that I happy take some more files.

Can I check "assigned" on the files I'm working on in the list in issue #8551 ?

### DIFF
--- a/tests/unit/analytics-test.js
+++ b/tests/unit/analytics-test.js
@@ -36,9 +36,8 @@ describe('analytics', function() {
     command = null;
   });
 
-  it('track gets invoked on command.validateAndRun()', function() {
-    return command.validateAndRun([]).then(function() {
-      expect(called, 'expected analytics.track to be called').to.be.true;
-    });
+  it('track gets invoked on command.validateAndRun()', async function() {
+    await command.validateAndRun([]);
+    expect(called, 'expected analytics.track to be called').to.be.true;
   });
 });


tests/unit/analytics-test.js
```diff
@@ -34,9 +34,8 @@ describe('analytics', function() {
     command = null;
   });
 
-  it('track gets invoked on command.validateAndRun()', function() {
-    return command.validateAndRun([]).then(function() {
-      expect(called, 'expected analytics.track to be called').to.be.true;
-    });
+  it('track gets invoked on command.validateAndRun()', async function() {
+    await command.validateAndRun([]);
+    return expect(called, 'expected analytics.track to be called').to.be.true;
```
Turbo87: the `return` here is no longer necessary. we only needed to return in the previous state because of the Promise, but since the async function already implicitly returns a Promise the explicit return is obsolete.

RichardOtvos: I know, I was debating it if I should add it, I wasn't sure if we want to return a Promise resolving to undefined. I feel this might be more explicit (although admittedly more noisy too).

I don't have strong opinions on this question, so I'm happy to remove it if you want.

Turbo87:
@RichardOtvos FYI I've rebased the branch and squashed the commits to retrigger TravisCI. we were apparently missing https://github.com/ember-cli/ember-cli/commit/be76597c2d817b1c54acd2f12f5d7080e0b04b35 before 🙈

PR #8557, opened 2019-04-06
#8511 Modernize install-blueprint-task.js test

RichardOtvos:


RichardOtvos:
@Turbo87 Do we want to replace the the `.to.eventually.equal` type of assertions with an await and `to.equal` type ones as part of this task?

Like this:
```
it('resolves "foobar" by looking up the "foobar" blueprint locally', function() {
      let foobarBlueprint = { name: 'foobar blueprint' };
      td.when(task._lookupBlueprint('foobar')).thenResolve(foobarBlueprint);
      return expect(task._resolveBlueprint('foobar')).to.eventually.equal(foobarBlueprint);
});
```

with something like this
```
it('resolves "foobar" by looking up the "foobar" blueprint locally', async function() {
      let foobarBlueprint = { name: 'foobar blueprint' };
      td.when(task._lookupBlueprint('foobar')).thenResolve(foobarBlueprint);
      
      const resolvedBlueprint = await task._resolveBlueprint('foobar');
      expect(resolvedBlueprint).to.equal(foobarBlueprint);
});
```

Turbo87:
yes, I think something like this makes sense:

```js
it('resolves "foobar" by looking up the "foobar" blueprint locally', async function() {
      let foobarBlueprint = { name: 'foobar blueprint' };
      td.when(task._lookupBlueprint('foobar')).thenResolve(foobarBlueprint);
      expect(await task._resolveBlueprint('foobar')).to.equal(foobarBlueprint);
});
```

RichardOtvos:
@Turbo87 Do you see any benefits in changing the way we handle testing promise rejections, from
```javascript
      // normal function (and what we currently have in the codebase)
      return expect(task.ensureBower()).to.be.rejectedWith('foobar');
```

to

```javascript
      //async function
      await expect(task.ensureBower()).to.be.rejectedWith('foobar');
```

The benefit I see in the second way, is that it would mean you always use async functions when you test Promises (rather than only use them when testing for resolved promises), but I'm not sure if it's worth switching for that alone.

I've experimented with some other ways to test rejections, but everything I could think of was a lot more verbose or invasive than chai-as-promised `.rejectedWith` assertion. 
I'm open to ideas if you can think of anything else. :)

Turbo87:
@RichardOtvos yep, that second snippet looks 👌 to me! :)

tests/unit/tasks/install-blueprint-test.js
```diff
@@ -1,18 +1,18 @@
...
 
 const expect = require('../../chai').expect;
+const promisify = require('util').promisify;
 const mkTmpDirIn = require('../../../lib/utilities/mk-tmp-dir-in');
 const InstallBlueprintTask = require('../../../lib/tasks/install-blueprint');
 
 let root = path.join(__dirname, '../../..');
 let tmproot = path.join(root, 'tmp');
-const remove = RSVP.denodeify(fs.remove);
+const remove = promisify(fs.remove);
```
Turbo87: I think this can be removed completely. `fs.remove()` (at least from `fs-extra`) will always return a Promise by default now.

tests/unit/tasks/install-blueprint-test.js
```diff
@@ -33,28 +33,28 @@ describe('InstallBlueprintTask', function() {
...
-      return expect(task._resolveBlueprint('foobar')).to.eventually.equal(foobarBlueprint);
+      expect(await task._resolveBlueprint('foobar')).to.equal(foobarBlueprint);
     });
 
     it('rejects invalid npm package name "foo:bar"', function() {
       let error = new Error('foobar not found');
       td.when(task._lookupBlueprint('foo:bar')).thenReject(error);
 
-      return expect(task._resolveBlueprint('foo:bar')).to.be.rejectedWith(error);
+      expect(task._resolveBlueprint('foo:bar')).to.be.rejectedWith(error);
```
Turbo87: `rejectedWith` calls need an `await`

RichardOtvos: Yeah, wanted to get you opinion about how to handle test cases with `.to.rejectedWith` assertions first. Will update the file shortly.

RichardOtvos:
This should be ready, let me know if you require any more changes.

PR #8558, opened 2019-04-06
#8511 Mondernize tasks/test.js

RichardOtvos:
We can just use native promises with the new node version, no need for RSVP.

### DIFF
--- a/lib/tasks/test.js
+++ b/lib/tasks/test.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const Task = require('../models/task');
-const Promise = require('rsvp').Promise;
 const SilentError = require('silent-error');
 
 class TestTask extends Task {


Turbo87:
unfortunately it's not quite so easy. `RSVP.Promise` has a `.finally()` method, but the native `Promise` class does not (yet). depending on whether the method that we return this Promise from is public API or not this might be a breaking change.

can you check if any of the methods that return Promises need to be considered public API?

RichardOtvos:
@Turbo87 As far as I see, TestTask is not part of the public API (it's not in the API docs either), and only used by the TestServerTask as a base class (which also overwrites the method using the promise).

That said, I'm not really familiar with Ember Addon development, is there a possibility it is used by third party addons?

PR #8560, opened 2019-04-06
#8511 Changes install-blueprint.js to use async/await

RichardOtvos:
What it says in the title. Please take a look, and see if there are any patterns you'd like me to change.

### DIFF
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -3,10 +3,10 @@
 const fs = require('fs-extra');
 const Blueprint = require('../models/blueprint');
 const Task = require('../models/task');
-const util = require('util');
 const temp = require('temp');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


lib/tasks/install-blueprint.js
```diff
@@ -43,26 +43,30 @@ class InstallBlueprintTask extends Task {
...
+    return blueprint.install(installOptions);
   }
 
-  _resolveBlueprint(name) {
+  async _resolveBlueprint(name) {
     name = name || 'app';
     logger.info(`Resolving blueprint "${name}" ...`);
 
-    return this._lookupLocalBlueprint(name).catch((error) => this._handleLocalLookupFailure(name, error));
+    let blueprint;
```
locks, line 55: It was this or `return await` from both branches or the tests fail.

PR #8563, opened 2019-04-08
Drop Node 6 support

Turbo87:
`ember-source@2.10.0-beta.1` has already dropped support for Node 6, which makes our test suite fail 😱 

Merging this PR will break our support policy by having `master` (and `beta`) not be compatible with Node.js 6 anymore before its official EOL (2019-04-30), but since our parent project has already broken that promise I guess this is fine...


.travis.yml
```diff
@@ -10,6 +10,7 @@ branches:
     - master
     - beta
     - release
+    - next
```
rwjblue: Probably can remove?

rwjblue: Though I guess it doesn't matter much...

Turbo87: will remove it once this PR is merged. otherwise PRs against `next` will not run CI.

Turbo87:
alright. we're doing this!

rwjblue:
:tada:

stefanpenner:
👯‍♂

PR #8564, opened 2019-04-08
Adds babel-eslint as the default ESlint parser

pzuraq:
`ember-cli-babel` now supports class fields and decorators out of the
box, and we're currently on track to ship decorators as on by default in
Ember 3.10. However, users will get ESLint errors if they attempt to use
decorators, since it's default parser does not yet support them.

This PR adds the `babel-eslint` parser, which will allow things to work
as expected and has been what we've been recommending to early adopters
for some time.

It's worth noting that the next major version of `babel-eslint` seems to
be breaking, and will apparently _require_ a `.babelrc`. I'm not sure
whether or not we'll be able to support that, given `ember-cli-babel`
handles that type of configuration for us. We can probably continue
using `babel-eslint@10` until that gets sorted out.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module'
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -20,6 +20,7 @@
   "devDependencies": {
     "@ember/jquery": "^0.6.0",
     "@ember/optional-features": "^0.7.0",
+    "babel-eslint": "^10.0.2",
     "broccoli-asset-rev": "^3.0.0",
     "ember-ajax": "^5.0.0",
     "ember-cli": "~<%= emberCLIVersion %>",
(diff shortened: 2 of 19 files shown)


Turbo87:
it's a little worrying that we're moving towards a solution where we already know that it will break in the near future. is there any more information available on that change in `babel-eslint`?

pzuraq:
The first v11 beta is out so it seems like it should be coming sooner rather than later. I agree it's worrying, but I also think this is going to be a deeper problem, if we add a `.babelrc` then folks will probably expect Babel to respect it as well. That would be a much larger lift for `ember-cli-babel` I think, but possible without breaking changes. Do you think it'd make sense to allow `ember-cli-babel` to read/use `.babelrc` in apps?

Turbo87:
> Do you think it'd make sense to allow ember-cli-babel to read/use .babelrc in apps?

that is a significant refactoring effort that we would have to move the whole userbase to and iirc there were quite a few issues with this approach when we did the Babel 6 migration which, I think, first introduced those files.

might become easier with embroider, but that's still some time away...

rwjblue:
> it's a little worrying that we're moving towards a solution where we already know that it will break in the near future.

Totally agree. I don't see that we have much choice at the moment though. Without this change (in the abstract) we can't even use the syntax. We should reach out to the Babel folks for guidance.

PR #8565, opened 2019-04-09
Modernize git-init.js and git-init-test.js

xg-wang:
cc @Turbo87 

Epic: #8511


tests/unit/tasks/git-init-test.js
```diff
@@ -27,58 +24,52 @@ describe('git-init', function() {
...
-      process.chdir(tmpdir);
-    });
+    let tmpdir = await mkTmpDirIn(tmproot);
+    process.chdir(tmpdir);
   });
 
   afterEach(function() {
     process.chdir(root);
-    return remove(tmproot);
+    return fs.remove(tmproot);
```
Turbo87: can you turn that `return` into an `await`?

PR #8566, opened 2019-04-09
#8511 Mondernize build-watch.js build-watch-test.js

xg-wang:
`defer` is used by lib/tasks/build-watch.js and lib/tasks/server.js

@Turbo87 how do you feel about introducing a new dependency here?


lib/tasks/build-watch.js
```diff
@@ -14,12 +14,12 @@ class BuildWatchTask extends Task {
     this._runDeferred = null;
   }
 
-  run(options) {
+  async run(options) {
```
xg-wang: @stefanpenner @rwjblue I think converting a public API to async function also needs the shim correct? Node will implicitly wrap the return value to a Promise.

PR #8567, opened 2019-04-09
[Modernize] shim Promise.prototype.finally

xg-wang:
Shim Promise.prototype.finally at cli entrance. I guess this should be enough.
cc @Turbo87

### DIFF
--- a/lib/utilities/promise.js
+++ b/lib/utilities/promise.js
@@ -1,0 +1,5 @@
+'use strict';
+
+require('promise.prototype.finally').shim();
+
+module.exports = Promise;
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "npm-package-arg": "^6.1.0",
     "portfinder": "^1.0.20",
     "promise-map-series": "^0.2.3",
+    "promise.prototype.finally": "^3.1.0",
     "quick-temp": "^0.1.8",
     "resolve": "^1.10.0",
     "resolve-package-path": "^1.2.6",
--- a/tests/unit/utilities/promise-test.js
+++ b/tests/unit/utilities/promise-test.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const Promise = require('../../../lib/utilities/promise');
+const expect = require('chai').expect;
+const semver = require('semver');
+
+const isNode8 = semver.satisfies(process.version, '^8');
+
+(isNode8 ? describe : describe.skip)('Promise', function() {
+  it('should shim Promise.prototype.finally', function(done) {
+    let fulfillmentValue = 1;
+    let promise = Promise.resolve(fulfillmentValue);
+
+    promise.finally(function() {
+      expect(arguments.length).to.equal(0);
+      done();
+    });
+  });
+});


RichardOtvos:
This would certainly make it easier to modernize our code, without breaking the public API, and solve questions like the one in the [tasks/test.js](https://github.com/ember-cli/ember-cli/pull/8558) PR.

stefanpenner:
One issue is, we have more than one entry point. As some folks use ember-cli as a lib or pull out test helpers...

xg-wang:
@RichardOtvos the problem here is not for internal `ember-cli` code using `finally`, but for external users.

@stefanpenner Is using as a lib referring to use cases other than ember addons? This change covers all ember commands use cases. For ember-cli as a lib use cases, I think the best we can do is to document this is breaking change, and the alternative is to try-catch async/await or do the similar shim.

RichardOtvos:
@xg-wang I might be wrong here, but the whole issue in the ticket I linked to is that while the Test task returns a promise (which we don't use internally with finally), there might be others using it, and relying  on the Promise having a finally method

xg-wang:
@RichardOtvos Yes that's what I'm thinking and confirming with Stef.

I think we have these 2 options:
1. Add shim to every pormise-export-file-entry (effectively swapping require rsvp to the new lib)
2. Add at cli entry as this RB does, this will not break for ember users since they always use ember-cli for ember related stuff. It will break for other users so we need an explicit release note mentioning migration path (adding the similar shim here)

kellyselden:
2. is difficult because of lockstep. We cannot release a breaking v4 version of cli without also releasing a v4 of emberjs.

xg-wang:
So should we merge this PR, then for each file where we remove `const Promise = RSVP.Promise;` add the shim?

rwjblue:
> So should we merge this PR, then for each file where we remove const Promise = RSVP.Promise; add the shim?

Ya, I think that is plausible, though I guess I'd prefer to have a simpler replacement. Something like:

```js
// lib/utilities/promise.js
require('promise.prototype.finally').shim();

module.exports = Promise;
```

Then swap each usage of `const Promise = RSVP.Promise` or `const Promise = require('rsvp').Promise` with `const Promise = require('../utilities/promise');`.


@stefanpenner - Are we certain that the only thing missing from `RSVP.Promise` is `.finally`?

Turbo87:
@rwjblue according to https://github.com/tildeio/rsvp.js/blob/92aff41f2c8b9d83b3bca09d07d10eb1c5e3dafd/lib/rsvp/promise.js `then`, `catch` and `finally` are the only public prototype methods

I think it would be fine to go over https://emberobserver.com/code-search?codeQuery=require(%27ember-cli%2F, ignore the addons with only 1 result, and use all of those existing import paths as potential entry points that need the polyfill.

I'm not sure about the `module.exports = Promise;` part. why do you want to reexport the native Promise class?

rwjblue:
> I'm not sure about the module.exports = Promise; part. why do you want to reexport the native Promise class?

* its pretty weird/odd/non-standard to import for side-effects
* importing the promise base class is easy enough
* the technique I suggested allows us to prevent accidental regressions via linting that forbids the `Promise` global globally (this will ensure **all** files that use `Promise` are properly shimming `.finally`)

Turbo87:
>     * its pretty weird/odd/non-standard to import for side-effects

yeah, but that is exactly what we're doing here

>     * importing the promise base class is easy enough

yeah, but it won't get used most of the time when we use async/await

rwjblue:
True indeed. How do we ensure that all files with async/await require this shim then?

Turbo87:
> How do we ensure that all files with async/await require this shim then?

by making sure that all (public) entrypoints (see above) call the shim before they do anything else

rwjblue:
> by making sure that all (public) entrypoints (see above) call the shim before they do anything else

Generally seems fine, but I do not believe that surveying addons alone is sufficient to determine public entrypoints.  I believe the only "real" way to ensure this isn't a breaking change is via some sort of custom rule to ensure the shim is required if _any_ async/await or `Promise` is used within a non-test file.

xg-wang:
Should we close this PR since node 8 EOLs soon?

rwjblue:
@xg-wang - ya, probably but we should work on removing existing usages of the shim once we drop support for Node 8...

PR #8570, opened 2019-04-09
#8511 Modernizes serve-test.js

RichardOtvos:
Makes serve-test.js use async await

### DIFF
--- a/tests/unit/tasks/serve-test.js
+++ b/tests/unit/tasks/serve-test.js
@@ -6,8 +6,6 @@ const Builder = require('../../../lib/models/builder');
 const MockProject = require('../../helpers/mock-project');
 const expect = require('chai').expect;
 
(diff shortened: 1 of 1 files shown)


tests/unit/tasks/serve-test.js
```diff
@@ -73,23 +70,23 @@ describe('serve task', function() {
...
     });
 
-    it(`Serves ember app from given path`, function() {
+    it(`Serves ember app from given path`, async function() {
       runServeTask('docs');
-      return RSVP.resolve().then(() => {
-        expect(ui.output).to.be.contains('– Serving on');
-      });
+
+      await Promise.resolve();
```
Turbo87: that looks like a very strange test to me, but it already was like that before, so I guess that's fine?! 😅

PR #8571, opened 2019-04-09
Support npm blueprints with versions

kellyselden:
There are various reasons you may want to generate an older blueprint (`ember new -b @ember/octane-app-blueprint@0.17.2`).

The main use I have planned is using this in ember-cli-update https://github.com/ember-cli/ember-cli-update/issues/546.

### DIFF
--- a/lib/tasks/install-blueprint.js
+++ b/lib/tasks/install-blueprint.js
@@ -10,7 +10,7 @@ const path = require('path');
 const merge = require('ember-cli-lodash-subset').merge;
 const execa = require('../utilities/execa');
 const SilentError = require('silent-error');
-const validateNpmPackageName = require('validate-npm-package-name');
+const npa = require('npm-package-arg');
 
 const logger = require('heimdalljs-logger')('ember-cli:tasks:install-blueprint');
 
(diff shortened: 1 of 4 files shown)


kellyselden:
Hmm my new acceptance test works locally but fails on CI.

lib/tasks/install-blueprint.js
```diff
@@ -77,20 +77,26 @@ class InstallBlueprintTask extends Task {
   _handleLookupFailure(name, error) {
     logger.info(`Blueprint lookup for "${name}" failed`);
 
-    if (!validateNpmPackageName(name).validForNewPackages) {
+    // I wish I didn't have to do this https://github.com/npm/validate-npm-package-name/issues/22
```
rwjblue: Hmm, is there a reason we don't use https://github.com/npm/npm-package-arg? It is _explicitly_ intending to be "parse anything you can pass to `npm install xxxx`", and internally uses `validate-npm-package-name`.

kellyselden: fixed.

kellyselden:
@rwjblue Do you have any additional problems with this?

lib/tasks/install-blueprint.js
```diff
@@ -87,10 +89,16 @@ class InstallBlueprintTask extends Task {
   }
 
   _tryNpmBlueprint(name) {
+    let [, nameWithoutVersion, version] = name.match(/^(@?[^@]+)(?:@(.*))?$/);
```
rwjblue: Hmm, why do we need to re-parse this manually? I'd much prefer to avoid being in the business of parsing npm's own arguments ourselves...

Does this work?

```js
let parsed = npa(name);
let nameWithoutVersion = parsed.name;
let version = parsed.fetchSpec;
```

lib/tasks/install-blueprint.js
```diff
@@ -10,7 +10,7 @@ const path = require('path');
 const merge = require('ember-cli-lodash-subset').merge;
 const execa = require('../utilities/execa');
 const SilentError = require('silent-error');
-const validateNpmPackageName = require('validate-npm-package-name');
+const npa = require('npm-package-arg');
```
rwjblue: Did we already have this in the package.json (I see that `validate-npm-package-name` is removed, but don't see this added)?

kellyselden: confirm

lib/tasks/install-blueprint.js
kellyselden: That worked perfectly. Great catch!

PR #8578, opened 2019-04-10
By default make ember test to pick ports automatically

SparshithNR:
1) This will avoid port collission when two instances of tests running.
2) Changes made to failing tests

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -187,7 +187,7 @@ module.exports = Command.extend({
         ui: this.ui,
         outputPath,
         project: this.project,
-        port: await this._generateTestPortNumber(commandOptions),
+        port: await this._generateTestPortNumber(commandOptions, this.ui),
       },
       this._generateCustomConfigs(commandOptions)
     );
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -8,6 +8,10 @@ const path = require('path');
 const Win = require('../utilities/windows-admin');
 const fs = require('fs');
 const temp = require('temp');
+const util = require('util');
+const PortFinder = require('portfinder');
+let getPort = util.promisify(PortFinder.getPort);
+
 temp.track();
 
 let defaultPort = 7357;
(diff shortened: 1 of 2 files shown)


stefanpenner:
LGTM, although I wonder if we can do what @Turbo87 mentioned, and try `4200` first before trying other stuff, to avoid confusion.

SparshithNR:
> LGTM, although I wonder if we can do what @Turbo87 mentioned, and try `4200` first before trying other stuff, to avoid confusion.

We didn't have `4200` as default. https://github.com/ember-cli/ember-cli/pull/8578/files#diff-39de54119935553ddd431dac09b49eafL13

rwjblue:
@SparshithNR - Can you dust the cobwebs off of this one with a rebase? I want to see if CI is passing now...

rwjblue:
Yea! It passes now...

> LGTM, although I wonder if we can do what @Turbo87 mentioned, and try 4200 first before trying other stuff, to avoid confusion.

We should probably still try this. We should try to use 7357 first, then fallback to dynamic search if we can't use 7357.

@SparshithNR - Mind trying to tweak that?

Turbo87:
why `7357`?

SparshithNR:
@Turbo87 It was by default set to `7357` so we are continuing with the same. https://github.com/ember-cli/ember-cli/blob/839d207b5ed6ffb7d8c6741f5885fe0dd28c1198/lib/commands/test.js#L14-L15

rwjblue:
@turbo87 because that is the default port of `ember test —server` (has been for a very long time). Maybe you were thinking that this changes `ember serve` (which _would_ need to default to 4200)?

SparshithNR:
@rwjblue Windows test failed for an unrelated issue I think. Is there a way to trigger a rebuild? or should I push a sample commit?

Turbo87:
@rwjblue yeah, sorry, got confused. I rarely use the test server anymore 😅

ef4 (approved):

PR #8592, opened 2019-04-16
Modernize create-and-step-into-directory.js

xg-wang:
Epic: #8511


xg-wang:
This PR removes the Promise polyfill. When we have a conclusion on https://github.com/ember-cli/ember-cli/pull/8567 we may need to add the shim import with all these kinds of changed files.

Issue #7779, opened 2018-04-19
environment undefined first time through


jamesarosen:
I've added the following to my `config/environment.js`:
```js
console.log('environment', environment)
```

When I run `ember serve`, I see
```
environment undefined
Could not start watchman
Visit https://ember-cli.com/user-guide/#watchman for more info.
environment development
```

That is, `config/environment.js` gets called multiple times (at least [once to boot the server](https://github.com/ember-cli/ember-cli/blob/6816daa6f5ca48cda04c2818fc761ffa2d29eb2f/lib/commands/serve.js#L46) and [once per middleware](https://github.com/ember-cli/ember-cli/blob/19e44befff0124a13e6c36526ba017bb462f7fea/lib/tasks/server/middleware/history-support/index.js#L22)), but `environment` only gets defaulted to `'development'` after the first invocation.

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.1.0-beta.1
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
os: darwin x64
5.5.1
1.3.2
```

rwjblue:
The argument passed in to your `config/environment.js` function is ultimately supplied by the caller of `project.config(environmentStringHere)`.  If anyone is calling `project.config()` (without an arg) then you will have your function's first argument be `undefined`.

jamesarosen:
Bingo! https://github.com/ember-cli/ember-resolver/blob/d37c06bd37eb08363dcd1cc5e6ab4329190552b1/index.js#L11

So is that "supported" behavior? If so, is the meaning of that call something like "project configuration that is not environment-dependent"?

givanse:
There is a report of `environment` being `undefined` [here](https://github.com/ember-cli-deploy/ember-cli-deploy/issues/474).

And I saw the same behavior in a brand new app. (I incorrectly reported it here first [In config/environment.js the value of environment is undefined](https://github.com/emberjs/ember.js/issues/16659))

```
 ember-cli: 2.18.2
 node: 8.11.2
 os: win32 x64
```
 
logs environment as `undefined`

givanse:
[...]

PR #8595, opened 2019-04-16
[ENHANCEMENT] Project#config should use EMBER_ENV as default environment when none is passed in

nlfurniss:
When no `env` is passed to `project#config`, it uses `undefined` which some devs have [found confusing](https://github.com/ember-cli/ember-cli/issues/7779).

I'm proposing using the environment from `EMBER_ENV` as the default if no `env` is supplied.

- [x] Tested in demo app
- [x] Wrote new test
- [x] Both fast and slow ember-cli tests pass

Closes https://github.com/ember-cli/ember-cli/issues/7779

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -251,10 +251,11 @@ class Project {
     @return {Object}     Merged configuration object
    */
   config(env) {
-    let key = `${this.configPath()}|${env}`;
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you @nlfurniss!

PR #8603, opened 2019-04-19
Fix ember test --environment=production: excluding testem.js from being fingerprinted

### LINKED ISSUE #8154: testem fails to connect when running a production environment
happens in both `ember-cli` `3.4.x` and `3.5.x`

does not happen in ember-cli `~3.3.0`

```
(node:73081) ExperimentalWarning: The fs.promises API is experimental
ember-cli: 3.4.3
http_parser: 2.8.0
node: 10.1.0
v8: 6.6.346.27-node.6
uv: 1.20.2
zlib: 1.2.11
ares: 1.14.0
modules: 64
nghttp2: 1.29.0
napi: 3
openssl: 1.1.0h
icu: 61.1
unicode: 10.0
cldr: 33.0
tz: 2018c
os: darwin x64
6.4.1
1.5.1
```

step2yeung:
this helps fix #8154

TLDR: prevent testem.js from being fingerprinted so it can serve the testem.js with code that enables the connection between the browser and testem server & test reporting.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -368,9 +368,9 @@ class EmberApp {
 
     // Add testem.js to excludes for broccoli-asset-rev.
     // This will allow tests to run against the production builds.
-    if (this.options.fingerprint && this.options.fingerprint.exclude) {
-      this.options.fingerprint.exclude.push('testem');
-    }
+    this.options.fingerprint = this.options.fingerprint || {};
+    this.options.fingerprint.exclude = this.options.fingerprint.exclude || [];
+    this.options.fingerprint.exclude.push('testem');
   }
 
   _emberCLIBabelConfigKey() {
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -368,7 +368,9 @@ class EmberApp {
 
     // Add testem.js to excludes for broccoli-asset-rev.
     // This will allow tests to run against the production builds.
-    if (this.options.fingerprint && this.options.fingerprint.exclude) {
+    if (this.isProduction || this.options.fingerprint) {
```
Turbo87: I'm wondering why we only have this for `isProduction`. Might make sense to add that regardless of the build mode 🤔

rwjblue: Ya, agree. @step2yeung any objection to always excluding?

step2yeung: make sense, updated!

rwjblue:
Thank you @step2yeung!

PR #8605, opened 2019-04-23
CONTRIBUTING: Fix the test:all command lines

MonsieurDart:
Fix the test:all command lines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -227,7 +227,7 @@ And in case we didn't emphasize it enough: **we love tests!**
 
 #### Testing
 
-Testing is done with mocha. You can invoke the tests by running `npm test` or `npm test:all`. Both scripts are running `node tests/runner` may be with additional params. To contribute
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -227,8 +227,7 @@ And in case we didn't emphasize it enough: **we love tests!**
 
 #### Testing
 
-Testing is done with mocha. You can invoke the tests by running `npm test` or `npm run-script
-test-all`. Both scripts are running `node tests/runner` may be with additional params. To contribute
+Testing is done with mocha. You can invoke the tests by running `npm test` or `npm test:all`. Both scripts are running `node tests/runner` may be with additional params. To contribute
```
Turbo87: I think when using `npm` it needs to be `npm run test:all`. Skipping the `run` only works in `yarn`.

MonsieurDart: Yup, you are right, my bad! And both work with `yarn`.

PR #8608, opened 2019-04-23
[Modernize] tests/unit/models/asset-size-printer-test.js

MonsieurDart:
cc @Turbo87

Epic: #8511

### DIFF
--- a/tests/unit/models/asset-size-printer-test.js
+++ b/tests/unit/models/asset-size-printer-test.js
@@ -40,113 +40,110 @@ describe('models/asset-size-printer', function() {
     });
   }
 
-  beforeEach(function() {
-    return mkTmpDirIn(tmpRoot).then(function(tmpdir) {
-      storedTmpDir = tmpdir;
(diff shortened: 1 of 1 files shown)


tests/unit/models/asset-size-printer-test.js
```diff
@@ -40,113 +40,110 @@ describe('models/asset-size-printer', function() {
...
+    fs.mkdirsSync(assetDir);
+    fs.mkdirsSync(assetChildDir);
 
-      writeFiles();
-    });
+    writeFiles();
   });
 
   afterEach(function() {
     return Promise.all([remove(storedTmpDir)]);
```
Turbo87: the `Promise.all()` here seems unnecessary. you can replace that while you're on it:

```js
afterEach(async function() {
  await fs.remove(storedTmpDir);
});
```

locks:
```suggestion
    await fs.remove(storedTmpDir);
```

PR #8627, opened 2019-05-07
blueprints/app: Update `ember-cli-babel` to v7.7.3

Turbo87:
... so that we can use them sweet sweet decorators 💖


Turbo87:
the reason for AppVeyor failing is reported in https://github.com/ember-cli/ember-cli/pull/8631#issuecomment-491499261

I'll merge anyway, as the other builds seem to be 🍏

PR #8633, opened 2019-05-10
Refactor template build pipeline to enable co-located templates.

rwjblue:
This refactors the processing of templates so that instead of filtering out templates from JavaScript we process them in a pipeline (first the templates, and the result of that is passed in to be compiled as JavaScript).

On its own, this change does not enable co-located templates, but this work is required so that ember-cli-htmlbars can implement the feature.

### DIFF
--- a/lib/broccoli/default-packager.js
+++ b/lib/broccoli/default-packager.js
@@ -237,24 +237,33 @@ module.exports = class DefaultPackager {
     if (this._cachedProcessedAppAndDependencies === null) {
       let config = this.packageConfig();
       let internal = this.packageEmberCliInternalFiles();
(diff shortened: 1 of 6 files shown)


rwjblue:
```
not ok 235 Addon - moduleName uses the module name function
  expected {} to deeply equal { Object (totes-not-fake-addon) }
  AssertionError: expected {} to deeply equal { Object (totes-not-fake-addon) }
      at Context.<anonymous> (tests/unit/broccoli/addon/module-name-test.js:67:37)
      at Generator.next (<anonymous>)
      at onFulfilled (node_modules/co/index.js:65:19)
```

PR #8641, opened 2019-05-14
Fixed a few spelling/grammar/link issues

Alonski:
This could probably be improved a bit more. Not sure exactly what this does.

### DIFF
--- a/lib/utilities/windows-admin.js
+++ b/lib/utilities/windows-admin.js
@@ -5,15 +5,15 @@ const chalk = require('chalk');
 class WindowsSymlinkChecker {
   /**
    *
-   * On windows users will have a much better experience if symlinks are enabled
-   * an usable. This object when queried informs windows users, if they can
-   * improve there build performance, and how.
+   * On Windows, users will have a much better experience if symlinks are enabled
+   * and usable. This object, when queried, informs Windows users regarding
+   * improving their build performance, and how.
    *
-   *  > Windows vista: nothing we can really do, so we fall back to junctions for folders + copying of files
-   *  <= Windows vista: symlinks are available but using them is somewhat tricky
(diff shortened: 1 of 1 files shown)


kellyselden:
You have a legitimate failure in the tests based on the console logging.

Alonski:
Doh... It seems that the tests check the content. The tests also need to be fixed. I will try to find time to get this done in the near future

Alonski:
BTW this makes it harder to contribute docs changes. Not sure what to do about it. I wouldn't have expected the change I made to be tested :)

rwjblue:
> I wouldn't have expected the change I made to be tested :)

If it were just updating the comment, I'd agree but this also updates console output which we think is fairly important to have tested...

Alonski:
Oh double doh... Makes sense :)

lib/utilities/windows-admin.js
```diff
@@ -5,15 +5,15 @@ const chalk = require('chalk');
...
+   * On Windows, users will have a much better experience if symlinks are enabled
+   * and usable. When queried, this object informs Windows users regarding
+   * improving their build performance, and how.
    *
-   *  > Windows vista: nothing we can really do, so we fall back to junctions for folders + copying of files
-   *  <= Windows vista: symlinks are available but using them is somewhat tricky
-   *    * if the users is an admin, the process needed to have been started with elevated privs
+   *  > Windows Vista: nothing we can really do, so we fall back to junctions for folders + copying of files
+   *  <= Windows Vista: symlinks are available but using them is somewhat tricky
+   *    * if the user is an admin, the process needs to have been started with elevated priveleges
```
locks:
```suggestion
   *    * if the user is an admin, the process needs to have been started with elevated privileges
```

Alonski:
Awesome work @locks! Thanks

PR #8657, opened 2019-05-21
Add npmignore to Blueprint.renamedFiles{}

Panman82:
The addons blueprint should also included a `npmignore` file that is renamed to `.npmignore`. This would allow the octane addon blueprint to include the npmignore file. The current, 3.10 addon blueprint does this via some hooks in `index.js`, which IMO would be unneeded if npmignore is added here.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -172,8 +172,6 @@ module.exports = {
 
     '^addon-config/environment.js': 'config/environment.js',
     '^addon-config/ember-try.js': 'config/ember-try.js',
-
-    '^npmignore': '.npmignore',
   },
 
   fileMapper(path) {
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1381,7 +1381,7 @@ Blueprint._readdirSync = function(path) {
 */
 Blueprint.renamedFiles = {
   gitignore: '.gitignore',
-  npmignore: '.npmignore'
+  npmignore: '.npmignore',
 };
 
 /**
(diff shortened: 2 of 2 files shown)


rwjblue:
> The current, 3.10 addon blueprint does this via some hooks in index.js

Shouldn't that also be updated in this PR?

Panman8201:
@rwjblue good point, I'm working on the [octane addon blueprint PR](https://github.com/ember-cli/ember-octane-blueprint/pull/105) but I could adjust the blueprint here too. Just to confirm, I think all that's needed is to remove this line:

https://github.com/ember-cli/ember-cli/blob/v3.10.1/blueprints/addon/index.js#L162

kellyselden:
Putting code that applies only to addons in all blueprints doesn't seem right. Could the octane addon blueprint include this line `'^npmignore': '.npmignore',` in the fileMap and accomplish the same goal?

Panman8201:
@kellyselden I'm not sure since the octane blueprint doesn't have most of the code from the 3.10 blueprint, and I'm having a tough time trying to figure out what's going on in the 3.10 blueprint `index.js` 
 (it's mixed in with the app blueprint somehow). Additionally, I feel like (with `.gitignore`), any blueprints published will probably run into issues with a `.npmignore` too.

Panman8201:
I guess either way I'm going to have to do something else in the Octane blueprint because (if merged) this change wouldn't apply for any existing ember-cli installs. Is it possible for the Octane addon blueprint to modify the `renamedFiles` property?

https://ember-cli.com/api/classes/Blueprint.html#props_renamedFiles

rwjblue:
@Panman8201 - Mind rebasing? I think that should pick up the required CI fixes to make it pass.

locks:
Closing as stale. Thanks for the effort!

PR #8662, opened 2019-05-24
use async/await in ember-try config

kellyselden:


kellyselden:
Good idea.

kellyselden:
@rwjblue I believe you lose the concurrency of Promise.all that way. I will make the change, but just wanted to point it out.

kellyselden:
@rwjblue changes and rebased.

rwjblue:
Ya, true (RE: the concurrency), but I think the clarity is much better and _probably_ still worth it...

pgengler:
FWIW, I tried this out on a project and I get an eslint parse failure on `async function`:
```
5:24  error  Parsing error: Unexpected token function
```

where line 5 is;
```module.exports = async function() {```

(Node v10.15.3 and eslint v5.16.0)

pgengler:
Nevermind, figured it out ... I didn't update the top-level .eslintrc.js to get rid of the `ecmaVersion` override (#8659). Sorry!

kellyselden:
Thanks @rwjblue!

PR #8614, opened 2019-04-28
[WIP] Replace watcher with Broccoli watcher

oligriffiths:
This PR is an attempt to replace the Ember CLI file watcher with the Broccoli one.
At this point I'm only replacing the watcher, not the server etc to keep the changes contained, but in the future I see these also moving to Broccoli.

There are quite a few differences in the API and functionality of the Ember-CLI watcher and the Broccoli one.
The Ember version https://github.com/ember-cli/broccoli-sane-watcher/blob/ember-cli/index.js#L116 sets up a bunch of handlers on the build that do the following:

```js
  return this.builder
    .build(addWatchDir, annotation) // 1
    .then(saveNode) // 2
    .then(totalTime) // 3
    .then(appendFilePath) // 4
    .then(triggerChange, triggerError) // 5
    .then(verboseOutput.bind(this)) // 6
    .finally(cleanup); // 2
```

1. Notify the watcher of watched directories - this isn't necessary as Broccoli watcher does it a different way https://github.com/broccolijs/broccoli/blob/master/lib/watcher.js#L39
2. Some heimdall cleanup
3. Total time calculations (weirdly this total time is different from the output nodes totalTime stat...)
4. `appendFilePath` which appears to do nothing as `filePath` is null https://github.com/ember-cli/broccoli-sane-watcher/blob/ember-cli/index.js#L58
5. Triggers a change in the Ember watcher model
6. Print slow nodes

All of the above (except 4) have been ported into the Ember watcher model.
These changes are needed in broccoli https://github.com/broccolijs/broccoli/pull/398 so will land that first.

PR #8666, opened 2019-05-30
Updating #8614 to fix tests and functionality 

thoov:
cc @stefanpenner @oligriffiths

### DIFF
--- a/lib/experiments/index.js
+++ b/lib/experiments/index.js
@@ -1,6 +1,12 @@
 'use strict';
 
-const availableExperiments = Object.freeze(['PACKAGER', 'MODULE_UNIFICATION', 'DELAYED_TRANSPILATION', 'SYSTEM_TEMP']);
+const availableExperiments = Object.freeze([
+  'PACKAGER',
+  'MODULE_UNIFICATION',
+  'DELAYED_TRANSPILATION',
+  'SYSTEM_TEMP',
+  'BROCCOLI_WATCHER',
+]);
 
 const enabledExperiments = Object.freeze(['SYSTEM_TEMP']);
 
--- a/lib/models/server-watcher.js
+++ b/lib/models/server-watcher.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Watcher = require('./watcher');
+const { isExperimentEnabled } = require('../experiments');
 
 module.exports = class ServerWatcher extends Watcher {
   constructor(options) {
@@ -8,10 +9,15 @@ module.exports = class ServerWatcher extends Watcher {
 
     this.watcher.on('add', this.didAdd.bind(this));
     this.watcher.on('delete', this.didDelete.bind(this));
+
+    if (isExperimentEnabled('BROCCOLI_WATCHER')) {
+      this.watcher.on('change', this.didChange.bind(this));
+    }
   }
 
+  // This branch can be removed once BROCCOLI_WATCHER is enabled by default
   constructWatcher(options) {
-    return new (require('sane'))(this.watchedDir, options);
+    return this.constructBroccoliWatcher(options);
   }
 
   constructBroccoliWatcher(options) {
--- a/lib/models/server-watcher.js
+++ b/lib/models/server-watcher.js
@@ -14,6 +14,10 @@ module.exports = class ServerWatcher extends Watcher {
     return new (require('sane'))(this.watchedDir, options);
   }
 
+  constructBroccoliWatcher(options) {
+    return new (require('sane'))(this.watchedDir, options);
+  }
+
   didChange(relativePath) {
     this.ui.writeLine(`File changed: "${relativePath}"`);
   }
--- a/lib/models/watcher.js
+++ b/lib/models/watcher.js
@@ -7,7 +7,7 @@ const serveURL = require('../utilities/get-serve-url');
 const { isExperimentEnabled } = require('../experiments');
 const printSlowTrees = require('broccoli-slow-trees');
 
-class Watcher extends CoreObject {
+module.exports = class Watcher extends CoreObject {
   constructor(_options) {
     super(_options);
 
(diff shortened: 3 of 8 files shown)


lib/models/watcher.js
```diff
@@ -16,17 +18,66 @@ class Watcher extends CoreObject {
...
+  _totalTime(hash) {
+    const sumNodes = (node, cb) => {
+      let total = 0;
+      node.visitPreOrder(node => {
+        total += cb(node);
+      });
+      return total;
+    };
+
+    return sumNodes(hash.graph.__heimdall__, function(node) {
```
stefanpenner: `=>`

lib/models/watcher.js
```diff
@@ -35,13 +86,12 @@ class Watcher extends CoreObject {
...
   }
 
-  then() {
-    return this.watcher.then.apply(this.watcher, arguments);
-  }
-
   didChange(results) {
     logger.info('didChange %o', results);
 
+    if (isExperimentEnabled('BROCCOLI_WATCHER') && !this.foobar) {
```
stefanpenner: `this.foobar`?

stefanpenner:
LGTM – will merge once tests pass, and then we will cut a `beta` asap.

lib/models/server-watcher.js
```diff
@@ -14,6 +14,10 @@ module.exports = class ServerWatcher extends Watcher {
     return new (require('sane'))(this.watchedDir, options);
   }
 
+  constructBroccoliWatcher(options) {
+    return new (require('sane'))(this.watchedDir, options);
```
rwjblue: This is exactly the same as `constructWatcher`, what is the point of having a second identical method?

lib/models/watcher.js
```diff
@@ -16,17 +18,65 @@ class Watcher extends CoreObject {
...
-    this.watcher.on('change', this.didChange.bind(this));
     this.serveURL = serveURL;
   }
 
   constructWatcher(options) {
     return new (require('ember-cli-broccoli-sane-watcher'))(this.builder, options);
   }
 
+  constructBroccoliWatcher(options) {
+    let watcher = new (require('broccoli')).Watcher(this.builder, this.builder.builder.watchedSourceNodeWrappers, {
```
rwjblue: this would be much easier to read if we separate the require from the instantiation

```suggestion
const { Watcher } = require('broccoli');
let watchedNodes = this.builder.builder.watchedSourceNodeWrappers
let watcher = new Watcher(this.builder, watchedNodes, {
```

lib/models/server-watcher.js
stefanpenner: This has to do with some complexity with how the inheritance + feature flagging. 
It likely could be refactored + given a comment. 

Once the flag is gone, this wont be as silly.

package.json
```diff
@@ -56,6 +56,7 @@
     "broccoli-middleware": "^2.0.1",
     "broccoli-module-normalizer": "^1.3.0",
     "broccoli-module-unification-reexporter": "^1.0.0",
+    "broccoli-slow-trees": "^3.0.1",
```
oligriffiths: Shall we expose lib/messages from broccoli and include the slowTrees printing from there? I think that’s better than the dependency here

lib/models/watcher.js
thoov: I agree and updated

package.json
thoov: I 100% agree that we should use broccoli's message util. However, we should explore that as a quick follow up so that we can start validating the watcher / broccoli 3 stuff now.

lib/models/watcher.js
```diff
@@ -16,17 +18,65 @@ class Watcher extends CoreObject {
...
 
     this.serving = _options.serving;
-    this.watcher = this.watcher || this.constructWatcher(options);
+
+    if (isExperimentEnabled('BROCCOLI_WATCHER')) {
+      this.watcher = this.watcher || this.constructBroccoliWatcher(options);
+    } else {
+      // This branch can be removed once BROCCOLI_WATCHER is enabled by default
+      this.watcher = this.watcher || this.constructWatcher(options);
+      this.watcher.on('change', this.didChange.bind(this));
```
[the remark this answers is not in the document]
thoov: I now moved both sets of event listeners out of the construct methods. This is left outside of that method intentionally because of how the inheritance model with the ServerWatcher class works. This is something that we should address once we turn this feature flag on fully.

lib/models/server-watcher.js
thoov: I cleaned this up a bit as well but stef's point above is the underlining cause for both of these nearly identical methods.

lib/experiments/index.js
```diff
@@ -1,6 +1,12 @@
 'use strict';
 
-const availableExperiments = Object.freeze(['PACKAGER', 'MODULE_UNIFICATION', 'DELAYED_TRANSPILATION', 'SYSTEM_TEMP']);
+const availableExperiments = Object.freeze([
+  'PACKAGER',
+  'MODULE_UNIFICATION',
+  'DELAYED_TRANSPILATION',
+  'SYSTEM_TEMP',
+  'BROCCOLI_WATCHER',
```
rwjblue: Each experiment should be tested independently, adding this here also requires adding a corresponding `.travis.yml` entry.

rwjblue: I pushed 1e11fb6 to address this.

oligriffiths:
@rwjblue can we move https://github.com/rwjblue/broccoli-slow-trees to the broccoli project?

package.json
oligriffiths: Cool, first step done here https://github.com/oligriffiths/broccoli-slow-trees/pull/1/files?w=1

stefanpenner:
🕺

stefanpenner:
@oligriffiths I believe @thoov was going to land that in a follow up cleanup PR next week.

oligriffiths:
🎉 Thanks @thoov

PR #8678, opened 2019-06-04
give ember-cli a progress indicator

stefanpenner:
learnings from: https://github.com/stefanpenner/ember-cli-progress
asciinema: https://asciinema.org/a/jYyqWp3096aWqUgNNSnBo9hb2

I have been running this in several non-trivial apps for the last few months, so far the feedback has been positive and we have not been running into any issues. So I would like to propose adding the functionality directly to ember-cli. 

- [x] feedback

Plan (assuming accepted):

- [x] deprecate https://github.com/stefanpenner/ember-cli-progress
- [x] change ember-cli-progress to warn + disable itself if a new version of cli with built-in progress is used.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -12,6 +12,7 @@ const _resetTreeCache = require('./addon')._resetTreeCache;
 const Sync = require('tree-sync');
 const heimdall = require('heimdalljs');
 const { isExperimentEnabled } = require('../experiments');
+const progress = require('../utilities/heimdall-progress');
 
 /**
  * Wrapper for the Broccoli [Builder](https://github.com/broccolijs/broccoli/blob/master/lib/builder.js) class.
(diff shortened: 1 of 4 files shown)


lib/models/builder.js
```diff
@@ -209,6 +210,12 @@ class Builder extends CoreObject {
       }
     }
 
+    this.ui.startProgress(progress.format(progress()));
+
+    const interval = setInterval(() => {
```
sarbbottam: mind renaming it to `intervalId` to be more explicit?

kellyselden: Can you continue the convention of using `let` please?

lib/models/builder.js
```diff
@@ -226,6 +233,10 @@ class Builder extends CoreObject {
         }
       )
       .then(this.checkForPostBuildEnvironmentIssues.bind(this))
+      .finally(() => {
+        this.ui.stopProgress();
+        clearInterval(interval);
```
kellyselden: Should this go above `stopProgress` in case it throws?

lib/models/builder.js
oligriffiths: For my own knowledge, why do we prefer let over const?

stefanpenner: I believe `const` is appropriate here.

stefanpenner: Good call, making it clear that we are differentiating between the actual interval, and the "interval cookie" provided by `setInterval` makes sense.

stefanpenner: Specifically, `interval` being changed later on, would in-fact break this algorithm.

stefanpenner:
Any concerns with merging this?

rwjblue:
> Any concerns with merging this?

Not sure what more than approval I can do here, but: no, I have no concerns 😝

stefanpenner:
The Travis build which is currently "not completed" has actually completed and is green. So I am just going to merge, and let this run the canary release train.

PR #8679, opened 2019-06-05
Adding change logging for backwards compatibility

thoov:
Adding logging to support broccoli-sane-watchers' logic

### DIFF
--- a/lib/models/server-watcher.js
+++ b/lib/models/server-watcher.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const Watcher = require('./watcher');
-const { isExperimentEnabled } = require('../experiments');
 
 module.exports = class ServerWatcher extends Watcher {
   constructor(options) {
--- a/lib/models/server-watcher.js
+++ b/lib/models/server-watcher.js
@@ -9,10 +9,6 @@ module.exports = class ServerWatcher extends Watcher {
 
     this.watcher.on('add', this.didAdd.bind(this));
     this.watcher.on('delete', this.didDelete.bind(this));
-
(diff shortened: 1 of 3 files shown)


thoov:
cc: @stefanpenner @oligriffiths

stefanpenner:
Looks like a relevant failure:

```
not ok 968 Server Watcher watcher:change logs that the file was changed
  expected 'file foo.txt undefined\nFile changed: "foo.txt"\n' to equal 'File changed: "foo.txt"\n'
  AssertionError: expected 'file foo.txt undefined\nFile changed: "foo.txt"\n' to equal 'File changed: "foo.txt"\n'
      at Context.<anonymous> (tests/unit/models/server-watcher-test.js:33:28)
```

thoov:
FYI: this logging isn't exactly one-to-one with brococli-sane-watcher. This PR will only log once per build (even if multiple things change as the change event is not fired if a rebuild is already happening: https://github.com/broccolijs/broccoli/blob/master/lib/watcher.js#L69)

Example if you delete 2 files at the same time the current watcher would log:

```
file deleted controllers/bar.js
file added controllers/foo.js
```

And the new watcher would just log:

```
file deleted controllers/bar.js
```

If we want to keep exact pairity then we need to move this logic to here: https://github.com/broccolijs/broccoli/blob/master/lib/watcher_adapter.js#L17

cc: @stefanpenner @oligriffiths

oligriffiths:
Once we do the console UI work we can revisit this in Broccoli

rwjblue:
FWIW, I do think we want to do this in Broccoli. Knowing only one of the changed files would essentially mean that things like the hot reloader are only able to reload the first changed file (not all).

PR #8680, opened 2019-06-05
[Fixes #8677] ensure watcher parity

### LINKED ISSUE #8677: beta broccoli watcher issue
(paired with @thoov)

On beta, introduced by: https://github.com/ember-cli/ember-cli/pull/8666 we have  introduced a breaking change. Specifically the meaning of `watcher.on('change'` has changed breaking https://github.com/lifeart/ember-ast-hot-load.

To reproduce:

1. first checkout `ember-ast-hot-load`
2. then link the beta branch of ember-cli
3. then run:

```
EMBER_CLI_BROCCOLI_WATCHER=true BROCCOLI_ENABLED_MEMOIZE=true ember s
```

4.  With the network tab open:
mutate `tests/dummy/app/templates/components/template-only.hbs`

You should only see `dummy.js` being loaded, and the result column of `template-only` being updated, but currently you see a full page reload.

Additionally when switching from `on('changes'` to `on('buildSuccess` we see only the `dummy.js` being refetched, but the UI isn't updating correctly. This leaves me to believe something else might be off.

Finally, when running CLI without `EMBER_CLI_BROCCOLI_WATCHER=true BROCCOLI_ENABLED_MEMOIZE=true` set, all appears to function correctly.

---

Thanks @lifeart for reporting this issue so quickly!

cc @oligriffiths

stefanpenner:
this fixes #8677 and should unblock beta.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "babel-plugin-module-resolver": "^3.2.0",
     "bower-config": "^1.4.1",
     "bower-endpoint-parser": "0.2.2",
-    "broccoli": "^3.0.0",
+    "broccoli": "^3.0.1",
     "broccoli-amd-funnel": "^2.0.1",
     "broccoli-babel-transpiler": "^7.2.0",
     "broccoli-builder": "^0.18.14",


stefanpenner:
cc @thoov

stefanpenner:
cc @thoov / @lifeart

PR #8687, opened 2019-06-09
Add .git directory to npmignore

rwwagner90:
I am not sure why this suddenly became an issue, but it appears that `.git` is being published to npm now, at least for ember-shepherd. It might be related to me recently updating node/npm and perhaps it used to be ignored by default, but we probably do not want to publish the `.git` directory, so this should ignore it. Related issue is here https://github.com/shipshapecode/ember-shepherd/issues/319

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -12,6 +12,7 @@
 /.env*
 /.eslintignore
 /.eslintrc.js
+/.git/
 /.gitignore
 /.template-lintrc.js
 /.travis.yml


rwwagner90:
I did some more research, and apparently `.git` is supposed to be ignored by default, according to https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

However, there seem to be some bugs in some of the newer npm versions that cause it to be published.

rwwagner90:
Also, random thought. Would it be easier to pivot to a `files` array and have the default for addons only ship the `addon` and `app` directories?

stefanpenner:
Is there an open issue on npm?

kellyselden:
+1 on files

rwwagner90:
@stefanpenner yes, there is https://npm.community/t/npm-pack-includes-items-from-git-folder-if-there-is-a-branch-called-readme-regression-in-6-9-0/6044/5

@kellyselden should I change this PR to adding a `files` array instead and remove the `npmignore` from the default blueprints? I'm happy to do whatever the best solution is, I just want to get something in 😃

stefanpenner:
@rwwagner90 cool thanks. In general I would prefer to not  work around an NPM bug here, but if it doesn't look like they would hot fix that, then maybe we need to.

> @kellyselden should I change this PR to adding a files array instead and remove the npmignore from the default blueprints? I'm happy to do whatever the best solution is, I just want to get something in 😃

If you are interested in exploring this, please do so via RFC.

rwwagner90:
@stefanpenner sure thing. I will write up an RFC for switching to `files`.

In the meantime, it might not hurt to have this npmignore change, as it should hopefully ensure no one accidentally publishes their `.git` directory and makes their addon unable to be installed.

rwjblue:
:sob: I do not think files is a good idea, unless we also move our main entry point into a subfolder (and have that folder be in the files list).

stefanpenner:
@rwjblue yup, I share similar concerns.

rwwagner90:
@rwjblue wouldn't the entry point be `index.js`? If so, I would envision adding `app`, `addon`, `index.js` and maybe `lib` and some other things to files. Are there some files for which this would be problematic? If so, probably no point in me writing an RFC 😛

PR #8682, opened 2019-06-06
fix: allow use of prettier on index.html

lougreenwood:
### Problem

When using Prettier to auto-format HTML files on save, if the `index.html` file is saved, Prettier will perform the following conversion where space padding is applied to `{{rootURL}}`:

from:
`<script src="{{rootURL}}assets/vendor.js"></script>`

to:
`<script src="{{ rootURL }}assets/vendor.js"></script>`

Because `ember-cli` expects exactly `{{rootURL}}` - the output of Prettier causes the app to fail to boot because `ember-cli` has not performed the usual replacements on `{{ rootURL }}`.

### Solution

Add optional spaces to the regex used to identify these `{{rootURL}}` tags in index.html.

This PR also allows space padding for the following tags: `{{EMBER_ENV}}` `{{MODULE_PREFIX}}`.

**Note:**
The fix was not added for `{{content-for ...}}` because Prettier only applies this padding when a tag is part of a string literal (`<script src="{{ rootURL }}...">`) and `{{content-for ...}}` seems to not be used in string literals.

This can easily be added for safety if required.

PR #8692, opened 2019-06-10
fix: allow use of prettier on index.html

lougreenwood:
Replaces #8682 

### Problem

When using Prettier to auto-format HTML files on save, if the `index.html` file is saved, Prettier will perform the following conversion where space padding is applied to `{{rootURL}}`:

from:
`<script src="{{rootURL}}assets/vendor.js"></script>`

to:
`<script src="{{ rootURL }}assets/vendor.js"></script>`

Because `ember-cli` expects exactly `{{rootURL}}` - the output of Prettier causes the app to fail to boot because `ember-cli` has not performed the usual replacements on `{{ rootURL }}`.

### Solution

Add optional spaces to the regex used to identify these `{{rootURL}}` tags in index.html.

This PR also allows space padding for the following tags: `{{EMBER_ENV}}` `{{MODULE_PREFIX}}`.

**Note:**
The fix was not added for `{{content-for ...}}` because Prettier only applies this padding when a tag is part of a string literal (`<script src="{{ rootURL }}...">`) and `{{content-for ...}}` seems to not be used in string literals.

This can easily be added for safety if required.

### DIFF
--- a/lib/utilities/ember-app-utils.js
+++ b/lib/utilities/ember-app-utils.js
@@ -167,13 +167,13 @@ function contentFor(config, match, type, options) {
 function configReplacePatterns(options) {
   return [
     {
-      match: /{{rootURL}}/g,
+      match: /{{\s?rootURL\s?}}/g,
       replacement(config) {
         return normalizeUrl(config.rootURL);
       },
     },
     {
-      match: /{{EMBER_ENV}}/g,
+      match: /{{\s?EMBER_ENV\s?}}/g,
       replacement(config) {
         return convertObjectToString(config.EmberENV);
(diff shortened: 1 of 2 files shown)


lougreenwood:
@rwjblue - this branch addresses your feedback in https://github.com/ember-cli/ember-cli/pull/8682#issuecomment-499494845 with none of my Git Rebase mess from the previous branch ;)

lougreenwood:
@rwjblue you're welcome and no problem - I'm just not used to rebasing, and that's my own fault :)

stefanpenner:
fancy!

PR #8698, opened 2019-06-12
Changed uninstalling and remove wording for delete

RecoX:
Changed uninstalling and remove wording for delete to maintain coherence with the command name

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -327,7 +327,7 @@ let Blueprint = CoreObject.extend({
     },
 
     remove(info) {
-      this._writeStatusToUI(chalk.red, 'delete', info.displayPath);
+      this._writeStatusToUI(chalk.red, 'destroy', info.displayPath);
       if (!this.dryRun) {
         return cleanRemove(info);
       }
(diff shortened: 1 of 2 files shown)


rwjblue:
Hmm, I don't have a _particular_ preference here but I don't exactly see `remove` as "incorrect".

@RecoX - Mind helping me understand the rationale of why `delete` is better than `remove`?

stefanpenner:
Deleting does sound/read better in this context

RecoX:
> Hmm, I don't have a _particular_ preference here but I don't exactly see `remove` as "incorrect".
> 
> @RecoX - Mind helping me understand the rationale of why `delete` is better than `remove`?

Is to maintain the word along the whole process, the command to delete something in ember is

`ember d component name` or `ember delete route name`

so that's why i change it, to easily remember the commands

a lot of times i tried with `ember u` or `ember r` because of this

kellyselden:
It might be useful to compare to how git does it, which also uses "remove":

```
C:\code\ember-awesome-macros>git clean -fdx
Removing .vscode/
Removing dist/
Removing node_modules/
Removing tmp/
```

rwjblue:
> ember d component name or ember delete route name

FWIW, the `d` in `ember d` doesn't stand for "delete" it is "destroy".

locks:
I renamed to destroy to match the command as @rwjblue mentioned. Let me know what is necessary to move this forward, whether it's reverting to delete, or even keeping the current status quo!

Turbo87:
IMHO "destroy" sounds quite aggressive... 🤔

locks:
Closing due to inactivity.

RecoX:
@locks  Just merge it 🥇

locks:
@RecoX I'm not on the CLI team so I can't make the call 😅 There wasn't concensus on the change.

PR #8701, opened 2019-06-13
Close #8589 addon templates were detected, but no template compilers …

### LINKED ISSUE #8589: addon templates were detected, but no template compilers were registered.
As an addon author, I would like a template compiler to be automatically registered for me when I need it, so I avoid the error message below. Is there a reasonable way to for ember-cli to do this?

Error message: 
"Addon templates were detected but there are no template compilers registered for (addon-name). Please make sure your template precompiler (commonly ember-cli-htmlbars) is listed in dependences (NOT devDependencies) in (addon-name)'s package.json."

haochuan:
For #8589. 

As mentioned by @rwjblue, moved `ember-cli-htmlbars` into the dependencies of the addon blueprint by default, since we will only see this error in ember addons.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -51,6 +51,12 @@ module.exports = {
     contents.dependencies['ember-cli-babel'] = contents.devDependencies['ember-cli-babel'];
     delete contents.devDependencies['ember-cli-babel'];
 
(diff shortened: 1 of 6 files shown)


rwjblue:
Thank you!!

PR #8656, opened 2019-05-20
fix: reset resolve-package-path caches in PackageInfoCache._clear()

rwjblue:
Backport changes from https://github.com/ember-cli/ember-cli/pull/8650 to `release` branch (for 3.10.x release).

PR #8752, opened 2019-07-15
fix: reset resolve-package-path caches in PackageInfoCache._clear()

chriskrycho:
- Backport changes from #8650 to `v3.10.x` branch.
- Replicates #8656, which landed in `release` and was intended as a backport for `v3.10.2`, but was never shipped as such.
- Includes the other changes that landed on v3.10.x since that branch tip was last updated. (I'm happy to change this I just don't know what the *right* procedure here is).

**Note:** I'm currently testing this change against ember-cli-typescript; will mark as *not*-WIP when finished.


chriskrycho:
Since 3.11.0 is now out, I think this is irrelevant. Thanks, folks!

PR #8761, opened 2019-07-19
lint vendor files for bare-bones IE11

kellyselden:
This come up before, but it has recently bit me again. I accidentally used newer features in a vendor file and regressed IE11. https://github.com/kellyselden/ember-macro-helpers/pull/273/files. This could have been easily been prevented with an older eslint parser, which I added with success. I'm offerring it here again for more debate, because I think it is a good idea and will help prevent errors.

I'm waiting for an answer [here](https://stackoverflow.com/questions/57107800/eslint-disable-extends-in-override) to see if I can clean up that override any.

### DIFF
--- a/blueprints/app/files/.eslintignore
+++ b/blueprints/app/files/.eslintignore
@@ -1,6 +1,5 @@
 # unconventional js
 /blueprints/*/files/
-/vendor/
 
 # compiled output
 /dist/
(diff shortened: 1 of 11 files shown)


rwjblue:
I'd rather not make this change. Ultimately, its a lot of extra boilerplate for a pretty uncommon scenario. In addition, I think its probably better to suggest folks transpile their "self authored" vendor (as opposed to literally copying a file from some other source) files so you'd just be following the same rules RE: IE11 support that you would everywhere else in your addon.

sandstrom:
Since IE11 is on it's way out, maybe we could close this PR?

https://github.com/emberjs/rfcs/pull/685

PR #8774, opened 2019-08-01
Remove --disable-gpu flag when starting headless chrome

stefanpenner:
* https://bugs.chromium.org/p/chromium/issues/detail?id=982977 is is causing issues
* https://bugs.chromium.org/p/chromium/issues/detail?id=737678 it appears to now no longer be needed

i think this should fix:

```
 message: >
                Error: Browser failed to connect within 30s. testem.js not loaded?
                Stderr:

                DevTools listening on ws://127.0.0.1:59131/devtools/browser/ba633bc8-25cf-49a3-8f32-d14cf24fb5d8
                [0801/105531.351003:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
                [0801/105531.498065:WARNING:gpu_process_host.cc(1188)] The GPU process has crashed 1 time(s)
                [0801/105532.794732:WARNING:gpu_process_host.cc(962)] Reinitialized the GPU process after a crash. The reported initialization time was 0 ms
                [0801/105533.131335:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
                [0801/105533.155083:WARNING:gpu_process_host.cc(1188)] The GPU process has crashed 2 time(s)
                [0801/105534.318595:WARNING:gpu_process_host.cc(962)] Reinitialized the GPU process after a crash. The reported initialization time was 0 ms
                [0801/105534.559282:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
                [0801/105534.656891:WARNING:gpu_process_host.cc(1188)] The GPU process has crashed 3 time(s)
                [0801/105534.656929:FATAL:gpu_data_manager_impl_private.cc(894)] The display compositor is frequently crashing. Goodbye.
                [0801/105535.025242:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)

```

PR #8775, opened 2019-08-01
Remove --disable-gpu flag from testem config blueprints

tomdale:
Unfortunately, Chrome v76 contains a regression [where passing the `--disable-gpu` flag on the Mac cause the process to crash](https://bugs.chromium.org/p/chromium/issues/detail?id=982977). This PR removes this flag from the testem config in several different blueprints, because it is no longer necessary.

This flag was originally required by headless Chrome when running on Windows. However, now that https://bugs.chromium.org/p/chromium/issues/detail?id=729961 has landed, this flag is no longer needed.


kellyselden:
Already covered #8774.

Issue #8790, opened 2019-08-16
What should we do with the module unification blueprint?


kellyselden:
It is currently a maintenence burden, and correct me if I'm wrong, it is not something that will be released?

rwjblue:
IMHO, we should remove module unification support code. Ember has already disabled its support (in https://github.com/emberjs/ember.js/pull/18066) in other words it's already not possible to actually use.

locks:
Does this work necessitate a major version?

dcyriller:
It was an experiment, behind a feature flag...

rwjblue:
> Does this work necessitate a major version?

No.

locks:
I believe most if not all of the work as been done, so I'll close this issue. Thanks everyone x)

PR #8791, opened 2019-08-16
WIP: Use Octane for default app and addon blueprints

NullVoxPopuli:
goals:
 - the existing commands should work with the new versions of both blueprints
   - app -> `@ember/octane-app-blueprint`
   - addon -> `@ember/octane-addon-blueprint`
 - the old app/addon blueprints get the `classic` prefix

questions:
 - do we want to straight up remove the classic blueprints?
 - should we also remove all the module unification code? (separate PR)


kellyselden:
My thoughts are:

> do we want to straight up remove the classic blueprints?

Yes. No point in maintaining two sets. If someone wants classic, they can `npx ember-cli@3.13 new my-app`.

> should we also remove all the module unification code? (separate PR)

#8790 Looks like the concensus is beginning to look like yes.

Panman8201:
FWIW, I think this can be closed. @rwjblue made changes to the "classic" blueprints in ember-cli (here) to now be "octane". I've also followed up with PR #8873 to make sure things lined up. Additionally, it looks like [the MU stuff was also removed in recent commits](https://github.com/ember-cli/ember-cli/compare/v3.14.0-beta.2...master).

NullVoxPopuli:
idk -- it should use the blueprint repo/package, right?

Panman8201:
> idk -- it should use the blueprint repo/package, right?

IMHO, no. It's my understanding `ember-octane-blueprints` was started as a temporary thing so that users had a way to try Octane until it was fully released, at which point using Octane is just using ember.

What you're proposing is to change how blueprints are managed and is something entirely different from how it has worked up to this point. And I'm not sure continuing with another release point is the best idea (although I'm not the one doing releases 🤷‍♂).

NullVoxPopuli:
>  is to change how blueprints are managed

it's already supported though. Just need to change the refs of app/addon to the octane app/addon packages. 

The main thing with this PR is that the tests aren't updated. Everything else in here works iirc

rwjblue:
Sorry for the run around here @NullVoxPopuli, I'm going to close this one...

PR #8798, opened 2019-08-19
Update blueprint reference for ember-source to 3.13.0-beta.2.

rwjblue:
This was changed when we merged release up into the beta branch, but the tests in beta branch were still expecting the 3.13 beta series.

Should fix beta and master (once merged upwards) CI once #8797 is also landed.


rwjblue:
Rebased on top of #8797

PR #8812, opened 2019-08-28
Clarify installation error message

jrjohnson:
I noticed this message from a question at https://stackoverflow.com/questions/57668319/ember-serve-node-modules-appears-empty-you-may-need-to-run-npm-install

I'm not sure it would have helped this user, but it seemed a touchup was in order anyway and I don't think this would add any friction for experienced developers.


kellyselden:
I guess this new message is also a bit future-proof for yarn pnp style setups?

PR #8813, opened 2019-08-28
Update NPM version check to avoid double `npm install` when using `npm@5.7.1` or higher.

deepan83:
- The issue with deletion of files/folders was resolved in npm 5.7.1. https://github.com/npm/npm/issues/17379#issuecomment-367924115
https://blog.npmjs.org/post/171139955345/v570 

This PR updates the npm version check to reflect this and only runs an additional npm install when below the version with the fix.

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -175,7 +175,7 @@ class NpmTask extends Task {
         // this ensures that we run a full `npm install` **after** any `npm
         // install foo` runs to ensure that we have a fully functional
         // node_modules hierarchy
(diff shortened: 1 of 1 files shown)


kellyselden:
I'm curious does this affect the 6.x line? Since npm maintained both majors at the same time for a while I think.

rwjblue:
@deepan83 - Mind rebasing against master to confirm CI is happy? I'm inclined to land this...

deepan83:
@rwjblue done

rwjblue:
Thank you!

Issue #8478, opened 2019-03-05
Changes in environment.js applied only on second save


nfc036:
I have properties defined in ```config/environment.js```, i.e. ```testProperty``` with value 1:
```
'use strict';
module.exports = function(environment) {
  let ENV = {
    testProperty: 1,  
    modulePrefix: 'testapp2',
    environment,
    rootURL: '/',
    locationType: 'auto',
...
```
To illustrate the problem I import ```ENV``` from config in application controller and show the current value of ```ENV.testProperty``` in the application template. 
Changing the value of ```testProperty``` from 1=>2 triggers a live reload, but the value shown in the browser is still "1" instead of "2" as expected. When changing the value from 2=>3, live reload is triggered again but not "3" as expected but "2" is shown (the value from the update before the last change).
Additional information: 
- when I add a ```console.log(ENV.testProperty')``` at the end of ```config/environment.js``` the console shows the (correct) current number twice before "Build successful".
- when I change any other file leading to a live reload, the current state of ```environment.js``` is still not reflected in the application. I have to trigger a second save on ```environment.js``` to get the correct content.
- restarting ```ember s``` always shows the current value correctly.

I noticed this with ember-cli@3.8.1 but checking with ember-cli@3.7.1 shows the same behaviour. It might be relevant that I'm working on **windows**.

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.8.1
http_parser: 2.7.0
node: 8.9.3
v8: 6.1.534.48
uv: 1.15.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 57
nghttp2: 1.25.0
openssl: 1.0.2n
icu: 59.1
unicode: 9.0
cldr: 31.0.1
tz: 2017b
os: win32 x64
6.8.0
1.3.2
```

rwjblue:
Hmm, I haven't seen/heard of this issue before now. I'll try to test tomorrow on OSX to see if the same thing is going on there and I just didn't notice...

nfc036:
Thanks for looking into this, Robert. I didn't notice it before either. I checked one of my projects. When going back in time to Juli 2018 with ember-cli@3.3.0 everything worked as expected. In September 2018 I ugraded my app from ember-cli@3.3.0 to 3.4.2. The issue described above seems to start with this update.

nfc036:
[...]

PR #8819, opened 2019-09-03
BUGFIX: Clear project.configCache before starting the build

colenso:
Fixes https://github.com/ember-cli/ember-cli/issues/8478
The problem I noticed was that `get-serve-url` would access `project.config` after `finalizeBuild` and hence the config would get cached and would be used in the next build. Hence each build was using the previous build's config.
This PR simply clears the `project.configCache` before the build is started.

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -216,6 +216,8 @@ class Builder extends CoreObject {
       this.ui.spinner.text = progress.format(progress());
     }, this.ui.spinner.interval);
 
+    this.project.configCache.clear();
+
     return this.processAddonBuildSteps('preBuild')
       .then(() => this.builder.build(this.broccoliBuilderFallback ? addWatchDirCallback : null))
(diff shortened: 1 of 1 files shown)


milindalvares:
Yes please!

rwjblue:
@colenso - Would you mind adding some tests here so that we don't regress again?
Also, I've updated the target branch of this PR to `beta` so that we can get the fix into a released version a bit faster.

rwjblue:
Still needs some tests...

locks:
@colenso I tried rebasing but I don't think I have access! Do you have some cycles to rebase and submit the test?

colenso:
@locks Sorry. I haven't seen this in a while, but I want to get this through so I rebased and made another PR https://github.com/ember-cli/ember-cli/pull/9489.
Closing this in favor of https://github.com/ember-cli/ember-cli/pull/9489

PR #8820, opened 2019-09-03
Issue deprecation when enabling MODULE_UNIFICATION flag.

rwjblue:
MODULE_UNIFICATION has always required using canary versions of Ember and ember-cli. `ember-source` has disabled the feature completely so that apps can no longer enable the feature at all.

This implements a basic system for ember-cli to tell folks that were leveraging that experiment that we will be removing support from ember-cli. This will _likely_ be removed in ember-cli@3.17 (after the next LTS release).

Begins addressing https://github.com/ember-cli/ember-cli/issues/8790.

### DIFF
--- a/lib/experiments/index.js
+++ b/lib/experiments/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const chalk = require('chalk');
 const availableExperiments = Object.freeze([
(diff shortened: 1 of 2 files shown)


stefanpenner:
🔥

PR #8834, opened 2019-09-09
Ensure addon tree is scoped to addon name before compiling templates.

rwjblue:
Without this, addon templates are compiled with an incorrect module name (they would be seen as `templates/components/foo-bar` instead of `addon-name/templates/components/foo-bar`).

The scoping was happening just before JS compilation (which is why that wasn't also incorrect), but not before template compilation. This fixes issues by doing the scoping first, and passing the scoped tree through the pipeline.

### DIFF
--- a/tests/unit/broccoli/template-precompilation-test.js
+++ b/tests/unit/broccoli/template-precompilation-test.js
@@ -19,15 +19,18 @@ describe('template preprocessors', function() {
   let input, output, addon;
 
   class FakeTemplateColocator extends BroccoliPlugin {
-    constructor(trees, options = { root: '' }) {
-      super(...arguments);
-      this.options = options;
-    }
-
     build() {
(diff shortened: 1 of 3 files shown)


tests/unit/broccoli/template-precompilation-test.js
```diff
@@ -26,8 +26,11 @@ describe('template preprocessors', function() {
 
     build() {
       let [inputPath] = this.inputPaths;
-      let root = fs.existsSync(path.join(inputPath, this.options.root)) ? this.options.root : '';
-
+      let entries = fs.readdirSync(inputPath);
+      if (entries.length > 1) {
+        throw new Error('all input files should be scoped to the addon or project name');
```
rwjblue: This would have failed prior to the changes here.

PR #8845, opened 2019-09-18
Fix: generated assets get out of sync with index.html when app name (package.json#name) changes

NullVoxPopuli:
Related: https://github.com/ember-cli/ember-octane-blueprint/pull/183

The workflow I'm trying to resolve:

```
ember new my-app
# build project, notice dist/assets has:
# - my-app.css, my-app.js

change package.json#name to "my-app-updated"
# build project, notice dist/assets has:
# - my-app-updated.css, my-app-updated.js
```
but the output index.html still expects my-app instead of my-app-updated

--------------------------------

I would have expected that the index.html still refers to the correct asset paths.

I don't know what all package.json#name is tied to -- is it tied to other things?
Is there a reason the index.html css/js names were static after app generation, even though the generated asset names are dynamic?

I guess there could maybe be somewhat undefined behavior here?

This PR locks the generated assets into the value of package.json which happens to be APP.name.
Another approach may be to generate assets based of MODULE_PREFIX instead of APP.name

### DIFF
--- a/lib/utilities/ember-app-utils.js
+++ b/lib/utilities/ember-app-utils.js
@@ -186,6 +186,13 @@ function configReplacePatterns(options) {
         return config.modulePrefix;
       },
     },
+    {
+      match: /{{\s?appName\s?}}/g,
+      replacement(config) {
+        // addons don't have an APP object, so default to dummy
+        return config.APP.name || 'dummy';
+      },
+    },
   ];
 }
 
--- a/lib/utilities/ember-app-utils.js
+++ b/lib/utilities/ember-app-utils.js
@@ -158,6 +158,7 @@ function contentFor(config, match, type, options) {
                     application or not
  * @param {Boolean} options.storeConfigInMeta Controls whether to include the
                     contents of config
+ * @param {Object} options.outputPaths The filenames referenced in app/index.html + tests/index.html
    @return {Array} An array of patterns to match against and replace
 */
 function configReplacePatterns(options) {
@@ -187,10 +188,15 @@ function configReplacePatterns(options) {
       },
     },
     {
-      match: /{{\s?appName\s?}}/g,
-      replacement(config) {
-        // addons don't have an APP object, so default to dummy
-        return config.APP.name || 'dummy';
+      match: /{{\s?outputPaths\.([^.]+)\.([^.]+)\s?}}/g,
+      replacement(_config, _match, category, fileType) {
+        return options.outputPaths[category][fileType];
+      },
+    },
+    {
+      match: /{{\s?outputPaths\.([^.]+)\.([^.]+)\.([^.]+)\s?}}/g,
+      replacement(_config, _match, category, fileType, subType) {
+        return options.outputPaths[category][fileType][subType];
       },
     },
   ];
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -10,7 +10,7 @@
     {{content-for "head"}}
 
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
-    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
+    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/{{appName}}.css">
 
     {{content-for "head-footer"}}
   </head>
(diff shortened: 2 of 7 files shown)


lib/utilities/ember-app-utils.js
```diff
@@ -190,6 +190,12 @@ function configReplacePatterns(options) {
         return config.modulePrefix;
       },
     },
+    {
+      match: /{{\s?asset-name\s?}}/g,
```
Turbo87: I think I would prefer `appName` 🤔

NullVoxPopuli: updated!

NullVoxPopuli:
> test where we change the package.json/name and check assets?

This is def needed. Just wanted an ok an implementation before I go learn how the blueprint tests work ;)

Thanks!

NullVoxPopuli:
- tests/index.html should be changed
 - should probably add additional tests? (CI passes with no test changes)
    - but I also don't know how to test the _renaming_ of an app, as I've lost a lot of context since November. lol :cry:

NullVoxPopuli:
I've added `{{appName}}` to the tests/index.html.

There is no acceptance test (that I found) for `ember build` -- should one be added in a separate PR?

NullVoxPopuli:
Just tried codesandbox's vercel integration, and it looks like this issue came up again, and codesandbox (or vercel) changes the name entry in package.json:

image

NullVoxPopuli:
anyone want to take a look at this? I'm lost / don't know what to do next and this is blocking codesandbox's integrations / forking behaviors :(

rwjblue (commented):
I've somewhat completely revamped how this works. It now uses the exact same structure to determine the actual path on disk that the files will be places as it does to determine what the scripts/links will use in the `index.html` files.

chancancode:
Hmm, we should check with @ef4 about this...

rwjblue:
> Hmm, we should check with @ef4 about this...

We have to confirm Embroider support for _any_ change in this system...

ef4:
At a glance, I suspect embroider's compatibility layer will be able to handle this. But yeah, I am 👎 on this entire API, both before and after this change. This PR is patching over a design flaw that we should just eliminate entirely instead.

**Code in your app should refer to the files that you author, not the files that get emitted from the build.**

This is already true almost everywhere, but index.html is an exception, and that is the source of this bug.

Here is the general idea of how I'd rather the index.html blueprint worked:

```diff
<!DOCTYPE html>
<html<% if(lang) { %> lang="<%= lang %>"<% } %>>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title><%= namespace %></title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    {{content-for "head"}}

-    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
-    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/<%= name %>.css">
+    <link integrity="" rel="stylesheet" href="./styles/app.css">

    {{content-for "head-footer"}}
  </head>
  <body>
    {{content-for "body"}}

-    <script src="{{rootURL}}assets/vendor.js"></script>
-    <script src="{{rootURL}}assets/<%= name %>.js"></script>
+    <script type="module" src="./app.js"></script>

    {{content-for "body-footer"}}
  </body>
</html>
```

The idea is, you're expressing *intent*. "My app's javascript goes here", "my app's CSS goes here". The links point at real files in the repo, in the same way that `import "./something"` in your JS refers to a real file in the repo, and it's up to the build to rewrite that as necessary to make it actually run in the browser.

The build would look for the `./app.js` script tag and replace it with one **or more** script tags that achieve the intended result of getting your app into the page. The traditional build always uses two, but that's not something we want to hard code, because it's a cache optimization decision that you should be able to tweak without changing the app's own source.

To actually make this change, we would need to figure out the upgrade & compatibility story, because it's definitely a different way of thinking of index.html. It would also definitely break embroider compatibility, but only because this design is already closer to how embroider really works under the hood, and we would need to detect it and turn off the compat code that is looking explicitly for a script tag with `{{rootURL}}{{outputPaths.app.js}}`, etc.

Probably the presence of a script tag with `type="module"` and `src` resolving to your app.js file is enough to reliably detect that we're using the "new" index.html format.

PR #8856, opened 2019-09-25
Prepare for first Octane beta.

rwjblue:
* Update ember-source to 3.14.0-beta.1
* Add `@ember/edition-utils` to new apps
* Move `.ember-cli` to `.ember-cli.js`
* Add octane defaulted optional features to
  `config/optional-features.json`
* `setEdition('octane')` for new apps / addons

### DIFF
--- a/blueprints/addon/files/config/optional-features.json
+++ b/blueprints/addon/files/config/optional-features.json
@@ -1,3 +1,6 @@
 {
-  "jquery-integration": false
+  "application-template-wrapper": false,
+  "default-async-observers": true,
+  "jquery-integration": false,
+  "template-only-glimmer-components": true
 }
--- a/blueprints/app/files/config/optional-features.json
+++ b/blueprints/app/files/config/optional-features.json
@@ -1,3 +1,6 @@
 {
-  "jquery-integration": false
+  "application-template-wrapper": false,
+  "default-async-observers": true,
+  "jquery-integration": false,
+  "template-only-glimmer-components": true
 }
--- a/blueprints/app/files/.ember-cli.js
+++ b/blueprints/app/files/.ember-cli.js
@@ -1,9 +1,13 @@
-{
+const { setEdition } = require('@ember/edition-utils');
+
+setEdition('octane');
+
+module.exports = {
   /**
     Ember CLI sends analytics information by default. The data is completely
     anonymous, but there are times when you might want to disable this behavior.
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
-}
+  disableAnalytics: false
+};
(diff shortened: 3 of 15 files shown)


Panman8201:
FWIW, PR #8791 needs to be finished too (I assume before an octane beta anyway).

rwjblue:
@Panman8201 - Ya, I'm not 100% sure we will ultimately decide to do #8791. I need to sync with @NullVoxPopuli on it still...

blueprints/addon/files/config/optional-features.json
```diff
@@ -1,3 +1,6 @@
 {
-  "jquery-integration": false
+  "default-async-observers": true,
```
kpfefferle: Nitpick: can this be moved under `application-template-wrapper` so they are alphabetical and match what you'd get if you run `ember feature:enable default-async-observers` on an app that already has the other three optional features?

blueprints/app/files/config/optional-features.json
```diff
@@ -1,3 +1,6 @@
 {
-  "jquery-integration": false
+  "default-async-observers": true,
```
kpfefferle: Nitpick: can this be moved under application-template-wrapper so they are alphabetical and match what you'd get if you run ember feature:enable default-async-observers on an app that already has the other three optional features?

blueprints/addon/files/config/optional-features.json
rwjblue: @kpfefferle - Great point, thank you for reviewing!

rwjblue:
> What are the main differences between this and the "Octane blueprint" after this?

Not much, we should diff once this lands. That is why I was saying I'm not sure that we actually need to do https://github.com/ember-cli/ember-cli/pull/8791.

Panman8201:
> Not much, but we should diff once this lands. That is why I was saying I'm not sure that we actually need to do #8791.

Ah, I see. This PR updates the blueprints as the exist/function today with the "Octane by default" things. @NullVoxPopuli PR runs the [Octane blueprints](https://github.com/ember-cli/ember-octane-blueprint) as a separate source, the way to "try Octane" today `ember new my-app -b @ember/octane-app-blueprint`, which would be a completely different approach to how it has worked up to this point.

rwjblue:
> This PR updates the blueprints as the exist/function today with the "Octane by default" things. @NullVoxPopuli PR runs the Octane blueprints as a separate source, the way to "try Octane" today ember new my-app -b @ember/octane-app-blueprint, which would be a completely different approach to how it has worked up to this point.

Yep, exactly!

chancancode:
@rwjblue turns out, one of the things that we forgot to include here is `@glimmer/component` 😬

rwjblue:
Ya, working on it (and some template-lint tweaks).

PR #8857, opened 2019-09-25
Tweaks to release scripts.

rwjblue:
* Update changelog generation system
  * Account for new style merge commit titles
  * Fetch the actual PR author (no more issues with sending a branch from
    ember-cli/ember-cli)
  * Ignore "dependabot" PRs
  * Add support for showing commits that are _not_ included in a PR
* Rewrite output repo updater in JavaScript (Bash 🤕 )


dev/update-output-repos.js
```diff
@@ -0,0 +1,68 @@
...
+
+const fs = require('fs-extra');
+const path = require('path');
+const execa = require('execa');
+const tmp = require('tmp');
+tmp.setGracefulCleanup();
+
+const currentVersion = require('../package').version;
+const EMBER_PATH = require.resolve('../bin/ember');
+const isStable = !currentVersion.includes('-beta');
```
Turbo87: should  we use the  semver package to parse for the actual version extension thingy? just in case we want to publish alphas eventually 😅

rwjblue: Ya, good point, will update.

PR #8861, opened 2019-09-27
Octane related tweaks

rwjblue:
* Add `@glimmer/component` 
* Update ember-template-lint configuration to use `octane` preset


kellyselden:
I take it the alpha will get bumped to 1.0.0 before 3.14?

rwjblue:
> I take it the alpha will get bumped to 1.0.0 before 3.14?

Confirm. Using a stable release is a 3.14 blocker.

Issue #8872, opened 2019-10-03
Sync beta blueprints with octane blueprints


chancancode:
This is the inverse of https://github.com/ember-cli/ember-octane-blueprint/issues/170. By the time 3.14 ships, we need them to be basically the same. Right now, the most notable omission is `ember-auto-import` in the beta blueprints.

kellyselden:
Would that be RFC worthy? I'm not up to speed with that project, but I remember things like a non trivial build speed reduction, and making sure it matches your Babel config or else errors. Would the guides be prepared to teach this in time for octane? (Instead of app.import and in repo add-on shims)

Panman8201:
FYI, I've submitted #8873 which completes this task.

PR #8873, opened 2019-10-03
App and addon blueprint updates to match octane blueprints

Panman8201:
I basically diffed the output from app & addon blueprints between ember-cli and [ember-octane-blueprint](https://github.com/ember-cli/ember-octane-blueprint) pulling out the important differences and including them in this PR. (I did the [same thing with the octane blueprints started](https://github.com/ember-cli/ember-octane-blueprint/pull/105).)

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -8,7 +8,7 @@
 # misc
 /.bowerrc
 /.editorconfig
-/.ember-cli
+/.ember-cli.js
 /.env*
 /.eslintignore
 /.eslintrc.js
--- a/blueprints/app/files/.ember-cli.js
+++ b/blueprints/app/files/.ember-cli.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const { setEdition } = require('@ember/edition-utils');
 
 setEdition('octane');
--- a/blueprints/app/files/app/app.js
+++ b/blueprints/app/files/app/app.js
@@ -3,12 +3,10 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
-const App = Application.extend({
-  modulePrefix: config.modulePrefix,
-  podModulePrefix: config.podModulePrefix,
-  Resolver
-});
+export default class App extends Application {
+  modulePrefix = config.modulePrefix;
+  podModulePrefix = config.podModulePrefix;
+  Resolver = Resolver;
+}
 
 loadInitializers(App, config.modulePrefix);
-
-export default App;
--- a/blueprints/app/files/app/router.js
+++ b/blueprints/app/files/app/router.js
@@ -1,12 +1,10 @@
 import EmberRouter from '@ember/routing/router';
 import config from './config/environment';
 
-const Router = EmberRouter.extend({
-  location: config.locationType,
-  rootURL: config.rootURL
-});
+export default class Router extends EmberRouter {
+  location = config.locationType;
+  rootURL = config.rootURL;
+}
 
 Router.map(function() {
 });
-
-export default Router;
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -23,6 +23,7 @@
     "@glimmer/component": "^0.14.0-alpha.13",
     "babel-eslint": "^10.0.3",
     "broccoli-asset-rev": "^3.0.0",
+    "ember-auto-import": "^1.5.2",
     "ember-cli": "~<%= emberCLIVersion %>",
     "ember-cli-app-version": "^3.2.0",
     "ember-cli-babel": "^7.11.1",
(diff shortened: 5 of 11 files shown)


blueprints/addon/index.js
```diff
@@ -57,6 +57,10 @@ module.exports = {
     contents.dependencies['ember-cli-htmlbars'] = contents.devDependencies['ember-cli-htmlbars'];
     delete contents.devDependencies['ember-cli-htmlbars'];
 
+    // Move ember-auto-import into the dependencies of the addon blueprint by default
+    contents.dependencies['ember-auto-import'] = contents.devDependencies['ember-auto-import'];
```
rwjblue: We should discuss this a bit more. I'm not sure that we want to make all addons depend on ember-auto-import 🤔 

@Turbo87 / @ef4 - Thoughts?

Turbo87: yeah, that's what I said on Discord too. I think it's fine adding it to apps, but I think for addons the authors should have to manually add it if they really need to. I'm assuming that with embroider the whole story will change again anyway.

Panman8201: Yeah, I wasn't sure but it *was* included in the octane blueprints so I threw it in there. Also brought it in the #st-octane channel in Discord and @chancancode said "it's a bug, it should include it", linking to this: https://www.notion.so/Octane-Preview-3-13-FAQ-and-Known-Issues-bbd25c2ffb0e4f079103db1fc0144175#284a1ee8f26448d6bd77febc2583aa7f

NullVoxPopuli:
idk -- it should use the blueprint repo/package, right? leaving the classic one alone?

Panman8201:
> idk -- it should use the octane blueprint repo/package, right? leaving the classic one alone?

This PR is just finishing up the "sync" with the Octane blueprints, which was already started & merged by @rwjblue.

blueprints/addon/index.js
kellyselden: I hate to be a bummer, but it would be nice to discuss if it is ready to be added by default at all. More here https://github.com/ember-cli/ember-cli/issues/8872#issuecomment-538023305.

ef4: I agree, I think adding ember-auto-import to the default blueprint has significant ecosystem-wide effects that need to be aired in an RFC. My own notes on what I see as the desirable TODOs before doing that are here: https://github.com/ef4/ember-auto-import/issues/182

ef4: I want to clarify that my caveats are almost all about usage by *addons*. Having it in apps is less risky and a huge ergonomics win, so I'm in favor of having it in the app blueprint.

(The difference is: addons need to just work with whatever version of auto-import their consuming app is using. Even future ones. We can gracefully roll out breaking changes to app authors, we cannot easily to addon authors.)

Panman8201: Sounds like there is a relative consensus to include it for apps but not addons.?. Let me know which way it needs to go and I'll make the changes, or feel free to push a commit to the PR.

rwjblue: Ya, I think we are on the same page now. Apps should get it by default, and addons would only have it for their dummy apps (in other words, it should stay in `devDependencies`).

rwjblue:
> idk -- it should use the octane blueprint repo/package, right? leaving the classic one alone?

I think we've discussed this a bit, but for 3.14 I think the `app` and `addon` blueprints should continue to be provided by ember-cli directly. I think most of us (cli team, framework team) are 👍 on the conceptual idea of moving these blueprints to be external packages, but there is a bit of thought that needs to go into how to manage things (like versioning). I don't think any of us wants this to turn into an Octane blocker, so decoupling that separation from Octane seems much preferable to me.

ef4:
> @rwjblue I'm a little hesitant to add it to the dummy apps by default too, because it would make it seem that it works in the addon code itself too. Similar issue to using `ember-maybe-import-regenerator`, I guess.

ember-auto-import itself guards against this risk. If you use it as a devDependency, it will only function inside the dummy app's files. The addon's files aren't able to auto-import anything.

Turbo87:
okay, if that's the case then I'm fine with it 👍

blueprints/addon/index.js
rwjblue: @kellyselden - I did some local tests between a 3.14.0-beta.2 app with and without `ember-auto-import` as a dependency. Adding `ember-auto-import` adds _roughly_ 2s to the initial build time (for cold builds they went from 17.67s to 19.73s; warm builds went from 7.94 to 8.96), and rebuild time approximately 93ms (went from 213ms to 306ms). These times seem fine to me (considering the significant improvement to development ergonomics).

Panman8201:
Ok, changed PR to leave `ember-auto-import` in `devDependencies` for the addon as well (tests running now). Is "Squash & Merge" enabled for this repo or should I `git rebase -i` to combine all my commits (if needed)?

rwjblue:
@Panman8201 - I'll take care of the rebase, thank you for working on this!

rwjblue:
@Panman8201 - Thank you again!

kellyselden:
Any apps that are `app.import`ing AMD assets so ES6 import shims work are going to get doubled assets if there is also a CommonJS version found. @ef4 said there may be a way to detect this.

kellyselden:
more info https://github.com/ef4/ember-auto-import/issues/250

PR #8878, opened 2019-10-08
Adds flag to throw for invalid addon locations

tmquinn:
This also surfaces the warning outside of the logger

### DIFF
--- a/lib/models/package-info-cache/package-info.js
+++ b/lib/models/package-info-cache/package-info.js
@@ -360,7 +360,7 @@ class PackageInfo {
         });
       }
 
-      if (process.env.ERROR_ON_INVALID_ADDON) {
+      if (process.env.EMBER_CLI_ERROR_ON_INVALID_ADDON) {
         throw msg;
       } else {
         this.project.ui.writeWarnLine(msg);
--- a/lib/models/package-info-cache/package-info.js
+++ b/lib/models/package-info-cache/package-info.js
@@ -346,20 +346,24 @@ class PackageInfo {
     if (invalidPackages.length > 0) {
       let typeName = this.project ? 'project' : 'addon';
 
-      logger.info('');
-      logger.info(`The 'package.json' file for the ${typeName} at ${this.realPath}`);
+      let msg = `The 'package.json' file for the ${typeName} at ${this.realPath}`;
 
       let relativePath;
-
       if (invalidPackages.length === 1) {
         relativePath = path.relative(this.realPath, invalidPackages[0].realPath);
-        logger.info(`  specifies an invalid, malformed or missing addon at relative path '${relativePath}'`);
+        msg = `${msg}\n  specifies an invalid, malformed or missing addon at relative path '${relativePath}'`;
       } else {
-        logger.info('  specifies invalid, malformed or missing addons at relative paths');
+        msg = `${msg}\n  specifies invalid, malformed or missing addons at relative paths`;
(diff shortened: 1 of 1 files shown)


lib/models/package-info-cache/package-info.js
```diff
@@ -346,21 +347,26 @@ class PackageInfo {
...
+        msg = `${msg}\n  specifies invalid, malformed or missing addons at relative paths`;
         invalidPackages.forEach(packageInfo => {
           let relativePath = path.relative(this.realPath, packageInfo.realPath);
-          logger.info(`    '${relativePath}'`);
+          msg = `${msg}\n    '${relativePath}'`;
         });
       }
+
+      const config = require('../../utilities/get-config')();
+      if (config.get('errorOnInvalidAddon')) {
```
rwjblue: Why is this a flag and not the default? What circumstances are we intentionally wanting to swallow invalid addon paths for?

tmquinn: I know, right? This has bitten us many times, I think the reason for the flag is that the behavior is radically different from the current behavior. Previously it didn't even throw a warning without the DEBUG flag set...

rwjblue: So the reason it is behind a flag is that it is a potentially breaking change?

tmquinn: Ya, per @stefanpenner

rwjblue: I think we should:

* have it always emit a warning to the console (this is the only way to get these cleaned up throughout the ecosystem)
* have it throw an error when a `process.env` variable is set

tmquinn: OK, so this PR does the first one already, so I will just switch the flag from .ember-cli to an env...

lib/models/package-info-cache/package-info.js
```diff
@@ -346,21 +347,26 @@ class PackageInfo {
...
+        throw msg;
+      } else {
+        console.warn(chalk.yellow(msg));
```
rwjblue: Do we not have access to a console-ui instance here? In general, we should try not to use `console` if we can help it 🤔

tmquinn:
I looked around for it... it's not on this in this context...

lib/models/package-info-cache/package-info.js
rwjblue: We (at least) sometimes have `this.project` which has `this.project.ui` on it.

lib/models/package-info-cache/package-info.js
```diff
@@ -346,21 +346,25 @@ class PackageInfo {
...
-        logger.info('  specifies invalid, malformed or missing addons at relative paths');
+        msg = `${msg}\n  specifies invalid, malformed or missing addons at relative paths`;
         invalidPackages.forEach(packageInfo => {
           let relativePath = path.relative(this.realPath, packageInfo.realPath);
-          logger.info(`    '${relativePath}'`);
+          msg = `${msg}\n    '${relativePath}'`;
         });
       }
+
+      if (process.env.ERROR_ON_INVALID_ADDON) {
```
rwjblue: Let's prefix this with `EMBER_CLI_`:

```suggestion
      if (process.env.EMBER_CLI_ERROR_ON_INVALID_ADDON) {
```

rwjblue:
Thank you!

PR #8882, opened 2019-10-09
blueprints/init: Simplify "Get started" message

dcyriller:
This fixes a typo in the message.

Closes https://github.com/ember-learn/guides-source/issues/1133.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -112,7 +112,7 @@ module.exports = Command.extend({
 
         this.ui.writeLine('');
         this.ui.writeLine(prependEmoji('🎉', `Successfully created project ${chalk.yellow(projectName)}.`));
-        this.ui.writeLine(prependEmoji('👉 ', 'We suggest you that you get started by typing:'));
+        this.ui.writeLine(prependEmoji('👉 ', 'Get started by typing:'));
         this.ui.writeLine('');
         const command = `cd ${projectName}`;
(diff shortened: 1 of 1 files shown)


lib/commands/init.js
```diff
@@ -112,12 +112,11 @@ module.exports = Command.extend({
 
         this.ui.writeLine('');
         this.ui.writeLine(prependEmoji('🎉', `Successfully created project ${chalk.yellow(projectName)}.`));
-        this.ui.writeLine(prependEmoji('👉 ', 'We suggest you that you get started by typing:'));
+        this.ui.writeLine(prependEmoji('👉 ', 'Get started by typing:'));
         this.ui.writeLine('');
         const command = `cd ${projectName}`;
         this.ui.writeLine(`  ${chalk.gray('$')} ${chalk.cyan(command)}`);
-        const packageManager = commandOptions.yarn ? 'yarn' : 'npm';
-        this.ui.writeLine(`  ${chalk.gray('$')} ${chalk.cyan(`${packageManager} start`)}`);
```
rwjblue: 😢 I actually like this better. It works much better with teams of folks, and especially if folks do `npx ember-cli new foo` (because they wouldn't even have a global `ember` installed).

dcyriller: I've removed the last commit.

rwjblue:
Thank you!!!

Issue #8863, opened 2019-09-27
Error when running `ember new` with `--skip-npm` on 3.14.0-beta.1+


rwjblue:
Ember 3.14.0-beta.1 introduced some changes to the project config (namely moving from `.ember-cli` to `.ember-cli.js`), and using `setEdition('octane')` in that project config. During any `ember-cli` command we instrument run times and what not so that we have a way to profile command execution so that we can improve it. Unfortunately, when we attempt to stop instrumentation we discover the new `.ember-cli.js` file and attempt to `require` it which throws an error (essentially a module not found for `@ember/edition-utils`) since we specifically instructed it to avoid installing npm packages.

The rough issue looks like:

```
ember new bar --skip-npm
installing app
Ember CLI v3.14.0-beta.2

✨  Creating a new Ember app in /Users/rjackson/sandbox/bar:
  create .editorconfig
  create .ember-cli.js
  create .eslintignore
  create .eslintrc.js
  create .template-lintrc.js
  create .travis.yml
  create .watchmanconfig
  create README.md
  create app/app.js
  create app/components/.gitkeep
  create app/controllers/.gitkeep
  create app/helpers/.gitkeep
  create app/index.html
  create app/models/.gitkeep
  create app/resolver.js
  create app/router.js
  create app/routes/.gitkeep
  create app/styles/app.css
  create app/templates/application.hbs
  create app/templates/components/.gitkeep
  create config/environment.js
  create config/optional-features.json
  create config/targets.js
  create ember-cli-build.js
  create .gitignore
  create package.json
  create public/robots.txt
  create testem.js
  create tests/helpers/.gitkeep
  create tests/index.html
  create tests/integration/.gitkeep
  create tests/test-helper.js
  create tests/unit/.gitkeep
  create vendor/.gitkeep

🎥  Initializing git repository.
Git: successfully initialized.

🎉  Successfully created project bar.
👉   We suggest you that you get started by typing:

  $ cd bar
  $ npm start

Happy coding!
Cannot read config file: /Users/rjackson/sandbox/bar/.ember-cli.js
```

sukima:
This is happening for latest Ember in NPM now when using the octane blueprint (3.13.1)

chancancode:
Yeah, basically the problem happens when you have a `.ember-cli.js` but didn't have the required packages installed. For the time being, it should be safe to ignore the error. Things should work normally after running npm install or yarn install to install the required packages.

locks:
[...]

PR #8899, opened 2019-10-23
[BUGFIX] Don't reload the config for instrumentation

pzuraq:
We currently will reload the config to setup instrumentation, which can
happen multiple times while the CLI is running. This sometimes results
in failures when the CLI generates new applications that haven't
installed `node_modules` yet, and in other cases. We really only need to
load the config once, so this PR changes it so that we do.

Fixes https://github.com/ember-cli/ember-cli/issues/8863


rwjblue:
Thank you @pzuraq!

PR #8900, opened 2019-10-23
[BUGFIX] Include legacyDecorators eslint option by default

pzuraq:
Ember supports the legacy decorators proposal, which originally
placed the decorators _before_ the export keywords. However, eslint
now throws errors if this is used, which is inline with the stage 2
spec that was rejected. This enables the legacyDecorators flag in
eslint to allow decorator placement anywhere.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -3,7 +3,10 @@ module.exports = {
   parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2018,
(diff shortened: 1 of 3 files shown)


mehulkar:
@pzuraq noticed this in my upgrade. Do you have an example of before and after? I'm not sure I understand what you mean by "placed decorators before the export keyword"

pzuraq:
```js
@decorator
export default MyClass {}

// vs
export default
@decorator
MyClass {}
```

PR #8901, opened 2019-10-23
[BUGFIX beta] Merge with existing `EmberENV`

chancancode:
This allows Ember Inspector or other code in `index.html` before the vendor script tag to set values in `EmberENV`.

### DIFF
--- a/lib/broccoli/vendor-prefix.js
+++ b/lib/broccoli/vendor-prefix.js
@@ -1,4 +1,11 @@
-window.EmberENV = {{EMBER_ENV}};
+window.EmberENV = (function(EmberENV, extra) {
+  for (var key in extra) {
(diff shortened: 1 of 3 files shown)


Turbo87:
why is this labeled as "bugfix"? what bug does it fix?

chancancode:
@turno87 currently, the output produced by Ember CLI prevents anything else from setting the EmberENV global (since it replaces it at the top of the vendor bundle). Since this is also the mechanism used by the inspector in 3.14+ to opt into more detailed debug information (“debug render tree” feature), this is currently preventing the inspector from working as designed.

chancancode:
It’s not an inspector specific problem, just that this is probably the first time we noticed the problem since the inspector need to set this variable before ember loads.

PR #8905, opened 2019-10-31
Use production environment for `npm run build` / `yarn build` by default

pichfl:
Output from `npm run build` should be optimized by default to make sure people deploy optimized versions of Ember.

This ensures that even newcomers or people comparing frameworks will compare the optimized versions and not the development builds.

It also ensures that builds on serverless services will be optimized out of the box. See [now.sh](https://zeit.co/guides/deploying-ember-with-zeit-now/) for a most recent example.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -11,7 +11,7 @@
     "test": "tests"
   },
   "scripts": {
-    "build": "ember build",
+    "build": "ember build --environment=production",
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
--- a/tests/fixtures/addon/npm/package.json
+++ b/tests/fixtures/addon/npm/package.json
@@ -13,7 +13,7 @@
     "test": "tests"
   },
   "scripts": {
-    "build": "ember build",
+    "build": "ember build --environment=production",
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
--- a/tests/fixtures/addon/yarn/package.json
+++ b/tests/fixtures/addon/yarn/package.json
@@ -13,7 +13,7 @@
     "test": "tests"
   },
   "scripts": {
-    "build": "ember build",
+    "build": "ember build --environment=production",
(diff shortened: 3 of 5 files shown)


mansona:
@rwjblue I just read through that RFC and while I do agree that it is something to consider and potentially push forward, I don't think that we should block this change on implementing it that way. Let me explain 😂 

Firstly I agree that the current situation is quite confusing and we should think about "cleaning up" the logic around this to make the defaults more sensible. When I discussed this with @pichfl it was mainly around the making sure that the command `npm run build` does the right thing out of the box (and hence this change). 

We also think that this is a very low impact change that can be merged now without any concern for backwards compatibility because: 

- it is very explicit - if someone were to look at the package.json and see `--environment=production` they would see that the `build` script is for building a production environment
- it's discoverable - right now people need to "know" from reading external documentation that they need to do a production build. Seeing `--environment=production` in the build script might prompt them to google to see what it means
- this change only affects **new** apps - as this is only changing the generator this will not affect anyone that updates the ember-cli package
- the change is explicit - if someone uses `ember-cli-update` and they see this in the diff they can choose to delete it and not update (if they don't want to)
- and lastly, any change that we make via the RFC we will need to support using `--environment=production` until at least ember-cli@4 so it has a relatively low impact adding this to the build script in the middle of the 3.x cycle

There is also a general "don't let best be the enemy of good" vibes to what I'm saying but I think that comes across with my bulleted list 😂

mansona:
I would  really like to see this landed 🙈 It seems that quite a few people are trying out Ember again thanks to Octane and they are getting confused by this exact issue 😫 https://news.ycombinator.com/item?id=21848806

mansona:
@wycats you suggested "changing the default" [here](https://news.ycombinator.com/item?id=21848919) but you seem to be suggesting that you want `ember build` to potentially have a different default output? 🤔 is that right?

rwjblue:
I'll try to chat with the ember-cli team about this, and update the thread here. The main discussion is around whether or not we need to consider this a bugfix (and can "just do it") or something that is more fundamental / mental model related and therefore requires an RFC.

PR #8907, opened 2019-11-05
Utilize a more universal .ui

tmquinn:
Addon packageInfos do not have a project. All packageInfos have a cache.ui

### DIFF
--- a/tests/unit/models/package-info-cache/package-info-cache-test.js
+++ b/tests/unit/models/package-info-cache/package-info-cache-test.js
@@ -175,6 +175,47 @@ describe('models/package-info-cache/package-info-cache-test.js', function() {
   });
 
   describe('packageInfo', function() {
+    describe('project with invalid paths', function() {
+      let project, fixturifyProject;
+      beforeEach(function() {
+        // create a new ember-app
+        fixturifyProject = new FixturifyProject('simple-ember-app', '0.0.0', project => {
+          project.addAddon('ember-resolver', '^5.0.1');
+          project.addAddon('ember-random-addon', 'latest');
+          project.addAddon('loader.js', 'latest');
+          project.addAddon('something-else', 'latest');
+          project.addInRepoAddon('ember-super-button', 'latest', function(project) {
+            project.pkg['ember-addon'].paths = ['lib/herp-not-here'];
+          });
+          project.addDevDependency('ember-cli', 'latest');
+          project.addDevDependency('non-ember-thingy', 'latest');
+          project.pkg['ember-addon'].paths.push('lib/no-such-path');
+        });
+
+        fixturifyProject.writeSync();
+
+        project = fixturifyProject.buildProjectModel(Project);
+      });
+
+      afterEach(function() {
+        fixturifyProject.dispose();
+      });
+
+      it('shows a warning with invalid ember-addon#path', function() {
+        project.discoverAddons();
+        expect(project.cli.ui.output).to.include(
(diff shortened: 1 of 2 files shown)


stefanpenner:
@rwjblue I'm all about testing, but is it worth it here?

rwjblue:
> @rwjblue I'm all about testing, but is it worth it here? 

@stefanpenner - I'm pretty surprised actually, I've been chided by you many times (rightfully so!) for not requiring test changes. For example, if I had required tests on the original PR we wouldn't have this regression...

> can you provide a recommendation how to best test this?

I think the test would go in this module:

https://github.com/ember-cli/ember-cli/blob/c1d0070641069a9de7673484ed92b55a37b58506/tests/unit/models/package-info-cache/package-info-cache-test.js#L177

It would:

* use `FixturifyProject` (like the `valid project` describe block just below)
* set up an _invalid_ addon reference (exactly the kind of thing this is supposed to catch)
* call `project.discoverAddons()`
* assert that the `mock-ui` used in the test has the information printed to it

For extra credit, add another test just like that but with the environment variable set that causes it to throw an error, and assert the error is thrown.

tmquinn:
I'm on it.

tmquinn:
Running into an issue. So the point is to have a path, that does not have a corresponding directory. So I am just `fixturifyProject.pkg['ember-addons'] = ['../path/to/nowhere'];` instead of using `.addInRepoAddon()`, unfortunately this makes `fixturifyProject.buildProjectModel(Project)` crash...

Trying to step thru and see if there's something I can do to fix this, but I just wanted to make sure this was what you intended me to do, or if I've missed the mark.

tmquinn:
uggh.. crap.. I see one thing I did wrong... hold on for update

tmquinn:
ok, putting the array in paths fixed the previous issue, but now discoverAddons() is crashing... I'll let you know what I find

tmquinn:
OK, it's not crashing, it's now timing out because I am in DevTools.... so I tried using one of the fixtures checked in to the repo, adding a bad path, and not seeing any results in output...? Will carry on though...

tmquinn:
@rwjblue could you expand on

> set up an invalid addon reference (exactly the kind of thing this is supposed to catch)

tmquinn:
I KNEW as soon as I @'d you I'd figure it out... ok NVM on previous question

tmquinn:
OK, so I've added two tests to cover the bad paths at the Project level, but this regression occurred because of a bad path in an Addon. It does not look like discoverAddons() runs at an addon level. Is there a way to do this?

tests/unit/models/package-info-cache/package-info-cache-test.js
```diff
@@ -175,6 +175,47 @@ describe('models/package-info-cache/package-info-cache-test.js', function() {
...
+          "specifies an invalid, malformed or missing addon at relative path 'lib/no-such-path'"
+        );
+      });
+
+      it('throws an error with flag on', function() {
+        process.env.EMBER_CLI_ERROR_ON_INVALID_ADDON = 'true';
+        expect(() => project.discoverAddons()).to.throw(
+          /specifies an invalid, malformed or missing addon at relative path 'lib\/no-such-path'/
+        );
+        delete process.env.EMBER_CLI_ERROR_ON_INVALID_ADDON;
```
stefanpenner: For safety we should likely do the globaly mutation of `process.env` in beforeEach/afterEach or in a try/finally. Otherwise we risk test failures causing test leakage.

PR #8910, opened 2019-11-06
restrict app Travis to master and PRs

kellyselden:
We did this to addons, where they are restricted to master, PRs, and npm tags. I think it makes sense to do the same with apps, minus the npm tags?


rwjblue:
Seems good to me.

PR #8915, opened 2019-11-08
Revert changes made to enable "octane" as the default for ember new

rwjblue:
Reverts octane specific changes that landed in ember-cli under the assumption that 3.14 would be "octane" now that we have decided that there is a bit of work yet to do before the octane experience is the new default.

See the blog post for more details:

https://blog.emberjs.com/2019/10/31/octane-release-update.html


auvipy:
will this be default in ember 3.16?

kellyselden:
@auvipy It's coming back in ember-cli 3.15

PR #8924, opened 2019-11-13
fix named UMD imports

kellyselden:
We've run into a package that publishes webpack-built named UMD files.

https://github.com/krisk/Fuse/blob/v3.4.5/configs/webpack.common.js#L49-L50
https://github.com/krisk/Fuse/blob/v3.4.5/dist/fuse.js#L9

The problematic line is

```js
"object"==typeof exports&&"object"==typeof module?module.exports=t():"function"==typeof define&&define.amd?define("Fuse",[],t):"object"==typeof exports?exports.Fuse=t():e.Fuse=t()
```

ember-cli assumes you are calling `define` without checking `define.amd` with a name, or checking `define.amd` and calling `define` without a name.

This means that calling

```js
app.import('node_modules/fuse.js/dist/fuse.js');
```

misses the `define.amd` check because we only set it to true when explicitly using the amd transform.

Also calling

```js
app.import('node_modules/fuse.js/dist/fuse.js', {
  using: [{
    { transformation: 'amd', as: 'Fuse' }
  }]
});
```

fails because ember-cli tries to double name it, since it's already a named define.

This solution fixes the problem by allowing

```js
app.import('node_modules/fuse.js/dist/fuse.js', {
  using: [{
    { transformation: 'amd' }
  }]
});
```

which sets `define.amd`, but doesn't try to name it.

We could also solve this by always setting `define.amd`, then all you would need is

```js
app.import('node_modules/fuse.js/dist/fuse.js');
```

but I don't know what kind of consequences that would have.


lib/tasks/transforms/amd/index.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
-const stew = require('broccoli-stew');
+// Why is this rule erroring here?
```
kellyselden: Any ideas? It is definitely published.

lib/broccoli/amd-shim.js
```diff
@@ -4,16 +4,17 @@ const stew = require('broccoli-stew');
...
-        name,
-        '"); return define.apply(null, args); }; newDefine.amd = true; return newDefine; })());',
-      ].join('');
-    } else {
-      return content;
+      sections.push(' args.unshift("');
+      sections.push(name);
+      sections.push('");');
     }
+    sections.push(' return define.apply(null, args); }; newDefine.amd = true; return newDefine; })());');
```
rwjblue: We have intentionally avoided specifying `define.amd`, our loader does **not** implement the full spec behavior (which is what `define.amd` is trying to imply).

lib/tasks/transforms/amd/index.js
rwjblue: Probably because there is a package.json in this directory. You’d need to add broccoli-stew to that package.json (like with a version constraint of `*`).

kellyselden: I removed broccoli-stew. I guess I need to get `'../../../broccoli/amd-shim'` in the package.json somehow.

lib/broccoli/amd-shim.js
kellyselden: I still keep the `newDefine` override convention that the existing AMD shim has. Did I break it in some way?

kellyselden:
@rwjblue Are there any outstanding issues preventing merge?

PR #8925, opened 2019-11-14
.npmignore => files for addons

kellyselden:
This has been bothering me for a long time. I've witnessed countless burns by things accidently getting published, whether it be private keys, IDE configs, or other differences between local and CI publishing. I've started doing this on all my non-ember addon packages and couldn't be happier.

The only downside is now instead of accidentally publishing too much, you may accidentally forget to publish something. This is an acceptable tradeoff in my opinion.


rwjblue:
Historically, I have resisted using `files` in `package.json` because I think it's pretty trolly (add a new folder with some shared utils, things seem fine, publish package :boom: ), but I think at this point as long as we are using `eslint-plugin-node` and running linting across `.` (instead of our older technique of running against specific foldres) that that "troll factor" is mitigated.

blueprints/addon/index.js
```diff
@@ -86,6 +86,19 @@ module.exports = {
     contents['ember-addon'] = contents['ember-addon'] || {};
     contents['ember-addon'].configPath = 'tests/dummy/config';
 
+    contents.files = [
+      'addon/**/!(.gitkeep)',
```
rwjblue: I don't think we need a complicated regexp here, publishing a few (maybe just one?) empty file is not an issue. Lets remove these globs in favor of:

```suggestion
      'addon',
```

blueprints/addon/index.js
```diff
@@ -86,6 +86,19 @@ module.exports = {
...
+      'addon-test-support',
+      'app/**/!(.gitkeep)',
+      'blueprints',
+      'config/**/!(ember-try.js)',
+      'lib',
+      'public',
+      'test-support',
+      'vendor/**/!(.gitkeep)',
```
rwjblue:
```suggestion
      'vendor',
```

blueprints/addon/index.js
```diff
@@ -86,6 +86,19 @@ module.exports = {
     contents['ember-addon'] = contents['ember-addon'] || {};
     contents['ember-addon'].configPath = 'tests/dummy/config';
 
+    contents.files = [
+      'addon/**/!(.gitkeep)',
+      'addon-test-support',
+      'app/**/!(.gitkeep)',
```
rwjblue:
```suggestion
      'app',
```

blueprints/addon/index.js
```diff
@@ -86,6 +86,19 @@ module.exports = {
...
+      'blueprints',
+      'config/**/!(ember-try.js)',
```
rwjblue: In general I think we should move `config/ember-try.js` into `tests/dummy/config/ember-try.js`.

```suggestion
      'config',
```

kellyselden: Can ember-try find it there?

rwjblue: Not sure, but we should make sure it can (if it can't) 😸

blueprints/addon/index.js
kellyselden: Could I convince you otherwise? I'd prefer for it to not be worse than before. Depending on the addon, this would be three files: addon, app, and vendor.

rwjblue: A few thoughts:

* they are three completely empty files and take up basically no weight
* we _probably_ shouldn't even be creating them to begin with 😛 
* we should lint for `.gitkeep` files in directories with other contents

kellyselden: I can agree with the second two points. Is that enough reason to start publishing these files? Could we iterate and have these globs temporarily?

rwjblue: > Could we iterate and have these globs temporarily?

We could, but I find the extra blueprint churn worse than the current state...

rwjblue:
@kellyselden - Let's get this landed, mind rebasing?

blueprints/addon/index.js
rwjblue: Dropping this concern, let's move forward...

Turbo87:
idk, I tend to prefer `.npmignore` 😅

kellyselden:
rebased

locks:
Ping @rwjblue @Turbo87!

Turbo87:
@locks my opinion has not changed. I'd rather have people ship too many files than not enough files 🤷‍♂️

rwjblue:
Ya, I tend to agree. `files` is a massive footgun...

kellyselden:
I disagree with both. `.npmignore` is a footgun. I'd rather have to hotfix because I missed a file than to have a security issue because I published a sensitive file.

locks:
Closing due to lack of consensus.

PR #8926, opened 2019-11-14
add application to invalid names

### LINKED ISSUE #8879: Naming an app "application" causes issues.
1. Run `ember init` inside a directory named `application`.
2. run `ember serve`

404 errors on `application.css` and `applicaion.js`, as these get named `app.*` on init. If the directory name is `app` the `ember init` call is rejected as invalid name.

In our case we are a single app company, with both front and and back end in a Yarn workspaces managed monorepo. Since the app is named for the company, Tara AI, and we are using Workspaces, we thought the app's `package.json` name should be `@taraai/app`, this was not allowed, so we tried `@taraai/application`, which had these bugs. (yes, I know the directory name does not need to match, but was trying to keep it simple for everyone) 

We can of course use other generic names for our application (program, webapp, etc) but CLI should either be more permissive of names, or more explicit on which ones will cause issues.

---

```
ember-cli: 3.13.1
http_parser: 2.8.0
node: 10.16.0
v8: 6.8.275.32-node.52
uv: 1.28.0
zlib: 1.2.11
brotli: 1.0.7
ares: 1.15.0
modules: 64
nghttp2: 1.34.0
napi: 4
openssl: 1.1.1b
icu: 64.2
unicode: 12.1
cldr: 35.1
tz: 2019a
os: darwin x64
6.9.0
1.17.3
```

kennethlarsen:
This PR adds `application` to the list of invalid app names since it is causing issues.

Closes #8879

### DIFF
--- a/lib/utilities/valid-project-name.js
+++ b/lib/utilities/valid-project-name.js
@@ -22,7 +22,7 @@ function startsWithNumber(name) {
  * @return {Boolean}
  */
 module.exports = function(name) {
-  let invalidProjectNames = ['test', 'ember', 'ember-cli', 'vendor', 'public', 'app'];
+  let invalidProjectNames = ['test', 'ember', 'ember-cli', 'vendor', 'public', 'app', 'application'];
   name = name.toLowerCase();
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you @kennethlarsen!

PR #8929, opened 2019-11-15
Drop Node 11 support in generated apps and addons

SergeAstapov:


rwjblue:
Eeck, sorry that I missed this for so long! Would you mind rebasing?

PR #8930, opened 2019-11-15
Drop Node 11 support

SergeAstapov:


rwjblue:
Can you also update https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md ?

SergeAstapov:
@rwjblue I did my best to pick proper Ember CLI versions for the support table.

Also added Node.js 13 as supported as of CLI v3.15 (might have wrong assumption)

SergeAstapov:
Updated Node.js supported matrix in #8932

PR #8927, opened 2019-11-14
feat(blueprint): allow cloning git repo with ref

buschtoens:
This PR adds the ability to checkout a specific ref (branch, tag, commit sha) when using a git repo as the blueprint source, similar to how you can specify named refs as npm dependencies.

```bash
ember new -b https://github.com/ember-cli/app-blueprint-test.git#named-ref
```

It also adds [`--single-branch`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---no-single-branch) and [`--depth 1`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt) to the `git clone` command to speed up the checkout in general.

PR #8941, opened 2019-11-25
feat(blueprint): resolve remote blueprints via package manager

buschtoens:
Builds on and closes #8927.

This PR refactors the `install-blueprint` task to always use npm for installing remote blueprints. This means that users can now invoke `ember new -b` (and friends) with exactly the same package specifiers as they would pass to `npm` / `yarn`. Valid invocations include, but are not limited to:

- `ember new foo -b ember-cli/app-blueprint-test`
- `ember new foo -b ember-cli/app-blueprint-test#named-ref`
- `ember new foo -b gitlab:foo/bar`
- `ember new foo -b https://github.com/ember-cli/app-blueprint-test.git#named-ref`

/cc @rwjblue


buschtoens:
Hm, a lot of tests in ['unit/tasks/install-blueprint-test.js'](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/install-blueprint-test.js) are now failing, of course. TBH I'm not really sure whether they actually add a lot of value, since most of the heavy lifting is done by npm and we have sufficient coverage by the acceptance test.

rwjblue:
Ya, I'm game to remove the unit tests (since we are actually removing our custom functionality for this also).

buschtoens:
Seems like there's a valid failure with Node v8. I'll look into it.

buschtoens:
So turns out that `npm@6.4.1`, the version that ships with `node@8`, is buggy. 🙄 

My (apparently too) clever trick to install the package of unknown name into a predefined location does not reliably work with git dependencies:

```json
{
  "dependencies": {
    "ember-cli-blueprint-1574873174904": "git+https://github.com/ember-cli/app-blueprint-test.git"
  }
}
```

In the current `npm` version the above works as expected and installs the blueprint to:

```
node_modules/ember-cli-blueprint-1574873174904/
```

In `npm@6.4.1` only the `package-lock.json` is created and `node_modules` is missing.
_Interestingly_, when I change `npm install` to `npm install --no-package-lock`, the `node_modules` directory _is_ created. Unfortunately the package is not installed at the predefined location but at:

```
node_modules/ember-blueprint-demo/
```

Which is the name used in the repo's [`package.json`](https://github.com/ember-cli/app-blueprint-test/blob/master/package.json#L2).

---

One solution would be:

```sh
echo "{}" > package.json
npm install --no-package-lock --save https://github.com/ember-cli/app-blueprint-test.git
```

This would create a `package.json` with only a single dependency, that has the correct name:

```json
{
  "dependencies": {
    "ember-blueprint-demo": "git+https://github.com/ember-cli/app-blueprint-test.git"
  }
}
```

buschtoens:
Alright, looking good, I think! 😊 
Let me know, whether you'd like me to add more tests or change something.

PR #8946, opened 2019-11-26
replace addon testing ember 3.4 for 3.12

kellyselden:
We are overdue for another bump in latest 2 ember LTS for addon testing support. This swaps 3.4 for 3.12.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -53,8 +53,8 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-lts-3.4
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.8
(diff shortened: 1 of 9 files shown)


Turbo87:
@kellyselden why did you target `release` with this? seems like this will be a breaking change for most addons that adopt this change 🤔

kellyselden:
We've always been shifting the last 2 LTS's in the blueprints. We actually should have done this in ember-cli 3.13's stable release when 3.12 became LTS, I just forgot. As with most our blueprint changes, addon's have a choice whether they want to track these changes and adjust their major version accordingly. Do you have concerns with this policy?

Turbo87:
problem is just that if we release this in a patch version bump we have no way of communicating this change in the release blog post since we only do those for minor and major version releases

kellyselden:
That's a good point. I forgot about release blog post.

knownasilya:
Maybe do a blogpost any way 🤷‍♂

rwjblue:
Tweaking this to target beta (I don't think its great to land a breaking change patch release of 3.14).

PR #8949, opened 2019-12-01
Use GitHub Actions for CI

Turbo87:
because it's usually much faster and supports Windows too

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,109 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - beta
+      - release
+      - next
+      - 'v*'
+      - 'release-*'
+      - 'lts-*'
+  pull_request:
+  schedule:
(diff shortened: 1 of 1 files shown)


kellyselden:
Is there a chance we'll run out of free minutes on the OSS plan?

Turbo87:
@kellyselden AFAIK they changed the policy and there is no limit anymore for public projects

kellyselden:
Looks successful. Do they allow stages for our linting and unit check?

Turbo87:
> Do they allow stages for our linting and unit check?

sort of. it's modeled as a graph, so for example later stage jobs care declare what previous jobs need to have run first using `needs`.

rwjblue:
@Turbo87 - What is still pending here?

Turbo87:
@rwjblue nothing... 😅 

Windows tests are not working yet for various reasons, but we can tackle that in a dedicated PR

rwjblue:
This just adds GH actions, we can reevaluate usage of Travis and AppVeyor as we gain confidence here.

PR #8959, opened 2019-12-07
Non-scope directories without a package.json should not be considered a node_module

stefanpenner:
…ge.json should not be considered a node_module

Sometimes dependency management tooling may leave empty directories in node_modules. This is likely a bug in their implementation, but as it turns out the node resolution algorithm accounts for it, by simply not considering node_modules who do not have a package.json.

Assuming this works as expected, and folks are onboard with the change, this is likely a candidate for a `[Bugfix release]` or similar backport.

This, in theory and pending additional local testing, should address https://github.com/ember-engines/ember-engines/issues/689

- [x] review
- [x] verify fix, now that it is code-complete and not a quick hack
- [x] decide how far back to backport
- [x] fix-up windows tests.

### DIFF
--- a/lib/models/package-info-cache/index.js
+++ b/lib/models/package-info-cache/index.js
@@ -636,17 +636,24 @@ class PackageInfoCache {
     logger.info('Creating new NodeModulesList instance for %o', realPath);
     nodeModulesList = new NodeModulesList(realPath, this);
 
-    let entries = fs.readdirSync(realPath); // should not fail because getRealDirectoryPath passed
+    const entries = fs.readdirSync(realPath).filter(fileName => {
+      if (fileName.startsWith('.') || fileName.startsWith('_')) {
+        // we explicitly want to ignore these, according to the
+        // definition of a valid package name.
+        return false;
+      } else if (fileName.startsWith('@')) {
+        return true;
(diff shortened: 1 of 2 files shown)


rwjblue:
> this is likely a candidate for a [Bugfix release] or similar backport.

Yeah, I actually think we should consider making an LTS back port (since I’m pretty sure 3.12 is also affected)...

stefanpenner:
I'm busy tomorrow, will complete once i have time again.

stefanpenner:
Confirmed This fix addresses the issue I saw in ember-engines, currently fixing up 2 small windows failures.

I'll likely also open a yarn issue, as it should likely not be leaving empty directories floating around...

rwjblue:
> fix-up windows tests.

I'm not sure that they are actually broken. They take too long, and timeout when the AppVeyor boxes are under load, and pass when they are not busy (e.g. mid-night).

From the appveyor log:
> Build execution time has reached the maximum allowed time for your plan (60 minutes).

stefanpenner:
I'm passing this off to @rwjblue, as he has a specific plan for releasing/backports.

rwjblue:
Rebased to target lts-3-12 branch.

PR #8961, opened 2019-12-11
Set Octane for 3.15 release

rwjblue:
* Adds `ember` property to `package.json` to implement emberjs/rfcs#558
* Adds `@glimmer/component@1.0.0` as a development dependency for both apps and addons
* Updates `ember-try` to at least 1.4.0 in order to support `config/ember-try.js` scenarios with `ember` `package.json` property (mentioned in emberjs/rfcs#558)
* Enables Octane related optional features
* Updates ember-template-lint configuration to use `octane` preset
* Update to ember-source@3.15 stable
* Updates all packages in the application blueprint to their latest version


blueprints/app/files/.template-lintrc.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-  extends: 'recommended'
+  extends: 'octane'
```
Turbo87: is that available in template-lint v1 already?

rwjblue: Yes

Issue #4372, opened 2015-07-02
Support scoped addon names 


Fed03:
Scoped packages are not supported by ember-cli.  
When you create a new addon like this `ember addon @company/package-foo` an error comes out.

```
ENOENT, mkdir 'D:\Work\@company\package-foo'
Error: ENOENT, mkdir 'D:\Work\@company\package-foo'
    at Error (native)
```

I think the error comes from here https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/create-and-step-into-directory.js#L32

After this, I thought to create the addon in the common way and than change the package name in
- `package.json`
- `bower.json`
- `index.js`

sadly this didn't work either. This is the error after `ember serve`, and the modification of `index.js` is the culprit

```
ENOENT, no such file or directory 'D:\Work\package-foo\tmp\reexporter-tmp_cache_dir-qa23yYYW.tmp\reexports\@company\package-foo.js'
Error: ENOENT, no such file or directory 'D:\Work\package-foo\tmp\reexporter-tmp_cache_dir-qa23yYYW.tmp\reexports\@company\package-foo.js'
    at Error (native)
    at Object.fs.openSync (evalmachine.<anonymous>:502:18)
    at Object.fs.writeFileSync (evalmachine.<anonymous>:1103:15)
    at Reexporter.read (D:\Work\package-foo\node_modules\ember-cli\lib\utilities\reexport.js:30:6)
    at D:\Work\ocelot-user-addresses\node_modules\ember-cli\node_modules\broccoli\lib\builder.js:79:23
    at lib$rsvp$$internal$$tryCatch (D:\Work\package-foo\node_modules\ember-cli\node_modules\rsvp\dist\rsvp.js:489:16)
    at lib$rsvp$$internal$$invokeCallback (D:\Work\package-foo\node_modules\ember-cli\node_modules\rsvp\dist\rsvp.js:501:17)
    at D:\Work\package-foo\node_modules\ember-cli\node_modules\rsvp\dist\rsvp.js:1095:13
    at lib$rsvp$asap$$flush (D:\Work\package-foo\node_modules\ember-cli\node_modules\rsvp\dist\rsvp.js:1290:9)
    at process._tickCallback (node.js:355:11)
```

I kept the name change only in `package.json` in order to publish to the private registry, but it's a half solution. `ember install @company/package-foo` does indeed work but, for example, the default blueprint does not run. Also the multiple `import` araound the application are not consistent.

[...]

PR #8964, opened 2019-12-16
Add support for `ember new @scope-here/name-here`.

rwjblue:
Closes https://github.com/ember-cli/ember-cli/issues/4372

### DIFF
--- a/lib/commands/new.js
+++ b/lib/commands/new.js
@@ -6,7 +6,7 @@ const Command = require('../models/command');
 const RSVP = require('rsvp');
 const Project = require('../models/project');
 const SilentError = require('silent-error');
-const validProjectName = require('../utilities/valid-project-name');
+const isValidProjectName = require('../utilities/valid-project-name');
 const normalizeBlueprint = require('../utilities/normalize-blueprint-option');
 const mergeBlueprintOptions = require('../utilities/merge-blueprint-options');
 
@@ -40,12 +40,12 @@ module.exports = Command.extend({
   beforeRun: mergeBlueprintOptions,
 
   run(commandOptions, rawArgs) {
-    let packageName = rawArgs[0],
+    let projectName = rawArgs[0],
       message;
 
     commandOptions.name = rawArgs.shift();
 
-    if (!packageName) {
+    if (!projectName) {
       message = `The \`ember ${
         this.name
       }\` command requires a name to be specified. For more details, use \`ember help\`.`;
@@ -57,24 +57,20 @@ module.exports = Command.extend({
       commandOptions.skipGit = true;
     }
 
-    if (packageName === '.') {
+    if (projectName === '.') {
       let blueprintName = commandOptions.blueprint === 'app' ? 'application' : commandOptions.blueprint;
       message = `Trying to generate an ${blueprintName} structure in this directory? Use \`ember init\` instead.`;
 
       return Promise.reject(new SilentError(message));
     }
 
-    if (!validProjectName(packageName)) {
-      message = `We currently do not support a name of \`${packageName}\`.`;
+    if (!isValidProjectName(projectName)) {
+      message = `We currently do not support a name of \`${projectName}\`.`;
 
       return Promise.reject(new SilentError(message));
     }
 
     commandOptions.blueprint = normalizeBlueprint(commandOptions.blueprint);
-
-    if (!commandOptions.directory) {
-      commandOptions.directory = packageName;
-    }
 
     let InitCommand = this.commands.Init;
 
@@ -86,16 +82,20 @@ module.exports = Command.extend({
     });
 
     return this.runTask('CreateAndStepIntoDirectory', {
+      projectName,
       directoryName: commandOptions.directory,
       dryRun: commandOptions.dryRun,
     }).then(opts => {
       initCommand.project.root = process.cwd();
 
       return initCommand.run(commandOptions, rawArgs).catch(err => {
-        let dirName = commandOptions.directory;
-        process.chdir(opts.initialDirectory);
-        return rmdir(dirName).then(() => {
-          console.log(chalk.red(`Error creating new application. Removing generated directory \`./${dirName}\``));
+        let { initialDirectory, projectDirectory } = opts;
+
+        process.chdir(initialDirectory);
+        return rmdir(projectDirectory).then(() => {
+          console.log(
+            chalk.red(`Error creating new application. Removing generated directory \`./${projectDirectory}\``)
+          );
           throw err;
         });
       });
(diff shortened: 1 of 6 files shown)


lib/utilities/project-name.js
```diff
@@ -1,5 +1,21 @@
 'use strict';
 
+const InvalidProjectNames = ['test', 'ember', 'ember-cli', 'vendor', 'public', 'app', 'application'];
```
Turbo87: what about `addon`?

Turbo87:
```suggestion
const INVALID_PROJECT_NAMES = ['test', 'ember', 'ember-cli', 'vendor', 'public', 'app', 'application'];
```

rwjblue: Seems good, I'll add `addon`.

rwjblue: Updated

lib/utilities/project-name.js
```diff
@@ -1,5 +1,21 @@
...
+
+function sanitizeProjectName(name) {
+  let isScoped = name[0] === '@' && name.includes('/');
+
+  let sanitizedName = name;
+
+  if (isScoped) {
+    sanitizedName = sanitizedName
+      .slice(1) // drop the `@`
+      .replace('/', '-'); // replace the first / char with -
```
kellyselden: I would think people would want just the right side of the / for the folder name. My idea is you are already in your scope dir that has all your addons (ie "/projects/my-scope"), and you make your addons via `ember new @my-scope/my-addon-2`, you expect a project name of `@my-scope/my-addon-2` with the project structure "/projects/my-scope/my-addon-2", not "projects/my-scope/my-scope-my-addon-2". I think this makes sense in a monorepo too, where your monorepo is the scope and your packages dirs don't have the scope in their basename. I guess it all comes down to how people organise their scopes, but this is not how I'd expect it to work.

kellyselden:
For future travellers, we can probably use this https://github.com/npm/npm-package-arg/pull/38 for parsing packages with scopes once I finish it.

lib/utilities/project-name.js
rwjblue: Hmm, for all scenarios where I've created a scoped package I've used this format (https://github.com/emberjs/ember-optional-features, https://github.com/emberjs/ember-edition-utils, https://github.com/emberjs/ember-test-helpers). I agree that within a mono-repo this would be a bit different (you'd likely make `packages/@scope-name/project-name`).

In this case, we are only providing the default (and I just picked the default that I've seen and personally used before). The escape hatch here is that if the user specifies a specific `--directory` then we will use that value instead of inferring one.

What is the path forward here?

kellyselden: In addition to the above, the report from [here](https://github.com/ember-cli/ember-cli-update/issues/464) also has this layout `@html-next/flexi-config` => "/Users/rwwagner/html-next/flexi-config/"

rwjblue: @kellyselden - So if we detect that `path.basename(process.cwd()) === scopeName` then we do _not_ prefix, otherwise we do?

kellyselden: I vote for always strip off scope, and escape to --directory if you want scope in the path. If I'm scaffolding a monorepo (ie /my-scope/packages/), it would fail that cwd check.

rwjblue: Hmm, don't most mono-repos that are actually publishing scoped packages include the scoped folder name!? Some examples:

* https://github.com/glimmerjs/glimmer-vm/tree/master/packages
* https://github.com/glimmerjs/glimmer.js/tree/master/packages
* https://github.com/emberjs/ember.js/tree/master/packages
* https://github.com/vuejs/vue-cli/tree/dev/packages

Also, here are some popular monorepo packages that actually use `scope-name-package-name` in repo (what this PR would do now):

* https://github.com/babel/babel/tree/master/packages
* https://github.com/zeit/next.js/tree/canary/packages

I think the fact of the matter is that there just _isn't_ a standard way of doing this.

---

@kellyselden - FWIW, I don't particularly think that either of us are going to change our opinions on a default here, but we need to come to some sort of compromise because the current situation is massively more horrible than having to `mv scope-name-package-name wherever-i-actually-want-it`.

kellyselden: Let's go with your compromise of checking cwd, otherwise dasherize. Since you're doing the work, you get more sway :)

runspired: fwiw we don't include the scoped folder name in EmberData. Maybe we should? Seems like an unnecessary extra folder if you're already in a monorepo for the scope.

runspired: (Also I didn't realize it was common to have the extra folder, I am now less surprised I find navigating the code for so many other workspaces including Ember's so arduous, I'd always thought Ember's was that way to handle migrating from legacy)

runspired: When not using a monorepo for scoped I'd say I usually don't have the `@` in the folder path. For instance with `html-next` (the github org and npm scope mentioned above), the repo will be `html-next/foo` and the package name will be `@html-next/foo`. Unlike the comment above on having the `@scope/` typically be part of the path in `packages` for a monorepo, I feel for the non-monorepo case this `org/name` structure is far more common.

rwjblue: @runspired my suggestion in https://github.com/ember-cli/ember-cli/pull/8964#discussion_r359464747 would have required `html-next` **not** `@html-next` (the `@` is not part of the "scope name" IMHO, it's mostly only a sigil used to denote that it is a scope). If we tweak that _slightly_ to look for _any_ parent directory including the scope name, I think that would also handle the html-next and ember-data case (with the exception of the outlier `packages/-ember-data` folder) as well as @kellyselden's original point in https://github.com/ember-cli/ember-cli/pull/8964#discussion_r359430711.

rwjblue: OK, updated to implement my suggestion just above. If any directory including or above `process.cwd()` are _exactly_ the scope name, the new folder created will be the package name without scope other wise it will be `scope-name-package-name`.

lib/commands/new.js
```diff
@@ -57,25 +57,21 @@ module.exports = Command.extend({
...
+      message = `We currently do not support a name of \`${projectName}\`.`;
 
       return Promise.reject(new SilentError(message));
     }
 
     commandOptions.blueprint = normalizeBlueprint(commandOptions.blueprint);
 
-    if (!commandOptions.directory) {
-      commandOptions.directory = packageName;
-    }
```
rwjblue: I removed this defaulting so that we can ensure that we _always_ use `--directory` if specified (the defaulting here makes the logic quite a bit harder inside `CreateAndStepIntoDirectory`).

rwjblue:
WAT

lib/utilities/project-name.js
kellyselden: Sorry to suggest another change, but in one of your examples above (https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue), the scope folder has @. Do you think we could/should check for without or _with_ the @?

PR #8965, opened 2019-12-18
Update ember-resolver to v7.0.0.

rwjblue:
The primary breaking change is that it drops support for Module Unification.

### DIFF
--- a/tests/fixtures/addon/simple/package.json
+++ b/tests/fixtures/addon/simple/package.json
@@ -8,7 +8,7 @@
     "paths": ["./lib/ember-super-button"]
   },
   "devDependencies": {
-    "ember-resolver": "^5.0.1",
+    "ember-resolver": "^7.0.0",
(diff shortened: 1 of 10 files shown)


tests/fixtures/addon/simple/package.json
```diff
@@ -8,7 +8,7 @@
     "paths": ["./lib/ember-super-button"]
   },
   "devDependencies": {
-    "ember-resolver": "^5.0.1",
```
kellyselden: I take it all these fixtures with 5.0.1 are not even used, otherwise where does that number come from lol.

rwjblue: I think it just happened to be the version at the time the fixtures were authored.

kellyselden: Right but they must not be asserted against, because there was no v5.0.1 in the src/

PR #8971, opened 2019-12-19
Add an ember-try scenario for Ember "classic" (pre-octane).

rwjblue:
This provides an additional Ember try scenario to help aide addons in confirming that while onboarding new Octane features they do not accidentally regress non-octane users.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -59,6 +59,7 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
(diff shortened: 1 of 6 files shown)


Turbo87:
@kellyselden I had that thought too, but since jQuery is technically unrelated to Octane I'm not sure if we should really do that

rwjblue:
Ya, I agree with @Turbo87. I'll land as is and we can consider merging if folks have that concern as they work on upgrading.

bitwolfe:
@rwjblue This new scenario actually has an error. According to ember-try docs the `ember`node should be wrapped in an `npm` node. Tested with this change and it's working.

```
...
npm: {
  ember: 'classic'
}
...
```

rwjblue:
Yikes, sorry about that!

rwjblue:
Submitted https://github.com/ember-cli/ember-cli/pull/8977 to fix.

rwjblue:
Fixed and released in v3.15.1.

PR #8974, opened 2019-12-19
Add maybeMakeCommand method for CLI

lifeart:
@stefanpenner @Turbo87 

looks like we can't use `cli` constructor for `evironment`, because `environment` needs `project` and `project` needs `cli`.

`setEnvironment` solve this

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -84,13 +84,54 @@ class CLI {
 
   /**
    * @private
+   * @method maybeMakeCommand
+   * @param commandName
+   * @param commandArgs
+   * @return {null|CurrentCommand}
+   */
+  maybeMakeCommand(commandName, commandArgs) {
+    if (this._environment === undefined) {
+      throw new Error('Unable to make command without environment, you have to execute "run" method first.');
+    }
+    let CurrentCommand = lookupCommand(this._environment.commands, commandName, commandArgs, {
+      project: this._environment.project,
+      ui: this.ui,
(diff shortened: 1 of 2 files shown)


lib/cli/cli.js
```diff
@@ -84,13 +84,55 @@ class CLI {
 
   /**
    * @private
-   * @method run
+   * @method setEnvironment
    * @param environment
+   */
+  setEnvironment(environment) {
```
stefanpenner: If we cant pass this in via the constructor, i think we should restore the original functionality and provide via `run`, and simply have run update an internal/private `._environment` property, that `maybeMakeCommand` can use.

Also, then we should make `maybeMakeCommand` fail, if the `this._environment` is not set.

tests/unit/cli/cli-test.js
```diff
@@ -194,6 +194,23 @@ describe('Unit: CLI', function() {
...
+  it('errors correctly if no environment', function() {
+    stubValidateAndRun('help');
+
+    let cli = new CLI({
+      ui,
+      analytics,
+      testing: true,
+    });
+
+    return cli.run().catch(function(err) {
```
stefanpenner: do we need to also assert that this does not resolve?

stefanpenner:
re-triggering appveyor build

lifeart:
@rwjblue we need you! :)

lifeart:
Thank you! 🎉  😃

PR #8975, opened 2019-12-19
Fix tests writing 'undefined' as file content

BridgeAR:
Node.js might soon start to validate input data passed through to `fs`. Doing so caught to errors in ember-cli (the test suite is run against Node.js using `citgm` https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2142/nodes=osx1011/testReport/junit/(root)/citgm/ember_cli_v3_14_0/).

The tests seem to work independent of the content of these files. This is not a good idea though and it was never meant to be an optional argument.

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -177,7 +177,7 @@ describe('Acceptance: ember new', function() {
 
   it('ember new with blueprint uses the specified blueprint directory with a relative path', async function() {
     fs.mkdirsSync('my_blueprint/files');
-    fs.writeFileSync('my_blueprint/files/gitignore');
+    fs.writeFileSync('my_blueprint/files/gitignore', '');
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you @BridgeAR!

BridgeAR:
@rwjblue is it possible to merge this some time soon? CITGM will fail as soon as my PR lands in Node.js if this is not in a public release.

kellyselden:
Thanks again.

BridgeAR:
@rwjblue @kellyselden thank you for the fast response! Is there already an estimate when you might publish your next release?

kellyselden:
These are test files and shouldn't be published. I'm confused how you're testing this against a published version?

BridgeAR:
We verify releases only. We therefore execute the tests that where present during the release by using the commit hash.

kellyselden:
OK. In that case, we have a 6 week release cycle, so somewhere between 6 - 12 weeks a change to master will get released to stable.

You could resubmit as a fix on the stable branch and it will get out sooner if you really need it.

PR #8978, opened 2019-12-20
GitHub Actions CI Tweaks

rwjblue:
* Add specific job for `linting` (this makes it a bit easier to quickly
  spot linting issues vs actual test failures)
* Add `--frozen-lockfile` and `--non-interactive` to CI
* Add Node 13.x testing
* Split slow and fast tests into separate CI jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      - run: yarn install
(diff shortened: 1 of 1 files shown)


rwjblue:
@Turbo87 - What do you think RE: running the `test` and `test:slow` jobs separately? It doesn't _seem_ to dramatically effect the overall time, so I'm vaguely inclined to remove that change...

Turbo87:
if there is no time advantage to it then I would say keep it in the same job for simplicity

PR #9012, opened 2020-01-14
Drop support for Node v8

jrjohnson:
In line with the Node support policy this drops support for Node v8
which reached end of life on December 31st 2019.  Also updated
blueprints to match.


jrjohnson:
I'm assuming that the appveyor test timeout isn't related to this PR, but if I'm wrong about that let me know.

docs/node-support.md
```diff
@@ -10,7 +10,7 @@
 | 5.x              | 1.13.9 - 2.6.3               |
 | 6.x              | 2.9.0 - 3.9.x                |
 | 7.x              | 2.10.0 - 2.16.x              |
-| 8.x              | 2.13.3 - Current             |
+| 8.x              | 2.13.3 - 3.15.x             |
```
kellyselden: I believe this will be 3.16, since beta will release with node 8 still turned on.

jrjohnson: I wasn't sure about that and meant to ask, thanks for noticing! I switched it to 3.16.x.

rwjblue:
Thank you @jrjohnson!

rwjblue:
🤦‍♂ wrong button, sorry!

PR #9013, opened 2020-01-15
Remove useless line break in `.editorconfig` file

dcyriller:


rwjblue:
Only Node 8 was failing here, and Node 8 has been removed from CI.

Landing...

PR #9014, opened 2020-01-15
Avoid errors when `ui` is not present and a warning will be emitted.

rwjblue:
Addon packageInfos do not have a project. All packageInfos have a cache.ui.

(cherry picked from commit 1652f93ed738168c9a0ce0b25e36063d5f99043c)


stefanpenner:
Failures so far appear to be semver drift in depends on the now EOL'd node 8. That seems unfortunate, but acceptable.

PR #9015, opened 2020-01-15
Allow failure of Node 8 CI jobs.

rwjblue:
Node 8 is EOL and dependencies are releasing versions that require Node 10+ without bumping major versions.

This allows Node 8 failures, so that we can continue to have a passing CI for the other supported platforms (so that as we land bugfixes and whatnot, we can be certain they work properly).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,9 @@ jobs:
     strategy:
       matrix:
         node-version: [8.x, 10.x, 12.x]
+        include:
+          - node-version: 8.x
+            continue-on-error: true
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


rwjblue:
See https://github.com/ember-cli/ember-cli-htmlbars/pull/437 for a writeup of one example of a dependency releasing a breaking change within an existing major version.

rwjblue:
> Do you think instead of allow failures, it might make sense to just yarn install with --ignore-engines on?

The issue is that that is much harder :sob:. Specifically, it isn't ember-cli's own `yarn install` that is failing, it is specific smoke tests that themselves run `ember new` with yarn. So allowing failure is much easier (because otherwise I'd have to modify many individual tests), and in addition all I'm trying to do here is make sure that we can _at least_ test the other platforms that do continue to work (for #9014).

kellyselden:
@rwjblue Ah that makes total sense. Thanks for clarifying.

PR #9023, opened 2020-02-10
Update to use Ember + Ember Data 3.16.

rwjblue:


rwjblue:
FYI - CI is failing for the module unification test runs because `@ember-data/private-build-infra` has a `./src` folder but no template preprocessors (e.g. it does not depend on `ember-cli-htmlbars`).

PR #9026, opened 2020-02-11
Add @glimmer/tracking to default blueprint.

### LINKED ISSUE #9003: Tracked properties module not included in new ember projects
---

```
ember-cli: 3.15.1
http_parser: 2.8.0
node: 10.13.0
v8: 6.8.275.32-node.36
uv: 1.23.2
zlib: 1.2.11
ares: 1.14.0
modules: 64
nghttp2: 1.34.0
napi: 3
openssl: 1.1.0i
icu: 62.1
unicode: 11.0
cldr: 33.1
tz: 2018e
os: darwin x64
6.4.1
1.12.3
```

The [guides](https://guides.emberjs.com/v3.15.0/upgrading/current-edition/tracked-properties/) imply that tracked properties are included out of the box in a new ember-cli project. 

However, generating a new ember-cli project on 3.15 does not include the module @glimmer/tracking -- leading to confusion for users.

Should a new ember-cli project include @glimmer/tracking?

rwjblue:
This package isn't _strictly_ required (because we polyfill replace it with a Ember.* path at runtime), but having it is much easier to explain (aides consumers in TypeScript land, matches the similar @glimmer/component dependency, etc).

Fixes #9003

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -20,6 +20,7 @@
   "devDependencies": {
     "@ember/optional-features": "^1.1.0",
     "@glimmer/component": "^1.0.0",
+    "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.0.3",
     "broccoli-asset-rev": "^3.0.0",
     "ember-auto-import": "^1.5.3",
--- a/tests/fixtures/addon/npm/package.json
+++ b/tests/fixtures/addon/npm/package.json
@@ -27,6 +27,7 @@
   "devDependencies": {
     "@ember/optional-features": "^1.1.0",
     "@glimmer/component": "^1.0.0",
+    "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.0.3",
     "broccoli-asset-rev": "^3.0.0",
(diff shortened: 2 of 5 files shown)


kellyselden:
Is the pinned version used (instead of the bundled version) if it is included like this?

rwjblue:
> Is the pinned version used (instead of the bundled version) if it is included like this?

No, the Ember bundled version is always used at the moment, but as we move Ember itself more towards "real" packages (vs our fake polyfilled ones) it will have better support for these kinds of things.

Also, `@glimmer/tracking` is not an actual addon ([see here](https://github.com/glimmerjs/glimmer.js/blob/master/packages/%40glimmer/tracking/package.json)) and _basically_ has no effect on the build.

kellyselden:
Hmm I'm not sure if this helps confusion or hurts confusion. I guess I'll defer to you on this one.

rwjblue:
@kellyselden - Ya, the thing for now is that we are already telling folks that they need to add any `@glimmer/*` packages as deps/devDeps (e.g. `@glimmer/component`). So following the rule ourselves is a bit better IMHO.

Issue #9029, opened 2020-02-11
empty "server" folder in project root broke build


lifeart:
I'm doing "step-by-step" app and rolling back commits from created server middleware I fased such issue (folder exists, because git don't remove it, but there is no files)
Produced error look unclear for cummon user and I think it make sence to improve it or silentrly handle empty "server" folder.

ember-cli 3.15.2
env: windows

octane app:
```
server <- folder
```

```
$ ember s
Cannot find module './server' from 'C:\Users\lifeart\Documents\repos\longestjs\pizza'
Stack Trace and Error Report: C:\Users\lifeart\AppData\Local\Temp/error.dump.5cf25c0b55910baad8d2f05e79b6c3b9.log
```

```
ERROR Summary:

  - broccoliBuilderErrorStack: [undefined]
  - code: MODULE_NOT_FOUND
  - codeFrame: [undefined]
  - errorMessage: Cannot find module './server' from 'C:\Users\lifeart\Documents\repos\longestjs\pizza'
  - errorType: [undefined]
  - location:
    - column: [undefined]
    - file: [undefined]
    - line: [undefined]
  - message: Cannot find module './server' from 'C:\Users\lifeart\Documents\repos\longestjs\pizza'
  - name: Error
  - nodeAnnotation: [undefined]
  - nodeName: [undefined]
  - originalErrorMessage: [undefined]
  - stack: Error: Cannot find module './server' from 'C:\Users\lifeart\Documents\repos\longestjs\pizza'
    at Function.resolveSync [as sync] (C:\Users\lifeart\Documents\repos\longestjs\pizza\node_modules\resolve\lib\sync.js:81:15)
    at Project.resolveSync (C:\Users\lifeart\Documents\repos\longestjs\pizza\node_modules\ember-cli\lib\models\project.js:354:20)
    at Project.require (C:\Users\lifeart\Documents\repos\longestjs\pizza\node_modules\ember-cli\lib\models\project.js:370:21)
    at ExpressServerTask.processAppMiddlewares (C:\Users\lifeart\Documents\repos\longestjs\pizza\node_modules\ember-cli\lib\tasks\server\express-server.js:120:33)
    at Promise.resolve.then.then (C:\Users\lifeart\Documents\repos\longestjs\pizza\node_modules\ember-cli\lib\tasks\server\express-server.js:255:24)
[...]

PR #9034, opened 2020-02-12
fix: check middleware path + index, instead of only path

lifeart:
issue: https://github.com/ember-cli/ember-cli/issues/9029


lib/tasks/server/express-server.js
```diff
@@ -116,7 +116,7 @@ class ExpressServerTask extends Task {
   }
 
   processAppMiddlewares(options) {
-    if (this.project.has(this.serverRoot)) {
+    if (this.project.has([this.serverRoot, 'index'].join(path.sep))) {
```
lifeart: has logic: https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L342

rwjblue: This would break folks using `server.js` (which is also supported).

lib/tasks/server/express-server.js
```diff
@@ -116,7 +116,7 @@ class ExpressServerTask extends Task {
   }
 
   processAppMiddlewares(options) {
-    if (this.project.has(this.serverRoot)) {
+    if (this.project.has(`${this.serverRoot}.js`) || this.project.has([this.serverRoot, 'index'].join(path.sep))) {
```
rwjblue: Now this breaks folks using `server/index.ts` along with ts-node.

Would you mind replacing this `if` completely, with something like:

```js
processAppMiddlewares(options) {
  let server;

  try {
    server = this.project.require(this.serverRoot);
  } catch(error) {
	if (error.code === 'MODULE_NOT_FOUND') {
      // do nothing, `server.js`, `server/index.js`, etc were not found
      return;
    }
    throw error;
  }

   // existing logic
}
```

lifeart: it will never support `ts` due to https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L323 implementation

rwjblue: > it will never support ts due to https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L323 implementation

Incorrect 😄 

I do agree that that implementation should be fixed (to use `resolve.sync` or something), but the example I gave was using `server/index.ts` and if you look at the existing logic we look for `path.join(this.root, 'server')` and it would exist so we'd never hit the fallback that looks for `path.join(this.root, 'server.js')`.

lifeart:
going to close it and reopen upsteamed, with tests

PR #9041, opened 2020-02-13
Remove usage of RSVP.

rwjblue:
* Rewrite all usages of RSVP to use native promise
* Use [promise.hash.helper](https://github.com/stefanpenner/promise.hash.helper)   as a `RSVP.hash` replacement
* Remove `promise.prototype.finally` (in favor of native promise `.finally`)

Closes https://github.com/ember-cli/ember-cli/pull/8567


lib/utilities/hash.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = async function hash(obj) {
```
stefanpenner: Tests? Should we extract? If so, I can gladly do so today.

rwjblue: @stefanpenner - Ya, I can write some tests for the util. I was wondering if we should think about a larger refactor in RSVP itself to make RSVP use the native promise if possible...

stefanpenner:
@rwjblue https://github.com/stefanpenner/promise.hash.helper/

lib/utilities/hash.js
rwjblue: @stefanpenner extracted into [promise.hash.helper](https://github.com/stefanpenner/promise.hash.helper)

PR #9042, opened 2020-02-13
Include API documentation `yuidoc` JSON output when publishing

sivakumar-kailasam:
The learning team would like to start publishing the latest version of the ember-cli api doc as a part of https://api.emberjs.com. To do so, we'll need the yuidoc generated `data.json` file to be published as a part of the npm package. 

This PR adds a new entry in the `.npmignore` file that enables this.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -9,3 +9,4 @@
 *.log
 !/.travis.yml
 /yarn.lock
+!docs/build/data.json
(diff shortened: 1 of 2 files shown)


rwjblue:
@sivakumar-kailasam - Won't we also need to add something to regenerate it in `prepare` script?

sivakumar-kailasam:
I didn't realise that the docs generation was a separate step on travis. `yarn pack` didn't  invoke the `prepare` script so I had to use `prepack` instead.

sivakumar-kailasam:
@rwjblue the CI failures seem unrelated.

PR #9062, opened 2020-02-24
include "yarn": true in .ember-cli if using yarn

kellyselden:
Normally this isn't a problem, because you have a yarn.lock. But in certain cases (automation/testing), you might not have the lockfile (you're using fixtures you generated with `--skip-npm --yarn`), and stuff like `addPackagesToProject` will use npm instead of yarn.

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -395,7 +395,7 @@ describe('Acceptance: ember new', function() {
       let namespace = 'app';
       let fixturePath = `${namespace}/npm`;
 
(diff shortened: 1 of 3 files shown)


kellyselden:
Correct

kellyselden:
This still doesn't fix it:

```
ember-cli:npm-task yarn.lock not found -> using npm +275ms
```

rwjblue:
Where are we with this? Do you intend to proceed?

kellyselden:
I'll pick it up later.

PR #9068, opened 2020-02-27
[minor] Adds link to CLI commands doc from README

entendu:
This adds a direct link to the basic CLI commands doc.

Rationale:

1. The readme says, "You can call `ember <command> --help` to find out more about all of the following commands..." but no command list follows.

2. From the project page/readme, it takes 3 clicks to find what the actual `ember-cli` commands are, one of which is hidden inside an accordion.


rwjblue:
Thank you!

Issue #9007, opened 2020-01-05
 ember-cli >= 3.15 or EMBER_CLI_BROCCOLI_WATCHER=true prior to 3.15 usage does not correctly print out errors


kellyselden:
In 3.14, an app.import of a missing file was an immediate exit with an expected stack trace. In 3.15, it is now a hung build with a UnhandledPromiseRejectionWarning.

3.14 https://travis-ci.org/kellyselden/ember-cli-test-server/jobs/633006370#L624
3.15 https://travis-ci.org/kellyselden/ember-cli-test-server/jobs/633037017#L560

My test suite is intentionally causing this error, that's the only way I discovered this. Here is the line
https://github.com/kellyselden/ember-cli-test-server/blob/764e703573/test/index-test.js#L112
https://github.com/kellyselden/ember-cli-test-server/blob/764e703573/test/index-test.js#L61

rwjblue:
Hmm, this is definitely odd. Not sure what we changed internally here to make this be an unhandled rejection now. :thinking:

stefanpenner:
3.15 first enabled BROCCOLI_WATCHER by default, I believe that experiment fails to correctly print errors.

stefanpenner:
Potentially introduced by https://github.com/ember-cli/ember-cli/pull/8666 and but enabled by default by https://github.com/ember-cli/ember-cli/pull/8906

stefanpenner:
I'll dig into this, most likely in the AM. I'm about to head to the gym...

stefanpenner:
some prep work, so we can fail the test on unhandled rejections: https://github.com/ember-cli/ember-cli/pull/9070

kellyselden:
@stefanpenner Thanks a lot for digging!

stefanpenner:
https://github.com/ember-cli/ember-cli/pull/9070
https://github.com/ember-cli/ember-cli/pull/9072

PR #9070, opened 2020-02-28
fail tests on unhandled rejections

stefanpenner:
prep work to address https://github.com/ember-cli/ember-cli/issues/9007

- [x] fix all failures

### DIFF
--- a/lib/utilities/open-editor.js
+++ b/lib/utilities/open-editor.js
@@ -13,14 +13,17 @@ function openEditor(file) {
 
   let editorArgs = openEditor._env().EDITOR.split(' ');
   let editor = editorArgs.shift();
-  let editProcess = openEditor._spawn(editor, [file].concat(editorArgs), { stdio: 'inherit' });
+  const args = [file].concat(editorArgs);
+  let editProcess = openEditor._spawn(editor, args, { stdio: 'inherit' });
 
   return new Promise((resolve, reject) => {
     editProcess.on('close', code => {
       if (code === 0) {
         resolve();
       } else {
-        reject();
+        reject(
+          new Error(`Spawn('${editor}', [${args.join(',')}]) exited with a non zero error status code: '${code}'`)
+        );
       }
     });
   });
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -1,7 +1,4 @@
 'use strict';
-// process.on('unhandledRejection', e => {
-//   throw e;
-// });
 const captureExit = require('capture-exit');
 captureExit.captureExit();
 
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -1,5 +1,7 @@
 'use strict';
-
+// process.on('unhandledRejection', e => {
+//   throw e;
+// });
 const captureExit = require('capture-exit');
 captureExit.captureExit();
 
--- a/tests/unit/docs-lint-test.js
+++ b/tests/unit/docs-lint-test.js
@@ -1,45 +1,61 @@
 'use strict';
 
+/*
+ * Turns out, YUIDoc attached a process.on('unhandledRejection' handler which
+ * incorrectly reports any unhandledRejection as a YUIDoc error. This makes
+ * debugging of unhandledRejections in the ember-cli test suite quite painful
+ *
+ * To address YUIDocs behavior, we run YUIDoc isolated in it's own process;
+ */
+if (!process.env['IS_CHILD_THREAD']) {
+  const execa = require('execa');
+
+  describe('YUIDoc', function() {
+    it('parses without warnings', async function() {
+      await execa('node', [`--unhandled-rejections=strict`, __filename], {
+        env: {
+          IS_CHILD_THREAD: true,
+        },
+      });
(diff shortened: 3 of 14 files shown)


tests/unit/docs-lint-test.js
```diff
@@ -1,45 +1,45 @@
...
-    });
-
-    if (message.length) {
-      let error = new Error(message);
-      delete error.stack;
-      throw error;
-    }
-  });
-});
+// 'use strict';
```
stefanpenner: this test file requires `yuidocjs` which makes more or less screws up the output of unhandled rejections, making debugging more annoying by branding all unhandled rejections as errors specific to yuidoc.... So i've disabled it for now, we should explore either removing these tests, or execute these tests in their own subprocess.

tests/runner.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
-
+// process.on('unhandledRejection', e => {
```
stefanpenner: can likely remove, assuming older node versions we support handle ` --unhandled-rejections=strict`

tests/unit/tasks/server/middleware/tests-server-test.js
```diff
@@ -42,7 +42,7 @@ describe('TestServerAddon', function() {
       });
     });
 
-    it('invokes next when the watcher fails', function(done) {
+    it.skip('invokes next when the watcher fails', function(done) {
```
stefanpenner: fix and unskip test

stefanpenner:
fix

```
not ok 1233 open-editor spawns EDITOR with passed file
  Cannot read property 'on' of undefined
  TypeError: Cannot read property 'on' of undefined
      at /home/runner/work/ember-cli/ember-cli/lib/utilities/open-editor.js:19:17
      at new Promise (<anonymous>)
      at openEditor (lib/utilities/open-editor.js:18:10)
      at Context.<anonymous> (tests/unit/utilities/open-editor-test.js:27:5)
```

lib/utilities/open-editor.js
```diff
@@ -13,14 +13,17 @@ function openEditor(file) {
...
+  let editProcess = openEditor._spawn(editor, args, { stdio: 'inherit' });
 
   return new Promise((resolve, reject) => {
     editProcess.on('close', code => {
       if (code === 0) {
         resolve();
       } else {
-        reject();
+        reject(
+          new Error(`Spawn('${editor}', [${args.join(',')}]) exited with a non zero error status code: '${code}'`)
```
mike-north:
```suggestion
          new Error(`Spawn('${editor}', [${args.join(', ')}]) exited with a non zero error status code: '${code}'`)
```

tests/unit/docs-lint-test.js
```diff
@@ -1,45 +1,61 @@
...
+let json = yuiDoc.run();
 
-      if (!warnings[file]) {
-        warnings[file] = [];
-      }
+let warnings = {};
+json.warnings.forEach(function(warning) {
+  let tmp = warning.line.split(':');
+  let file = tmp[0].trim();
+  let line = tmp[1];
```
mike-north: Seems like a good place to destructure, and I'm looking for a chance to make a multi-line suggestion

```suggestion
  let [file, line] = warning.line.split(':').map(item => item.trim());
```

tests/unit/tasks/server/livereload-server-test.js
```diff
@@ -60,24 +60,22 @@ describe('livereload-server', function() {
...
-            `Livereload failed on http://localhost:1337.  It is either in use or you do not have permission.`
-          );
-        })
-        .finally(function() {
-          preexistingServer.close(done);
         });
+      } catch (e) {
+        expect(e.message).to.include(`Livereload failed on 'http://localhost:1337', It may be in use.`);
+      } finally {
+        await new Promise(resolve => preexistingServer.close(resolve));
```
mike-north: I'm curious about whether there's there a reason you're not using `Promise.resolve` here?
```suggestion
        await Promise.resolve(preexistingServer.close(resolve));
```

tests/unit/docs-lint-test.js
stefanpenner: Ya, that's a fine suggestion. But in general I try to avoid cosmetic cleanup in bug-fixes.

tests/unit/tasks/server/livereload-server-test.js
stefanpenner: We need a promise that fulfills only when the callback passed to `prexistingServer.close` is invoked. 

Unfortunately the return value of `preexistingServer.close` is not meaningful when it comes to the above mentioned goal as `preexistingServer.close(cb) === preexistingServer` which when awaited fullfills on the next tick, regardless if the server has closed or not.


Two other concerns with the provided recommendation:

* `await Promise.resolve(value)` can be safely reduced to `await value` assuming `Promise` is the platform provided promise, which in this case it should be.
* the `resolve` argument passed to `close` is not defined, and will trigger a thrown `ReferenceError`.

PR #9079, opened 2020-03-08
Parse PORT environment variable as a number

mnutt:
`ember serve -p XXXX` works just fine, but `PORT=XXXX ember serve` does not work for any value of `XXXX`, always resulting in `Port XXXX is already in use.`, even if the port is not in use. I believe it's due to:

```javascript
commandOptions.port !== foundPort
```
(https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L111)

`process.env.PORT` comes in as an environment variable string, and as `foundPort` is a number, the comparison fails.

The workaround (and probably recommended path) is to use `-p XXXX` instead, but since the env-based functionality is here I thought it would be good to fix it. I'm happy to add an acceptance test but the change was so trivial I wasn't sure if it was worth the overhead.


ro0gr:
> I'm happy to add an acceptance test but the change was so trivial I wasn't sure if it was worth the overhead.


There is an unit test which seems suitable  https://github.com/ember-cli/ember-cli/blob/master/tests/unit/commands/serve-test.js

mnutt:
It's actually a little tricky to test since `defaultPort` is set at module load time for `commands/serve`. Would it make more sense to add an acceptance test which spins up a new process with the environment variable set, or change the `defaultPort` variable declaration to make it more testable?

PR #9080, opened 2020-03-08
Run basic tests against Windows, Linux, and macOS in GH Actions

rwjblue:
Adds Windows and macOS runs to the GitHub Actions setup. At the moment, it only runs them for "basic tests" (not the longer running smoke tests), but we can work towards fixing the smoke tests on macOS and Windows in follow ups.


rwjblue:
Next up: smoke tests!

PR #9086, opened 2020-03-13
Add Windows to GH Actions CI for test:all.

rwjblue:
* Adds `windows-latest` and `macOS-latest` to run for each supported Node version.
* Tweaks the job names to be a bit simpler / easier to read.
* Fix issue on Windows when killing an already killed child process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,8 @@ jobs:
     strategy:
       matrix:
         feature-flag:
-          - EMBER_CLI_ENABLE_ALL_EXPERIMENTS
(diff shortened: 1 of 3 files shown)


rwjblue:
I expect this to largely pass (I [tested it on just Node 13 in a branch](https://github.com/ember-cli/ember-cli/runs/503476128?check_suite_focus=true) before PRing), though there may be specific failures that need more triage/debugging.

rwjblue:
LOL, passes on Windows, fails on macOS 😝

stefanpenner:
🕺

PR #9093, opened 2020-03-14
Migrate supported Node version detection to use GitHub Actions config.

rwjblue:
The `PlatformChecker` reads the Node versions that are being tested in CI to decide if we should print a message to the console upon booting.

When running on a supported (based on our `package.json`'s `engines.node`) but untested (e.g. not tested in CI) the following message is printed:

> Node ${process.version} is not tested against Ember CLI on your platform. We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.

However, when running on an unsupported version of Node (e.g. a version not covered by our `package.json`s `engines.node`) the following message is printed:

> Node ${process.version} is no longer supported by Ember CLI.  We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.

---

This change updates the `PlatformChecker` to use the GitHub Actions configuration (`.github/workflows/ci.yml`) instead of using `.travis.yml` and `appveyor.yml`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,6 +7,12 @@
 /tmp/
 /.*
 *.log
-!/.travis.yml
 /yarn.lock
 !docs/build/data.json
+
+# the GitHub Action CI configuration is used by
+# lib/utilities/platform-checker.js to determine what versions of Node are
+# under test by the current version of ember-cli being used, this means we
+# **must** publish this file (but it would be ignored by the `/.*` line just
+# above)
+!/.github/workflows/ci.yml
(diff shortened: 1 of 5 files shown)


.npmignore
```diff
@@ -7,6 +7,6 @@
 /tmp/
 /.*
 *.log
-!/.travis.yml
+!/.github/workflows/ci.yml
```
kellyselden: Maybe a comment for future travellers why this is in here?

rwjblue: Great idea! Updated!

PR #9101, opened 2020-03-15
Migrate code coverage to GitHub Actions.

rwjblue:


rwjblue:
FYI - I've removed this PR status checking over on the code climate side:

> codeclimate — 1 issue to fix

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,32 @@
+name: Code Coverage
```
rwjblue: This new CI job only runs for `master` branch, but while testing I enabled running on my test branch. You can see that run here (to check/confirm that it works properly):

https://github.com/ember-cli/ember-cli/runs/509294687?check_suite_focus=true

PR #9108, opened 2020-03-16
Cleanup of async in `CLI` / `Builder` while digging into issues around progress clean up.

rwjblue:
See individual commits for details.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -128,10 +128,11 @@ class CLI {
     if (environment === undefined) {
       return Promise.reject(new Error('Unable to execute "run" command without environment argument'));
     }
-    this._environment = environment;
     let shutdownOnExit = null;
     return hash(environment)
       .then(environment => {
+        this._environment = environment;
+
         let args = environment.cliArgs.slice();
         let commandName = args.shift();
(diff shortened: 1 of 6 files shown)


lib/cli/cli.js
```diff
@@ -118,139 +118,142 @@ class CLI {
...
+      }
 
-            return command.validateAndRun(commandArgs);
-          })
-          .then(result => {
-            instrumentation.stopAndReport('command', commandName, commandArgs);
+      let instrumentation = this.instrumentation;
+      let onCommandInterrupt;
+
+      let runPromise = Promise.resolve()
```
rwjblue: This is left out of the async/await refactor specifically because it is intertwined with the `onCommandInterrupt` function in a way that is not trivially possible with `async` / `await`.

stefanpenner: Let's created a `async () => {}` here, so we can coordinate on `onCommandInterrupt` and `runPromise` as currently implemented, but using cleaner more maintainable syntax

rwjblue: Ya, great suggestion! Will update.

rwjblue: Updated with that...

PR #9134, opened 2020-03-31
[DOC] Update locals hook example

locks:
Closes https://github.com/ember-cli/ember-cli/pull/8210.

Could not update the original PR, so cherry picked into a new branch.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -526,7 +526,7 @@ let Blueprint = CoreObject.extend({
     When the following is called on the command line:
 
     ```sh
-    ember generate controller foo --type=array --dry-run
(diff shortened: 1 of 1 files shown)


rwjblue:
Ugh, looks like CI is failing because of the way our version generation works. This branch has a `/` in its name, and that results in an invalid semver version (with the way we generate local versions to include the branch name).

Specifically, we calculate: `3.18.0-beta.1-docs/locals-hook-1ad27f26d8`, and I think we should probably replace that `/` with a `-` or something...

rwjblue:
I think that failure should be fixed if you rebase against master (fixed by #9139).

locks:
woohoo!

PR #9139, opened 2020-04-01
make slashes into dashes for version string

locks:


lib/utilities/version-utils.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     if (fs.existsSync(gitPath)) {
       let repoInfo = getRepoInfo(gitPath);
 
-      output.push(repoInfo.branch);
+      output.push(repoInfo.branch.replace(/\//g, '-'));
```
rwjblue, line 15: `repoInfo.branch` is sometimes not defined, so this `repoInfo.branch.replace` fails

```
error: TypeError: Cannot read property 'replace' of null
      at emberCLIVersion (/home/runner/work/ember-cli/ember-cli/lib/utilities/version-utils.js:15:35)
```

PR #9146, opened 2020-04-04
Convert express-server task to async await

locks:


lib/tasks/server/express-server.js
```diff
@@ -153,29 +153,30 @@ class ExpressServerTask extends Task {
...
-          this.ui.writeLine('');
-        })
-        .catch((err) => {
-          this.ui.writeError(err);
-        })
-        .finally(() => {
-          this.serverRestartPromise = null;
-        });
+      try {
+        this.serverRestartPromise = await this.stopHttpServer()
```
Turbo87, line 159: I'm not sure about the `await` here. `serverRestartPromise` will no longer contain the promise in this case, but instead the resolution value of the promise 🤔

locks: Good point! I think with the new commit the correct logic is preserved.

lib/tasks/server/express-server.js
```diff
@@ -153,29 +153,30 @@ class ExpressServerTask extends Task {
     this.scheduleServerRestart();
   }
 
-  restartHttpServer() {
+  async restartHttpServer() {
     if (!this.serverRestartPromise) {
-      this.serverRestartPromise = this.stopHttpServer()
-        .then(() => {
+      try {
```
rwjblue, line 158: This seems wrong to me, the `try`/`catch`/`finally` should be around the async just below now the creation of the `this.serverRestartPromise` chain.

You need something like this:

```js
this.serverRestartPromise = (async () => {
  try {
    await this.stopHttpServer();

    this.emit('restart');
    this.ui.writeLine('');
    this.ui.writeLine(chalk.green('Server restarted.'));
    this.ui.writeLine('');
  } catch(err) {
    this.ui.writeError(err);
  } finally {
    this.serverRestartPromise = null;
  }
})();
```

locks: Updated!

PR #9149, opened 2020-04-04
Update blueprint dependencies to latest

bmish:
The biggest change is updating to [eslint-plugin-ember v8](https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v8.0.0) which includes additional recommended rules like [ember/no-jquery](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-jquery.md), so we no longer need to manually enable that rule.


rwjblue:
Thanks!

PR #9158, opened 2020-04-09
Convert clean-remove to async/await syntax

locks:


lib/utilities/clean-remove.js
```diff
@@ -4,45 +4,41 @@ const path = require('path');
...
-        (chainedPromise, thePath) =>
-          chainedPromise.then((wasShortCircuited) => {
-            if (wasShortCircuited) {
-              // optimization that says since my child dir wasn't empty,
-              // I can't be empty, so keep skipping
-              return true;
-            }
+    return paths.reduce(
+      (chainedPromise, thePath) =>
+        chainedPromise.then(async (wasShortCircuited) => {
```
rwjblue, line 15: Can we kill this reduce?

AFAICT this whole thing can be boiled down to this (which I think is much simpler):

```js
for (let thePath of paths) {
  try {
    let childPaths = await fs.readdir(thePath);
    if (childPaths.length > 0) {
      return;
    }
  
    await fs.remove(thePath);
  } catch(err) {
    if (err.code === 'ENOENT') {
      // do nothing, already removed
    }

    throw err;
  }
}
```

rwjblue:
Also can you migrate tests/integration/utilities/clean-remove-test.js?

locks:
Tests converted!

lib/utilities/clean-remove.js
```diff
@@ -4,45 +4,36 @@ const path = require('path');
...
-            return fs.readdir(thePath).then((paths) => {
-              if (paths.length) {
-                // don't check parent dirs since this one isn't empty
-                return true;
-              }
+        await fs.remove(thePath);
+      } catch (err) {
+        if (err.code === 'ENOENT') {
+          // do nothing, already removed
+        }
```
rwjblue, line 24: I'm not sure that both this `catch` and the outer one are needed.

Can you remove this try/catch (and we'll just rely on the outer one)?

locks: Done :)

PR #9163, opened 2020-04-12
Convert in-option-generate-test to async/await syntax

locks:


tests/acceptance/in-option-generate-test.js
```diff
@@ -46,86 +46,57 @@ describe('Acceptance: ember generate with --in option', function () {
...
-            expect(file('lib/other-thing/addon/foos/bar.js')).to.exist;
-          })
-        )
-    );
+    await initApp();
+    await generateUtils.inRepoAddon('lib/other-thing');
+    await generateUtils.tempBlueprint();
+    await ember(['generate', 'foo', 'bar', '--in=lib/other-thing']);
+
+    return expect(file('lib/other-thing/addon/foos/bar.js')).to.exist;
```
Turbo87, line 56: no need for the return here

locks: Updated!

PR #9174, opened 2020-04-16
Enable embroider as an opt-in experiment for new apps

thoov:
This PR adds the "EMBROIDER" experiment such that new apps can be created with embroider enabled and defaulted. This experiment enables:

```
EMBER_CLI_EMBROIDER=true ember new foobar
```

Which will properly setup the app as specified from [embroider's readme](https://github.com/embroider-build/embroider#how-to-try-it)

cc: @ef4 @rwjblue @stefanpenner


thoov:
@rwjblue Adding the experiment to the CI workflow causes some of the tests to fail and I have skipped them. What are your suggestions about these?

tests/acceptance/new-test.js
```diff
@@ -374,6 +375,27 @@ describe('Acceptance: ember new', function () {
     expect(pkgJson.name).to.equal('@foo/bar', 'uses addon name for package name');
   });
 
+  it('embroider experiment creates the correct files', async function () {;
```
locks, line 378:
```suggestion
  it('embroider experiment creates the correct files', async function () {
```

thoov: Done, thanks for the catch

blueprints/app/files/ember-cli-build.js
```diff
@@ -20,5 +20,10 @@ module.exports = function(defaults) {
   // please specify an object with the list of modules as keys
   // along with the exports of each module as its value.
 
+  <% if (embroider) { %>
```
kellyselden, line 23: This should have a fixture test now, as I believe you have some unintended whitespace now.

Turbo87:
can I ask **why** we're doing this? adding embroider manually doesn't seems like a lot of work but adding and supporting is here will lead to increased maintenance work 🤔

thoov:
@Turbo87 I agree that the actual diff to enable embroider is small and the recommended way for existing apps to try it out would be to follow the readme steps. The why for this is to prepare for how embroider apps will be created (ie with `ember new`) and to help generally push it forward. 

I'm happy to chat about how we can lower any added costs. Do you have some particular maintenance costs you are worried about?

Turbo87:
> I'm happy to chat about how we can lower any added costs. Do you have some particular maintenance costs you are worried about?

we'd have to keep our blueprint up-to-date with every new embroider release and test it properly. it also adds significant cost to our CI times for running another scenario.

> and to help generally push it forward.

I agree that we should push it forward, but adding it to new apps will not help much IMHO. the problem with embroider is that currently existing apps are often not compatible yet and that won't change by adding it as a flag to the blueprint for new apps.

thoov:
@Turbo87:
 
> we'd have to keep our blueprint up-to-date with every new embroider release and test 
> it properly.

This should just be the occasional version bump. I don't see anything additional over how we manage the other dependencies. As for testing it properly, the majority of that should remain within the embroider project.

> it also adds significant cost to our CI times for running another scenario.

I'm fine not including it as a scenario if we feel the time cost is not worth it. I don't think there is much value that it adds for this case but it was a suggestion from @rwjblue which is why I added it.

> I agree that we should push it forward, but adding it to new apps will not help much 
> IMHO. the problem with embroider is that currently existing apps are often not 
> compatible yet and that won't change by adding it as a flag to the blueprint for new 
> apps.

While I agree that its current state is definitely still beta (and I wouldn't suggest this become the default until it exits beta) this seems like a low hanging fruit to help get people to try and experiment with embroider especially in a situation were embroider is working well (new simple default apps). If now is not the correct time to introduce this change the when do feel that time would be?

blueprints/app/files/ember-cli-build.js
thoov: Updated with a fixture test for this

ef4:
An alternative to this implementation that might address maintainability concerns would be:

 - adding the embroider packages directly as dependencies of ember-cli
 - adding a feature flag so that ember-cli's `app.toTree()` would invoke embroider.

This would make embroider an implementation detail of ember-cli, which is where we want to head anyway.

The main thing that would need figuring out is where to put the embroider-specific options, once the user is no longer configuring it all manually. Probably that can be arguments directly passed to the app, under an `@embroider/core` property.

rwjblue:
> An alternative to this implementation that might address maintainability concerns would be:
> 
> * adding the embroider packages directly as dependencies of ember-cli
> * adding a feature flag so that ember-cli's `app.toTree()` would invoke embroider.
> 
> This would make embroider an implementation detail of ember-cli, which is where we want to head anyway.

Just to be clear, my position (and I believe that Stef is also of the same opinion) all "build pipeline code" that exists in `ember-cli` should be removed and the blueprint needs to be able to iterate how _new_ apps are emitted without breaking existing ones. The only way (AFAICT) to do this without significant juggling is to allow the build pipeline code to be provided by a different package (what the current Embroider README.md says, and this PR does).

rwjblue:
Going to land this, and iterate things on master. The additional CI cost (at the moment) is very low (because the smoke tests are skipped here), so it should not significantly impact our ability to iterate.

@thoov - Would you mind making a new issue that is specifically for tracking the experiment?

PR #9210, opened 2020-05-04
Revert "test version-utils"

rwjblue:
This reverts commit 5a8152a645622b5ea1e4172535825ef44e672c10.

This is being reverted as it causes a failure when ember-cli is ran as part of CITGM (see https://github.com/nodejs/citgm/pull/804). We should rework the test to avoid mocking `require('git-repo-info')` in order to bring it back.

/cc @locks


locks:
I'll take a look in a follow up PR.

PR #9212, opened 2020-05-05
Ensure that the captured exit is released.

rwjblue:
Prior to this, an error thrown during the bulk of `lib/cli/index.js` would have caused the process to be captured (which happens very early) but not released. This is because prior to this change we only ensured cleanup around the `cli.run` asynchrony, but not for any of the other work there.


rwjblue:
The diff doesn't make this clear, but it really is just moving the `try` that was already present up near the top of the function and re-indenting things.

PR #9223, opened 2020-05-11
Add timeout to CI runs

rwjblue:
Prevents CI runs from running much much longer (on accident) when they should probably just timeout.


rwjblue:
This was incorrect

PR #9238, opened 2020-05-22
Add config/ember-cli-update.json to app and addon blueprints.

### LINKED ISSUE #9237: Add ember-cli-update.json file to new app / addon blueprint?
Now that ember-cli-update will honor this file by default, I think we should add it to the default blueprints. Without it, using ember-cli-update is a bit annoying in a world with dependabot / renovate / green keeper.

Thoughts?

rwjblue:
Enables apps and addons to update ember-cli without impacting future `ember-cli-update` runs.

Without this `ember-cli-update` will use the current version of `ember-cli` in the `package.json` to determine which version to attempt to update _from_. Unfortunately, this means that if you were to update your `ember-cli` version without _also_ running the full blueprint update, you would have a much harder time of updating your general project structure in the future.

Support for this format and process was nailed down and merged in [emberjs/rfcs#477](https://github.com/emberjs/rfcs/blob/master/text/0477-blueprints-update.md).

Closes #9237.


blueprints/app/index.js
```diff
@@ -21,6 +21,18 @@ module.exports = {
...
+    let blueprintOptions = '';
+    if (hasOptions) {
+      let indent = `\n            `;
+      let outdent = `\n          `;
+
+      blueprintOptions =
+        indent +
+        [!options.welcome && '"--no-welcome"', options.yarn && '"--yarn"'].filter(Boolean).join(',\n            ') +
+        outdent;
+    }
```
rwjblue, line 34: This is very similar between app and addon blueprints, and I tried to end up sharing the code but the shared version was much harder to reason about. This is specifically because the default values for `--welcome` (and therefore the flag value) **change** between the `app` and `addon` blueprint.

tests/acceptance/new-test.js
```diff
@@ -379,16 +379,56 @@ describe('Acceptance: ember new', function () {
...
+      let fixturePath = path.join(__dirname, '../fixtures', fixtureName, fileName);
       let fixtureContents = fs
         .readFileSync(fixturePath, { encoding: 'utf-8' })
         .replace('<%= emberCLIVersion %>', currentVersion);
 
-      expect(file('package.json')).to.equal(fixtureContents);
+      expect(file(fileName)).to.equal(fixtureContents);
     }
 
+    it('app defaults', async function () {
```
rwjblue, line 392: I created this because we actually didn't have any of our blueprint templating tests for `ember new foo` (we have tests for a two different permutations `--yarn` and `--no-welcome`).

tests/acceptance/new-test.js
```diff
@@ -432,29 +474,8 @@ describe('Acceptance: ember new', function () {
         expect(file(filePath)).to.equal(file(path.join(__dirname, '../fixtures', fixturePath, filePath)));
       });
 
-      checkPackageJson(fixturePath);
-    });
-
-    it('addon + npm + !welcome', async function () {
```
rwjblue, line 438: This test was moved (unchanged) up and renamed to `addon defaults`.

blueprints/app/files/config/ember-cli-update.json
```diff
@@ -0,0 +1,18 @@
+{
+  "schemaVersion": "1.0.0",
+  "packages": [
+    {
+      "name": "ember-cli",
+      "version": "<%= emberCLIVersion %>",
+      "blueprints": [
+        {
+          "name": "<%= blueprint %>",
+          "outputRepo": "https://github.com/ember-cli/ember-<%= blueprint === 'app' ? 'new' : 'addon' %>-output",
```
kategengler, line 10: I don't see `outputRepo` in the associated RFC, and I only found minimal documentation on `ember-cli-update`'s guides -- will the `outputRepo` generated be valid when generating with options that that don't match the ones that we use to generate those repos? For example, if I generate with `--yarn` and this file has both `ember-new-output` as the `outputRepo` and `--yarn` in the options, what happens?

rwjblue: The options (and version) specified in the file are used to generate (from scratch) the source diff. So specifically, using this system will actually mean you get correct diffs applied when in a project that uses yarn.

kategengler: Great -- that wasn't clear to me from the documentation on ember-cli-update. It sounded like `outputRepo` was a thing to specify only if it could be used in place of the local generation & diffing.

kellyselden: Katie you are right. outputRepo is the old style, where you don't generate the diff locally. It is still used if you run `ember-cli-update compare`, to get a quick diff on github, without doing a bunch of work locally. It can also be used by custom blueprints to keep track of their diffs in the same way.

rwjblue: @kellyselden so should I remove it? Also why does `npx ember-cli-update bootstrap` emit it?

kellyselden: I would say keep it. It's a super quick way to see what has changed. `ember-cli-update compare` opens it up in your browser with the diff.

rwjblue: kk, thanks again @kellyselden!

PR #9259, opened 2020-06-13
RFC #635: `ember new --lang`

josephdsumner:
@rwjblue  @melSumner @hergaiety @jgwhite 

I'm so excited to see the decision to move forward with [RFC #635](https://github.com/emberjs/rfcs/pull/635) and get started on working through this!
 
This is an updated rebase + squash of the feature branch that was developed within the Ember.js Accessibility Strike Team and proposed as a candidate implementation in the RFC. There have been a few updates since the RFC was first submitted, and of course, some changes to make per review. To the best of my knowledge, this is an accurate, aggregate summary of action items:

From last st-a11y review: 
- [x] rebase + squash onto ember-cli master as a working starting point + make sure recent blueprint changes are resolved

- [x] tidy up evaluation for html tag in index.html
-- @rwjblue you mentioned a 'one-liner' for this, I think I understood what you meant

- [x] update the coverage for programming languages to include file extensions, abbreviations, etc.
-- found + marked edge cases

- [x] change util logic to check for misuse cases first
-- will need to be updated for edges per review

- [x] Update description / help for `addon` command
-- clarify that it is the dummy app getting the lang value assigned

- [x] Test coverage: index-test.html

- [x] Test coverage: put entire file in fixtures for `index.html` 
-- read result file, confirm it matches full expected result (long String, literal, etc.)

- [x] Change default type/value from empty String to undefined

From RFC Discussion (briefly):
- [x] Handling of edge cases (valid language code and programming language abbreviation)

- [x] Handling of error messages (verbosity)

- [x] Handling of error work flow (misuse cases: when to build, when to break, etc.)

- [x] An appropriate implementation for working around the out-of-scope `parser` bug


lib/commands/init.js
```diff
@@ -29,6 +30,13 @@ module.exports = Command.extend({
     },
     { name: 'yarn', type: Boolean }, // no default means use yarn if the blueprint has a yarn.lock
     { name: 'name', type: String, default: '', aliases: ['n'] },
+    {
+      name: 'lang',
+      type: String,
+      default: '',
+      description: 'Sets the base human language of the application via index.html',
+      aliases: ['l'],
```
kellyselden, line 38: Does it need a short alias?

josephdsumner: > Does it need a short alias?

That's a great point! Probably doesn't need one. We started out with the --language option (with -lang and -l as the aliases), but decided to make the option --lang so it directly matched the html attribute it was assigning. If -lang was short enough to be an alias before that change, makes sense that --lang doesn't need a short alias.

hergaiety:
When this is finished and merged I'd love to follow-up with an update to the Ember guides with a mention of how to use Ember CLI to specify the language on `guides/release/accessibility/application-considerations.md` under the `Lang Attribute` section.

blueprints/addon/index.js
```diff
@@ -144,6 +144,8 @@ module.exports = {
         outdent;
     }
 
+    let lang = options.lang;
```
rwjblue, line 147: I'd vaguely prefer for this to follow suit with the style below. We either need to update `welcome: options.welcome` and `yarn: options.yarn` to destructure, or change this one to use `lang: options.lang`. Either way seems totally fine to me.

blueprints/app/index.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
         [!options.welcome && '"--no-welcome"', options.yarn && '"--yarn"'].filter(Boolean).join(',\n            ') +
         outdent;
     }
+    let lang = options.lang;
```
rwjblue, line 35: I'd vaguely prefer for this to follow suit with the style below. We either need to update `welcome: options.welcome` and `yarn: options.yarn` to destructure, or change this one to use `lang: options.lang`. Either way seems totally fine to me.

lib/commands/addon.js
```diff
@@ -15,6 +15,13 @@ module.exports = NewCommand.extend({
     { name: 'skip-git', type: Boolean, default: false, aliases: ['sg'] },
     { name: 'yarn', type: Boolean }, // no default means use yarn if the blueprint has a yarn.lock
     { name: 'directory', type: String, aliases: ['dir'] },
+    {
+      name: 'lang',
+      type: String,
+      default: '',
```
rwjblue, line 21: I think technically we do not have a default. 

```suggestion
```

As it stands this says the default is `''` (empty string) but the blueprint only adds `lang=` if a value is set...

lib/commands/init.js
```diff
@@ -29,6 +30,13 @@ module.exports = Command.extend({
     },
     { name: 'yarn', type: Boolean }, // no default means use yarn if the blueprint has a yarn.lock
     { name: 'name', type: String, default: '', aliases: ['n'] },
+    {
+      name: 'lang',
+      type: String,
+      default: '',
```
rwjblue, line 36: Same point RE: default. 

```suggestion
```

lib/commands/init.js
rwjblue: Seems fine to keep, though I do want to point out that we should **not** use the shortened versions of _any_ flags in our own documentation or guides (other than to document the shorthand :wink:).

lib/commands/init.js
```diff
@@ -78,6 +86,14 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(`We currently do not support a name of \`${packageName}\`.`));
     }
 
+    // [WIP] ember new foo --lang
```
rwjblue, line 89: What is the status of this comment? Does it stick around long term? If so, it needs to be tweaked to make a bit more sense (and remove the `WIP` label).

lib/commands/init.js
```diff
@@ -78,6 +86,14 @@ module.exports = Command.extend({
...
+    let langDef = this.availableOptions.find((opt) => opt.name === 'lang').default;
+    let langInfo = getLangArgInfo(commandOptions.lang);
+    if (!langInfo.result && langInfo.langArgValue !== langDef) {
+      return Promise.reject(new SilentError(langInfo.message));
+    }
+    commandOptions.lang = langInfo.langArgValue;
```
rwjblue, line 95: I would think that this would be sufficient:

```suggestion
    if (commandOptions.lang) {
      let langInfo = getLangArgInfo(commandOptions.lang);
      commandOptions.lang = langInfo.langArgValue;
    }
```

RE: using the custom `langInfo.message`, I would expect `getLangArgInfo` to directly `throw new SilentError(langInfo.message)` internally.

lib/commands/new.js
```diff
@@ -64,6 +72,14 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(message));
     }
 
+    // [WIP] ember new foo --lang
+    let langDef = this.availableOptions.find((opt) => opt.name === 'lang').default;
+    let langInfo = getLangArgInfo(commandOptions.lang);
+    if (langInfo.langArgValue !== langDef && !langInfo.result) {
+      return Promise.reject(new SilentError(langInfo.message));
+    }
+    commandOptions.lang = langInfo.langArgValue;
```
rwjblue, line 81: Same comment as above, though I don't understand why we need the logic in _both_ `commands/init.js` and `commands/new.js` (since `commands/new.js` is directly invoking `commands/init.js` for the initialization functionality).

package.json
```diff
@@ -146,6 +147,7 @@
     "fixturify": "^2.1.0",
     "latest-version": "^5.1.0",
     "mocha": "^7.2.0",
+    "is-language-code": "^1.0.7",
```
rwjblue, line 150: This should not be both in `dependencies` and `devDependencies`. Should probably remove this one.

```suggestion
```

tests/acceptance/addon-dummy-generate-test.js
```diff
@@ -240,4 +240,69 @@ describe('Acceptance: ember generate in-addon-dummy', function () {
     await generateInAddon(['server', '--dummy']);
     expect(file('server/index.js')).to.exist;
   });
+
+  // [WIP] ember addon --lang flag
```
rwjblue, line 244: Should probably kill this `WIP` comment

```suggestion
  // ember addon --lang flag
```

tests/acceptance/addon-dummy-generate-test.js
```diff
@@ -240,4 +240,69 @@ describe('Acceptance: ember generate in-addon-dummy', function () {
...
+  // -------------------------------
+  // Good: Default
+  it('ember addon without --lang flag (default) has no lang attribute in dummy index.html', async function () {
+    await ember(['addon', 'foo', '--skip-npm', '--skip-bower', '--skip-git']);
+    expect(file('tests/dummy/app/index.html')).to.contain('<html>');
+  });
```
rwjblue, line 250: I'd prefer to account for this in the main blueprinting tests for default output of `new`/`addon` output here:

https://github.com/ember-cli/ember-cli/blob/5237dc7f67e766c8da1a56e329b6c91369c0c34c/tests/acceptance/new-test.js#L439-L460

tests/acceptance/addon-dummy-generate-test.js
```diff
@@ -240,4 +240,69 @@ describe('Acceptance: ember generate in-addon-dummy', function () {
...
+    let err = await expect(ember(['addon', 'foo', '--skip-npm', '--skip-bower', '--skip-git', '--lang'])).to.be
+      .rejected;
+    expect(err.name).to.equal('SilentError');
+    expect(err.message).to.be.ok;
+    expect(err.message).to.include('An error with the `--lang` flag returned the following message:');
+    expect(err.message).to.include('Information about using the `--lang` flag:');
+    expect(err.message).to.not.include('set the app programming language');
+    expect(err.message).to.include('ember-cli command option');
+    expect(err.message).to.include('--disable-analytics');
+  });
```
rwjblue, line 295: I think we should remove this test

tests/acceptance/init-test.js
```diff
@@ -166,4 +166,67 @@ describe('Acceptance: ember init', function () {
 
     expect(dir('.git')).to.not.exist;
   });
+
+  // [WIP] ember init --lang flag
```
rwjblue, line 170:
```suggestion
  // ember init --lang flag
```

tests/acceptance/init-test.js
```diff
@@ -166,4 +166,67 @@ describe('Acceptance: ember init', function () {
...
+  // -------------------------------
+  // Good: Default
+  it('ember init without --lang flag (default) has no lang attribute in index.html', async function () {
+    await ember(['init', '--skip-npm', '--skip-bower', '--skip-git']);
+    confirmBlueprinted();
+    // expect(file('app/index.html')).to.contain('<html>');
+  });
+
```
rwjblue, line 178:
```suggestion
```

tests/acceptance/init-test.js
```diff
@@ -166,4 +166,67 @@ describe('Acceptance: ember init', function () {
...
+  // Good: Correct Usage
+  it('ember init with --lang flag and valid code assigns lang attribute in index.html', async function () {
+    await ember(['init', '--skip-npm', '--skip-bower', '--skip-git', '--lang=en-US']);
+    // expect(file('app/index.html')).to.contain('<html lang="en-US">');
```
rwjblue, line 182: Should probably be uncommented 😄

tests/acceptance/init-test.js
```diff
@@ -166,4 +166,67 @@ describe('Acceptance: ember init', function () {
...
+    let err = await expect(ember(['init', '--skip-npm', '--skip-bower', '--skip-git', '--lang'])).to.be.rejected;
+    expect(err.name).to.equal('SilentError');
+    expect(err.message).to.be.ok;
+    expect(err.message).to.include('An error with the `--lang` flag returned the following message:');
+    expect(err.message).to.include('Information about using the `--lang` flag:');
+    expect(err.message).to.not.include('set the app programming language');
+    expect(err.message).to.include('ember-cli command option');
+    expect(err.message).to.include('--disable-analytics');
+  });
+
```
rwjblue, line 221: I think we should remove these tests, our normal command option parsing should enforce this. If it doesn't, please make a (somewhat) detailed issue and then we can track that down after landing this.
 
```suggestion
```

tests/acceptance/new-test.js
```diff
@@ -101,6 +101,104 @@ describe('Acceptance: ember new', function () {
     expect(file('app/templates/application.hbs')).to.contain('Welcome to Ember');
   });
 
+  // [WIP] ember new --lang flag
```
rwjblue, line 104:
```suggestion
  // ember new --lang flag
```

tests/acceptance/new-test.js
```diff
@@ -101,6 +101,104 @@ describe('Acceptance: ember new', function () {
...
+  // -------------------------------
+  // Good: Default
+  it('ember new without --lang flag (default) has no lang attribute in index.html', async function () {
+    await ember(['new', 'foo', '--skip-npm', '--skip-bower', '--skip-git']);
+    expect(file('app/index.html')).to.contain('<html>');
+  });
+
```
rwjblue, line 111: I'd prefer to account for this in the main blueprinting tests for default output of `new`/`addon` output here:

https://github.com/ember-cli/ember-cli/blob/5237dc7f67e766c8da1a56e329b6c91369c0c34c/tests/acceptance/new-test.js#L439-L460


```suggestion
```

tests/acceptance/new-test.js
```diff
@@ -101,6 +101,104 @@ describe('Acceptance: ember new', function () {
...
+    let err = await expect(ember(['new', 'foo', '--skip-npm', '--skip-bower', '--skip-git', '--lang'])).to.be.rejected;
+    expect(err.name).to.equal('SilentError');
+    expect(err.message).to.be.ok;
+    expect(err.message).to.include('An error with the `--lang` flag returned the following message:');
+    expect(err.message).to.include('Information about using the `--lang` flag:');
+    expect(err.message).to.not.include('set the app programming language');
+    expect(err.message).to.include('ember-cli command option');
+    expect(err.message).to.include('--disable-analytics');
+  });
+
```
rwjblue, line 189: See above. I think we should fix this issue holistically not per-option, as such I would prefer to avoid adding specific tests for this. 

I'm happy to help someone work through a fix for the general parsing issue though...

```suggestion
```

tests/unit/commands/init-test.js
```diff
@@ -227,4 +227,86 @@ describe('init command', function () {
       expect(reason).to.equal('Called run');
     });
   });
+
+  // [WIP] ember init --lang flag
```
rwjblue, line 231:
```suggestion
  // ember init --lang flag
```

tests/unit/commands/init-test.js
```diff
@@ -227,4 +227,86 @@ describe('init command', function () {
...
+
+  // // Misuse: No specification + hidden option
+  // it('ember init with --lang flag but no specification fails with an error message; absorbs appended ember-cli option (hidden)', async function () {
+  //   let err = await expect(command.validateAndRun(['foo', '--skip-npm', '--skip-bower', '--skip-git', '--lang'])).to.be.rejected;
+  //   expect(err.name).to.equal('SilentError');
+  //   expect(err.message).to.be.ok;
+  //   expect(err.message).to.not.include('set the app programming language');
+  //   expect(err.message).to.include('ember-cli command option');
+  //   expect(err.message).to.include('--disable-analytics');
+  // });
```
rwjblue, line 299: Should be uncommented (or deleted)...

tests/unit/commands/new-test.js
```diff
@@ -106,4 +106,91 @@ describe('new command', function () {
     let reason = await command.validateAndRun(['foo', '--custom-option=customValue']);
     expect(reason).to.equal('Called run');
   });
+
+  // [WIP] ember new --lang flag
```
rwjblue, line 110:
```suggestion
  // ember new --lang flag
```

tests/unit/commands/new-test.js
```diff
@@ -106,4 +106,91 @@ describe('new command', function () {
...
+  it('ember new with --lang flag and invalid code fails with an error message', async function () {
+    let err = await expect(command.validateAndRun(['foo', '--skip-npm', '--skip-bower', '--skip-git', '--lang=en-UK']))
+      .to.be.rejected;
+    expect(err.name).to.equal('SilentError');
+    expect(err.message).to.be.ok;
+    expect(err.message).to.include('An error with the `--lang` flag returned the following message:');
+    expect(err.message).to.include('Information about using the `--lang` flag:');
+    expect(err.message).to.not.include('set the app programming language');
+    expect(err.message).to.not.include('ember-cli command option');
+  });
```
rwjblue, line 195: I would prefer for this logic to live inside `init` command, and only be tested in one place (likely `tests/acceptance/new-test.js`).

tests/unit/commands/init-test.js
```diff
@@ -227,4 +227,86 @@ describe('init command', function () {
...
+  it('ember init with --lang flag and invalid code fails with an error message', async function () {
+    let err = await expect(command.validateAndRun(['foo', '--skip-npm', '--skip-bower', '--skip-git', '--lang=en-UK']))
+      .to.be.rejected;
+    expect(err.name).to.equal('SilentError');
+    expect(err.message).to.be.ok;
+    expect(err.message).to.include('An error with the `--lang` flag returned the following message:');
+    expect(err.message).to.include('Information about using the `--lang` flag:');
+    expect(err.message).to.not.include('set the app programming language');
+    expect(err.message).to.not.include('ember-cli command option');
+  });
```
rwjblue, line 311: Lets delete all of these tests here, and rely on them in the `tests/acceptance/new-test.js` (they are effectively duplicated here, in new commands unit tests, and in new's acceptance tests).

lib/commands/init.js
kellyselden: I vote for removing it. --lang is already short, and we consume -l when we may want it for a future longer option starting with L.

blueprints/app/index.js
josephdsumner: Thanks for spotting this! I destructured it before the utilities and template logic were consolidated and forgot to revert it once I finished. I can absolutely update `lang` to be consistent with `yarn` and `welcome`.

blueprints/addon/index.js
josephdsumner: Thanks for spotting this! I destructured it before the utilities and template logic were consolidated and forgot to revert it once I finished. I can absolutely update `lang` to be consistent with `yarn` and `welcome`.

package.json
josephdsumner: Can do -- thanks!

lib/commands/init.js
josephdsumner: It can definitely go! Thanks!

lib/commands/addon.js
josephdsumner: > I think technically we do not have a default.
> 
> As it stands this says the default is `''` (empty string) but the blueprint only adds `lang=` if a value is set...

As in, there's no consensus about whether or not to have a valid language code by default or what that code would be? AFAIK there isn't, and I definitely think that discussion exceeds the scope of this 'tooling-only' RFC.

I absolutely agree with you that we should use `undefined` instead. My rationale for choosing the empty String initially was that it was falsy enough to handle the 'no-input' logic planned for the utility *and* it would serve as a convenient vehicle for receiving a default language code *in the event that such a decision ever got made*. However, the choice to use the String could definitely be (mis)interpreted as a biased one, i.e., that it treats the outcome of any discussion about a default as a foregone conclusion. I'd love to rule that out, and `undefined` should do that.

I think I've covered the change appropriately with [this set of changes](https://github.com/josephdsumner/ember-cli/pull/6/files). The only notable change in control is the utility: it asserts that the input is defined before trying to do any formatting / evaluation / validation.

lib/commands/init.js
josephdsumner: [Addressed above](https://github.com/ember-cli/ember-cli/pull/9259#discussion_r459765499).

lib/commands/new.js
```diff
@@ -64,6 +71,13 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(message));
     }
 
+    let langDef = this.availableOptions.find((opt) => opt.name === 'lang').default;
+    let langInfo = getLangArgInfo(commandOptions.lang);
+    if (langInfo.langArgValue !== langDef && !langInfo.result) {
+      return Promise.reject(new SilentError(langInfo.message));
```
rwjblue, line 77: Since we are already in an async function, there is no need to `return Promise.reject` here. Doing this is the same thing:

```suggestion
      throw new SilentError(langInfo.message);
```

lib/commands/new.js
```diff
@@ -64,6 +71,13 @@ module.exports = Command.extend({
       return Promise.reject(new SilentError(message));
     }
 
+    let langDef = this.availableOptions.find((opt) => opt.name === 'lang').default;
```
rwjblue, line 74: AFAICT there is no default anymore, we should be able to remove this.

lib/utilities/valid-lang-flag.js
```diff
@@ -0,0 +1,170 @@
+/* 
+[WIP] 
+
```
rwjblue, line 3:
```suggestion
```

rwjblue:
Also went through and resolved the comments that have already been addressed...

josephdsumner:
> Also went through and resolved the comments that have already been addressed...

Thank you so much -- feel free to wait for a request for re-review as I consolidate the cleanup :)

tests/acceptance/addon-dummy-generate-test.js
josephdsumner: Moved in 29e0a4b1 👍

tests/acceptance/addon-dummy-generate-test.js
josephdsumner: Removed in d3b0f8c 👍

tests/acceptance/new-test.js
josephdsumner: Removed in d3b0f8c 👍

tests/acceptance/init-test.js
josephdsumner: Removed when consolidating testing suite to acceptance/new-test in 72e10e9 👍

lib/commands/new.js
josephdsumner: Removed when consolidating `lang`-related logic into `init`  in f95d3a9  👍

lib/commands/new.js
josephdsumner: This got removed when consolidating lang-related logic into init in f95d3a9, but I did make the corresponding change int init. I haven't yet updated the utility throw the error internally.

tests/unit/commands/init-test.js
josephdsumner: Removed as part of 72e10e9 👍

lib/commands/init.js
josephdsumner: This has been partially resolved in f95d3a9 in that the logic here has been updated. The error handling hasn't been moved to `getLangArgInfo` yet, but in the meantime, it has been updated per the comment about using `throw`.

lib/commands/new.js
josephdsumner: Removed when consolidating lang-related logic into init in f95d3a9 👍

tests/acceptance/init-test.js
josephdsumner: 😄 Going to be hard to beat that one! Removed as part of 72e10e9 👍

tests/unit/commands/init-test.js
josephdsumner: Removed as part of 72e10e9 👍

tests/unit/commands/new-test.js
josephdsumner: This should be covered by f95d3a9 and 72e10e9, respectively 👍

tests/acceptance/new-test.js
josephdsumner: Relocated in 5ca7693f 👍

tests/acceptance/init-test.js
josephdsumner: Removed in 72e10e9 👍

lib/commands/init.js
josephdsumner: Update:  the error handling has been moved to the utility (and removed from `init`) as part of b3f03f4 👍

lib/commands/init.js
josephdsumner: I don't think it's really my place to 'vote' here, but I did think about this throughout the RFC process from as many angles as possible:

Reasons to keep:
- no reason not to consume for a feature that was fairly well-received by the community
- an a11y-related flag is something you don't see often (if at all), so landing the feature has a bit of a 'making a statement' PR bonus to it. I don't know if that would be diminished by making `--lang` not as 'fully-featured' as other things
- there was definitely a great deal of 'how do we name this' discussion during review, but there wasn't much concern raised about the alias. However, I'm willing to admit that this might be due to there being more pressing concerns to discuss (e.g., confusion WRT programming languages) and a group bias (people working on an a11y-st are probably more likely to think the flag 'deserves' an alias)
- there is definitely a precedent for using two-to-three letter aliases as needed, and in principle, the future longer option starting with L would be more suitable to one of those anyway

Reasons to remove:
- don't want to be wasteful in general -- *especially* in `new`
- don't want to cause confusion with some other widely-used `-l` (here or elsewhere, things like `--list` and `--local` come to mind)
- don't want to consume the alias if there's a longer, more sensible choice coming in the roadmap
- the decision to go with `--lang` over `--language` was largely shaped by `lang` being the specific attribute set in `index.html` -- i.e., that it was an a11y-related teaching opportunity. I know that we do plan to teach this appropriately (and only with the full name) where the new flag is used in the documentation, but I suppose there's a chance for that opportunity to be lost / the meaning to be easily forgotten if we provide a shortcut

Taken together, I'd probably say that unless there's a concrete precedent that we will cause confusion with some other widely-used `-l` or evidence that we've got something in the roadmap that would definitely be a more suitable choice, it makes most sense to keep it.

kellyselden: We can always add it later once the pain of the full option is known. It's however not as easy to remove it or repurpose it once a better use comes along.

josephdsumner:
@rwjblue Requesting a re-review with the hope that the following have been resolved:
- test consolidation into `acceptance/new-test`
- command logic consolidation into `init`
- error handling moved directly to the `valid-lang-flag` utility
- extraneous comment removal

AFAIK, that would take care of a lot of the architectural cleanup, but would still leave at least few more TODOs to reach the MVP. There's probably a missing item or two here, but it should become much more apparent what those are with the above taken care of first:
- `addon` help text (can do)
- `index-test` tests (suggestion would be helpful here)
- utility update: implementation to match whatever consensus was reached re: when to break and what to report. For example, what do to when there's a valid human AND programming language overlap. The latest info I have is from the RFC comment section itself, but have no idea whether or not framework agreed on a spec, etc.

lib/commands/init.js
hergaiety: How do we determine which flags are "better" than another flag? Currently this RFC was accepted and is being implemented, which seems "better" or "more useful" than a hypothetical potential thing that may exist in the future. I for one vote to give this flag as much importance as any other flag that we have added in the past.

rwjblue:
> implementation to match whatever consensus was reached re: when to break and what to report. For example, what do to when there's a valid human AND programming language overlap. The latest info I have is from the RFC comment section itself, but have no idea whether or not framework agreed on a spec, etc.

We decided to emit a warning instead of an error, and continue along using the user provided value.

josephdsumner:
> > implementation to match whatever consensus was reached re: when to break and what to report. For example, what do to when there's a valid human AND programming language overlap. The latest info I have is from the RFC comment section itself, but have no idea whether or not framework agreed on a spec, etc.
> 
> We decided to emit a warning instead of an error, and continue along using the user provided value.

Thanks! I will make that the model for all the mis-use cases:
- log warning (issue description)
- log result (inform user about what will happen in index.html)
- log help-related info
- never break
- only assign lang if valid used properly

lib/commands/addon.js
```diff
@@ -18,7 +18,7 @@ module.exports = NewCommand.extend({
     {
       name: 'lang',
       type: String,
-      description: 'Sets the base human language of the application via index.html',
+      description: 'Sets the base human language of the dummy application via index.html',
```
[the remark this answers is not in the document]
josephdsumner: Definitely clearer! Updated in 2057331 👍

rwjblue:
I've just rebased / squashed things down here in prep for getting this landed. I've also removed the `-l` shorthand based on convo with a few folks.

This should be good to go once CI is happy, great work here folks!

josephdsumner:
So great to see this land! 
@MelSumner @rwjblue @jgwhite @hergaiety and the rest of the strike team --
Thank you all **so** much for all of the time and effort devoted to guidance and review!

PR #9263, opened 2020-06-25
Use workers to print asset sizes for production builds

### LINKED ISSUE #9260: Production builds takes a lot of time to print compression stats
If the build assets are larger then printing the compression statistics like gzipped sizes is taking time since all the assets are serially processed by compressing using `gzip` and then printing the size. Can we make this faster by making it run concurrently or spawning multiple processes for the same? I see some room for optimization there.

## Why?
- We can make use of the multiple cores present in a machine to do faster builds.
- This will eventually result in quicker prod builds and faster deployments.

@rwjblue What are your thoughts?

rajasegar:
Fixes #9260


lib/models/asset-size-printer.js
```diff
@@ -63,16 +63,16 @@ module.exports = class AssetPrinterSize {
...
-          let contentsBuffer = fs.readFileSync(file);
-          return gzip(contentsBuffer).then((buffer) => ({
-            name: file,
-            size: contentsBuffer.length,
-            gzipSize: buffer.length,
-            showGzipped: contentsBuffer.length > 0,
-          }));
+          return pool
+            .exec('gzipStats', [file])
+            .then((result) => result)
```
bekzod, line 68: can't we remove this ?

lib/models/asset-size-printer.js
```diff
@@ -63,16 +63,16 @@ module.exports = class AssetPrinterSize {
...
+            .catch((err) => console.error(err));
         });
```
bekzod, line 71: can't we just pass all assets at once, instead of mapping it like 
```javascript
assets = Promise.all(pool.exec('gzipStats', assets));
```

lib/models/asset-size-printer.js
bekzod, line 68: shouldn't we remove this ?

lib/models/asset-size-printer.js
```diff
@@ -63,16 +63,16 @@ module.exports = class AssetPrinterSize {
...
-          return gzip(contentsBuffer).then((buffer) => ({
-            name: file,
-            size: contentsBuffer.length,
-            gzipSize: buffer.length,
-            showGzipped: contentsBuffer.length > 0,
-          }));
+          return pool
+            .exec('gzipStats', [file])
+            .then((result) => result)
+            .catch((err) => console.error(err));
```
bekzod, line 69: should we be handling error like this ? `console.error` was not there before

lib/models/gzipStats.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const fs = require('fs');
+const zlib = require('zlib');
+const util = require('util');
+const gzip = util.promisify(zlib.gzip);
+
+module.exports = function (file) {
+  let contentsBuffer = fs.readFileSync(file);
```
bekzod, line 9: shouldn't we use promisified version of `fs.readFile` to be consistent with code below

lib/models/asset-size-printer.js
rajasegar: Remove what? @bekzod Can you please elaborate?

lib/models/asset-size-printer.js
rajasegar: I agree we shouldn't be handling error like this, any pointers?

lib/models/gzipStats.js
rajasegar: It used to be sync version previously also here
https://github.com/ember-cli/ember-cli/blob/8b168056ccd18679a16aa29c3a2de1a20df0e368/lib/models/asset-size-printer.js#L66
I just moved it to a separate file.

lib/models/asset-size-printer.js
bekzod: sorry, I meant ` .then((result) => result)`

rwjblue: `.then((something) => something)` is the same as not adding the `.then` at all.

lib/models/asset-size-printer.js
```diff
@@ -63,16 +63,16 @@ module.exports = class AssetPrinterSize {
...
         });
 
-      return resolve(Promise.all(assets));
+      return Promise.all(assets).then((data) => {
+        pool.terminate(); // terminate all workers when done
```
rwjblue, line 73: This should probably be in a `.finally`

lib/models/asset-size-printer.js
```diff
@@ -49,12 +49,12 @@ module.exports = class AssetPrinterSize {
 
   makeAssetSizesObject() {
     return new Promise((resolve) => {
```
rwjblue, line 51: I think we can remove this manual `new Promise` wrapper now

lib/models/gzipStats.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const fs = require('fs');
+const zlib = require('zlib');
+const util = require('util');
+const gzip = util.promisify(zlib.gzip);
+
+module.exports = function (file) {
```
rwjblue, line 8:
```suggestion
module.exports = async function (file) {
```

package.json
```diff
@@ -125,6 +125,7 @@
     "uuid": "^8.1.0",
     "walk-sync": "^2.1.0",
     "watch-detector": "^1.0.0",
+    "workerpool":"^3.1.1",
```
rwjblue, line 128: The current release is 6.0.0, why use such an old version?

lib/models/asset-size-printer.js
rajasegar: done

lib/models/asset-size-printer.js
rajasegar: done

lib/models/asset-size-printer.js
rajasegar: It seems like the test is failing if I remove the new promise wrapper so I reverted only that change.

package.json
rajasegar: Because that's what there in yarn.lock and I don't want to introduce any breaking changes.

rwjblue: Hmm, but AFAICT we didn't use it before. If it was in the `yarn.lock`, it was from some transitive dependency right?

I _believe_ that this PR is ember-cli's first direct usage of `workerpool` (though we have used it indirectly for a while through `broccoli-babel-transpiler`), we should add the current version...

lib/models/asset-size-printer.js
rwjblue: Hmm, I don't understand what you mean. How is this done?

rwjblue: My point above was that we should make the `makeAssetSizesObject` method itself an async function, then we don't need to `return new Promise` at all (but we can use `async`/`await` to wait for the `pool.exec` results.

rajasegar:
@rwjblue Thanks for the tip
- `workerpool` updated to use the latest version 6.0.0
- Refactor `makeAssetSizesObject` function to use async/await
Please check and let me know

lib/models/asset-size-printer.js
```diff
@@ -47,36 +47,30 @@ module.exports = class AssetPrinterSize {
...
+      // Skip test files
+      .filter((file) => {
+        let filename = path.basename(file);
+        return !testFileRegex.test(filename);
+      })
+      // Print human-readable file sizes (including gzipped)
+      .map((file) => {
+        return pool
+          .exec('gzipStats', [file])
+          .then((result) => result)
```
rwjblue, line 67: This `.then(result => result)` doesn't do anything (it chains a promise, returning the passed in value), we should be able to delete it.

```suggestion
```

lib/models/asset-size-printer.js
```diff
@@ -47,36 +47,30 @@ module.exports = class AssetPrinterSize {
...
+          .catch((err) => console.error(err));
+      });
```
rwjblue, line 69: I think we'll want this in a try/finally block to ensure the pool is terminated:

```suggestion
    try {
      let assets = files
        // Skip test files
        .filter((file) => {
          let filename = path.basename(file);
          return !testFileRegex.test(filename);
        })
        // Print human-readable file sizes (including gzipped)
        .map((file) => {
          return pool
            .exec('gzipStats', [file])
            .catch((err) => console.error(err));
        });
        
        return await Promise.all(assets);
      } finally {
        pool.terminate();
      }
```

lib/models/asset-size-printer.js
```diff
@@ -47,36 +47,30 @@ module.exports = class AssetPrinterSize {
...
-        .then((data) => {
-          return resolve(data);
-        })
-        .finally(() => {
-          pool.terminate(); // terminate all workers when done
-        });
-    });
+    const result = await Promise.all(assets);
+    pool.terminate(); // terminate all workers when done
+    return result;
```
rwjblue, line 73: Replaced by my suggestion above...

```suggestion
```

rajasegar:
@rwjblue Review suggestions applied, please check and let me know

lib/models/asset-size-printer.js
```diff
@@ -63,17 +63,13 @@ module.exports = class AssetPrinterSize {
...
         // Print human-readable file sizes (including gzipped)
         .map((file) => {
-          let contentsBuffer = fs.readFileSync(file);
-          return gzip(contentsBuffer).then((buffer) => ({
-            name: file,
-            size: contentsBuffer.length,
-            gzipSize: buffer.length,
-            showGzipped: contentsBuffer.length > 0,
-          }));
+          return pool.exec('gzipStats', [file]).catch((err) => console.error(err));
```
rwjblue, line 66: Do we need this `.catch`? What does it provide us? I _think_ it is emitting the error then swallowing the failure, is that intentional?

Previously, an error during this computation would have failed the command (which as far as I can tell is what would happen if this `.catch(err => console.error(err))` was removed). 

Lets try removing it:

```suggestion
          return pool.exec('gzipStats', [file]);
```

rajasegar:
@rwjblue I made the changes, but for some reason the windows builds are failing

PR #9277, opened 2020-07-10
[ENHANCEMENT] Allow ember install to work with Yarn v2

### LINKED ISSUE #9276: "ember install" fails with yarn v2
_Originally posted by @gilest in https://github.com/ember-cli/ember-cli/issues/8164#issuecomment-621597177_
> v2 of the `yarn` CLI uses `-i` or `--interactive` rather than `--non-interactive`.

See [here in the documentation](https://yarnpkg.com/cli/add#details).

```
ember install ember-file-upload

🚧  Installing packages... This might take a couple of minutes.
Command failed: yarn add --dev ember-file-upload --non-interactive

Unknown Syntax Error: Command not found; did you mean one of:

  0. yarn add [--json] [-E,--exact] [-T,--tilde] [-C,--caret] [-D,--dev] [-P,--peer] [-O,--optional] [--prefer-dev] [-i,--interactive] [--cached] ...
  1. yarn add [--json] [-E,--exact] [-T,--tilde] [-C,--caret] [-D,--dev] [-P,--peer] [-O,--optional] [--prefer-dev] [-i,--interactive] [--cached] ...

While running add --dev ember-file-upload --non-interactive
```

```
ember-cli: 3.18.0
node: 14.3.0
v8: 8.1.307.31-node.33
uv: 1.37.0
zlib: 1.2.11
brotli: 1.0.7
ares: 1.16.0
modules: 83
nghttp2: 1.40.0
napi: 6
llhttp: 2.0.4
openssl: 1.1.1g
cldr: 37.0
icu: 67.1
tz: 2019c
unicode: 13.0
os: darwin x64
6.14.5
2.1.1
```

caassandra:
Closes #9276 

The new version of yarn is [non-interactive by default](https://yarnpkg.com/cli/add#details):
> If the `-i`,`--interactive` option is used (or if the `preferInteractive` settings is toggled on) the command will first try to check whether other workspaces in the project use the specified package and, if so, will offer to reuse them.

This breaks `ember install` when using v2. This PR adds a check to see if the user is using Yarn 2, and if they are, it doesn't add the `--non-interactive` flag.


lib/tasks/npm-task.js
```diff
@@ -131,10 +143,11 @@ class NpmTask extends Task {
 
     logger.info('yarn.lock found -> trying yarn');
     try {
-      await this.checkYarn();
+      const { yarnVersion } = await this.checkYarn();
 
       logger.info('yarn found -> using yarn');
       this.useYarn = true;
+      this.yarnVersion = yarnVersion;
```
rwjblue, line 150: I find it vaguely odd to set `this.yarnVersion` in two different ways. Can we refactor this a bit (possibly making `this.checkYarn` write the value itself) to avoid having to set the value twice based on control flow?

lib/tasks/npm-task.js
```diff
@@ -110,13 +118,17 @@ class NpmTask extends Task {
...
+        .catch((error) => {
+          if (error.code === 'ENOENT') {
+            throw new SilentError('Yarn could not be found.');
+          }
+
+          throw error;
+        })
+        .then(({ yarnVersion }) => {
+          this.yarnVersion = yarnVersion;
+        });
```
rwjblue, line 131: Would you mind tweaking this to use `async`/`await` while you are making changes? Not sure why we missed doing that as part of the other work (adding `async` to `findPackageManager`)...

I'm thinking something like:


```suggestion
      try {
        await this.checkYarn()
      } catch(error) {
        if (error.code === 'ENOENT') {
          throw new SilentError('Yarn could not be found.');
        }

        throw error;
      }
```

rwjblue:
Sorry for the delay here @caassandra, thank you for picking this up!

lib/tasks/npm-task.js
caassandra: Ah yeah looking back now it's hard to understand why I wrote it this way. I think I wanted to mirror the logic of returning an object of shape `{ xxxxVersion: version }` from `checkNpmVersion()`, but it also was necessary to set the `useYarn` flag.

Given that the only times `checkYarn()` is called is when `useYarn` is already `true` or when `yarn.lock` is found, I think it's fine to set the `useYarn` flag to `true` from inside `checkYarn()` right before returning. What I mean to say is that there is never a time that yarn is successfully found where we do not also want to use yarn.

The only exception I can think of is that someone has a project with a `yarn.lock` but is actually using `npm` to manage their packages, but I suspect that since yarn already warns about finding a `package-lock.json` & mixing package managers is pretty heavily discouraged, that's ok.

```js
try {
  const yarnVersionResult = await this.checkYarn();

  logger.info('yarn found -> using yarn');

  // if we get here, useYarn & yarnVersion are already set
  // this.useYarn = true;
  // this.yarnVersion = yarnVersion;

  // keeping this as a separate line from this.checkYarn() to allow logging
  return yarnVersionResult

} catch (_err) {
  logger.info('yarn not found -> using npm');
  return this.checkNpmVersion();
}
```

As a result, the only place that `this.yarnVersion` is set is right before the actual call to `yarn install`:

```js
if (this.useYarn) {
  this.yarnVersion = result.yarnVersion;
  let args = this.toYarnArgs(this.command, options);
  promise = this.yarn(args);
} else {
  let args = this.toNpmArgs(this.command, options);
  promise = this.npm(args);
// more stuff...
```

lib/tasks/npm-task.js
caassandra: I spent a while fixing the test stubs because of the refactoring of where the errors are getting thrown due to your other comment, and I think I also removed the `.then()` flow incidentally. Whatever happened, the only results for <kbd>⌘</kbd> + <kbd>F</kbd> `.then()` brings up the lines for the workaround for npm 5 and 6.

caassandra:
@rwjblue no prob! I'm sure there's a lot more work to do in getting full Yarn v2 compatibility with PnP and all that. I appreciate you taking time to review this PR! Lmk if there's more changes you'd like.

PR #9299, opened 2020-08-02
remove explicit `yarn install` in `.travis.yml`

kellyselden:
I accidentally screwed up the lockfile, and realized the yarn command is missing `--frozen-lockfile`. The default travis command for yarn is `yarn --frozen-lockfile` https://travis-ci.org/github/ember-vr/ember-vr-shopping/jobs/714192550#L511. Instead of trying to match what Travis is doing, I propose removing the line, and allowing Travis to handle it, just like it does for NPM (`npm ci`).


rwjblue:
Seems good to me.

Issue #8969, opened 2019-12-19
Sending intr (ctrl + c / interrupt) to ember serve doesn't fully terminate the app 


pichfl:
Sending <kbd>ctrl</kbd> + <kbd>c</kbd> to a running `ember serve` should terminate the server and running ember-cli instance, but for some reason it does not completely clean up after itself and leaves my command prompt in an unfinished state, making it necessary for me to press <kbd>return</kbd> to actually get back to my regular prompt.

```sh
source/project on 🌱 master [$] via ⬢ v12.13.1 took 3s
❯ yarn start
```
(Ember project runs as expected)

<kbd>ctrl</kbd> + <kbd>c</kbd>
```sh
cleaning up...
⠋ cleaning up⏎

source/project on 🌱 master [$] via ⬢ v12.13.1 took 3s
_
```
("❯ " character missing, no more autocompletion from my shell, etc)

I'm using fish, but the same thing happens on zsh and bash, both with [my current prompt](https://github.com/starship/starship) enabled and disabled.

It's really annoying and I wonder if it breaks more stuff beyond just me having to remember to do the extra key press which resets my prompt like that:

<kbd>↩</kbd>
```
source/project on 🌱 master [$] via ⬢ v12.13.1 took  49s


source/project on 🌱 master [$] via ⬢ v12.13.1 took  49s
❯

source/project on 🌱 master [$] via ⬢ v12.13.1 took  49s
❯
```

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.14.0
node: 12.4.0
v8: 7.4.288.27-node.18
uv: 1.29.1
zlib: 1.2.11
brotli: 1.0.7
ares: 1.15.0
modules: 72
nghttp2: 1.38.0
napi: 4
llhttp: 1.1.3
http_parser: 2.8.0
openssl: 1.1.1b
cldr: 35.1
icu: 64.2
tz: 2019a
unicode: 12.1
os: darwin x64
6.12.1
1.21.1
```

rwjblue:
@pichfl - Can you help us track this down by providing as close to step by step instructions on reproducing? My guess is that ora isn't being reset properly in all pathways, but a repro will really help.

I also use starship for my prompt atm, so _hopefully_ I'll be able to get this to reproduce.

runspired:
I've been experiencing this for the last few versions, probably should have reported it, but many of the apps I work on do crazy things so it took me a while to realize I hit it with new apps too.

[...]

PR #9300, opened 2020-08-05
Update init.js

chrism:
This is a simple fix to solve the issue outlined here:
https://github.com/ember-cli/ember-cli/issues/8969

@rwjblue recommended this.


rwjblue:
Sorry for the mega long delay here. I think we actually _want_ to keep saying `yarn start` or `npm start`, I'd really like to dig into what is leaving the terminal in a bad state. If it turns out to be `yarn` itself, we should fix it there (or start moving back to npm?). If it turns out to be us not clearing things up, but only when executed in specific ways: we should fix that...

PR #9329, opened 2020-09-11
ignore package.json of yarn workspace

### LINKED ISSUE #9328: Can not generate a new project if ember-cli is installed in yarn workspace
Ember CLI wrongly assumes being inside an Ember CLI project if `ember-cli` is installed in yarn workspace. When running `ember new <app-name>` and `ember addon <addon-name>` it throws `You cannot use the new/addon command inside an ember-cli project.`

## Use case

Installing `ember-cli` in workspace would have some benefits:

1. No need to install `ember-cli` globally or to use `npx` to create new ember projects.
2. Version used to generate new projects can be controlled per workspace and shared with colleagues.
3. Generating a new project using an `ember-cli` installed in workspace is as fast as if using a globally installed version. It's much faster than using `npx`.

## Reproduction

To reproduce create a yarn workspace and install `ember-cli` in it:

1.  `mkdir test-workspace && cd test-workspace/`
2. create `package.json` with the following content:
    ```json
    {
      "private": true,
      "workspaces": [
        "packages/*"
      ]
    }5
     ```
3.  `yarn add -W ember-cli`
4. `mkdir packages`

Running `ember new test-app` or `ember new test-addon` in `./packages` fails:

```
$ ember new test-project
You cannot use the new command inside an ember-cli project.

$ ember addon test-project
You cannot use the addon command inside an ember-cli project.
```

## Debug

The assertion is thrown here: https://github.com/ember-cli/ember-cli/blob/627934f91b2aa0e19b041fdb1b547873c1855793/lib/models/command.js#L279-L281

The `isWithinProject` method of `Command` is initialized with return value of `isEmberCLIProject` method of `Project`: https://github.com/ember-cli/ember-cli/blob/627934f91b2aa0e19b041fdb1b547873c1855793/lib/models/command.js#L86

The `isEmberCLIProject` method checks if `ember-cli` is in dependencies: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/lib/models/project.js#L187-L197

The dependencies are read by `dependencies` method: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/lib/models/project.js#L355-L374

As no argument is passed it uses `pkg` property, which is passed in when initializing the `Project` instance: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/lib/models/project.js#L27-L43

If I got the in-source comment right it should only be initialized using the static `closestSync` method: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/lib/models/project.js#L645-L679

That one uses the `findupPath` function to find a `package.json` in any parent directory: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/lib/models/project.js#L772-L779

`findupPath` throws if it does not find a `package.json`. But that should be the expected in most cases. :confused:

jelhan:
Fixes #9328

I wanted to write a test case but it seems as if the current test setup does not support testing these case: https://github.com/ember-cli/ember-cli/blob/066e9efc9034e8ddde57151ebf76d35398037c20/tests/helpers/ember.js#L78-L83 Is not having test coverage acceptable? If not it would be great if you can give me a hint how to test it.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -771,7 +771,7 @@ function ensureUI(_ui) {
 
 function findupPath(pathName) {
   let pkgPath = findup.sync('package.json', { cwd: pathName });
-  if (!pkgPath || isYarnWorkspacke(pkgPath)) {
+  if (!pkgPath || isYarnWorkspace(pkgPath)) {
     throw new NotFoundError(`No project found at or up from: \`${pathName}\``);
   }
 
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -778,7 +778,7 @@ function findupPath(pathName) {
   return path.dirname(pkgPath);
 }
 
-function isYarnWorkspacke(pkgPath) {
+function isYarnWorkspace(pkgPath) {
   let packageJson = JSON.parse(fs.readFileSync(pkgPath, { encoding: 'utf-8' }));
   return Object.keys(packageJson).includes('workspaces');
 }
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -771,11 +771,16 @@ function ensureUI(_ui) {
 
 function findupPath(pathName) {
   let pkgPath = findup.sync('package.json', { cwd: pathName });
-  if (!pkgPath) {
+  if (!pkgPath || isYarnWorkspacke(pkgPath)) {
     throw new NotFoundError(`No project found at or up from: \`${pathName}\``);
   }
 
   return path.dirname(pkgPath);
+}
+
+function isYarnWorkspacke(pkgPath) {
+  let packageJson = JSON.parse(fs.readFileSync(pkgPath, { encoding: 'utf-8' }));
+  return Object.keys(packageJson).includes('workspaces');
 }
 
 function isEmberCliProject(pkg) {
--- a/tests/helpers/fixturify-project.js
+++ b/tests/helpers/fixturify-project.js
@@ -21,6 +21,12 @@ function prepareAddon(addon) {
 }
 
 module.exports = class EmberCLIFixturifyProject extends FixturifyProject {
+  constructor() {
+    super(...arguments);
+
+    this.inRepoAddonLocation = 'lib';
+  }
+
   writeSync() {
     super.writeSync(...arguments);
     this._hasWrriten = true;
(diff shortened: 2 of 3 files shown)


jelhan:
Good catch @rwjblue!

rwjblue (commented):
My concern here is that this is actually a perfectly valid scenario. Imagine a project that has:

* ember app in the project root
* some "in repo addons" in `packages/*/package.json` that have dependencies

In that case (which seems pretty reasonable to me) you would possibly reach for adding `workspaces` to that root `package.json`, and this change would (if I'm understanding the logic correctly) make it **no longer match**.

🤔

kellyselden:
Could you do "has `workspaces` and not `devDeps->ember-cli"?

rwjblue:
@jelhan - I've added a commit showing how to test this functionality (adding a basic test for the "normal" setup, and another test for the scenario that I mentioned just above that is failing with this PR's changes). At the moment, the two tests (one passes one fails) show the issue I'm describing: there is no difference from the case of "hoisting ember-cli to a shared root workspaces package.json, but all projects are in another directory" and the "the root of the repo is an ember-cli app that also uses workspaces to ensure the dependencies for its in-repo addons are installed.

The task here I guess is to get that test passing, and I think in order to do that we need to come up with another reasonable heuristic for identifying it as a project...

rwjblue:
> Could you do "has workspaces and not `devDeps->ember-cli"?

@kellyselden - Yea, that was what I was originally thinking but re-reading the steps over in https://github.com/ember-cli/ember-cli/issues/9328 suggests that @jelhan is actually trying to add `ember-cli` to that root `package.json` (likely to hoist and ensure all of the sub-packages get the shared ember-cli installation).

jelhan:
> @jelhan is actually trying to add `ember-cli` to that root `package.json` (likely to hoist and ensure all of the sub-packages get the shared ember-cli installation).

Exactly. It was motivated by https://github.com/jelhan/ember-addon-tests. This would allow me to add a `emberCliVersion` option to `new TestProject(options)`.

It would allow consumers to write a more predictable test suite. Currently the test suite is executed against the Ember CLI version installed globally. That means that the test suite may pass on one computed but fail on another. Very bad developer experience.

I tried to use `npx` instead to have a predictable Ember CLI version. But the performance is very bad compared to Ember CLI installed either globally or in workspace.

I know that there are some work-a-rounds:

1. I could rely on the bug described in #9331. But that doesn't sound like a stable long-term solution.
2. I could install the desired `ember-cli` version to another temporary folder and use that binary. But that would mean installing the same `ember-cli` package twice. That wouldn't be good for performance. Prepopulating `node_modules` of workspace with `node_modules` created by `ember-cli` standalone installation might help with performance. But add more complexity.

bertdeblock:

PR #9343, opened 2020-09-25
Fix yuidoc for private APIs

jenweber:
Changes:
- Mark all methods private in `package-cache` except methods below the line that says `// PUBLIC API BELOW`
- Mark all methods private in `hardware-info`
- Mark all methods private in `command-generator`
- Create a fake class for some private API test helpers in acceptance.js (`PrivateTestHelpers`)

To see what this looks like as rendered docs, run `npx yuidoc . --server` from the root of the ember-cli repo.

### DIFF
--- a/lib/models/hardware-info.js
+++ b/lib/models/hardware-info.js
@@ -233,6 +233,7 @@ const hwinfo = {
    * Inidcates whether the host is running on battery power.  This can cause
    * performance degredation.
    *
+   * @private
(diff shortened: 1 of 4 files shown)


jenweber:
cc @mansona here's the PR for the yuidoc changes we discussed today.

mansona (approved):
from what I understand of yuidoc this looks good to me 🎉 😂

PR #9363, opened 2020-10-18
use travis build matrix/expansion instead of include

kellyselden:
I learned more about Travis CI stages and wanted to share. Travis has a default stage called "test". This is were all build matrix expansion jobs go. If we move all the `include` env jobs to the real env section, we can allow Travis to expand them into the "test" stage. You can see how it looks [here](https://github.com/kellyselden/ember-a11y-blueprint/pull/58).

Even though this is technically more "correct", it's a little more verbose, and you are stuck with the name "test" instead of how we preferred to name the stages. I'll leave this here for debate on whether it's a good change or not.

### DIFF
--- a/blueprints/addon/files/.travis.yml
+++ b/blueprints/addon/files/.travis.yml
@@ -22,11 +22,26 @@ env:
     # See https://git.io/vdao3 for details.
     - JOBS=1
 
+  jobs:
+    # we recommend new addons test the current and previous LTS
+    # as well as latest stable release (bonus points to beta/canary)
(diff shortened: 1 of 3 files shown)


rwjblue (commented):
FWIW, I did know that you could do this when we set up the original stages configuration. I don't think it's major deal to go either way, but I vaguely preferred naming the stage (which is why I went that way originally).

I'm not sure what changing would get us overall, any specific benefit that you can see that would justify the overall churn?

Turbo87:
I assume this can be closed?! :)

PR #9366, opened 2020-10-19
Move ember-cli-build.js to tests/dummy for addons

astronomersiva:
Another attempt at addressing #7486 based on @rwjblue's suggestions during the recent Ember Contributors workshop. 

This is still a WIP and I just wanted to get some feedback on this before I go ahead and change the blueprints and add tests.


lib/commands/build.js
```diff
@@ -20,6 +20,7 @@ module.exports = Command.extend({
     { name: 'watch', type: Boolean, default: false, aliases: ['w'] },
     { name: 'watcher', type: String },
     { name: 'suppress-sizes', type: Boolean, default: false },
+    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/' },
```
rwjblue, line 23: Should probably the full path to the file (we want the path to the file, not the path to the directory that contains the file).

```suggestion
    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/ember-cli-build.js' },
```

lib/commands/serve.js
```diff
@@ -88,6 +88,7 @@ module.exports = Command.extend({
       description: 'Specify the certificate to use for SSL.',
     },
     { name: 'path', type: 'Path', description: 'Reuse an existing build at given path.' },
+    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/' },
```
rwjblue, line 91: Same as above 
```suggestion
    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/ember-cli-build.js' },
```

lib/commands/test.js
```diff
@@ -76,6 +76,7 @@ module.exports = Command.extend({
     { name: 'path', type: 'Path', description: 'Reuse an existing build at given path.' },
     { name: 'query', type: String, description: 'A query string to append to the test page URL.' },
     { name: 'output-path', type: 'Path', aliases: ['o'] },
+    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/' },
```
rwjblue, line 79:
```suggestion
    { name: 'build-file-path', type: 'Path', default: 'tests/dummy/ember-cli-build.js' },
```

lib/models/builder.js
```diff
@@ -44,9 +44,14 @@ class Builder extends CoreObject {
    */
   readBuildFile(path) {
```
rwjblue, line 45: Let's refactor this function to not take in the path argument here, it is more confusing overall than having the method itself control all of the logic (especially when the thing we get isn't the path to the file, its a directory path).

lib/models/builder.js
```diff
@@ -44,9 +44,14 @@ class Builder extends CoreObject {
...
+    let project = this.project;
+    let projectRoot = project.root;
+
+    let buildFile = findBuildFile('ember-cli-build.js', { projectRoot, path });
     if (buildFile) {
-      return buildFile({ project: this.project });
+      return buildFile({ project });
     }
 
     throw new SilentError('No ember-cli-build.js found.');
```
rwjblue, line 57: I also don't really think that the `findBuildFile` utility is very helpful at all.

I think we should do this (modulo any silly mistakes I've made while prototyping...):

```js
  readBuildFile() {
    // Load the build file
    let buildFile;
    if (this.buildFilePath) {
      buildFile = require(this.buildFilePath);
    } else {
      let buildFilePath = findUp.sync('ember-cli-build.js', { cwd: this.project.root });
      
      buildFile = require(buildFilePath);
    }

    if (typeof buildFile !== 'function') {
      throw new SilentError('Could not find a valid ember-cli-build.js file.');
    }

    process.chdir(projectRoot);

    return buildFile({ project });
  }
```

And delete that utility file.

rwjblue (changes requested):
Thanks for working on this!!! I left a few comments (hopefully helpful) inline, but we also need a few other things in order to land this:

* tests confirming that an arbitrary file can be used (e.g. `--build-file-path some/path/lol.js` should use `lol.js`)
* updates to the existing `--help` fixture tests to add the new option
* tests confirming that the fallback behavior to root level `ember-cli-build.js` works

astronomersiva:
Thanks for the feedback @rwjblue!

I've done the following:

* Refactored the `readBuildFile` method to handle all the logic and removed the `findBuildFile` utility function.
* Updated the addon blueprints to generate the `ember-cli-build.js` at `tests/dummy` from now on.
* Added a test to ensure that backward compatibility is maintained in addons.
* Added a test to ensure that any file can be used along with the `build-file-path` argument.

I haven't updated the `--help` fixtures as I am not sure about how to update them. Is there anything like `jest --updateSnapshot` that I can use here for this?

blueprints/addon/index.js
```diff
@@ -166,7 +166,12 @@ module.exports = {
     let addonFilesPath = this.filesPath(this.options);
     let addonFiles = walkSync(addonFilesPath);
 
-    return uniq(appFiles.concat(addonFiles));
+    let files = [].concat(
+      appFiles.filter((f) => f !== 'ember-cli-build.js'),
+      addonFiles
+    );
```
rwjblue, line 172: Hmm, why do you need this? I would have expected that the `fileMap` just below here would have been updated to add:

```
'^ember-cli-build.js': 'tests/dummy/ember-cli-build.js'
```

lib/models/builder.js
```diff
@@ -38,23 +38,37 @@ class Builder extends CoreObject {
...
 
-    let buildFile = findBuildFile('ember-cli-build.js', { projectRoot, path });
-    if (buildFile) {
-      return buildFile({ project });
+    let defaultBuildFilePath = project.isEmberCLIAddon() ? 'tests/dummy/ember-cli-build.js' : 'ember-cli-build.js';
+    let buildFilePath = path.resolve(projectRoot, this.buildFilePath || defaultBuildFilePath);
+    // backward compatibility for addons with `ember-cli-build.js` at project root.
+    if (!fs.existsSync(buildFilePath)) {
+      buildFilePath = findUp.sync('ember-cli-build.js', { cwd: projectRoot });
+    }
```
rwjblue, line 55: Let's simplify, and ensure that the "magic" defaulting / find-up behavior only happens when a `--build-file-path` was **not** passed.

tests/fixtures/addon/ember-cli-build-at-root/ember-cli-build.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
```
rwjblue, line 1: Can we test this with our custom fixturify-project test helper (instead of making more static fixture files)?

rwjblue:
> I haven't updated the --help fixtures as I am not sure about how to update them. Is there anything like jest --updateSnapshot that I can use here for this?

No, but I wish there was (or we'd migrate to using jest here).

lib/models/builder.js
rwjblue: Something like:

```js
let { buildFilePath } = this;
if (!buildFilePath) {
  let defaultBuildFilePath = project.isEmberCLIAddon()
    ? 'tests/dummy/ember-cli-build.js'
    : 'ember-cli-build.js';
  buildFilePath = path.join(project.root, defaultBuildFilePath);
  
  if (!fs.existsSync(buildFilePath)) {
    // backward compatibility for addons with `ember-cli-build.js` at project root.
    buildFilePath = findUp.sync('ember-cli-build.js', { cwd: project.root });
  }
}
```

bertdeblock:
Is this still something we want to pursue, knowing this won't be an issue with v2 addons?

locks:
I think @bertdeblock is correct and further time investment should go into ushering in the new embroider era 😁

PR #9383, opened 2020-11-04
[feature] support h2 when starting https server

gabrielcsapo:
# H2! 

> There is currently a limitation with the built in http2 module and express@4. This will be fixed in express@5 and can be migrated off of spdy. (spdy has been maintained and is being downloaded over 7 million times a day!)

## What do we get from h2?

We get a lot out of the box! https://developers.google.com/web/fundamentals/performance/http2 The biggest take away is

```
he primary goals for HTTP/2 are to reduce latency by enabling full request and response multiplexing, minimize protocol overhead via efficient compression of HTTP header fields, and add support for request prioritization and server push. To implement these requirements, there is a large supporting cast of other protocol enhancements, such as new flow control, error handling, and upgrade mechanisms, but these are the most important features that every web developer should understand and leverage in their applications.
```

And importantly:

```
HTTP/2 does not modify the application semantics of HTTP in any way. All the core concepts, such as HTTP methods, status codes, URIs, and header fields, remain in place.
```

## What happens when we use it? 

> All tests are being run against https://github.com/travis-ci/travis-web.

### http 1.1

> http 1.1 (no caching) (~3.45s)
 
Screen Shot 2020-11-04 at 2 50 43 PM

> http 1.1 (with caching) (~2.70s)

Screen Shot 2020-11-04 at 2 50 49 PM

### http2

> http2 (no caching) (~2.3s) `33% improvement`

Screen Shot 2020-11-04 at 2 49 10 PM

> http2 (caching) (~1.50s) `44% improvement`

Screen Shot 2020-11-04 at 2 52 30 PM


gabrielcsapo:
@rwjblue what are the next steps for this change?

package.json
```diff
@@ -164,7 +165,7 @@
     "yuidocjs": "0.10.2"
   },
   "engines": {
-    "node": "10.* || >= 12"
+    "node": ">= 10.10.0 || >= 12"
```
rwjblue, line 168: Hmm, in general we can't don't change engines 🤔. Why is this needed?

lib/tasks/server/express-server.js
```diff
@@ -16,7 +16,9 @@ class ExpressServerTask extends Task {
     this.emitter = new EventEmitter();
     this.express = this.express || require('express');
     this.http = this.http || require('http');
-    this.https = this.https || require('https');
+    // This should be migrated to require('http2') when we have migrated to express@5 which supports the built-in http2 bindings
```
rwjblue, line 19: Why can't we require express@5?

tests/unit/tasks/server/express-server-test.js
```diff
@@ -228,6 +229,43 @@ describe('express-server', function () {
         });
     });
 
+    // TODO: this is not supported by supertest https://github.com/visionmedia/supertest/pull/494 should fix that
```
rwjblue, line 232: We might want to consider a switch, that PR is pretty old and doesn't really have any traction as far as I can tell...

package.json
gabrielcsapo: to use the http2 module for our tests we needed to bump this as it wasn't released to public usage until 10.10.0

lib/tasks/server/express-server.js
gabrielcsapo: express@5 is still in beta, do we want to use the beta?

tests/unit/tasks/server/express-server-test.js
gabrielcsapo: will follow up with the pr to see what is happening there.

gabrielcsapo: followed up here https://github.com/visionmedia/supertest/pull/691.

lib/tasks/server/express-server.js
grconrad: express@5 has been in alpha for years (still at alpha.8, published 9 months ago)

package.json
rwjblue: But we don't actually test locally with Node < 10.10, so it shouldn't really matter

lib/tasks/server/express-server.js
rwjblue: ugh, ok, that is a tad frustrating 😝

locks:
Hey @gabrielcsapo. I was thinking of closing this PR, as getting up to speed seems more daunting than redoing it? And personally I think it would be more interesting to work towards an RFC for the proposed migration here https://github.com/ember-cli/ember-cli/issues/9676 instead.
Let us know if you agree, or if I should reopen this PR!

PR #9387, opened 2020-11-23
Add support for specifying a path for the generate command

NullVoxPopuli:
This is needed for this project: 
  - https://github.com/NullVoxPopuli/ember-codemod-blueprint

Which has [these remaining TODOs](https://github.com/NullVoxPopuli/ember-codemod-blueprint/pull/4/files) before the idea can even work
- [ ] add an API to allow accessing what files are generated in `afterInstall` (not the focus of _this_ PR)
- [x] add a path in place of the blueprint name, similar to how `new` can use a path to a local blueprint, 
   rather than rely on the name resolution system currently in place.
   this is useful, for wanting to chain blueprint executions, as is the focus of ember-codemod-blueprint
   (what this PR is focused on)

I'm open to advice on approach here, but those two bullets are what I need to be able to build more straightforward and extensible blueprint in user-space.

I'm kind of thinking about blueprints being like this at a high level (after the above work):
```
generateBaseBlueprint('ember-source/blueprints/component')
 |> renameToTypeScript
 |> addTypeArgument
 |> etc
```
actual api in https://github.com/NullVoxPopuli/ember-codemod-blueprint -- main difference is no pipes, and you'd want to work with specific file paths, rather than whatever the blueprint output is.
So a snippet from ember-codemod-blueprint,

```ts
module.exports = codemodBlueprint({
  upstream: 'ember-source/blueprints/component',
 
  transform(options, helpers) {
    let { transformScript, transformTemplate } = helpers;
    let { generatedFiles } = options;

    for (let fileInfo of generatedFiles) {
      
      if (fileInfo.name.includes('app/components') && fileInfo.name.endsWith('.js')) {
        await transformScript(
          scriptPath, 
          transformA, 
          transformB, 
          transformC
       );
     }
     // etc, and more transforms for the other files
    }
});
```

This approach should work for app-generation as well, but the `transform` hook may be _much_ bigger, depending on how different your team's default app is from the base blueprint

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1318,6 +1318,23 @@ let Blueprint = CoreObject.extend({
 Blueprint.lookup = function (name, options) {
   options = options || {};
 
+  if (name.includes(path.sep)) {
+    let blueprintPath = path.resolve(name);
+    let isNameAPath = Boolean(blueprintPath);
+
+    if (isNameAPath) {
+      if (Blueprint._existsSync(blueprintPath)) {
+        return Blueprint.load(blueprintPath);
+      }
+
+      if (!options.ignoreMissing) {
+        throw new SilentError(`Unknown blueprint: ${name}`);
+      }
+
+      return;
+    }
+  }
+
   let lookupPaths = generateLookupPaths(options.paths);
 
   let lookupPath;
--- a/lib/commands/generate.js
+++ b/lib/commands/generate.js
@@ -9,6 +9,11 @@ const _ = require('ember-cli-lodash-subset');
 const EOL = require('os').EOL;
 const SilentError = require('silent-error');
 
+const UNKNOWN_BLUEPRINT_ERROR =
+  'The `ember generate` command requires a ' +
+  'blueprint name to be specified. ' +
+  'For more details, use `ember help`';
+
 module.exports = Command.extend({
   name: 'generate',
   description: 'Generates new code from blueprints.',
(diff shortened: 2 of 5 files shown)


mehulkar:
Some high level feedback: instead of a second arg, you could use the same one? `ember g named-blueprint` and `ember g path/to/blueprint` could be nicer

NullVoxPopuli:
Just tried this in a monorepo, and it was succesful:
```bash
❯ ember g ../../../node_modules/ember-source/blueprints/component my-component-via-path
installing component
  create addon/components/my-component-via-path/index.hbs
  skip addon/components/my-component-via-path/index.js
  tip to add a class, run `ember generate component-class my-component-via-path`
installing component-test
  create tests/integration/components/my-component-via-path-test.js
installing component-addon
  create app/components/my-component-via-path.js
  ```

**Comment by NullVoxPopuli** [comment]
Related: https://github.com/emberjs/rfcs/pull/699

File: lib/models/blueprint.js [hunk 1]
```diff
@@ -1279,6 +1279,23 @@ let Blueprint = CoreObject.extend({
 Blueprint.lookup = function (name, options) {
   options = options || {};
 
+  if (name.includes(path.sep)) {
+    let blueprintPath = path.resolve(name);
```
lifeart, line 1283: could we add check for relative/absolute paths, is cwd = curren project dir?

lifeart (approved):

ef4 (approved):
This seems OK to land, we just need to rebase and make sure tests still pass.

PR #9402, opened 2020-12-02
Prevent build cycles when app is within a watched dir

ef4:
Embroider v2 packages can put their source code wherever they want to (as they are "just regular node packages"). That means when we're interactively rebuilding them, we need to watch their root. But if a v2 addon contains its own test suite, and that test suite emits a `dist` folder, we will end up watching our own output, and cause an infinite build loop.

This fixes that by always ignoring the build's own output directory.  This depends on https://github.com/broccolijs/broccoli/pull/474 and will be updated to stop pointing at my fork once we have a broccoli release.

I'm unsure where you want to put test coverage for this. It will necessarily be covered by embroider's own tests, and here we are really just threading one option down to the layers that actually implement this behavior, so IMO I don't think an expensive end-to-end test here is necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "babel-plugin-module-resolver": "^4.0.0",
     "bower-config": "^1.4.3",
     "bower-endpoint-parser": "0.2.2",
-    "broccoli": "^3.4.2",
+    "broccoli": "^3.5.0",
     "broccoli-amd-funnel": "^2.0.1",
     "broccoli-babel-transpiler": "^7.8.0",
     "broccoli-builder": "^0.18.14",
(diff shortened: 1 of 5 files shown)


ef4:
CI is failing for bogus reasons at the moment.

package.json
```diff
@@ -43,8 +43,7 @@
     "babel-plugin-module-resolver": "^4.0.0",
     "bower-config": "^1.4.3",
     "bower-endpoint-parser": "0.2.2",
-    "broccoli": "^3.4.2",
-    "broccoli-amd-funnel": "^2.0.1",
+    "broccoli": "git+https://github.com/broccolijs/broccoli.git#0c56d222634097542f571ff8a79532f8580e3c0c",
```
rwjblue, line 46: Ready for update to https://github.com/broccolijs/broccoli/releases/tag/v3.5.0

rwjblue (changes requested):

package.json
ef4: Updated to the released version, and fixed a mistake where I dropped a line from package.json unintentionally.

rwjblue (approved):

PR #9391, opened 2020-11-30
Add prettier to blueprint

bmish:
Implements this RFC: https://github.com/emberjs/rfcs/blob/master/text/0628-prettier.md (https://github.com/emberjs/rfcs/pull/628)

Changes included:
* Installs prettier dev dependencies (package.json)
* Updates lint scripts (package.json)
* Adds `plugin:prettier/recommended` to .eslintrc.js
* Creates new file `.prettierrc.js`
* Adds `.prettierrc.js` to Node files override in .eslintrc.js
* Adds `.eslintcache` to .eslintiginore
* Creates new file `.prettierignore` (duplicate of .eslintignore)
* Fixes prettier violations in blueprints and test fixtures

TODO:

- [x] update tests
- [x] manually test this
- [x] ensure ember-cli generated code has no prettier violations
- [ ] update `ember generate` to run `lint:fix`  to fix prettier violations in external blueprints? ([RFC section](https://github.com/emberjs/rfcs/blob/master/text/0628-prettier.md#blueprint-changes))

PR #9407, opened 2020-12-08
Ensure correct lint scripts run when running test script

bmish:
The `test` script can just run the top-level `lint` script, instead of running `lint:*` which will erroneously run the `lint:fix` script too.

The bug was introduced in #9391.


rwjblue (approved):
Eeck, thanks for catching this!

PR #9410, opened 2020-12-10
gitignore .eslintcache

simonihmig:
#9391 added the `-cache` flag to the eslint script, so we should ignore the cache file.

### DIFF
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -10,6 +10,7 @@
 /.editorconfig
 /.ember-cli
 /.env*
+/.eslintcache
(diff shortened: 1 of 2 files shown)


rwjblue (commented):
Good catch! Can you also add it to https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/npmignore ?

simonihmig:
> Can you also add it to https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/npmignore ?

Ah yes, sure, done!

rwjblue:
Thanks @simonihmig!

PR #9413, opened 2020-12-11
Deprecation Compliance: Adding declarations to EmberENV

mehulkar:
Work in progress, opening up for initial thoughts

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -232,20 +232,12 @@ class Project {
     @return {Object}     Merged configuration object
    */
   config(env) {
-    try {
-      throw new Error();
-    } catch (e) {
-      // console.log(`CALLING config() from: ${e.stack.split('\n')[4].trim()}`);
-    }
     let _env = env === undefined ? process.env.EMBER_ENV : env;
     let key = `${this.configPath()}|${_env}`;
     let config = this.configCache.get(key);
     if (config === undefined) {
-      console.log('no config in cacheffor ', _env);
       config = this.configWithoutCache(_env);
       this.configCache.set(key, config);
-    } else {
-      console.log('found config in cache', key);
     }
     return _.cloneDeep(config);
   }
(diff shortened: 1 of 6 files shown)


lib/models/project.js
```diff
@@ -255,7 +263,24 @@ class Project {
...
+
+      this.addons.forEach((x) => {
+        if (x.pkg.ember && x.pkg.ember.deprecationCompliance) {
+          console.log('✅', 'deprecation compliance: ', x.name, x.pkg.version, x.pkg.ember.deprecationCompliance);
+          deprecationCompliance.addons[x.name] = x.pkg.ember.deprecationCompliance;
+        } else {
+          console.log('💩', 'deprecation compliance: ', x.name, x.pkg.version, );
+        }
+      });
+      mergedConfig.EmberENV.DEPRECATION_COMPLIANCE = deprecationCompliance;
```
mehulkar, line 280: tried this out w/ a sample app and looks like `this.addons` is already the canonical list and handles nested dependencies, etc, but I'm not 100% sure.

Seems like just passing the whole thing into EmberEnv is ok? @rwjblue super initial pass, but let me know if you had anything else in mind.

lib/models/project.js
```diff
@@ -255,12 +255,28 @@ class Project {
...
+    deprecationCompliance.addons = {};
+
+    this.addons.forEach((x) => {
+      if (x.pkg.ember && x.pkg.ember.deprecationCompliance) {
+        deprecationCompliance.addons[x.name] = x.pkg.ember.deprecationCompliance;
+      }
+    });
+
+    return deprecationCompliance;
+  }
```
rwjblue, line 278: The addon hierarchy is a tree structure. As it stands, this will only grab from the top level addons. In order to traverse all addons you'll probably want to recurse.

Something like:

```js
  getDeprecationComplianceConfig() {
    const deprecationCompliance = {
      self: this.pkg.ember.deprecationCompliance,
      addons: {},
    };

    this._gatherDeprecationComplianceFromAddon(this.project, deprecationCompliance);

    return deprecationCompliance;
  }

  _gatherDeprecationComplianceFromAddon(addonOrProject, deprecationCompliance) {
    addonOrProject.addons.forEach((x) => {
      if (x.pkg.ember && x.pkg.ember.deprecationCompliance) {
        deprecationCompliance.addons[x.name] = x.pkg.ember.deprecationCompliance;
      }
    });
  }
```

(Assuming that we want to merge things into a flat list)

rwjblue:
👋 @mehulkar how's it going?

mehulkar:
Will pick this back up after 4.0 release, since all the deprecated features in 3.x will be removed anyway then.

mehulkar:
Yeah we can close it 😭 , sorry I was never able to close it out!

PR #9430, opened 2021-01-14
'live-reload-js-url' and 'live-reload-options' configuration from command-line

lifeart:
related issue: https://github.com/ember-cli/ember-cli-inject-live-reload/issues/150


lifeart:
looks like fixtures update requred @toranb @rwjblue any easy way to update it?
I could poke with https://github.com/suchipi/chai-jest-snapshot if needed

ElenaSpb (approved):

sandstrom:
Looks great! 🏅 

We've had to work around this by having our scripts write to `.ember-cli`, which isn't ideal. This would be much better.

lib/commands/serve.js
```diff
@@ -58,6 +58,20 @@ module.exports = Command.extend({
     { name: 'watcher', type: String, default: 'events', aliases: ['w'] },
     { name: 'live-reload', type: Boolean, default: true, aliases: ['lr'] },
     { name: 'live-reload-host', type: String, aliases: ['lrh'], description: 'Defaults to host' },
+    {
+      name: 'live-reload-js-url',
+      type: String,
+      aliases: ['lrju'],
```
rwjblue, line 64: Let's remove the alias, I don't think this option is common enough to need an alias.

lib/commands/serve.js
```diff
@@ -58,6 +58,20 @@ module.exports = Command.extend({
...
+      description:
+        'The absolute URL used to load livereload.js. If specified, this overrides the liveReloadPort option.',
+    },
+    {
+      name: 'live-reload-options',
+      type: String,
+      aliases: ['lro'],
+      description:
+        'Stringified JSON object for LiveReload options. LiveReload supports a number of options for configuring websocket communication, including https, host, port, and others. See advanced example below.',
```
rwjblue, line 73: I think the last part of this sentence should be removed ("see below for more info" doesn't seem right). Maybe replace it with a git.io link to the README where it is listed?

rwjblue (changes requested):
Overall seems good, thanks for working on it!

Some todos:

* needs tests 
* needs to add an error (early, when command options are parsed) if incompatible options are specified

lifeart:
@rwjblue could you help with fixtures updated approach? Should I somehow regenrate existing one? (is so - I have no idea how to update it), or move touched checks to `chai-jest-snapshot`?

rwjblue:
What fixtures?

lifeart:
> What fixtures?
for example: tests\fixtures\help\help.txt

image


CI is failing on it, but I don't know how to update it in a right way

rwjblue:
Ya, I'd love to make those snapshots. Updating them is mega painful everytime 😩. If you have the time/energy, I'd love a separate PR that migrates those to use snapshots (then we can update those new snapshots in this PR).

lifeart:
@rwjblue looks like it's alive now

bertdeblock:

PR #9431, opened 2021-01-14
replace fixtures with jest-like snapshots

lifeart:
spillover from https://github.com/ember-cli/ember-cli/pull/9430

cc @rwjblue


sandstrom:
Hello @lifeart! Just checking in to hear what's left in this one?

I noticed that this seems to be a blocker for https://github.com/ember-cli/ember-cli/pull/9430 (or maybe it isn't?), and that enhancement (live-reload-js-url as CLI param) would be great for us 😄

lifeart:
Hi @sandstrom! Can't get snapshots diffing working fine locally & on ci (locally it's works fine).
I added you to my fork to have ability to push into this branch in case if you have some ideas how to fix it.

sandstrom:
thanks for the update @lifeart 

I'm afraid I don't think I have any ideas on how to fix it.

Is it possible to decouple https://github.com/ember-cli/ember-cli/pull/9430 from this PR? (basically to advance #9430 even though this is still pending)

lifeart:
@sandstrom we have to update existing `snapshot` for #9430, without snapshots update, it looks little bit tricky too :/ there is no guide on it.

lifeart:
I'm thinking about binary data format for snapshots to get CI working fine

sandstrom:
@lifeart Just curious if you found a solution to the problem in this PR?

The updates you added in https://github.com/ember-cli/ember-cli/pull/9430 are great, and I'd love to see it merged 😄 

Taking a wild guess here, because I don't know much about jest snapshots, but are any of these links helpful?

https://github.com/facebook/jest/blob/main/e2e/__tests__/console.test.ts
https://github.com/ai/print-snapshots

lifeart:
Hi @sandstrom! Nope, looks like because of "colorization" we have snapshot serialization issues here, and it's also has different serialization based on platform (windows/*nix).

I'm thinking about binary snapshots, but it's sounds like quite a big overhead here..

rwjblue:
@lifeart - Can you separate the functionality from the migration of these specific test fixtures? It would be nice to start using snapshot for some of the other areas where we currently use fixtures, and this PR looks like it would already allow that (since those cases aren't subject to the colorization / serialization problems).

lifeart:
strange CI fails 
image

cc @rwjblue

rwjblue:
@lifeart - The failures you shared in the screenshot were unrelated to your changes (I fixed them in https://github.com/ember-cli/ember-cli/pull/9643).

lifeart:
great, thanks @rwjblue!

PR #9436, opened 2021-01-26
Enable Embroider test scenario for addons

thoov:
Leverage [@embroider/test-setup](https://github.com/embroider-build/embroider/tree/master/packages/test-setup) to add both `embroider-safe` and `embroider-optimized` scenarios for addons test matrix.


blueprints/addon/additional-dev-dependencies.json
```diff
@@ -1,5 +1,6 @@
 {
   "devDependencies": {
+    "@embroider/test-setup": "0.36.0",
```
rwjblue, line 3: We should probably allow a range here

```suggestion
    "@embroider/test-setup": "^0.36.0",
```

rwjblue (changes requested):
Generally looks good, only one minor tweak.

PR #9446, opened 2021-02-02
Update blueprint for index.html

MelSumner:
Updates the blueprint for index.html to be more accessible by default, as inspired by the ember-template-lint rule that covers this issue, [no-invalid-meta](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-meta.md)

### DIFF
--- a/blueprints/app/files/app/index.html
+++ b/blueprints/app/files/app/index.html
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title><%= namespace %></title>
     <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=yes">
 
     {{content-for "head"}}
 


rwjblue (commented):
Based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name#standard_metadata_names_defined_in_other_specifications I don't think we need to do this. 

> If set to no, the user is not able to zoom in the webpage. The default is yes. Browser settings can ignore this rule, and iOS10+ ignores it by default.

Since `user-scalable=yes` is the default value, we don't need to manually specify it (and AFAICT manually specifying provides no additional value).

rwjblue:
I just double checked, the ember-template-lint rule checks for explicit `no` values:

https://github.com/ember-template-lint/ember-template-lint/blob/8ee7ac81f6f999b5ebb625cb34dc94b6ebbfb0ed/lib/rules/no-invalid-meta.js#L76-L84

I've also created https://github.com/ember-template-lint/ember-template-lint/pull/1735 to show that the default ember-cli value does pass the `no-invalid-meta` rule.

PR #9448, opened 2021-02-15
move release instructions to wiki

kellyselden:
This was discussed on a call, but I want to make sure everyone is still on board with this change.


rwjblue (commented):
Hmm, I'm not sure this actually works well. A number of the things that are instructed on are _in_ the actual branch being released. If any of those scripts change the docs in the wiki are going to be out of sync... 🤔

kellyselden:
Oh I didn't know they were tailored to their branches. I just though they were out of sync.

Something interesting is the notes mention switching branches, but then you get a new set of notes.

rwjblue:
> Oh I didn't know they were tailored to their branches.

The thing I was meaning was that the release process has you run commands for scripts _in the repo_. Those scripts (their arguments, etc) are tied to the branch that you are on.

PR #9454, opened 2021-02-23
Add --embroider as an option for new and init

thoov:
Enables --embroider along with EMBER_CLI_EMBROIDER to default the build to embroider.

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -35,7 +35,7 @@ module.exports = Command.extend({
       type: String,
       description: 'Sets the base human language of the application via index.html',
     },
-    { name: 'embroider', type: Boolean, default: false },
+    { name: 'embroider', type: Boolean, default: false, description: 'Enables the build system to use Embroider' },
   ],
 
   anonymousOptions: ['<glob-pattern>'],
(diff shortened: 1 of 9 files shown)


lib/commands/new.js
```diff
@@ -34,6 +34,7 @@ module.exports = Command.extend({
       type: String,
       description: 'Sets the base human language of the application via index.html',
     },
+    { name: 'embroider', type: Boolean, default: false },
```
rwjblue, line 37: Can you add a description here too?

lib/commands/init.js
```diff
@@ -35,6 +35,7 @@ module.exports = Command.extend({
       type: String,
       description: 'Sets the base human language of the application via index.html',
     },
+    { name: 'embroider', type: Boolean, default: false },
```
rwjblue, line 38: Can you add a description for the flag?

rwjblue (changes requested):
Overall 👍👍, I think we need to add a description for the flag (so the help output has some conceptual explanation of what its for).

thoov:
@rwjblue should be good now

rwjblue (approved):

PR #9455, opened 2021-02-23
Move embroider experiment checks to each test

thoov:
As we work to enable each test to work under embroider moving the guards down a level will allow us to introduce smaller PRs enabling a few tests at a time.


tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -43,6 +40,9 @@ describe('Acceptance: brocfile-smoke-test', function () {
   });
 
   it('a custom EmberENV in config/environment.js is used for window.EmberENV', async function () {
+    if (isExperimentEnabled('EMBROIDER')) {
+      this.skip();
+    }
```
rwjblue, line 45: I think these will have to return, right? Otherwise we still run to completion (e.g. running the tests) only to mark them as skipped later.

Am I mising something?

thoov:

rwjblue (approved):

Issue #8070, opened 2018-09-26
Update default .eslintignore to ensure JS dotfiles get linted


rwjblue:
By default `eslint` will not lint any files starting with a `.` (including `.eslintrc.js`), in order to fix this we need to add `!.*` to the `.eslintignore` file.

See https://github.com/eslint/eslint/issues/10341 for details.

elwayman02:
Ironically, the default `.eslintrc.js` file generated by ember-cli fails at linting now that this is "fixed" in v3.5.0...

```
/Users/axle/Projects/ember-is-visible/.eslintrc.js
   1:1   error  'module' is not defined   no-undef
  47:32  error  'require' is not defined  no-undef
```

Thoughts @rwjblue?

kellyselden:
Fixed via #8127

PR #9456, opened 2021-02-24
Add `.*/` to eslint ignore

chancancode:
`!.*` was added to ensure js-based config files are linted (#8070)

However, this was probably too broad. An unintended consequence of that change was that the `.git` directory is now linted by eslint. The easiest way to observe this is to create a branch with `.js` in the name:

```shell
git checkout -b "foo.js"
yarn lint:js
```

This will now error since git created a file called `foo.js` in `.git` to maintain the metadata for this branch, but the file does not contain valid JavaScript.

In addition to these error scenarios, it causes unnecessary file scans. There are probably other caches by other tools other than git that could cause problems here as well.

### DIFF
--- a/blueprints/app/files/.eslintignore
+++ b/blueprints/app/files/.eslintignore
@@ -13,6 +13,7 @@
 # misc
 /coverage/
 !.*
+.*/
 .eslintcache
 
 # ember-try


kellyselden (approved):

wagenet (approved):

rwjblue:
Thank you!

elwayman02:
@chancancode should this also be added to `.prettierignore`, or is this only an issue for `eslint` specifically?

PR #9469, opened 2021-03-02
[FEATURE] Lint fix after blueprint generation

### LINKED ISSUE #9429: Automatically run `lint:fix` after `ember init` and `ember generate`
The [Prettier RFC](https://github.com/emberjs/rfcs/blob/master/text/0628-prettier.md) includes the following:

> In order to ensure blueprint output follows each individual projects custom stylistic linting settings as much as possible. The following will be updated to run lint:fix (when available) after their existing functionality:
> 
> * ember generate
> * ember init
> * ember-cli-update

This is specifically to avoid every addon that provides blueprints from failing linting after generation. 

Related to:

* https://github.com/emberjs/ember.js/issues/19333
* https://github.com/emberjs/data/issues/7418

rpemberton:
Fixes #9429

This adds auto lint fixing for files generated by `ember generate` and `ember init`.

This is my first PR on ember-cli so looking forward to any feedback and happy to make improvements to it.


lib/commands/generate.js
```diff
@@ -22,6 +22,7 @@ module.exports = Command.extend({
     { name: 'classic', type: Boolean, default: false, aliases: ['c'] },
     { name: 'dummy', type: Boolean, default: false, aliases: ['dum', 'id'] },
     { name: 'in-repo-addon', type: String, default: null, aliases: ['in-repo', 'ir'] },
+    { name: 'skip-lint-fix', type: Boolean, default: false, aliases: ['slf'] },
```
rwjblue, line 25: Let's remove the alias (this is uncommon enough that I don't think forcing folks to use the longer but more descriptive `--skip-lint-fix` flag).

Also, what do you think about doing `--lint-fix` as the flag, then folks could do `--no-lint-fix` "for free"?

lib/tasks/generate-from-blueprint.js
```diff
@@ -12,6 +14,19 @@ class GenerateTask extends Task {
...
 
   async run(options) {
+    let fileInfos = await this.createBlueprints(options);
+
+    if (!options.skipLintFix) {
+      try {
+        let fileDisplayPaths = fileInfos.map(({ displayPath }) => displayPath);
+        await lintFix.run(fileDisplayPaths);
+      } catch (error) {
+        logger.error('Lint fix failed');
```
rwjblue, line 24: I think we should include a little more context about the error (maybe interpolate `error.message`)

lib/tasks/install-blueprint.js
```diff
@@ -45,7 +46,16 @@ class InstallBlueprintTask extends Task {
...
     logger.info(`Installing blueprint into "${target}" ...`);
-    return blueprint.install(installOptions);
+    let fileInfos = await blueprint.install(installOptions);
+
+    if (!options.skipLintFix) {
+      try {
+        let fileDisplayPaths = fileInfos.map(({ displayPath }) => displayPath);
+        await lintFix.run(fileDisplayPaths);
+      } catch (error) {
+        logger.error('Lint fix failed');
```
rwjblue, line 56: Add a little more context RE: the error here

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,32 @@
...
+
+const execa = require('execa');
+
+function eslint(filePaths) {
+  return execa('eslint', [...filePaths, '--fix'], { cwd: process.cwd(), preferLocal: true });
+}
+
+function emberTemplateLint(filePaths) {
+  return execa('ember-template-lint', [...filePaths, '--fix'], { cwd: process.cwd(), preferLocal: true });
+}
```
rwjblue, line 11: I was thinking that we should check for `lint:fix` and run it (that will support more things than just eslint/ember-template-lint if they've configured them)

rwjblue (changes requested):
Overall this is great, thanks for working on it!!

I've left a few small inline comments to work on, but please ping me when you are ready for another review.

lib/utilities/lint-fix.js
rpemberton: I considered the same thing however there were a couple of things that steered me away from this approach:

- It could affect files beyond what's generated. I wouldn't expect `ember generate` to lint fix unrelated files. 
- The performance difference. I expect that linting the whole app is likely to be slower than targeting only what's changed.

Though you make a good point about `lint:fix` providing support for more than just the default lint tools. Would `lint:fix` be your preference?

rpemberton: Thanks for the feedback, and for reviewing it so quickly 👍

rwjblue: > Would lint:fix be your preference?

Yes.

RE: the two main concerns:

> It could affect files beyond what's generated. I wouldn't expect ember generate to lint fix unrelated files.

IMHO we have to think of these as idempotent. It _shouldn't_ change unrelated files because you _shouldn't_ be in a "failing linting" state.

> The performance difference. I expect that linting the whole app is likely to be slower than targeting only what's changed.

Ya, this is definitely true though I think the usage of `--cache` by default in `lint:js` should mitigate.

kellyselden: Could we combine the two? Pseudo code `execa('npm', ['run', 'lint:fix', '--', ...filePaths])`. I'm not sure how the npm-run-all package passes args.

rpemberton: I looked into this however as `lint:fix` calls the `lint:*:fix` scripts and they already use a `.`, e.g. `eslint . --fix`, I think that will always guarantee that everything will be lint fixed.

I tested this by running `./node_modules/.bin/eslint . --fix ./app/router.js` and the additional path had no effect.

lib/commands/generate.js
rpemberton: Good suggestions, done 👍

lib/tasks/generate-from-blueprint.js
rpemberton: done

lib/tasks/install-blueprint.js
rpemberton: done

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const chalk = require('chalk');
+const execa = require('execa');
+
+async function run() {
+  let lintFixScriptName = `lint:fix`;
+  let cwd = process.cwd();
+  let npmRunResult = await execa('npm', ['run'], { cwd });
```
rwjblue, line 9: We should probably use either npm or yarn based on the normal "is yarn.lock present" trigger that we normally use

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,28 @@
...
+  let lintFixScriptRegex = new RegExp(`^  ${lintFixScriptName}$`, 'm');
+  let hasLintFixScript = lintFixScriptRegex.test(npmRunResult.stdout);
```
rwjblue, line 11: Hmm, why do we need to run `npm run` just to figure out if they have a lint:fix script? Couldn't we directly read the JSON file to see?

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,28 @@
...
+
+  if (!hasLintFixScript) {
+    let warningMessage = `'${lintFixScriptName}' not found in package.json scripts`;
+
+    console.warn(chalk.yellow('Lint fix skipped:', warningMessage));
+
+    return Promise.reject({
+      shortMessage: warningMessage,
+    });
```
rwjblue, line 20: Inside an async context (like here where you are already within an async function execution) you shouldn't return a rejected promise you should throw directly.

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  await execa('npm', ['run', lintFixScriptName], { cwd });
```
rwjblue, line 23: Same here RE: using yarn or npm

rwjblue (changes requested):

lib/tasks/generate-from-blueprint.js
```diff
@@ -12,6 +14,18 @@ class GenerateTask extends Task {
   }
 
   async run(options) {
+    await this.createBlueprints(options);
+
+    if (options.lintFix) {
+      try {
+        await lintFix.run();
+      } catch (error) {
+        logger.warn('Lint fix failed: %o', error);
```
rwjblue, line 23: Can you check if we can use `this.ui` here (it is a `console-ui` instance)?

lib/tasks/install-blueprint.js
```diff
@@ -45,7 +46,15 @@ class InstallBlueprintTask extends Task {
...
     let blueprint = await this._resolveBlueprint(blueprintOption);
     logger.info(`Installing blueprint into "${target}" ...`);
-    return blueprint.install(installOptions);
+    await blueprint.install(installOptions);
+
+    if (options.lintFix) {
+      try {
+        await lintFix.run();
+      } catch (error) {
+        logger.warn('Lint fix failed: %o', error);
```
rwjblue, line 55: Can you check if we can use `this.ui` here (it is a `console-ui` instance)?

rwjblue (commented):
Looks really good, thanks for keeping at it!

 I'm not 100% sure why CI is failing, but the actual code changes here look really good. I think you might need to rebase (we recently merged a few other PRs).

lib/utilities/lint-fix.js
```diff
@@ -0,0 +1,32 @@
...
+  let packageJson = fs.readJsonSync('package.json');
+
+  let hasLintFixScript = !!packageJson.scripts[lintFixScriptName];
+
+  if (!hasLintFixScript) {
+    ui.writeWarnLine(
+      'Lint fix skipped: "lint:fix" not found in package.json scripts. New files might not pass linting.'
+    );
+
+    throw new Error('"lint:fix" not found in package.json scripts');
```
rwjblue: Just to clarify, we expect this to be absorbed by the `try`/`catch` up in `lib/tasks/install-blueprint.js` and `lib/tasks/generate-from-blueprint.js` right? 

We definitely don't want to bubble out an error just because folks don't have `lint:fix` defined...

rwjblue (approved):
Just one clarification, but I think this is good to go.

lib/utilities/lint-fix.js
rpemberton: That's correct, however I wasn't keen on it so I replaced the error throw with a `return`.

rwjblue:
I restarted CI (macOS had a test timeout).

rpemberton:
Thanks, yeh this test appears to be a little flaky:
`Acceptance: ember new ember new npm blueprint with old version`
https://github.com/ember-cli/ember-cli/pull/9469/checks?check_run_id=2057118456#step:5:92

rwjblue:
Thank you!

charlesfries:
What's the point of linting right after the files are generated? Why not just update the source files to use the default Ember project linting/formatting options?

This change *really* slows down the generate command in a large app, especially if you chain multiple generate commands together.

bertdeblock:
The reason for this is mentioned in the linked issue as well as in the corresponding RFC itself. Note that you can also disable this functionality globally via your `.ember-cli` config file.

Z-Zhao:
> The reason for this is mentioned in the linked issue as well as in the corresponding RFC itself. Note that you can also disable this functionality globally via your `.ember-cli` config file.

Can anyone suggest how could I disable this lint:fix globally via the .ember-cli config file. As mentioned above?

charlesfries:
@Z-Zhao 

```js
// .ember-cli

{
  "lintFix": false
}

```

PR #9476, opened 2021-03-08
move fixturify-project to devDeps

ef4:
I think fixturify-project is only used in the tests, so we don't need it in deps.


ef4:
This revealed another bug. We were publishing `tests/helpers/*` to NPM.

rwjblue:
That wasn't a bug, we intended to provide these helpers to consumers.

ef4:
Oh, ok. TIL.

rwjblue:
yeah 😕 sorry, we (Stef and I) have wanted to expose them as public API with a nice `require('ember-cli').testHelpers` kinda API for a while, just never got around to it 😭

PR #9480, opened 2021-03-09
Refactor getPort to only check required port

Cartmanishere:
Portfinder library gets port as follows -- 

1. `getPort({:port 0})`:
    - Returns a random port.

2. `getPort({:port <non-zero>})`:
    - Returns the first open port between `<non-zero>` and upper limit (which is `40000`)

3. `getPort({:port <lower-limit>, :stopPort <upper-limit>})`
    - Returns a first open port between the range `<lower-limit>`  and `<upper-limit>`

In the ember script, we should only check for the specified port and not check for a range.
Hence, for a non-zero required port, we check for an open port in the range -- `[portNum, portNum]`
So if this port is not available, then `getPort` will give an exception.

If the requested port is less than `1024`, then the reason for port being unavailable can be that --
- Port is already in use
- You do not have sufficient permissions available

So we set appropriate error message based on the requested port.

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -107,15 +107,25 @@ module.exports = Command.extend({
   },
 
   async _checkExpressPort(commandOptions) {
-    let foundPort = await getPort({ port: commandOptions.port, host: commandOptions.host });
-    if (commandOptions.port !== foundPort && commandOptions.port !== 0) {
-      let message = `Port ${commandOptions.port} is already in use.`;
-      return Promise.reject(new SilentError(message));
+    let portOptions = { port: commandOptions.port, host: commandOptions.host };
+    if (commandOptions.port !== 0) {
+      portOptions.stopPort = commandOptions.port;
     }
-
-    // otherwise, our found port is good
-    commandOptions.port = foundPort;
-    commandOptions.liveReloadPort = commandOptions.liveReloadPort || commandOptions.port;
-    return commandOptions;
+    try {
+      let foundPort = await getPort(portOptions);
+      commandOptions.port = foundPort;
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
Generally seems good to me, has some linting issues to clean up.

Cartmanishere:
Fixed the linting issues. Thanks.

lib/commands/serve.js
```diff
@@ -107,15 +107,25 @@ module.exports = Command.extend({
...
+    } catch (err) {
+      let message;
+      if (commandOptions.port === 0) {
+        message = `No open port found above ${commandOptions.port}`;
+      } else if (commandOptions.port < 1024) {
+        message = `Port ${commandOptions.port} is already in use or you do not have permissions to use this port.`;
+      } else {
+        message = `Port ${commandOptions.port} is already in use.`;
+      }
       return Promise.reject(new SilentError(message));
```
rwjblue, line 128: Apologies for not spotting this initially (and I know that you didn't introduce this), but we shouldn't return a rejected promise here from within an async function. We should just throw instead.

```suggestion
      throw new SilentError(message);
```

Cartmanishere: Sure, I'll make this change and update.

Cartmanishere: @rwjblue I have made the change but I would like to understand the rationale behind throwing an error here. My understanding is that return promise would have the same behavior as throwing an error.

rwjblue: Yep! The resulting behavior is the same (the promise that invoking the outer function returns would reject). It really is just about writing idiomatic code to me.

rwjblue:
Thank you!

PR #9483, opened 2021-03-09
Remove EMBROIDER guards from smoke-test-slow

thoov:
Enables embroider experiment to run on the smoke tests as they now pass. Keeping the skip of the fingerprinting tests under Embroider.


thoov:

rwjblue (approved):

PR #9392, opened 2020-11-30
Add eslint-plugin-qunit to blueprint

bmish:
[eslint-plugin-qunit](https://github.com/platinumazure/eslint-plugin-qunit) a great plugin for enforcing QUnit best practices in tests, and I consider it to be the most useful/relevant linting plugin that we haven't added to the blueprint yet. I use in many of my Ember apps/addons.

Fixes #9390.

TODO:

- [x] Wait until fix for https://github.com/platinumazure/eslint-plugin-qunit/issues/75 fixed is published since this is a pretty severe bug in some recommended rules
- [x] Wait for RFC approval: https://github.com/emberjs/rfcs/pull/702

PR #9484, opened 2021-03-15
Update eslint-plugin-qunit to v6 in blueprint

bmish:
https://github.com/platinumazure/eslint-plugin-qunit/blob/master/CHANGELOG.md#600
* The `two` config has been merged into the `recommended` config so this simplifies things for us even more.
* Note: this version of the dependency drops support for Node 13.

Follow-up to #9392.


rwjblue (approved):

rwjblue:
Thank you!

PR #9487, opened 2021-03-17
Add support for creating a single addon instance per bundle root (which enables dramatically reducing the total number of addon instances)

davecombs:
This creates a cache of addon instances (one per "bundle host", i.e. project or lazy engine) that meet criteria that they are simple enough and have no behaviors that would prevent them from being a 'singleton', i.e. that we could replace other instances of them (and, importantly, their child addons) with a proxy to the original addon instance. Each proxy contains a pointer to the singleton instance and a local copy of the 'app' and 'parent' properties. It defines 'set' and 'get' to proxy through most property or method requests to the singleton copy. In applications with a large number of references to a given addon that itself has a large number of dependent addons, this can save huge amounts of space (in the largest applications, on the order of several GB of heap and process memory). It does, however, mean that traversals up the 'parent' tree will only work correctly when the addon is being called in the one location that it is not proxied.

As part of creating tests for the new feature, made a significant number of changes to the `fixturify-project' utility in tests/helpers/ to make it more flexible and allow creating fixtures for engines (both in-repo and in node_modules). Those changes are then used in the per-bundle-addon-cache tests.


lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +421,190 @@ class PackageInfo {
...
+   * roots.
+   *
+   * @method findBundleRoot
+   * @param {Project|Addon} parent
+   * @return {Object} the object in the 'parent' chain that is a bundle root.
+   */
+  findBundleRoot(parent) {
+    let tmp = parent;
+    while (tmp) {
+      if (tmp.isBundleRoot) {
```
rwjblue, line 472: Let's add a small utility that checks lazyLoading (not `isBundleRoot`) here.

Something like:

```js
if (isBundleRoot(tmp)) {
```

lib/models/project.js
```diff
@@ -96,6 +97,8 @@ class Project {
     if (this.packageInfoCache.hasErrors()) {
       this.packageInfoCache.showErrors();
     }
+
+    this.singletonAddonCache = new SingletonAddonCache(this);
```
rwjblue, line 101: This should be underscored:

```suggestion
    this._singletonAddonCache = new SingletonAddonCache(this);
```

lib/models/singleton-addon-cache/cache-entry.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+/**
+ * Class for the entries in the SingletonAddonCache. The cache itself is keyed
+ * on the addon's PackageInfo.root.
+ *
+ * @class CacheEntry
+ */
+class CacheEntry {
```
rwjblue, line 9: Hmm, do we really need this class? Seems like we can handle this shaping concern (built or not)  at callsites.

lib/models/singleton-addon-cache/index.js
```diff
@@ -0,0 +1,95 @@
...
+ * Class that stores entries by packageInfo.root that are a singleton instance.
+ * Note that there's a separate instance of this cache for each 'bundle root'
+ * (i.e. the Project and any lazy engine). Each then contains all the
+ * singleton instances within itself (the Project automatically contains
+ * all the bundle roots other than itself.)
+ *
+ * @public
+ * @class SingletonAddonCache
+ */
+class SingletonAddonCache {
```
rwjblue, line 15: Should this be "bundle root addon cache"?

lib/models/singleton-addon-cache/cache-entry.js
```diff
@@ -0,0 +1,35 @@
...
+  constructor(instance) {
```
rwjblue, line 10: Instance of what? Addon instance?

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +421,190 @@ class PackageInfo {
...
+      if (cacheEntry && cacheEntry.isBuilt) {
+        // Return a proxy for the already-set-up cache entry.
+        let proxy = getAddonProxy(cacheEntry, parent);
+        return proxy;
+      } else {
+        // The cache entry may or may not exist, but the instance isn't set up yet.
+        // Create a new instance, install it in the bundleRoot's cache (add/update
+        // the cache entry) and return the new instance.
+        addon = this.constructAddonInstance(parent, project);
+        if (ctor.isBundleRoot) {
```
brendenpalmer, line 584: this should use the aforementioned util that @rwjblue recommended here: https://github.com/ember-cli/ember-cli/pull/9487/files#r598920189

lib/models/singleton-addon-cache/cache-entry.js
davecombs: resolving as outdated.

lib/models/singleton-addon-cache/cache-entry.js
davecombs: Good idea - removing in next version.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +421,190 @@ class PackageInfo {
...
+    // Find some parent, if there is one, that defines the 'isBundleRoot' property and that
+    // returns a truthy value. 'Project' is always the ultimate bundle.
+
+    // An optimization that lazy engines does that we can do as well. If the bundle
+    // root is not the project, but the project does or is going to have an instance
+    // of this singleton in its own cache, then use the project cache.
+    let bundleRoot = this.findBundleRoot(parent);
+    let bundledProjectAddons = project.getPackageInfo().getBundledProjectAddons();
+
+    if (bundleRoot === project || (bundleRoot !== project && !bundledProjectAddons.has(this))) {
```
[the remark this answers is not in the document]
rwjblue: Let's simplify this conditional a bit (likely by making local variables with good names).

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+
+      // Luckily for us, both types of bundle root (Project and lazy engine) have a 
+      // _packageInfo field, just what we need!
+      let bundleRootPkgInfo = bundleRoot._packageInfo;
+
+      let cacheEntry = project.perBundleAddonCache.getPerBundleCacheEntry(bundleRootPkgInfo, this);
+
+      if (cacheEntry && cacheEntry.isBuilt) {
+        // Return a proxy for the already-set-up cache entry.
+        console.log(`BR ${bundleRootPkgInfo.realPath}: PROXY: ${this.realPath}`);
```
rwjblue, line 584: Note: ✂️

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+        let addonProxy = getAddonProxy(cacheEntry, parent);
+        return addonProxy;
+      } else {
+        // The cache entry may or may not exist, but the instance isn't set up yet.
+        // Create a new instance, install it in the bundleRoot's cache (add/update
+        // the cache entry) and return the new instance.
+        console.log(`BR ${bundleRootPkgInfo.realPath}: NEWENTRY: ${this.realPath}`);
```
rwjblue, line 591: Note: 🔪

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+        addon = this.constructAddonInstance(parent, project);
+        project.perBundleAddonCache.addPerBundleInstance(bundleRootPkgInfo, this, addon);
+
+        if (addon.initializeAddons) {
+          addon.initializeAddons();  // XXX based on addon.setupRegistry, I think this is basically a no-op by this point.
```
rwjblue, line 596: We still have to call it (because addon's could decide to have custom constructors and not have the `setupRegistry` in constructor issue). Should tweak the comment but keep the actual implementation.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+        }
+
+        return addon;
+      }
+    } else {
+      // The bundleRoot is not the project but the project bundles the addon too, so
+      // the cache entry needs to go in the project's cache. Create/use an empty
+      // cache entry and return a proxy to it. The project will set the instance later.
+      let cacheEntry = project.perBundleRootAddonCache.addPerBundleInstance(project._packageInfo, this, null);
+      console.log(`BR ${project._packageInfo.realPath}: EMPTY: ${this.realPath} (ALSOPROJECT)`);
```
rwjblue, line 608: Note: 💀

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+        return parent;
+      }
+
+      if (property === 'app') {
+        return _app;
+      }
+
+      if (property === 'isProxy') {
+        return true;
+      }
```
rwjblue, line 37: Who / what checks this property?

lib/models/per-bundle-addon-cache/cache-entry.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+/**
+ * Class for the entries in the SingletonAddonCache. The cache itself is keyed
```
rwjblue, line 4: Tweak this comment (we renamed away from the singleton verbiage)

lib/models/per-bundle-addon-cache/cache-entry.js
```diff
@@ -0,0 +1,35 @@
...
+ * on the addon's PackageInfo.root.
+ *
+ * @class CacheEntry
+ */
+class CacheEntry {
```
rwjblue, line 9: This class doesn't obviously provide value over using a properly shaped POJO. 🤔

lib/models/per-bundle-addon-cache/index copy.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
```
rwjblue, line 1: This file is odd (`index copy.js`), maybe it is a duplicate or artifact of some refactoring?

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * For use with per-bundle addon caching, is a given object an instance of a Project?
+   * 
+   * @public
+   * @method isProject
+   * @param {Object} obj 
+   * @returns {Boolean}
+   */
+  static isProject(obj) {
+    return obj?.constructor?.name === 'Project' || false;
```
rwjblue, line 180: Can't use optional chaining, also it seems pretty odd to check the constructor like this. Why wouldn't we use the methods on the project instance?

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,220 @@
...
+   * @method isLazyEngine 
+   * @param {Object} obj 
+   * @returns {Boolean} true if the object is a lazy engine instance or a constructor
+   * function/class for a lazy engine. 
+   */
+  static isLazyEngine(obj) {
+    let val = (obj?.lazyLoading?.enabled === true ||
+      obj?.options?.lazyLoading?.enabled === true ||
+      obj?.prototype?.lazyLoading?.enabled === true ||
+      obj?.prototype?.options?.lazyLoading?.enabled === true);
```
rwjblue, line 197: optional chaining will have to go...

rwjblue (changes requested):

lib/models/singleton-addon-cache/cache-entry.js
davecombs: Actually, we do need some sort of 'wrapper' around the instance value (though we don't need the 'isBuilt' method) because we have to have an object for a proxy to point at, potentially before an instance actually exists. In the case where an addon instance is referenced by both a lazy engine and the project, the project's one takes precedence. But if we encounter the lazy-engine dependency first, we need the engine to point to a proxy and then the project to fill in the proxy instance later when we get to that dependency processing.

lib/models/project.js
davecombs: I'll change the name to _perBundleAddonCache, but otherwise agree.

lib/models/singleton-addon-cache/cache-entry.js
davecombs: actually, ended up calling it cacheableAddonInstance and changing the internal name to _addonInstance.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+   * @param {} parent the addon/project that is to be the direct parent of the
+   * addon instance created here
+   * @param {*} project the project that is to contain this addon instance
+   * @return {Object} the constructed instance of the addon
+   */
+  getAddonInstance(parent, project) {
+    let ctor = this.getAddonConstructor();
+    let addon;
+
+    if (!SUPPORT_ADDON_INSTANCE_CACHING || !ctor.allowCachingPerBundle) {
```
brendenpalmer, line 549: will the `ctor.allowCachingPerBundle` conditional always work? eg, if my addon entry point looks something like:

```js
module.exports = {
  name: 'foo',
  allowCachingPerBundle: true
};
```
ultimately this will create a constructor that extends from `lib/models/addon.js`, but `ctor.allowCachingPerBundle` will be `undefined` AFAICT; we should also check `ctor.prototype.allowCachingPerBundle`

lib/models/instrumentation.js
```diff
@@ -60,6 +60,10 @@ class Instrumentation {
 
     this.instrumentations = {
       init: options.initInstrumentation,
+      initAddons: {
+        token: null,
+        node: null,
+      },
```
rwjblue, line 66: ✂️

lib/models/instrumentation.js
```diff
@@ -227,6 +244,8 @@ class Instrumentation {
     let filename = `instrumentation.${name}`;
     if (name === 'build') {
       filename += `.${this.instrumentations.build.count}`;
+    } else if (name === 'initAddons') {
+      filename += `.${this.project.name().replace(/[^\w]/g, '')}`;
```
rwjblue, line 248: 🔪

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +483,119 @@ class PackageInfo {
...
+   * @param {} parent the addon/project that is to be the direct parent of the
+   * addon instance created here
+   * @param {*} project the project that is to contain this addon instance
+   * @return {Object} the constructed instance of the addon
+   */
+  getAddonInstance(parent, project) {
+    let ctor = this.getAddonConstructor();
+    let addon;
+
+    if (!SUPPORT_ADDON_INSTANCE_CACHING || !ctor.isSingleton) {
```
rwjblue, line 542: Probably still want to change `isSingleton` to the other name I mention

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +483,119 @@ class PackageInfo {
...
+        addon.addons = [];
+      }
+
+      return addon;
+    }
+
+    let perBundleAddonCache = project._perBundleAddonCache;
+    let bundleRoot = this.findBundleRoot(parent);
+
+    if (bundleRoot === project || (bundleRoot !== project && !perBundleAddonCache.bundledProjectAddonPkgInfos.has(this))) {
```
rwjblue, line 559: Let's simplify this down into either a utility / helper or a named variable

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +483,119 @@ class PackageInfo {
...
+        addon.initializeAddons();
+      } else {
+        addon.addons = [];
+      }
+
+      return addon;
+    }
+
+    let perBundleAddonCache = project._perBundleAddonCache;
+    let bundleRoot = this.findBundleRoot(parent);
```
rwjblue, line 557: Can we tweak this function to actually let us remove some of these low level details (from below), which should make the actual flow in this method much clearer.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +483,119 @@ class PackageInfo {
...
+
+    if (bundleRoot === project || (bundleRoot !== project && !perBundleAddonCache.bundledProjectAddonPkgInfos.has(this))) {
+      // the instance belongs to its bundle root, whether it's the project or not.
+
+      // Luckily for us, both types of bundle root (Project and lazy engine) have a 
+      // _packageInfo field, just what we need!
+      let bundleRootPkgInfo = bundleRoot._packageInfo;
+
+      let cacheEntry = perBundleAddonCache.getPerBundleCacheEntry(bundleRootPkgInfo, this);
```
rwjblue, line 566: Let's make `getPerBundleCacheEntry` handle returning a valid cache entry in all cases (which would either create or return a proxied cache entry automatically).

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+      // trap addons & don't return the real addons array; this is done for primarily
+      // two reasons: (1) `ember-engines` dedupe logic is stateful, so this will just
+      // completely break, and (2) with all dependencies declared explicitly, the build
+      // is _extremely_ slow (eg, it takes over 40 minutes!)
+      if (property === 'addons') {
+        return [];
+      }
+
+      // allow access to `instance`
+      if (property === 'instance') {
```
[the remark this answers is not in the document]
rwjblue: YA, let's use a private symbol.

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      if (targetCacheEntry.isBuilt) {
+        if (property !== 'constructor' && typeof targetCacheEntry.instance[property] === 'function') {
+          // we don't want the proxy to be the context of these methods, so wrap
+          // the original methods to maintain the correct context
+          return function _originalAddonPropMethodWrapper() {
+            return targetCacheEntry.instance[property](...arguments);
+          };
+        }
```
rwjblue, line 65: Chatted with @brendenpalmer and @davecombs, the reason for this is somewhat subtle and I'd like to make sure that the comment explains it. Specifically, if we fall through to the `Reflect.get` just below the `this` context of the function when invoked is **the proxy** not the original instance (so its local state is incorrect).

Annnnyyyyyywwwwwaaaaayyyyssssss, lets update the comment to make that clear.

lib/models/per-bundle-addon-cache/cache-entry.js
rwjblue: Chated with @davecombs about this, he is working to move it inline (will still likely use a class though)

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+  /**
+     * Get the list of addon PackageInfo objects that are a dependency of this project,
+     * directly or transitively. Do not check any dependency that is itself a bundle root.
+     * This is used as part of the per-bundle addon caching optimization.
+     *
+     * @method getBundledAddonPackageInfos
+     * @return {Set} a Set of {PackageInfo} objects.
+     */
+  setupBundledProjectAddonPackageInfos() {
+    debugger;
```
rwjblue, line 65: 🕳️

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+  /**
+  * Find all lazy engine packageInfos in the project. We cache only a single copy
+  * of each, by NAME, not by PATH.
+  * 
+  * @public
+  * @method setupBundleRoots
+  * @param {Project} project the project that we use to get the PackageInfoCache
+  */
+  setupLazyEngineCache() {
+    debugger;
```
rwjblue, line 102: 🎯

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+
+        lazyEnginePkgInfos.set(pkgInfo.name, pkgInfo);
+      }
+
+      let addonPackageList = pkgInfo.discoverAddonAddons();
+      addonPackageList.forEach((pkgInfo) => findAddons(pkgInfo));
+    }
+
+    let addonPackageList = this.project._packageInfo.discoverProjectAddons();
+    addonPackageList.forEach((pkgInfo) => findAddons(pkgInfo));
```
rwjblue, line 123:
```suggestion
    addonPackageList.forEach(findAddons);
```

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+      if (pkgInfo.isLazyEngine()) {
+        if (lazyEnginePkgInfos.has(pkgInfo.name)) {
+          return;
+        }
+
+        lazyEnginePkgInfos.set(pkgInfo.name, pkgInfo);
+      }
+
+      let addonPackageList = pkgInfo.discoverAddonAddons();
+      addonPackageList.forEach((pkgInfo) => findAddons(pkgInfo));
```
rwjblue, line 119:
```suggestion
      addonPackageList.forEach(findAddons);
```

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+
+    let addonPackageList = this.project._packageInfo.discoverProjectAddons();
+    addonPackageList.forEach((pkgInfo) => findAddons(pkgInfo));
+
+    //   engineName: xxx
+    //   engineInstance: yyy (initially null, filled in as we instantiate an engine
+    //   addonInstanceCache: zzz (the cache of addon instances and proxies bundled with this lazy engine))
+    let cache = new Map();
+    lazyEnginePkgInfos.forEach((pkgInfo) => {
+      cache.set(pkgInfo.name, new Object({engineName: pkgInfo.name, engineInstance: null, addonInstanceCache: new Map()}));
```
rwjblue, line 130:
```suggestion
      cache.set(pkgInfo.name, {engineName: pkgInfo.name, engineInstance: null, addonInstanceCache: new Map()});
```

lib/models/project.js
```diff
@@ -14,9 +14,14 @@ const versionUtils = require('../utilities/version-utils');
...
-const PackageInfoCache = require('../models/package-info-cache');
+const PackageInfoCache = require('./package-info-cache');
+const PerBundleAddonCache = require('./per-bundle-addon-cache');
 
-const instantiateAddons = require('../models/instantiate-addons');
+const instantiateAddons = require('./instantiate-addons');
+
+// Support for per-bundle addon caching is opt OUT. Caching any particular
+// addon class is currently opt IN.
+const SUPPORT_ADDON_INSTANCE_CACHING = process.env.EMBER_CLI_ADDON_INSTANCE_CACHING !== 'false';
```
rwjblue, line 24: Let's make `per-bundle-addon-cache/index.js` the "source of truth" for this value, and import it from there.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,276 @@
...
+   * For use with per-bundle addon caching, indicate if a given object is an instance 
+   * or a constructor function/Class for a lazy engine.
+   * 
+   * @public
+   * @method isLazyEngine 
+   * @param {Object} obj 
+   * @returns {Boolean} true if the object is a lazy engine instance or a constructor
+   * function/class for a lazy engine. 
+   */
+  static isLazyEngine(obj) {
```
rwjblue, line 249: I think this can be deleted (since you extracted to a stand alone utility).

lib/utilities/is-project.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function isProject(obj) {
+  return obj.constructor && obj.constructor.name === 'Project';
```
rwjblue, line 4: I do want to change this to something else (not relying on Function.name), _maybe_ using:

https://github.com/ember-cli/ember-cli/blob/b24b73b388934796ca915ca665b48a27c857199b/lib/models/project.js#L195

Or `obj instanceof Project`?

lib/utilities/is-project.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function isProject(obj) {
```
rwjblue, line 3: also mentioned by @brendenpalmer, but lets move to directly checking if `obj === this.project` (the project the per-addon-bundle-cache was created for).

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +421,190 @@ class PackageInfo {
...
+   * @param {} parent the addon/project that is to be the direct parent of the
+   * addon instance created here
+   * @param {*} project the project that is to contain this addon instance
+   * @return {Object} the constructed instance of the addon
+   */
+  getAddonInstance(parent, project) {
+    let ctor = this.getAddonConstructor();
+    let addon;
+
+    if (!isSupportingSingletons || !ctor.isSingleton) {
```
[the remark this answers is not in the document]
rwjblue: The original source location here is changed (so this shows as outdated), but the change requested here is still not quite applied yet.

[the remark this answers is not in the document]
davecombs: I just manually added the change to my code. Will get in in the next push.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: For Brenden's second comment - this is the part that is affected by the redesign of what's in PerBundleAddonCache. The intention there is that when a lazy engine is being created, the PBAC will see it's a lazy engine, see if there is already a cache entry (in the cache-by-name I mentioned earlier), and if so, just return a proxy to that instance. The first time a given lazy engine (by name) is requested, it will be created as a real instance and that cache-by-name entry will be created. (So lazy-engine instances behave like other addons do - one "real" copy with real parent, everybody else a Proxy).

lib/models/package-info-cache/package-info.js
```diff
@@ -220,10 +227,50 @@ class PackageInfo {
...
   }
 
+  /**
+   * Indicate if this packageInfo is for a project. Should be called only after the project
+   * has been loaded.
+   * @returns {Boolean} true if this packageInfo is for a project, false otherwise.
+   */
+  isProject() {
+    return !!this.project;
+  }
```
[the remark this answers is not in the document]
davecombs: I may not actually need this method any more. Holding the comment open until that's resolved.

lib/cli/cli.js
```diff
@@ -76,6 +77,7 @@ class CLI {
       });
 
     this.packageInfoCache = new PackageInfoCache(this.ui);
+    this.perBundleAddonCache = new PerBundleAddonCache();
```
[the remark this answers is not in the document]
davecombs: It's now actually tied to the project, and the PBAC has more functionality than previously.

lib/models/instantiate-addons.js
```diff
@@ -58,38 +59,18 @@ function instantiateAddons(parent, project, addonPackages) {
         addonInitializationNode: true,
       });
 
-      let start = Date.now();
-
       let pkgInfo = parent.packageInfoCache.getEntry(addonInfo.path);
 
       if (!pkgInfo || !pkgInfo.valid) {
+        console.log(`##### The \`${addonInfo.pkg.name}\` addon could not be found at \`${addonInfo.path}\` or was otherwise invalid.`);
```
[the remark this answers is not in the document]
davecombs: Yep. I just had it in temporarily for debugging, forgot to remove it. Gone now.

lib/models/instrumentation.js
```diff
@@ -169,6 +173,19 @@ class Instrumentation {
...
+      totalTime: totalTime(tree),
+      platform: {
+        name: process.platform,
+      },
+    };
+
+    _getHardwareInfo(summary.platform);
+
+    return summary;
+  }
```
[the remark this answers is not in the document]
davecombs: Removed. Mentioned it to Richard Conrad, who'll do a PR for this shortly (probably for 3.28). Or I will if he doesn't mind.

lib/models/per-bundle-addon-cache/cache-entry.js
davecombs: Removed the CacheEntry class as not providing a ton of value over plain POJO.

lib/models/per-bundle-addon-cache/cache-entry.js
davecombs: No longer used.

lib/models/per-bundle-addon-cache/index copy.js
davecombs: removed (it was a temporary copy as I was refactoring).

lib/models/per-bundle-addon-cache/index.js
davecombs: Agree. I was originally just following the 'isLazyEngine' style, but yeah, (obj instanceof Project is obvious and much more straightforward.

lib/models/per-bundle-addon-cache/index.js
davecombs: removed for the new lib/utilities/is-lazy-engine.js, which doesn't use chaining.

lib/models/package-info-cache/package-info.js
```diff
@@ -5,7 +5,14 @@ const ErrorList = require('./error-list');
 const Errors = require('./errors');
 const AddonInfo = require('../addon-info');
 const isAddon = require('../../utilities/is-addon');
+const isEngine = require('../../utilities/is-engine');
 const logger = require('heimdalljs-logger')('ember-cli:package-info-cache:package-info');
+const PerBundleAddonCache = require('../per-bundle-addon-cache');
+const { getAddonProxy } = require('../per-bundle-addon-cache/addon-proxy');
+const { values } = require('ember-cli-lodash-subset');
```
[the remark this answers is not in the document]
davecombs: The first 2 are, I'm not sure how 'values' snuck in there. Removing.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,220 @@
...
+   * 
+   * NOTE: this does not indicate that the object is already in the bundle root cache!
+   * 
+   * @public
+   * @method isBundleRoot
+   * @param {Object} obj
+   * @returns {Boolean} true if the object is a lazy engine instance or a constructor
+   * function/class for a lazy engine or is a project instance.
+   */
+  static isBundleRoot(obj) {
```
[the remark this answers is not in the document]
davecombs: I need to check usage, but the idea of the one in PackageInfo is "is this PackageInfo for a bundle root, in other words, does this packageInfo get created into a Project or a lazy engine"? The one here is "is this object a Project or lazy engine"? Those aren't the same, though I guess I could do something similar to your checking of 'constructor' and 'prototype'. I need to look into how these are each used.

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      if (property === 'isProxy') {
+        return true;
+      }
+
+      // trap addons & don't return the real addons array; this is done for primarily
+      // two reasons: (1) `ember-engines` dedupe logic is stateful, so this will just
+      // completely break, and (2) with all dependencies declared explicitly, the build
+      // is _extremely_ slow (eg, it takes over 40 minutes!)
```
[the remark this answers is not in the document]
davecombs: Good idea! I'll reword.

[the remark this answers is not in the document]
davecombs: Wrote a more detailed description of the whole "too many addon instances" problem in the index.js file and refer to that discussion with a link here. We needed that overall discussion of why the PBAC exists and what problem it solves in the code anyway, that seemed the best place.

lib/models/package-info-cache/package-info.js
davecombs: I need to check this. Will put the logic into a utility function that checks the prototype, too, like 'is-lazy-engine.js'.

lib/models/instrumentation.js
davecombs: removed.

lib/models/package-info-cache/package-info.js
davecombs: yep - allowCachingPerBundle. Done.

lib/models/per-bundle-addon-cache/index.js
davecombs: Done. As I'm changing my own code directly, won't commit the suggestion itself.

lib/models/project.js
davecombs: much better - thanks.

lib/models/per-bundle-addon-cache/index.js
davecombs: Done. I still have the one in PackageInfo, but that's specific to checking the PackageInfo itself, not a random object.

lib/models/package-info-cache/package-info.js
```diff
@@ -220,10 +224,49 @@ class PackageInfo {
...
 
+  /**
+   * Indicate if this packageInfo represents a lazy engine.
+   *
+   * @method isForLazyEngine
+   * @return {Boolean} true if this pkgInfo is configured as an engine and the
+   * module this represents has lazyLoading enabled, false otherwise.
+   */
+  isForLazyEngine() {
+    return isEngine(this.pkg.keywords) && isLazyEngine(require(this.realPath));
```
brendenpalmer, line 256: we want to use `require(this.addonMainPath)` here; it's not guaranteed that addon entry points are located at `{realPath}/index.js`

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: resolving this as it's now all been addressed.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: I'm only keeping this open now because I need to add the check in the caching code for whether deduping is enabled or not. I've addressed the rest of it.

lib/models/instantiate-addons.js
```diff
@@ -101,8 +80,10 @@ function instantiateAddons(parent, project, addonPackages) {
     addons.map((addon) => ({
       name: addon.name,
       times: {
-        initialize: addon.constructor._meta_.initializeIn,
-        lookup: addon.constructor._meta_.lookupIn,
+        //initialize: addon.constructor._meta_.initializeIn,
+        //lookup: addon.constructor._meta_.lookupIn,
+        initialize: 0,
+        lookup: 0,
```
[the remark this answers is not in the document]
davecombs: It was a temporary "fix" to override things. When we have a proxy, we have to figure out what to do with the constructor timing. The above is wrong, and needs to be fixed, but the "fix" is not just to uncomment the code, as far as I can tell quickly.

lib/models/package-info-cache/package-info.js
```diff
@@ -220,10 +224,49 @@ class PackageInfo {
...
-  isAddon() {
+  /**
+   * Indicate if this packageInfo is for a project. Should be called only after the project
+   * has been loaded.
+   *
+   * @method isForProject
+   * @return {Boolean} true if this packageInfo is for a Project, false otherwise.
+   */
+  isForProject() {
+    return isProject(this.project);
```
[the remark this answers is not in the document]
davecombs: No, not that I remember. But I will re-check.

lib/models/package-info-cache/package-info.js
davecombs: Good point - thanks!

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,91 @@
...
+      }
+
+      // See the {@link index.js} file for a discussion of why the proxy 'addons'
+      // property returns an empty array.
+      if (property === 'addons') {
+        return [];
+      }
+
+      // allow access to `instance`
+      if (property === 'targetInstance') {
```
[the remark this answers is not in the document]
davecombs: I changed the implementation in `ember-engines` to just check if `PerBundleAddonCache` exists on the project then call a method (getAddonAddons) to do what's needed, rather than having to export this name/symbol.

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,91 @@
...
+      // called again; instead we return a no-op implementation here
+      if (property === 'included') {
+        return () => undefined;
+      }
+
+      if (targetCacheEntry.targetInstance) {
+        if (property !== 'constructor' && typeof targetCacheEntry.targetInstance[property] === 'function') {
+          // we don't want the proxy to be the context of these methods, so wrap
+          // the original methods to maintain the correct context
+          return function _originalAddonPropMethodWrapper() {
```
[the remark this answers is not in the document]
davecombs: Brenden, can you give me a better explanation for this text? You wrote the original comment, I just left it alone, so I wasn't in on precisely what you meant to say. I'm happy to include whatever you like.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,387 @@
...
+ * Some implementation details given the above desired behavior:
+ * (1) There are actually 2 types of caches in this object:
+ * (1a) the first is keyed by lazy engine name. Note: the real instance of
+ * the lazy engine is created as it is encountered while traversing the addon
+ * tree (just like any real addon instance), it's just also referenced
+ * in this cache so we can create proxies to this single instance from
+ * anywhere in the addon tree.
+ * (2) The cache (one item per lazy engine + project) of real addon instances
+ * found while traversing the dependency tree.
+ *
```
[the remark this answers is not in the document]
davecombs: thanks!

lib/models/per-bundle-addon-cache/index.js
davecombs: good idea. Added it.

lib/models/per-bundle-addon-cache/index.js
davecombs: good idea - added it.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,220 @@
...
+    let bundleRootCache = this.bundleRootCaches.get(bundleRootPkgInfo);
+
+    let cacheEntry = bundleRootCache.get(addonPkgInfo);
+    if (cacheEntry) {
+      cacheEntry.instance = addonInstance;
+    } else {
+      cacheEntry = new CacheEntry(addonInstance);
+      bundleRootCache.set(addonPkgInfo, cacheEntry);
+    }
+  }
```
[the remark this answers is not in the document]
davecombs: The code in `getAddonInstance` that was related to the cache was actually moved into the cache as a new method there, also named `getAddonInstance`.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    // We know now we're asking for a 'regular' (non-bundle-root) addon instance.
+
+    let bundleRoot = this.findBundleRoot(parent);
+
+    // the following should not be able to happen given that Projects are always
+    // bundle roots, but just in case, throw an error if we didn't find one.
+    if (!bundleRoot) {
+      throw new Error(`Addon at path\n  ${addonPkgInfo.realPath}\n has 'allowCachingPerBundle' but has no bundleRoot`);
```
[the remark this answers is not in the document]
davecombs: i like it! Moved.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,387 @@
...
+
+  /**
+   * Called by engine-addon when it needs to retrieve the real child addons for a given
+   * object (i.e., during ancestorHostAddons), even though the object might be a proxy.
+   * 
+   * @method getAddonAddons
+   * @param {Addon|Proxy} addonInstance an instance of an addon or a Proxy generated by this class
+   * @return {Array} an array of addon instances
+   */
+  getAddonAddons(addonInstance) {
```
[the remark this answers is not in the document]
davecombs: This is the method that `ember-engines` should call inside `ancestorHostAddons`, instead of futzing around with knowing that `targetInstance` exists or replacing the method with a custom one. I'll add a comment.

lib/models/project.js
```diff
@@ -414,6 +419,17 @@ class Project {
...
   }
 
+  /**
+   * Return the packageInfo entry for this project.
+   *
+   * @public
+   * @method getPackageInfo
+   * @return {PackageInfo} the packageInfo entry for this project
+   */
+  getPackageInfo() {
```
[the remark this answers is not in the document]
davecombs: Checked, and no. It was in an earlier version of my stuff. Removed.

lib/utilities/is-project.js
davecombs: That removes the last use of isBundleRoot, so removed that. Took the contents of this function and moved it to PackageInfo.isForProject, which is the only place isProject was used after isBundleRoot was removed.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: I've now added the check in PerBundleAddonCache's function bundleRootOwnsInstance(). I just have a const that equals `process.env.EMBER_ENGINES_ADDON_DEDUPE`. Since it's in process I don't need to do any trickery with checking if we can require `ember-engines` (though that would ultimately be preferable, if ember-engines exported this process var, so we didn't have to rely on how it's named.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: Removed this.

lib/models/package-info-cache/package-info.js
```diff
@@ -220,10 +227,50 @@ class PackageInfo {
...
+    if (!this.isEngine()) {
+      return false;
+    }
+
+    let module = require(this.realPath);
+
+    let val = (module?.lazyLoading?.enabled === true ||
+      module?.options?.lazyLoading?.enabled === true ||
+      module?.prototype?.lazyLoading?.enabled === true ||
+      module?.prototype?.options?.lazyLoading?.enabled === true);
```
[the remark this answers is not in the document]
davecombs: Reworked this to not use ?

lib/models/package-info-cache/package-info.js
```diff
@@ -368,11 +415,39 @@ class PackageInfo {
...
+   * find the addon's bundle root object (see PerBundleAddonCache.isBundleRoot).
+   * Because Projects are always bundle roots, this should always pass. Just in case,
+   * if it does not, throw an error. 
+   *
+   * @method findBundleRoot
+   * @param {Project|Addon} parent
+   * @return {Object} the object in the 'parent' chain that is a bundle root.
+   */
+  findBundleRoot(parent) {
+    let tmp = parent;
```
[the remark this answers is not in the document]
davecombs: Changed to 'curr'

lib/models/package-info-cache/package-info.js
davecombs: Converted these to `logger.debug` statements.

lib/models/package-info-cache/package-info.js
davecombs: same

lib/models/package-info-cache/package-info.js
davecombs: same

lib/models/per-bundle-addon-cache/addon-proxy.js
davecombs: removed this. I do have an equivalent now to allow references to the Symbol('_targetInstance_'), which is what now replaces the enumerable member 'targetInstance'.

lib/models/per-bundle-addon-cache/index.js
davecombs: Removed 'isProject' entirely here as it wasn't needed.

lib/models/per-bundle-addon-cache/addon-proxy.js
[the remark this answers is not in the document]
davecombs: Done.

lib/models/per-bundle-addon-cache/index.js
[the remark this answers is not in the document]
davecombs: I changed the PackageInfo method names to the form 'isForxxx', like 'isForBundleRoot' that makes things more obvious that they're not quite the same as the utilities in lib/utilities. And I removed some that are no longer needed there.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+   * @param {} parent the addon/project that is to be the direct parent of the
+   * addon instance created here
+   * @param {*} project the project that is to contain this addon instance
+   * @return {Object} the constructed instance of the addon
+   */
+  getAddonInstance(parent, project) {
+    let ctor = this.getAddonConstructor();
+    let addon;
+
+    if (!isSupportingSingletons || !ctor.isSingleton) {
```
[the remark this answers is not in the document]
davecombs: Moved all that to checks for `allowCachingPerBundle`. and check for both the ctor and its prototype.

lib/models/package-info-cache/package-info.js
```diff
@@ -415,6 +490,126 @@ class PackageInfo {
...
+          addon.addons = [];
+        }
+
+        return addon;
+      }
+    } else {
+      // The bundleRoot is not the project but the project bundles the addon too, so
+      // the cache entry needs to go in the project's cache. Create/use an empty
+      // cache entry and return a proxy to it. The project will set the instance later.
+      let cacheEntry = project.perBundleRootAddonCache.addPerBundleInstance(project._packageInfo, this, null);
```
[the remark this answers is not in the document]
davecombs: Fixed.

lib/models/package-info-cache/package-info.js
davecombs: Fixed to check the prototype.

lib/models/instrumentation.js
davecombs: Removed - this will go in with the PR to set up addon-initialization instrumentation.

lib/models/package-info-cache/package-info.js
davecombs: Done

lib/models/package-info-cache/package-info.js
davecombs: done.

lib/models/package-info-cache/package-info.js
davecombs: Moved the block of 4 lines starting at 595 to a separate function, which is then invoked (and which is also invoked in the PerBundleAddonCache), and put a more enlightening comment.

lib/models/package-info-cache/package-info.js
davecombs: outdated, handled differently.

lib/models/per-bundle-addon-cache/index.js
davecombs: updated the @method to the real one.

lib/models/per-bundle-addon-cache/index.js
davecombs: Fixed the comment fields.

lib/models/project.js
davecombs: Done.

lib/utilities/is-project.js
davecombs: Done - I check isEmberCliProject() after testing for the symbol being present.

lib/models/per-bundle-addon-cache/addon-proxy.js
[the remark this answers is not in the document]
davecombs: Rob had a better explanation so I included a version something like that. Definitely more obvious.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: No. The project field is added by PackageInfoCache.loadProject().

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,402 @@
...
+    // The cache is keyed by name (since there is only a single instance of any given
+    // engine name, even if 2 engine addons have the same engine name).
+    // To allow the project in the cache, it will use the empty string for a name,
+    // to not collide with any possible engine name.
+    // Each cache entry consists of the bundle host instance (Project or Lazy engine)
+    // and an addon instance cache to bundle with that bundle host.
+    this.bundleHostCache = this._setupBundleHostCache();
+
+    // Indicate if ember-engines deduping is supported.
+    this.engineAddonDedupeEnabled = !!process.env.EMBER_ENGINES_ADDON_DEDUPE;
```
brendenpalmer, line 86: *nit*: might be worth renaming this to `engineAddonTransitiveDedupeEnabled` since `ember-engines` dedupes _direct_ engine dependencies by default

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,402 @@
...
+    this.bundledProjectAddonPkgInfos = this._setupBundledProjectAddonPackageInfos();
+
+    // The cache of bundle-host instances and their individual addon caches.
+    // The cache is keyed by name (since there is only a single instance of any given
+    // engine name, even if 2 engine addons have the same engine name).
+    // To allow the project in the cache, it will use the empty string for a name,
+    // to not collide with any possible engine name.
+    // Each cache entry consists of the bundle host instance (Project or Lazy engine)
+    // and an addon instance cache to bundle with that bundle host.
+    this.bundleHostCache = this._setupBundleHostCache();
```
brendenpalmer, line 83: do we want to set these caches up eagerly? if nobody enables `allowCachingPerBundle` for any addon, then ppl will incur this cost and these caches will go unused

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,402 @@
...
+
+  /**
+   * Called by engine-addon when it needs to retrieve the real child addons for a given
+   * object (i.e., during ancestorHostAddons), even though the object might be a proxy.
+   *
+   * @method getAddonAddons
+   * @param {Addon|Proxy} addonInstance an instance of an addon or a Proxy generated by this class
+   * @return {Array} an array of addon instances
+   */
+  getAddonAddons(addonInstance) {
```
brendenpalmer, line 389: why does this method live on `PerBundleAddonCache`? i'd probably just remove this for now, especially w/ plans to move over host addons/LCA host logic to `ember-cli`: https://github.com/ember-cli/ember-cli/issues/9501

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -0,0 +1,402 @@
...
+      addonInstance[TARGET_INSTANCE] && addonInstance[TARGET_INSTANCE].addons
+        ? addonInstance[TARGET_INSTANCE].addons
+        : addonInstance.addons;
+    return result;
+  }
+}
+
+module.exports = PerBundleAddonCache;
+
+// Support for per-bundle addon caching is opt OUT. Caching any particular
```
brendenpalmer, line 400: i'd think i'd clarify this a bit; it's a _global_ opt-out (even if any addon has itself enabled `allowCachingPerBundle`)

lib/models/package-info-cache/package-info.js
```diff
@@ -220,10 +223,49 @@ class PackageInfo {
...
-  isAddon() {
+  /**
+   * Indicate if this packageInfo is for a project. Should be called only after the project
+   * has been loaded.
+   *
+   * @method isForProject
+   * @return {Boolean} true if this packageInfo is for a Project, false otherwise.
+   */
+  isForProject() {
+    return !!this.project && this.project.isEmberCLIProject && this.project.isEmberCLIProject();
```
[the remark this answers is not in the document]
davecombs: I believe there are test cases where it is not set (I have other spots in the package-info code that hit this issue for other parts of Project). Yes, I will add a comment about 'this.project'.

lib/models/per-bundle-addon-cache/index.js
davecombs: Without my having to wade into the middle of engine code TOO much, is that default deduping going to be affected by the caching too, or not? If I need to learn the engine code before finding an answer, just let me know.

lib/models/per-bundle-addon-cache/index.js
davecombs: I get your point. Just FYI the cost of setup is << 1 second. But I'll look more into the complexity of what has to be checked where to make sure the caches are or are not created. Remember that the addon instance cache for a given lazy engine MUST exist BEFORE the lazy engine itself has finished its constructor, because of that g*&(&mn `setupRegistry` call to create the children before the parent is done. It cannot wait until there is an instance of a child that wants to be put into the addon cache for that bundle host, as we don't have access to the inside of the bundle host's constructor function.

lib/models/per-bundle-addon-cache/index.js
davecombs: It's because I didn't want to have to completely replace the ancestorHostAddons function in ember-engines. This is what I was talking about when we discussed releasing ember-engines first (so that its function can call back into here.) The code there just says:

if (project.perBundleAddonCache) {
  addons = perBundleAddonCache.getAddonAddons(addon);
} else {
  addons = addon.addons;
}

lib/models/per-bundle-addon-cache/index.js
davecombs: Sounds good.

lib/models/instantiate-addons.js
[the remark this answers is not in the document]
davecombs: I have a separate PR up (9503) to move them to a _timings_ object on the addon instance. That's only used once, to dump init timing to the log file, in instantiate-addons. Waiting for @rwjblue or some other maintainer to approve that PR.

lib/models/package-info-cache/package-info.js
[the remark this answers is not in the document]
davecombs: Added the comment

lib/models/instantiate-addons.js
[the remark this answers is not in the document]
davecombs: Submitted PR #9503 to fix this particular issue separate from this PR.

lib/models/per-bundle-addon-cache/index.js
davecombs: Changed to use your naming.

lib/models/per-bundle-addon-cache/index.js
davecombs: Updated the comment to be more explanatory.

lib/models/per-bundle-addon-cache/addon-proxy.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const TARGET_INSTANCE = Symbol.for('_targetInstance_');
```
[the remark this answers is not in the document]
davecombs: Converted to another file, (target_instance.js) in the same directory, and then 'require' that from both the proxy code and the PerBundleAddonCache module.

lib/models/per-bundle-addon-cache/index.js
davecombs: Until we have the fix to move the `findLCAHost` and other stuff in `ancestorHostAddons` over from ember-engines to ember-cli, we need some sort of call like the above. I'll hold this open until we decide what to do and how to stage the changes.

rwjblue (commented):
This is looking nice, thanks for continuing to hone it @davecombs!

I think the main thing left is tests for the new functionality. When you get that going, please ping me for another round of review.

rwjblue:
Where are we at here @davecombs?

lib/models/per-bundle-addon-cache/index.js
davecombs: Ultimately decided to leave setting up the caches as is for now, given the small cost.

davecombs:
added a ton of test cases and updates to fixturify-project to support the proxy feature.

brendenpalmer (changes requested):

tests/unit/models/per-bundle-addon-cache/enable-cache-test.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function createProject() {
+  let fixture = createStandardCacheFixture();
+  let project = fixture.buildProjectModel(Project);
+  return project;
+}
+
+describe('Unit | per-bundle-addon-cache enable caching', function () {
+  it('perBundleAddonCache should be set in Project if EMBER_CLI_ADDON_INSTANCE_CACHING is not false', function () {
```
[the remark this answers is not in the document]
davecombs: Sure. When I ran it, though, tests passed. Oh, crap - I was running each file separately.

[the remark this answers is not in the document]
davecombs: Done

PR #9485, opened 2021-03-16
Add Ember 3.24 LTS to ember-try configuration

bertdeblock:

PR #9488, opened 2021-03-18
Update supported Ember version in addon blueprint

bertdeblock:
Forgot to update this when doing https://github.com/ember-cli/ember-cli/pull/9485.

### DIFF
--- a/blueprints/addon/files/README.md
+++ b/blueprints/addon/files/README.md
@@ -7,7 +7,7 @@
 Compatibility
 ------------------------------------------------------------------------------
 
-* Ember.js v3.16 or above
+* Ember.js v3.20 or above
 * Ember CLI v2.13 or above
 * Node.js v10 or above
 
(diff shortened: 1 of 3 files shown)


blueprints/addon/files/README.md
```diff
@@ -7,7 +7,7 @@
 Compatibility
 ------------------------------------------------------------------------------
 
-* Ember.js v3.16 or above
+* Ember.js v3.20 or above
 * Ember CLI v2.13 or above
```
rwjblue, line 11: Mind sending another PR updating this to current versions too?

rwjblue (approved):

blueprints/addon/files/README.md
bertdeblock: Sure, which version should this be?
Was also wondering if we could automatically fill this in when generating an addon, one less thing to update manually when we have a new LTS version.

rwjblue: Should be in sync with ember version.

Issue #8782, opened 2019-08-12
config changes not picked up by development server


chancancode:
Likely duplicate of #8478, which has more information and a potential PR

1. Generate an ember app
1. Run `FOO=bar ember s`
1. Go to `https://localhost:4200/`
1. Run in the console `require("my-app/config/environment").default.FOO` => `undefined`
1. Add `ENV.FOO = process.env.FOO;` to `config/environment.js`
1. The file _is_ being watched, and it looks like the server rebuilt the app as expected
1. Refresh `https://localhost:4200/` (it was refreshed automatically after the build anyway)
1. Run in the console `require("my-app/config/environment").default.FOO` => still `undefined`
1. Restart the server, passing `FOO=bar` again
1. Go to `https://localhost:4200/`
1. Run in the console `require("my-app/config/environment").default.FOO` => `"bar"` as expected

It seems like this is not related to reading `FOO` from `process.env`. Changing the line to `ENV.FOO = 'bar'` seems to have the same issue.

chancancode:
Asked @rwjblue on discord, he seems to think this is a bug!

chancancode:
I think @rwjblue had confirmed that this is a bug, and it does look like it's probably the same as #8478, and it happens on OS X as well. This is affecting the tutorial, specifically [this chapter](https://octane-guides-preview.emberjs.com/release/tutorial/07-reusable-components/). When this is fixed, we can remove a bunch of steps from there.

bertdeblock:
This is still a valid issue, but I'm going to close this one as it's a duplicate of https://github.com/ember-cli/ember-cli/issues/8782 indeed.

PR #9489, opened 2021-03-19
BUGFIX: Clear project.configCache before starting the build

### LINKED ISSUE #8478: Changes in environment.js applied only on second save

colenso:
Fixes #8478.
Fixess https://github.com/ember-cli/ember-cli/issues/8782.

The problem I noticed was that get-serve-url would access project.config after finalizeBuild and hence the config would get cached and would be used in the next build. Hence each build was using the previous build's config.
This PR simply clears the project.configCache before the build is started.


colenso:
Can someone help me out with tests? @locks perhaps?

sly7-7:
Since the time I use ember-cli, I just always assumed it was "normal" to have to `ember s` again when changing the environnment.js file.

colenso:
Yes. But it always wasn't so. And I find it very handy to have it rebuild automatically when I change something in the environment.js file

rwjblue:
I definitely agree that the behavior is important to fix, but we do need some tests.

I think you could add a test to this file, like this test:

https://github.com/ember-cli/ember-cli/blob/6252533f5cf9ba1e77528f7711fe311324901a29/tests/unit/broccoli/builder-test.js#L32

In your case, you don't need a special broccoli plugin (like that test I linked above). I think you mostly just want to ensure that the config content is correct per build.

bertdeblock:

PR #9491, opened 2021-03-19
Update supported Ember CLI version in addon blueprint

bertdeblock:
[More info](https://github.com/ember-cli/ember-cli/pull/9488#discussion_r597218016)

### DIFF
--- a/blueprints/addon/files/README.md
+++ b/blueprints/addon/files/README.md
@@ -8,7 +8,7 @@ Compatibility
 ------------------------------------------------------------------------------
 
 * Ember.js v3.20 or above
-* Ember CLI v2.13 or above
+* Ember CLI v3.20 or above
 * Node.js v10 or above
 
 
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

kellyselden:
I don't think this is right. Addons work in current - 1 of ember-cli. Otherwise every ember-cli-update run would "break" addons, requiring everyone update their version of ember-cli.

rwjblue:
> Addons work in current - 1 of ember-cli. 

Sorry, what does this mean? Addons should work with all supported versions of ember-cli, which is roughly the same support schedule as ember-source. By the time this branch is published (3.27.0 in ~ 8 weeks) that would be only ember-cli@3.20 ember-cli@3.24 and ember-cli@3.27 (which is what this documentation update implies).

kellyselden:
Oh my bad I misread. I see now that it is tracking the ember version with -2 LTS. I revoke my concern :)

kellyselden (approved):

PR #9493, opened 2021-03-22
Add CLASSIC as a case for CI

thoov:
This allows to keep classic compatibility while we transition towards Embroider as the default.


thoov:

rwjblue (approved):

PR #9505, opened 2021-04-19
Pass `realPath` as `root` rather than the dirname for `addonMainPath`

brendenpalmer:
Currently it's possible for package infos representing the same addon to differ throughout the project. This is only true if an addon has an entry point located outside of `realPath` (eg, `realPath/lib/main.js`). This fixes this bug and adds a test-case to guard against this in the future.

### DIFF
--- a/lib/models/package-info-cache/package-info.js
+++ b/lib/models/package-info-cache/package-info.js
@@ -390,18 +390,17 @@ class PackageInfo {
     // TODO: Future work - allow a time budget for loading each addon and warn
     // or error for those that take too long.
     let module = require(this.addonMainPath);
-    let mainDir = path.dirname(this.addonMainPath);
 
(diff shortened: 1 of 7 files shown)


rwjblue (approved):

ef4:
I think this is a breaking change. Existing addons that customized their `ember-addon.main` were previously forced to also relocate things like their default `vendor` directory, so they did. This breaks them.

This is the case for `@embroider/macros` for example, which points `ember-addon.main` to `src/ember-addon-main.js` and therefore has to put its vendor in `src/vendor`.

rwjblue:
😭 that is a bit unfortunate, but you are totally correct

rwjblue:
@brendenpalmer - Would you mind sending a PR to revert?

PR #9532, opened 2021-05-10
Replace deprecated `babel-eslint` with `@babel/eslint-parser`

### LINKED ISSUE #9307: Replace deprecated babel-eslint with new @babel/eslint-parser
This is mostly a tracking issue as I am not in a place to open a PR for this atm, but,

here in the app blueprint: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L29
we're still using babel-eslint,
but, per, https://github.com/babel/babel-eslint, the eslint parser for babel has moved into their monorepo as https://github.com/babel/babel/tree/master/eslint/babel-eslint-parser

bmish:
Fixes #9307.

Old: https://www.npmjs.com/package/babel-eslint
New: https://www.npmjs.com/package/@babel/eslint-parser

Note: `@babel/eslint-parser requires @babel/core@>=7.2.0` (which is already a transitive dependency through ember-cli-babel among other dependencies)

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -2,13 +2,14 @@
 
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@babel/eslint-parser',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
     ecmaFeatures: {
       legacyDecorators: true,
     },
+    requireConfigFile: false,
   },
   plugins: ['ember'],
   extends: [
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -9,7 +9,6 @@ module.exports = {
     ecmaFeatures: {
       legacyDecorators: true,
     },
-    requireConfigFile: false,
   },
   plugins: ['ember'],
   extends: [
@@ -28,6 +27,7 @@ module.exports = {
         '.eslintrc.js',
         '.prettierrc.js',
         '.template-lintrc.js',
+        'babel.config.js',
         'ember-cli-build.js',<% if (blueprint !== 'app') { %>
         'index.js',<% } %>
         'testem.js',
--- a/blueprints/app/files/babel.config.js
+++ b/blueprints/app/files/babel.config.js
@@ -1,0 +1,17 @@
+const { buildEmberPlugins } = require('ember-cli-babel');
+
+module.exports = function (api) {
+  api.cache(true);
+
+  return {
+    presets: [
+      [
+        require.resolve('@babel/preset-env'),
+        {
+          targets: require('./config/targets'),
+        },
+      ],
+    ],
+    plugins: [...buildEmberPlugins(__dirname)],
+  };
+};
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -20,6 +20,7 @@
 /.template-lintrc.js
 /.travis.yml
 /.watchmanconfig
+/babel.config.js
 /bower.json
 /config/ember-try.js
 /CONTRIBUTING.md
--- a/blueprints/app/files/ember-cli-build.js
+++ b/blueprints/app/files/ember-cli-build.js
@@ -4,6 +4,9 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function (defaults) {
   let app = new EmberApp(defaults, {
+    'ember-cli-babel': {
+      useBabelConfig: true,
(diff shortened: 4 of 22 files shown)


tehhowch:
One issue with this PR as of 68c6bc4 is that linting the quickstart app fails due to the experimental decorators.

```
  5:2  error  Parsing error: /home/bjh/projects/ember_new/my-app/app/components/people-list.js: Support for the experimental syntax 'decorators-legacy' isn't currently enabled (5:3):

  3 |
  4 | export default class PeopleListComponent extends Component {
> 5 |   @action
    |   ^
  6 |   showPerson(person) {
  7 |     alert(`The person's name is ${person}!`);
  8 |   }
```
I believe the solution is to use a babel config file that then specifies the legacy plugin-proposal-decorators behavior:
```json
{
  "plugins": [["@babel/plugin-proposal-decorators", { "legacy": true }]]
}
```

It's possible others are needed (such as class properties and any other plugins that get added to ember-cli-babel).

bmish:
@tehhowch good point. I definitely needed these two plugins in my app:

```
"plugins": [
  "@babel/plugin-proposal-class-properties",
  ["@babel/plugin-proposal-decorators", { "decoratorsBeforeExport": true }]
]
```

Open questions:
1. Should we use the .eslintrc.js config to specify the Babel plugins, or is it time to start a new `.babelrc` file?
2. Presumably, since ember-cli-babel already specifies these plugins as transitive dependencies, we shouldn't add them as explicit dependencies ourselves (letting the wrapper manage it for us)?
3. It's kind of unfortunate we need to manage the list of Babel plugins in an additional place now...

blueprints/app/files/.eslintrc.js
```diff
@@ -2,13 +2,14 @@
...
   root: true,
-  parser: 'babel-eslint',
+  parser: '@babel/eslint-parser',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
     ecmaFeatures: {
       legacyDecorators: true,
     },
+    requireConfigFile: false,
```
rwjblue, line 12: Instead of this, we should try to work towards leveraging the `baber.config.js` support in recent versions of ember-cli-babel.

See https://github.com/rwjblue/___demo-babel-config as a demo of how that can work.

rwjblue: Without that change, the configuration for `@babel/eslint-parser` will have to be duplicated like mentioned in https://github.com/babel/ember-cli-babel/issues/366 (specifically https://github.com/babel/ember-cli-babel/issues/366#issuecomment-729839454).

bmish: Neat! I just copied the changes from https://github.com/rwjblue/___demo-babel-config (following the README) (and also added the `babel.config.js` to `.npmignore`). Note that this is a much larger change now...

tehhowch:
This is enough to build the quickstart app, but a first attempt to use in my workplace's monorepo failed with module resolution errors in the webpack + auto-import stage. Given that it's on 3.24 LTS and I had to bump ember-cli-babel for this config to be possible to try, I probably missed something.

(The app I tried building does specify `includePolyfill: true`; is that something we would need to move elsewhere when using the Babel config file option?)

bmish:
@tehhowch I think it's expected that you had to bump ember-cli-babel due to @rwjblue's comment: https://github.com/ember-cli/ember-cli/pull/9532#discussion_r630253943

tehhowch:
Yep, I was just clarifying that my failure to apply this to an existing monorepo is not necessarily due to this PR not specifying something it should, but rather that I may have not done all the needed background bits. For example, I probably should have retried with the latest v1.x of ember-auto-import, rather than v1.10.1. I will give it another go next week.

rwjblue:
@bmish - Mind rebasing? We fixed a few issues on master branch in the last week that I _think_ should make this landable.

bmish:
@rwjblue rebased.

blueprints/app/files/babel.config.js
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = function (api) {
+  api.cache(true);
+
+  return {
+    presets: [
+      [
+        require.resolve('@babel/preset-env'),
+        {
+          targets: require('./config/targets'),
```
rwjblue, line 11: This require is failing when running inside an addon (because the file is in a different location: `tests/dummy/config/targets.js`)

bmish: Fixed in #9556. I had a problem with this branch and had to re-create the PR.

bmish:
This PR has been replaced by #9556. Apologies about the confusion.

PR #9537, opened 2021-05-19
Implement LCA host/host addons logic in `ember-cli`

brendenpalmer:
This PR implements host addons/LCA host logic in `ember-cli` and updates addon bundling caching to use these host addons, rather than _only_ considered bundled addons by the project.

An LCA host in this context is defined as the lowest common ancestor that is considered a host amongst _all_ bundle hosts (engines/project) by the same name in the project.

The behavior in `ember-cli` needs to differ than the [original implementation](https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/lib/utils/find-lca-host.js) in `ember-engines` because we need to know this information up-front _during_ addon instantiation, rather than after all addons in the project have been instantiated. As a result, the method for determining the LCA host/host addons is different: because we don't have access to addon instances, we use `PackageInfo` objects to compute this information. It's worth noting that `PackageInfo` objects _do not_ have knowledge of their parents, so we do a breadth-first traversal, keeping track of all paths until we discover the intended engine (represented by a `PackageInfo`).

For example, given the following project addon structure:

```
        --Project--
         /      \
        /        \
   Lazy Engine A  \
                Addon A
                  |
                  |
             Lazy Engine B
              /          \
             /            \
        Lazy Engine A   Lazy Engine C
```

- The LCA host for Lazy Engine A is the project
- The LCA host for Lazy Engine B is the project
- The LCA host for Lazy Engine C is Lazy Engine B

Similarly, we're also interested in all of the host package infos for a given LCA host; in the above example: `hostAndAncestorBundledPackageInfos` for each are considered to be:

- `hostAndAncestorBundledPackageInfos` for lazy engine A includes all non-lazy dependencies of its LCA host & above (in this case, just the project)
- `hostAndAncestorBundledPackageInfos` for lazy engine B includes all non-lazy dependencies of its LCA host & above (in this case, just the project)
- `hostAndAncestorBundledPackageInfos` for lazy engine C includes non-lazy deps of lazy engine B & non-lazy deps of the project (LCA host & above)

As part of this, we also expose a `getLCAHost` method so that `ember-engines` can prefer to use this method if it exists; we prefer that the computation for all of this has a single source of truth for both `ember-engines` and bundle addon caching.

### DIFF
--- a/lib/models/host-info-cache.js
+++ b/lib/models/host-info-cache.js
@@ -1,0 +1,334 @@
+'use strict';
+
+function allPkgInfosEqualAtIndex(paths, index) {
+  const itemToCheck = paths[0][index];
+  return paths.every((pathToLazyEngine) => pathToLazyEngine[index] === itemToCheck);
+}
+
+class HostInfoCache {
+  constructor(project) {
+    this.project = project;
+    this._bundledPackageInfoCache = new Map();
+    this._hostAddonInfoCache = new Map();
+    this._lcaHostCache = new Map();
+  }
+
+  /**
+   * Given a path (calculated as part of `getHostAddonInfo`), return the correct
+   * "bundle host". A bundle host is considered the project or lazy engine.
+   *
+   * For example, given the following package structure:
+   *
+   *      --Project--
+   *       /      \
+   *      /        \
+   * Lazy Engine A  \
+   *              Addon A
+   *                |
+   *                |
+   *           Lazy Engine B
+   *            /          \
+   *           /            \
+   *      Lazy Engine A   Lazy Engine C
+   *
+   * The provided paths for lazy engine A would look like:
+   *
+   * - [Project]
+   * - [Project, Addon A, Lazy Engine B]
+   *
+   * For this project structure, this function would return [Project, [Project]]
+   *
+   * Similarly, given the following project structure:
+   *
+   *            --Project--
+   *             /      \
+   *            /        \
+   *     Lazy Engine A    \
+   *          /        Lazy Engine B
+   *         /               |
+   *        /                |
+   *  Lazy Engine C     Lazy Engine C
+   *
+   * The provided paths for lazy engine C would look like:
+   *
+   * - [Project, Lazy Engine A]
+   * - [Project, Lazy Engine B]
+   *
+   * In this case, the host is the project and would also return [Project, [Project]]
+   *
+   * @method _findNearestBundleHost
+   * @param {Array<PackageInfo[]>} paths The found paths to a given bundle host
+   * @return {[PackageInfo, PackageInfo[]]}
+   * @private
+   */
+  _findNearestBundleHost(paths, pkgInfoForLazyEngine) {
+    // building an engine in isolation (it's considered the project, but it's
+    // also added as a dependency to the project by `ember-cli`)
+    if (this.project._packageInfo === pkgInfoForLazyEngine) {
+      return [this.project._packageInfo, [this.project._packageInfo]];
+    }
+
+    const shortestPath = paths.reduce(
+      (acc, pathToLazyEngine) => Math.min(acc, pathToLazyEngine.length),
+      Number.POSITIVE_INFINITY
+    );
+
+    const pathsEqualToShortest = paths.filter((pathToLazyEngine) => pathToLazyEngine.length === shortestPath);
+    const [firstPath] = pathsEqualToShortest;
+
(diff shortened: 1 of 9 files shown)


lib/models/per-bundle-addon-cache/index.js
```diff
@@ -235,16 +146,19 @@ class PerBundleAddonCache {
 
   /**
    * An optimization we support from lazy engines is the following:
+   *
    * if an addon instance is supposed to be bundled with a particular lazy engine, and
-   * same addon is also to be bundled in the project, prefer the one in the project.
+   * same addon is also to be bundled by a common LCA host, prefer the one bundled by the
+   * host (since it's ultimatly going to be deduped later by `ember-engines`).
```
davecombs, line 152: ultimately

davecombs (changes requested):
Some minor changes to ask for more descriptive member names and some more documentation of methods, but otherwise this looks good! Nice work.

lib/models/project.js
```diff
@@ -18,6 +18,7 @@ const PackageInfoCache = require('./package-info-cache');
 const PerBundleAddonCache = require('./per-bundle-addon-cache');
 
 const instantiateAddons = require('./instantiate-addons');
+const { HostInfoCache } = require('./host-info-cache');
```
davecombs, line 21: Given that all the other modules here are just the class, not a class inside an object, and there is nothing else in this particular object, would it be better to just have `const HostInfoCache` w/o the {} wrapper? More consistent at least, if not any better.

tests/unit/models/per-bundle-addon-cache/proxy-test.js
```diff
@@ -271,29 +271,317 @@ describe('models/per-bundle-addon-cache', function () {
...
-      expect(cacheEntries).to.exist;
+      const lazyEngineBPkgInfo = project.addons.find((addon) => addon.name === 'lazy-engine-b')._packageInfo;
+      cacheEntries = findAddonCacheEntriesByName(project.perBundleAddonCache, lazyEngineBPkgInfo, 'test-addon-a');
+      expect(cacheEntries).to.deep.equal([], 'should exist; lazy engine B has opted into bundle caching');
 
-      // project cache should be used
-      expect(cacheEntries.length).to.equal(0);
+      cacheEntries = findAddonCacheEntriesByName(project.perBundleAddonCache, lazyEngineBPkgInfo, 'test-addon-b');
+      expect(cacheEntries).to.deep.equal([], 'should exist; lazy engine B has opted into bundle caching');
+      expect(cacheEntries.length).to.equal(0, 'project cache should be used');
```
davecombs, line 287: Is this necessary, since you have the 'to.deep.equal' directly above?  Seems superfluous.

davecombs: Also, should there be positive tests here? The above shows that the 2 `test-addon...` objects aren't in the lazy engine caches - should it be shown that they ARE in the project cache?

davecombs (changes requested):
Found a few more cases that look like something is missing.

lib/models/host-info-cache.js
```diff
@@ -0,0 +1,332 @@
...
+   *
+   * In this case, the host is the project and would also return [Project, [Project]]
+   *
+   * @method _findNearestBundleHost
+   * @param {Array<PackageInfo[]>} paths The found paths to a given bundle host
+   * @return {PackageInfo}
+   * @private
+   */
+  _findNearestBundleHost(paths, pkgInfoForLazyEngine) {
+    // building an engine in isolation (it's considered both the project, but it's
```
davecombs, line 65: can remove 'both' here.

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -68,21 +59,16 @@ class PerBundleAddonCache {
   constructor(project) {
     this.project = project;
 
-    // cache of the addon packageInfos for addons that are descendants of the project
-    // (excluding going into lazy engines). Lazy engines that refer to any of these
-    // will defer to the project-level one.
-    this.bundledProjectAddonPkgInfos = this._setupBundledProjectAddonPackageInfos();
-
     // The cache of bundle-host instances and their individual addon caches.
```
davecombs, line 62: Could you update this comment, since the cache is no longer a cache of instances but a cache of PackageInfos. Also, the 'empty string for a name' thing has gone away, right?

rwjblue (approved):

PR #9540, opened 2021-05-23
Use relative override paths in blueprint ESLint config

loganrosen:
Since eslint/eslint#11799 landed in ESLint v6 some time back, we can now use relative paths in the ESLint override configuration in the blueprints.

ESLint didn't previously support relative paths in overrides, so we had to list directories containing files (which are Ember files) with the same names as those listed in `files` (Node files in the root directory) in `excludedFiles` (see #7529). Now that the paths in `files` are relative, we can be sure that it will only match against the files in the root directory, not files with the same name in subdirectories.

For good measure, I've also changed the subdirectory globs to be relative to the root as well, in case e.g. there are other directories in the tree with the same name that should not be matched.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -24,24 +24,18 @@ module.exports = {
     // node files
     {
       files: [
-        '.eslintrc.js',
-        '.prettierrc.js',
-        '.template-lintrc.js',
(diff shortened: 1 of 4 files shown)


rwjblue (approved):
This is great, thank you @loganrosen!

PR #9541, opened 2021-05-23
Use a better dummy package than fs-extra in a test

pgengler:
The 'fs-extra' package, as of version 10, has dropped Node 10 support, leading to failing builds for beta (which still has Node 10 support).

This changes the test that uses that package (via a blueprint) to instead use the 'ember-try-test-suite-helper' package (which was created as a dummy package for ember-try to use). Since that package doesn't have any dependencies/devDependencies or an 'engines' statement it should not end up with any conflicts down the line.


kellyselden (approved):

rwjblue:
Thank you @pgengler!

PR #9542, opened 2021-05-24
Add validation checks for addon instance bundle caching

brendenpalmer:
This PR implements two validation checks for addon bundle caching:

1. If an `EmberApp` has a project that has opted in to addon bundle caching _and_ that app has utilized the addon `whitelist`/`blacklist` this is treated as an error condition. Addon bundle caching needs to know the host addons _during_ addon instantiation, but these configurations change these semantics afterwards, so we're treating this as an error condition.
2. If a proxy _doesn't_ return stable cache keys for a given tree type, this is treated as an error condition. Bundle addon caching is assumed to be used on addons w/ stable cache keys (eg, for this to work properly w/ `ember-engines` the cache key needs to be stable).

I've also added tests to cover these error conditions.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -168,6 +168,30 @@ class EmberApp {
 
     this._isPackageHookSupplied = typeof this.options.package === 'function';
     this._cachedAddonBundles = {};
+
+    if (this.project.perBundleAddonCache && this.project.perBundleAddonCache.numProxies > 0) {
+      if (this.options.addons.whitelist && this.options.addons.whitelist.length) {
+        throw new Error(
+          [
+            '[ember-cli] addon bundle caching is disabled for apps that specify an addon `whitelist`',
+            '',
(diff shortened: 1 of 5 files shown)


lib/broccoli/ember-app.js
```diff
@@ -168,6 +168,16 @@ class EmberApp {
 
     this._isPackageHookSupplied = typeof this.options.package === 'function';
     this._cachedAddonBundles = {};
+
+    if (this.project.perBundleAddonCache && this.project.perBundleAddonCache.numProxies > 0) {
+      if (this.options.addons.whitelist && this.options.addons.whitelist.length) {
+        throw new Error('[ember-cli] addon bundle caching is disabled for apps that specify an addon `whitelist`');
```
rwjblue, line 174: Hmm, I think we need to beef up these error messages. It may not be obvious which addons have opt-ed into bundle caching, can we include the list in the message?

rwjblue (approved):

PR #9543, opened 2021-05-24
Add ability to specify a custom `ember-addon.perBundleAddonCacheUtil` utility

brendenpalmer:
This PR adds the ability to specify a custom configuration in the project's `package.json` via the `ember-addon.perBundleAddonCacheUtil` configuration.

If a consuming application specifies a relative path to a custom utility via the `ember-addon.perBundleAddonCacheUtil` configuration, we prefer the custom implementation provided by the consumer; however, the default implementation here is to indicate if the original addon entry point has the `allowCachingPerBundle` flag set either on itself or on its prototype.

### DIFF
--- a/lib/models/package-info-cache/package-info.js
+++ b/lib/models/package-info-cache/package-info.js
@@ -262,7 +262,7 @@ class PackageInfo {
    * module this represents has lazyLoading enabled, false otherwise.
    */
   isForLazyEngine() {
-    return this.isForEngine() && isLazyEngine(require(this.addonMainPath));
+    return this.isForEngine() && isLazyEngine(this._getAddonEntryPoint());
   }
 
   /**
@@ -435,14 +435,7 @@ class PackageInfo {
       return this.addonConstructor;
     }
 
-    if (!this.addonMainPath) {
-      throw new Error(`${this.pkg.name} at ${this.realPath} is missing its addon main file`);
-    }
-
-    // Load the addon.
-    // TODO: Future work - allow a time budget for loading each addon and warn
-    // or error for those that take too long.
-    let module = require(this.addonMainPath);
+    let module = this._getAddonEntryPoint();
     let mainDir = path.dirname(this.addonMainPath);
 
     let ctor;
@@ -516,7 +509,7 @@ class PackageInfo {
    * @return {Object} the constructed instance of the addon
    */
   getAddonInstance(parent, project) {
-    let addonEntryPointModule = require(this.addonMainPath);
+    let addonEntryPointModule = this._getAddonEntryPoint();
     let addonInstance;
 
     if (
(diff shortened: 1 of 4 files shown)


lib/models/package-info-cache/package-info.js
```diff
@@ -530,10 +516,14 @@ class PackageInfo {
    * @return {Object} the constructed instance of the addon
    */
   getAddonInstance(parent, project) {
-    let ctor = this.getAddonConstructor();
+    let addonEntryPointModule = require(this.addonMainPath);
```
rwjblue, line 519: `getAddonConstructor` checks to ensure that addonMainPath is present and throws a nice error, if we move to directly requiring (which seems fine) we should ensure that we do the same nice error

lib/models/package-info-cache/package-info.js
```diff
@@ -530,10 +516,14 @@ class PackageInfo {
...
     let addonInstance;
 
-    if (PerBundleAddonCache.isEnabled() && this.allowCachingPerBundle(ctor)) {
+    if (
+      PerBundleAddonCache.isEnabled() &&
+      project &&
```
rwjblue, line 524: Is it possible for `project` to be missing? That seems surprising 🤔

lib/models/per-bundle-addon-cache/index.js
```diff
@@ -1,10 +1,48 @@
...
+const path = require('path');
 const isLazyEngine = require('../../utilities/is-lazy-engine');
 const { getAddonProxy } = require('./addon-proxy');
 const logger = require('heimdalljs-logger')('ember-cli:per-bundle-addon-cache');
 const { TARGET_INSTANCE } = require('./target-instance');
 
+function defaultAllowCachingPerBundle({ addonEntryPointModule }) {
+  return (
+    addonEntryPointModule.allowCachingPerBundle ||
+    (addonEntryPointModule.prototype && addonEntryPointModule.prototype.allowCachingPerBundle)
```
rwjblue, line 13: Hmm, this seems wrong? `addonEntryPointModule` is the plain / non-"tweaked" export (usually `module.exports = { name: require('./package').name }`) now (it used to be the result of `getAddonConstructor()` but that was changed).

rwjblue (changes requested):

lib/models/per-bundle-addon-cache/index.js
rwjblue: I was mistaken here, the code is accounting for either case

@brendenpalmer is going to double check to ensure there are tests for each of these cases

lib/models/package-info-cache/package-info.js
brendenpalmer: Ah, good call. I'll update

lib/models/package-info-cache/package-info.js
brendenpalmer: yeah, it is i guess, I was a bit surprised too 😞. i'm unsure of the exact case(s) that'd trigger this, but for example, there is a test-case to test that `treeGenerator` works without `project`: https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/addon-test.js#L35

lib/models/per-bundle-addon-cache/index.js
brendenpalmer: ah yeah, this case is just to cover if someone tries to opt-in as part of exporting a subclass of the base model addon directly; eg, `module.exports = Addon.extend({ allowCachingPerBundle: true });`. there wasn't a test-case to cover this, but i just updated this PR to test for this

lib/models/package-info-cache/package-info.js
rwjblue: Ya, I think we should fix up those tests (they just have bad/invalid setup). We can do that in a separate PR though.

rwjblue (approved):

PR #9563, opened 2021-06-03
Add `pnpm` support to `ember install` command

Turbo87:
see https://pnpm.io/

note that this is only to support `ember install` with `pnpm` for now. this does not add support for e.g. `--pnpm` to `ember new`, which can be done in a follow-up PR.

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -13,7 +13,15 @@ module.exports = Command.extend({
     { name: 'save', type: Boolean, default: false, aliases: ['S'] },
     { name: 'save-dev', type: Boolean, default: true, aliases: ['D'] },
     { name: 'save-exact', type: Boolean, default: false, aliases: ['E', 'exact'] },
-    { name: 'yarn', type: Boolean, description: 'Use --yarn to enforce yarn usage, or --no-yarn to enforce npm' }, // no default means use yarn if the blueprint has a yarn.lock
+    {
+      name: 'package-manager',
+      type: ['npm', 'pnpm', 'yarn'],
+      description:
+        'Use this option to force the usage of a specific package manager. ' +
+        'By default, ember-cli will try to detect the right package manager ' +
+        'from any lockfiles that exist in your project.',
+      aliases: [{ yarn: 'yarn' }, { pnpm: 'pnpm' }],
+    },
   ],
 
   anonymousOptions: ['<addon-name>'],
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -142,20 +142,20 @@ class NpmTask extends Task {
       return this.checkNpmVersion();
     }
 
-    if (!this.hasYarnLock()) {
+    if (this.hasYarnLock()) {
+      logger.info('yarn.lock found -> trying yarn');
+      try {
+        const yarnResult = await this.checkYarn();
+        logger.info('yarn found -> using yarn');
+        return yarnResult;
+      } catch (_err) {
+        logger.info('yarn not found -> using npm');
+      }
+    } else {
       logger.info('yarn.lock not found -> using npm');
-      return this.checkNpmVersion();
     }
 
-    logger.info('yarn.lock found -> trying yarn');
-    try {
-      const yarnResult = await this.checkYarn();
-      logger.info('yarn found -> using yarn');
-      return yarnResult;
-    } catch (_err) {
-      logger.info('yarn not found -> using npm');
-      return this.checkNpmVersion();
-    }
+    return this.checkNpmVersion();
   }
 
   async run(options) {
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -29,7 +29,7 @@ class NpmTask extends Task {
   }
 
   get packageManagerOutputName() {
-    return this.useYarn ? 'Yarn' : 'npm';
+    return this.packageManager.name;
   }
 
   npm(args) {
@@ -58,7 +58,7 @@ class NpmTask extends Task {
         logger.info('yarn --version: %s', version);
       }
 
-      return { yarnVersion: version };
+      return { name: 'yarn', version };
     } catch (error) {
       logger.error('yarn --version failed: %s', error);
 
@@ -97,7 +97,7 @@ class NpmTask extends Task {
         );
       }
 
-      return { npmVersion: version };
+      return { name: 'npm', version };
     } catch (error) {
       logger.error('npm --version failed: %s', error);
 
@@ -154,8 +154,7 @@ class NpmTask extends Task {
   }
 
   async run(options) {
-    let result = await this.findPackageManager({ useYarn: options.useYarn });
-    this.useYarn = Boolean(result.yarnVersion);
+    this.packageManager = await this.findPackageManager({ useYarn: options.useYarn });
 
     let ui = this.ui;
     let startMessage = this.formatStartMessage(options.packages);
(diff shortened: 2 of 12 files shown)


rwjblue (commented):
In general, I'm 👍

lib/tasks/npm-task.js
```diff
@@ -28,6 +29,10 @@ class NpmTask extends Task {
     this.versionConstraints = '3 || 4 || 5 || 6';
   }
 
+  get packageManagerOutputName() {
+    return this.packageManager.name === 'yarn' ? 'Yarn' : 'npm';
```
rwjblue, line 33: Still seems like `pnpm` needs to be added here

Turbo87:
I still want to add a few more tests but wanted to wait first if CI is actually green with the existing ones. Unfortunately CI seems to take a significant amount of time... 😅

lib/tasks/npm-task.js
```diff
@@ -124,37 +158,50 @@ class NpmTask extends Task {
...
-    if (this.useYarn === false) {
+    if (useYarn === false) {
       logger.info('npm requested -> using npm');
       return this.checkNpmVersion();
     }
 
-    if (!this.hasYarnLock()) {
-      logger.info('yarn.lock not found -> using npm');
-      return this.checkNpmVersion();
+    if (this.hasYarnLock()) {
```
rwjblue, line 172: I think we should swap this around to look for `package-lock.json` actually, since `npm@7` supports maintaining a `yarn.lock` as well.

lib/tasks/npm-task.js
```diff
@@ -124,37 +158,50 @@ class NpmTask extends Task {
    * @method findPackageManager
    * @return {Promise}
    */
-  async findPackageManager() {
-    if (this.useYarn === true) {
+  async findPackageManager({ useYarn } = {}) {
```
rwjblue, line 161: I'd like for this to be modified (e.g. `packageManager` string instead of a boolean for `useYarn`).

lib/tasks/npm-task.js
```diff
@@ -165,31 +212,37 @@ class NpmTask extends Task {
...
+        // * https://github.com/npm/npm/issues/16853
+        // * https://github.com/npm/npm/issues/17379
+        //
+        // this ensures that we run a full `npm install` **after** any `npm
+        // install foo` runs to ensure that we have a fully functional
+        // node_modules hierarchy
+        let npmVersion = this.packageManager.version;
+        if (npmVersion && semver.lt(npmVersion, '5.7.1')) {
+          await this.npm(['install']);
+        }
```
rwjblue, line 239: We can _probably_ remove this now

lib/tasks/npm-task.js
rwjblue: I was gonna comment on the changes to `new` and `addon` command, but looks like you didn't need to make any modifications.

I'd like to make `--package-manager` option, so you could do `--package-manager=yarn` or `--package-manager=pnpm`. Then `--yarn` would be an alias for `--package-manager=yarn` (and we could add `--pnpm` as an alias of `--package-manager=pnpm`).

lib/tasks/npm-task.js
Turbo87: 🤦 yep, missed that 😅

lib/tasks/npm-task.js
Turbo87: if a `yarn.lock` file is detected I would assume the devs expect to use `yarn` to install it. if `yarn` is not available we fall back to `npm`, which in the case of `npm@7` would then also work. the `versionConstraints` in the class say that only npm 3-6 are supported yet though.

lib/tasks/npm-task.js
Turbo87: yeah, I agree. I wanted to keep the API of the class the same for now though.

lib/tasks/npm-task.js
Turbo87: probably, but it's unrelated to the pnpm stuff, so I'd rather not conflate the two things 😉

lib/tasks/npm-task.js
Turbo87: I've added support for `--package-manager` now.

lib/commands/install.js
```diff
@@ -13,7 +13,15 @@ module.exports = Command.extend({
...
-    { name: 'yarn', type: Boolean, description: 'Use --yarn to enforce yarn usage, or --no-yarn to enforce npm' }, // no default means use yarn if the blueprint has a yarn.lock
+    { name: 'yarn', type: Boolean, description: 'This is deprecated. Use --package-manager instead.' }, // no default means use yarn if the blueprint has a yarn.lock
+    {
+      name: 'package-manager',
+      type: String,
+      description:
+        'Use this option to force the usage of a specific package manager. ' +
+        'By default, ember-cli will try to detect the right package manager ' +
+        'from any lockfiles that exist in your project. Supported values: npm, yarn, pnpm',
+    },
```
rwjblue, line 24: Hmm, seems like we don't have to deprecate `--yarn`. I was thinking we could do something like:

https://github.com/ember-cli/ember-cli/blob/70ae16c5eb972076580b23e9c22579234749862e/lib/commands/build.js#L16

```suggestion
    {
      name: 'package-manager',
      type: String,
      description:
        'Use this option to force the usage of a specific package manager. ' +
        'By default, ember-cli will try to detect the right package manager ' +
        'from any lockfiles that exist in your project. Supported values: npm, yarn, pnpm',
      aliases: [{ yarn: 'yarn' }, { pnpm: 'pnpm' }],
    },
```

lib/tasks/npm-task.js
```diff
@@ -28,6 +29,10 @@ class NpmTask extends Task {
     this.versionConstraints = '3 || 4 || 5 || 6';
   }
 
+  get packageManagerOutputName() {
+    return this.packageManager.name === 'yarn' ? 'Yarn' : this.packageManager.name;
```
rwjblue, line 33: Seems odd to capitalize here, why can't we just use this:

```suggestion
    return this.packageManager.name;
```

?

Turbo87: Yarn capitalizes the name on their website 🤷‍♂️

rwjblue: K, let's go with my suggestion (don't accept it though because you'll have to make other changes too).

Turbo87:
@rwjblue are you okay with the updated changes? :)

lib/commands/install.js
```diff
@@ -13,7 +13,15 @@ module.exports = Command.extend({
     { name: 'save', type: Boolean, default: false, aliases: ['S'] },
     { name: 'save-dev', type: Boolean, default: true, aliases: ['D'] },
     { name: 'save-exact', type: Boolean, default: false, aliases: ['E', 'exact'] },
-    { name: 'yarn', type: Boolean, description: 'Use --yarn to enforce yarn usage, or --no-yarn to enforce npm' }, // no default means use yarn if the blueprint has a yarn.lock
+    {
+      name: 'package-manager',
+      type: String,
```
rwjblue, line 18: I think this works, right?

```suggestion
      type: ['yarn','npm','pnpm'],
```

Basically, we don't want to allow "any" string; only specfic ones.

lib/commands/install.js
```diff
@@ -13,7 +13,15 @@ module.exports = Command.extend({
...
+      description:
+        'Use this option to force the usage of a specific package manager. ' +
+        'By default, ember-cli will try to detect the right package manager ' +
+        'from any lockfiles that exist in your project. Supported values: npm, yarn, pnpm',
```
rwjblue, line 22: If you use the array of strings above, I don't think you have to have this in the description

Turbo87: yep, you're right. fixed now 👍

rwjblue (approved):

rwjblue:
Thanks @Turbo87!

PR #9558, opened 2021-05-27
Switch from `octane` template lint config to `recommended`

bmish:
ember-template-lint v3 promoted the `octane` config (now deprecated) to be `recommended`.

https://github.com/ember-template-lint/ember-template-lint/blob/master/CHANGELOG.md#v300-2021-03-02

PR #9564, opened 2021-06-03
Move template config from Octane to Recommended

elwayman02:
I've been informed by @scalvert that v3 of `ember-template-lint`, which is already included in this blueprint, moved from "octane" back to "recommended" as the default config, and that "octane" is deprecated.

cc: @MelSumner


rwjblue:
Landed originally in https://github.com/ember-cli/ember-cli/pull/9558, I just haven't merged that back into master just yet (will do shortly).

That change will be in ember-cli@3.27.0 when released.

PR #9575, opened 2021-06-15
Update gitignore to include .lint-todo folder

MelSumner:
Since `ember-template-lint` is installed by default, it would be helpful to have the `.lint-todo` folder already in the .gitignore file. It's consistent with the experience we designed for developers.

### DIFF
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -13,6 +13,7 @@
 /.pnp*
 /.sass-cache
 /.eslintcache
+/.lint-todo
 /connect.lock
 /coverage/
 /libpeerconnection.log


bertdeblock:
The docs state that this folder should be checked in with the other source code.
Adding it to the `.npmignore` file seems like a good idea though?

rwjblue:
Hmm, I think we actually explicitly **need** these files to be included into the git history. That is the only way that todos would be able to work for teams of developers.

scalvert:
It should be in `.npmignore`, not `.gitignore`.

scalvert:
I think we probably could have added this to the [.npmignore blueprint file](https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/npmignore) rather than close this PR. WDYT?

PR #9595, opened 2021-08-07
Add support for `addons.exclude` and `addons.include` options

bertdeblock:
Implements [RFC 639](https://emberjs.github.io/rfcs/0639-replace-blacklist-whitelist.html).

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -84,7 +84,7 @@ class EmberApp {
    - trees, defaults to `{}`
    - jshintrc, defaults to `{}`
    - vendorFiles, defaults to `{}`
-   - addons, defaults to `{ blacklist: [], whitelist: [] }`
+   - addons, defaults to `{ exclude: [], include: [] }`
 
    @class EmberApp
    @constructor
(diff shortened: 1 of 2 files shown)


lib/broccoli/ember-app.js
```diff
@@ -366,6 +370,22 @@ class EmberApp {
 
     this.options = defaultsDeep(options, detectedDefaultOptions, DEFAULT_CONFIG);
 
+    if (this.options.addons.blacklist) {
```
bertdeblock, line 373: Should I set `this.options.addons.blacklist` and `this.options.addons.whitelist` when not defined to respectively `this.options.addons.exclude` and `this.options.addons.include`? Just in case e.g. a project starts using `include` / `exclude` and one of its addons uses `blacklist` / `whitelist` to perform a build-time check.

lib/broccoli/ember-app.js
```diff
@@ -84,7 +84,7 @@ class EmberApp {
    - trees, defaults to `{}`
    - jshintrc, defaults to `{}`
    - vendorFiles, defaults to `{}`
-   - addons, defaults to `{ blacklist: [], whitelist: [] }`
+   - addons, defaults to `{ exclude: [], include: [] }`
```
bertdeblock, line 87: Do we want to keep `blacklist` and `whitelist` mentioned in the docs here?

rwjblue:
@bertdeblock - Thanks for picking this up!!! I think if you rebase, it should fix the CI failures...

lib/broccoli/ember-app.js
rwjblue: No, I think what you have here is good

lib/broccoli/ember-app.js
rwjblue: Let's add a getter and issue a warning if someone does touch `this.options.addons.blacklist` etc

rwjblue:
Thank you @bertdeblock!

PR #9607, opened 2021-09-02
Add online editor files to update-output-repos script

NullVoxPopuli:
adds .stackblitzrc to ember-new-output and ember-addon-output repos.

I figure we can use this new online-editors folder to support whichever online environments/editors we want 🙃


kellyselden:
This isn't the right place for it. These repos are for `ember new` output only. This should go in a separate repo maintained by someone (maybe us).

NullVoxPopuli:
what is the purpose of the new output repos?

I thought it specifically was so that people could start a new project easily by "just cloning"?
(which is exactly what the online editors do).

I just thought it would be way easier to add these sorts of files to ember-new-output and ember-addon-output than to make two duplicate repos per online editor we wanted to support :thinking: :shrug: :upside_down_face:

kellyselden:
The purpose is to see what `ember new` would generate at any point in time at any version. It's only historical reference.

NullVoxPopuli:
ah ok -- that makes sense.
would you be open to a new set of repos for all online editors?
like,
```
ember-cli/ember-new-output-for-online-editors
ember-cli/ember-addon-output-for-online-editors
```
?

rwjblue (commented):
In general, I think this is a good idea but I'm a bit concerned with the (now) in ability to tell what is in the default blueprint with default options vs what is added for ease of consumption.

One possible way to handle that would be to have two repos (one for the static output of the blueprint and one for consumption by things like CodeSandbox, StackBlitz, etc). Another possible way might be to push specific branches for each of the supported services.

Thoughts?

NullVoxPopuli:
doing branches for each of the specific services is something I can do :tada: 
_however_, I think it's useful to be able to have the ability to test on stable/lts/canary via the stable/lts/master branches of ember-new-output :thinking: 
So, a separate repo per service? or a single repo for all services?

NullVoxPopuli:
how about: all online editors use the same set of branches? (reflected in most recent commit)

kellyselden:
This seems pretty good. My only concerns is adding multiple editors. Then you would see codesandbox files showing up in stackblitz and so on. It could be confusing. Thoughts on having branches per editor?

NullVoxPopuli:
Branches per editor make sense to me

NullVoxPopuli:
I ran this a few times locally on my machine, and it works:
```
❯ node dev/update-output-repos.js 
Running ember new my-app
cloning git@github.com:ember-cli/editor-output.git in to /tmp/tmp-122363-2jIYKEWf6TCV
preparing updates for online editors
clearing git@github.com:ember-cli/editor-output.git in /tmp/tmp-122363-2jIYKEWf6TCV/editor-output
copying generated contents to output repo
copying online editor files
commiting updates
pushing commit
Running ember addon my-addon
cloning git@github.com:ember-cli/editor-output.git in to /tmp/tmp-122363-wwhT354l0pCJ
preparing updates for online editors
clearing git@github.com:ember-cli/editor-output.git in /tmp/tmp-122363-wwhT354l0pCJ/editor-output
copying generated contents to output repo
copying online editor files
commiting updates
pushing commit

```

NullVoxPopuli:
so, this is ready to review / prepare for merge?

cc @kellyselden @rwjblue

kellyselden:
Thanks!

Issue #9606, opened 2021-08-30
`ember install` fails on first attempt, succeeds on second (ember-cli 3.28.0)


lfrost:
Thank you for taking the time to open an issue!

For bug reports please include the following section in your issue details.
If you include instructions on how to reproduce the bug or a failing test case
it will make it easier for us to track down the issue you're having.

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.28.0
node: 14.17.5
v8: 8.4.371.23-node.76
uv: 1.41.0
zlib: 1.2.11
brotli: 1.0.9
ares: 1.17.2
modules: 83
nghttp2: 1.42.0
napi: 8
llhttp: 2.1.3
openssl: 1.1.1k
cldr: 39.0
icu: 69.1
tz: 2021a
unicode: 13.0
os: linux x64
6.14.14
1.22.11
```
*Description*
`ember install` fails on first attempt, succeeds on second.

*Steps to Reproduce*
```
$ ember new --no-welcome --yarn test-app
$ cd test-app
$ ember install ember-composable-helpers

🚧
Yarn: Installed ember-composable-helpers
Install failed. Could not find addon with name: ember-composable-helper
```
This happens on every Ember add-on tested, including ember-cli-typescript, ember-cli-mirage, ember-cli-string-helpers, ember-composable-helpers, and ember-uuid.

bertdeblock:
I'm seeing the exact same issue. Could it be that this PR is related => https://github.com/ember-cli/ember-cli/pull/9487? Since it seems to touch code related to initializing addons.

bertdeblock:
@davecombs 👋 Would you mind checking if #9487 might be related to this issue?

mansona:
So I'm seeing this as an issue when running `ember install` for any addon that has a default blueprint (a blueprint with the same name as the addon). For example when you run `ember install empress-blog` it throws an error 

```
Install failed. Could not find addon with name: empress-blog
```

I have done a `git bisect` and I have confirmed that it is indeed https://github.com/ember-cli/ember-cli/pull/9487 that caused the issue, specifically this commit: https://github.com/ember-cli/ember-cli/pull/9487/commits/22b799b51410546d8e07685c9b621da8c11998d7

Unfortunately this is a **massive** PR and I'm wondering if we can actually just revert it while we figure out what went wrong? I don't even know how to test this to make sure that it doesn't break 🙈

bertdeblock:
[...]

PR #9608, opened 2021-09-02
Revert "adding PerBundleAddonCache feature"

mansona:
This reverts commit 22b799b51410546d8e07685c9b621da8c11998d7 which is the merge commit for https://github.com/ember-cli/ember-cli/pull/9487

~This is a very naive revert. I'm not really sure what I am looking at when I was resolving the conflicts 🙈 I kinda just wanted to see if this will pass CI and open the discussion~

So it turns out there was a lot of follow-on work on this feature and I needed to do a bit more of an involved revert 😫 I have no idea what I'm looking at, I'm just trying to pull out the whole feature (which is unlikely what we actually want) and to see if I can get the tests to still pass

(hopefully) Fixes https://github.com/ember-cli/ember-cli/issues/9606


mansona:
On the plus this does seem to fix the issue when running it locally 😂

ef4:
I am in favor of this reversion. I am very skeptical about the whole proxy-based thing and if I had seen it while it was happening I would have advised against it.

It broke ember-engines (https://github.com/ember-engines/ember-engines/pull/769) and required changes there.

It broke embroider (https://github.com/embroider-build/embroider/pull/941) and required changes there.

I suspect we will find more fallout as people test more addons. If we're going to incur this much addon ecosystem breakage, we should **fix the underlying problem instead of patching over it** with proxies. The underlying problem is that it makes no sense to instantiate addons per-consumption rather than per on-disk copy. The vast majority of addons don't care either way and we can automatically detect that they don't care. We even already have code for that, in embroider, that could also be used directly by ember-cli.

mansona:
Thanks for the comment @ef4 🎉 I appreciate more of the context but also I'm glad that I'm not going crazy suggesting something like this 😂 

The question is: how do we move forward from here? Do we need to get more consensus to get this merged? I'll see if I can get CI passing so at least we're not going to have that holding us back, but other than that I am struggling to know what to do to help this get merged. 

I'm also concerned about a protracted discussion about this topic when we know it is currently breaking a huge part of the ecosystem 🙈

mansona:
Oh actually it looks like CI is failing on master in the same place 🤔 https://github.com/ember-cli/ember-cli/runs/3438949128#step:5:1964

Do we know if embroider is working against ember-cli master right now? is it a known issue?

kategengler:
embroider is not currently working against master, see https://github.com/ember-cli/ember-cli/pull/9597

kategengler:
pinging @rwjblue as a reviewer here

bertdeblock:
Is this PR still needed since a fix for #9606 was merged and released?

kategengler:
I agree this is no longer needed since #9606

PR #9610, opened 2021-09-03
Failing test default blueprint

mansona:
This adds a failing test for the default blueprint issue in #9606 

I needed to fix something else to be able to `ember install` from a folder and have it run the default blueprint but this is a valid test that recreates the issue 👍 

~I will open a new PR that targets the same test to the 3.27 branch to show that the test actually passes~

Edit: there isn't actually a branch for the 3.27 release so I can't open a PR but I can confirm that it works fine in 3.27 and you can run it manually if you clone my branch https://github.com/mansona/ember-cli/tree/check-backport

### DIFF
--- a/lib/tasks/addon-install.js
+++ b/lib/tasks/addon-install.js
@@ -4,6 +4,8 @@ const Task = require('../models/task');
 const SilentError = require('silent-error');
 const merge = require('ember-cli-lodash-subset').merge;
 const getPackageBaseName = require('../utilities/get-package-base-name');
+const fs = require('fs');
+const path = require('path');
 
 class AddonInstallTask extends Task {
   constructor(options) {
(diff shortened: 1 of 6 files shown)


bertdeblock:
I'm no `ember-cli` expert, but I was under the impression that the issue is caused by something within the following call:

```js
let addon = this.project.findAddonByName(packageName);
```

If I install e.g. `ember-truth-helpers` with v3.27 and v3.28, `givenName` and `packageName` are the same for both versions, but it's the result of `findAddonByName` that's different (`null` vs the actual addon instance).

I'm just guessing though.

bertdeblock:
Do we still need this one? #9606 was resolved a while ago in v3.28.1

bertdeblock:
Going to close this since #9606 was resolved in v3.28.1.

PR #9611, opened 2021-09-05
use more standard markdown for addon readme

mansona:
This just standardises the type of headers used in the addon README to be using the normal `# Header 1` and `# Header 2` markdown style that is much more widely used.

In fact this standardises the style to be the same as the app readme and also the same as the contributing file in the addon blueprint 😂


locks:
Thank you! The team decided to go with this change since Prettier seems to agree with the proposed syntax and we want to avoid unnecessary changes when the user saves the file.

PR #9612, opened 2021-09-06
Embroider update

mansona:
This is just an extension of https://github.com/ember-cli/ember-cli/pull/9597 but updating to `v0.44.0` so that it might fix tests (because https://github.com/embroider-build/embroider/pull/948 was released in that version)


stefanpenner:
Let's keep the original PR as the place to collaborate on this. I've added you as a contributor, so that should make it possible for you.

PR #9613, opened 2021-09-08
Fix test that started failing with v2.17.0 of qunit

kategengler:
Change test to look for output that asserts the same results, the
underlying behavior of qunit changed in version 2.17.0 and altered the
result.

### DIFF
--- a/tests/acceptance/brocfile-smoke-test-slow.js
+++ b/tests/acceptance/brocfile-smoke-test-slow.js
@@ -171,6 +171,7 @@ describe('Acceptance: brocfile-smoke-test', function () {
     await runCommand(path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember'), 'test');
   });
 
+  // See https://github.com/ember-cli/ember-cli/issues/9615 if test begins to fail; qunit had a regression
   it('addon trees are not jshinted', async function () {
     await copyFixtureFiles('brocfile-tests/jshint-addon');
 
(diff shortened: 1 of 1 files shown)


kategengler:
Submitted this issue to qunit https://github.com/qunitjs/qunit/issues/1651

stefanpenner:
This unblocks an issue, but it seems like we may want to restore the original functionality if qunit fixes this issue. Given that, could you leave an inline commit + an issue so we eventually get back to this?

stefanpenner (approved):

tests/acceptance/brocfile-smoke-test-slow.js
```diff
@@ -183,9 +183,8 @@ describe('Acceptance: brocfile-smoke-test', function () {
 
     let ember = path.join('.', 'node_modules', 'ember-cli', 'bin', 'ember');
 
-    let error = await expect(runCommand(ember, 'test', '--filter=jshint')).to.eventually.be.rejected;
-
-    expect(error.output.join('')).to.include('Error: No tests matched the filter "jshint"');
+    let response = await expect(runCommand(ember, 'test', '--filter=jshint')).to.eventually.be.ok;
+    expect(response.output.join('')).to.include('tests 0');
```
Krinkle, line 187: Hi :)

I'm trying to connect the dots between this and https://github.com/qunitjs/qunit/issues/1651.

Is the response output here by any chance based on the DOM's `Node.textContent` in some way? It appears not, but if not then I'm not exactly sure why this changed.

stefanpenner:
'

stefanpenner:
@kategengler alternatively, should we pin to an older QUnit for the time beind?

I'm open to either approach, just curious as to your thoughts

tests/acceptance/brocfile-smoke-test-slow.js
kategengler: `ember test` runs the qunit browser runner through `testem` and would reject when running and having a failing test.

kategengler:
@stefanpenner I think we should just change the test -- the qunit is in the blueprints and it wouldn't prevent our users from upgrading it.

PR #9614, opened 2021-09-08
fixturify-project should be a devDep

stefanpenner:


stefanpenner:
nevermind, i see it now :P

PR #9617, opened 2021-09-09
Fix comment consistency

charlesfries:
Cleanup of recent commits

### DIFF
--- a/blueprints/app/files/.prettierignore
+++ b/blueprints/app/files/.prettierignore
@@ -13,7 +13,6 @@
 # misc
 /coverage/
 !.*
-.*/
 .eslintcache
 
 # ember-try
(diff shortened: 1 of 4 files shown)


blueprints/app/files/.prettierignore
```diff
@@ -13,6 +13,7 @@
 # misc
 /coverage/
 !.*
+.*/
```
kellyselden, line 16: What does this do?

charlesfries: It shields dot-directories from Prettier linting, e.g. the `.git` directory. Although in hindsight I'm realizing that this will keep directories with non-JS contents like `.github` from getting linted, when in reality they probably should.

Will remove.

kellyselden:
Thanks!

PR #9618, opened 2021-09-09
Reload `_packageInfo` as part of `reloadPkg`

brendenpalmer:
This PR resolves https://github.com/ember-cli/ember-cli/issues/9606.

Some context: the issue here is that `_packageInfo` is not properly reset as part of `reloadPkg`, so (without this fix) you get into a state using `ember install` where the package info cache itself is properly reset, but we're still referencing the old package info object on the project instance. In this case, the newly installed package won't exist as expected, which causes us to run into a failure case as part of install.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -594,6 +594,11 @@ class Project {
 
     this.packageInfoCache.reloadProjects();
 
+    // update `_packageInfo` after reloading projects from the `PackageInfoCache` instance
+    // if we don't do this we get into a state where `_packageInfo` is referencing the old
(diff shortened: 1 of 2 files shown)


jeffsawatzky:
Will there be a patch release soon with this fix?

rwjblue:
Published in https://github.com/ember-cli/ember-cli/releases/tag/v3.28.1

jeffsawatzky:
> Published in https://github.com/ember-cli/ember-cli/releases/tag/v3.28.1

Thanks!

Issue #9531, opened 2021-05-10
"Invalid watchman found" when using new watchman timestamp-style versions


ballPointPenguin:
At some point (May 2020 I believe), Facebook stopped using SemVer for Watchman releases and switched to a timestamp-style versioning system. See https://github.com/facebook/watchman/releases
When I run ember-cli commands, such as `ember test` I get an error like:

```
Invalid watchman found, version: [20210328.072220.0] did not satisfy [>= 3.0.0].
Visit https://ember-cli.com/user-guide/#watchman for more info
```

---
Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 3.26.1
node: 12.21.0
v8: 7.8.279.23-node.45
uv: 1.40.0
zlib: 1.2.11
brotli: 1.0.9
ares: 1.16.1
modules: 72
nghttp2: 1.41.0
napi: 7
llhttp: 2.1.3
http_parser: 2.9.4
openssl: 1.1.1j
cldr: 37.0
icu: 67.1
tz: 2019c
unicode: 13.0
os: linux x64
npm: 6.14.11
watchman: 20210508.164824.0
```

gabrielcsapo:
Having a similar issue, since watchman@4.9 is no longer on brew.

rwjblue:
Ya we should remove the guard, basically just check for watchman and assume it’s good enough. The versions older than 3.0.0 are unlikely to be around anymore anyways.

gabrielcsapo:
I have found the culprit and opened and issue https://github.com/ember-cli/watch-detector/issues/12.

I have removed the guard locally and will push up the changes.

gabrielcsapo:
The fix for the issue can be found here https://github.com/ember-cli/watch-detector/pull/13.

rwjblue:
This should be resolved now (the fix was released in watch-detector@1.0.1).

yads:
Will this update be rolled into ember-cli as well?

PR #9619, opened 2021-09-13
Update watch-detector to 1.0.1

colenso:
Mainly because it includes a fix for this issue https://github.com/ember-cli/ember-cli/issues/9531


rwjblue (approved):

rwjblue:
Thank you!

sly7-7:
@rwjblue It's quite an old one, but I encounter this after ugrading watchman, and running ember-cli-3.28.5. It seems like it was not backported to 3.28. That beeing said, I guess ember-cli fallback on an other watcher (NODE) (I don't quite know what I'm talking about). So I wonder if I could just ignore the warning, or if this is worth an other patch release to 3.28 ?

PR #9638, opened 2021-09-15
Update ember-cli's own dependencies to latest

rwjblue:


lupestro:
"ember-cli-preprocess-registry": "^3.3.0" => 4.0.0? would remove dependency on broccoli-clean-css and hence on older version of chalk (and old ansi-regex with CVE on it.)

PR #9673, opened 2021-10-20
Update blueprint dependencies to latest

rwjblue:


rwjblue:
Whoops, accidentally included eslint@8 with this one. I've rolled that back to eslint@7, so I think this should pass CI now...

PR #9678, opened 2021-10-25
Add required dependencies for beta/canary try scenarios

jrjohnson:
ember-source beta currently requires ember-auto-import v2, without this
no addon ember-beta test scenarios will run. As far as I can figure
adding these as requirements for the scenarios will actually enhance
addon compatibility as real errors in Webpack config will be surfaced
instead of a dependency warning.

### DIFF
--- a/blueprints/addon/files/addon-config/ember-try.js
+++ b/blueprints/addon/files/addon-config/ember-try.js
@@ -36,6 +36,8 @@ module.exports = async function () {
         npm: {
           devDependencies: {
             'ember-source': await getChannelURL('beta'),
+            'ember-auto-import': '^2.0.0',
+            webpack: '^5.0.0',
           },
         },
       },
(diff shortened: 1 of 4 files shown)


kategengler:
I believe it is unnecessary to update the blueprint. 

Addons may currently need to make this ember-try config change to pass beta and canary scenarios, but this blueprint change would only make it to an addon when they updated to a new version of ember-cli, probably v4.0, at which point, these ember-try dependencies are no longer necessary because in updating to 4.0, the addon will have added the appropriate dependencies.

jrjohnson:
@kategengler I was hoping to maybe get this into the last `3.28` release so addons can update to that without trouble and then pull it out again with the `4.0` release. It's ok with me to close this and just add it where it's needed in each add-on, but it's a bit of a burden on maintainers reviewing my random update PRs to understand this instead of just trusting the ember-cli-update changes.

PR #9680, opened 2021-10-28
Ignore default folder name used by broccoli-debug

notmessenger:
https://cli.emberjs.com/release/advanced-use/debugging/#logging talks about how debugging is ready to be used and only an environment variable needs to be set to use it.  When a user does so though a `/DEBUG` folder is created, which they then need to ensure does not get checked into their repo.

Keeping this from being checked in can certainly fall within the realm of the end-application's responsibility, as a user should know how to a manage such resources, etc, but I am of the opinion that this folder falls in the same category as other files and paths already being ignored via the _.gitignore_ file, as they are by-products of running commands provided by the Ember CLI ecosystem and not something the user is necessarily aware they are opting into to when running the command.

### DIFF
--- a/blueprints/addon/files/gitignore
+++ b/blueprints/addon/files/gitignore
@@ -1,0 +1,2 @@
+# broccoli-debug
+/DEBUG/
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -27,3 +27,6 @@
 /package.json.ember-try
 /package-lock.json.ember-try
 /yarn.lock.ember-try
+
+# broccoli-debug
+/DEBUG/


rwjblue:
This seems great, can you confirm that this also ends up in the addon blueprint?

locks:
@notmessenger poke!

notmessenger:
> @notmessenger poke!

@locks have pushed changes

locks:
Thanks! I've approved CI and enabled auto-merge.

bertdeblock:
I don't think `blueprints/addon/files/gitignore` should have been added?
The `.gitignore` file for newly generated addons now _only_ contains:

```
# broccoli-debug
/DEBUG/
```

This means that files and folders like `dist`, `node_modules`, ... are not being git-ignored anymore starting from Ember CLI v4.3.

This will also affect all addons that update to v4.3 using `ember-cli-update`:
https://github.com/ember-cli/ember-addon-output/blob/v4.3.0/.gitignore

PR #9694, opened 2021-11-08
test in node 16 LTS

kellyselden:
This should remove the warning for untested node 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x]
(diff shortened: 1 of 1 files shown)


kellyselden:
release branch is already broken https://github.com/ember-cli/ember-cli/runs/3933032539?check_suite_focus=true, so going ahead with merge.

nelstrom:
The [Node Support](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md) documentation lists Node `v14.x` as the latest version to be supported by ember-cli. Would it be appropriate to update that documentation to list `v16.x` as well? Or is there more work to do before we can list `v16.x` as fully supported by ember-cli?

PR #9697, opened 2021-11-16
feat: cancel stale actions workflows

runspired:
This uses Github Action's [concurrency](docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) feature to kill the workflow runs of commits that are not the current
head commit of a PR.  `head_ref` is only set for PRs [source](docs.github.com/en/actions/learn-github-actions/contexts#github-context), which is why we fall back to `ref` (so push workflows will run one job for each branch for the most recent commit).

### DIFF
--- a/blueprints/addon/files/.github/workflows/ci.yml
+++ b/blueprints/addon/files/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ on:
       - master
   pull_request: {}
 
+concurrency:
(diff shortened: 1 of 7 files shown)


rwjblue (approved):
👍  - This is closer to what we had with the old Travis CI setup, and definitely will help prevent a bunch of stray CI jobs blocking the queue if folks are "push happy" on pull requests.

PR #9699, opened 2021-11-21
Add Ember v3.28 LTS to `addon` blueprint (drop support for Ember v3.20 LTS)

bertdeblock:


SergeAstapov:
@bertdeblock do you mind to update PR title as it say v3.24 and we actually drop v3.20

bertdeblock:
Lol 😄, thanks for pointing that out!

kellyselden:
It looks like this will need to be based on the release branch instead of master if we want this released for 3.28 LTS.

PR #9730, opened 2021-12-14
Add Ember v3.28 LTS to `addon` blueprint (drop support for Ember v3.20 LTS)

kellyselden:
Don't release until 4.0 is released. That's when 3.28 becomes LTS.


ef4:
The embroider scenario failure here goes away if you upgrade the embroider packages to the latest (`0.48.1`).

PR #8171, opened 2018-11-07
Add `--ssl` options to `ember test`

nathanhammond:
`testem` supports SSL configuration but we don't expose or forward configuration of that via Ember CLI.

Issue #5029, opened 2015-11-02
Can't run ember test with ssl flag enabled


olivierlesnicki:
Our apps are setting/reading a number of secure cookies (set via our login component). While `ember serve --ssl` is supported, `ember test --ssl` is not.

As a results acceptance tests are failing (the login component set a `secure` cookie but none of the components are able to read it).

Any suggestions?

olivierlesnicki:
So after reading through the `testem` documentation, one can request `testem` to run in https by passing a path to a `server.key` and a `server.certificate` in the `testem.json` like so:

```json
  "key": "ssl/server.key",
  "cert": "ssl/server.crt",
```

However, this configuration causes the `ember test` command to timeout - the app builds and then nothing happen. Running `ember test --server` actually runs the test correctly in all browsers but `phantom`.

Steps to replicate:
- Clone this app https://github.com/olivierlesnicki/ember-test-app-ssl
- Run `ember test`

olivierlesnicki:
Any views on what can be done?

stefanpenner:
@olivierlesnicki support for this needs to be added.

olivierlesnicki:
Adding these properties to `testem.json` achieves the requirement:
```
  "key": "ssl/server.key",
  "cert": "ssl/server.crt",
  "phantomjs_args": [
    "--ignore-ssl-errors=true"
  ],
```

stefanpenner:
> Adding these properties to testem.json achieves the requirement:

PR #9755, opened 2022-01-10
[BUGFIX] fixes `--ssl=false` when using `ember test`

linearza:
At the moment when you have the ssl keys defined in `.ember-cli` such as:
```
{
 ...
  "ssl": true,
  "ssl-key": "ssl/server.key.pem",
  "ssl-cert": "ssl/server.crt.pem"
}
```

And you run `ember test --ssl=false` then the key values are still passed through, causing testem to enable SSL - which in turn fails in CI as it looks for the ssl certs

Original change made here: https://github.com/ember-cli/ember-cli/pull/8171
Related issue here: https://github.com/ember-cli/ember-cli/issues/5029#issuecomment-154475311


tests/unit/tasks/test-test.js
```diff
@@ -107,4 +107,31 @@ describe('test task test', function () {
       sslCert: 'ssl/server.cert',
     });
   });
+
+  it('supports properly disabling SSL by nullifying related keys', function () {
```
bertdeblock, line 111: This test passes without the code changes as well, which makes it hard to tell if these changes are desired.
Also, the expectations in this test are already covered here:
https://github.com/ember-cli/ember-cli/blob/master/tests/unit/tasks/test-test.js#L55-L56

bertdeblock (commented):
I can't seem to reproduce this issue locally.

If I add the following to my `.ember-cli` file:

```
"ssl": true,
"ssl-key": "ssl/server.key.pem",
"ssl-cert": "ssl/server.crt.pem",
```

and run `ember test --ssl=false`, everything works as expected.

bertdeblock:
Closing for now, feel free to reopen if you would like to pick this back up!

cyril-sf:
@bertdeblock I'm running into this issue as well.

In my case, it fails on the CI because the ssh key and cert are not there at all. 

Try to set
```
"ssl": true,
"ssl-key": "ssl/non-existing.key.pem",
"ssl-cert": "ssl/non-existing.crt.pem",
```

and then run `ember test --ssl=false`

PR #9760, opened 2022-01-18
Add `timeout-minutes` to GitHub CI jobs in `app` and `addon` blueprints

mansona:
This PR adds `timeout-minutes: 5` to each of the tests that can mess with the dependencies of an addon. This is because you can have specific cases where a floating dependency might cause the ember tests to fail but not return correctly, just causing Github to run them each for the default **6 Hours!!** 😱 (why is it such a large default timeout GitHub!?) 

I guess we could make this a general policy and I could add it to all of the blueprints that have any Github CI jobs? What do we think?


blueprints/addon/files/.github/workflows/ci.yml
```diff
@@ -33,6 +33,7 @@ jobs:
   floating:
     name: "Floating Dependencies"
     runs-on: ubuntu-latest
+    timeout-minutes: 5
```
rwjblue, line 36: let's make this 10min, that seems like a better default than 5 to me

rwjblue (changes requested):
Overall 👍  here. Some questions:

* why is this not adding a timeout to the main "Tests" job?
* Can you update the associated tests so that CI will pass (we have a number of fixtures around the `ci.yml` contents)
* can you update the app blueprint as well?

bertdeblock:
@mansona Are you still interested in working on this? If you are short on time, I can also open a new PR and get this merged if you are okay with that?

mansona:
Sorry for the delay folks 🙈  I have a few too many PRs open out there in the world and I should do a bit more maintenance a bit more often 😉 

I've updated the PR as requested and updated the timeout defaults to 10 mins 👍

blueprints/addon/files/.github/workflows/ci.yml
mansona: So I was feeling that this was probably a good idea for the **app** tests but I can see a benefit for the default to be the same for everything. This PR really isn't designed to cut things off in the most efficient way, just to prevent the absolute worst case scenario, and 10 mins is a lot less than 6 hours 😂

bertdeblock (approved):

rwjblue (approved):

PR #9737, opened 2021-12-16
Remove jQuery integration scenario from ember-try

NullVoxPopuli:
The release version of: https://github.com/ember-cli/ember-cli/pull/9736

PR #9761, opened 2022-01-18
fix: ember-default-with-jquery not passing in new addons

knownasilya:
Previously it ran against default which was < v4 and so it worked, but now the default is v4 and this test fails.


bertdeblock:
Scenario has now been removed here: https://github.com/ember-cli/ember-cli/pull/9737.

rwjblue:
Thanks for opening the PR @knownasilya! Closing as it's landed in release branch though...

knownasilya:
Nice!

Issue #9765, opened 2022-01-23
Deprecation warning for locationType=auto


denik1981:
**What did happen?**

A fresh installation using `ember-cli` throws a deprecation warning from ember.js for using `locationType="auto"`.
```
DEPRECATION: Router location 'auto' is deprecated. 
Most users will want to set `locationType` to 'history' in config/environment.js for no change in behavior. 
See deprecation docs for details. 
[deprecation id: deprecate-auto-location] 
See https://emberjs.com/deprecations/v4.x#toc_deprecate-auto-location for more details.
```

**What should happen?**

New apps generated with ember-cli should stay in sync with ember.js LTS list of deprecations. 
The deprecation warning came in ember 4.x.
    
**Steps to reproduce the issue**

- `ember new my-new-ember-app`
-  `cd my-new-ember-app`
- `ember serve`
- Navigate to [http://localshost:4200](http://localshost:4200)
- Check the console in your browser devtools.

**References:**
[ember-v4-deprecate-auto-location](https://emberjs.com/deprecations/v4.x#toc_deprecate-auto-location)

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 4.1.0
node: 16.13.2
v8: 9.4.146.24-node.14
uv: 1.42.0
zlib: 1.2.11
brotli: 1.0.9
ares: 1.18.1
modules: 93
nghttp2: 1.45.1
napi: 8
llhttp: 6.0.4
openssl: 1.1.1l+quic
cldr: 39.0
icu: 69.1
tz: 2021a
unicode: 13.0
ngtcp2: 0.1.0-DEV
nghttp3: 0.1.0-DEV
os: linux x64
8.1.2
1.22.17
```

denik1981:
I have submitted PR #9766 to address this issue

bertdeblock:
It seems you've created the same issue twice. Also, I think this should already be addressed by https://github.com/ember-cli/ember-cli/pull/9739.

denik1981:
GH sent me to an error page the first time I submitted the issue so I created it again. 
I didn't though it was created. I will close it.

> Also, I think this should already be addressed by #9739.
Seems so. It was looking for this in the pool of issues (not in the PRs).
Double-check for #9739 because I'm not seeing in there the code needed to adapt CI tests.

denik1981:
From I found in the merged code, this twot test utilities are still referencing `locationType:auto` in the release branch.

[...]

PR #9766, opened 2022-01-23
Fix console warnings for deprecated locationType='auto' in ember 4.x. #9765

denik1981:
Fixes: #9765

**Why is it needed?**
As mentioned in **ember 4.x** deprecations list:

>Several years ago only some browsers supported the (then) new History Location API. Others could only serialize the router location into an url hash my/path#/ember/route. To handle this dynamically, Ember built an AutoLocation that would feature-detect and use either 'hash' or 'history' as the underlying mechanism. These days, virtually all browsers support the history API, so this is what 'auto' will resolve to in almost every case. Since 'auto' has served its purpose, it's being removed. 

**What it does do?**
- It complies with the suggested adaptations ember suggests for this deprecated feature.
- It replaces the location type from `auto` to `history` in both, the template app and in all tests.

>Set locationType in config/environment.js to 'history'. This is it, your app should work just like it used to (*)

_* from ember 4.x deprecations list._

**References:**
[ember-v4-deprecate-auto-location](https://deprecations.emberjs.com/v4.x/#toc_deprecate-auto-location)

### DIFF
--- a/blueprints/app/files/config/environment.js
+++ b/blueprints/app/files/config/environment.js
@@ -5,7 +5,7 @@ module.exports = function (environment) {
     modulePrefix: '<%= modulePrefix %>',
     environment,
     rootURL: '/',
-    locationType: 'auto',
+    locationType: 'history',
     EmberENV: {
       FEATURES: {
         // Here you can enable experimental features on an ember canary build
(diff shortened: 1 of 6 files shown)


kategengler:
This was handled in https://github.com/ember-cli/ember-cli/pull/9739 and released in v4.1.1

Thank you!

PR #9768, opened 2022-01-26
feat: Add support for TypeScript blueprints

cafreeman:
This PR implements the `ember-cli` side of https://github.com/emberjs/rfcs/pull/776 by adding support for generating blueprints authored in TypeScript and outputting either JavaScript or TypeScript. 

More specifically, this PR adds the following:
- The capability to down-level blueprints authored in TypeScript and write out a JavaScript file
- The ability to destroy files generated from TypeScript blueprints (regardless of if their output was JS or TS)
- A new `--typescript` flag for generating TypeScript from TypeScript blueprints on-demand
- A `shouldTransformTypeScript` flag on `Blueprint` that indicates whether a blueprint is eligible for down-leveling.
- The capability to look for an `isTypeScriptProject: true` value in `.ember-cli` and generate TypeScript files rather than JavaScript files wherever possible. 

Once this PR is merged, there will be a follow-on PR to [the ember.js repo](https://github.com/emberjs/ember.js) that actually converts the existing blueprints to TypeScript.


lib/models/blueprint.js
```diff
@@ -196,6 +197,17 @@ let Blueprint = CoreObject.extend({
 
   _printableProperties: ['name', 'description', 'availableOptions', 'anonymousOptions', 'overridden'],
 
+  // Indicates whether or not a blueprint is a candidate for automatic transpilation from TS to JS.
+  // This property could be false in the case that the blueprint is written in JS and is not intended
+  // to work with TS at all, OR in the case that the blueprint is written in TS and the author does
+  // not intend to support transpilation to JS.
```
rwjblue, line 203: Lets move this down into the YUIDoc comment just below

lib/models/blueprint.js
```diff
@@ -429,13 +441,72 @@ let Blueprint = CoreObject.extend({
...
+    // them the JS output
+    if (options.typescript && !isTypeScriptFile(fileInfo.inputPath)) {
+      this.ui.writeLine(
+        chalk.yellow(
+          "You passed the '--typescript' flag but there is no TypeScript blueprint available. " +
+            'A JavaScript blueprint will be generated instead.'
+        )
+      );
+      return false;
+    }
```
rwjblue, line 485: Shouldn't this be above the short circuit return just above?

lib/models/blueprint.js
```diff
@@ -429,13 +441,72 @@ let Blueprint = CoreObject.extend({
...
+
+    // If it's a TypeScript file, we transform it *unless* the user has asked for TypeScript either
+    // globally (by setting `isTypeScriptProject` to true) or locally (by passing the `--typescript`
+    // flag when they invoked the generator).
+    return !(options.isTypeScriptProject || options.typescript) && isTypeScriptFile(fileInfo.inputPath);
```
rwjblue, line 490: Let's beef this comment up slightly to indicate where we expect this to come from, and why it **isn't** just using `{ "typescript": true }`.

Also, maybe split into a variable declaration so we can explain the flag independently of the logic explanation.

rwjblue: We should also consider adding `isTypeScriptProject` to the default `.ember-cli` file (with the default value being `false`)...

tests/acceptance/typescript-destroy-test.js
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  it('deletes TS files generated from typescript blueprints in a typescript project', async function () {
+    let commandArgs = ['foo', 'bar'];
+    let files = ['app/foos/bar.ts'];
+    await initApp();
+
+    await fs.writeJson('.ember-cli', {
+      isTypeScriptProject: true,
+    });
```
rwjblue, line 120: Can you make a test _just_ like this test, but setting `{ "typescript": true }`. The output / results should be same, but I just wanna confirm/assert that the "expected" (yes, I know it's weird, but some folks internalize it this way).

tests/acceptance/typescript-generate-test.js
```diff
@@ -0,0 +1,217 @@
...
+
+    const testGenerated = file('tests/foos/bar-test.js');
+    expect(testGenerated).to.contain('export default function bar(a, b) {');
+  });
+
+  it('outputs typescript when isTypeScriptProject is true', async function () {
+    await initApp();
+    await fs.writeJSON('.ember-cli', {
+      isTypeScriptProject: true,
+    });
```
rwjblue, line 92: Same as the other comment, can you confirm that writing `{ "typescript": true }` to `.ember-cli` has the same output behaviors here?

lib/models/blueprint.js
```diff
@@ -1630,11 +1701,31 @@ function finishProcessingForInstall(infos) {
   return sequence(infosNeedingConfirmation).then(() => infos);
 }
 
-function finishProcessingForUninstall(infos) {
-  let validInfos = infos.filter((info) => fs.existsSync(info.outputPath));
+function finishProcessingForUninstall(infos, shouldTransformTypeScript) {
+  let validInfos = infos.filter((info) => {
+    if (shouldTransformTypeScript && isTypeScriptFile(info.outputPath)) {
```
rwjblue, line 1706: Consider (and let me know) if we should incorporate the `--typescript` flag value here also.

rwjblue (changes requested):
Overall looks great! Thanks for pushing this forward!!!

blueprints/app/files/.ember-cli
```diff
@@ -5,5 +5,11 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+
+  /**
+  Setting `isTypeScriptProject` to true will force the blueprint generators to generate TypeScript
+  rather than JavaScript by default
```
rwjblue, line 12:
```suggestion
  Setting `isTypeScriptProject` to true will force the blueprint generators to generate TypeScript
  rather than JavaScript by default, when a TypeScript version of a given blueprint is available.
```

lib/models/blueprint.js
```diff
@@ -430,12 +443,82 @@ let Blueprint = CoreObject.extend({
...
+
+    // Indicates when the user has asked for TypeScript either globally (by setting
+    // `isTypeScriptProject` to true) or locally (by passing the `--typescript` flag when they
+    // invoked the generator). Since ember-cli merges `.ember-cli` and the all of the flag values into
+    // one object, we could theoretically *just* have a `typescript` flag that is either passed in
+    // to `generate` OR set in `.ember-cli`, but we thought the DX would be improved by differentiating
+    // between what is intended to be global vs. local config. Note that, because of this behavior,
+    // the user could *technically* also just set `typescript: true` in `.ember-cli` or pass
+    // `--is-typescript-project` when invoking `generate`, but we don't expect this to happen often.
+    // const userRequestedTypeScript = options.typescript !== false && options.isTypeScriptProject;
```
rwjblue, line 495: 🔪

lib/models/blueprint.js
```diff
@@ -430,12 +443,82 @@ let Blueprint = CoreObject.extend({
...
+    }
+
+    // Indicates when the user has asked for TypeScript either globally (by setting
+    // `isTypeScriptProject` to true) or locally (by passing the `--typescript` flag when they
+    // invoked the generator). Since ember-cli merges `.ember-cli` and the all of the flag values into
+    // one object, we could theoretically *just* have a `typescript` flag that is either passed in
+    // to `generate` OR set in `.ember-cli`, but we thought the DX would be improved by differentiating
+    // between what is intended to be global vs. local config. Note that, because of this behavior,
+    // the user could *technically* also just set `typescript: true` in `.ember-cli` or pass
+    // `--is-typescript-project` when invoking `generate`, but we don't expect this to happen often.
```
rwjblue, line 494: Can you rework this comment slightly? Specifically removing the alternative that we discarded, and just focusing on the goal / intent of the actual code we are going with...

lib/models/blueprint.js
```diff
@@ -430,12 +443,82 @@ let Blueprint = CoreObject.extend({
...
+    // const userRequestedTypeScript = options.typescript !== false && options.isTypeScriptProject;
+    let userRequestedTypeScript = !noTypeScript && (options.typescript === true || options.isTypeScriptProject);
+
+    // If it's a TypeScript file, we transform it *unless* the user has specifically requested
+    // TypeScript locally or globally (see the comment above for more info)
+    return !userRequestedTypeScript && isTypeScriptFile(fileInfo.inputPath);
```
rwjblue, line 500:
```suggestion
    let userExplicitlySelectedTypeScriptStatus = options.typescript !== undefined;
    
    let shouldUseTypeScript = userExplictlySelectedTypeScriptStatus
      ? options.typescript
      : options.isTypeScriptProject;
      
    if (shouldUseTypeScript) {
      return isTypeScriptFile(fileInfo.inputPath);
    }

    return false;
```

lib/models/blueprint.js
```diff
@@ -770,7 +864,45 @@ let Blueprint = CoreObject.extend({
 
     this._ignoreUpdateFiles();
 
-    return finishProcessingForUninstall(fileInfos.filter(isValidFile));
+    fileInfos = fileInfos.filter(isValidFile).reduce((acc, info) => {
+      // if it's possible that this blueprint could have produced either typescript OR javascript, we have to do some
+      // work to figure out which files to delete.
+      if (this.shouldTransformTypeScript) {
+        if (this.options.typescript === true) {
```
rwjblue, line 871: lets make `shouldTransformTypeScript === false && options.typescript === true` a nice error (using `SilentError`) for the user (telling them that they asked to delete the `--typescript` version of this blueprint, but it doesn't support typescript anyways)

tests/acceptance/help-test.js
```diff
@@ -121,6 +121,12 @@ describe('Acceptance: ember help', function () {
     let output = options.ui.output;
 
     let fixturePath = path.join(__dirname, '..', 'fixtures', 'help', 'generate-blueprint.txt');
+
+    // makes updating this fixture much much easier...
+    if (process.env.WRITE_HELP_FIXTURES) {
```
rwjblue, line 126: 👍

rwjblue (changes requested):

lib/models/blueprint.js
```diff
@@ -757,6 +840,17 @@ let Blueprint = CoreObject.extend({
...
       .then(finishProcessingForInstall)
+      .then((fileInfos) => {
+        return Promise.all(
+          fileInfos.map((info) => {
+            if (this.shouldConvertToJS(this.options, info)) {
+              return this.convertToJS(info);
+            }
+
+            return info;
+          })
```
rwjblue, line 851: for the polyfill here, you'd have to do something like this:

```js
init() {
  this._super.init.apply(this, arguments);

  const originalWTFBBQ = this.processFiles;
  this.processFiles = function() {
     // supering
    let upstreamResult = originalWTFBBQ.apply(this, arguments);
    return upstreamResult.then((fileInfos) => {
      // basically the code just above
    });
  };
}
```

lib/models/blueprint.js
```diff
@@ -770,7 +864,45 @@ let Blueprint = CoreObject.extend({
...
+      if (this.shouldTransformTypeScript) {
+        if (this.options.typescript === true) {
+          // if the user explicitly passed `--typescript`, we only want to delete TS files, so we stick with the existing
+          // info object since it will contain a .ts outputPath (since we know this blueprint is authored in TS because
+          // of our check above)
+          acc.push(info);
+          return acc;
+        }
+
+        const jsInfo = new FileInfo({
```
rwjblue, line 879: You can probably do this in the polyfill (to avoid making a require on the internal `file-info.js`)
```js
let jsInfo = Object.create(info);
Object.assign(jsInfo, { /* new stuff */ });
```

lib/models/blueprint.js
```diff
@@ -770,7 +864,45 @@ let Blueprint = CoreObject.extend({
 
     this._ignoreUpdateFiles();
 
-    return finishProcessingForUninstall(fileInfos.filter(isValidFile));
+    fileInfos = fileInfos.filter(isValidFile).reduce((acc, info) => {
```
rwjblue, line 867: For the polyfill, you can move htis logic into `_getFileInfos` directly (you will have to test that you are in the uninstall pathway before doing the new logic, but you should be able to determine that fairly easily)

rwjblue (commented):
Here is some info on how to polyfill

lib/models/blueprint.js
cafreeman: There's a possibility that people could have blueprints that are *exclusively* TypeScript, in which case this error message wouldn't end up being accurate.

That said, we're not at any risk of accidentally deleting someone's JS files in this scenario since `--typescript` forces the command to only look for `.ts` files. It'll simply overlook the JS files.

lib/models/blueprint.js
cafreeman: The logic here ended up being a bit more involved, but I took this suggestion and tweaked it to work with the logic so that it's still much easier to parse.

lib/models/blueprint.js
```diff
@@ -430,12 +443,85 @@ let Blueprint = CoreObject.extend({
...
+    // `isTypeScriptProject` to true) or locally (by passing the `--typescript` flag when they
+    // invoked the generator). Although ember-cli merges `.ember-cli` and all of the flag values into
+    // one object, we thought the DX would be improved by differentiating between what is intended
+    // to be global vs. local config.
+    const shouldUseTypeScript = userExplicitlySelectedTypeScriptStatus
+      ? options.typescript
+      : options.isTypeScriptProject;
+
+    // if the user wants TS output and we have a TS file available, we do *not* want to downlevel to JS
+    if (shouldUseTypeScript && isTypeScriptFile(fileInfo.inputPath)) {
```
rwjblue, line 499: Aren't we already inside an `isTypeScriptFile` guard from 466 ? (maybe I misread the indentation levels in the diff)

rwjblue (approved):

lib/models/blueprint.js
cafreeman: Yep, good catch. Removed the guard here.

PR #9770, opened 2022-01-28
added mention of the addon that contains a command override to console output

davecombs:
When CLI prints an error message saying that a particular addon tried to override a core command (e.g. 'test' is overridden by 'ember-exam'), added the name of the offending addon to the message so we know where to find it. Tested locally

### DIFF
--- a/lib/cli/lookup-command.js
+++ b/lib/cli/lookup-command.js
@@ -27,11 +27,15 @@ module.exports = function (commands, commandName, commandArgs, optionHash) {
   // Attempt to find command in ember-cli core commands
   let command = findCommand(commands, commandName);
 
+  let addonWithCommand;
   let addonCommand;
   // Attempt to find command within addons
   if (project && project.eachAddonCommand) {
     project.eachAddonCommand((addonName, commands) => {
       addonCommand = findCommand(commands, commandName);
+      if (addonCommand) {
+        addonWithCommand = addonName;
+      }
       return !addonCommand;
     });
   }
@@ -39,7 +43,7 @@ module.exports = function (commands, commandName, commandArgs, optionHash) {
   if (command && addonCommand) {
     if (addonCommand.overrideCore) {
       ui.writeWarnLine(
-        `An ember-addon has attempted to override the core command "${command.prototype.name}". ` +
+        `An ember-addon (${addonWithCommand}) has attempted to override the core command "${command.prototype.name}". ` +
           `The addon command will be used as the overridding was explicit.`
       );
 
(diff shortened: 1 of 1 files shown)


bertdeblock:
@davecombs This seems like a helpful addition, are you still interested in getting this merged?
It seems we only need to update the related warnings [here](https://github.com/ember-cli/ember-cli/blob/b24b73b388934796ca915ca665b48a27c857199b/tests/unit/cli/lookup-command-test.js#L140) and [here](https://github.com/ember-cli/ember-cli/blob/b24b73b388934796ca915ca665b48a27c857199b/tests/unit/cli/lookup-command-test.js#L162) to include the addon's name.
The fake addon [here](https://github.com/ember-cli/ember-cli/blob/b24b73b388934796ca915ca665b48a27c857199b/tests/unit/cli/lookup-command-test.js#L23) needs a name as well, otherwise it will log `(undefined)`, which I assume can't happen with real addons.

davecombs:
Thanks, Bert - I'd basically forgotten about this PR from a while back. I'll fix the warnings and the fake addon and update the PR.

davecombs:
@bertdeblock - The failure right now is with a blueprint. Since I didn't do anything with any blueprints, and others besides Windows passed, could you retrigger this or push it?

bertdeblock (approved):

tests/unit/cli/lookup-command-test.js
```diff
@@ -136,9 +137,10 @@ describe('cli/lookup-command.js', function () {
       ui,
     });
 
-    expect(ui.output).to.match(
-      /WARNING: An ember-addon has attempted to override the core command "serve"\. The core command will be used.*/
+    let re = new RegExp(
+      `WARNING: An ember-addon \\(${project.addons[0].name}\\) has attempted to override the core command "serve"\\. The core command will be used.*`
```
kellyselden, line 141: I think using the string FakeAddon would be better here. I would hate for the code to not be working, but the test pass because `project.addons[0].name` was "correctly" undefined.

kellyselden (changes requested):

kellyselden:
Just a test issue, then looks good for me

tests/unit/cli/lookup-command-test.js
davecombs: There are 2 tests like this. I switched both of them to include the hardcoded addon name rather than the expression to fetch the name.

kellyselden: Hmm I'm not seeing the change. Perhaps it was a bad push?

davecombs:
Hi, Kelly. Yes, it was a bad push. I `git add`ed but forgot to `git
commit`! Just pushed the real commit. :-)

Thanks,
Dave

kellyselden (approved):

PR #9775, opened 2022-02-01
Add ability to skip printing files based on user provided regex

### LINKED ISSUE #9763: Exclude files/folders from asset size printer
The asset size printer currently prints everything in the `dist/` directory. We have some files we're putting in there that we don't need to see in our export. Moreover, for \*reasons\*, a single build job actually builds ~10 ember builds, so the output gets really verbose. I'd like to be able to exclude asset sizes for files in some folders based on some configuration.

---

```
ember-cli: 3.23.0
node: 14.17.1
v8: 8.4.371.23-node.67
uv: 1.41.0
zlib: 1.2.11
brotli: 1.0.9
ares: 1.17.1
modules: 83
nghttp2: 1.42.0
napi: 8
llhttp: 2.1.3
openssl: 1.1.1k
cldr: 39.0
icu: 69.1
tz: 2021a
unicode: 13.0
os: darwin x64
# npm
6.14.13
#version
1.22.17
```

mehulkar:
This can be configured using the `.ember-cli` file. I chose this because neither `ember-cli-build.js` or `config/environment.js` seemed appropriate. I'm not 100% sure if this also means that this can be passed as a flag to `ember build` and `ember asset-sizes`.

I saw there aren't any tests for reading these options yet: https://github.com/ember-cli/ember-cli/blob/60d55de1a3cfb2eb751e4e76ac46b722464884a9/tests/unit/commands/show-asset-sizes-test.js#L56-L58. Happy to try to add some if we want them!

Closes #9763.

### DIFF
--- a/lib/commands/asset-sizes.js
+++ b/lib/commands/asset-sizes.js
@@ -9,6 +9,7 @@ module.exports = Command.extend({
   availableOptions: [
     { name: 'output-path', type: 'Path', default: 'dist/', aliases: ['o'] },
     { name: 'json', type: Boolean, default: false },
+    { name: 'asset-sizes-skip-file-regex', type: String, default: null },
   ],
 
   run(commandOptions) {
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -20,6 +20,7 @@ module.exports = Command.extend({
     { name: 'watch', type: Boolean, default: false, aliases: ['w'] },
     { name: 'watcher', type: String },
     { name: 'suppress-sizes', type: Boolean, default: false },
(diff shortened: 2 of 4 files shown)


rwjblue:
Overall this is fine, but we need to add an option to `build` command (so it shows up in help), which would alos make testing much easier. The merging of `.ember-cli` file contents would still work though.

mehulkar:
@rwjblue done. Don't love the flag name, and feel like there's probably opportunity to combine it with the functionality of `--suppress-sizes`, but that feels more invasive (and probably a breaking change that doesn't seem worth it)

PR #9781, opened 2022-02-06
[ENHANCEMENT] Remove `X-UA-Compatible` meta tag for IE browser

bobisjan:
This meta tag should not be needed since IE support was dropped in v4, https://docs.microsoft.com/en-us/openspecs/ie_standards/ms-iedoco/380e2488-f5eb-4457-a07a-0cb1b6e4b4b5


locks:
Nicely spotted, thanks!

PR #9784, opened 2022-02-10
Unskip tests

locks:
The test suite skips various tests for various reasons.
In this PR I attempt to unskip said tests and see what gives.


bertdeblock:
@locks Are you still planning on continuing the work here, or should we close for now?

locks:
The tests I tried to unskip were proving to be quite challenging, so I think it's better to close this!

PR #9823, opened 2022-03-14
[RFC 801] Deprecate `blacklist` and `whitelist` build options

bertdeblock:
RFC PR: https://github.com/emberjs/rfcs/pull/801

~~Putting this in Draft, as the RFC is still in FCP.~~


kategengler:
Thank you!

PR #9824, opened 2022-03-16
[RFC 638] Interactive way to create new Ember apps and addons

bertdeblock:
RFC PR: https://github.com/emberjs/rfcs/pull/638

Went with `inquirer` for the following reasons:
- Development of `enquirer` seems to have stalled
- `prompts` seems nice as well, but `inquirer` is more populair (based on download and GH star count)
- `inquirer` seems to be the most active library at the moment

Result:

https://user-images.githubusercontent.com/7403183/193288180-0b5107fc-ccc1-4a3c-b50f-fd7eab33498f.mov


bertdeblock:
I overlooked some tests, e.g. `ember help` output, will fix asap!

NullVoxPopuli:
excited to try this out!

bertdeblock:
Can't seem to figure out why CI is failing at the moment.
It seems this line is the culprit:
https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/smoke-test-slow.js#L236-L237
It hangs, and thus the timeout is exceeded.
Seems to fail here as well:
https://github.com/ember-cli/ember-cli/runs/5564343698?check_suite_focus=true#step:5:2237

package.json
```diff
@@ -94,6 +94,7 @@
     "heimdalljs-logger": "^0.1.10",
     "http-proxy": "^1.18.1",
     "inflection": "^1.13.1",
+    "inquirer": "^8.2.1",
```
rwjblue, line 97: Currently we get inquirer via console-ui, can we refactor to leverage that?

bertdeblock: We could, but `console-ui` still depends on v6, and v6 doesn't support providing fixed answers, which I'm using during tests. This would also need an API change in `console-ui` I think.
What's the benefit of using `ui.prompt` instead of using `inquirer` directly?
Seems like an unnecessary abstraction at first glance, but maybe I'm missing some context.

NullVoxPopuli:
Really excited for this!!! An interactive mode is so clutch for folks new to programming and are _extremely_ intimidated by help screens and cli flags.

(It's also just a nice way to walk folks through stuff)

package.json
NullVoxPopuli: I second the "unnecessary abstraction" assessment. `ui.prompt`  / `console-ui` does not have very many docs (and it's not clear what's it is using under its APIs), and folks wanting to contribute to the interactive flow in ember-cli will appreciate knowing that `inquirer` is used, which has _toooons of docs_.

rwjblue: Well, we do need to update/clean-up the console-ui stuff more generally, but I think this is fine for now (other than that we have two or more copies of inquirer in our stack, which is a bit annoying).

rwjblue (commented):
Overall this looks great, thanks for working on it!!

I left a couple inline questions/suggestions, and I think we need at least one test that goes through a full flow "acceptance" / "slow" test that actually runs `ember new` and enters the values for the prompts (to make sure that things actually work in reality).

lib/tasks/interactive-new.js
```diff
@@ -0,0 +1,160 @@
...
+          },
+        ],
+      },
+      {
+        name: 'ciProvider',
+        type: 'list',
+        message: 'Which CI provider do you want to use?',
+        // `newCommandOptions.ciProvider` has `github` as a default value,
+        // so we need to check the presence of the `--ci-provider` flag to know
+        // if the user provided a CI provider manually.
```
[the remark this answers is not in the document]
bertdeblock: I've removed the default value and explicitly assign `github` as the default after the interactive wizard, 
if no value for `ciProvider` exists.

lib/tasks/interactive-new.js
```diff
@@ -0,0 +1,160 @@
...
+    return langChoices;
+  }
+
+  getUserLocale() {
+    return osLocale();
+  }
+}
+
+function isCliOptionProvided(name) {
+  return process.argv.some((arg) => arg.includes(name));
```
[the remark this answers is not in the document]
bertdeblock: Should not be needed anymore with the change mentioned above.

bertdeblock:
Addressed the smaller issues, still need to come up with an acceptance test the completes the entire flow.

bertdeblock:
In case anyone is following along, I could use some help writing a full acceptance test where we simulate user input.

kategengler:
Discussed in the ember-cli meeting today, we've decided not to block on the interactive testing but hope to add that test coverage later.

PR #9848, opened 2022-04-09
Update `ember-cli-dependency-checker` to v3.3.1

### LINKED ISSUE #6277: `ember init` to skip dependency checking
I'm brainstorming the command `init` skipping the dependency checking.

The only time I'm using `init` is when updating addons (usually very out-of-date). So I clone the repo, cd into it, run `npm link ember-cli`, and then try to `ember init` with my up-to-date global version. Then I get the dependencies missing message. I then have to install all these old packages, which I'm about to update or remove anyways. Seems to me that dependency checking when doing `init` is not useful.

I don't think any missing packages other than ember-cli would cause `init` to fail, correct?

gnclmorais:
Fixes #6277. Follows the simple strategy I found on https://github.com/ember-cli/ember-cli/pull/8331.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -36,7 +36,7 @@
     "ember-cli": "~<%= emberCLIVersion %>",
     "ember-cli-app-version": "^5.0.0",
     "ember-cli-babel": "^7.26.11",
-    "ember-cli-dependency-checker": "^3.2.0",
+    "ember-cli-dependency-checker": "^3.3.1",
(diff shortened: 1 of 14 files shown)


rwjblue:
We need to remove the new peer dependency (introduced in 3.3.0, but a loooong time ago) before updating to it

rwjblue:
@gnclmorais let's go with 3.3.1 now that it is released

gnclmorais:
@rwjblue, updated to 3.3.1 👍

rwjblue (approved):

Issue #9793, opened 2022-02-20
Fix broccoli-test-helper buildOutput deprecation


locks:
Deprecation:

```
`buildOutput` is deprecated, use `createBuilder` or `fromBuilder` followed by `builder.build()` instead
```

PR #9857, opened 2022-04-20
Use `createBuilder` instead of deprecated `buildOutput` in test suite

geneukum:
Let's avoid using the deprecated broccoli-test-helper function
`buildOutput(outputNode)`, instead preferring to use
`createBuilder(outputNode)` followed by `await output.build()` as
suggested by the JSdocs.

This closes issue #9793.


bertdeblock (approved):

bertdeblock:
Thanks!

PR #9864, opened 2022-04-30
Addon blue prints should depend on on ember-source

optikalefx:
Because ember has started to remove some deprecations, if you generate and addon without the correct ember-source then your addon won't work.

This happened to me when I was using ember-cli v3.24 to generate an addon. @ember/component was not available because the addon was using the latest version of ember and not the ember-cli version that was used to generate it.

By having the addon generate with the same ember-source as the ember-cli that created it, this is avoided.

### DIFF
--- a/blueprints/addon/index.js
+++ b/blueprints/addon/index.js
@@ -49,6 +49,9 @@ module.exports = {
     contents.dependencies = contents.dependencies || {};
     contents.devDependencies = contents.devDependencies || {};
 
+		// addon should depend on the ember-source that matches ember-cli
+    contents.dependencies['ember-source'] = contents.devDependencies['ember-cli'];
+
     // npm doesn't like it when we have something in both deps and devDeps
     // and dummy app still uses it when in deps
     contents.dependencies['ember-cli-babel'] = contents.devDependencies['ember-cli-babel'];
--- a/tests/unit/blueprints/addon-test.js
+++ b/tests/unit/blueprints/addon-test.js
@@ -184,6 +184,19 @@ describe('blueprint - addon', function () {
         expect(json.dependencies).to.deep.equal({ a: '1', b: '1' });
         expect(json.devDependencies).to.deep.equal({ a: '1', b: '1' });
       });
+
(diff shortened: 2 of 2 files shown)


blueprints/addon/index.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
     contents.dependencies = contents.dependencies || {};
     contents.devDependencies = contents.devDependencies || {};
 
+		// addon should depend on the ember-source that matches ember-cli
+    contents.dependencies['ember-source'] = contents.devDependencies['ember-cli'];
```
NullVoxPopuli, line 53:
```suggestion
    contents.peerDependencies['ember-source'] = `>= ${contents.devDependencies['ember-source']}`;
```

I think this is what you want. This means that the consuming app chooses what ember-source, and the addon is expected to be compatible with at least this version

blueprints/addon/index.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
     contents.dependencies = contents.dependencies || {};
     contents.devDependencies = contents.devDependencies || {};
```
NullVoxPopuli, line 50:
```suggestion
    contents.devDependencies = contents.devDependencies || {};
    contents.peerDependencies = contents.peerDependencies || {};
```

NullVoxPopuli:
Thanks for getting this going! It's a huge help! <3

blueprints/addon/index.js
NullVoxPopuli, line 53:
```suggestion
    contents.peerDependencies['ember-source'] = '^3.24.0 || ^4.0.0';
```

Another option, maximum compat, matching ember try's config

ef4:
It would not be correct for an addon to have ember-source in `dependencies` because that would bring an extra copy of ember-source into applications, and there should really only be one copy.

The comments by @NullVoxPopuli showing `peerDependencies` would be a better solution. I think he's correct that the peer dependency range should match the same range of ember versions that appear in the default ember-try config.

bertdeblock:
@optikalefx Do you have time to address the comments?

PR #9666, opened 2021-10-14
Remove IE11 comments from `config/target.js` in app blueprint

bertdeblock:
Closes #9622.

PR #9885, opened 2022-05-02
[CLEANUP] Remove `ie 11` from default targets

### LINKED ISSUE #9884: Remove `ie 11` from default targets
https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/default-targets.js

bertdeblock:
IE 11 was removed from the app's `targets` file here https://github.com/ember-cli/ember-cli/pull/9666.
So I'm guessing the default targets used should reflect that as well?

Closes #9884.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -311,7 +311,7 @@ class Project {
     if (fs.existsSync(`${targetsPath}.js`)) {
       this._targets = this.require(targetsPath);
     } else {
-      this._targets = require('../utilities/default-targets');
+      this._targets = require('../../blueprints/app/files/config/targets');
     }
     return this._targets;
   }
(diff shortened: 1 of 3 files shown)


lib/models/project.js
```diff
@@ -311,7 +311,7 @@ class Project {
     if (fs.existsSync(`${targetsPath}.js`)) {
       this._targets = this.require(targetsPath);
     } else {
-      this._targets = require('../utilities/default-targets');
+      this._targets = require('../../blueprints/app/files/config/targets');
```
bertdeblock, line 314: Or is this trying to be too clever?
I know some blueprint files aren't valid JS because of the string interpolation, but here it works.

ef4 (approved):

PR #9890, opened 2022-05-05
Make sure addons are discovered only once

wagenet:


chancancode (approved):

bertdeblock:
👋 It seems [this test](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/project-test.js#L437-L446) is failing, because we already call `project.initializeAddons();` [here](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/project-test.js#L298) in the `beforeEach` hook.
Since this PR fixes the early return check, calling `project.initializeAddons();` [here](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/models/project-test.js#L443) has no effect anymore.
I think we can just remove the `project.initializeAddons();` call from the `beforeEach` hook and simply call it per test where needed.

bertdeblock (approved):

kellyselden (approved):

bertdeblock:
Thanks @wagenet!

PR #9904, opened 2022-05-11
Update Node compatibility warning

### LINKED ISSUE #9783: Tweak Node LTS warning when using recent active LTS
At the moment, when you are using the most-recent Node active LTS, but that LTS came out after the ember-cli you are using, you get this annoying error:

```
WARNING: Node v16.13.2 is not tested against Ember CLI on your platform. We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.
```

A suggestion given at the meeting was to instead link to the place that lists which versions of Node your ember-cli _was_ tested against.

bertdeblock:
New warning:

```
WARNING: Ember CLI v3.21.2 is not tested against Node v16.15.0. See "https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md" to find out which version of Node is best to use.
```

Closes #9783.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -151,20 +151,22 @@ class CLI {
 
       let platformCheckerToken = heimdall.start('platform-checker');
 
+      const emberCLIVersion = require('../../package.json').version;
       const PlatformChecker = require('../utilities/platform-checker');
       let platform = new PlatformChecker(process.version);
       let recommendation =
-        ' We recommend that you use the most-recent "Active LTS" version of Node.js.' +
-        ' See https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md for details.';
+        ' See "https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md" to find out which version of Node is best to use.';
 
       if (!this.testing) {
         if (platform.isDeprecated) {
-          this.ui.writeDeprecateLine(`Node ${process.version} is no longer supported by Ember CLI.${recommendation}`);
+          this.ui.writeDeprecateLine(
+            `Node ${process.version} is no longer supported by Ember CLI v${emberCLIVersion}.${recommendation}`
+          );
         }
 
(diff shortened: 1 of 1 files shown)


lib/cli/cli.js
```diff
@@ -151,20 +151,22 @@ class CLI {
 
       let platformCheckerToken = heimdall.start('platform-checker');
 
+      const emberCLIVersion = require('../../package.json').version;
       const PlatformChecker = require('../utilities/platform-checker');
       let platform = new PlatformChecker(process.version);
       let recommendation =
-        ' We recommend that you use the most-recent "Active LTS" version of Node.js.' +
-        ' See https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md for details.';
+        ' See "https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md" to find out which version of Node is best to use.';
```
locks, line 158: Why did you wrap the link in double quotes? Can you still cmd-click (or equivalent) it in terminals?

bertdeblock: > Why did you wrap the link in double quotes?

Looked more highlighted to me, but I can revert.

> Can you still cmd-click (or equivalent) it in terminals?

Yes.

I can link to the CI file, the only downside is that code changes might affect the link and that it needs to be updated again (if we use line numbers).

Another (but probably far-fetched) solution could be to define `[12.x, 14.x, 16.x]` in a separate file we can link to and access that file in CI using [fromJson](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson).

bertdeblock: We can also include the node versions directly in the message? We already load the CI file here:
https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/platform-checker.js#L7.

bertdeblock: @locks Any suggestions on how I should update this message?

ef4 (approved):

Issue #9905, opened 2022-05-11
Test against Node v18


bertdeblock:
Node v18 is a current stable release and will be the next LTS. According to our Node support document, we should already be testing against.

boris-petrov:
I personally can't run `ember-cli` on Node 18. Running my project leads to:

```
Build Error (PostcssCompiler)                                                                                                                                 
                                                                                                                                                              
You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
```

I believe that's because of a [bug](https://github.com/mozilla/source-map/issues/432) in `source-map` that has since been fixed but only released in a beta `0.8.0-beta.0`. I've got a lower version than that coming from many, many dependencies. One of them is `fast-sourcemap-concat`. Even if I pin `source-map` to version `0.8.0-beta.0`, I get an error from it:

```
Cannot find module 'source-map/lib/source-map/base64-vlq'
Require stack:
- /home/boris/project/node_modules/fast-sourcemap-concat/lib/vlq.js
- /home/boris/project/node_modules/fast-sourcemap-concat/lib/coder.js
- /home/boris/project/node_modules/fast-sourcemap-concat/lib/source-map.js
- /home/boris/project/node_modules/fast-sourcemap-concat/lib/index.js
- /home/boris/project/node_modules/broccoli-concat/index.js
- /home/boris/project/node_modules/ember-source/lib/concat-bundle.js
- /home/boris/project/node_modules/ember-source/lib/index.js
- /home/boris/project/node_modules/ember-cli/lib/models/package-info-cache/package-info.js
- /home/boris/project/node_modules/ember-cli/lib/models/package-info-cache/index.js
- /home/boris/project/node_modules/ember-cli/lib/models/project.js
- /home/boris/project/node_modules/ember-cli/lib/utilities/get-config.js
- /home/boris/project/node_modules/ember-cli/lib/utilities/instrumentation.js
- /home/boris/project/node_modules/ember-cli/lib/cli/index.js
- /usr/lib/node_modules/ember-cli/bin/ember
```

So I guess it's going to be pretty difficult to run `ember-cli` on Node 18...

jacobq:
[...]

PR #9906, opened 2022-05-11
Drop support for Node v12

### LINKED ISSUE #9893: Drop support for Node v12
Node v12 is EOL since 2022-04-30, which means we should be able to drop support according to [our Node support doc](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md).
Is this something we can do right away or do we need to do other things first before we can drop support here?

bertdeblock:
Node v12 is EOL since 2022-04-30, which means we can drop support following [our Node support doc](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md).

Closes #9893.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -64,7 +64,7 @@
     "webpack": "^5.72.1"
   },
   "engines": {
-    "node": "12.* || 14.* || >= 16"
+    "node": "14.* || >= 16"
   },
   "ember": {
     "edition": "octane"
--- a/docs/node-support.md
+++ b/docs/node-support.md
@@ -14,7 +14,7 @@
 | 9.x              | 2.16.2 - 3.2.x               |
 | 10.x             | 3.1.3 - 3.28.0               |
 | 11.x             | 3.9.0 - 3.13.0               |
-| 12.x             | 3.10.0 - Current             |
+| 12.x             | 3.10.0 - 4.6.0               |
 | 13.x             | 3.15.0 - 3.20.0              |
 | 14.x             | 3.19.0 - Current             |
 | 16.x             | 3.28.0 - Current             |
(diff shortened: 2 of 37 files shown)


elwayman02 (approved):

kellyselden (approved):

blueprints/app/files/package.json
```diff
@@ -64,7 +64,7 @@
     "webpack": "^5.72.0"
   },
   "engines": {
-    "node": "12.* || 14.* || >= 16"
+    "node": "14.* || >= 16"
```
SergeAstapov, line 67: @bertdeblock should this actually become `14.* || 16.* || >= 18`? Node.js 17 is EOL as well as it's non LTS

SergeAstapov: actually, worth having it in separate PR as we should add v18 elsewhere, like `docs/node-support.md`, `.github/workflows/ci.yml`, etc.

SergeAstapov (approved):

blueprints/app/files/package.json
bertdeblock: Hi @SergeAstapov, it was indeed my intention to add support for v18 in a separate PR.
There's also a separate ticket for it: https://github.com/ember-cli/ember-cli/issues/9905.

docs/node-support.md
```diff
@@ -14,7 +14,7 @@
 | 9.x              | 2.16.2 - 3.2.x               |
 | 10.x             | 3.1.3 - 3.28.0               |
 | 11.x             | 3.9.0 - 3.13.0               |
-| 12.x             | 3.10.0 - Current             |
+| 12.x             | 3.10.0 - 4.5.0               |
```
kellyselden, line 17: Is this still correct?

bertdeblock: Updated to `4.6.0`.

PR #9908, opened 2022-05-12
Remove support for finding an addon by its unscoped name

### LINKED ISSUE #9878: Remove support for finding an addon by its unscoped name
This happens here:

https://github.com/ember-cli/ember-cli/blob/master/lib/utilities/find-addon-by-name.js#L55-L64

Seems to have been deprecated a while ago.

bertdeblock:
Closes #9878.

### DIFF
--- a/lib/utilities/find-addon-by-name.js
+++ b/lib/utilities/find-addon-by-name.js
@@ -1,34 +1,14 @@
 'use strict';
 
-function unscope(name) {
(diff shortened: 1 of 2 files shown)


ef4 (approved):
I'm OK with this since it was marked deprecated before 4.0. I'd consider it a bug that we didn't start throwing immediately at 4.0, and this is the bug fix.

PR #9910, opened 2022-05-16
update stable deps before release

kellyselden:


bertdeblock:
@josephdsumner Nope, `ember-source` needs a patch release to fix the issue you are describing.

PR #9914, opened 2022-05-21
Temporarily skip failing ember new test for npm versions <= v6.0.0

bertdeblock:
This test has been failing consistently for Node versions that ship with npm v6 by default, being v12 and v14.
See https://github.com/npm/cli/issues/4896 for more info. The test is still run for Node v16.
Not sure if this is the best approach, but this should make CI green again. Feedback appreciated!

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const execa = require('execa');
+const semver = require('semver');
 const fs = require('fs-extra');
 const ember = require('../helpers/ember');
 const walkSync = require('walk-sync');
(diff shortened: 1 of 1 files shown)


tests/acceptance/new-test.js
```diff
@@ -276,20 +277,24 @@ describe('Acceptance: ember new', function () {
     expect(file('.named-ref')).to.exist;
   });
 
-  it('ember new with shorthand git blueprint and ref checks out the blueprint with the correct ref and uses it', async function () {
-    this.timeout(20000); // relies on GH network stuff
+  // Temporarily skipped for Node versions that ship with npm v6 by default.
+  // See https://github.com/npm/cli/issues/4896 for more info.
+  if (semver.gte(process.version, 'v16.0.0')) {
```
rwjblue, line 282: As far as I can tell from reading that issue, the issue is with `npm@6` and is not a problem with `npm >=7`, right?

If that is the case, can we check for the `npm` version instead of just assuming that we are using the bundled `npm`? There are a few different ways to get `npm` onto our CI machines, and additionally there are many different combos that folks might be using locally. Checking for the actual thing we care about seems much better here...

rwjblue (changes requested):
Overall I think it's totally fine to guard the test when we know we are working in a degraded environment, but I'd rather swap around the guard to be a bit more specific.

rwjblue (approved):

PR #9920, opened 2022-05-27
[BUGFIX] Make sure a blueprint’s options object and project instance are always available for all public hooks

### LINKED ISSUE #5481: this.project is undefined in Blueprint.supportsAddon()
The `install()` method of the `Blueprint` class is currently responsible for setting `this.project`. Since `supportsAddon()` is called before the `install()` method though `this.project` is not yet set which can cause issues in certain implementations of the `filesPath()` method (e.g. different paths depending on the installed test framework).

### LINKED ISSUE #7287: Blueprint hook files() does not receive options
https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L227-L249

The `files()` hook is documented as "Override the list of files provided by the blueprint. Useful if you want to exclude certain files conditionally." However, it does not receive the locals as a parameter so there is nothing to do a conditional check on. 

It is kind of possible to get the entity options from `this.options`, but the `files()` hook is called multiple times and the first time `this.options` is still undefined. The default implementation of the `files()` hook makes this error when calling `this.filesPath(this.options)`, but it doesn't fail because the default `filesPath` doesn't use the parameter. 

This also means that, when not overriding the `files()` hook, using the `options` parameter for overriding the  `filesPath` hook can fail.

### LINKED ISSUE #8160: Blueprint hooks invoked before options are set
I'm running into an issue when trying to define a new blueprint, where certain hooks are being called before the options (accessible as `this.options`) are provided to the blueprint.

`this.options` is set when calling the `install` hook here:

Call site:
https://github.com/ember-cli/ember-cli/blob/b52ce2e2614a08cf3c8690d668677900e523e383/lib/tasks/generate-from-blueprint.js#L70

Definition:
https://github.com/ember-cli/ember-cli/blob/b52ce2e2614a08cf3c8690d668677900e523e383/lib/models/blueprint.js#L440

However, when `supportsAddon` is called here:

https://github.com/ember-cli/ember-cli/blob/b52ce2e2614a08cf3c8690d668677900e523e383/lib/tasks/generate-from-blueprint.js#L31

The options haven't yet been set on the blueprint object. It, in turn, calls `files` which calls `filesPath`.

If your addon wants to depend on the user-provided options to alter the files generated (for instance, generating TypeScript or JavaScript files based on user's input), you run into a problem because the options are not yet available.

Because the default `files` implementation assumes that `this.options` is set (and calls `filesPath` with that value)

https://github.com/ember-cli/ember-cli/blob/b52ce2e2614a08cf3c8690d668677900e523e383/lib/models/blueprint.js#L241

you end up with the `options` sometimes being undefined when overriding `filesPath`

---

```
ember-cli: 3.5.0
http_parser: 2.8.0
node: 10.5.0
v8: 6.7.288.46-node.8
uv: 1.20.3
zlib: 1.2.11
ares: 1.14.0
modules: 64
nghttp2: 1.32.0
napi: 3
openssl: 1.1.0h
icu: 61.1
unicode: 10.0
cldr: 33.0
tz: 2018c
os: darwin x64
6.1.0
1.10.1
```

### LINKED ISSUE #8866: When using blueprint, filesPath hook not receiving options parameter
Ember CLI: 3.12.0

`filesPath()` is being called in `files()` that has [caching](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L235-L237), so it's only getting called once.

When is called for the [first time](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L716), [this.options](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L447) doesn't have a value, so the [value is never read by filesPath()](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L239)

There are several options for this:

1. this.options is being available in `_getFilesForInstall()`
2. remove cache for `files()`
3. add a flag to force recompute `files()`

bertdeblock:
Still need a way to properly test this. Suggestions are welcome!

Closes #5481.
Closes #7287.
Closes #8160.
Closes #8866.

### DIFF
--- a/lib/tasks/generate-from-blueprint.js
+++ b/lib/tasks/generate-from-blueprint.js
@@ -3,7 +3,6 @@
 const Blueprint = require('../models/blueprint');
 const Task = require('../models/task');
 const parseOptions = require('../utilities/parse-options');
-const { merge } = require('ember-cli-lodash-subset');
 const logger = require('heimdalljs-logger')('ember-cli:generate-from-blueprint');
 const lintFix = require('../utilities/lint-fix');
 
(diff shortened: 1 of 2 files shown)


lib/tasks/generate-from-blueprint.js
```diff
@@ -37,7 +56,7 @@ class GenerateTask extends Task {
 
     // otherwise, use default addon-import
     if (noAddonBlueprint.indexOf(name) < 0 && !addonBlueprint && options.args[1]) {
-      let mainBlueprintSupportsAddon = mainBlueprint && mainBlueprint.supportsAddon();
+      let mainBlueprintSupportsAddon = mainBlueprint && mainBlueprint.supportsAddon(blueprintOptions);
```
bertdeblock, line 59: This feels slightly weird. I was also thinking about passing the options to the blueprint upon instantiation, but that's a bit more involved. The options would need to be passed to the `Blueprint.lookup` method and so on.

bertdeblock: I just noticed that https://github.com/ember-cli/ember-cli/issues/5481 is a similar issue. I will update this PR to pass along the blueprint options and project upon creation of the blueprint instance.

bertdeblock: Done!

lib/tasks/generate-from-blueprint.js
```diff
@@ -29,18 +29,51 @@ class GenerateTask extends Task {
...
+      project: this.project,
+      settings: this.settings,
+      testing: this.testing,
+      taskOptions: options,
+      originBlueprintName: name,
+    };
+
+    let mainBlueprintOptions = merge({}, baseBlueprintOptions, options);
+    let testBlueprintOptions = merge({}, mainBlueprintOptions, { installingTest: true });
+    let addonBlueprintOptions = merge({}, mainBlueprintOptions, { installingAddon: true });
```
rwjblue, line 51: Let's use either `Object.assign` or `{ ... }` here, I don't think we need to use `merge` from lodash

rwjblue (commented):
Overall, LGTM, one minor nitpick though

lib/tasks/generate-from-blueprint.js
bertdeblock: Yup, updated!

I assume we can get rid of `merge` in other places as well?

rwjblue: Yep!

rwjblue (approved):

PR #9936, opened 2022-06-08
Allow app property to pass through addon proxy to addon instance

raycohen:
When the EmberApp constructor runs, the presence of nested addon dependencies can result in the `project.addons` list containing a mixture of addon instances and addon proxy objects. For the addon proxy objects, the EmberApp setting their `app` property would prevent the addon *instance* from having it set, which goes against the expectation described in the models/addon.js app property comment: 

https://github.com/ember-cli/ember-cli/blob/ee6d372df2e16384a92f06800329eeb23c83a7d1/lib/broccoli/ember-app.js#L132

https://github.com/ember-cli/ember-cli/blob/ee6d372df2e16384a92f06800329eeb23c83a7d1/lib/models/addon.js#L188-L193

Later on, when an addon attempts to read `this.app` it will be undefined if the addon was a Proxy by the time the EmberApp constructor method attempted to set it. This seems like a bug.

The comment explaining the proxy holding the app claims that ember-cli might set that app property, but I'm not sure what situation it is referring to. I didn't see any test coverage of this behavior. Simply removing this fixes the issue where addons in my project that expect to be able to read `this.app` in their index.js functions were erroring.

### DIFF
--- a/lib/models/per-bundle-addon-cache/addon-proxy.js
+++ b/lib/models/per-bundle-addon-cache/addon-proxy.js
@@ -41,10 +41,7 @@ function validateCacheKey(realAddonInstance, treeType, newCacheKey) {
 
 /**
  * Returns a proxy to a target with specific handling for the
- * `parent` property, as well has to handle the `app` property;
- * that is, the proxy should maintain correct local state in
- * closure scope for the `app` property if it happens to be set
- * by `ember-cli`. Other than `parent` & `app`, this function also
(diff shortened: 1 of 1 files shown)


rwjblue:
@brendenpalmer - Would you mind reviewing?

bertdeblock:
Anyone who can verify this change? Would be nice to get this in if this actually fixes an issue.

bertdeblock:

PR #9943, opened 2022-06-18
ENHANCEMENT: Add .DS_Store file in gitignore

msaaddev:
The `.gitignore` file inside the project that the CLI generates does not contain `.DS_Store` file. The file is automatically generated in macOS and pollutes the repo if it is accidentally committed.

So the PR is adding the `.DS_Store` file in the gitignore file created in the Ember project using the CLI.

### DIFF
--- a/blueprints/app/files/gitignore
+++ b/blueprints/app/files/gitignore
@@ -19,6 +19,7 @@
 /npm-debug.log*
 /testem.log
 /yarn-error.log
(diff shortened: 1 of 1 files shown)


kellyselden:
Thanks for the PR. We try to avoid editor and OS-specific changes. This kind of change would be better suited for a global .gitconfig.

PR #9944, opened 2022-06-21
Run tests with Node 18 in CI

ddzz:
This PR adds the latest version of Node to the matrix configuration in the `tests` job in CI.

PR #9945, opened 2022-06-21
[ENHANCEMENT/BREAKING] Add Node v18 to `engines` in `app` and `addon` blueprint (removes support for Node v17)

bertdeblock:
Probably best to merge after #9944.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -64,7 +64,7 @@
     "webpack": "^5.74.0"
   },
   "engines": {
-    "node": "14.* || >= 16"
+    "node": "14.* || 16.* || >= 18"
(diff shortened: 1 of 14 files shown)


blueprints/app/files/package.json
```diff
@@ -64,7 +64,7 @@
     "webpack": "^5.74.0"
   },
   "engines": {
-    "node": "14.* || >= 16"
+    "node": "14.* || 16.* || >= 18"
```
rwjblue, line 67: Note: this explicitly drops support for Node 17; we should make sure to call this out as a breaking change in the release / changelog / etc.

rwjblue (approved):

PR #9972, opened 2022-07-22
[ENHANCEMENT] Add support for `--typescript` flag to `app` and `addon` blueprints

simonihmig:
This adds the `--typescript` flag to the `ember new` and `ember addon` commands, as per the [CLI integration section of RFC800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan#cli-integration).

While the blueprint changes itself were not very difficult, what did cost *a lot* of time was debugging and identifying problems with regard to Ember CLI itself and the test coverage. 

What is special here, and what *uncovered* (not *caused*) the mentioned problems, is that the blueprint delegates much of the work to `ember-cli-typescript` and its own default blueprint, when the flag is present. This is per se not bad, and it is what the RFC actually says. But by calling `addAddonToProject()`, it forcefully installs all the dependencies, even in tests, regardless of the `--skip-npm` flag. This is inevitable I think, as otherwise `ember-cli-typescript`'s blueprint couldn't run in tests. But it makes the test naturally *quite* slow, and it uncovered some unexpected issues unrelated to the actual changes, which I will walk you through in the following...

1. ~It seems the `project` is shared between tests (see below). When `initializeAddons()` is called, it is supposed to also call `discoverAddons()` [here](https://github.com/ember-cli/ember-cli/pull/9972/files#diff-b6deaa28f3fa9f3dbd9f2016784825ced589e12bb5f77b6d5e6ff6028560e309R503). But it happens that `_didDiscoverAddons` is true here, so `discoverAddons()` return early. The thing is here, that usually in blueprint tests you only deal with the internal addons of ember-cli. But with the new typescript tests, and external addons are *actually* installed from npm, we now deal with different sets of `this.addons` depending on the test. In the typescript test, `this.addons` still would only contain the internal addons (due to the early return), which when trying to run `ember-cli-typescript`'s default blueprint makes this fail, saying that `ember-cli-typescript` is an unknown addon (as `this.addons` hasn't been updated). So I *think* the fix I added [here](https://github.com/ember-cli/ember-cli/pull/9972/files#diff-b6deaa28f3fa9f3dbd9f2016784825ced589e12bb5f77b6d5e6ff6028560e309R499) is legit!?~
Own fix replaced by #10014 

2. When running the typescript blueprint tests (which again install real external dependencies) with npm, an (unrelated) peer dependency issue arises, which is described here: https://github.com/emberjs/ember-test-helpers/issues/1236. This makes the test fail, but the issue is not related to the test itself, it's real. Therefor for now, the added tests only work with yarn, which does not share this behavior around peer deps.

3. ~When running the blueprint acceptance tests, they call `await ember(['new', 'foo', '--skip-npm', '--skip-bower', '--skip-git']);` or similar. This eventually calls into [`Project#nullProject()`](https://github.com/ember-cli/ember-cli/blob/ca24ccde85b5024d1561624074ae2def84de404b/lib/models/project.js#L178-L200), which return a *singleton* project instance. But this means this instance is shared (leaked!) between tests. This seems to not have caused issues before, as the instance was not fundamentally different for different tests. But it is now! If you look back at item 1., you see that for the typescript tests real external addons are added to `this.addons`, which is not the case for all the other tests. But now that this state is leaking between tests, the following happens:
Any tests would pass when running in isolation. But any of the tests that *follows* one of the new typescript tests would fail when calling `await ember(...)`, with `Cannot find module './commands'` thrown when `@ember/optional-features` tries to (lazily!) require it [here](https://github.com/emberjs/ember-optional-features/blob/master/index.js#L12). See [this CI run](https://github.com/ember-cli/ember-cli/runs/7476653237?check_suite_focus=true#step:5:531). This happens because in the (leaked) project instance `this.addons` still contains all the external addons *in memory* (including `@ember/optional-features`, which just *happens* to cause the error as it's the first one), while the actual modules on disk in the tmp folder have been deleted by the (previous) test's `afterEach`. So that `require` naturally fails. But the root issue here is that the project's state is leaking here between tests I think.
I was able to overcome this with a quick'n'dirty hack in the last [commit](https://github.com/ember-cli/ember-cli/pull/9972/commits/e8ea613729658570c2c90884265b4463a0e3af49), which basically removes the singleton behavior. In the latest test run you can see that all of the blueprint tests are passing now! But...~

Fixed, see https://github.com/ember-cli/ember-cli/pull/9972#issuecomment-1219535120.  

4. ~CI is now failing with a `JavaScript heap out of memory` error. Probably this is due to this hack, which might be causing some memory leaks? However I didn't want to spend even more time on investigating this, before the real path to overcoming 3. is not determined.~

~So 3. is really the point where I would like some assistance from some CLI experts here on how to proceed with that, as I don't fully understand what the deal is with this "null project", and what implications it would have to not treat it as a singleton to prevent it from leaking between tests, or if some other approach is better suited. I'll keep this in a draft state for now, but please review nevertheless!~


tests/fixtures/app/typescript/.eslintrc.js
```diff
@@ -0,0 +1,67 @@
...
+    'eslint:recommended',
+    'plugin:ember/recommended',
+    'plugin:prettier/recommended',
+  ],
+  env: {
+    browser: true,
+  },
+  rules: {},
+  overrides: [
+    // ts files
```
simonihmig, line 24: Putting the TS `extends` into an override for *only* .ts files is not something the RFC explicitly mandated, but it is what I have been doing everywhere. Otherwise, when putting this into the global `extends`, you would get false positives from TS rules when ESLint runs on plain .js files.

If you suggest a different solution, please tell me!

lib/models/project.js
```diff
@@ -311,7 +311,7 @@ class Project {
     if (fs.existsSync(`${targetsPath}.js`)) {
       this._targets = this.require(targetsPath);
     } else {
-      this._targets = require('../../blueprints/app/files/config/targets');
+      this._targets = { browsers: ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'] };
```
simonihmig, line 314: When requiring `../../blueprints/app/files/config/targets`, node would for some reason also try to read from `../../blueprints/app/files/package.json`, which is a *template* file, not a real one. 

Previously this caused no issues, as the only parts with "template tags" (or whatever you call these `<% %>`) was *within* `.name` [here](https://github.com/ember-cli/ember-cli/pull/9972/files#diff-e4c0828d65b51abb30d80befe6fe7870dc36a73dad0083a9dcbb09d48f58e1cfL2). But now we have more conditionals in place, which make this package.json template *invalid* JSON (in the raw format), so node would fail parsing it.

As `targets.js` is a JS file, I also couldn't just simply read it without making node *interpret* it (like with something like `fs.readJSONSync()`), so the only fix I could come up with is to unfortunately just duplicate value here. There is a test covering the output of this, so hopefully the chance of this and the actual targets file diverging is very low! 🤔

lib/models/project.js
```diff
@@ -176,9 +176,9 @@ class Project {
   }
 
   static nullProject(ui, cli) {
-    if (NULL_PROJECT) {
-      return NULL_PROJECT;
-    }
+    // if (NULL_PROJECT) {
+    //   return NULL_PROJECT;
+    // }
```
simonihmig, line 181: This shouldn't stay like this, see the PR descriptions's item 3.!

blueprints/app/files/package.json
```diff
@@ -17,7 +17,8 @@
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
-    "lint:js:fix": "eslint . --fix",
+    "lint:js:fix": "eslint . --fix",<% if (typescript) { %>
+    "lint:ts": "tsc --noEmit",<% } %>
```
simonihmig, line 21: The RFC also doesn't say anything about this. But I *believe* something like this should be added!?

chriskrycho: Something longer it, yes, but it should absolutely not be named as a lint. I know a bunch of addons do treat it this way, and some people choose to think of it this way, but from a pedagogical perspective it is not how we (specifically the Typed Ember team) want to encourage people to think of TypeScript. Type errors are a different class of errors than lint errors. They aren’t opinions or best practices, however string: they are *correctness issues*, and so deserve a different treatment.

I would suggest something like `type-check` or, if we wanted to generalize, introduce a `check:` namespace which could include `check:types` and `check:lint:*`.

simonihmig: I see your point, and you are certainly not wrong here. 

However the solution here seems like the most pragmatic, as putting it under `lint:*` makes the main `lint` script also run the TS check, alongside ESLint and template lint (by means of the `npm-run-all` setup). Putting it under a different name(space) would require users to account for it in their CI setup, and us to update the CI-parts of the blueprints. 

Or we define `lint` as something like `"lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\" && yarn type-check"`, but this doesn't make things much better with regard to your argument of not confusing type checking with linting.

blueprints/app/files/app/router.__ext__
```diff
@@ -6,4 +6,4 @@ export default class Router extends EmberRouter {
   rootURL = config.rootURL;
 }
 
-Router.map(function () {});
+Router.map(function () {});<% if (typescript) { %> // eslint-disable-line @typescript-eslint/no-empty-function<% } %>
```
dfreeman, line 9: I'm surprised `no-empty-function` is part of the `@typescript-eslint` recommended ruleset but not the regular `eslint:recommended` one. (But that's neither here nor there 😄)

Would changing this to something like the snippet below appease the rule without us needing to explicitly disable it?

```ts
Router.map(function () {
  // Add route declarations here
});
```

blueprints/app/files/package.json
dfreeman: I have super mixed feelings on this one. In a vacuum designing this from scratch, I like the idea of a `check:` namespace, but I suspect that would be a complicated change to make today. I think there would be a fair amount of friction in migrating, both in terms of existing users' muscle memory and custom scripts, and how `lint:fix` is enshrined in `ember-cli` as the "run this after generating new files to clean them up" script.

I agree that Including `type-check` (or `typecheck`, depending on what color we paint that shed) in the `lint` script isn't ideal, but it may be an ok pragmatic compromise.

blueprints/app/files/app/router.__ext__
simonihmig: > Would changing this to something like the snippet below appease the rule without us needing to explicitly disable it?

Yes, it would!

I just wanted to try out what happens with the comment when you add a route by `ember g route`, only to find out that it actually does not work at all with a `router.ts` file! Is this known?

Not sure if https://github.com/ember-cli/ember-router-generator would actually support parsing TS, but the error is thrown even before that lib comes into play, in the ember-source blueprint [here](https://github.com/emberjs/ember.js/blob/master/blueprints/route/index.js#L165-L166), as it only checks for a `router.js`.

dfreeman: Yep, that's [a known issue](https://github.com/typed-ember/ember-cli-typescript/issues/165) (Which we should definitely fix now that TS support is official! Though obviously not a hard blocker for this PR in particular)

blueprints/app/files/package.json
ef4: > type errors are a different class of errors than lint errors. They aren’t opinions or best practices, however strong: they are correctness issues, and so deserve a different treatment.

I don't think this is a helpful framing because lots of lint rules are *also* correctness issues, up to and including "instant runtime failure".

ef4: I think we should zoom in on the actual pedagogical goal here, which is probably something like "helping people understand that some checks are pretty important for correctness while others are just opinions so they don't turn off the important ones".

But trying to draw a hard line between tsc and any other static checks that are happening is not an effective solution to that problem. There are eslint and template-lint rules that are *also* about correctness just as much as typescript is, so if we're trying to teach people what's actually important, the boundary between important and opinion should include some subset of the eslint and template-lint rules too.

I also think you're never going to get everyone to agree that tsc is not a linter, so unless you want to keep having that argument forever you can just accept that it sure looks very linter-esque, especially when it's running `--noEmit` and literally has the exact same developer interactions that eslint does ("During CI it does static analysis to tell me about problems and in my editor it makes red squiggles.").

Finally, I would argue that typescript-the-project thinks about itself as a linter. That's why `noEmitOnError` defaults to false. It's **normal and expected** during active development **to run and debug code that still has type errors**, and that is central to the design of typescript.

simonihmig: > it does static analysis to tell me about problems and in my editor it makes red squiggles.

Yes, that pretty much sums up how I personally understand the term linting. For me it was always about *how* it does its job (static analysis, i.e. basically "looking" at code instead of running it, which is what tests are about).

Similarly, the [definition of linting on Wikipedia](https://en.wikipedia.org/wiki/Lint_(software)) emphasizes static analysis, but does not really limit the scope of what kind of code problems it cares about: (see "programming errors" which I understand as *correctness* type of problems)

> Lint, or a linter, is a [static code analysis](https://en.wikipedia.org/wiki/Static_program_analysis) tool used to flag programming errors, [bugs](https://en.wikipedia.org/wiki/Software_bug), stylistic errors and suspicious constructs.

simonihmig:
For the record: it seems @ef4 has fixed the main `nullProject` problem here, so thanks a bunch for that! :tada:

For my own record: I recently used this branch to create a real new project, and realized that we need to also...

- [ ] convert [tests/helpers/index.js ](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/tests/helpers/index.js) to TS (you'll get a type error, as this is imported from TS).

ef4:
Yes, the NullProject question is resolved. The summary is that NullProject is supposed to be for the case where ember-cli is running outside of any project. That is the situation when you run `ember new`. But then to install addons into the newly generated project, we need to have a real Project and not NullProject.

blueprints/app/files/tests/helpers/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  setupTest as upstreamSetupTest,
+} from 'ember-qunit';
+
+// This file exists to provide wrappers around ember-qunit's / ember-mocha's
+// test setup functions. This way, you can easily extend the setup that is
+// needed per test type.
+
+function setupApplicationTest(
+  hooks: Parameters<typeof upstreamSetupApplicationTest>[0],
+  options?: Parameters<typeof upstreamSetupApplicationTest>[1]
```
simonihmig, line 13: TS team: please have a look here!

I could have typed the first arg as `hooks: NestedHooks`, as `@types/ember-qunit` declares this type globally. But not the `SetupOptions` interface, which is not even exported. So the only way I found to type this was like this. And for consistency I chose to do this for both args...

simonihmig:
I think this PR is good now, with only the somewhat philosophical question remaining of how to name and integrate the type checking script.

So please give this a/another review, Ember-CLI team & TS-Team (@dfreeman @chriskrycho @jamescdavis)!

ef4:
(Please don't take my discussion above over the definition of "lint" as a blocker, I've said my piece but I can live with whatever will get this moving.)

chriskrycho:
Will review Monday or Tuesday. And for what it's worth the Typed Ember team generally shared the "not going to block on this" POC about naming. It's probably unsurprising that we *do* disagree about whether to think about type checking and linting as equivalent—but we would all rather ship this with a name we don't love than not ship it while arguing over that name!

blueprints/app/files/tests/helpers/index.ts
dfreeman: This could also be `(...[hooks, options]: Parameters<typeof upstreamSetupApplicationTest>)`, not that that's _much_ better. I think we should probably just export the `SetupOptions` type from `@types/ember-qunit`.

chriskrycho: I am doing that now!

chriskrycho: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62073 – I exported it as `@internal`, since it is expressly *not* part of the public API, but *is* important for collaborators here. I also opened https://github.com/emberjs/ember-qunit/issues/957 to track converting `ember-qunit` to TS!

blueprints/app/files/package.json
chriskrycho: I take the point that ESLint and ember-template-lint do both catch actual *correctness* issues; I was wrong when I distinguished along that axis. That being said, I still think there is a meaningful difference here—types, lints, formal methods, preconditions-and-postconditions, and more all exist on the spectrum of “static code analysis tools for proving correctness.” That doesn’t mean they’re the same things! However, I’m *much* more interested in seeing this move forward than in getting lost in the weeds about that detail. I am fine with going with `lint:types` for now, and we can think about whether we want to move the ecosystem to the idea of `check:*` or such at some later point.

I'm suggesting `lint:types` rather than `lint:ts` because I think `lint:ts` might imply "do ESLint or similar on TS files".

```suggestion
    "lint:types": "tsc --noEmit",<% } %>
```

chriskrycho (changes requested):
All right, one suggested tweak here, but looks good otherwise!

lib/models/project.js
```diff
@@ -496,6 +496,7 @@ class Project {
       return;
     }
     this._addonsInitialized = true;
+    this._didDiscoverAddons = false;
```
bertdeblock, line 499: I assume this was added to fix the `ember install` error?
I made https://github.com/ember-cli/ember-cli/pull/10014 as well in an attempt to resolve that.
I'm not entirely sure which fix is the right one, but it is probably better to have a separate PR and changelog entry for this.

simonihmig: It addresses the issue I described in 1. in the PR description. Let me check if your fix does also fix the issue for me, then I'll revert this and wait for your change to get merged first...

simonihmig: Yeah, just reran the tests with your change, and they were passing. So I'll just discard my change and rebase when your PR has been merged!

bertdeblock: Good to know, thanks for confirming!

simonihmig:
Status: Will wait for #10014 to get merged and rebase, after that this PR should hopefully be good then!

tests/acceptance/new-test.js
```diff
@@ -646,6 +650,65 @@ describe('Acceptance: ember new', function () {
...
+      ['.ember-cli'].forEach((filePath) => {
+        checkFile(filePath, path.join(__dirname, '../fixtures', fixturePath, filePath));
+      });
+      checkFileWithEmberCLIVersionReplacement(fixturePath, 'tests/dummy/config/ember-cli-update.json');
+      checkEslintConfig(fixturePath);
+
+      // smoke test for the existence of essential TypeScript features... (see app test)
+      let pkgJson = fs.readJsonSync('package.json');
+      expect(pkgJson.scripts['lint:ts']).to.equal('tsc --noEmit');
+      expect(pkgJson.devDependencies['ember-cli-typescript']).to.exist;
```
bertdeblock, line 706:
```suggestion
      expect(pkgJson.dependencies['ember-cli-typescript']).to.exist;
```

`ember-cli-typescript` needs to end up in `dependencies` for addons:

```
WARNING: `ember-cli-typescript` should be included in your `dependencies`, not `devDependencies`
```

bertdeblock (commented):
Noticed a few things:

1. When generating an app or addon, I'm seeing the following warning logged about 40 times:
```
You passed the '--typescript' flag but there is no TypeScript blueprint available. A JavaScript blueprint will be generated instead.
```

2. Running `ember s` in a new app or addon fails with the following error:
```
Error: Typechecking failed
```

3. Do we want/need to specify/control the version of `ember-cli-typescript`?

blueprints/app/files/package.json
simonihmig: Updated!

tests/acceptance/new-test.js
simonihmig: Great catch! Fixed.

blueprints/app/files/tests/helpers/index.ts
simonihmig: Using the new export now!

blueprints/app/files/app/router.__ext__
simonihmig: Now using the suggested comment in the (otherwise still empty) function body!

simonihmig:
> When generating an app or addon, I'm seeing the following warning logged about 40 times:
You passed the '--typescript' flag but there is no TypeScript blueprint available. A JavaScript blueprint will be generated instead.

The logic around this warning message would emit the warning whenever the `--typescript` flag is explicitly passed, and the blueprint files contain *any* non-ts files, like hbs. So this seems to be a general bug, that probably way only uncovered now with the blueprint shipping so many other non-ts files. Fixed that in https://github.com/ember-cli/ember-cli/pull/9972/commits/538acd0c677a6d71feb72cf2e2266ec6a8640653. 

> Running ember s in a new app or addon fails with the following error:
Error: Typechecking failed

I realized that some types packages like `@types/ember-qunit` were not installed by e-c-ts' own blueprint, as the fix by @ef4 would leave the stale version of `project` in `blueprintOptions`, so the e-c-ts blueprint would not see that `ember-qunit` is installed [here](https://github.com/typed-ember/ember-cli-typescript/blob/master/ts/blueprints/ember-cli-typescript/index.js#L188-L191) and therefore *not* install its types. Fixed in https://github.com/ember-cli/ember-cli/pull/9972/commits/005db393a333eab78c3da606e12ef5f44ca323b4. 

In my tests, a generated TS app lints, tests and runs fine now!

> Do we want/need to specify/control the version of ember-cli-typescript?

Idk, why would we not want to have the latest version?

simonihmig:
Hey folks (@bertdeblock @chriskrycho @dfreeman @ef4), I think I addressed all feedback here! (thank you!)
CI seems happy, hope you are too! :wink:

bertdeblock:
Hi @simonihmig, RE:

> Do we want/need to specify/control the version of ember-cli-typescript?

I didn't have a specific reason in mind, it was just a general wonderment, since the versions of all the other app and addon dependencies are defined in the `package.json` file. I guess my thinking was, if in the future, you generate or `ember-cli-update` to an older version of `ember-cli`, you get the `ember-cli-typescript` version that was defined and compatible at that time. It could be that I'm mistaken and/or overthinking this though.

simonihmig:
I see your point. I am still a bit undecided though. It requires then continuously updating the version specifiers, at least for major versions. The thing is that all the other versions are easily discoverable by looking into the blueprint's `package.json`. But this one will live somewhere unexpectedly hidden in `index.js`...

chriskrycho:
Might be a good use case for `optionalPeerDependencies`. That way it appears in `package.json` and can specify the supported ranges at any given time.

bertdeblock:
I _think_ we can add `ember-cli-typescript` to `dependencies`/`devDependencies` in the `package.json` file, so it gets picked up by the [update-blueprint-dependencies](https://github.com/ember-cli/ember-cli/blob/master/dev/update-blueprint-dependencies.js) script. And instead of _installing_ `ember-cli-typescript` post install, we would only need to run the `ember-cli-typescript` default blueprint post install.

bertdeblock (approved):

bertdeblock:
I'm unsure if my "concern" is a valid one and/or if it should be blocking progress here.
As far as I can tell, this seems good to go!

simonihmig:
Seems I need to fix some conflicts here. Will do that once I have general approval from someone from the TS team at least!

chriskrycho:
I’ll make sure once of us looks this coming week. Thanks for keeping it moving forward!

dfreeman (approved):
High level this makes sense to me and seems like a great step forward!

The `ember-cli-typescript` versioning question is an interesting one, but at least in my opinion doesn't need to be a blocker here. As we consider what a world without `ember-cli-typescript` may look like (look for an RFC from @jamescdavis in the near-ish future), that question may answer itself :)

simonihmig:
I rebased this, so I think this is good to go now!

bertdeblock:
Thanks @simonihmig!

PR #9975, opened 2022-07-28
Fix `prefer-const` lint violations in `app` and `addon` blueprints

bmish:
Many users use the [prefer-const](https://eslint.org/docs/latest/rules/prefer-const) ESLint lint rule. We can make a minor/trivial/harmless tweak to the blueprint so that there won't be any violations from this lint rule. There's no reason anyone should be reassigning the value of `ENV` or `app` anyway.


bertdeblock (approved):

bertdeblock:
I think this is okay, thanks @bmish!

PR #9985, opened 2022-08-03
[WIP] Default workspace settings for Visual Studio Code

kaermorchen:
This PR adds a new argument `--code-editor=vscode` to `init` and `new` commands. It adds the default workspace settings for Visual Studio Code.


bertdeblock:
I've put this in draft to reflect the WIP status.

bertdeblock:
I'm going to close this one for now if that's okay. I think it would be better to submit the RFC first.
I think it would be good to get some community feedback on the files added here, the contents of these files and if we want to support this in the first place. If we offer support for VS Code, users might expect support for other editors as well.

Issue #9404, opened 2020-12-03
On Node 15, errors during build are not recoverable


kategengler:
On Node 15, something that causes a build error (such as a syntax error in hbs or js) causes `ember s` to exit 1 when a request is made.

For example, you type part of a line of code, your editor saves, the build error displays in your log, livereload is triggered and a request is made, then the server exits 1. 

<details>
<summary>Output</summary>

```bash
file changed components/selected-item.hbs.hbs
Build Error (broccoli-persistent-filter:Babel > [Babel: my-app]) in my-app/components/my-component.js

/Users/katie/dev/my-app/my-app/components/my-component.js: Parse error on line 79:
...           @extra={{(hash driveTime=@dri
-----------------------^
Expecting 'ID', 'STRING', 'NUMBER', 'BOOLEAN', 'UNDEFINED', 'NULL', 'DATA', got 'OPEN_SEXPR'


Stack Trace and Error Report: /var/folders/wq/_yzzcnwn3wlfm5x_6gg5d8vm0000gn/T/error.dump.64eb8cc3228b3c62aa5d1e0a7a89a0e7.log
node:internal/process/promises:218
          triggerUncaughtException(err, true /* fromPromise */);
          ^

BuildError: /Users/katie/dev/my-app/my-app/components/my-component.js: Parse error on line 79:
...           @extra={{(hash driveTime=@dri
-----------------------^
Expecting 'ID', 'STRING', 'NUMBER', 'BOOLEAN', 'UNDEFINED', 'NULL', 'DATA', got 'OPEN_SEXPR'
    at Parser.parseError (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:13397:15)
    at Parser.parse (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:13472:20)
    at parseWithoutProcessing (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:14657:26)
    at parse (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:14662:15)
    at preprocess (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:8885:35)
    at precompile (/Users/katie/dev/my-app/node_modules/ember-source/dist/ember-template-compiler.js:6103:38)
[...]

PR #9987, opened 2022-08-03
Handle rebuild failures without exiting

### LINKED ISSUE #9584: v3.15 w/ Node 14+ crashes consistently when postBuild hook throws
Discovered during some investigative work by @dfreeman over here https://github.com/typed-ember/ember-cli-typescript/issues/1439#issuecomment-878244006

tldr: 
 - postBuild returns a rejected promise, ember-cli crashes
 - rejected promises need to be caught from postBuild

bendemboski:
When a rebuild fails, mark it as a builder failure so the watcher doesn't re-throw it, causing it to be an unhandled error that results in the process exiting.

Fixes #9584 #9404

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -221,6 +221,10 @@ class Builder extends CoreObject {
     } catch (error) {
       await this.processAddonBuildSteps('buildError', error);
 
+      // Mark this as a builder error so the watcher knows it has been handled
+      // and won't re-throw it
+      error.isBuilderError = true;
+
       throw error;
     } finally {
       clearInterval(uiProgressIntervalID);
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -49,7 +49,16 @@ class HistorySupportAddon {
 
     app.use(async (req, _, next) => {
       try {
-        const results = await watcher;
+        let results;
+        try {
+          results = await watcher;
+        } catch (e) {
+          // This means there was a build error, so we won't actually be serving
+          // index.html, and we have nothing to do. We have to catch it here,
+          // though, or it will go uncaught and cause the process to exit.
+          return;
+        }
+
         if (this.shouldHandleRequest(req, options)) {
           let assetPath = req.path.slice(baseURL.length);
           let isFile = false;
--- a/tests/unit/models/builder-test.js
+++ b/tests/unit/models/builder-test.js
@@ -478,6 +478,20 @@ describe('models/builder.js', function () {
       await expect(builder.build()).to.be.rejected;
       expect(hooksCalled).to.deep.equal(['preBuild', 'build', 'postBuild', 'outputReady', 'buildError']);
     });
+
+    it('sets `isBuilderError` on handled addon errors', async function () {
+      addon.postBuild = function () {
+        return Promise.reject(new Error('preBuild Error'));
+      };
+
+      let error;
+      try {
+        await builder.build();
+      } catch (e) {
+        error = e;
+      }
+      expect(error).to.haveOwnProperty('isBuilderError', true);
+    });
   });
 
   describe('fallback from broccoli 2 to broccoli-builder', function () {


ef4:
I confirmed that this does not fix either kind of crash.

(Meaning, an exception thrown out of template compilation or a rejected promise out of a postBuild hook.)

Remember to check that you can do a successful HTTP request after the build error. That is where the crash actually happens.

bendemboski:
@ef4 I guess I haven't tested the template compilation scenario, but I am seeing it working for addons rejecting in the post-build hook. This doesn't address the case where that happens during the initial build, only during rebuilds -- I'm pretty sure initial build failures have "always" (i.e. as long as I can remember) caused the process to exit, so I wasn't aiming to address that. The main scenario I was targeting is that I introduce a typescript error while editing code and that causes `ember-cli` to exit, so even after I've fixed the typescript error I have to restart the whole process and new a full build.

[here](https://github.com/bendemboski/ember-cli-pr-9987-repro) is an Ember app I threw together with instructions in the README to I believe demonstrates that this PR fixes the issue I'm describing.

ef4:
Thanks for sharing the repro app. For anybody else following along: make sure you run the repro on Node 15 or newer to see the crash.

Following the README, I can confirm that I observe the crash that you documented in the unmodified app. But then I followed the instructions to switch to this PR, and I can still make it crash. Here's the diff compared to the steps you wrote in the README:

```diff
To see that it fixes it:

1. In `package.json`, change the `ember-cli` version to git+ssh://git@github.com/bendemboski/ember-cli.git#handle-build-failures
2. `yarn`
3. `ember serve` (build succeeds)
2. Open http://localhost:4200 in a browser (UI loads)
3. Make a change to `application.hbs` (build fails, process does _not_ exit, browser does not reload)
-4. Make another change to `application.hbs` (build succeeds, browser reloads)
-5. Repeat steps 3-4 as desired
+4. Manually refresh localhost:4200 in the browser. (ember-cli crashes)
```

ef4:
Just to check that I'm not doing something dumb, here's the diff I get in yarn.lock when I switch to the PR:

```diff
diff --git a/yarn.lock b/yarn.lock
index 2c25036..8817fea 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -4658,10 +4658,9 @@ ember-cli-version-checker@^5.1.1, ember-cli-version-checker@^5.1.2:
     semver "^7.3.4"
     silent-error "^1.1.1"
 
-ember-cli@~4.6.0:
-  version "4.6.0"
-  resolved "https://registry.yarnpkg.com/ember-cli/-/ember-cli-4.6.0.tgz#ff409b99d074414a8944bdadf6db820eaf2dfb7b"
-  integrity sha512-4OAPHJJt4wARrTzfpfEM49KA1JJjMZ/8mhUncdqJFJG19/mS6yJiAUemnsLhZLnbjicB/gYSQ3rrUv/vxeX3cA==
+"ember-cli@git+ssh://git@github.com/bendemboski/ember-cli.git#handle-build-failures":
+  version "4.7.0-beta.0"
+  resolved "git+ssh://git@github.com/bendemboski/ember-cli.git#a8f619d108fd5ff2980f23a2664863f524b2844c"
   dependencies:
     "@babel/core" "^7.18.6"
     "@babel/plugin-transform-modules-amd" "^7.18.6"
```

bendemboski:
Okay, @ef4 I see what you're seeing now. There are two distinct issues, one where the process exits immediately upon an addon throwing from `postBuild()` (and likely other hooks), and another where the process exits when a request to serve the app is received when the most recent build failed.

My first commit fixed the first issue, and I just pushed a second commit that fixes the second issue. So now using my repro app,

1. In `package.json`, change the `ember-cli` version to git+ssh://git@github.com/bendemboski/ember-cli.git#handle-build-failures
2. `yarn` (might have to `yarn --force` to get the latest commit from my branch if the previous one is cached?)
3. `ember serve` (build succeeds)
2. Open http://localhost:4200 in a browser (UI loads)
3. Make a change to `application.hbs` (build fails, process does _not_ exit, browser does not reload)
4. Manually refresh localhost:4200 in the browser. (process does _not_ exit, browser loads build error page)
5. Make another change to `application.hbs` (build succeeds, browser reloads)

Note that in step (4) the error page looks odd:

localhost_4200

I'm pretty sure this is a pre-existing issue where we aren't wrapping random errors thrown from addon hooks properly in the broccoli wrapper magic to make the error page look right -- it looks exactly as expected when the error is a template compilation error rather than my contrived addon hook error. I poked around a bit and couldn't easily figure out how best to address this, but IMO it should not hold up getting the core issues here fixed.

ef4 (approved):

bertdeblock:
Thanks @bendemboski and @ef4!

jdaviderb:
When will this feature be released?

MichalBryxi:
@jdaviderb from what I can see [in the CHANGELOG](https://github.com/ember-cli/ember-cli/blob/master/CHANGELOG.md#v480) the [related issue](https://github.com/ember-cli/ember-cli/issues/9404#issuecomment-792335093) was released in `v4.8.0`. But maybe I'm reading the history here wrong?

Issue #9966, opened 2022-07-04
Replace `npm-run-all` with `concurrently` in blueprints?


bertdeblock:
`npm-run-all` seems unmaintained and `concurrently` seems like a good / modern alternative.

PR #9500, opened 2021-03-29
Implemented one-line fix for yarn lint not working in Windows (+ test fixtures) 

lupestro:
This is a proposed fix for ember-cli issue #9428 - the fix tested locally with a project on Windows, Mac, and WSL - test suite run locally.

PR #9988, opened 2022-08-04
[BUGFIX] - Address npm-run-all and Yarn 3 conflict & Removed warning

christianarty:
## Details
* There is an ongoing [issue with `npm-run-all`, Yarn 2+, and Glob Matching ](https://github.com/mysticatea/npm-run-all/issues/200)
  * The general fix for it is instead of `npm-run-all lint:*` we would put `npm-run-all 'lint:*'`
  * The `-n` is for printing the name of the task. For Yarn 2+ users, without this flag, you would not know which subtasks are running from stdout
* Disabled warning for users using Yarn 2+ and linking it with the node-modules folder instead of the new plug n' play mechanism

## Testing
* Ran `yarn test:all` and it succeeded. 
```
  1592 passing (22m)
  13 pending
```
* Tested my changes to a blueprint and verified that the `npm-run-all` command still works as expected for Yarn 1 and Yarn 2+ versions

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -12,14 +12,14 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
-    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
-    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint": "npm-run-all --print-name --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
+    "lint:fix": "npm-run-all --print-name --aggregate-output --continue-on-error --parallel \"lint:*:fix\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
     "start": "ember serve",
-    "test": "npm-run-all lint test:*",
+    "test": "npm-run-all --print-name \"lint\" \"test:*\"",
     "test:ember": "ember test"
   },
   "devDependencies": {
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -63,7 +63,11 @@ class NpmTask extends Task {
 
       if (semver.gte(version, '2.0.0')) {
         logger.warn('yarn --version: %s', version);
-        this.ui.writeWarnLine(`Yarn v2 is not fully supported. Proceeding with yarn: ${version}`);
+        let yarnConfig = await this.yarn(['config', 'get', 'nodeLinker']);
+        let nodeLinker = yarnConfig.stdout.trim();
+        if (nodeLinker !== 'node-modules') {
+          this.ui.writeWarnLine(`Yarn v2 is not fully supported. Proceeding with yarn: ${version}`);
+        }
       } else {
         logger.info('yarn --version: %s', version);
       }
--- a/tests/fixtures/addon/defaults-travis/package.json
+++ b/tests/fixtures/addon/defaults-travis/package.json
@@ -15,13 +15,13 @@
   "scripts": {
     "build": "ember build --environment=production",
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
-    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:*:fix\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
     "start": "ember serve",
-    "test": "npm-run-all lint test:*",
+    "test": "npm-run-all --print-name \"lint\" \"test:*\"",
     "test:ember": "ember test",
     "test:ember-compatibility": "ember try:each"
   },
--- a/tests/fixtures/addon/defaults-travis/package.json
+++ b/tests/fixtures/addon/defaults-travis/package.json
@@ -14,8 +14,8 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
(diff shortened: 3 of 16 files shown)


blueprints/app/files/package.json
```diff
@@ -12,14 +12,14 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
-    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
-    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint": "npm-run-all -n --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
+    "lint:fix": "npm-run-all -n --aggregate-output --continue-on-error --parallel 'lint:*:fix'",
```
rwjblue, line 16: Is there a long form for `-n`? I'd much prefer to have the long form names for commited scripts like this.

lib/tasks/npm-task.js
```diff
@@ -63,7 +63,11 @@ class NpmTask extends Task {
 
       if (semver.gte(version, '2.0.0')) {
         logger.warn('yarn --version: %s', version);
-        this.ui.writeWarnLine(`Yarn v2 is not fully supported. Proceeding with yarn: ${version}`);
+        let yarnConfig = await this.yarn(['config', 'get', 'nodeLinker']);
+        let nodeLinker = yarnConfig.stdout.trim();
+        if (nodeLinker !== 'node-modules') {
+          this.ui.writeWarnLine(`Yarn v2 is not fully supported. Proceeding with yarn: ${version}`);
+        }
```
rwjblue, line 70: Does `yarn config get nodeLinker` work on all Yarns (e.g. 1.x, 2.x, 3.x)? If it does, maybe we could just do that instead of calling `yarn --version` **and** `yarn config get nodeLinker`?

rwjblue: Locally in a yarn@1 repo, it does seem like it works (just returns `"undefined"`)

```
❯ yarn config get nodeLinker
warning package.json: No license field
undefined
```

rwjblue (changes requested):

blueprints/app/files/package.json
rwjblue: Why do we move to `'` for the `lint:fix` script but not `lint`?

christianarty: > Is there a long form for `-n`? I'd much prefer to have the long form names for commited scripts like this.
Fair point. There is a long form. `--print-name` Ill make a revision to include that. 

> Why do we move to ' for the lint:fix script but not lint?
I'll address this in my next revision.

lib/tasks/npm-task.js
christianarty: Im not opposed to doing that since the `yarn config get`command is present in all versions of Yarn, however, how would we distinguish the difference between yarn 1 vs yarn 3 , specifically for the warning message? `nodeLinker === undefined` as Yarn 1 whereas `[pnpm,pnp,node_modules].includes(nodeLinker)` for Yarn 3?  Additionally, I'm not sure if there is a case for Yarn 3 where the nodeLinker is undefined.

[Reference for Yarn 3 available values](https://yarnpkg.com/configuration/yarnrc#nodeLinker)

bertdeblock:
FYI, this PR seems to undo the fix implemented over in https://github.com/ember-cli/ember-cli/pull/9500.
Do these scripts still work on windows?

christianarty:
@bertdeblock 
>FYI, this PR seems to undo the fix implemented over in https://github.com/ember-cli/ember-cli/pull/9500.
Do these scripts still work on windows?

Yea Upon further research, I have found that the single-quotes will not be interpreted the same on Mac and Windows. See here: https://github.com/mysticatea/npm-run-all/issues/45

I will throw in a fix for that in my latest revision. 

> Somewhat related: I also created https://github.com/ember-cli/ember-cli/issues/9966 a while ago.

I like `concurrently`, however, I think there could be issues with migrating to it from what i know (please let me know if im wrong)
1. Concurrently allows for parallization of npm scripts, however there is not a clear way to run scripts sequentially. By default, `npm-run-all` runs scripts sequentially cross platform (i.e npm run script-1 && npm run script-2)
2. Additionally, how will this effect those using blueprints that build on `npm-run-all`. Would that be part of a larger change?

I don't think we are opposed to it, but i think that would be encompassed in a separate pr rather than this one. @rwjblue  do you have any comments about using `concurrently` instead of `npm-run-all`

bertdeblock:
@christianarty Thanks for the update. Yeah, I thought I'd just mention the `concurrently` issue just in case, but agreed, not in scope for this PR.

lib/tasks/npm-task.js
rwjblue: Maybe I'm misunderstanding the output on yarn 3, but won't `yarn config get nodeLinker` always reply consistently?

- Yarn 1 - `undefined`
- Yarn 2 - `pnp`
- Yarn 3 - whichever linker you have setup

I would imagine the logic to be to explicitly check for `pnp` (and trigger the warning) or we could probably also make it so that If we see any value _other than_ `undefined` or `node-modules` we would trigger the warning.

Does that make sense? Would that work?

rwjblue:
FWIW, I definitely agree that we should look into migrating to `concurrently` (or something else) since `npm-run-all` isn't being actively maintained.

lib/tasks/npm-task.js
christianarty: Ah okay. I see what you are saying. From thinking about it, i think that would work fine. I'll push out another revision to account for it. Thanks @rwjblue!

christianarty: Upon further investigation, I think we should keep the version call (although I would like to remove that extra call) . If we remove it, it would:

1) provide an inconsistent API for the `packageManager`. (AKA checkNPM and checkPNPM return name & version whereas yarn would return only the name)

2) Refactoring the `npm-tasks.js` to where version is not needed, would cause some issues for people on older versions of NPM. Basically, other PMs need the version number as well.

rwjblue: Sounds good, thanks for digging!

rwjblue (approved):

rwjblue:
Thanks @christianarty!

PR #9994, opened 2022-08-11
Handle build failures without exiting

ef4:
This builds on top of https://github.com/ember-cli/ember-cli/pull/9987 to solve the remaining piece of the problem.

@bendemboski's change is correct to avoid an *immediate* explosion when a postProcess hook rejects. But it still gives you a *delayed* explosion when the most recent build has errored and then somebody hits the express middleware.

The cause of that was in the history-support middleware. Since it only used `finally` and not `catch` around `await watcher`, if the most recent watcher build was rejected this would allow the resulting exception to escape, unhandled, killing Node 15+.

It's not easy to identify this class of bug because the stack trace itself is useless. We know where the exception originated -- the problem is figuring out who is bringing it back to life by consuming it and not capturing the resulting errors. I had to step through the express stack turning things off one-by-one to identify history-support middleware as the culprit.


ef4:
@bendemboski independently fixed this same bug on his earlier PR at the same time I was making this one. Closing in favor of #9987.

PR #10013, opened 2022-09-07
[BUGFIX] Handle rebuild failures without exiting

kellyselden:
Cherry picking https://github.com/ember-cli/ember-cli/pull/9987 into lts-3-28.


bertdeblock:
@kellyselden I _think_ we need to cherry pick https://github.com/ember-cli/ember-cli/pull/9914 as well if we want CI to pass.

bertdeblock (approved):

PR #10014, opened 2022-09-08
[BUGFIX release] Make sure newly installed addons are discovered when running `ember install`

bertdeblock:
Running `ember install addon-name` currently throws the following error:

```
Install failed. Could not find addon with name: addon-name
```

This regression seems to have been introduced by #9890, even though the fix in that PR seems logical.

### DIFF
--- a/lib/models/project.js
+++ b/lib/models/project.js
@@ -658,6 +658,7 @@ class Project {
   reloadAddons() {
     this.reloadPkg();
     this._addonsInitialized = false;
+    this._didDiscoverAddons = false;
     return this.initializeAddons();
   }
 


kellyselden:
Do you mind targeting the release branch?

bertdeblock:
@kellyselden Done!

simonihmig:
The fix here is needed for https://github.com/ember-cli/ember-cli/pull/9972, so would appreciate if we could land this soon! 🙏 😀

csprocket777:
When using ember-cli to create a new project, it created the project with Ember 3.22. I then upgraded it using ember-cli-update to 4.7.0 and am experiencing this issue. Am I experiencing this because of this upgrade process or is this still present in 4.7.0?

bertdeblock:
> When using ember-cli to create a new project, it created the project with Ember 3.22.

Probably because your globally installed version is v3.22?

> Am I experiencing this because of this upgrade process or is this still present in 4.7.0?

I don't think this fix has been released yet.

PR #10015, opened 2022-09-09
Update `glob` to v8

bertdeblock:
Using forward slashes in glob patterns is now required.
More info here: https://github.com/isaacs/minimatch#windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "fs-tree-diff": "^2.0.1",
     "get-caller-file": "^2.0.5",
     "git-repo-info": "^2.1.1",
-    "glob": "^7.2.0",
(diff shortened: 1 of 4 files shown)


kellyselden:
Would it make sense to switch glob packages like fast-glob, if they have better Windows paths support?

bertdeblock:
@kellyselden `fast-glob` has the same requirement regarding using forward slashes in glob patterns:
https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows

I can still do the switch if you feel `fast-glob` is better?

kellyselden:
In that case, I'm OK with this change. I thought it was a glob exclusive issue.

kellyselden (approved):

PR #10024, opened 2022-09-20
[ENHANCEMENT] Add `ember-source` to `peerDependencies` in `addon` blueprint

bertdeblock:
Didn't have the right permissions to modify the original PR: #9864.

Closes #9864.

### DIFF
--- a/blueprints/addon/additional-package.json
+++ b/blueprints/addon/additional-package.json
@@ -1,0 +1,20 @@
+{
+  "keywords": [
+    "ember-addon"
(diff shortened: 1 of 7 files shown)


blueprints/addon/additional-package.json
```diff
@@ -0,0 +1,20 @@
+{
```
bertdeblock, line 1: Seemed easier to move more stuff into an "additional" package.json file, instead of adding everything in JS land.

PR #10025, opened 2022-09-27
[ENHANCEMENT] Update NPM version constraints

bertdeblock:
This should resolve the following warning for apps using an NPM version higher than v6:

```
WARNING: Ember CLI is using the global npm, but your npm version has not yet been verified to work with the current Ember CLI release.
```


NullVoxPopuli (approved):
Thanks! Npm 8 has been out for ages!

kellyselden (approved):

PR #10026, opened 2022-09-29
[ENHANCEMENT] Display info message when running the `lint:fix` script post blueprint generation

bertdeblock:
This will make it clearer that something is still happening (and what) and should also improve the perceived speed when generating blueprints.

Motivation:
- https://github.com/ember-cli/ember-cli/pull/9995
- https://discord.com/channels/480462759797063690/480501885837770763/1025092185839894578

Result:
https://user-images.githubusercontent.com/7403183/193127543-b0f5ff56-c6d7-4212-ada0-63375fe33eaa.mov

### DIFF
--- a/lib/utilities/lint-fix.js
+++ b/lib/utilities/lint-fix.js
@@ -1,26 +1,27 @@
 'use strict';
 
-const fs = require('fs-extra');
 const execa = require('../utilities/execa');
+const isYarnProject = require('../utilities/is-yarn-project');
+const prependEmoji = require('../utilities/prepend-emoji');
 
-async function run(ui) {
+async function run(project) {
   let lintFixScriptName = 'lint:fix';
-  let cwd = process.cwd();
-  let packageJson = fs.readJsonSync('package.json');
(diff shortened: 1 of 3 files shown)


lib/utilities/lint-fix.js
```diff
@@ -1,26 +1,27 @@
 'use strict';
 
-const fs = require('fs-extra');
 const execa = require('../utilities/execa');
+const isYarnProject = require('../utilities/is-yarn-project');
+const prependEmoji = require('../utilities/prepend-emoji');
 
-async function run(ui) {
+async function run(project) {
```
bertdeblock, line 7: Passing in the project allows us to check the presence of the fix script via `project.pkg.scripts` instead of having to read the package file again. It also allows us to use `isYarnProject` instead of check the current directory for a `yarn.lock` file, which I assume didn't work properly when used in a Yarn workspace?

Windvis (approved):

kategengler (approved):

PR #10038, opened 2022-10-01
Update `filesize` to v10

bertdeblock:


lib/models/asset-size-printer.js
```diff
@@ -11,7 +11,7 @@ module.exports = class AssetPrinterSize {
   }
 
   print() {
-    const filesize = require('filesize');
+    const { filesize } = require('filesize');
```
bertdeblock, line 14: Only named exports now.

kategengler (approved):

PR #10044, opened 2022-10-04
[ENHANCEMENT] Improve readability of asset-size printer and include Brotli sizes

### LINKED ISSUE #10023: idea: Report compressed sizes with brotli at 11
Since the entire world now does this (right? right?!!?! if you don't you should) it would be awesome if the little print out after a production build or when using broccoli-concat-analyzer similarly did this.

bertdeblock:
Human readable output:

- Uses a table layout for better readability
- Sorted based on asset size

Screenshot 2022-10-05 at 10 25 33

JSON output:

- Contains a new `brotliSize` field for each file

Screenshot 2022-10-04 at 23 00 31

Downside: The addition of the Brotli size slows down the generation of asset sizes.

Closes #10023.


lib/models/asset-size-printer.js
```diff
@@ -5,46 +5,48 @@ const path = require('path');
 const walkSync = require('walk-sync');
 const workerpool = require('workerpool');
 
-module.exports = class AssetPrinterSize {
+module.exports = class AssetSizePrinter {
```
bertdeblock, line 8: We use `AssetSizePrinter` everywhere else, and the filename is `asset-size-printer` as well.

lib/models/asset-size-printer.js
```diff
@@ -5,46 +5,48 @@ const path = require('path');
...
   constructor(options) {
-    Object.assign(this, options);
+    this.outputPath = options.outputPath;
```
bertdeblock, line 10: Makes it clearer what's expected / passed in.

lib/models/asset-size-printer.js
```diff
@@ -5,46 +5,48 @@ const path = require('path');
...
+    this.ui = options.ui;
   }
 
-  print() {
-    const { filesize } = require('filesize');
-    let ui = this.ui;
+  async print() {
```
bertdeblock, line 14: Switched to "async - await" for both print functions.

lib/models/asset-size-printer.js
```diff
@@ -5,46 +5,48 @@ const path = require('path');
...
-        ui.writeLine(JSON.stringify({ files: entries }));
-      } else {
-        ui.writeLine(chalk.red(`No asset files found in the path provided: ${this.outputPath}`));
-      }
-    });
+  async printJSON() {
+    let files = await this.makeAssetSizesObject();
+
+    if (files.length !== 0) {
+      this.ui.writeLine(JSON.stringify({ files }));
```
bertdeblock, line 46: The `showGzipped` property was removed, so we can print everything as is.

lib/models/gzip-stats.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
bertdeblock, line 1: File was also renamed from `gzipStats.js` to `gzip-stats.js` for consistency.

bertdeblock:
This is fun.

Screenshot 2022-10-05 at 08 37 57

PR #10049, opened 2022-10-14
[ENHANCEMENT] Remove the `config/environment.js` file from the `addon` blueprint

bertdeblock:
Motivation:

- Most addons (that I know of) don't use or need this
- Since apps also have a `config/environment.js` file, it sometimes happens that addon authors mistakingly use this file to configure the dummy app - an unwanted side effect of this is that, the configuration also ends up in the consuming app

This doesn't mean that the file is not supported anymore of course, it's just not included by default.

Resolves part of https://github.com/ember-cli/ember-cli/issues/7486#issuecomment-811095821.

### DIFF
--- a/blueprints/addon/files/addon-config/environment.js
+++ b/blueprints/addon/files/addon-config/environment.js
@@ -1,5 +1,0 @@
-'use strict';
-
-module.exports = function (/* environment, appConfig */) {
-  return {};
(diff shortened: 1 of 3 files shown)


rwjblue (approved):
Awesome to finally see this go away!

PR #10053, opened 2022-10-17
Add support for node ESM addons

hjdivad:
Node has support for packages that export files in ES module syntax. See [module packages](https://nodejs.org/api/packages.html#modules-packages) in the node documentation.

This feature doesn't quite work with ember addons due to in-addon files being loaded:
  1. absolutely and
  2. via `require`

Relative `require`s can be made to work with [conditional exports](https://nodejs.org/api/packages.html#conditional-exports), but even that would require addon authors to write those files in commonjs format.

Absolute requires cannot be made to work unless the file loaded has a `cjs` suffix.

This PR makes it possible to write ember addons that opt in to ES module syntax.  The support is imperfect since config/environment.js still has to be in commonjs format, so in practice addon authors need to write:

```javascript
// config/package.json
{
  "type": "commonjs"
}
```

### DIFF
--- a/lib/models/builder.js
+++ b/lib/models/builder.js
@@ -44,7 +44,7 @@ class Builder extends CoreObject {
    */
   readBuildFile(path) {
     // Load the build file
-    let buildFile = findBuildFile('ember-cli-build.js', path);
+    let buildFile = findBuildFile(path);
     if (buildFile) {
       return buildFile({ project: this.project });
     }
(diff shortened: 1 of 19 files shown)


hjdivad:
@rwjblue obviously don't merge while there's a fixup! commit

I left it that way to make it easier to see the additional diff re the slow tests + watcher support

rwjblue (approved):

Issue #10055, opened 2022-10-18
Switch to eslint-plugin-n


Windvis:
[eslint-plugin-node](https://github.com/mysticatea/eslint-plugin-node) has been unmaintained for a while now. The repo owner is unresponsive to requests for adding extra maintainers to the repo, as can be seen in these issues:
- https://github.com/mysticatea/eslint-plugin-node/issues/341
- https://github.com/mysticatea/eslint-plugin-node/issues/300

The wider ecosystem is switching to [eslint-plugin-n](https://github.com/eslint-community/eslint-plugin-n), which is a fork of the original package that is being maintained by the community. I think we should also switch to that package in the blueprints (and probably ember-cli itself as well).

PR #10056, opened 2022-10-18
[INTERNAL] Use `eslint-plugin-n` instead of `eslint-plugin-node`

bertdeblock:
Related to #10055.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,8 +5,8 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2020,
   },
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -5,8 +5,8 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2020,
   },
-  plugins: ['node', 'prettier'],
```
bertdeblock, line 8: The `plugin:n/recommended` config already adds the `n` plugin.

Windvis (approved):

PR #10062, opened 2022-10-24
[ENHANCEMENT] Update the Prettier config in the blueprints

### LINKED ISSUE #10061: Tweak the Prettier config so it only applies to JS files by default
The [.prettierrc.js](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.prettierrc.js) file in the app/addon blueprints was added as part of the [Prettier integration RFC](https://rfcs.emberjs.com/id/0628-prettier#prettier).

One of the requirements was that Prettier should use single quotes for JS files instead of the double quotes that it uses by default. Since the RFC was only intended for JS files and Prettier was integrated via eslint a choice was made to use a config like this since it would only run over JS files:

```js
module.exports = {
  singleQuote: true,
};
```

This works fine, but it has the side effect that all other file types will also use single quotes which isn't always the most popular option. For example `.hbs` templates usually use double quotes.

Since more people are starting to use Prettier for other types of files it is not uncommon to add overrides so that it would use double quotes instead. This works in practice but it requires users to update the config file every time they want to use Prettier for different file types (and single quotes aren't the standard for that file type). A better solution would be to let Prettier use the default configuration for all file types, _but_ still use single quotes for JS files. This can be achieved with a config to like this:

```js
module.exports = {
  overrides: [
    {
      files: '*.{js,ts}',
      options: {
        singleQuote: true,
      },
    },
  ],
};
```

I think it would be a nice improvement to update the config in the blueprints to a setup like that and I can open a PR if there are no objections to it.

Windvis:
This updates the `.prettierrc.js` file so that it only applies the singleQuote config to `.js` (and `.ts`) files. This has the benefit that other file types will still use the default Prettier settings and apps no longer have to add config overrides to ensure that.

Closes #10061

### DIFF
--- a/blueprints/app/files/.prettierrc.js
+++ b/blueprints/app/files/.prettierrc.js
@@ -1,5 +1,12 @@
 'use strict';
 
 module.exports = {
-  singleQuote: true,
+  overrides: [
+    {
+      files: '*.{js,ts}',
+      options: {
+        singleQuote: true,
+      },
+    },
+  ],
 };


bertdeblock (approved):

rwjblue (approved):

jelhan:
Thanks a lot for this change! This will ease adoption of Prettier for Handlebars, YAML, Markdown and others.

PR #10078, opened 2022-11-03
[INTERNAL] CI updates

bertdeblock:
The `test:all` script runs _all basic tests_ and _all slow tests_. We run this script in CI for all supported Node versions on ubuntu and windows. We also have a `basic-tests` job that only runs the basic tests and that uses Node v14 and runs on ubuntu, macOS and windows. This means that, we run the basic tests twice for Node v14 on ubuntu and windows. This PR attempts to resolve that.


.github/workflows/ci.yml
```diff
@@ -34,12 +34,8 @@ jobs:
       - run: yarn lint
 
   basic-tests:
-    name: "Basic Tests - ${{ matrix.os }}"
-    runs-on: "${{ matrix.os }}-latest"
-
-    strategy:
-      matrix:
-        os: [ubuntu, macOS, windows]
```
bertdeblock, line 42: All basic tests are still run on `ubuntu` and `windows` in the `tests` job below via the `test:all` script.

.github/workflows/ci.yml
```diff
@@ -52,11 +48,9 @@ jobs:
       - run: yarn test
 
   tests:
-    name: "Node ${{ matrix.node-version }} - ${{ matrix.os }} "
+    name: "Node ${{ matrix.node-version }} - ${{ matrix.os }}"
     runs-on: "${{ matrix.os }}-latest"
 
-    needs: [linting, basic-tests]
```
bertdeblock, line 58: These jobs have no output that's required for the `tests` job.
Do we really need `needs` here? Omitting it should speed up CI.

PR #10091, opened 2022-12-02
[BUGFIX] Fix looking up the default blueprint for scoped addons

### LINKED ISSUE #10090: Ember cli can not find default blueprint for scoped addons
Starting from version 4.3, ember cli can not find default blueprint for addon if addon is a scoped package (i.e. `@scope/addon-name`). If I install such addon, ember cli behaves like it does not have default blueprint. If I run `ember g @scope/addon-name`, it gives `Unknown blueprint: @scope/addon-name`. Unfortunately I can not provide a specific addon example to reproduce this issue, because it is published to private registry. 

The cause of problem is [line 1511 of lib/models/blueprint.js](https://github.com/ember-cli/ember-cli/blob/master/lib/models/blueprint.js#L1511), which checks if blueprint name is a path ([related PR](https://github.com/ember-cli/ember-cli/pull/9387)). Scoped packages contain `/`, which is a `path.sep`.

GendelfLugansk:
Fixes #10090 by checking if addon name starts with '@'

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1508,28 +1508,20 @@ let Blueprint = CoreObject.extend({
 Blueprint.lookup = function (name, options) {
   options = options || {};
 
-  if (name.includes(path.sep)) {
-    let blueprintPath = path.resolve(name);
-    let isNameAPath = Boolean(blueprintPath);
-
-    if (isNameAPath) {
-      if (Blueprint._existsSync(blueprintPath)) {
-        return Blueprint.load(blueprintPath, options.blueprintOptions);
-      }
-
-      if (!options.ignoreMissing) {
-        throw new SilentError(`Unknown blueprint: ${name}`);
-      }
-
-      return;
-    }
-  }
-
   let lookupPaths = generateLookupPaths(options.paths);
 
   let lookupPath;
   for (let i = 0; (lookupPath = lookupPaths[i]); i++) {
     let blueprintPath = path.resolve(lookupPath, name);
+
+    if (Blueprint._existsSync(blueprintPath)) {
+      return Blueprint.load(blueprintPath, options.blueprintOptions);
+    }
+  }
+
+  // Check if `name` itself is a path to a blueprint:
+  if (name.includes(path.sep)) {
+    let blueprintPath = path.resolve(name);
 
     if (Blueprint._existsSync(blueprintPath)) {
       return Blueprint.load(blueprintPath, options.blueprintOptions);
--- a/tests/acceptance/addon-generate-test.js
+++ b/tests/acceptance/addon-generate-test.js
@@ -143,4 +143,17 @@ describe('Acceptance: ember generate in-addon', function () {
     await generateInAddon(['server']);
     expect(file('server/index.js')).to.exist;
   });
+
+  it('successfully generates the default blueprint for scoped addons', async function () {
+    await initAddon('@foo/bar');
+    await ember(['g', 'blueprint', '@foo/bar']);
+    await fs.outputFile('blueprints/@foo/bar/files/__name__.js', '');
+    await ember(['g', '@foo/bar', 'baz']);
+
+    expect(file('baz.js')).to.exist;
+  });
+
+  it(`throws the unknown blueprint error when \`name\` matches a folder's name, but doesn't include the \`${path.sep}\` char`, async function () {
+    await expect(generateInAddon(['tests'])).to.be.rejectedWith('Unknown blueprint: tests');
+  });
 });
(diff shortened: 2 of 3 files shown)


bertdeblock:
Thanks for reporting the issue and making this PR.
I had a look at the PR that introduced this regression and I think we can actually avoid the sep + `@` check if we write the lookup by path after the original lookup code:

```diff
Blueprint.lookup = function (name, options) {
  options = options || {};

- if (name.includes(path.sep)) {
-   let blueprintPath = path.resolve(name);
-   let isNameAPath = Boolean(blueprintPath);
-
-   if (isNameAPath) {
-     if (Blueprint._existsSync(blueprintPath)) {
-       return Blueprint.load(blueprintPath, options.blueprintOptions);
-     }
-
-     if (!options.ignoreMissing) {
-       throw new SilentError(`Unknown blueprint: ${name}`);
-     }
-
-     return;
-   }
- }

  let lookupPaths = generateLookupPaths(options.paths);

  let lookupPath;
  for (let i = 0; (lookupPath = lookupPaths[i]); i++) {
    let blueprintPath = path.resolve(lookupPath, name);

    if (Blueprint._existsSync(blueprintPath)) {
      return Blueprint.load(blueprintPath, options.blueprintOptions);
    }
  }

+ // Check if `name` itself is a path to a blueprint.
+ let blueprintPath = path.resolve(name);
+ if (Blueprint._existsSync(blueprintPath)) {
+   return Blueprint.load(blueprintPath, options.blueprintOptions);
+ }

  if (!options.ignoreMissing) {
    throw new SilentError(`Unknown blueprint: ${name}`);
  }
};
```

bertdeblock:
I also noticed that the tests added in https://github.com/ember-cli/ember-cli/pull/9387 actually pass _with and without_ the added code for supporting paths, so I'm going to try and fix that first to make sure we don't introduce any additional regressions.

GendelfLugansk:
In this case, can we just add empty line to `lookupPaths` array? Inside `generateLookupPaths`, for example. But I am not sure this approach would not break support for specifying a path for the generate command.

kategengler:
Should we add a test for this case?

bertdeblock:
https://github.com/ember-cli/ember-cli/pull/10092 was merged. Added a test for this fix as well.

tests/acceptance/addon-generate-test.js
```diff
@@ -143,4 +143,10 @@ describe('Acceptance: ember generate in-addon', function () {
     await generateInAddon(['server']);
     expect(file('server/index.js')).to.exist;
   });
+
+  it('successfully generates the default blueprint for scoped addons', async function () {
+    await initAddon('@foo/bar');
```
kategengler, line 148: should this use the `generateInAddon` helper the other tests use?

Should there also be a test in `generate-test.js` for the non-addon case?

lib/models/blueprint.js
```diff
@@ -1536,6 +1519,12 @@ Blueprint.lookup = function (name, options) {
     }
   }
 
+  // Check if `name` itself is a path to a blueprint:
+  let blueprintPath = path.resolve(name);
```
kategengler, line 1523: What happens if someone accidentally tries, for example, `ember g tests foo`? I believe right now they would get the unknown blueprint error, but after this change it would attempt to load it as a blueprint and throw in unknown ways.

tests/acceptance/addon-generate-test.js
bertdeblock: > should this use the generateInAddon helper the other tests use?

`generateInAddon` currently uses a fixed addon name:
https://github.com/ember-cli/ember-cli/blob/61b839456af26d35bd9546e6932b85fdf94fcff1/tests/acceptance/addon-generate-test.js#L48

and uses `initAddon` internally.

I could make the addon name configurable, but to avoid having to edit the other tests, I would have to use `await generateInAddon(['@foo/bar'], '@foo/bar');`, which I find less readable than having the separate `initAddon` + `ember` calls. `generateInAddon` also only runs a single `ember` command.

> Should there also be a test in generate-test.js for the non-addon case?

I'll look into it.

lib/models/blueprint.js
bertdeblock: Good call, I guess we still need a `/` check?

tests/acceptance/addon-generate-test.js
bertdeblock: > I could make the addon name configurable

It seems the util already looks at the second argument.

lib/models/blueprint.js
bertdeblock: Should be fixed now.

kategengler (approved):

bertdeblock:
@GendelfLugansk Thanks for opening the issue and this PR!

PR #10092, opened 2022-12-06
[INTERNAL] Fix tests related to generating blueprints using a path instead of a name

bertdeblock:
These tests were added in #9387, but they actually pass _with_ and _without_ [the added code](https://github.com/ember-cli/ember-cli/pull/9387/files#diff-0be5f01e6d22431d64f3da013c8f82658807a994ab75b368a150da89220585d2R1321-R1336), which makes them unreliable and makes it hard to know if the added code actually does what it's supposed to.

This was uncovered when reviewing #10091.

### DIFF
--- a/tests/acceptance/generate-test.js
+++ b/tests/acceptance/generate-test.js
@@ -145,13 +145,10 @@ describe('Acceptance: ember generate', function () {
   });
 
   it('allows a path to be specified to a blueprint', async function () {
-    await outputFile(
-      'blueprints/http-proxy/files/server/proxies/__name__.js',
-      "import Ember from 'ember';\n" + 'export default Ember.Object.extend({ foo: true });\n'
-    );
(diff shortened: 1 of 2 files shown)


tests/acceptance/generate-test.js
```diff
@@ -145,13 +145,10 @@ describe('Acceptance: ember generate', function () {
   });
 
   it('allows a path to be specified to a blueprint', async function () {
-    await outputFile(
-      'blueprints/http-proxy/files/server/proxies/__name__.js',
-      "import Ember from 'ember';\n" + 'export default Ember.Object.extend({ foo: true });\n'
-    );
-    await generate([path.resolve(`${__dirname}/../../blueprints/http-proxy`), 'foo', 'http://localhost:5000']);
```
bertdeblock, line 152: I'm guessing this also worked without the path-support code because this used an absolute path?

kategengler (approved):

PR #10108, opened 2022-12-12
[BUGFIX release] Correctly instantiate server watcher

bertdeblock:
Fixes:

```
instantiate Watcher with (await Watcher.build()).watcher, not new Watcher()
```

when running:

```
ember s
```

in a project that has a `server` directory.

Similar to https://github.com/ember-cli/ember-cli/pull/10101.

### DIFF
--- a/tests/unit/models/server-watcher-test.js
+++ b/tests/unit/models/server-watcher-test.js
@@ -17,10 +17,17 @@ describe('Server Watcher', function () {
     analytics = new MockAnalytics();
     watcher = new MockServerWatcher();
 
-    await ServerWatcher.build({
+    class ServerWatcherMock extends ServerWatcher {
+      setupBroccoliWatcher() {
+        this.watcher = watcher;
+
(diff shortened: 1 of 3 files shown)


tests/unit/models/server-watcher-test.js
```diff
@@ -17,10 +17,17 @@ describe('Server Watcher', function () {
...
+        this.watcher = watcher;
+
+        return super.setupBroccoliWatcher(...arguments);
+      }
+    }
+
+    await ServerWatcherMock.build({
       ui,
       analytics,
-      watcher,
```
bertdeblock, line 23: Assigning the `watcher` upon creation is not really how things work at the moment, and is why this test didn't catch `this.watcher` being `undefined` in the server-watcher's `constructor` when running `ember s` in a real project. This test now fails without the changes in this PR and succeeds with.

kellyselden (approved):

Issue #10063, opened 2022-10-25
Clean up ignore files in `app` and `addon` blueprints


bertdeblock:
Files included:
- `.eslintignore`
- `.gitignore`
- `.prettierignore`

Some "issues" I've noticed:

- All ignore files in an app contain `ember-try` related ignores even though we don't ship `ember-try` for apps
- These files still include `/bower_components/` - even though `ember-cli` still supports Bower, I don't think we should include this for new apps and addons
- Same goes for `/vendor/` - the `vendor` folder was recently removed here: https://github.com/ember-cli/ember-cli/pull/10050
- Is including `/node_modules/` in `.eslintignore` and `.prettierignore` even needed nowadays? I thought this folder was ignored by default by those tools
- Do we need to include `/coverage/`? Apps and addons don't ship with anything OOTB that generates a `coverage` folder

PR #10122, opened 2023-01-10
[ENHANCEMENT] Add Stylelint setup to `app` and `addon` blueprints [RFC 814]

bmish:
Implements this RFC:
* https://github.com/emberjs/rfcs/pull/814

### DIFF
--- a/blueprints/app/files/.stylelintignore
+++ b/blueprints/app/files/.stylelintignore
@@ -1,0 +1,8 @@
+# unconventional files
+/blueprints/*/files/
+
+# compiled output
+/dist/
+
+# addons
+/.node_modules.ember-try/
--- a/blueprints/app/files/app/styles/app.css
+++ b/blueprints/app/files/app/styles/app.css
@@ -1,0 +1,1 @@
+/* Ember supports plain CSS out of the box. More info: https://cli.emberjs.com/release/advanced-use/stylesheets/ */
--- a/tests/fixtures/addon/defaults-travis/package.json
+++ b/tests/fixtures/addon/defaults-travis/package.json
@@ -15,6 +15,8 @@
   "scripts": {
     "build": "ember build --environment=production",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
+    "lint:css": "stylelint \"**/*.css\"",
+    "lint:css:fix": "concurrently \"npm:lint:css -- --fix\"",
     "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
@@ -62,7 +64,11 @@
     "loader.js": "^4.7.0",
     "prettier": "^2.3.0",
     "qunit": "^2.15.0",
-    "qunit-dom": "^1.6.0"
+    "qunit-dom": "^1.6.0",
+    "stylelint": "^14.16.1",
+    "stylelint-config-prettier": "^9.0.4",
+    "stylelint-config-standard": "^29.0.0",
+    "stylelint-prettier": "^2.0.0"
   },
   "engines": {
     "node": "14.* || 16.* || >= 18"
--- a/tests/fixtures/app/nested-project/actual-project/package.json
+++ b/tests/fixtures/app/nested-project/actual-project/package.json
@@ -13,6 +13,8 @@
   "scripts": {
     "build": "ember build --environment=production",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
+    "lint:css": "stylelint \"**/*.css\"",
+    "lint:css:fix": "concurrently \"npm:lint:css -- --fix\"",
     "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
@@ -58,7 +60,11 @@
     "loader.js": "^4.7.0",
     "prettier": "^2.2.1",
     "qunit": "^2.14.1",
-    "qunit-dom": "^1.6.0"
+    "qunit-dom": "^1.6.0",
+    "stylelint": "^14.16.1",
+    "stylelint-config-prettier": "^9.0.4",
+    "stylelint-config-standard": "^29.0.0",
+    "stylelint-prettier": "^2.0.0"
   },
   "engines": {
     "node": "14.* || 16.* || >= 18"
--- a/blueprints/addon/files/npmignore
+++ b/blueprints/addon/files/npmignore
@@ -18,6 +18,8 @@
 /.gitignore
 /.prettierignore
 /.prettierrc.js
+/.stylelintignore
+/.stylelintrc.js
 /.template-lintrc.js
 /.travis.yml
 /.watchmanconfig
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -35,6 +35,7 @@ module.exports = {
       files: [
         './.eslintrc.js',
         './.prettierrc.js',
+        './.stylelintrc.js',
         './.template-lintrc.js',
         './ember-cli-build.js',<% if (blueprint !== 'app') { %>
         './index.js',<% } %>
--- a/blueprints/app/files/.stylelintrc.js
+++ b/blueprints/app/files/.stylelintrc.js
@@ -1,0 +1,5 @@
+'use strict';
+
+module.exports = {
+  extends: ['stylelint-config-standard', 'stylelint-prettier/recommended'],
+};
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -13,6 +13,8 @@
   "scripts": {
     "build": "ember build --environment=production",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
+    "lint:css": "stylelint \"**/*.css\"",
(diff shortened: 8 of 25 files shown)


blueprints/app/files/.stylelintignore
```diff
@@ -0,0 +1,25 @@
+# unconventional js
```
bmish, line 1: This file is currently an exact match of `.prettierignore`.

blueprints/app/files/app/styles/app.css
```diff
@@ -0,0 +1 @@
+/* placeholder */
```
bmish, line 1: Needed to avoid a lint violation:

```
app/styles/app.css
 1:1  ✖  Unexpected empty source  no-empty-source
```

blueprints/app/files/.stylelintignore
elwayman02: This should be sufficient, in my experience. The stylelintignore in our app at work is very similar.

tests/fixtures/addon/defaults-travis/package.json
```diff
@@ -62,7 +64,11 @@
     "loader.js": "^4.7.0",
     "prettier": "^2.3.0",
     "qunit": "^2.15.0",
```
elwayman02, line 66: Someone unrelated, but is it weird that this package.json file lacks dependencies like webpack and tracked-built-ins?

tests/fixtures/app/nested-project/actual-project/package.json
```diff
@@ -56,6 +58,10 @@
     "eslint-plugin-qunit": "^7.0.0",
     "loader.js": "^4.7.0",
     "prettier": "^2.2.1",
+    "stylelint": "^14.16.1",
```
elwayman02, line 61: Should this be alphabetized to be below `qunit-dom`?

tests/fixtures/addon/defaults-travis/package.json
elwayman02: A couple of the other package.json files below are missing one or both of those packages, inconsistently. Not something to fix in this PR, but maybe worth looking into if it wasn't intentional.

tests/fixtures/app/nested-project/actual-project/package.json
bmish: Good point, fixed.

tests/fixtures/addon/defaults-travis/package.json
bmish: Yeah, I've noticed that's been an inconsistency for a long time, not sure what's going on. But shouldn't affect my PR.

blueprints/app/files/.stylelintignore
Windvis: Why does it need to be an exact match? Some of these aren't applicable to .css files I think? Adding them just to be consistent seems a bit strange to me. It looks like a mistake when reading the file 😄.

bmish: That's a fair point. But I'm unclear which to omit. Most of these folders could have various types of files in them, and we might as well just tell stylelint to ignore searching them regardless.

bmish: That's a fair point. But I'm unclear which to omit. Most of these folders could have various types of files in them, and we might as well just tell stylelint to ignore searching them regardless.

elwayman02 (approved):
Thank you for taking this on!

blueprints/app/files/.stylelintignore
bertdeblock: We run the same setup at work for projects and all we really need at the moment is:

```
/dist/
```

I guess for addons, it might be slightly different.

blueprints/app/files/app/styles/app.css
bmish: That's one possibility. However, I lean against that because:

* It gets messy having to include disable comments in the blueprint
* It makes this file "aware" of stylelint
* If someone starts adding actual styles to `app.css`, they have to know to remove this stylelint disable comment (I'm sure we would all know that, but can't assume everyone would know to clean it up)

Instead, what I recommend is we add a brief, helpful comment to explain what `app.css` is for, perhaps just:

```css
/* Ember CLI supports plain CSS out of the box. You can add your CSS styles to here. More info: https://cli.emberjs.com/release/advanced-use/stylesheets/ */
```

That would be useful regardless of the presence of stylelint / this violation.

bertdeblock: Seems much better than just `/* placeholder */` to me 👍

blueprints/app/files/.stylelintignore
bmish: Yeah, I imagine most addons could get away with a small/minimal list, but the same could be said to some extent for the JS/prettier ignore lists. Since those lists have been seemingly built to include everything conceivable, it seems like we might as well use the same list. But if we know of some items that cannot possibly have CSS files under them, then I'm open to removing them.

bertdeblock: I made https://github.com/ember-cli/ember-cli/issues/10063 a while ago, but haven't got to it yet.
I think a lot of entries aren't relevant anymore for new(er) apps.
Personally, I don't see why we should include entries that are not needed.

bmish: I went ahead and removed:
* `bower`-related items - I suspect it would be very rare for anyone to use that with a new app
* `.lint-todo/`- not related (from ember-template-lint) 
* `.eslintcache` - not related
* `coverage/` - not related

Kept:
* `blueprints/` - if someone creates their own Ember CLI [blueprints](https://cli.emberjs.com/release/advanced-use/blueprints/)
* `dist/` - common for various build tool output
* `tmp/` - common for temp files
* `vendor/` - this is a common folder used for manually storing third-party deps, even if it doesn't come with Ember apps by default
* `node_modules/` - obviously we want to ignore this, I don't see it mentioned as a default value in the [doc](https://stylelint.io/user-guide/ignore-code/#files-entirely) but this might not be documented, also note that default values could change
* `.node_modules.ember-try/` - common dependency in Ember apps

My general approach here is to ensure that folders from common dependencies (e.g. ember-try) or folders that are common in the frontend ecosystem (e.g. `vendor`) get ignored, with the hope that the vast majority of users won't have to worry about this / touch this file.

elwayman02: @bmish please keep `.lint-todo/`, that is not a technology unique to ember-template-lint. TODO infrastructure exists for ember-template-lint, eslint, and stylelint.

elwayman02: More info here: https://github.com/lint-todo/stylelint-formatter-todo

bmish: Thanks! Just added back.

bertdeblock: - The `tmp` folder in Ember projects was only a thing before the system temp folder was used I think, so I don't think it's needed for Ember apps that will get this update
- `vendor` is not included by default anymore in the blueprint as it's an outdated concept with tools like `ember-auto-import`, so to me it makes sense to omit it here as well
- `node_modules` is ignored by default, it's mentioned here: https://stylelint.io/user-guide/configure/#ignorefiles
- `.lint-todo` doesn't contain any css files, so I don't see why it should be added?

bmish: I removed `node_modules`, thanks for confirming that.

I understand `tmp` and `vendor` aren't used by default anymore, but I still think they're useful to include. Many large/complex/older Ember apps I work with still use them, both for custom and legacy reasons. 

- `tmp` is commonly used for experimentation / temporary / custom script output 
- `vendor` is a common pattern in the web world for storing third-party resources, even if not necessary in Ember anymore as a result of ember-auto-import, and it was only just removed from Ember

I think there's also something to be said for keeping these ignore lists reasonably consistent between JS/prettier/stylelint. Seeing different behavior from each tool because their ignore lists are inconsistent could cause confusion.

bertdeblock: Older apps won't be affected by this afaict. If you update to the version this ships in, you won't be using the `tmp` folder anymore. If it's for custom reasons, I feel that's not really Ember CLI's responsibility. I have no projects where these folders are used. I also don't think these ignore files should be consistent for the sake of consistency, that's how they got bloated in the first place I think, and how people might become confused about what's actually needed and what's not. The app blueprint is also used for creating new projects, which means new projects start off with things in these files that are no longer relevant.

bmish: Okay, I have removed `tmp` and `vendor`.

bmish: Does it look good now? We should merge soon.

bertdeblock: `.lint-todo` doesn't contain any css files afaik, so I think we can safely remove that one as well.

blueprints/app/files/.stylelintignore
```diff
@@ -0,0 +1,12 @@
+# unconventional files
+/blueprints/*/files/
+
+# compiled output
+/dist/
+
+# misc
+!.*
```
bertdeblock, line 8: Why did we need this again?

blueprints/app/files/.stylelintignore
bmish: I removed `.lint-todo` again. If CSS files are expected in that folder, let us know.

Also removed `!.*` since I don't think hidden CSS files are common.

blueprints/app/files/.stylelintignore
bmish: Usually this is for un-ignoring hidden files. Removed since I don't imagine hidden CSS files are common.

bertdeblock (approved):

bertdeblock:
Thank you @bmish!

Issue #10112, opened 2022-12-21
CI is failing


kategengler:
See https://github.com/ember-cli/ember-cli/actions/runs/3737153532/jobs/6342229833#step:5:3032

ef4:
@kategengler, @kellyselden, and I spent some time poking at this. It seems like a possible race condition because if you start putting in breakpoints the bug goes away. If you look after the tests fail the unresolvable files are in fact present. So it seems like the resolution failure is happening during a small time window in which there's a race between the test setup infrastructure and the code that's looking for these modules.

kategengler:
It is always the same 5 failures in acceptance/smoke-test-slow:

1. Acceptance: smoke-test ember new foo, make sure addon template overwrites
2. Acceptance: smoke-test ember test --path with previous build
3. Acceptance: smoke-test ember test wasm
4. Acceptance: smoke-test lint fixing after file generation does not fix lint errors with --no-lint-fix
5. Acceptance: smoke-test lint fixing after file generation does fix lint errors with --lint-fix

They fail when run with `yarn test:all` or `yarn test:slow` but do not fail individually or when that entire file is run.

kategengler:
Narrowed to this:
` node --unhandled-rejections=strict tests/runner tests/acceptance/addon-smoke-test-slow.js tests/acceptance/smoke-test-slow.js`

If the former is run before the later, those 5 tests fail.

kategengler:
Narrowed down to `ember addon foo, clean from scratch` in the addon-smoke-test-slow file causing the error. Changing the command run there from `test` to `help` has no affect, so I determined it was something to do with the `ember` helper, however, other tests using the `ember` helper before `smoke-test-slow` do not cause the failures.

For now, changing the test to `await runCommand('node_modules/ember-cli/bin/ember', 'test')` solves the issue.

PR incoming.

kategengler:
The above solves this but does not explain what about the `ember` helper used in that particular test causes the later tests to fail.

PR #10123, opened 2023-01-11
Added missing broccoli-plugin to app and addon because of failing CI

candunaj:


kategengler:
Thanks for looking into CI. This does make CI pass but I don't believe it is the right fix; we wouldn't want to add broccoli-plugin to the app blueprint. 

I've been also looking at the problem yesterday and today and have narrowed it down to a single test that is polluting the setup for the failing tests. 

References https://github.com/ember-cli/ember-cli/issues/10112

PR #10134, opened 2023-01-18
Fix beta CI?

kategengler:
This commit makes tests pass locally, but no idea why


kategengler:
The current failures are now what was fixed on master with https://github.com/ember-cli/ember-cli/pull/10124

ef4 (approved):

PR #10161, opened 2023-02-02
[ENHANCEMENT] Remove NPM version checking

bertdeblock:
Resolves:

```
WARNING: Ember CLI is using the global npm, but your npm version has not yet been verified to work with the current Ember CLI release.
```

for all future majors.

### DIFF
--- a/lib/tasks/npm-task.js
+++ b/lib/tasks/npm-task.js
@@ -25,8 +25,6 @@ class NpmTask extends Task {
 
     // The command to run: can be 'install' or 'uninstall'
     this.command = '';
-
-    this.versionConstraints = '3 || 4 || 5 || 6 || 7 || 8';
   }
 
   get packageManagerOutputName() {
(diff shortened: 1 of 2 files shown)


lib/tasks/npm-task.js
```diff
@@ -25,8 +25,6 @@ class NpmTask extends Task {
 
     // The command to run: can be 'install' or 'uninstall'
     this.command = '';
-
-    this.versionConstraints = '3 || 4 || 5 || 6 || 7 || 8';
```
bertdeblock, line 29: Seems better to not check this anymore than to keep bumping this every new major.

NullVoxPopuli: I imagine `engines` in package.json should be sufficient, too

NullVoxPopuli (approved):

lib/tasks/npm-task.js
bertdeblock: I think so yes. We could even add `"npm": ">= 3"`, if we want to keep erroring on NPM versions below v3 I think? Though, not sure if that's really needed.

kellyselden (approved):

kellyselden:
This wouldn't be [INTERNAL] though as your title suggests though right? It is an external feature removal.

NullVoxPopuli:
how is this an external feature? :thinking:

kellyselden:
because it is user facing.

bertdeblock:
@kellyselden Updated to `ENHANCEMENT`, let me know if that's okay.

PR #10162, opened 2023-02-03
[ENHANCEMENT] Deprecate `outputPaths` build option

### LINKED ISSUE #9478: Deprecate outputPaths usage under Embroider
Under Embroider the output paths of `/dist` are no longer predetermined. For example the application's javascript which would normally be `/dist/assets/app_name.js` will become for example: `/dist/assets/chunk.adjasld33lk23jol.js` + `/dist/assets/chunk.zsiojladj20ljk23jl.js`. The amount of "chunks" can vary based on the size / structure of your app especially if you have route based code splits, lazy engines, and/or dynamic imports. Each of these will create their own chunk(s). This issue is a reminder that we need to add the appropriate deprecations before enabling Embroider.

bertdeblock:
Closes #9478.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -9,7 +9,7 @@ const p = require('ember-cli-preprocess-registry/preprocessors');
 const chalk = require('chalk');
 const resolve = require('resolve');
 
-const assert = require('../debug/assert');
+const { assert, deprecate } = require('../debug');
 const Project = require('../models/project');
 
 const concat = require('broccoli-concat');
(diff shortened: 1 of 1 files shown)


bertdeblock:
Draft, as I don't think we can land this before v5.1.0.

locks:
@bertdeblock I think it's time 😅 Might need to update the version values.

NullVoxPopuli:
what would they be? 5.2 or 5.3?

bertdeblock:
Updated. Beta has its own branch, so master should be `5.3.0` at the moment.

locks (approved):

boris-petrov:
I can't seem to find a migration path anywhere. What are we supposed to change `outputPaths` with?

bertdeblock:
It was deprecated without a replacement if I remember correctly:
https://deprecations.emberjs.com/ember-cli/v5.x#toc_ember-cli-outputPaths-build-option

PR #10170, opened 2023-02-05
[ENHANCEMENT] Exclude `broccoli-asset-rev` when generating apps using the `--embroider` option

### LINKED ISSUE #10164: Exclude `broccoli-asset-rev` when generating apps using the `--embroider` option
`broccoli-asset-rev` doesn't work under Embroider.

bertdeblock:
As `broccoli-asset-rev` doesn't / won't ever work under Embroider.

Closes #10164.


bertdeblock:
One test is failing on Embroider, because a dummy package is `require`ing `broccoli-filter` without it being installed.
It used to be provided by `broccoli-asset-rev`.

PR #10173, opened 2023-02-09
[ENHANCEMENT] Don't delete the newly-generated app directory when an error occurs

ef4:
If anything fails during app generation, the entire app gets deleted. This means you can't debug what went wrong.

It's especially painful when generating anything that has a typescript type-checking step in its postinstall (like the `@embroider/addon-blueprint`), because if any type errors leak in due to dependency drift you have no way to move forward without patching this `fm.remove` out of ember-cli.

### DIFF
--- a/lib/commands/new.js
+++ b/lib/commands/new.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const fs = require('fs-extra');
 const chalk = require('chalk');
 const Command = require('../models/command');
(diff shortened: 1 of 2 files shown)


bertdeblock:
There is a related test we should remove I think: `ember new cleans up after itself on error`.

bertdeblock (approved):

PR #10174, opened 2023-02-12
[INTERNAL] Clean up removal of `bower.json` and `package.json` files in `addon` blueprint

bertdeblock:


blueprints/addon/index.js
```diff
@@ -111,13 +111,6 @@ module.exports = {
   },
 
   async afterInstall(options) {
-    let packagePath = path.join(this.path, 'files', 'package.json');
```
bertdeblock, line 114: We don't generate these files anymore, so no need to keep around this clean-up-related code.
For `package.json`, we now use [a replacer](https://github.com/ember-cli/ember-cli/pull/10174/files#diff-0ee18a1bec40c2bce48535d75189607df32ab18631183de0297b3c3755d352aaR20-R22).
Generating a `bower.json` file was removed long ago.

kategengler (approved):

PR #10175, opened 2023-02-12
[CLEANUP] Clean up `blacklist-whitelist-build-options` deprecation

bertdeblock:


tests/unit/models/addon-test.js
```diff
@@ -193,7 +193,6 @@ describe('models/addon.js', function () {
 
         it('addon', function () {
           let app = {
-            importWhitelist: {},
```
bertdeblock, line 196: Unrelated, but came across this when scanning the codebase for the word `whitelist`.
`importWhitelist` is not used anymore, so might as well remove it.

ef4 (approved):
Looks good to me, thanks.

PR #10177, opened 2023-02-12
[CLEANUP] Clean up `ember-cli-jshint-support` deprecation

bertdeblock:


lib/broccoli/ember-addon.js
```diff
@@ -42,10 +42,6 @@ class EmberAddon extends EmberApp {
         }),
         vendor: null,
       },
-      jshintrc: {
```
bertdeblock, line 45: Used by `ember-cli-jshint` [here](https://github.com/ember-cli/ember-cli-jshint/blob/master/index.js#L39).

PR #10185, opened 2023-02-14
[CLEANUP] Drop support for the `EMBER_CLI_ERROR_ON_INVALID_ADDON` env flag

bertdeblock:
Having an invalid addon path in `ember-addon.paths` will now always throw an error instead of a warning.

### DIFF
--- a/tests/fixtures/addon/with-app-styles/package.json
+++ b/tests/fixtures/addon/with-app-styles/package.json
@@ -4,10 +4,7 @@
   "dependencies": {
     "something-else": "latest"
   },
-  "ember-addon": {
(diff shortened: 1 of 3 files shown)


tests/fixtures/addon/with-app-styles/package.json
```diff
@@ -4,10 +4,7 @@
   "dependencies": {
     "something-else": "latest"
   },
-  "ember-addon": {
-    "paths": ["./lib/ember-super-button"]
```
bertdeblock, line 8: The `ember-super-button` addon does not exist for this test fixture. Not needed as well.

kategengler (approved):

PR #10187, opened 2023-02-15
[ENHANCEMENT] Remove Node version checking

### LINKED ISSUE #10167: Remove Node version checking
Similar to https://github.com/ember-cli/ember-cli/pull/10161.

At the moment, Ember CLI warns when:

1. The detected Node version is not supported anymore
2. The detected Node version is not tested against

Theoretically, `1.` should not be necessary anymore. I think we can fully rely on `engines` for this?
And `2.` is slightly annoying for users who always follow Node's current version, especially since, most of the time, everything works fine. I've also seen cases where the warning is shown, and users just ignore it, even though the used Node version is the cause of their issue.

Normally, according to our Node support document, we should also test against Node's current version, but it seems it's been a while since we've done this (hence the warning users get). CI would probably take longer as well, and at the moment, CI can already take up to one hour or more.

We could also test against current Node, but only for the basic tests on ubuntu? Which doesn't take that long.
Though, I would get rid of the warning nonetheless.

bertdeblock:
Motivation is written out in #10167.

Closes #10167.


locks:
@bertdeblock do you have time to rebase this?

bertdeblock:
Rebased.

locks (approved):

PR #10188, opened 2023-02-17
[ENHANCEMENT] Update ESLint parser `ecmaVersion` to `latest` in `app` blueprint

elwayman02:
There's no particular reason we should still be using the 2018 version of the parser. This should be forward-compatible, and we certainly would want new application to run against the latest parser.

### DIFF
--- a/blueprints/app/files/.eslintrc.js
+++ b/blueprints/app/files/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
   root: true,
   parser: '<%= typescript ? '@typescript-eslint/parser' : '@babel/eslint-parser' %>',
   parserOptions: {
-    ecmaVersion: 2018,<% if (!typescript) { %>
+    ecmaVersion: 'latest',<% if (!typescript) { %>
     sourceType: 'module',
     requireConfigFile: false,
     babelOptions: {
--- a/tests/fixtures/addon/.eslintrc.js
+++ b/tests/fixtures/addon/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
   root: true,
   parser: '@babel/eslint-parser',
   parserOptions: {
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -3,14 +3,13 @@
...
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: 2022,
   },
   plugins: ['prettier'],
   extends: ['eslint:recommended', 'plugin:n/recommended', 'plugin:prettier/recommended'],
   env: {
     browser: false,
     node: true,
-    es6: true,
```
elwayman02, line 13: Per the docs, this overrides the parserOptions and sets it back to version 6 instead of the latest: https://eslint.org/docs/latest/use/configure/language-options

.eslintrc.js
```diff
@@ -3,14 +3,13 @@
 module.exports = {
   root: true,
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: 2022,
```
bertdeblock, line 6: Can you drop these changes from the PR? I prefer to not include internal changes in enhancement PRs.

blueprints/app/files/.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   root: true,
   parser: '<%= typescript ? '@typescript-eslint/parser' : '@babel/eslint-parser' %>',
   parserOptions: {
-    ecmaVersion: 2018,<% if (!typescript) { %>
+    ecmaVersion: 2022,<% if (!typescript) { %>
```
bertdeblock, line 7: Can we use `'latest'` instead? I think that's the default as well in the new ESLint config system.

bertdeblock (commented):
Thanks for picking this up!

PR #10189, opened 2023-02-18
[ENHANCEMENT] Add v4.8 LTS to `addon` blueprint - Remove v3.28 LTS and `ember-classic` scenario

bertdeblock:


blueprints/addon/additional-package.json
```diff
@@ -11,7 +11,7 @@
     "ember-try": "^2.0.0"
   },
   "peerDependencies": {
-    "ember-source": "^3.28.0 || ^4.0.0"
+    "ember-source": "^4.4.0"
```
kategengler, line 14: I think it would make sense for this to remain "^4.0.0" even if only testing newer 4.x. It would be a shame for updating addons to break in projects only because of the peerDep requirements.

kategengler:
I think the peerDep on ember-source in the blueprint should remain pretty loose, other than that, it lgtm.

blueprints/addon/additional-package.json
bertdeblock: I understand the concern, but that would also mean we need to update the Compatibility section in the README to be in sync with this I think?

AFAIK, we've always used the oldest LTS version in the README:

```
* Ember.js v4.4 or above
* Ember CLI v4.4 or above
```

kategengler: IIRC having a peerDependencies section is relatively new. The README has matched what the ember-try scenarios have had. I think it will cause unnecessary pain when addons update the blueprint but otherwise do not make changes and still remain compatible with older versions, even if the peerDeps say otherwise. It's really on the addon author to not make that bump without a major bump but I'm not sure how many will realize that. Once again, I wish we could put comments in package.json. 

cc @ef4 for an opinion.

ef4:
Given how ember-cli-update works, I think I agree with @kategengler that it's better to be looser here.

Ideally, we could put the stricter value here for newly-generated addons without accidentally causing upgrading addons to break their users unnecessarily. But that would require changes to ember-cli-update so that this peer dep (and the embrer-try config) get a different update strategy.

bertdeblock:
Thanks both! Updated the PR accordingly.

kategengler (approved):

PR #10190, opened 2023-02-19
[ENHANCEMENT] Improve concurrently-based scripts in `app` blueprint

bertdeblock:


blueprints/app/files/package.json
```diff
@@ -12,17 +12,17 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
-    "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
+    "lint": "concurrently --group --raw \"npm:lint:*(!fix)\"",
```
bertdeblock, line 15: - `--group`: orders the output _as if_ the commands are run sequentially
- `--raw`: only log the raw output of tasks (no prefixes)

This improves readability of the output.

blueprints/app/files/package.json
```diff
@@ -12,17 +12,17 @@
...
+    "lint:css:fix": "stylelint \"**/*.css\" --fix",
+    "lint:fix": "concurrently --group --raw \"npm:lint:*:fix\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix<% if (typescript) { %>",
     "lint:types": "tsc --noEmit<% } %>",
     "start": "ember serve",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
+    "test": "concurrently --group --kill-others-on-fail --raw \"npm:lint\" \"npm:test:*\"",
```
bertdeblock, line 25: `--kill-others-on-fail`: make sure to exit early when the a script fails.
AFAICT, this is the closest we can get to the old behaviour when we were using `npm-run-all`.
`npm-run-all` would actually run `lint` and `test:*` sequentially. `test:*` would not be run if `lint` failed.

blueprints/app/files/package.json
```diff
@@ -12,17 +12,17 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
-    "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
+    "lint": "concurrently --group --raw \"npm:lint:*(!fix)\"",
     "lint:css": "stylelint \"**/*.css\"",
-    "lint:css:fix": "concurrently \"npm:lint:css -- --fix\"",
-    "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
+    "lint:css:fix": "stylelint \"**/*.css\" --fix",
```
bertdeblock, line 17: Avoid nested `concurrently` invocations. This makes the output more readable.
Repeating the glob seems like an acceptable trade off, as it's not something that will change often. Reading / looking at the output will happen far more often.

blueprints/app/files/package.json
NullVoxPopuli: The error case:
image

The nothing is wrong case is pretty nice: 
image


but I think the error case is more confusing?

PR #10211, opened 2023-03-01
Automate output repos

NullVoxPopuli:
To help reduce human involvement with updating the output repos, I've moved the running of the output repo scripts from release-it to run based on github actions triggers.

Comments are in the new workflow which describe what happens when.

The workflow is configured to use a `GH_PAT` token that has access to each of the repos.
For testing, I used a "Classic" token, but a "fine grained" one could be used as well.       

-------------------------

Additionally, due to the varying circumstances in which the two styles of output repos need to be updated, the code for updating the editor-output repo has been extracted to its own script. 
Notable changes in the editor-output scripts:
 - extra log at the top when the script runs, but is not applicable (hopefully shouldn't happen because the workflow only runs when release-branches are pushed)
 - JS/TS generation was extracted to an env variable so that we can run JS/TS output in separate jobs and they can error separately. This will help with debugging.

I've tested this change on my fork of [ember-cli here, using v4.10.0](https://github.com/NullVoxPopuli/ember-cli/actions/runs/4308484790)
 - C.I. logs:
   - when a tag is pushed: 
     - https://github.com/NullVoxPopuli/ember-cli/actions/runs/4308079722/jobs/7513916182
       - errorred, because my GH_PAT does not have push permissions to ember-cli (good)
   - when a release branch is pushed (previously, there was risk of these branches getting over-written by patch-releases older LTS versions): 
     - https://github.com/NullVoxPopuli/ember-cli/actions/runs/4305578543
       - successfully early exited as editor-output repo does not currently care about pre-releases
     - Successful publish for existing release v4.10.0:
       - well, JS errored, because it's already published: https://github.com/NullVoxPopuli/ember-cli/actions/runs/4308484790/jobs/7514815116
       - TS succeeded because it wasn't previously published: https://github.com/NullVoxPopuli/ember-cli/actions/runs/4308484790/jobs/7514815267
     - Branches on editor-output: https://github.com/ember-cli/editor-output/branches

### DIFF
--- a/dev/update-editor-output-repos.js
+++ b/dev/update-editor-output-repos.js
@@ -12,6 +12,8 @@ const isStable = !currentVersion.includes('-beta');
 const ONLINE_EDITOR_FILES = path.join(__dirname, 'online-editors');
 
 const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
+const VARIANT = process.env.VARIANT;
+const VALID_VARIANT = ['javascript', 'typescript'];
 
 if (!GITHUB_TOKEN) {
   throw new Error('GITHUB_TOKEN must be set');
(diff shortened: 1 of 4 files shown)


dev/update-editor-output-repos.js
```diff
@@ -12,6 +12,8 @@ const isStable = !currentVersion.includes('-beta');
 const ONLINE_EDITOR_FILES = path.join(__dirname, 'online-editors');
```
NullVoxPopuli, line 12: view this commit with invisible character changes hidden

dev/update-output-repos.js
```diff
@@ -57,12 +57,8 @@ async function updateRepo(repoName) {
 }
 
 async function main() {
-  try {
```
NullVoxPopuli, line 134: This change may be of importance.
I thought it ok to remove the try-catch, because running the script will print the error anyway.
Back when invoking this script via release-it, it probably made sense to swallow the error so release-it didn't fail.
But since the intent is to allow automation: https://github.com/ember-cli/ember-cli/pull/10211/files#diff-884b451b7291de286406620874239b6f24260847ef0b28d43beaa0b873d7032eR37

we no longer need to swallow the errors.

kategengler:
Thank you!

PR #10212, opened 2023-03-01
[INTERNAL] Fix version of Code Climate action

bertdeblock:


.github/workflows/coverage.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - run: yarn install --frozen-lockfile --non-interactive
 
       - name: Test && Report to Code Climate
-        uses: paambaati/codeclimate-action@v3
+        uses: paambaati/codeclimate-action@3.2.0
```
bertdeblock, line 21: Taken from the action's README:

> Please use a specific version of this action – for example, v3.2.0, instead of using only major versions like v3 or v3.2 – these will not work!

kategengler (approved):

PR #10195, opened 2023-02-25
[CLEANUP] Remove automatic inclusion of jQuery in legacy Ember versions

bertdeblock:

PR #10220, opened 2023-03-22
[CLEANUP] Drop support for including Ember builds via Bower

bertdeblock:
~~Still need to update the corresponding tests, but I'm going to wait until #10195 is merged, as that will make updating the tests easier.~~

I think after merging this, it should be easier to clean up the `ember-cli.building-bower-packages` deprecation.

### DIFF
--- a/lib/broccoli/ember-app.js
+++ b/lib/broccoli/ember-app.js
@@ -9,6 +9,7 @@ const p = require('ember-cli-preprocess-registry/preprocessors');
 const chalk = require('chalk');
 const resolve = require('resolve');
 
+const assert = require('../debug/assert');
 const Project = require('../models/project');
 
 let preprocessJs = p.preprocessJs;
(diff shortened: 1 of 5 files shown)


lib/broccoli/ember-app.js
```diff
@@ -369,46 +370,26 @@ class EmberApp {
...
-      productionEmber = `${this.bowerDirectory}/ember/ember.prod.js`;
-      developmentEmber = `${this.bowerDirectory}/ember/ember.debug.js`;
-      if (!fs.existsSync(this._resolveLocal(developmentEmber))) {
-        developmentEmber = `${this.bowerDirectory}/ember/ember.js`;
-      }
-      emberTesting = `${this.bowerDirectory}/ember/ember-testing.js`;
-    }
+    let emberSource = this.project.findAddonByName('ember-source');
+
+    assert(
```
bertdeblock, line 375: Theoretically, this should not be needed. Some tests were failing because we now always access `paths` on `ember-source` on line 384. This mainly provides a clearer error message during development.

tests/unit/broccoli/ember-addon-test.js
```diff
@@ -7,6 +7,18 @@ const EmberApp = require('../../../lib/broccoli/ember-app');
 const { expect } = require('chai');
 const MockCLI = require('../../helpers/mock-cli');
 
+const EMBER_SOURCE_ADDON = {
```
bertdeblock, line 10: I've repeated this a couple of times, simply because I couldn't find a decent sport to centralise this 😬.

kategengler (approved):

PR #10221, opened 2023-03-24
[CLEANUP] Clean up `building-bower-packages` deprecation

bertdeblock:
This cleans up the last remaining Bower deprecation. We still have some Bower related functionality / checks / references in place. Will be cleaned up in follow-up PR(s).


tests/unit/broccoli/default-packager/styles-test.js
```diff
@@ -44,20 +44,18 @@ describe('Default Packager: Styles', function () {
       },
       templates: {},
     },
-    bower_components: {
```
bertdeblock, line 47: Kept the files, but moved to the `vendor` tree.

kategengler (approved):

PR #10244, opened 2023-04-05
Update ember-cli-preprocess-registry and add ember-cli-clean-css to the blueprints

kategengler:
This moves css minification out of
ember-cli-preprocess-registry and into the addon.

I had failing unit tests for styles with the default packager. I have deleted them because the behavior is no longer built-in, but provided by ember-cli-clean css.

There are also failing unit tests around hasPodTemplates/hasTemplates. I have set these to skip because I believe we could get them to pass with the correct setup for the tests. `ember-cli-htmlbars` does not seem to be in these tests and I unfortunately do not know where in the test setup to correct that. I have confirmed by using my branch in `ember-new-output` that templates correctly compile with the updated ember-cli-preprocess-registry.

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -44,6 +44,7 @@
     "ember-cli": "~<%= emberCLIVersion %>",
     "ember-cli-app-version": "^6.0.0",
     "ember-cli-babel": "^7.26.11",
+    "ember-cli-clean-css": "^2.0.0",
     "ember-cli-dependency-checker": "^3.3.1",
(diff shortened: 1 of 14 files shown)


ef4 (approved):

kategengler:
two "slow" tests consistently fail with this PR under embroider https://github.com/ember-cli/ember-cli/actions/runs/4887941676/jobs/8725291333?pr=10244#step:5:1812 and https://github.com/ember-cli/ember-cli/actions/runs/4887941676/jobs/8725291333?pr=10244#step:5:1929

PR #10245, opened 2023-04-06
change guarding condition for output repos

NullVoxPopuli:
The branch condition assumed there were release branches, and there are none

### DIFF
--- a/.github/workflows/sync-output-repos.yml
+++ b/.github/workflows/sync-output-repos.yml
@@ -18,9 +18,6 @@ on:
     # for addon and new output
     tags:
       - 'v*'
-    # for editor output
-    branches:
-      - 'release-*'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-output-repos.yml
```diff
@@ -57,7 +57,7 @@ jobs:
 
   push-editors:
     name: "Push to editor output repos (${{ matrix.variant }})"
-    if: github.ref_type == 'branch'
+    if: github.ref_type == 'tag'
```
kategengler, line 60: Could you instead remove the guard on both jobs and change the `on` part to only run on tags?

NullVoxPopuli: can do!

PR #10249, opened 2023-04-06
Serve app on root url without trailing slash

### LINKED ISSUE #10246: Serve app at /rooturl when using the history location type
I noticed that ember-cli started defaulting locationType to history ~a year ago. This is great!

Unlike `auto` location, `history` location supports visiting the bare root url (i.e. no trailing slash) directly. For example, if your config contains `rootURL: '/example/``, with history location the app will work correctly when served at http://localhost/example (e.g. via nginx).

The issue I'm running into is that the ember-cli development server doesn't serve the app at this URL:

image

I think that we should modify the HistoryMiddleware to serve the app on the bare root url when using the history location type. Somewhere around here: https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/middleware/history-support/index.js#L89-L91

mmun:
Closes #10246.

I confirmed that my test fails without the middleware change and passes after the middleware change.

I also tried this out locally in an application that specifies `rootURL: "/example"`. Note that after the application boots up, Ember does a client-side route transition to normalize the URL to include the trailing slash.

https://user-images.githubusercontent.com/1151810/230454814-8f3d406b-af3b-4484-aaa4-7deeed3d3d03.mov

### DIFF
--- a/lib/tasks/server/middleware/history-support/index.js
+++ b/lib/tasks/server/middleware/history-support/index.js
@@ -87,8 +87,14 @@ class HistorySupportAddon {
       return false;
     }
     let rootURL = options.rootURL === '' ? '/' : cleanBaseURL(options.rootURL);
-    let rootURLRegexp = new RegExp(`^${rootURL}`);
-    return rootURLRegexp.test(req.path);
+    if (req.path.startsWith(rootURL)) {
+      return true;
+    }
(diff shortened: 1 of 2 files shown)


locks (approved):

NullVoxPopuli (approved):

locks:
Thank you @mmun! Always lovely to see you around 😁

PR #10270, opened 2023-05-08
FEATURE: Enable Ember's stable types

chriskrycho:
- In TS, add an `import 'ember-source/types'` statement.
- In JS, add a `/** @typedef {import('ember-source/types')} */` comment.

This is sufficient to make the types shipped in Ember itself as of emberjs/ember.js#20449 "enabled" for end users out of the box. It should be coordinated with an `ember-cli-typescript` release which *removes* the `@types/ember` packages from its own blueprint so that users do not get those installed needlessly.

(Maintainers: what's the best spot to test this?)

---

> **Note**: this is currently a Draft while we discuss how best to roll it out, given that Ember Data's types (`@types/ember-data` etc.) are not currently compatible with these types.

### DIFF
--- a/blueprints/addon/files/local-types/index.ts
+++ b/blueprints/addon/files/local-types/index.ts
@@ -1,0 +1,12 @@
+<% if (typescript) { %>
+// This "side-effect"-type import provides auto-complete, go-to-def, etc. for
+// Ember's internals throughout your application, so don't remove it!
+import 'ember-source/types';
+<% } else { %>
+// This "type definition" import comment provides auto-complete, go-to-def, etc.
+// for Ember's internals throughout your application, so don't remove it, even
+// if you do not use TypeScript at all.
+/**
+  @typedef {import('ember-source/types')} Types
+*/
+<% } %>
--- a/blueprints/app/files/app/app.ts
+++ b/blueprints/app/files/app/app.ts
@@ -10,3 +10,15 @@ export default class App extends Application {
 }
 
 loadInitializers(App, config.modulePrefix);
+
+<% if (typescript) { %>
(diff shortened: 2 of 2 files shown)


yarn.lock
```diff
@@ -22,6 +22,11 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.21.4.tgz#457ffe647c480dff59c2be092fc3acf71195c87f"
   integrity sha512-/DYyDpeCfaVinT40FPGdkkb+lYSKvsVuMjDAG7jPOWWiM1ibOaB9CXJAlc4d1QpP/U2q2P9jbrSlClKSErd55g==
 
+"@babel/compat-data@^7.21.5":
```
kategengler, line 25: Why is yarn.lock edited?

chriskrycho: Weird. Presumably the answer is `git commit -am "..."` and I just missed seeing this. Let me yank it.

chriskrycho: Force pushed with a version that has only the blueprint changes.

chriskrycho: And #10271 will prevent this happening to other folks.

kategengler:
I'm not positive on where to test it (I'm somewhat surprised nothing fails with this added). I'd think in https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/new-test.js and/or https://github.com/ember-cli/ember-cli/blob/master/tests/acceptance/init-test.js

kategengler:
Is this still relevant?

NullVoxPopuli:
blocked on ember-data being compatible with the stable types

Windvis:
This can probably be closed since #10506 was merged. That didn't include the type import comment for js projects though.

PR #10271, opened 2023-05-11
INTERNAL: fix the lock file

chriskrycho:
A fresh checkout and running `yarn` produces this diff. Commit it so contributors do not end up accidentally checking it in on unrelated changes.


chriskrycho:
@kategengler this is why #10270 had a lock file change. This should minimize the likelihood of that happening to other folks!

PR #10283, opened 2023-06-01
Refactor `--typescript` support in blueprints

### LINKED ISSUE #10045: Creating a new app or addon with the `--typescript` flag overrules the use of the `--skip-npm` flag
I.e. all node modules are installed regardless of if the user passed in the `--skip-npm` flag.
Probably because we use `addAddonToProject` to add the `ember-cli-typescript` addon to the generated app or addon.

simonihmig:
This is refactoring the `--typescript` support in app and addon blueprints according to the RFC update of https://github.com/emberjs/rfcs/pull/902. More specifically it will
* not use `ember-cli-typescript` for the scaffolding - which fixes #10045 and helps with a bunch of issues where the blueprint is used downstream: v2 addon blueprint, migration tools etc.
* not use `ember-cli-typescript` in the actual app/addon
* ~introduce Glint~
  actually I reverted my change to not do this (yet), because I had [some concerns about it](https://github.com/emberjs/rfcs/pull/902#discussion_r1213469949). We do add some Glint setup to make the (dummy) app what I was calling "Glint-ready" in the sense that it is able to consume Glint-enabled addons without running into type errors, as discussed [here on Discord](https://discord.com/channels/480462759797063690/484421406659182603/1115926217149915206), but it is not enabling full Glint type checking for the app itself. Omitting this potentially controversial choice felt like increasing the chances of getting this merged faster. If we still think full Glint-enablement should be part of the [revised RFC0800](https://github.com/emberjs/rfcs/pull/902), then we can add that later as a separate PR easily. 
* replace the `prepack`/`postpack` scripts for addon publishing from the commands provided by `ember-cli-typescript` to simple commands referring to `tsc` (`rimraf`) directly. For this I had to introduce a separate `tsconfig.declarations.json` (that replicates the [previous `tsc` options](https://github.com/typed-ember/ember-cli-typescript/blob/master/ts/lib/commands/precompile.ts#L29-L36)), as we need to change `includes` to not produce type declarations for tests, and there is no `--includes` we could pass to the `tsc` CLI. Also the declarations are now written to a separate `declarations` folder (`dist` is already taken), and mapped to by `typesVersions`. This was needed so we can easily clean up those files as part of `postpack` (actually, we could also just not clean up, and just gitignore the folder!? 🤔 ). Previously the `.d.ts` were written into the root (so `typesVersions` was not needed), but there was some special housekeeping in place to selectively remove those files again as part of the e-c-ts commands, which we cannot replicate easily without introducing custom commands again...

To Do:
- [x] figure out how to do the `prepack`/`postpack` scripts for addons _without_ e-c-ts
- [x] finalize discussion and get https://github.com/emberjs/rfcs/pull/902 merged? - are we blocked on this?
- [x] battle-test the produced output (especially TS addons consumed by TS apps)

---
I cannot request reviewers here, so pinging a few (active and former) CLI + TS folks for some of you to review please! 😀 
@kategengler @kellyselden @bertdeblock @ef4 @NullVoxPopuli @mansona @chriskrycho @gitKrystan @jamescdavis @wagenet @dfreeman


blueprints/addon/index.js
```diff
@@ -77,6 +82,19 @@ module.exports = {
...
     delete contents.devDependencies['ember-cli-app-version'];
 
+    // add scripts to build type declarations for TypeScript addons
+    if (this.options.typescript) {
+      contents.scripts.prepack = 'tsc --project tsconfig.declarations.json';
+      contents.scripts.postpack = 'rimraf declarations';
+
+      contents.typesVersions = {
+        '*': {
+          'test-support/*': ['declarations/addon-test-support/*'],
```
simonihmig, line 92: This is doing Ember's conventional mapping of `addon-test-support` (author) → `test-support` (consumer).

I have seen some issues though importing things from index-modules using that `typesVersions` approach, like having a `addon/foo/index.ts` module and trying to import that as `import foo from 'my-addon/foo'`, which did cause type errors, but `import foo from 'my-addon/foo/index'` did not. This seems to be some [bug in TypeScript itself](https://github.com/microsoft/TypeScript/issues/43133), but maybe someone here has a suggestion how to get around this? 

That's where definitely some more testing is needed...

blueprints/addon/index.js
```diff
@@ -77,6 +82,19 @@ module.exports = {
     // 100% of addons don't need ember-cli-app-version, make it opt-in instead
     delete contents.devDependencies['ember-cli-app-version'];
 
+    // add scripts to build type declarations for TypeScript addons
+    if (this.options.typescript) {
+      contents.scripts.prepack = 'tsc --project tsconfig.declarations.json';
+      contents.scripts.postpack = 'rimraf declarations';
```
IgnaceMaes, line 88: The `rimraf` command is not available by default when running `yarn pack`. Not sure if that's considered an issue?

blueprints/addon/files/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+  "extends": "@tsconfig/ember/tsconfig.json",
+  "compilerOptions": {
+    // The combination of `baseUrl` with `paths` allows Ember's classic package
+    // layout, which is not resolvable with the Node resolution algorithm, to
+    // work with TypeScript.
+    "baseUrl": ".",
+    "paths": {
+      "dummy/tests/*": ["tests/*"],
+      "dummy/*": ["tests/dummy/app/*", "app/*"],
+      "my-addon": ["addon"],
```
NullVoxPopuli, line 11: could `my-addon` be substituted with the `package.json#name`
Not everyone will use `my-addon` 😅

blueprints/addon/index.js
NullVoxPopuli: > here has a suggestion how to get around this?

```
"*": {
  "test-support": ["declarations/addon-test-support/index.d.ts"]
}
```
should cover that?

tests/fixtures/addon/typescript/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+  "extends": "@tsconfig/ember/tsconfig.json",
+  "compilerOptions": {
+    // The combination of `baseUrl` with `paths` allows Ember's classic package
+    // layout, which is not resolvable with the Node resolution algorithm, to
+    // work with TypeScript.
+    "baseUrl": ".",
+    "paths": {
+      "dummy/tests/*": ["tests/*"],
+      "dummy/*": ["tests/dummy/app/*", "app/*"],
+      "my-addon": ["addon"],
```
NullVoxPopuli, line 11: another "my-addon" reference

blueprints/addon/files/tsconfig.json
simonihmig: Oh geez! 🙈 😂

But tests are passing at least... 🙃

blueprints/app/files/package.json
```diff
@@ -36,8 +36,40 @@
     "@embroider/webpack": "^2.1.1<% } %>",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2<% if (typescript) { %>",
-    "@typescript-eslint/eslint-plugin": "^5.57.1",
-    "@typescript-eslint/parser": "^5.57.1<% } %>",
+    "@glint/core": "^1.0.2",
+    "@glint/environment-ember-loose": "^1.0.2",
+    "@glint/template": "^1.0.2",
+    "@tsconfig/ember": "^2.0.0",
+    "@types/ember": "^4.0.3",
```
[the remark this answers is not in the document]
simonihmig: Probably, but also I just didn't want to put multiple more or less unrelated changes into the same PR. Right now this is basically trying to replicate what e-c-ts did for us before...

blueprints/addon/files/tsconfig.json
simonihmig: Fixed!

tests/fixtures/addon/typescript/package.json
```diff
@@ -0,0 +1,127 @@
...
+    "lint:css": "stylelint \"**/*.css\"",
+    "lint:css:fix": "concurrently \"npm:lint:css -- --fix\"",
+    "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
+    "lint:hbs": "ember-template-lint .",
+    "lint:hbs:fix": "ember-template-lint . --fix",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "lint:types": "tsc --noEmit",
+    "prepack": "tsc --project tsconfig.declarations.json",
+    "postpack": "rimraf declarations",
```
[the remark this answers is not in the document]
simonihmig: Fixed!

blueprints/addon/index.js
simonihmig: Should be fixed now!

blueprints/addon/index.js
simonihmig: What about arbitrary modules you might want to export? Like `import foo from 'my-addon/foo/bar/baz'`

However coincidentally I was running into that also in another project, and we got it working _I think_ with something like that:

```
{
        '*': {
          'test-support/*': ['declarations/addon-test-support/*', 'declarations/addon-test-support/*/index.d.ts'],
          '*': ['declarations/addon/*', 'declarations/addon/*/index.d.ts'],
        },
```

- [ ] I need to double check that though...

NullVoxPopuli (approved):
nice work! so happy to see this landing :tada:

blueprints/addon/files/tsconfig.declarations.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "declarationDir": "declarations",
+    "emitDeclarationOnly": true,
+    "isolatedModules": false,
```
dfreeman, line 6:
```suggestion
    "rootDir": ".",
```

The `isolatedModules: false` flag that we had in `ts:precompile` was specifically to work around some incompatibilities with other flags that we needed to set in order to understand the structure of the output. Since we don't have tooling that's having to tweak flags any more and users can change their config however they like, forcing `isolatedModules` off should be unnecessary.

On the other hand, `tsc` can be a little cute about how it lays out its output: even though you have multiple directories in `include` below, if it only actually discovers `.ts` files in one of them, it will implicitly treat that as the input root. In other words, if the only TS file in your addon is `addon/index.ts`, it will generate `declarations/index.d.ts` without the intermediate `addon` directory. Explicitly setting `rootDir` here should ensure it always reflects the input structure appropriately.

blueprints/addon/files/tsconfig.declarations.json
```diff
@@ -0,0 +1,10 @@
...
+    "noEmit": false
+  },
+  "include": ["addon/**/*", "types/**/*", "addon-test-support/**/*"]
```
dfreeman, line 9: Should we be including `types` here? Since that's the equivalent of v2's `unpublished-development-types` directory, it seems to me like we'd want to ensure the addon's declarations can be successfully generated without relying on declarations in there, but I could be missing something?

(Also, as an aside: I think we can drop the `**/*` here and just use the directory names. IIRC requiring the globs was something we had to do to account for a weird corner case in `tsc`'s FS watcher back in `ember-cli-typescript@1.x` when we used `tsc` as part of the build to do transpilation.)

blueprints/addon/index.js
```diff
@@ -77,6 +82,21 @@ module.exports = {
     // 100% of addons don't need ember-cli-app-version, make it opt-in instead
     delete contents.devDependencies['ember-cli-app-version'];
 
+    // add scripts to build type declarations for TypeScript addons
+    if (this.options.typescript) {
+      contents.devDependencies.rimraf = '^5.0.1';
+
+      contents.scripts.prepack = 'tsc --project tsconfig.declarations.json';
+      contents.scripts.postpack = 'rimraf declarations';
```
dfreeman, line 90: Should we also add `declarations/` to `.gitignore`?

blueprints/app/files/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    // layout, which is not resolvable with the Node resolution algorithm, to
+    // work with TypeScript.
+    "baseUrl": ".",
+    "paths": {
+      "<%= name %>/tests/*": ["tests/*"],
+      "<%= name %>/*": ["app/*"],
+      "*": ["types/*"]
+    }
+  },
+  "include": ["app/**/*", "tests/**/*", "types/**/*"]
```
dfreeman, line 14: As above, I don't think we should need explicit `include`s any more.
```suggestion
  }
```

blueprints/app/files/package.json
```diff
@@ -36,8 +36,40 @@
     "@embroider/webpack": "^2.1.1<% } %>",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2<% if (typescript) { %>",
-    "@typescript-eslint/eslint-plugin": "^5.57.1",
-    "@typescript-eslint/parser": "^5.57.1<% } %>",
+    "@glint/core": "^1.0.2",
```
dfreeman, line 39: If we aren't setting up Glint for use in the actual app, then `@glint/core` is unneeded. That package doesn't have any types in it; it's just the CLI and language server machinery.

blueprints/addon/files/ember-cli-build.js
```diff
@@ -4,7 +4,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
   const app = new EmberAddon(defaults, {
-    // Add options here
+    <% if (typescript) {%>'ember-cli-babel': { enableTypeScriptTransform: true },
```
dfreeman, line 7: We also need to set this flag in the addon's `index.js` for it to work on the `.ts` files under `addon/`, right?

blueprints/addon/index.js
simonihmig: Yeah, thought about this as well... but then we don't _really_ need that rimraf script anymore, right? The previous behaviour of e-c-ts was to create the declarations only before publishing and immediately delete them afterwards, without them being git-ignored. And this is what is basically happening here as well. 

So should we git-ignore `declarations` and remove rimraf, or have both?

blueprints/addon/files/ember-cli-build.js
simonihmig: Oh, right! I had this already in my test addon I played with, but seems I forgot to apply that back to the blueprint 🤦‍♂️. Great catch!

blueprints/addon/index.js
dfreeman: Now that the declarations are in their own directory instead of scattered directly throughout the project, I think removing them isn't such a big deal, but I don't feel particularly strongly about it either way 🙂

blueprints/addon/files/ember-cli-build.js
dfreeman: We may also need to double check that this works correctly under Embroider. I had a recollection that there were issues with it a while back, and a quick discord search turned up [evidence of that being true back in early 2022](https://github.com/dfreeman/tsc-only/actions/runs/2201587701), but I'm not sure if it's been fixed in the meantime.

simonihmig: > I have absolutely no memory of having made that repo or I would have shared it much earlier

No problem, I feel you! 😜
But that's interesting, the changes there seem quite similar to what I have been doing here, so that makes me feel a bit more confident now! 😀

I just tested this with a TS-addon consumed in a TS-Embroider-app (with the new blueprint), and that was working fine actually! Maybe what was failing back then was fixed meanwhile (not sure what that was, given that the GitHub actions log is not available anymore. But I guess when you don't remember the repo, then you also don't remember the error message... :wink:)

blueprints/addon/index.js
simonihmig: I tested different variations of imports, and this is what seems to be working fine now:

```
  "typesVersions": {
    "*": {
      "test-support": [
        "declarations/addon-test-support/index.d.ts"
      ],
      "test-support/*": [
        "declarations/addon-test-support/*",
        "declarations/addon-test-support/*/index.d.ts"
      ],
      "*": [
        "declarations/addon/*",
        "declarations/addon/*/index.d.ts"
      ]
    }
  }
```

simonihmig:
Thanks @dfreeman for your review! I just pushed the changes addressing that. I resolved all the conversations here where I basically applied your suggestion. 

Only mildly controversial thing left is that rimraf vs. gitignore thing, which I also don't have strong opinions on (also personally I'm not really interested in creating new v1 addons anyway 😅). I'll leave this here for others to chime in...

blueprints/addon/files/ember-cli-build.js
dfreeman: > not sure what that was, given that the GitHub actions log is not available anymore

I think the issue then was that Embroider wasn't fully honoring the `enableTypeScriptTransform` flag and still required `ember-cli-typescript` to be present in the addon's `dependencies` to process any TS sources correctly.

> Maybe what was failing back then was fixed meanwhile

I think that's likely, since not processing out TS syntax would fail pretty spectacularly as soon as you tried to run the app 😄 
I also just did a quick check with a fresh Embroider app and everything seemed to work correctly, so this bit seems good to go 👍

blueprints/addon/index.js
ef4: Discussed with embroider core meeting and we recommend adding to gitignore and also calling rimraf either before a fresh build or after (so that extra files are not left around and picked up by typescript).

simonihmig: > Should we also add declarations/ to .gitignore?

Just did that!

The `.gitignore` blueprint file is used for both apps and addons, so apps now also get this change. Which isn't really needed. But also that file is already a bit blurry (yarn vs. npm, ember-try stuff for apps), so I think it's ok?

chriskrycho:
I think it's totally good and fine to merge this as it stands! The RFC states *intent* and we can keep iterating toward that. Thanks so much for driving this forward, @simonihmig!

NullVoxPopuli:
doh, looks like a conflict is present!

simonihmig:
Rebased this, and synced some fixtures, all blueprint tests are passing now, including [those for Embroider](https://github.com/ember-cli/ember-cli/actions/runs/5410307682/jobs/9831779204?pr=10283#step:5:138)!

The failing scenarios for Embroider seem unrelated, as these are also [failing on master](https://github.com/ember-cli/ember-cli/actions/runs/5408129731)! 🤔 

So I think this is good to merge now!? 🤞

kategengler (approved):
😭 Unfortunately CI started failing last night. I will merge this as soon as we get it green, since we cannot release until then.

simonihmig:
@kategengler seems the Embroider issue was fixed. I forced pushed to kick off another CI run, and... ✅

blueprints/addon/index.js
```diff
@@ -77,6 +82,22 @@ module.exports = {
     // 100% of addons don't need ember-cli-app-version, make it opt-in instead
     delete contents.devDependencies['ember-cli-app-version'];
 
+    // add scripts to build type declarations for TypeScript addons
+    if (this.options.typescript) {
+      contents.devDependencies.rimraf = '^5.0.1';
```
kellyselden, line 87: How are we going to keep this up to date automatically?

Issue #907, opened 2014-06-05
Defining a baseURL does not work


jerbob92:
When you define a baseURL other than '/' in config/environment.js, it will result in a broken application. 

The template will apply the basepath, so all assets will be requested from the basePath, but the ember-cli asset server does not use the basePath. 
I think the issue is in lib/commands/serve.js, when I hardcode my basePath in there, the app starts working again. 
this.project.config('development').baseURL is not set, and this.project.config('development') is empty...

PR #10287, opened 2023-06-07
Implementation of RFC 907 - pnpm support

NullVoxPopuli:
Supersedes https://github.com/ember-cli/ember-cli/pull/10130
RFC: https://github.com/emberjs/rfcs/pull/907
RFC Tracking PR: Ready for Release: https://github.com/emberjs/rfcs/pull/932
Docs: https://github.com/ember-learn/cli-guides/pull/290

The diff is mostly additional fixture files.


lifeart (approved):

kategengler:
This either needs a feature flag or to wait to be merged until the RFC moves to the Ready for Release stage.

NullVoxPopuli:
> This either needs a feature flag

How do consumers opt in to a feature flag once they've updated to a sufficently new ember-cli, once this is merged?

I'm not familiar with any cli-based feature flagging (I only have seen the flags from the ember-source consuming side)

> RFC moves to the Ready for Release stage

how can that happen without implementation to try out? :thinking:

kategengler:
I'm endeavoring to get us to follow the RFC Stages process. 

CLI has feature flagging, but it is called [experiments](https://github.com/ember-cli/ember-cli/blob/master/docs/experiments.md?plain=1)

We would turn on the feature flag as part of moving that RFC to "Released" once the RFC has been moved to ["Ready for Release"](https://github.com/emberjs/rfcs#ready-for-release).

NullVoxPopuli:
Thanks for the links! 
I still have a couple questions tho:

So we move the RFC to "Released" when code is isn't enabled by default?, but in the codebase -- folks have to _REALLY_ want to use `--pnpm` by _also_ prefixing their `ember-cli` commands with `EMBER_CLI_PNPM=true`

I think I'm confused?

According to `Ready for Release` docs:

> may be available for use behind a feature flag

Which means that it's not required? 
It seems like a lot of hubub for such a small change (_most_ of this PR is just fixtures)
it just feels redundant, I suppose -- I feel there is more risk in feature flagging this than not, and tests with the feature flag would be redundant with `--pnpm`. :shrug: 

I saw that embroider had a feature flag? but `--embroider` works today without adding an environment variable prefixing your command.
I think I'm _more confused now_. lol

I could implement a feature flag, but I think I need to not be confused about it first. lol

kategengler:
I'm confused by some of your questions 🙈 

RFC needs to get to the "Ready for release" stage and then we do the minor work to enable feature flags and actually release the code to move it to the "Released" stage. The thing we are releasing may or may not be a default for everybody but that depends on specifics of the particular RFC.

While there is a command line flag that is required for using this feature, by adding it, we are committing to the stability of that flag, and that is what the "Ready for Release" stage is there to get agreement upon. 

I believe the `EMBROIDER` experiment is not for the flag itself, but for enabling embroider by default.

[RFC #907](https://github.com/emberjs/rfcs/pull/907) is one of the first RFCs to fully go through the RFC stages process so I'm endeavoring to ensure we try following the process.

NullVoxPopuli:
Excellent! Thanks for explaining, you've been most helpful! 🎉 , i'll get those flags added before Tuesday

NullVoxPopuli:
I just pushed up all the feature flagging I thought was in the spirit of the RFC staging process, lemme know what you think.
I kept it all in one commit, so we can revert for the next stage.

blueprints/addon/files/.github/workflows/ci.yml
```diff
@@ -18,34 +18,41 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v3<% if (pnpm) { %>
+      - uses: NullVoxPopuli/action-setup-pnpm@v2
```
kategengler, line 22: I believe this is the first action we've used in a blueprint that isn't from an official source. 

Should we request it moved into an org we control (as we would do with an addon in the default blueprint). Thoughts @ef4

blueprints/addon/files/.npmrc
```diff
@@ -0,0 +1,8 @@
+# these settings allow us to be good stewards of the ecosystem
+# end-consumers can flip these settings in their own projects,
+# but library authors should not allow for less-than-strictest
+# behavior.
+#
+# For more information: https://pnpm.io/npmrc
+auto-install-peers=false
```
kategengler, line 7: I thought, from the RFC discussion, that auto-install-peers had become a default

NullVoxPopuli: Yes, as mentioned in the RFC, it's not good for library authors, because we must declare peers and deps correctly so that consumers _can_ rely on auto-install peers behaving correctly.

The auto install behavior does not propagate through the graph layers.

lib/commands/init.js
```diff
@@ -98,9 +100,15 @@ module.exports = Command.extend({
...
     await this.runTask('InstallBlueprint', blueprintOpts);
 
+    if (commandOptions.pnpm && !isExperimentEnabled('PNPM')) {
+      throw new Error(`--pnpm support is not enabled by default. Please prefix your command with EMBER_CLI_PNPM=true`);
+    }
+
     if (!commandOptions.skipNpm) {
       let packageManager =
-        commandOptions.yarn === true
+        commandOptions.pnpm === true
```
kategengler, line 109: This is getting pretty hard to read; would you mind refactoring it to have fewer ternaries?

lib/commands/init.js
```diff
@@ -125,7 +133,7 @@ module.exports = Command.extend({
     this.ui.writeLine('');
     const command = `cd ${projectName}`;
     this.ui.writeLine(`  ${chalk.gray('$')} ${chalk.cyan(command)}`);
-    const packageManager = commandOptions.yarn ? 'yarn' : 'npm';
+    const packageManager = commandOptions.pnpm ? 'pnpm' : commandOptions.yarn ? 'yarn' : 'npm';
```
kategengler, line 136: Request to simplify the ternaries again 🙏

lib/tasks/addon-install.js
```diff
@@ -37,7 +37,9 @@ class AddonInstallTask extends Task {
 
     let packageManager =
       commandOptions.packageManager ||
-      (commandOptions.yarn === true
+      (commandOptions.pnpm === true
```
kategengler, line 40: Maybe this should all move out to a utility (without the triple ternary), since it is in many places

tests/acceptance/new-test.js
```diff
@@ -540,6 +541,27 @@ describe('Acceptance: ember new', function () {
...
+      it('app + pnpm + welcome', async function () {
+        await ember(['new', 'foo', '--skip-npm', '--skip-git', '--pnpm']);
+
+        let fixturePath = 'app/pnpm';
+
+        ['app/templates/application.hbs', '.github/workflows/ci.yml', 'README.md'].forEach((filePath) => {
+          checkFile(filePath, path.join(__dirname, '../fixtures', fixturePath, filePath));
+        });
+
+        expect(file('.travis.yml')).to.not.exist;
```
kategengler, line 554: I know this is copied from other tests, but this assertion seems weird considering we have removed travis as an option afaik

tests/acceptance/new-test.js
```diff
@@ -561,6 +583,30 @@ describe('Acceptance: ember new', function () {
...
+          'tests/dummy/app/templates/application.hbs',
+          '.github/workflows/ci.yml',
+          'README.md',
+          'CONTRIBUTING.md',
+          '.npmrc',
+        ].forEach((filePath) => {
+          checkFile(filePath, path.join(__dirname, '../fixtures', fixturePath, filePath));
+        });
+
+        expect(file('.travis.yml')).to.not.exist;
```
kategengler, line 603: Again seems weird. I wonder what the absence of .travis.yml was supposed to be testing.

kategengler (changes requested):

tests/acceptance/new-test.js
NullVoxPopuli: Since travis is [still in the blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.travis.yml), i would bet that the assertion is, at most, half of the checks you'd need for supporting multiple ci providers.

With any ci provider option, you'd need to
- assert what you want exists in the output
- assert that the other options don't exist in the output

tests/acceptance/new-test.js
NullVoxPopuli: Seems like it should be done in another PR, as we don't want to mix purposes of PRs.
If we're going to remove travis, it should be done full-assedly.

Since the file still exists in the blueprint, i think it makes sense to include the assertion that travis is not present by default in the output

lib/tasks/addon-install.js
NullVoxPopuli: Can do!

lib/commands/init.js
NullVoxPopuli: rogerroger.gif

tests/acceptance/new-test.js
kategengler: Ok if it still exists in the blueprint. I think I was misremembering Bert removing travis support completely.

lib/tasks/addon-install.js
NullVoxPopuli: Done :tada:

lib/commands/init.js
NullVoxPopuli: Done :tada:

blueprints/addon/files/.github/workflows/ci.yml
NullVoxPopuli: I'm happy to transfer to _some org_ but it doesn't feel like it belongs under ember because it's also used for a bunch of non-ember projects.

I have another organization on GH that I started with @wycats, where, we have an action that wraps the above action here: https://github.com/wyvox/action -- the goal is to get get our repositories set up quickly, and efficiently, without bloating the ci configs and making them hard to manage _while also_ not forcing folks in to all of the features  (for example, the turbo server is only set up if you pass the GH token required for it to store artifacts).

I think it would make the most sense to transfer `NullVoxPopuli/action-setup-pnpm` to `wyvox/action-pnpm`

Thoughts?

(We can also add more folks to the org if that would put people at ease :tada: )

lib/commands/init.js
NullVoxPopuli: done :tada:

kategengler (approved):

blueprints/addon/files/.github/workflows/ci.yml
kategengler: I'm going to move the question of where the action to live to the RFC tracking PR and merge this.

PR #10289, opened 2023-06-10
Show a file diff when files don't match in the acceptance/new tests

NullVoxPopuli:
Work extracted from: https://github.com/ember-cli/ember-cli/pull/10287
This is helpful when working with blueprints, because the files are kind of large, and all chai-files says is "expect file to have content <the full expected content>" -- this behavior from chai-files isn't helpful when we're looking for minor differences between package.jsons, for example.

Whenever there is an error comparing files,
**before**
image

**after**

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -428,11 +428,16 @@ describe('Acceptance: ember new', function () {
       expect(file('.eslintrc.js')).to.equal(file(path.join(__dirname, '../fixtures', fixturePath, '.eslintrc.js')));
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
hm, nevermind, apparently there _are_ actual diffs after the meaningless text... :upside_down_face: oof
dat signal to noise ratio

NullVoxPopuli:
eh, I think this is entirely reporter related, tbh. 

At least the PR exists for folks to search for if they get stuck.

Meanwhile, I think we should probably change the reporter -- because trying to figure out diffs in CI-only failures is *rough* without a diff.

PR #10290, opened 2023-06-10
Update .eslintignore, so saving the doesn't run eslint on tests/fixtures

NullVoxPopuli:
Extracted from https://github.com/ember-cli/ember-cli/pull/10287

Without this change, if anyone opens the fixtures files and saves, and happens to have their editor set up to format on save, these files get formatted.

Without this change, eslint running on the files results in a big diff:
image

I do not know why removing the leading `/` fixes the problem, if anyone knows how to actually debug _why_ things happen with ignore files, that'd be immensely helpful.
I suppose I could node-debugger eslint -- but that's high effort for a little change.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,7 +4,6 @@
 /coverage/
 /docs/build/
 /node_modules/
-/tests/fixtures/
 /tmp/
 
 /lib/broccoli/app-*.js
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
potentially nevermind, this behavior doesn't reproduce in vscode -- so I guess I do have to debug something now

PR #10297, opened 2023-06-19
Update pnpm fixtures

NullVoxPopuli:
Looks we had a merge-order mishap. 

Maybe dependency updates were merged before the pnpm PR? idk.


blueprints/app/files/package.json
```diff
@@ -58,7 +58,7 @@
     "ember-qunit": "^7.0.0",
     "ember-resolver": "^10.1.0",
     "ember-source": "~5.0.0",
-    "ember-template-lint": "^5.10.1<% if (welcome) { %>",
+    "ember-template-lint": "^5.10.3<% if (welcome) { %>",
```
kategengler, line 61: It is probably fine to bump this but I am trying to reconcile it with your commit message of "How did ember-template-lint get bumped?"

kategengler:
LGTM just waiting on an answer to the one question

blueprints/app/files/package.json
NullVoxPopuli: The only thing I can think of is that I _accidentally_ applied a find and replace, where I thought I cancelled it.

NullVoxPopuli: I just pushed a rename of the commit with mentioning that 5.10.2 had a bug

kategengler: That follow up commit is the one that did the bumping though...

PR #10306, opened 2023-07-03
Set resolution-mode to highest for apps using pnpm

NullVoxPopuli:
Discussion here: https://discord.com/channels/480462759797063690/480501885837770763/1122582845873983529

tl;dr: 
```
# For more information: https://pnpm.io/npmrc
#
# The default is lowest-direct,
# which is good for library authors as it ensures they continue
# testing and developing against the bottom of their
# supported semver range for retaining maximum compatibility.
#
# but for apps we always want the latest bugfixes
```
Addons have not changed the `resolution-mode`.

pnpm made the decision for doing lowest-direct, which is great for ensuring maximum compatibility. 

There is a ton to read up on if you want more context: https://github.com/pnpm/pnpm/issues/6463

_NOTE_: for addons/libraries which are using secret / intimate dependencies (`@glimmer/syntax`, vm, etc), due to the pre-1.0 nature of semver behavior, folks will want to make sure that the versions are carefully managed -- as each minor version is actually a major, per semver, and if folks have too-low a version set up in their package.json, they'll likely get something very old) 

-----------------

Additionally, v1 addons will inherently have broken behavior (though it probably will not be noticed by many) because their dummy app setups do not correctly simulate a fresh app install (this includes ember-try, regardless of using npm, yarn, etc)

### DIFF
--- a/blueprints/app/files/.npmrc
+++ b/blueprints/app/files/.npmrc
@@ -1,0 +1,9 @@
+# For more information: https://pnpm.io/npmrc
+#
+# The default is lowest-direct,
+# which is good for library authors as it ensures they continue
+# testing and developing against the bottom of their
+# supported semver range for retaining maximum compatibility.
+#
+# but for apps we always want the latest bugfixes
+resolution-mode=highest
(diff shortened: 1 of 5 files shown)


bertdeblock:
Maybe we should wait on the outcome of https://twitter.com/pnpmjs/status/1693707270897517022?

NullVoxPopuli:
this is no longer needed, as of: https://github.com/pnpm/pnpm/releases/tag/v8.7.0

PR #10314, opened 2023-07-11
[BUGFIX] Do not try to wire up Testem unless a test framework is dete…

NullVoxPopuli:
…cted -- this is for legacy compat -- modern libraries, such as ember-qunit, call hookIntoTestFramework themselves, because it has better knowledge of when Testem needs to be configured

Update test to reflect changes to internal default-packager files

Backport of https://github.com/ember-cli/ember-cli/pull/10300

### DIFF
--- a/tests/acceptance/new-test.js
+++ b/tests/acceptance/new-test.js
@@ -21,6 +21,10 @@ const assertVersionLock = require('../helpers/assert-version-lock');
 
 let tmpDir = './tmp/new-test';
 
+let { stdout: nodeOut } = execa.sync('node', ['--version']);
+
+let isNode14 = semver.satisfies(nodeOut, '14.x');
+
 describe('Acceptance: ember new', function () {
   this.timeout(300000);
   let ORIGINAL_PROCESS_ENV_CI;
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
looking like floating deps (DT) caused the failures?

mansona:
The discussion on the meeting was to skip that failing test in Node 14 👍

tests/acceptance/new-test.js
```diff
@@ -625,7 +625,7 @@ describe('Acceptance: ember new', function () {
       checkFileWithEmberCLIVersionReplacement(fixturePath, 'tests/dummy/config/ember-cli-update.json');
     });
 
-    it('app + typescript', async function () {
+    it.skip('app + typescript', async function () {
```
ef4, line 628: The intention was to skip them in node 14, not skip them in every node.

NullVoxPopuli: I added a node check in the latest commit :partying_face: 

(I had tried checking NODE_VERSION on the env, but our test infra doesn't expose that, and I didn't think it made sense to mess with the test infra rn)

NullVoxPopuli: image


image

PR #10317, opened 2023-07-22
inline pnpm action

NullVoxPopuli:
after seeing how many lines this is, I think I'm just gonna move the action in to wyvox after researching what happens when I move an action


NullVoxPopuli:
superseded by https://github.com/ember-cli/ember-cli/pull/10319

PR #10318, opened 2023-07-26
[ENHANCEMENT] Bump `ember-resolver` and remove `@ember/string` in `app` blueprint

bertdeblock:


bertdeblock:
Never mind, `ember-data` still needs it.

runspired:
yeah I honestly think the removal from ember-resolver should have been RFC'd. Not sharing the cache between ember-resolver and ember-data is actually bad. We might want to create a configuration hook for both projects and let someone choose their string utils, but not great to have multiple copies that don't share the memoization.

NullVoxPopuli:
why do they need to share anything at all? 
what's ember-data using, memoization wise, that is also used in ember-reslover? 

(I want to get rid of ember-resolver entirely, and getting rid of complicated dependency sub-graphs, makes it much easier)

runspired:
@NullVoxPopuli camelize, dasherize, underscore. dasherize most heavily (which is what resolver also uses most heavily)

NullVoxPopuli:
oh I know that those functions would be copied, but you said you wanted shared memoization.

Do you have stats on how much of an impact not sharing memoization with ember-resolver would cause?
and is it worth a tougher migration to pure strict everything / v2 app / etc?

runspired:
1. how is it a tougher migration if it's just a peer-dep?
2. broad strokes, there's a number of apps I can think of where this would add ~50-100ms to TTI

NullVoxPopuli:
At what size of app would that cost incur? My intuition is that its an app with many _hundreds_ of models (where string transforming could take up to 100ms), and then the cost to TTI is already super slow due to a ton of other reasons so string transformation cost is a _small_ percentage of total TTI)

runspired:
50-100ms on these apps would be ~10-15% of their TTI even with hundreds of models. I don't disagree we want to kill the resolver, that would ultimately have a larger offsetting effect here, I just feel trying to nuke @ember/string isn't necessary.

PR #10332, opened 2023-08-02
feat: support converting gts files in blueprint

IgnaceMaes:
Add support to convert `.gts` blueprint files to `.gjs`.

Unblocks https://github.com/emberjs/ember.js/pull/20511#discussion_r1278580407


lib/models/blueprint.js
```diff
@@ -529,13 +529,15 @@ let Blueprint = CoreObject.extend({
   async convertToJS(fileInfo) {
     let rendered = await fileInfo.render();
 
-    const { removeTypes } = require('remove-types');
-    const transformed = await removeTypes(rendered);
-
-    fileInfo.rendered = transformed;
+    // removeTypes does not support GTS file format
+    if (path.extname(fileInfo.displayPath) === '.ts') {
```
IgnaceMaes, line 533: Can this realistically be merged without support? For now it would unblock basic `.gts` blueprint files that don't contain any types. (e.g. template-only components)

I wonder what the actual support would look like. Should we first transform it to valid TS, remove types, and transform back to GJS?

NullVoxPopuli: I think we'd need a custom parser? or maybe a light process -- like -- extracting the template(s), running babel to remove types, placing the template(s) back

IgnaceMaes: I've implemented a simple, naive version of extracting the template - removing the types - and placing back the template.

It does the job for simple cases. However, as the `removeTypes` function from `remove-types` also does formatting it kind of messes with the replacing. Because of this I'm not confident it would handle every possible case well.

ef4 (approved):

IgnaceMaes:
Is there any action to take here, or are we just hoping CI will succeed on another retry? 😬

ef4:
I think we need to turn of the Merge Queue feature. It doesn't support rerunning failed tests, and we have some flaky windows tests that make it hard to get things through on a single pass.

Issue #10344, opened 2023-08-19
WARNING: Ember CLI v5.2.0 is not tested against Node v20.4.0. S


Adesoji1:
Thank you for taking the time to open an issue!

For bug reports please include the following section in your issue details.
If you include instructions on how to reproduce the bug or a failing test case
it will make it easier for us to track down the issue you're having.

ember generate route register
WARNING: Ember CLI v5.2.0 is not tested against Node v20.4.0. See "https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md" to find out which version of Node is best to use.
You have to be inside an ember-cli project to use the generate command
---

Output from `ember version --verbose && npm --version && yarn --version`:
```
[Replace this line with the output.]
```

NullVoxPopuli:
thanks for reporting this issue!

I (personally) would like to remove this message entirely. :upside_down_face: 

In the meantime though, you can safely ignore that warning if everything is working fine, and I've submitted a PR here to address the warning: https://github.com/ember-cli/ember-cli/pull/10347

bertdeblock:
@NullVoxPopuli => https://github.com/ember-cli/ember-cli/pull/10187

NullVoxPopuli:
ah excellent! so in sufficiently new enough ember-cli there is no warning!

bertdeblock:
The warning check was removed and we now test against v20 => https://github.com/ember-cli/ember-cli/pull/10352

PR #10347, opened 2023-08-19
Test against node 20

NullVoxPopuli:
Resolves: https://github.com/ember-cli/ember-cli/issues/10344

In my searching, I couldn't find where the WARNING message is printed -- where is that?


NullVoxPopuli:
ooo interesting, live reload has a problem on node 20!

bertdeblock:
Closing in favour of #10352.

PR #10352, opened 2023-08-23
[ENHANCEMENT] Add official support for Node.js v20

bertdeblock:


tests/unit/tasks/server/livereload-server-test.js
```diff
@@ -100,7 +100,9 @@ describe('livereload-server', function () {
           port: 4200,
         })
         .then(function () {
-          http.get('http://localhost:4200/_lr/livereload.js', function (response) {
+          let httpGetOptions = { headers: { Connection: 'close' } };
+
+          http.get('http://localhost:4200/_lr/livereload.js', httpGetOptions, function (response) {
```
bertdeblock, line 105: I'm unsure if this is the right fix, but on Node.js v20, this `GET` request is being sent with `Connection: keep-alive` as opposed to `Connection: close` on older Node.js versions. `Connection: keep-alive` causes the `done` callback below to not being called anymore.

NullVoxPopuli: was there ever a problem with `ember test`? or was it just the tests here?

NullVoxPopuli (approved):

tests/unit/tasks/server/livereload-server-test.js
bertdeblock: I assume it's just this test. I think if `ember test` or any other command didn't work on Node.js v20, we would've gotten more bug reports already. Just a guess though!

kategengler:
Thank you!

PR #10354, opened 2023-08-25
[INTERNAL] Remove `@babel/core` as a dependency

bertdeblock:
Was added in https://github.com/ember-cli/ember-cli/pull/8427, to satisfy `@babel/plugin-transform-modules-amd`, but `@babel/plugin-transform-modules-amd` was removed in https://github.com/ember-cli/ember-cli/pull/10232, which means we can remove `@babel/core` again.


kategengler:
Thank you!

PR #10355, opened 2023-08-28
fix: Be more explicit about page-title being a helper rather than a component

ynotdraw:
With embroider, you need to be more explicit on differentiating components versus helpers.  By wrapping this in `(` and `)`, we are explicitly telling it that this is a helper rather than a component.  Even though embroider may not be adopted everywhere yet, I think getting folks in the habit of doing this now may make their migration path a lot easier.


NullVoxPopuli:
thank you!!!

bertdeblock:
See https://discord.com/channels/480462759797063690/485447409296736276/1015243372505608253.

kategengler:
The parens aren't necessary with embroider. @NullVoxPopuli Do you think this is relevant?

NullVoxPopuli:
It's needed whet `staticComponents: true` isn't set in ember-cli-build.js, because there is no way to statically differentiate `{{page-title}}` from component or helper.

I don't agree that it should be a part of the blueprint though, because (imo!!), the blueprint should be static-all-the-things-first, rather than partially migrated static-all-the-things

bertdeblock:
What do we want to do with this PR?

NullVoxPopuli:
We should probably close it. We lint against unneeded parens now

PR #10370, opened 2023-09-08
When generating a new app with --embroider use all optimisation flags

mansona:
~Firstly I don't know if this sort of change needs an RFC 🤔 let me know if you think it does 👍~ Edit: we agreed that this wasn't RFC material 👍 

My reasoning for this change is that we are at the point now that when someone generates a new app with `--embroider` it's reasonable for them to have all the optimisations turned on by default and for them to turn them off as needs be. This will prevent people from staring with a new Embroider app and adding code that can't be optimised **accidentally** rather than actually consciously opting in/out of certain behaviours


NullVoxPopuli:
I'm a big fan. I would be pretty annoyed if this were an additional RFC. 
We'll still need an RFC for making embroider _default_, but we can cross that when we get there

mansona:
@NullVoxPopuli I don't think it would be annoying for this to be an RFC, I'm happy to go through that process if we think it warrants it 👍 I actually started writing this as an RFC first but then thought I would chance it as a PR first 😂

kategengler:
I don't think this needs an RFC. It is for new projects and as you say we would want them to start optimized. Anybody upgrading with the `--embroider` flag would see the change and presumably understand whether they can turn on those flags.

blueprints/app/files/ember-cli-build.js
```diff
@@ -11,6 +11,12 @@ module.exports = function (defaults) {
 
   <% if (embroider) { %>const { Webpack } = require('@embroider/webpack');
   return require('@embroider/compat').compatBuild(app, Webpack, {
+    staticAddonTestSupportTrees: true,
+    staticAddonTrees: true,
+    staticHelpers: true,
+    staticModifiers: true,
+    staticComponents: true,
+    staticEmberSource: true,
     skipBabel: [
```
NullVoxPopuli, line 20: in my ember-cli-build, I also have these:
```
    // ember-inspector does not work with this flag
    // staticEmberSource: true,
    splitControllers: true,
    splitRouteClasses: true,
    packagerOptions: {
      webpackConfig: {
        // highest fidelity sourcemaps at the cost of speed
        devtool: 'source-map'
      }
    }
```

kategengler: Hmm, if ember-inspector does not work with that flag, I am less sure about recommending it be on by default, even with the `--embroider` flag

NullVoxPopuli: well, its a very new flag, and still opt in.

all the other options are about preventing userland code from regressing to be non-vite compatible (fully static)

staticEmberSource is kind of along those same lines (e.g.: "make sure you don't use the private require APIs!"), 

but small progress is still progress, and I think this PR is _good_ progress. Because there is no worse feeling than "trying the new thing", only to find out that "you weren't good enough", because the static flags weren't enabled by default.

in the afternoon RFC meeting today, we talked about a strategy for fixing the inspector in the long term for future ember-source changes (similar to what we do for supporting ember-data), and I really like that approach.

ef4 (approved):

mansona:
I converted this PR to draft because I have a WIP commit on the end to see if it fixes issues with embroider on CI. If it does I will go through and do a release of embroider and update this PR so it's ready to merge 👍

PR #10372, opened 2023-09-13
Reduce confusion when wanting to run tests

NullVoxPopuli:
For folks new to ember and who "just want to run tests", I don't think we should not be recommending `ember test` -- because it's headless.

`ember test --server` is also buggy, because testem is buggy, and kind of unmaintained.


kategengler:
Is this still relevant? It is a draft

NullVoxPopuli:
sort of -- but I'm waiting to see how ergonomics of testing will end up with Vite -- gonna close for now

PR #10373, opened 2023-09-18
add broccoli-plugin to new app to fix hoisting for tests

mansona:
first of all: this seems like a bad idea and we probably shouldn't merge but I don't really know how to fix this with the current CI setup 🤔 

CI is broken because of this file: https://github.com/ember-cli/ember-cli/blob/master/tests/fixtures/brocfile-tests/app-test-import/node_modules/ember-test-addon/index.js#L16

That fake addon/package needs to have access to `this.output` which seems to have been added as broccoli-plugin v3.1 in this PR: https://github.com/broccolijs/broccoli-plugin/pull/41

When i stopped in the debugger the `broccoli-plugin` version that it was hitting was v1.x that was being hoisted from a whole bunch of places. 

I don't know what the right solution to do here is, maybe we should add a valid dependency to https://github.com/ember-cli/ember-cli/blob/master/tests/fixtures/brocfile-tests/app-test-import/node_modules/ember-test-addon/package.json and then run yarn as part of the CI? or I could just commit a node modules folder to that addon with the correct `broccoli-plugin` version so that when fixtures are copied across they are correctly found 🤷 

Any thoughts on the right solution?

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -71,6 +71,7 @@
     "@typescript-eslint/eslint-plugin": "^6.4.0",
     "@typescript-eslint/parser": "^6.4.0<% } %>",
     "broccoli-asset-rev": "^3.0.0",
+    "broccoli-plugin": "^4.0.0",
     "concurrently": "^8.2.0",
     "ember-auto-import": "^2.6.3",
     "ember-cli": "~<%= emberCLIVersion %>",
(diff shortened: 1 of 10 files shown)


kategengler:
I don't think this is the right solution, as you say. The question is why is it suddenly getting the wrong version of broccoli-plugin. That fixture has been using `this.output` for two years.

mansona:
> The question is why is it suddenly getting the wrong version of broccoli-plugin

my current guess is that maybe something in our direct dependencies dropped broccoli (maybe went v2 or something) and we're falling back to an older version of broccoli-plugin being hoisted 🤔

mansona:
This isn't needed since https://github.com/ember-cli/ember-cli/pull/10374 was merged

PR #10374, opened 2023-09-18
Fix CI by adding broccoli-plugin to the test fixture so that it uses the correct

kategengler:
version -- there is no install step for the fixtures


mansona (approved):
Looks good to me 👍

PR #10375, opened 2023-09-18
use broccoli-plugin v1 api for test-addon

mansona:
closes #10373


mansona:
This isn't needed since https://github.com/ember-cli/ember-cli/pull/10374 was merged

PR #10393, opened 2023-10-26
[ENHANCEMENT] feat: add skip-install alias to skip-npm

IgnaceMaes:
The `--skip-npm` flag seems rather confusing now that three package managers are supported.

e.g. a valid way to generate an Ember app with pnpm today is:

```sh
ember new example-app --pnpm --skip-npm
```

With this alias it could be written as follows:

```sh
ember new example-app --pnpm --skip-install
```

I don't have a lot of context about the impact of this change, so please let me know if it isn't as simple as just adding an alias 😄

### DIFF
--- a/lib/commands/addon.js
+++ b/lib/commands/addon.js
@@ -11,7 +11,7 @@ module.exports = NewCommand.extend({
     { name: 'dry-run', type: Boolean, default: false, aliases: ['d'] },
     { name: 'verbose', type: Boolean, default: false, aliases: ['v'] },
     { name: 'blueprint', type: String, default: 'addon', aliases: ['b'] },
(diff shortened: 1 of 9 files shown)


NullVoxPopuli (approved):
lgtm -- do we have any docs/guides pages that need updated as well?

IgnaceMaes:
There doesn't appear to be any references [in the guides](https://github.com/search?q=repo%3Aember-learn%2Fguides-source%20skip-npm&type=code).

In the [cli docs](https://cli.emberjs.com/release/advanced-use/cli-commands-reference/) there are three references for which I opened a PR here: https://github.com/ember-learn/cli-guides/pull/308

kategengler:
Thank you!

PR #10394, opened 2023-10-27
[ENHANCEMENT] automatically select a port by default with `ember serve`

### LINKED ISSUE #9512: Thoughts on automatically selecting a new port if there is a port conflict?
I was starting an unrelated tool and had this output, which was quite the delight!
image

```
HTML Pages is Online! 🚀
Serving {project} at http://10.0.2.15:4201 (on port 4201, because 4200 is already in use)
The address was copied to the clipboard :)

Server is ready!
```
the port changing and dumping the address to the clipboard is neat (even though I'm just going to use localhost, cause reasons)

sportshead:
Fixes #9512

`ember serve` will now automatically choose a port (starting with 4200), unless a port is explicitly set with the `--port` flag or the `$PORT` env variable.  

## Notes:
The base port has also been changed from 7020 (which seems to have been arbitrarily chosen in 43878e2) to 4200, the current default. The old value of 7020 also doesn't seem to do anything since it's overridden by the default 4200 anyways.

The `should throw error when -p PORT is taken` test block has been moved out of a legacy if statement referencing appveyor.

### DIFF
--- a/lib/commands/serve.js
+++ b/lib/commands/serve.js
@@ -1,16 +1,12 @@
 'use strict';
 
 const Command = require('../models/command');
-const util = require('util');
 const SilentError = require('silent-error');
 const PortFinder = require('portfinder');
 const Win = require('../utilities/windows-admin');
 const EOL = require('os').EOL;
 
-PortFinder.basePort = 7020;
-
-let getPort = util.promisify(PortFinder.getPort);
-let defaultPort = process.env.PORT || 4200;
+const DEFAULT_PORT = 4200;
 
 module.exports = Command.extend({
   name: 'serve',
@@ -21,9 +17,11 @@ module.exports = Command.extend({
     {
       name: 'port',
       type: Number,
-      default: defaultPort,
+      default: process.env.PORT || DEFAULT_PORT,
       aliases: ['p'],
-      description: `To use a port different than ${defaultPort}. Pass 0 to automatically pick an available port.`,
+      description: `Overrides $PORT (currently ${
+        process.env.PORT || 'blank'
+      }). If the port 0 or the default port 4200 is passed, ember will use any available port starting from 4200.`,
     },
     { name: 'host', type: String, aliases: ['H'], description: 'Listens on all interfaces by default' },
     { name: 'proxy', type: String, aliases: ['pr', 'pxy'] },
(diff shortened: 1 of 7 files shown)


lib/commands/serve.js
```diff
@@ -21,9 +18,9 @@ module.exports = Command.extend({
     {
       name: 'port',
       type: Number,
-      default: defaultPort,
+      default: process.env.PORT,
```
NullVoxPopuli, line 21:
```suggestion
      default: process.env.PORT || DEFAULT_PORT,
```

`$PORT` may not exist, so we should fall back to the true default

tests/unit/commands/serve-test.js
```diff
@@ -61,44 +61,47 @@ describe('serve command', function () {
...
+      });
+    });
+  });
+
+  it('allows OS to choose port by default', function () {
+    return testServer({ port: '4200' }, function () {
+      return command.validateAndRun().then(function () {
+        let captor = td.matchers.captor();
+        td.verify(tasks.Serve.prototype.run(captor.capture()), {times: 1});
+        expect(captor.value.port).to.be.within(4201, 65535, 'has correct port');
```
NullVoxPopuli, line 102: oo nice!

NullVoxPopuli (approved):
Thanks a ton for working on this! this is a massive DX win!

lib/commands/serve.js
```diff
@@ -21,9 +18,9 @@ module.exports = Command.extend({
     {
       name: 'port',
       type: Number,
-      default: defaultPort,
+      default: process.env.PORT || DEFAULT_PORT,
```
sportshead, line 21: I got rid of `DEFAULT_PORT` from the default in the options originally to detect if it was left blank, perhaps this would be better? @NullVoxPopuli 
```suggestion
      default: process.env.PORT ?? 0,
```

sportshead:
I'll also make a patch to fix the test and lint errors

lib/commands/serve.js
NullVoxPopuli: the issue I see is that it doesn't communicate that the actual default is 4200.

when you say `0`, I think you're talking more about the _behavior_ when 4200 is taken -- which is more of an implementation detail of the behavior imo

sportshead: I'll keep 4200 on the default param and check for `DEFAULT_PORT` on line 108 instead

NullVoxPopuli:
@sportshead this is looking good -- can you get the tests passing? <3

sportshead:
@NullVoxPopuli - all the failed tests are referring to `ember help`, I'll try to rebase and see if that helps.

NullVoxPopuli:
the help output shows the description for the flags, so since the `-p` description changed, the help text change -- it's a type of snapshot test, if that makes sense

sportshead:
@NullVoxPopuli ah, that makes sense. I've updated the PR with the new changes for the help tests.

NullVoxPopuli (approved):

PR #10403, opened 2023-11-10
Fix some docs that were showing up weirdly in generated api docs

kategengler:


NullVoxPopuli (approved):

lib/models/per-bundle-addon-cache/target-instance.js
```diff
@@ -5,8 +5,9 @@
  * putting the symbol into the Symbol global cache. The symbol is used in per-bundle
  * cache entries to refer to the field that points at the real instance that a Proxy
  * refers to.
- * @property
+ *
```
kategengler, line 8:
```suggestion
 * @class TARGET_INSTANCE
```

lib/utilities/is-lazy-engine.js
```diff
@@ -3,7 +3,7 @@
 /**
  * Indicate if a given object is a constructor function or class or an instance of an Addon.
  *
- * @module is-lazy-engine
+ * @private
```
kategengler, line 6:
```suggestion
 * @private
 * @method
```

PR #10404, opened 2023-11-10
Add a workflow to deploy api docs to ember-learn/ember-cli.github.io

kategengler:
which will then auto deploy to the website

Also bump some actions.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,73 @@
...
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        type: string
+        description: 'Specify the released version of ember-cli to use to generate / update the output repos. Should be full semver version, and without a leading "v"'
```
NullVoxPopuli, line 12: This description may need updating 😅

kategengler, line 12:
```suggestion
        description: 'Specify the released version of ember-cli to use to generate and deploy documentation. Should be full semver version, and without a leading "v"'
```

PR #10414, opened 2023-12-11
update beta deps

kellyselden:


ef4:
The second fix is https://github.com/embroider-build/embroider/pull/1717

ef4:
The embroider fix is released in `@embroider/compat` 3.4.2.

PR #10416, opened 2023-12-18
Group tests' logs together so you know what log/error/warning/deprecation/etc came from which test.

NullVoxPopuli:
I cannot function without stuff like this, so I figure why not add it to the blueprint? :sweat_smile:


NullVoxPopuli:
I don't do much of anything in ember-cli anymore, so I'm going to close

PR #10417, opened 2023-12-18
Set a breakpoint when the window is about to unload.

NullVoxPopuli:
Any time the window is going to unload during tests, it's inanely hard and annoying to debug which test is being problematic.

Setting a debugger in code gives folks a stack that they can trace back to their test to see what was going wrong.

### DIFF
--- a/blueprints/app/files/tests/test-helper.ts
+++ b/blueprints/app/files/tests/test-helper.ts
@@ -5,6 +5,14 @@ import { setApplication } from '@ember/test-helpers';
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
 
+window.addEventListener('beforeunload', function() {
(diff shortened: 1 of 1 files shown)


blueprints/app/files/tests/test-helper.ts
```diff
@@ -5,6 +5,14 @@ import { setApplication } from '@ember/test-helpers';
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
 
+window.addEventListener('beforeunload', function() {
```
NullVoxPopuli, line 8: Note, and this is potentially workflow altering -- _live reload_ triggers this as well.

idk if there is a way to skip around this during live reload... but that'd be ideal.

NullVoxPopuli:
I don't do much in ember-cli anymore, so I'm going to close

PR #10418, opened 2023-12-21
[ENHANCEMENT] Use `content-tag` to parse GTS in blueprints

IgnaceMaes:
In https://github.com/ember-cli/ember-cli/pull/10332 support was added to use `.gts` files as blueprint source. It used [ember-template-tag](https://github.com/patricklx/ember-template-tag) to parse the source as it was able to give match ranges.

As of [v1.1.0](https://github.com/embroider-build/content-tag/releases/tag/v1.1.0), content-tag now also provides functionality to return match ranges. This PR refactors to use that package instead as it's the official one.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -556,7 +556,7 @@ let Blueprint = CoreObject.extend({
       let strippedCode = code;
       for (let i = 0; i < templateTagMatches.length; i++) {
         const match = templateTagMatches[i];
-        const templateTag = code.substring(match.range.start, match.range.end);
+        const templateTag = substringBytes(code, match.range.start, match.range.end);
         strippedCode = strippedCode.replace(templateTag, templateTagIdentifier(i));
       }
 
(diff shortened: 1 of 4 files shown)


lib/models/blueprint.js
```diff
@@ -547,16 +547,16 @@ let Blueprint = CoreObject.extend({
...
       const templateTagIdentifier = (index) => `template = __TEMPLATE_TAG_${index}__;`;
       const templateTagIdentifierBraces = (index) => `(template = __TEMPLATE_TAG_${index}__);`;
-      const templateTagMatches = parseTemplates(code).filter((match) => match.type === 'template-tag');
+      const templateTagMatches = preprocessor.parse(code);
       let strippedCode = code;
       for (let i = 0; i < templateTagMatches.length; i++) {
         const match = templateTagMatches[i];
-        const [matchStart, matchEnd] = match.range;
-        const templateTag = code.substring(matchStart, matchEnd);
+        const templateTag = code.substring(match.range.start, match.range.end);
```
NullVoxPopuli, line 559: Note, content-tag gives byte ranges, not character ranges.
See:
- https://github.com/embroider-build/content-tag/issues/45
- https://github.com/embroider-build/content-tag/issues/39#issuecomment-1832443310


Probably fine for now since the blueprints contain no multibyte characters tha

ef4: We should probably fix this, and ideally we should just offer char ranges from content-tag to make it easy.

IgnaceMaes: Good catch!

I've added a substring util based on byte ranges and added a test case to cover this.

ef4 (approved):

bertdeblock (approved):

bertdeblock:
This would also close https://github.com/ember-cli/ember-cli/pull/10418.
@NullVoxPopuli Can you merge?

PR #10425, opened 2024-01-08
fix GitHub Action floating deps scenario for PNPM

jelhan:
The _Floating Dependencies_ scenario is broken for a newly created Ember addon using PNPM. This change fixes it.

It seems that `wyvox/action-setup-pnpm`'s `no-lockfile` option is not working as expected. It throws the following error:

> Error: Dependencies lock file is not found in /home/runner/work/test-ember-addon-with-pnpm/test-ember-addon-with-pnpm. Supported file patterns: pnpm-lock.yaml

The option has been removed in `wyvox/action-setup-pnpm@v3`. So upgrading is not an option. But using `--no-lockfile` flag of PNPM is the recommended upgrade path anyways: https://github.com/wyvox/action-setup-pnpm/pull/3

Please find a demonstration of the bug here: https://github.com/jelhan/test-ember-addon-with-pnpm/actions/runs/7453402485/job/20278700434

A demonstration that the fix work could be found here: https://github.com/jelhan/test-ember-addon-with-pnpm/pull/1

@NullVoxPopuli would be great if you could review. I think you are the expert here as you implemented the PNPM support in Ember CLI _and_ maintain `wyvox/action-setup-pnpm`.

### DIFF
--- a/blueprints/addon/files/.github/workflows/ci.yml
+++ b/blueprints/addon/files/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: wyvox/action-setup-pnpm@v2
         with:
           node-version: 18
-          no-lockfile: true<% } else { %>
+          args: '--no-lockfile'<% } else { %>
       - uses: actions/setup-node@v3
         with:
           node-version: 18
--- a/tests/fixtures/addon/pnpm/.github/workflows/ci.yml
+++ b/tests/fixtures/addon/pnpm/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: wyvox/action-setup-pnpm@v2
         with:
           node-version: 18
-          no-lockfile: true
+          args: '--no-lockfile'
       - name: Run Tests
         run: pnpm test:ember
 


NullVoxPopuli:
yeah, the things you call out are exactly the motivation for the change in v3 :+1: 

> So upgrading is not an option. 

Upgrading to v3 should done (probably as separate PR), as there are important fixes in there (specifically around determining what tooling versions to use, based on your package.json (inherited upstream by way of upgrading setup-node and configuring it))

NullVoxPopuli (approved):

jelhan:
Thanks a lot for quick review.

> > So upgrading is not an option.
> 
> Upgrading to v3 should done (probably as separate PR), as there are important fixes in there (specifically around determining what tooling versions to use, based on your package.json (inherited upstream by way of upgrading setup-node and configuring it))

Sorry. I phrased it badly. I meant, upgrading is not an option _to fix this bug_. The changes landed with this PR unlock the upgrade. The upgrade to `wyvox/action-setup-pnpm@v3` should only require bumping the version number now. At least that's what RenovateBot done in my project and it worked find. :laughing:

PR #10440, opened 2024-02-07
Update optional-features.json

achambers:
Adding `no-implicit-route-model` for https://github.com/emberjs/rfcs/pull/970

## TODO

- [x]  Modify this to default to `true`

### DIFF
--- a/blueprints/app/files/config/optional-features.json
+++ b/blueprints/app/files/config/optional-features.json
@@ -3,5 +3,5 @@
   "default-async-observers": true,
   "jquery-integration": false,
   "template-only-glimmer-components": true,
-  "no-implicit-route-model": false
+  "no-implicit-route-model": true
 }
--- a/blueprints/app/files/config/optional-features.json
+++ b/blueprints/app/files/config/optional-features.json
@@ -2,5 +2,6 @@
   "application-template-wrapper": false,
   "default-async-observers": true,
   "jquery-integration": false,
-  "template-only-glimmer-components": true
+  "template-only-glimmer-components": true,
+  "no-implicit-route-model": false,
 }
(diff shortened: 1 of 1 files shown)


blueprints/app/files/config/optional-features.json
```diff
@@ -2,5 +2,6 @@
   "application-template-wrapper": false,
   "default-async-observers": true,
   "jquery-integration": false,
-  "template-only-glimmer-components": true
+  "template-only-glimmer-components": true,
+  "no-implicit-route-model": false
```
mansona, line 6:
```suggestion
  "no-implicit-route-model": true
```

I feel like we should be defaulting to the new (desired) behaviour when generating a new app with this blueprint 🤔 

I don't know if we have a policy for this written down for when we introduce a new optional feature, but considering we are hoping to remove this code in 6.0 it seems reasonable for new apps generated between now and then to be generated with the new stuff

mansona:
The CI is failing with the message: 

> Unknown feature "no-implicit-route-model" found in config/optional-features.json

Does this mean that we can't merge this until the version of ember-source that defines this optional feature is actually released? I don't know if we've done any version guarding on optional-features before 🤔

mansona:
Ah looking into it it's probably because @ember/optional-features hasn't been shipped. I'll work on get that released and this should start passing, and we should probably bump the version in the blueprint too to be safe ;)

blueprints/app/files/config/optional-features.json
bertdeblock: I agree. I think the same was done for the Octane optional features, and we also do the same for the `EXTEND_PROTOTYPES` env var => https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/environment.js#L10

kategengler: The defaults of the optional features in Octane were set to the "on" behavior only at Octane (the edition) being released. Prior to that they were the defaults. 

I agree with @mansona however, especially considering it targets 6.0, but that is not what the RFC says. @mansona could you ask this question on the RFC / propose a change to the text over there?

PR #10451, opened 2024-03-21
[ENHANCEMENT] Use a flat ESLint config file in the `app` blueprint

bertdeblock:
First time constructing a flat config, so I probably missed or misconfigured some things.
Still need to test this across various projects.

TODO:

- [ ] Ignore files and folders based on old `.eslintignore` file
- [ ] Use `browser` globals probably
- [ ] GJS / GTS by default
- [ ] Update tests


blueprints/app/files/eslint.config.mjs
NullVoxPopuli, line 1: imo, we should probably include gjs/gts support in here by default, since there is no cost to having config for gjs/gts if folks aren't using it.

bertdeblock: Could you provide a todo list for this? I know the recommended flat config of eslint-plugin-ember already brings in the correct parser, etc.

blueprints/app/files/eslint.config.mjs
```diff
@@ -0,0 +1,69 @@
...
+      parserOptions: {
+        babelOptions: {
+          plugins: [
+            [
+              '@babel/plugin-proposal-decorators',
+              { decoratorsBeforeExport: true },
+            ],
+          ],
+        },
+        ecmaVersion: 'latest',
```
bertdeblock, line 26: `latest` should be the default when using a flat config, so this can go?

blueprints/app/files/eslint.config.mjs
```diff
@@ -0,0 +1,69 @@
...
+        requireConfigFile: false,
+        sourceType: 'module',
```
bertdeblock, line 28: `module` should be the default when using a flat config, so this can go?

blueprints/app/files/eslint.config.mjs
bertdeblock: I need to use `recommended-gjs` or `recommended-gts` when needed it seems?

blueprints/app/files/eslint.config.mjs
NullVoxPopuli: sourceType is determined by the package.json `type` field, so I think we still need this since we lie about what type of package our projects are (CJS)

NullVoxPopuli:
@bertdeblock any chance you want to pick this back up?
would solve
- https://github.com/ember-cli/ember-cli/issues/10466
- https://github.com/ember-cli/ember-cli/issues/10467

bertdeblock:
Closing, as I don't have time to finish this.

PR #10458, opened 2024-04-22
Use Lodash's `_.template` instead of `lodash.template` package

### LINKED ISSUE #10459: Security vulnerability in lodash.template package
ember-cli is using a lodash function with a known security vulnerability. The vulnerability is not new, but due to how lodash used to publish functions individually as well as the entire 'lodash' package, it seems auto-scanning security systems were not picking up the issue until just recently. 

There is already a proposed fix [here](https://github.com/ember-cli/ember-cli/pull/10458): 

Hopefully we can get this patch into all ember-cli versions with security support (back to 4.12 right now I think). 

Thanks!

gorner:
edit: Resolves #10459.

[CVE-2021-23337](https://github.com/advisories/GHSA-35jh-r3h4-6jhm) was recently updated to reflect an issue previously patched in Lodash also being present in the `lodash.template` package, which from all appearances is [no longer maintained](https://github.com/lodash/lodash/issues/5738#issuecomment-1737782579). I should emphasize there's nothing I've seen to suggest there's any significant risk in terms of how ember-cli uses this package currently, but it still raises alerts on NPM/GitHub that would be best to clear if possible.

This PR replaces `lodash.template` with the overall `lodash` package – which is already a transitive dependency – and substitutes it in the two places where it was being used. I ran the existing tests for these two modules and they passed.

~~Note that this does **not** yet fully remove `lodash.template` from the dependency tree; it is also a dependency of `broccoli-concat` via `sourcemap-validator` (v1.1.1) which I intend to raise separately shortly.~~ 

**Edit**: On further review I noticed that the current versioning would accept bumping `fast-sourcemap-concat` to v2.1.1 and that doing so would remove the `lodash.template` dependency entirely.

If anything else is needed to ensure this can be back-merged to the LTS versions, please let me know.

### DIFF
--- a/lib/tasks/git-init.js
+++ b/lib/tasks/git-init.js
@@ -47,7 +47,7 @@ module.exports = class GitInitTask extends Task {
   }
 
   async _gitCommit() {
-    const template = require('lodash.template');
+    const template = require('lodash/template');
     let commitTemplate = fs.readFileSync(path.join(__dirname, '../utilities/COMMIT_MESSAGE.txt'));
     let commitMessage = template(commitTemplate)(pkg);
     let env = this.buildGitEnvironment();
--- a/lib/utilities/process-template.js
+++ b/lib/utilities/process-template.js
@@ -6,5 +6,5 @@ module.exports = function processTemplate(content, context) {
     interpolate: /<%=([\s\S]+?)%>/g,
     escape: /<%-([\s\S]+?)%>/g,
   };
-  return require('lodash.template')(content, options)(context);
+  return require('lodash/template')(content, options)(context);
 };
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "isbinaryfile": "^5.0.0",
     "js-yaml": "^4.1.0",
     "leek": "0.0.24",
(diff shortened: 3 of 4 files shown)


LucasHill:
Lodash has confirmed that @gorner's approach is the [recommended resolution](https://github.com/lodash/lodash/issues/5851#issuecomment-2069634388).

If possible, we should get this fix into all the currently supported LTS versions, which I believe is back to 4.12 for security patches.

gorner:
@LucasHill just want to clarify the "recommended resolution" you've linked is a comment I posted, I am not associated with the Lodash project as I said there. I'm not sure where Lodash would have endorsed this as their recommended fix.

LucasHill:
@gorner sorry I completely missed that and just made some assumptions. It was a well written explanation of what is happening and why projects should stop using the individually published functions.

kellyselden:
Could you retarget the `lts-4-12` branch? Then it will get into all LTS's.

gorner:
@kellyselden Done

kellyselden:
Released in 4.12.3, 5.4.2, 5.8.1, and 5.9.0-beta.1. Thanks!

PR #10461, opened 2024-05-09
Remove DT Types

NullVoxPopuli:
ember-data has real types now.

Removing DT types enables downstream blueprints to also remove DT types 
(like the v2 addon blueprint https://github.com/embroider-build/addon-blueprint )

Removing the DT types also gives us a tooon of fixes to the types and broader compatibility with libraries released in the last couple years :tada: :muscle:


Windvis:
@NullVoxPopuli What's the blocker for this one? Does it just need more work or was it discussed somewhere and put on hold?

NullVoxPopuli:
no blocker -- just needs someone to finish

Windvis:
@NullVoxPopuli I'll try to find some time to work on this :+1: 

One thing that concerns me a bit is that the official EmberData types are quite different from the DT ones. If we update the setup in the blueprints then apps will automatically switch to the new setup when they update using ember-cli-update, which can potentially break their setups.

I know it's a blueprint and people are free to choose what changes they accept when updating but I do wonder if we can make this easier to deal with somehow.

I did the DT -> official data types conversion manually in a smallish app and it was still quite a bit of work.

I fear apps might be stuck on older blueprint versions until they switch to the new types manually before updating.

NullVoxPopuli:
The DT types are more than wrong, and i think we shouldn't try to support them <3

They were a stopgap.

Windvis:
(I think this one can be closed now, since #10506 was merged)

PR #10468, opened 2024-06-05
update stable deps

kellyselden:


blueprints/app/files/package.json
```diff
@@ -99,8 +99,8 @@
     "eslint-plugin-prettier": "^5.1.3",
     "eslint-plugin-qunit": "^8.1.1",
     "loader.js": "^4.7.0",
-    "prettier": "^3.2.5",
-    "qunit": "^2.20.1",
+    "prettier": "^3.3.1",
+    "qunit": "^2.21.0",
     "qunit-dom": "^2.0.0",
```
NullVoxPopuli, line 104: can we bump this to v3? qunit-dom@v3 is a non-ember-addon, and better for embroider/vite

tests/fixtures/addon/defaults/package.json
```diff
@@ -64,8 +64,8 @@
     "eslint-plugin-prettier": "^5.1.3",
```
NullVoxPopuli, line 64: eslint-plugin-ember@v11 isn't compatible with gjs/gts, can we bump to 12.1.1?

blueprints/app/files/package.json
kellyselden: ./dev/update-blueprint-dependencies.js explicitly excludes major bumps

PR #10481, opened 2024-07-09
Bump ember-cli-app-version

NullVoxPopuli:
Resolves deprecations that landed in 5.10


mansona (approved):
Do we not have any tests to make sure that a newly generated app doesn't have any deprecations? 🤔

NullVoxPopuli:
> Do we not have any tests to make sure that a newly generated app doesn't have any deprecations? 🤔

nope! 🙈

kategengler:
@mansona No - we could have one here for sure. We do have the tutorial build which will fail on deprecation.

PR #10482, opened 2024-07-09
Bump qunit-dom

NullVoxPopuli:
qunit-dom is now a proper library (not an ember-addon)
(faster build!)


NullVoxPopuli:
oof, this is not helpful:
```
not ok 1512 Acceptance: smoke-test lint fixing after file generation does not fix lint errors with --no-lint-fix
  expected promise to be rejected with an error including 'foo-bar.hbs' but got 'Command failed with exit code 1: embe…'
  AssertionError: expected promise to be rejected with an error including 'foo-bar.hbs' but got 'Command failed with exit code 1: embe…'
```
the actual error is hidden

NullVoxPopuli:
seems like that error is a flake

PR #10487, opened 2024-07-09
feat: add GH action caching for node_modules

scotttball:
This PR adds node_module caching using the GH action backed cache. This should reduce times for builds after the cache has been built. The cache will get destroyed if the yarn.lock changes.

[Caching dependencies to speed up workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

[Github Action - Cache - Yarn](https://github.com/actions/cache/blob/v3/examples.md#node---yarn)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 18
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -26,13 +26,25 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 18
           cache: 'yarn'
 
-      - run: yarn install --frozen-lockfile --non-interactive
+      - name: Fetch yarn cache directory path
```
NullVoxPopuli, line 35: why isn't this handled by setup-node's `cache: yarn`?

scotttball: @NullVoxPopuli it's too costly to build it in for GH as a default option. They cache Node.js version and npm, but not node_modules.

NullVoxPopuli:
This seemed to not provide much (if any) improvement to windows' test performance :-\

PR #10489, opened 2024-07-11
[INTERNAL] Don't run basic tests on Windows

bertdeblock:
The `basic-tests` job runs all basic/faster tests. The `tests` job runs all basic/faster tests, _plus all slow tests_. This separation probably exists to provide quicker feedback for PRs, but it does slow down the entire run, as `basic-tests` on Windows already takes +10 minutes. This change speeds up the entire run, while keeping the same coverage.
This is just a quick fix/win. The entire CI setup could be improved I think.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu, macOS, windows]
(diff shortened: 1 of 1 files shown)


mansona (approved):
:shipit:

void-mAlex (approved):

PR #10490, opened 2024-07-11
[INTERNAL] Restructure CI

bertdeblock:
This will mainly run the basic tests and slow tests in parallel. This reduces CI time to about 25 - 30 minutes. Still slow, but way better than 40 - 45 minutes.

Example runs
- https://github.com/ember-cli/ember-cli/actions/runs/9905624209
- https://github.com/ember-cli/ember-cli/actions/runs/9905086757


kategengler:
Thank you!

PR #10491, opened 2024-07-16
fix: Ignore pnpm-lock.yaml in .prettierignore

tsmalls93:


Windvis:
@tsmalls93 I don't think the prettier setup formats .yaml files at the moment?

PR #10494, opened 2024-07-21
[INTERNAL] Remove `EOLNEWLINE` env var

bertdeblock:
This seemed to be needed for testing on Windows, but AFAICT, not used or needed anymore.

### DIFF
--- a/tests/runner.js
+++ b/tests/runner.js
@@ -15,10 +15,6 @@ chai.use(chaiFiles);
 chai.use(chaiAsPromised);
 chai.use(chaiJestSnapshot);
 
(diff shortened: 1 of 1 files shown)


mansona (approved):
I'm happy that tests pass with this change 👍

kategengler:
This wasn't to accommodate windows users running the tests locally, was it? cc @kellyselden

bertdeblock:
If that's still the case, I could add a comment instead explaining why it's needed.

kategengler:
If @kellyselden hasn't responded, probably safe to remove.

PR #10500, opened 2024-07-28
[INTERNAL] Move `mk-tmp-dir-in` to `tests/helpers`

bertdeblock:
As it's only used in tests, and always nice to ship a few less LOCs.


kategengler:
Thank you!

PR #10505, opened 2024-09-18
use our fork of remove-types with config: false

mansona:
The reason for this PR is because of the questionable default that any time you call transform with babel it looks for a local babel config file 😭 and in the new embroider/vite app blueprint we have an explicit user-owned babel config file that we need to make sure isn't picked up by the parts of the wider ecosystem when they call `transform()` from babel.

The `remove-types` dependency is used in generators and it makes the same mistake of not disabling the babel config lookup which means that generators currently fail in the embroider/vite blueprint: https://github.com/embroider-build/app-blueprint/pull/74

I opened a PR 4 months ago in `remove-types` but there has been no response 🙈 https://github.com/cafreeman/remove-types/pull/3 

I intend to fork remove-types into the ember-cli org to fix this issue but I'm opening this PR now to test the end-to-end and make sure the strategy works 👍 

Edit: I've tested it in https://github.com/embroider-build/app-blueprint/pull/75 and I'm sure this fixes the problem 👍 

Edit: this is now using our [fork of remove-types](https://github.com/ember-cli/babel-remove-types) that has been deployed to npm as [babel-remove-types](https://www.npmjs.com/package/babel-remove-types)

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -529,7 +529,7 @@ let Blueprint = CoreObject.extend({
     @return {Promise}
   */
   async removeTypes(extension, code) {
-    const { removeTypes } = require('remove-types');
+    const { removeTypes } = require('babel-remove-types');
 
     if (extension === '.gts') {
       const { Preprocessor } = require('content-tag');
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "promise-map-series": "^0.3.0",
     "promise.hash.helper": "^1.0.8",
     "quick-temp": "^0.1.8",
-    "remove-types": "^1.0.0",
+    "babel-remove-types": "^1.0.0",
     "resolve": "^1.22.1",
     "resolve-package-path": "^4.0.3",
     "safe-stable-stringify": "^2.4.3",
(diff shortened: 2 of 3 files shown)


Windvis:
Any idea if https://github.com/cyco130/detype might suit our use-case in ember-cli? That seems pretty active at least, and it's credited in the remove-types readme as well. They look fairly similar on first glance at least.

mansona:
So the strategy here is that we need to stick with remove-types because there are versions of ember-cli that have already shipped with it. If we move to a forked version then we can tell people who have issues in older versions of ember-cli to just write an override in their package.json files, but if we switch to another package we can't do that 😞 

The ideal situation would be to get a patch released upstream but I don't think that's going to happen 😞

NullVoxPopuli:
We can use whatever package we want, and have overrides: 
https://pnpm.io/package_json#pnpmoverrides

Screenshot_20240924-133225

Look at that alias!

mansona:
Yes but we can't change from one package to a completely different package with a totally different signature 🤣 that's what I meant. It needs to be a fork of the original with the same API so we can change it out from under ember-cli

Windvis:
> So the strategy here is that we need to stick with remove-types because there are versions of ember-cli that have already shipped with it. If we move to a forked version then we can tell people who have issues in older versions of ember-cli to just write an override in their package.json files, but if we switch to another package we can't do that

I don't understand the reasoning, sorry. If users on an older ember-cli run into this issue, they can still use an override for the patched package, even if we switched to a different package in a newer version, no?

NullVoxPopuli (approved):

PR #10447, opened 2024-03-05
[ENHANCEMENT] Add gjs/gts support for apps and addons

IgnaceMaes:
Adds setting up template tag dependencies for new Ember apps. (dependencies, configs)

Open question:
- Do we want this behind a flag?
  - Any downsides to just _always_ enabling it? This should have no real impact on existing apps, even if the format is not used.
  - Upside of not having a flag is keeping the API surface small, and preventing complicated scenarios. E.g. there's a mix in options depending if `typescript` is enabled or not, making this complexer.

Related: https://github.com/emberjs/tracking-polaris/issues/33

PR #10516, opened 2024-10-10
Use ESLint 9 and Flat Config

NullVoxPopuli:
Notes:
- PR https://github.com/ember-cli/ember-cli/pull/10521 is in this PR -- for sanity.
- includes gjs / gts out of the box (but template-imports not added -- that's for here: https://github.com/ember-cli/ember-cli/pull/10447 )
  - js config copied from https://github.com/embroider-build/app-blueprint
  - ts config copied from https://github.com/embroider-build/app-blueprint/pull/80 
- is otherwise the same, afaict
- Files are split, and does _not_ have a bunch of "if ts else js" conditions, because those have proven annoying to maintain, and something the v2 app blueprint exploration has proven out is that separate folders (js, shared, ts) to be merged later is actually very nice -- see for example: https://github.com/embroider-build/app-blueprint/pull/80
  This _does_ mean there is a bit of duplication, but I think the tradeoff is worth it.
  - To get this to work, I had to thread `options` through a couple methods in the blueprint and then add a `.filter(Boolean)` to filter out "null" (files not applicable based on options).
  - This couldn't use the same technique as the other TS files, because there are two hardcoded, and irrelevant features we have so far:
    - Delete files by name, when typescript
    - Convert JS to TS
    
    And what we want here is "conditionally rename files based on options, deleting the conditionally irrelevant variant" -- if you look in buildFileInfo, you'll see -- the logic is pretty easy


blueprints/app/files/tsconfig.json
```diff
@@ -11,7 +11,7 @@
       "*": ["types/*"]
     },
     "types": [
-      "ember-source/types",
+      "./node_modules/ember-source/types/stable",
```
NullVoxPopuli, line 14: 🤷

bertdeblock:
You can do the file filtering [here](https://github.com/ember-cli/ember-cli/blob/ca8cd79b3d8b56b07576d9dfe60ca417f777ac3a/blueprints/app/index.js#L99-L103), and avoid all the additional code changes? Some of those methods are public API.

blueprints/addon/files/npmignore
```diff
@@ -21,6 +21,7 @@
 /.watchmanconfig
 /CONTRIBUTING.md
 /ember-cli-build.js
+/eslint.config.mjs
```
bertdeblock, line 24: These can go?

```
/.eslintignore
/.eslintrc.js
```

NullVoxPopuli: Yeah, ignore is specified in the config, and the eslint config is named different now

NullVoxPopuli:
> Maybe the same can be achieved using the mapping functionality => 

File map doesn't let you do canditional mapping

blueprints/app/files/_ts_eslint.config.mjs
```diff
@@ -0,0 +1,155 @@
...
+      globals: {
+        ...globals.node,
+      },
+    },
+  },
+  /**
+   * ESM node files
+   */
+  {
+    files: ['*.mjs'],
```
SergeAstapov, line 132: maybe 
```suggestion
    files: ['**/*.mjs'],
```
like it's done on line 107 for .cjs?

blueprints/app/files/package.json
```diff
@@ -77,21 +79,24 @@
...
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-ember": "^12.2.1",
-    "eslint-plugin-n": "^16.6.2",
+    "eslint-plugin-n": "^17.11.1",
     "eslint-plugin-prettier": "^5.2.1",
     "eslint-plugin-qunit": "^8.1.2",
+    "globals": "^15.11.0",
     "loader.js": "^4.7.0",
     "prettier": "^3.3.3",
+    "prettier-plugin-ember-template-tag": "^2.0.2",
```
SergeAstapov, line 91: seem like this is new here, don't you need to configure it in `.prettierrc.js` per https://github.com/gitKrystan/prettier-plugin-ember-template-tag?tab=readme-ov-file#usage?

NullVoxPopuli: Looks like what I have work in neovim, but not in vscode.

ugh. thanks for double checking

NullVoxPopuli: fixed!

blueprints/app/files/_ts_eslint.config.mjs
NullVoxPopuli: We don't know that we want to support more depths of mjs tho?

blueprints/addon/files/npmignore
NullVoxPopuli: done

lib/models/blueprint.js
```diff
@@ -781,9 +781,10 @@ let Blueprint = CoreObject.extend({
     @param {Function} destPath
     @param {Object} templateVariables
     @param {String} file
-    @return {FileInfo}
+    @param {Object} options may be used when buildFileInfo is customized in a blueprint
+    @return {FileInfo | null}
   */
-  buildFileInfo(intoDir, templateVariables, file) {
+  buildFileInfo(intoDir, templateVariables, file /*, options */) {
```
SergeAstapov, line 787: @NullVoxPopuli if this is commented, is this actually needed?

NullVoxPopuli: It's just how I document that an argument is passed to this function.
It's used in blueprints/app/index.js, and passed where buildFileInfo is called in this file

blueprints/app/files/_ts_eslint.config.mjs
SergeAstapov: @NullVoxPopuli fine by me, just was confused why we would support nested .cjs but not nested .mjs

SergeAstapov (approved):

blueprints/app/files/_ts_eslint.config.mjs
NullVoxPopuli: yeah, for me, it's about wanting to get rid of mjs soonish (with type=module)

SergeAstapov: ✅

SergeAstapov (approved):

ef4 (approved):

blueprints/app/files/_js_eslint.config.mjs
```diff
@@ -0,0 +1,119 @@
...
+    ],
+  },
+};
+
+export default [
+  js.configs.recommended,
+  prettier,
+  {
+    ignores: ['dist/', 'node_modules/', 'coverage/', '!**/.*'],
+    linterOptions: {
```
Windvis, line 31: I tried this setup in a v5.12 project and I noticed the global ignores config wasn't working as expected.

It seems that [`ignores` _must_ be the only key in the object](https://eslint.org/docs/latest/use/configure/ignore#ignoring-files) to be considered the "global ignore" config.

We should split the `linterOptions` into a separate object.

NullVoxPopuli: makes sense -- thank you for trying this out!

blueprints/app/files/_js_eslint.config.mjs
```diff
@@ -0,0 +1,119 @@
...
+      globals: {
+        ...globals.browser,
+      },
+    },
+    plugins: {
+      ember,
+    },
+    rules: {
+      ...emberRecommended.rules,
+      ...gjsRecommended.rules,
```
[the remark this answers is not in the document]
Windvis: Resolved by https://github.com/ember-cli/ember-cli/pull/10533

PR #10536, opened 2024-10-25
[ENHANCEMENT] Make deprecations throw when the `until` for `ember-cli` has passed

MehulKChaudhari:
Throw deprecations if the until of ember-cli has passed

For - https://rfcs.emberjs.com/id/0830-evolving-embers-major-version-process


lib/debug/deprecate.js
```diff
@@ -34,6 +35,14 @@ const assert = require('./assert');
  * - `url` A URL that refers to additional information about the deprecation
  */
 function deprecate(description, condition, options) {
+  const isDeprecationRemoved = semver.gt(emberCLIVersion, options.until);
```
kategengler, line 38: I think this should be `gte` -- the until is the version at which it breaks

kategengler: I also think this needs to check the `options.for` is `ember-cli` -- this method is public and can be used by addons

kategengler (changes requested):

tests/unit/debug/deprecate-test.js
```diff
@@ -182,4 +182,38 @@ ID     foo
 UNTIL  5.0.0
 URL    https://example.com`);
   });
+
+
+  it('throws an deprecation if the current Ember-CLI version is greater than the until version of deprecation', function () {
+    expect(() => {
+      deprecate('The `foo` method is deprecated.', false, {
```
kategengler, line 189: The period is in the message here but also in the concatenated message in the implementation of `deprecate`

MehulKChaudhari: Thanks, Yes removed it

lib/debug/deprecate.js
```diff
@@ -34,6 +35,16 @@ const assert = require('./assert');
  * - `url` A URL that refers to additional information about the deprecation
  */
 function deprecate(description, condition, options) {
+  if (options && options.for) {
```
bertdeblock, line 38: This should be moved under:

```js
if (condition) {
  return;
}
```

Because we want to assert all options first, that way you don't need the if here, and we want to take the `condition` value into account I think.

kategengler:
This is in-progress version of https://github.com/emberjs/ember.js/pull/20669 for ember-cli

.github/workflows/ci.yml
```diff
@@ -112,3 +112,15 @@ jobs:
       - run: pnpm test:all
         env:
           'EMBER_CLI_${{ matrix.feature-flag }}': true
+
+  deprecations-broken-test:
+    name: Debug and Prebuilt with Deprecations as Errors
```
kategengler, line 117: This name is copy and pasted from ember-source?

.github/workflows/ci.yml
```diff
@@ -112,3 +112,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [basic-tests, linting]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-environment
+      - name: Run Tests with Deprecations as Errors
+        env:
+          OVERRIDE_DEPRECATION_VERSION: '5.0.0'
+        run: pnpm test
```
kategengler, line 126: I think this should be `pnpm test:all`

lib/debug/deprecate.js
kategengler: I agree that the check should move below the assert calls in the lines below.

kategengler (changes requested):

.github/workflows/ci.yml
MehulKChaudhari: Yes, Isn't it correct one?

.github/workflows/ci.yml
MehulKChaudhari: Yes, changed it

.github/workflows/ci.yml
MehulKChaudhari: Changes the name. Thank you for pointing it out

lib/debug/deprecate.js
kategengler: This should also move below the `if (condition)` as Bert says -- the condition is whether or not the deprecation is triggered, so we wouldn't want to throw on ones that are not triggered.

MehulKChaudhari: yes, made the changes

lib/debug/deprecate.js
```diff
@@ -34,6 +35,8 @@ const assert = require('./assert');
  * - `url` A URL that refers to additional information about the deprecation
  */
 function deprecate(description, condition, options) {
+  const currentEmbeCLIVersion = process.env.OVERRIDE_DEPRECATION_VERSION ?? emberCLIVersion;
```
bertdeblock, line 38:
```suggestion
  const currentEmberCLIVersion = process.env.OVERRIDE_DEPRECATION_VERSION ?? emberCLIVersion;
```

I would also declare this above `isDeprecationRemoved`.

MehulKChaudhari: Ok, I put it above the `isDeprecationRemoved` declaration.

kategengler:
I am suspicious of the tests all passing. I would expect, at least, most tests that had to be modified in https://github.com/ember-cli/ember-cli/pull/10538/ to fail with the deprecations broken.

.github/workflows/ci.yml
```diff
@@ -112,3 +112,15 @@ jobs:
...
+  deprecations-broken-test:
+    name: Deprecations "broken"
+    runs-on: ubuntu-latest
+    needs: [basic-tests, linting]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-environment
+      - name: Run Tests with Deprecations as Errors
+        env:
+          OVERRIDE_DEPRECATION_VERSION: '5.0.0'
```
kategengler, line 125:
```suggestion
          OVERRIDE_DEPRECATION_VERSION: '22.0.0'
```

We want to break all possible deprecations for the test run

kategengler:
@bertdeblock caught that the override deprecation version was too low to cause deprecations to break

tests/acceptance/init-test.js
```diff
@@ -196,6 +198,11 @@ describe('Acceptance: ember init', function () {
   });
 
   it('configurable CI option', async function () {
+
+    if (isDeprecationRemoved(travisDeprecation.until)){
```
kategengler, line 202: @bertdeblock Do you have thoughts on the best way to do the test infrastructure for deprecations that have since "broken"?

NullVoxPopuli (approved):

PR #10554, opened 2024-11-04
[ENHANCEMENT] Lazy load commands and tasks

bertdeblock:


lib/commands/index.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+module.exports = {
```
bertdeblock, line 3: Aliases are needed to look up commands, so this doesn't do anything.

PR #10555, opened 2024-11-09
Bump pnpm/action-setup to v4 in app/addon blueprints

SergeAstapov:
Just noticed those are not using latest version

### DIFF
--- a/blueprints/addon/files/.github/workflows/ci.yml
+++ b/blueprints/addon/files/.github/workflows/ci.yml
@@ -19,9 +19,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3<% if (pnpm) { %>
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
         with:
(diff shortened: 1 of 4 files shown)


NullVoxPopuli (approved):

blueprints/addon/files/.github/workflows/ci.yml
```diff
@@ -19,9 +19,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3<% if (pnpm) { %>
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
```
NullVoxPopuli, line 22: is this the one that reads `packageManager`?

If so, maybe next we set packageManager and remove with version?

SergeAstapov: @NullVoxPopuli yeah, fine by me as a follow up - so we do one thing in one PR

bertdeblock (approved):

PR #10580, opened 2024-12-02
[ENHANCEMENT] Update LTS versions in blueprints

bertdeblock:


bertdeblock:
https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/additional-package.json#L14
Do we want to bump this as well now?

void-mAlex (approved):

kategengler:
Thank you!

PR #10583, opened 2024-12-17
[ENHANCEMENT] Update `app` blueprint to support `ember-qunit` v9

ef4:
ember-qunit 9.0.0 is published, these are the corresponding blueprint changes.

The motivation here is:
 - ember-qunit currently takes responsibility for doing AMD-based loading of tests. That is not a thing on new-enough embroider apps.
 - in the v2 app blueprint we currently use `start({ loadTests: false })`, and that is enough to not get a runtime error, but it still pulls a v1 addon (ember-cli-test-loader) into your app, which causes you to need the whole v1 addon compatibility system just to use ember-qunit, even though ember-qunit is a v2 addon.

So the design is to make `import { start } from 'ember-qunit'` really only responsible for starting tests, and separately make `import { loadTests } from 'ember-qunit/test-loader'` responsible for the classical AMD loading behavior. Classic apps will us `loadTests` and newer apps simply will not import or call it.

In addition, we realized that this change presents an upgrade hazard since users might upgrade ember-qunit without touching their blueprint and get a green but empty test suite. QUnit has built-in protection for this situation, but ember-qunit was defeating that protection by always adding the `Ember.onerror validation` test by default, thus giving you a "not empty" test suite. So this change also removes that default behavior from `start` in favor of making it explicit. 

`Ember.onerror` is already slated for deprecation in https://github.com/emberjs/rfcs/pull/1003, so removing test coverage for that feature from the default `start` behavior seems like a positive step. It means people who clear the deprecation can also drop the corresponding test.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,7 +10,7 @@ importers:
     dependencies:
       '@pnpm/find-workspace-dir':
         specifier: ^7.0.2
-        version: 7.0.2
+        version: 7.0.3
       babel-remove-types:
         specifier: ^1.0.0
         version: 1.0.0
@@ -64,13 +64,13 @@ importers:
         version: 4.1.2
       ci-info:
         specifier: ^4.0.0
-        version: 4.0.0
+        version: 4.1.0
       clean-base-url:
         specifier: ^1.0.0
         version: 1.0.0
       compression:
         specifier: ^1.7.4
-        version: 1.7.4
+        version: 1.7.5
       configstore:
         specifier: ^5.0.1
         version: 5.0.1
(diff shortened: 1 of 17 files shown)


kategengler (approved):

pnpm-lock.yaml
```diff
@@ -1,6105 +1,1412 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
```
bertdeblock, line 1: Woops 😬

NullVoxPopuli: if only there were some way to declare which `packageManager` a project uses :thinking: 

haha

NullVoxPopuli: https://github.com/ember-cli/ember-cli/pull/10584

ef4:
This needs fixture updates, which will be easier with https://github.com/ember-cli/ember-cli/pull/10585

NullVoxPopuli:
@ef4 can you rebase -- I added packageManager -- so now, if you use proto, things should happen to work accidentally in the correct way

PR #10585, opened 2024-12-18
[INTERNAL] Support `WRITE_FIXTURES` in more test files

ef4:
The fixture-based tests support a WRITE_FILES=true environment variable when you want to update them, which is really the only pleasasnt way to maintain them. But this flag doesn't apply to files that were being checked via `checkFileWithEmberCLIVersionReplacement`.

This refactors those tests to use more targeted replacement and makes `WRITE_FIXTURES` apply to them.


bertdeblock (approved):
Just need to revert the lockfile changes I think.

PR #10592, opened 2024-12-25
[BUGFIX] Fix ESLint config for v1 addons

bertdeblock:
See #10586.

### DIFF
--- a/tests/fixtures/app/eslint.config.mjs
+++ b/tests/fixtures/app/eslint.config.mjs
@@ -81,8 +81,10 @@ export default [
     files: [
       '**/*.cjs',
       'config/**/*.js',
+      'tests/dummy/config/**/*.js',
       'testem.js',
       'testem*.js',
+      'index.js',
       '.prettierrc.js',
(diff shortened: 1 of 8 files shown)


NullVoxPopuli (approved):

tests/fixtures/app/eslint.config.mjs
```diff
@@ -81,8 +81,10 @@ export default [
     files: [
       '**/*.cjs',
       'config/**/*.js',
+      'tests/dummy/config/**/*.js',
```
SergeAstapov, line 84: is there simple way to exclude this for actual apps?

NullVoxPopuli: We'd need to copy the files to the addon blueprint, rather than continuing to use the app's files

SergeAstapov: thanks for explanation @bertdeblock and @NullVoxPopuli !

SergeAstapov (approved):

PR #10596, opened 2024-12-28
[ENHANCEMENT] Vanilla Prettier setup in blueprints [RFC 1055]

bertdeblock:


tests/acceptance/smoke-test-slow.js
```diff
@@ -86,15 +86,6 @@ describe('Acceptance: smoke-test', function () {
     expect(output).to.include('***CUSTOM_TESTEM_JS**');
   });
 
-  it('eslint passes after running ember new', async function () {
```
bertdeblock, line 89: Removed, because https://github.com/ember-cli/ember-cli/pull/10586 covers everything (ESLint, Stylelint, TypeScript, ...)

blueprints/app/files/package.json
```diff
@@ -12,10 +12,12 @@
   },
   "scripts": {
     "build": "ember build --environment=production",
+    "format": "prettier . --cache --write",
     "lint": "concurrently \"<%= packageManager %>:lint:*(!fix)\" --names \"lint:\" --prefixColors auto",
     "lint:css": "stylelint \"**/*.css\"",
     "lint:css:fix": "concurrently \"<%= packageManager %>:lint:css -- --fix\"",
     "lint:fix": "concurrently \"<%= packageManager %>:lint:*:fix\" --names \"fix:\" --prefixColors auto",
```
bertdeblock, line 19:
```suggestion
    "lint:fix": "concurrently \"<%= packageManager %>:lint:*:fix\" --names \"fix:\" --prefixColors auto && <%= invokeScriptPrefix %> format",
```

blueprints/app/files/tests/index.html
```diff
@@ -1,22 +1,25 @@
...
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
     <title><%= namespace %> Tests</title>
-    <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="description" content="" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
 
+    <!-- prettier-ignore -->
```
bertdeblock, line 9: Prevents:

```
{{content-for "head"}} {{content-for "test-head"}}
```

blueprints/app/files/.prettierignore
```diff
@@ -5,7 +5,9 @@
 /dist/
 
 # misc
+/config/ember-cli-update.json
```
bertdeblock, line 8: Both additions are generated files.

ef4 (changes requested):
Suggested alternative: can we put `**/*.html` in the prettierignore?

Because the formatting prettier is producing here is not great. You were forced to add the special prettier-ignore comment line. And prettier is making several tags self-closing that are not actually supposed to be self-closing in the HTML spec.

The current blueprint doesn't do prettier formatting for html, so this is not a loss of existing functionality.

Separate question: is this introducing prettier formatting to markdown, and if so does it also rewrite people's READMEs? If so, that would possibly be another extension to ignore.

NullVoxPopuli:
hbs / gjs / etc are not explicitly HTML tho -- and this goes for jsx, vue, svelte, ng / angular, it's all shorthand to (eventually) generate HTML.

the HTML file is the only real HTML we have.

mansona:
apparently there is a plugin that fixes the issue? https://github.com/prettier/prettier/issues/15336#issuecomment-2509738565

bertdeblock:
I'm only saying that formatting-wise the same issue occurs in hbs, gjs, ...

bertdeblock:
Indeed, but I didn't really want to include yet another dependency in the blueprint. I would prefer to just follow Prettier. To me, that's kind of the point of using Prettier. Let it decide what's best readability/formatting wise.

mansona:
I agree with the point of prettier (I may not like it but I agree with what it's trying to do)... but in this case it's just wrong 🤔 Void elements are not and should not be "self closing"

mansona:
https://html.spec.whatwg.org/multipage/syntax.html#start-tags

> Then, if the element is one of the [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), or if the element is a [foreign element](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements), then there may be a single U+002F SOLIDUS character (/), which on [foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) marks the start tag as self-closing. On [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), it does not mark the start tag as self-closing but instead is unnecessary and has no effect of any kind. For such void elements, it should be used only with caution — especially since, if directly preceded by an [unquoted attribute value](https://html.spec.whatwg.org/multipage/syntax.html#unquoted), it becomes part of the attribute value rather than being discarded by the parser.

SergeAstapov:
if another vote matters, I agree with @mansona that Prettier is surely wrong and ignorant for that issue (the reason is most likely because of React and JSX, like this reddit thread suggest https://www.reddit.com/r/reactjs/comments/jc8964/self_closing_tags_in_react/)

Hence, my vote goes to doing right thing and either ignore .html files (but *do* format .hbs, .gts, .gjs, etc.) or use Prettier plugin (only if it works perfectly).

bertdeblock:
fwiw, I'm not arguing that Prettier is doing the right thing. I agree that it's not according to the spec. I just feel that's a discussion for the Prettier project.

SergeAstapov:
maybe we should do RFC amendment to clarify this in RFC rather than in this thread? just a thought

bertdeblock:
Yeah, maybe it's safer to ignore html files for now then. Most projects only have 2 anyways.

ef4:
> I'm now slightly worried that my RFC wasn't clear enough.
>The intention was to format all files Prettier supports, including (but not limited to) *.html and *.md files.

Looking at the RFC I think your intent is clear, but speaking for myself the thing I was paying attention to in the RFC is stopping running prettier inside eslint. That is, IMO, the most important motivation. Having the *option* of running prettier on more file types is only one of the benefits of that decision, and not nearly the most important one.

(Having prettier inside eslint is miserable. I've seen way too many developers reacting to red squiggles from the prettier eslint plugin, which is defeats the point of having prettier, which should free you to *stop* paying attention to formatting).

bertdeblock:
HTML changes are reverted, and `*.html` is added to the Prettier ignore file.
Ready for another review I think!

blueprints/app/files/_js_eslint.config.mjs
```diff
@@ -16,7 +16,7 @@ import globals from 'globals';
 import js from '@eslint/js';
 
 import ember from 'eslint-plugin-ember/recommended';
-import prettier from 'eslint-plugin-prettier/recommended';
+import eslintConfigPrettier from 'eslint-config-prettier';
```
SergeAstapov, line 19: @bertdeblock why would we still need this if prettier runs outside of ESLint?

SergeAstapov: ah, right, so it's to disable any stylistic related rules in ESLint to avoid conflicts

SergeAstapov (approved):

kategengler (approved):

ef4 (approved):

PR #10598, opened 2024-12-29
[INTERNAL] Remove "locked versions for pre-v1.0.0 packages" tests

bertdeblock:
The purpose of these tests is to ensure that pre-v1.0.0 packages in ember-cli's `package.json` file,
and the `app` blueprint's `package.json` file, are defined using a fixed version (e.g. `0.5.0` instead of `^0.5.0`).
The PR that initially introduced these tests, did not seem to include any motivation as to why this was done.

Also, these tests have been broken since 2014, by [this commit](https://github.com/ember-cli/ember-cli/commit/780473fccf5293da6dec4707707f4c472aba3243).
ember-cli's `package.json` file contains various pre-v1.0.0 packages, that are not defined using a fixed version:

```
"broccoli-debug": "^0.6.5",
"exit": "^0.1.2",
"heimdalljs": "^0.2.6",
"heimdalljs-logger": "^0.1.10",
"markdown-it-terminal": "^0.4.0",
"promise-map-series": "^0.3.0",
"quick-temp": "^0.1.8",
"ember-cli-blueprint-test-helpers": "^0.19.2",
"ember-cli-internal-test-helpers": "^0.9.1",
"tmp": "^0.2.1",
```

Maybe defining pre-v1.0.0 packages using a fixed version was a thing back then?

AFAICT, it's okay to use `^` for these packages, and we can simply remove these broken tests.


kategengler (approved):

mansona:
so... if these have been broken for years how come the CI hasn't been failing? I don't see a skip test in there 🤔 I am just asking out of curiosity btw

bertdeblock:
@mansona They just kept passing by accident, because the assertion was defined in an `if`.
See this [commit](https://github.com/ember-cli/ember-cli/commit/780473fccf5293da6dec4707707f4c472aba3243). It introduces `semver.valid(deps[name])` to the `if` check.

```ts
// before
semver.gtr('1.0.0', '^0.6.5') // true === execute the assertion, and it will fail because '^0.6.5' is a range and not a valid SemVer / fixed version. Has to be defined as `0.6.5`.

// after
semver.valid('^0.6.5') && semver.gtr('1.0.0', '^0.6.5') // false === will not execute the assertion, because the valid / fixed check is pulled inside the `if`.
```

So suddenly, all pre-v1.0.0 packages using a range, were not asserted anymore since that commit.

I don't know if I explained it very well, it's a bit of a weird situation.

mansona:
wow that's a great find 😂 thanks for the explanation 🎉

PR #10599, opened 2025-01-01
[INTERNAL] Bump `content-tag` to v3

SergeAstapov:
Updates `content-tag` to v3.

Changelog https://github.com/embroider-build/content-tag/releases/tag/v3.0.0-content-tag

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "compression": "^1.7.4",
     "configstore": "^5.0.1",
     "console-ui": "^3.1.2",
-    "content-tag": "^2.0.1",
+    "content-tag": "^3.1.0",
     "core-object": "^3.1.5",
     "dag-map": "^2.0.2",
     "diff": "^5.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -75,8 +75,8 @@ importers:
         specifier: ^3.1.2
         version: 3.1.2
       content-tag:
-        specifier: ^2.0.1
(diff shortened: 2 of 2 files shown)


bertdeblock:
@SergeAstapov Thanks! I think you also need to take [this PR](https://github.com/embroider-build/content-tag/pull/86) into account. The return type changed.

SergeAstapov:
@bertdeblock yeap, was thinking it's one line change but appears it's not - will debug why it's failing

lib/models/blueprint.js
```diff
@@ -537,7 +537,7 @@ let Blueprint = CoreObject.extend({
       // Strip template tags
       const templateTagIdentifier = (index) => `template = __TEMPLATE_TAG_${index}__;`;
       const templateTagIdentifierBraces = (index) => `(template = __TEMPLATE_TAG_${index}__);`;
-      const templateTagMatches = preprocessor.parse(code);
+      const { code: templateTagMatches } = preprocessor.parse(code);
```
bertdeblock, line 540: We're using `parse` here, not `process`, so I think this change is not needed.

SergeAstapov: @bertdeblock thank you @bertdeblock for pointing out! turns out I was not paying an attention

bertdeblock (approved):

lib/models/blueprint.js
bertdeblock: No worries, I didn't notice it as well :smile:

Issue #10611, opened 2025-01-07
6.1, ember-data TS config is broken in monorepo when using npm or yarn@1


NullVoxPopuli:
Even more minimal repro:
- https://github.com/NullVoxPopuli/ts-types-package-json-exports-proof-for-ember-data
  This confirms my beliefs from when I originally reported this

Fix here: https://github.com/emberjs/data/pull/9633

--------------------

Repro steps:
- have monorepo with either npm or yarn@1
- create 6.1 app in some sub-folder with `--typescript`
- run lints
- see errors

I believe this happens because the tsconfig.json specifies ./node_modules/* paths for types, and that's an unsafe thing to do.

ember-data deps are not installed to the test-app's node_modules:

![image](https://github.com/user-attachments/assets/9f3de2ff-93da-4fe5-b0ce-0bc1c3c84b64)

Error output

```
[lint:css]
[lint:css] > test-app@0.0.0 lint:css
[lint:css] > stylelint "**/*.css"
[lint:css]
[lint:js]
[lint:js] > test-app@0.0.0 lint:js
[lint:js] > eslint . --cache
[lint:js]
[lint:hbs]
[lint:hbs] > test-app@0.0.0 lint:hbs
[lint:hbs] > ember-template-lint .
[lint:hbs]
[lint:types]
[lint:types] > test-app@0.0.0 lint:types
[lint:types] > tsc --noEmit
[lint:types]
[lint:hbs]
[lint:hbs] npm run lint:hbs exited with code 0
[lint:css] npm run lint:css exited with code 0
[lint:types] error TS2688: Cannot find type definition file for './node_modules/@ember-data/adapter/unstable-preview-types'.
[lint:types]   The file is in the program because:
[lint:types]     Entry point of type library './node_modules/@ember-data/adapter/unstable-preview-types' specified in compilerOptions
[lint:types] error TS2688: Cannot find type definition file for './node_modules/@ember-data/graph/unstable-preview-types'.
[lint:types]   The file is in the program because:
[lint:types]     Entry point of type library './node_modules/@ember-data/graph/unstable-preview-types' specified in compilerOptions
[lint:types] error TS2688: Cannot find type definition file for './node_modules/@ember-data/json-api/unstable-preview-types'.
[lint:types]   The file is in the program because:
[lint:types]     Entry point of type library './node_modules/@ember-data/json-api/unstable-preview-types' specified in compilerOptions
[...]

PR #10613, opened 2025-01-09
Add new flag for `new`: `--ember-data` (default), and `--no-ember-data`

NullVoxPopuli:
> [!IMPORTANT]  
> ember-cli 6.1+ are broken for all monorepo projects (due to ember-data tsconfig.json paths)

----------------

This is, in part, because I want to be unblocked on the v2 addon blueprint:
- https://github.com/embroider-build/addon-blueprint/pull/320
 blocked by: 
  - https://github.com/ember-cli/ember-cli/issues/10611
    - https://github.com/ember-cli/ember-cli/pull/10612
    - Related: 
      - ~~https://github.com/emberjs/data/issues/9630~~
      - https://github.com/emberjs/data/pull/9633

And also in part, because I haven't used fetch or any remote data in a test-app for any of my addons, and ember-data is always the first thing I kick out (along with ember-fetch, and a few others, but those will have other PRs)

There is precedence for this in existing code here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/index.js#L74

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -7,6 +7,9 @@ const chaiJestSnapshot = require('chai-jest-snapshot');
 module.exports = {
   timeout: 5000,
   reporter,
+  reporterOption: {
+    maxDiffSize: Infinity,
+  },
   retries: 2,
   rootHooks: {
     beforeEach() {
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -30,7 +30,7 @@
   "devDependencies": {
     "@babel/core": "^7.26.0<% if (!typescript) { %>",
     "@babel/eslint-parser": "^7.25.9",
-    "@babel/plugin-proposal-decorators": "^7.25.9<% } %><% if (typescript) { %>",
+    "@babel/plugin-proposal-decorators": "^7.25.9<% } %><% if (typescript && emberData) { %>",
     "@ember-data/adapter": "~5.4.0-beta.12",
     "@ember-data/graph": "~5.4.0-beta.12",
     "@ember-data/json-api": "~5.4.0-beta.12",
@@ -68,8 +68,8 @@
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-inject-live-reload": "^2.1.0<% if (!embroider) { %>",
     "ember-cli-sri": "^2.1.1",
-    "ember-cli-terser": "^4.0.2<% } %>",
-    "ember-data": "~5.4.0-beta.12",
+    "ember-cli-terser": "^4.0.2<% } %>",<% if (emberData) { %>
+    "ember-data": "~5.4.0-beta.12",<% } %>
     "ember-fetch": "^8.1.2",
     "ember-load-initializers": "^3.0.1",
     "ember-modifier": "^4.2.0",
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -68,8 +68,8 @@
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-inject-live-reload": "^2.1.0<% if (!embroider) { %>",
     "ember-cli-sri": "^2.1.1",
-    "ember-cli-terser": "^4.0.2<% } %>",<% if (emberData) { %>
-    "ember-data": "~5.4.0-beta.12",<% } %>
+    "ember-cli-terser": "^4.0.2<% } %><% if (emberData) { %>",
+    "ember-data": "~5.4.0-beta.12<% } %>",
     "ember-fetch": "^8.1.2",
     "ember-load-initializers": "^3.0.1",
     "ember-modifier": "^4.2.0",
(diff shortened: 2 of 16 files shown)


blueprints/app/files/tsconfig.json
```diff
@@ -26,7 +26,8 @@
       "./node_modules/@ember-data/model/unstable-preview-types",
       "./node_modules/@ember-data/serializer/unstable-preview-types",
       "./node_modules/@ember-data/tracking/unstable-preview-types",
-      "./node_modules/@warp-drive/core-types/unstable-preview-types"
+      "./node_modules/@warp-drive/core-types/unstable-preview-types"<% } else { %>
+      "ember-source/types"<% } %>
```
NullVoxPopuli, line 30: this also fixes a bug where in npm and yarn, none of the types work, since it's possible for there to be no local node_modules

NullVoxPopuli: see: https://github.com/ember-cli/ember-cli/pull/10612

noting, however, that ember-data doesn't support this, because the `types` exports are currently stripped on publish.
Which is something I'm aiming to address here:
- https://github.com/emberjs/data/pull/9631

.mocharc.js
```diff
@@ -7,6 +7,9 @@ const chaiJestSnapshot = require('chai-jest-snapshot');
 module.exports = {
   timeout: 5000,
   reporter,
+  reporterOption: {
+    maxDiffSize: Infinity,
```
NullVoxPopuli, line 11: this tells mocha to print diffs for our help files / objects

blueprints/app/files/package.json
```diff
@@ -66,8 +66,8 @@
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-inject-live-reload": "^2.1.0<% if (!embroider) { %>",
     "ember-cli-sri": "^2.1.1",
-    "ember-cli-terser": "^4.0.2<% } %>",
-    "ember-data": "~5.4.0-beta.12",
+    "ember-cli-terser": "^4.0.2<% } %>",<% if (emberData) { %>
```
bertdeblock, line 69: Template tags must be defined inside the quotes, so the auto-update script still works.
I made https://github.com/ember-cli/ember-cli/pull/10616 to help with this.

tests/fixtures/help/generate.txt
```diff
@@ -1,21 +1,44 @@
+Requested ember-cli commands:
```
bertdeblock, line 1: This diff seems odd.

blueprints/app/files/package.json
NullVoxPopuli: fixed!

tests/fixtures/help/generate.txt
NullVoxPopuli: yea, but tests pass, so either:
- this is unused
- tests don't care about this change
- it's been broken?
- a mystery!

bertdeblock (commented):
It would also be nice if we can exclude `app/models/.gitkeep` from the generated files I think.

NullVoxPopuli:
> It would also be nice if we can exclude app/models/.gitkeep from the generated files I think.

aye -- arguably -- ember-data stuff should be its own blueprint layered on top, I think. would make all the little details easier to manage

ef4 (approved):

lib/commands/new.js
```diff
@@ -43,6 +43,12 @@ module.exports = Command.extend({
       type: ['github', 'none'],
       description: 'Installs the optional default CI blueprint. Only Github Actions is supported at the moment.',
     },
+    {
```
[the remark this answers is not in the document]
NullVoxPopuli: I'd prefer this to be a followup <3

lib/tasks/interactive-new.js
```diff
@@ -121,6 +121,22 @@ class InteractiveNewTask extends Task {
           },
         ],
       },
+      {
+        name: 'emberData',
+        type: 'confirm',
+        message: 'Do you want to include ember-data?',
+        when: undefined === newCommandOptions.emberData,
+        choices: [
```
bertdeblock, line 129: These had to go.

lib/tasks/interactive-new.js
```diff
@@ -121,6 +121,22 @@ class InteractiveNewTask extends Task {
           },
         ],
       },
+      {
+        name: 'emberData',
+        type: 'confirm',
+        message: 'Do you want to include ember-data?',
+        when: undefined === newCommandOptions.emberData,
```
bertdeblock, line 128: This will always be `false` since the `ember-data` flag has a default value.

NullVoxPopuli: https://github.com/ember-cli/ember-cli/pull/10622

lib/tasks/interactive-new.js
NullVoxPopuli: https://github.com/ember-cli/ember-cli/pull/10622

PR #10617, opened 2025-01-17
[ENHANCEMENT] Use `staticInvokables` in the `app` blueprint

bertdeblock:
Resolves:

```
Setting 'staticComponents' is deprecated. Use 'staticInvokables' instead.
Setting 'staticHelpers' is deprecated. Use 'staticInvokables' instead.
Setting 'staticModifiers' is deprecated. Use 'staticInvokables' instead.
```


mansona (approved):
nice one!

PR #10619, opened 2025-01-24
[BUGFIX] Only remove type imports when removing the types from `.gts` files in blueprints

Windvis:
When implementing https://github.com/emberjs/ember.js/pull/20835#discussion_r1927321171 I ran into an issue when the types are removed from the .gts file. I tried reproducing the issue in a test.

It looks like the imports that are only used inside the template context are removed incorrectly. It also seems to generate an `export {}` sometimes, ~~but I'm not 100% sure what causes that~~ which also seems related to the removal of the imports (when there are no other imports or exports: https://www.typescriptlang.org/docs/handbook/2/modules.html#non-modules).

Edit: The underlying issue was fixed in babel-remove-types and released in v1.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@pnpm/find-workspace-dir": "^7.0.2",
-    "babel-remove-types": "^1.0.0",
+    "babel-remove-types": "^1.0.1",
     "broccoli": "^3.5.2",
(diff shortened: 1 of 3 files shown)


bertdeblock (approved):

Windvis:
(The current CI failures seem unrelated to the changes here)

PR #10620, opened 2025-01-28
[BUGFIX] Handle errors gracefully when proxy target is down

dwickern:
If there is an error connecting to the proxy target, proxied http requests hang indefinitely. Now the proxy server will respond with a 502.

### DIFF
--- a/lib/tasks/server/middleware/proxy-server/index.js
+++ b/lib/tasks/server/middleware/proxy-server/index.js
@@ -26,9 +26,11 @@ class ProxyServerAddon {
         timeout: options.proxyInTimeout,
       });
 
-      proxy.on('error', (e) => {
+      proxy.on('error', (e, req, res) => {
         options.ui.writeLine(`Error proxying to ${options.proxy}`);
         options.ui.writeError(e);
+        res.status(502);
+        res.end();
       });
 
       const morgan = require('morgan');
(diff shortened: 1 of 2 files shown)


bertdeblock (approved):

lib/tasks/server/middleware/proxy-server/index.js
```diff
@@ -26,9 +26,13 @@ class ProxyServerAddon {
         timeout: options.proxyInTimeout,
       });
 
-      proxy.on('error', (e) => {
+      proxy.on('error', (e, req, res) => {
         options.ui.writeLine(`Error proxying to ${options.proxy}`);
         options.ui.writeError(e);
+        if (typeof res.status === 'function') {
```
dwickern, line 32: `res` can be a http response or websocket. node-http-proxy emits an error for both:

https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/web-incoming.js#L165

https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/ws-incoming.js#L157

node-http-proxy cleans up websockets automatically. We need to explicitly end http requests or they will hang indefinitely in the browser.

ef4:
Unsure about the CI failure here, I think it's occurring on multiple PRs and needs to be investigated.

PR #10633, opened 2025-02-03
[BUGFIX release] Update all EmberData deps to stable

Windvis:
It seems the automation script can't handle the new `@ember-data/*` packages yet, so they weren't converted to stable versions before the release.

This would result it a mix of stable and beta deps in the generated package.json file when the `--typescript` flag is used.

Edit: It seems the issue is EmberData is no longer guaranteed to create a new stable release from the beta release we use in the blueprints, before ember-cli releases. As a result, the beta versions aren't switched to a stable version since the `latest-version` package doesn't "downgrade" a minor version: `~5.4.0-beta.12` wouldn't go back to the `~5.3.*` stable since that would violate the semver version.

The ember-data package would have the same problem, but the semver version is overruled and replaced by the `stable` npm tag in that flow.


ef4:
This looks good, we just need to resolve conflicts.

PR #10642, opened 2025-02-18
Only include warp-drive/core-types if emberData is true

evoactivity:
This should not be installed if `--no-ember-data` is passed.

I've nested the if emberData within the typescript if, not sure if that works or not.

This package is currently breaking new installs.


evoactivity:
The failures here I guess I would expect since this package is breaking making a new app.

Windvis:
Strange that there are no fixture tests for the `--no-ember-data` flag :sweat_smile: .

Windvis:
The version issue is fixed on master now, so rebasing should fix CI :+1:

evoactivity:
@Windvis Okay, I guess I can close this PR and we just land your PR with tests?

PR #10641, opened 2025-02-17
[Bugfix] Fix missing lint dependency for typescript projects

NullVoxPopuli:
uh, I'm so bad at managing fixtures in this project.
something is different about my machine (too modern? lol) that won't make local match CI

PR #10646, opened 2025-02-24
[Bugfix release]: Fix eslint parser for js when using --typescript

NullVoxPopuli:
backport of https://github.com/ember-cli/ember-cli/pull/10641

### DIFF
--- a/blueprints/app/files/package.json
+++ b/blueprints/app/files/package.json
@@ -28,7 +28,7 @@
   "devDependencies": {
     "@babel/core": "^7.26.7",
     "@babel/eslint-parser": "^7.26.5",
-    "@babel/plugin-proposal-decorators": "^7.25.9<% if (typescript && emberData) { %>",
+    "@babel/plugin-proposal-decorators": "^7.25.9<% if (typescript) { %>",
     "@ember-data/adapter": "~5.3.10",
     "@ember-data/graph": "~5.3.10",
     "@ember-data/json-api": "~5.3.10",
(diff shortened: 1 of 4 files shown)


blueprints/app/files/package.json
```diff
@@ -26,9 +26,9 @@
     "test:ember": "ember test"
   },
   "devDependencies": {
-    "@babel/core": "^7.26.7<% if (!typescript) { %>",
+    "@babel/core": "^7.26.7",
     "@babel/eslint-parser": "^7.26.5",
-    "@babel/plugin-proposal-decorators": "^7.25.9<% } %><% if (typescript) { %>",
+    "@babel/plugin-proposal-decorators": "^7.25.9<% if (typescript && emberData) { %>",
```
Windvis, line 31: The emberData check isn't needed on 6.2 and seems to be the reason for the CI faillure.
```suggestion
    "@babel/plugin-proposal-decorators": "^7.25.9<% if (typescript) { %>",
```

Windvis:
Ok, CI now has some other errors:

1. Something seems to have bumped minimatch to v10, which is not compatible with node 18:

>    error minimatch@10.0.1: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.20.5"
  error Found incompatible module.

2. Some [smoke tests](https://github.com/ember-cli/ember-cli/actions/runs/13636016509/job/38115010974?pr=10646) are failing with a `Error: Cannot find module 'typescript'` error

I think both of these are caused by eslint-plugin-n. They added `ts-declaration-location` as a dependency which depends on both minimatch v10 and typescript (as a peerDep).

This issue tracks the first problem: https://github.com/eslint-community/eslint-plugin-n/issues/414

Windvis:
@NullVoxPopuli They reverted the issue in 17.16.2 so I think this is unblocked now if you restart CI.

PR #10659, opened 2025-03-01
[BUGFIX] Improve detyping `.gts` files

bertdeblock:
Fixes https://github.com/emberjs/ember.js/pull/20862#issuecomment-2692161233.

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -535,14 +535,18 @@ let Blueprint = CoreObject.extend({
       const { Preprocessor } = require('content-tag');
       const preprocessor = new Preprocessor();
       // Strip template tags
-      const templateTagIdentifier = (index) => `template = __TEMPLATE_TAG_${index}__;`;
-      const templateTagIdentifierBraces = (index) => `(template = __TEMPLATE_TAG_${index}__);`;
+      const replacementClassMember = (i) => `template = __TEMPLATE_TAG_${i}__;`;
+      const replacementExpression = (i) => `__TEMPLATE_TAG_${i}__`;
       const templateTagMatches = preprocessor.parse(code);
       let strippedCode = code;
       for (let i = 0; i < templateTagMatches.length; i++) {
         const match = templateTagMatches[i];
         const templateTag = substringBytes(code, match.range.start, match.range.end);
-        strippedCode = strippedCode.replace(templateTag, templateTagIdentifier(i));
+        if (match.type === 'class-member') {
+          strippedCode = strippedCode.replace(templateTag, replacementClassMember(i));
+        } else {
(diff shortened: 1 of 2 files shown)


lib/models/blueprint.js
```diff
@@ -553,8 +557,14 @@ let Blueprint = CoreObject.extend({
       for (let i = 0; i < templateTagMatches.length; i++) {
         const match = templateTagMatches[i];
         const templateTag = substringBytes(code, match.range.start, match.range.end);
-        transformedWithTemplateTag = transformedWithTemplateTag.replace(templateTagIdentifier(i), templateTag);
-        transformedWithTemplateTag = transformedWithTemplateTag.replace(templateTagIdentifierBraces(i), templateTag);
+        if (match.type === 'class-member') {
+          transformedWithTemplateTag = transformedWithTemplateTag.replace(replacementClassMember(i), templateTag);
+        } else {
+          // babel-remove-types uses prettier under the hood, and adds trailing `;` where allowed,
```
bertdeblock, line 563: Seems we can also disable this, so that might be the better option:

```ts
// Remove types
const transformed = await removeTypes(strippedCode, false);
```

`false` === disable prettier.

bertdeblock: Comes with other issues, so ignoring this for now.

lifeart (approved):

lib/models/blueprint.js
```diff
@@ -535,14 +535,18 @@ let Blueprint = CoreObject.extend({
...
       // Strip template tags
-      const templateTagIdentifier = (index) => `template = __TEMPLATE_TAG_${index}__;`;
-      const templateTagIdentifierBraces = (index) => `(template = __TEMPLATE_TAG_${index}__);`;
+      const replacementClassMember = (i) => `template = __TEMPLATE_TAG_${i}__;`;
+      const replacementExpression = (i) => `__TEMPLATE_TAG_${i}__`;
       const templateTagMatches = preprocessor.parse(code);
       let strippedCode = code;
       for (let i = 0; i < templateTagMatches.length; i++) {
         const match = templateTagMatches[i];
         const templateTag = substringBytes(code, match.range.start, match.range.end);
```
NullVoxPopuli, line 544: you may also be interested in content-tag-utils, which does the byte math for you: https://github.com/nullVoxPopuli/content-tag-utils

NullVoxPopuli (approved):

Windvis (approved):
I applied the same changes as a patch in the ember PR, and it fixed the issue I was running into. Thanks! :pray:

Windvis:
What's the release strategy for this one? It blocks the PR in ember-source, since we need to list the ember-cli as a devDependency there, but ember-cli releases _after_ ember-source, so we can't merge it without a patch I think? Unless this is backported to a previous release.

PR #10663, opened 2025-03-10
Update the test script so it only runs the ember tests

Windvis:


NullVoxPopuli (approved):
yaaas, finally, no hiding lintssues behind test. <3

tests/fixtures/addon/defaults/package.json
```diff
@@ -25,8 +25,7 @@
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
     "start": "ember serve",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\" --prefixColors auto",
-    "test:ember": "ember test",
+    "test": "ember test",
```
kategengler, line 28: Fairly sure what `test` does was define in an RFC. I'll look for it.

Windvis: > Fairly sure what test does was define in an RFC. I'll look for it.

Maybe when we switched from running the linters as part of `ember test` into separate scripts?

> Also may not be a great choice for addons because they often also have test:node or similar

Addons also have a [`test:ember-compatibility`](https://github.com/ember-cli/ember-cli/blob/c18b9e55fbf5f928f4812c5447e438d61489da85/blueprints/addon/additional-package.json#L6) script so I would be very surprised if anyone is actually running the test script in CI since that would be quite heavy.

NullVoxPopuli: what was the RFC? I think if that can be  found, we can quick RFC to undo it, because it was a mistake. haha

Windvis: I think it's this one: https://github.com/emberjs/rfcs/blob/master/text/0121-remove-ember-cli-eslint.md#detailed-design

But I can't find the corresponding RFC PR to see the discussions there.

Windvis: It's not very well motivated in my opinion. I assume they simply wanted to match what the test script did before ember-cli-eslint was removed?

NullVoxPopuli: yea, feels silly.

do you want to whip up a counter RFC?

kategengler: That's not the one. I'll look when I get a few (soon).

I remember the motivating argument was that things run `npm test` by default and not wanting failing lints to to let default CI's pass.

NullVoxPopuli: ya -- I think we need to argue that we can't encapsulate all of CI into a single command

NullVoxPopuli:
@Windvis we could keep the PR until an RFC is accepted <3

PR #10667, opened 2025-03-22
Add a failing test for the TO gts + signature scenario

Windvis:
While working on the [strict component blueprint support PR](https://github.com/emberjs/ember.js/pull/20862), I encountered this issue which _seems_ like a bug in the type removing functionality of ember-cli.

This adds a failing test which makes it easier to debug.

It seems to be related to the "implicit `export default`" feature of .gts. If I change the code to this then the output no longer includes the `export {}` line.

Input:
```diff
import type { TOC } from '@ember/component/template-only';

export interface FooSignature {
  // The arguments accepted by the component
  Args: {};
  // Any blocks yielded by the component
  Blocks: {
    default: []
  };
  // The element to which \`...attributes\` is applied in the component template
  Element: null;
}

+export default <template>
-<template>
  {{yield}}
</template> satisfies TOC<FooSignature>;
```

Output:
```diff
+export default <template>
-<template>
  {{yield}}
</template>
```
I also tried updating content-tag to v3.1.2, since that includes a fix very similar to this issue, but it doesn't solve this one it seems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "compression": "^1.8.0",
     "configstore": "^5.0.1",
     "console-ui": "^3.1.2",
-    "content-tag": "^3.1.1",
+    "content-tag": "^3.1.2",
     "core-object": "^3.1.5",
     "dag-map": "^2.0.2",
     "diff": "^7.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -75,8 +75,8 @@ importers:
         specifier: ^3.1.2
         version: 3.1.2
       content-tag:
-        specifier: ^3.1.1
-        version: 3.1.1
+        specifier: ^3.1.2
+        version: 3.1.2
       core-object:
         specifier: ^3.1.5
         version: 3.1.5
@@ -1644,8 +1644,8 @@ packages:
     resolution: {integrity: sha512-FveZTNuGw04cxlAiWbzi6zTAL/lhehaWbTtgluJh4/E95DqMwTmha3KZN1aAWA8cFIhHzMZUvLevkw5Rqk+tSQ==}
     engines: {node: '>= 0.6'}
 
-  content-tag@3.1.1:
-    resolution: {integrity: sha512-94puwVk6X8oJcbRIEY03UM80zWzA3dYgGkOiRJzeY1vXgwrFUh3OolDDi/D7YBa6Vsx+CgAvuk4uXlB8loZ1FA==}
+  content-tag@3.1.2:
+    resolution: {integrity: sha512-Z+MGhZfnFFKzYC+pUTWXnoDYhfiXP9ojZe3JbwsYufmDuoeq2EvuDyeFAJ/RnKokUwz5s9bQhDOrbvSYRShcrQ==}
 
   content-type@1.0.5:
     resolution: {integrity: sha512-nTjqfcBFEipKdXCv4YDQWCfmcLZKm81ldF0pAopTvyrFGVbcR6P/VAAd5G7N+0tTr8QqiU0tFadD6FK4NtJwOA==}
@@ -6954,7 +6954,7 @@ snapshots:
     dependencies:
       safe-buffer: 5.2.1
 
-  content-tag@3.1.1: {}
+  content-tag@3.1.2: {}
 
   content-type@1.0.5: {}
 


NullVoxPopuli:
Can you try upgrading babel-plugin-ember-template-compilation?

Windvis:
> Can you try upgrading babel-plugin-ember-template-compilation?

AFAIK, `babel-plugin-ember-template-compilation` isn't used by the type removal feature in ember-cli. `pnpm why babel-plugin-ember-template-compilation` doesn't show it either.

Windvis:
It's babel-remove-types that adds the `export {}`;

This is the code that's passed into it, after the template tag is replaced with a placeholder. 

```ts
import type { TOC } from '@ember/component/template-only';

export interface FooSignature {
  // The arguments accepted by the component\n  Args: {};
  // Any blocks yielded by the component\n  Blocks: {
    default: []
  };
  // The element to which `...attributes` is applied in the component template
  Element: null;
}

__TEMPLATE_TAG_0__ satisfies TOC<FooSignature>;

```

Which results in this code:

```js
__TEMPLATE_TAG_0__;
export {};

```

So _I think_ we need to add more information to content-tag's .parse return value, about which content tag depends on the implicit default export, so we can add more logic here to deal with that scenario? We can then prepend it with export default before passing it over to `babel-remove-types`. Maybe babel-remove-types has an option to not output this `export {}` line though.

NullVoxPopuli:
tl;dr: content-tag needs to be at least 3.1.2, else you hit the missing `export default` bug

--------------


 content-tag emits the `export default` for plain templates, but not ones wrapped in a satisfies node:
 
 https://stackblitz.com/edit/stackblitz-starters-4fpuuz1c?file=index.js,package.json

```ts
import { template as template_ee404ce5d1a14c47b1acf1c491050df9 } from "@ember/template-compiler";
import type { TOC } from '@ember/component/template-only';
template_ee404ce5d1a14c47b1acf1c491050df9(`
  
  `, {
    eval () {
        return eval(arguments[0]);
    }
}) satisfies TOC<{
}>;
```

vs
```ts
import { template as template_245e215fd1a84f21bb99d45f1734555c } from "@ember/template-compiler";
export default template_245e215fd1a84f21bb99d45f1734555c(`
  
  `, {
    eval () {
        return eval(arguments[0]);
    }
});
```

_BUT!_

If I upgrade content-tag to 3.1.2, it`s fixed:
```ts
import { template as template_fd9b2463e5f141cfb5666b64daa1f11a } from "@ember/template-compiler";
import type { TOC } from '@ember/component/template-only';
export default template_fd9b2463e5f141cfb5666b64daa1f11a(`
  
  `, {
    eval () {
        return eval(arguments[0]);
    }
}) satisfies TOC<{
}>;
```

Windvis:
> tl;dr: content-tag needs to be at least 3.1.2, else you hit the missing export default bug

This PR literally includes the 3.1.2 version bump and it didn't fix the issue :sweat_smile: . 

3.1.2 fixed it when transforming .gts code, but here we are using the `.parse` method to then do something with the result and pass it to babel-remove-types. So we need to handle the implicit `export default` here as well, before handig it off to babel-remove-types.

NullVoxPopuli:
> This PR literally includes the 3.1.2 version bump and it didn't fix the issue 😅 .

apologies. I got too excited :facepalm:

NullVoxPopuli:
Gonna see if content-tag-utils handles this

NullVoxPopuli:
So, it seems that because content-tag-utils operates on the tag range, there is no issue: https://github.com/NullVoxPopuli/content-tag-utils/pull/13/files#diff-04471e09412bbfa48026ce0732d9ce4efd6dea7514e80516d583cbe2a11d0233R44

NullVoxPopuli:
> So we need to handle the implicit export default here as well, before handig it off to babel-remove-types.

I don't think this is true -- and `parse` can't provide this information, because it's the node _within_ the satisfies type node.

I think in order to remove the satisfies type node, `removeTypes` or the stripping code needs to match on the `satisfies` (outer) node first, since it wraps the template tag node?

Windvis:
> Where does this happen?

Here: https://github.com/ember-cli/ember-cli/blob/487ae48c517597aad20970f1471b6883f8b8fa7c/lib/models/blueprint.js#L539-L576

Windvis:
> Maybe content-tag-utils should have an unprocess method that converts plain js/ts back to gjs/gts.
> 
> That way strip-babel-types can operate on plain js, and we don't need parse

If it's possible that would indeed be ideal, but I'm not sure if it is :smile: (if we want to keep the same formatting, implicit export defaults, ...). Still, Worth looking into I guess, this does indeed feel like something that shouldn't be present in ember-cli.

Semi-off-topic but I am starting to wonder if the whole "remove types from blueprint files" was the way to go though. I think things would be a _lot_ simpler if we simply maintained a couple of extra .js blueprint files and didn't have to deal with these kind of issues at all..

NullVoxPopuli:
> unprocess

here it is, a way to go from js/ts back to gjs/gts

https://github.com/NullVoxPopuli/content-tag-utils/pull/15

NullVoxPopuli:
So now the code would be:
```js
let ts = p.process(gts);
let js = stripTypes(ts);
let gjs = unprocess(js);
```

PR #10670, opened 2025-03-28
testing: don't generate temp directories in the repo

mansona:
I noticed this when I was trying to do a new feature that relies on resolving that if we generate our temp folders **inside** the checked out repo then we will have things being accidentally resolvable 🙈 

This will also make test rely on system TEMP which I think is usually MemFS nowadays 🤔 so that would make things quicker 🎉


tests/acceptance/new-test-slow.js
```diff
@@ -1,23 +1,24 @@
...
-    await tmp.setup(tmpDir);
-    chdir(tmpDir);
+    const { path } = await tmp.dir();
+    tmpDir = path;
+    process.chdir(path);
   });
 
-  afterEach(async function () {
-    await tmp.teardown(tmpDir);
+  this.afterEach(function () {
```
kategengler, line 20: typo?

tests/acceptance/new-test.js
```diff
@@ -35,11 +36,11 @@ describe('Acceptance: ember new', function () {
     } else {
       process.env.CI = ORIGINAL_PROCESS_ENV_CI;
     }
-    return tmp.teardown(tmpDir);
+    process.chdir(root);
   });
 
   function confirmBlueprintedForDir(blueprintDir, expectedAppDir = 'foo', typescript = false) {
-    let blueprintPath = path.join(root, blueprintDir, 'files');
+    let blueprintPath = path.join(blueprintDir, 'files');
```
kategengler, line 43: All the below assertions appear to be dynamic, so changing the test makes me nervous. Could you add an assertion below that expected isn't an empty array?

tests/helpers/mk-tmp-dir-in.js
```diff
@@ -1,14 +0,0 @@
-'use strict';
-
-const fs = require('fs-extra');
-const temp = require('temp');
```
kategengler, line 4: Is `temp` still used elsewhere?

tests/acceptance/new-test-slow.js
mansona: I think maybe a strange auto-complete or something 🤔

tests/acceptance/new-test.js
mansona: sure 👍

tests/helpers/mk-tmp-dir-in.js
mansona: yes temp is actually a dependency and is used in the `ember test` command (for some reason 🤷)

tests/acceptance/new-test-slow.js
mansona: Fixed now

PR #10671, opened 2025-03-28
Implement the built in app, addon, and blueprint blueprints by package lookup

mansona:
This is mostly to open up a discussion of the implementation `lib/models/blueprint.js`.

Personally I think it would be better for maintainability if we just extract the blueprints into their own repos (not even a monorepo with each blueprint).


packages/app-blueprint/files/package.json
```diff
@@ -60,7 +60,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^9.1.2",
     "ember-auto-import": "^2.10.0",
-    "ember-cli": "~<%= emberCLIVersion %>",
+    "ember-cli": "~6.4.0-beta.1",
```
kategengler, line 63: So this would need to be bumped along with other packages?

mansona: yes exactly, I would need to add that to the script thanks for the reminder

packages/app-blueprint/files/config/ember-cli-update.json
```diff
@@ -3,7 +3,7 @@
   "packages": [
     {
       "name": "ember-cli",
-      "version": "<%= emberCLIVersion %>",
```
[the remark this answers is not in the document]
mansona: so the infrastructure of this hasn't actually changed at all, ember-cli is still in control of the generation of the blueprint even though you could do it directly with the new package

packages/app-blueprint/files/package.json
mansona: in fact this will automatically be bumped by release-plan, I have configured it correctly now to match the main ember-cli version 👍

packages/addon-blueprint/index.js
```diff
@@ -175,10 +179,8 @@ module.exports = {
   },
 
   beforeInstall() {
-    const version = require('../../package.json').version;
-    const prependEmoji = require('../../lib/utilities/prepend-emoji');
+    const prependEmoji = require('@ember/blueprint-model/utilities/prepend-emoji');
```
kategengler, line 183: Should this write the blueprint version so it is obvious what they are getting?

tests/integration/models/blueprint-test.js
```diff
@@ -181,12 +180,14 @@ describe('Blueprint', function () {
...
 
     it('finds blueprints in the ember-cli package', function () {
-      let expectedPath = path.resolve(defaultBlueprints, 'app');
+      let expectedPath = path.dirname(require.resolve('@ember/classic-app-blueprint'));
       let expectedClass = Blueprint;
 
       let blueprint = Blueprint.lookup('app');
 
-      expect(blueprint.name).to.equal('app');
+      // TODO it's really strange that a blueprint's name is defined by its folder
```
kategengler, line 188: I have vague memories that it can be overridden -- I think there are examples in ember-source?

packages/app-blueprint/index.js
```diff
@@ -119,10 +119,8 @@ module.exports = {
   },
 
   beforeInstall() {
-    const version = require('../../package.json').version;
-    const prependEmoji = require('../../lib/utilities/prepend-emoji');
+    const prependEmoji = require('@ember/blueprint-model/utilities/prepend-emoji');
 
-    this.ui.writeLine(chalk.blue(`Ember CLI v${version}`));
```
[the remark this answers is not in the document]
mansona: ah yes, I removed this because I hadn't solved the versioning yet 👍 I can put something else in here for sure

kategengler (approved):

PR #10681, opened 2025-04-02
Enable declaration generation for addons

SergeAstapov:
Ref https://github.com/tsconfig/bases/pull/303, as this change breaks declaration generation for v1 addons


NullVoxPopuli:
Alt: https://github.com/ember-cli/ember-cli/pull/10682

SergeAstapov:
for next release, https://github.com/ember-cli/ember-cli/pull/10682 makes sense. I guess this change can be release as patch if we'd like

NullVoxPopuli:
the primary backstory is that we haven't had control over patch vs major when releasing --- it sems the tsconfig/bases repo doesn't do semver in any way. By moving to our own packages, we can do semver properly

SergeAstapov:
@NullVoxPopuli sorry for nudging, is there a chance to check if tsconfig/bases does have an option to follow SemVer in some way?

NullVoxPopuli:
why's that? on some PRs to tsconfig/bases, we've said what type of impact a change should have, and those comments have been ignored (either by accident or for some other reason, idk)

SergeAstapov:
thank you for explanation @NullVoxPopuli ! makes sense. then this PR is not needed, should we go with https://github.com/ember-cli/ember-cli/pull/10682

PR #10686, opened 2025-04-28
don't generate temp directories in the repo

mansona:
This is an extraction from #10671 that I noticed during the monorepo work, if we're trying to verify things correctly it might be better for us to not generate temp folders inside the repo but to use system temp instead


mansona:
looks like I opened this already 🙈

PR #10704, opened 2025-05-26
Update init.js,   no 'cd...' proposition needed when ending the user interaction

jjffb4a:
in opposition to ember new, we are already in the targeted directory

remove lines 134+135


NullVoxPopuli:
> , we are already in the targeted directory

this is not true.

When you generate a new project,
```bash
# ~/somedir
npx ember-cli new my-app
```
you are still in `~/somedir`

in order for us to run dependency install, we need to cd in to `~/somedir/my-app`.
And the same is true for you, the user, if you want to run the dev server.

Because your terminal is in `~/somedir`, you also need to cd in to `~/somedir/my-app`

PR #10708, opened 2025-06-02
Simplified eslint configs

ijlee2:


blueprints/app/files/_ts_eslint.config.mjs
```diff
@@ -38,7 +38,6 @@ export default tseslint.config(
   eslint.configs.recommended,
   eslintPluginEmber.configs.base,
   eslintPluginEmber.configs.gjs,
-  eslintPluginEmber.configs.gts,
```
ijlee2: Redundant (declared below), if not, possibly incorrect as it may affect `*.{js,gjs}` if defined at a global level.

blueprints/app/files/_js_eslint.config.mjs
```diff
@@ -13,14 +13,14 @@
...
-import ember from 'eslint-plugin-ember/recommended';
+import eslintPluginEmber from 'eslint-plugin-ember/recommended';
 import eslintConfigPrettier from 'eslint-config-prettier';
-import qunit from 'eslint-plugin-qunit';
-import n from 'eslint-plugin-n';
+import eslintPluginQunit from 'eslint-plugin-qunit';
+import eslintPluginN from 'eslint-plugin-n';
 
-import babelParser from '@babel/eslint-parser';
+import babelEslintParser from '@babel/eslint-parser';
```
ijlee2: Unfortunately, the plugins don't follow a uniform naming convention when we look at all their `README`'s. The solution that I've come up with for [my configs](https://github.com/ijlee2/frontend-configs/blob/2.0.0/packages/eslint/ember/v1-app/index.mjs#L1-L12) was to camelize the package name.

NullVoxPopuli:
This is fairly new (as of this weekend), but we now have labels for everything so debugging is easier:

image
image



These particular screenshots are using https://github.com/NullVoxPopuli/ember-eslint, which is a more encapsulated version of the official lints -- allow consumers to delete a ton of deps from their package.json

PR #10717, opened 2025-06-11
pass --publish-branch to release-plan publish

mansona:
This is needed because release-plan (and maybe pnpm) has checks that confirms if you want to continue publishing if you're not on main or master. When publishing on a branch that is not main or master you need to pass a `--publish-branch`


mansona:
I'm cancelling the main CI run on this because it's a publish-only change 👍

PR #10724, opened 2025-06-12
Update RELEASE.md to fully document new release-plan setup

mansona:
I was tracking the written release process while putting together #10721 and I noticed a few issues, so I updated it as I went along 👍


mansona:
I'm merging this in spite of the flakey test failure since it's only changing documentation 👍

PR #10735, opened 2025-06-27
remove unnecessary project override

mansona:
This simplifies the interaction between the Project and the Blueprint models. Now a blueprint only uses the project that is passed in 👍

### DIFF
--- a/lib/models/blueprint.js
+++ b/lib/models/blueprint.js
@@ -1250,23 +1250,6 @@ let Blueprint = CoreObject.extend({
 
     let previousCwd;
     if (!this.project.isEmberCLIProject()) {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

kategengler:
This looks to have been added because of the crossed out #3 here https://github.com/ember-cli/ember-cli/pull/9972

If the tests pass, either it is no longer an issue, or we no longer have that scenario described in the PR.

mansona:
We've discussed this in Discord and apparently is to support the use of the `addAddonToProject` API during `ember new` or `ember addon`

The conclusion is that it's unlikely that anyone is using this and our own app and addon blueprints haven't used this for a while so we can safely remove it 👍

PR #10748, opened 2025-07-26
update blueprint version of ember-cli as part of release plan

mansona:
This is to fix an issue that was hidden and only became apparent **after** the last release. We have enabled CI on release-plan PRs to make sure we catch this issue before it becomes a problem. You can see the CI is failing here: https://github.com/ember-cli/ember-cli/pull/10746

This PR adds a step to update the ember-cli version in the app blueprint during release-plan planning. After this is merged release-plan PRs will be green again and CI won't be broken directly after release any more 👍

### DIFF
--- a/.github/workflows/plan-release.yml
+++ b/.github/workflows/plan-release.yml
@@ -80,6 +80,14 @@ jobs:
         env:
           GITHUB_AUTH: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

mansona:
I'm merging this over CI because a) it's flakey and b) this can literally have no effect on CI until after release-plan has run, and it can't run until we merge this.

It's unfortunate but we need to "test this in production"

PR #10777, opened 2025-08-21
Fix instrumentation tests by properly resetting fs.statSync

pichfl:
Fixes the following error that popped up with `node@24.6`:

```
 1) models/instrumentation.js
       ._enableFSMonitorIfInstrumentationEnabled
         "before each" hook for "if instrumentation is enabled, monitor":

      AssertionError: expected [Function] to equal [Function statSync]
      + expected - actual

      -{
      -  "__restore": [Function]
      -}
      +[Function]

      at Context.<anonymous> (tests/unit/models/instrumentation-test.js:38:30)
      at process.processImmediate (node:internal/timers:505:21)
```

### DIFF
--- a/tests/unit/models/instrumentation-test.js
+++ b/tests/unit/models/instrumentation-test.js
@@ -40,6 +40,10 @@ describe('models/instrumentation.js', function () {
 
     afterEach(function () {
       td.reset();
+      // Manually reset statSync as node@24.6 returns [Function: statSync],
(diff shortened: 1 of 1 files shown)


pichfl:
Scratch this, `node@24.6` breaks [heimdall-fs-monitor](https://github.com/heimdalljs/heimdalljs/tree/master/packages/%20heimdall-fs-monitor)

PR #10783, opened 2025-08-27
remove unused changelog script

mansona:
now that we're using release-plan and we're mostly ok with the auto-generated changelogs we don't need our own custom script any more 👍


evoactivity:
Test failures don't seem related.

mansona:
> Test failures don't seem related.

story of my life at the moment 😭 there is some strange code in the Project class that has a very strange test that is just odd. I updated the odd stuff 🤷

PR #10790, opened 2025-09-01
Reorganize tests for `new` and `addon` commands

pichfl:
- no functional changes
- logical groups will make it easier to disable specific block in upcoming experiments
- replace code comments with visible describe groups

I recommend going commit by commit for this one. I tried to keep the steps small.


mansona (approved):
I checked this carefully to make sure that code wasn't lost during the move 👍

PR #10791, opened 2025-09-01
update the format of the ember-cli-update.json

mansona:
Fundamentally this PR fixes a problem that has been surfaced by https://github.com/ember-cli/ember-cli/pull/10766

We can now have a difference in versions between `ember-cli` and the two app and addon blueprint packages. This means that expecting the app and addon blueprints to have a version the same as ember-cli is no longer valid.

This also sets us up to correctly update people across the vite boundary to the "old blueprints" without to much trouble 👍

### DIFF
--- a/packages/app-blueprint/files/config/ember-cli-update.json
+++ b/packages/app-blueprint/files/config/ember-cli-update.json
@@ -2,13 +2,11 @@
   "schemaVersion": "1.0.0",
   "packages": [
     {
-      "name": "ember-cli",
+      "name": "<%= blueprint %>",
       "version": "<%= blueprintVersion %>",
       "blueprints": [
         {
           "name": "<%= blueprint %>",
-          "outputRepo": "https://github.com/ember-cli/ember-<%= blueprint === 'app' ? 'new' : 'addon' %>-output",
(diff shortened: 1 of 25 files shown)


NullVoxPopuli (approved):

packages/app-blueprint/files/config/ember-cli-update.json
```diff
@@ -2,13 +2,11 @@
...
   "packages": [
     {
-      "name": "ember-cli",
+      "name": "<%= blueprint %>",
       "version": "<%= blueprintVersion %>",
       "blueprints": [
         {
           "name": "<%= blueprint %>",
-          "outputRepo": "https://github.com/ember-cli/ember-<%= blueprint === 'app' ? 'new' : 'addon' %>-output",
-          "codemodsSource": "ember-<%= blueprint %>-codemods-manifest@1",
```
kategengler, line 11: We probably do want a codemodsSource? This removes it for classic as well and so I don't believe ember-cli-update's run codemods will work without this. 

They are kept here in a monorepo https://github.com/ember-cli/ember-cli-update-codemods-manifest/

PR #10798, opened 2025-09-06
Add import from ember-data breakage/deprecation

NullVoxPopuli:
This should get the tutorial's nightly CI passing


kategengler:
@runspired Is this something that should be in the blueprint now?

runspired:
Yes

kategengler (approved):

kategengler:
Just to note this won't fix the tutorial until we publish

mansona:
shouldn't this be backported to all versions that need this? or is it only the latest version of ember-data/warpdrive where this is the fix?

kategengler:
As I understand it, only ember-data 5.7 has the fix for the deprecation that this resolves

PR #10806, opened 2025-09-10
skip build watch tests when vite is enabled

mansona:
This is needed for https://github.com/ember-cli/ember-cli/pull/10802


NullVoxPopuli (approved):
What about the compatBuild watch mode?

mansona:
> What about the compatBuild watch mode?

This is specifically testing the whole outside build with watch mode and the only thing that was failing was some clean up task on windows so it was a problem with the vite build 😭

PR #10826, opened 2025-10-03
move resolution of @ember/app-blueprint to prevent loading latest

mansona:
this makes sure that we change what the `app` blueprint means at a point where we can resolve a true path and return a real blueprint module. This means that we're now resolving the blueprint in the context of ember-cli's dependencies and not always getting latest

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -12,8 +12,6 @@ const { isPnpmProject, isYarnProject } = require('../utilities/package-managers'
 const getLangArg = require('../../lib/utilities/get-lang-arg');
 const { deprecate, DEPRECATIONS } = require('../debug');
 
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

lib/commands/init.js
```diff
@@ -126,10 +124,6 @@ module.exports = Command.extend({
     let blueprintOpts = clone(commandOptions);
     let blueprint = normalizeBlueprint(blueprintOpts.blueprint || this._defaultBlueprint());
 
-    if (isExperimentEnabled('VITE') && blueprint === 'app') {
```
kategengler, line 129: Is this change in pattern an issue for ember-cli-update using the blueprint from config?

PR #10843, opened 2025-10-11
Further contextualize help output and error when options and commands that are no longer supported in Vite-based projects are used

kategengler:


kategengler:
It looks like `--output-path` or `-o` is used by compat prebuild with vite projects? (I'm guessing based on failing tests) @mansona?

PR #10846, opened 2025-10-13
[beta bugfix] allow build --watch only in EMBROIDER_PREBUILD

mansona:
In https://github.com/ember-cli/ember-cli/pull/10843 we added an error for anyone calling `ember build --watch` because it doesn't work in an embroider world. We missed the fact that Embroider itself needs to be able to call it 🙈 

This PR fixes that and makes sure that there are no extra warnings shown to the end-user 👍

### DIFF
--- a/lib/commands/build.js
+++ b/lib/commands/build.js
@@ -38,9 +38,10 @@ module.exports = Command.extend({
   },
 
   async run(commandOptions) {
-    if (this.isViteProject) {
-      // --watch is used during Embroider prebuild but should never be used directly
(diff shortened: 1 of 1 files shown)


kategengler:
I'm 👍 but wish there were tests

mansona:
@kategengler your instinct was correct, the fix was actually in the wrong place 🙈 whenever you call `--watch` then `--watcher` is implicitely set with a value of `watchman` so I needed to move the check out a level. I pushed the test before the fix to show the problem 👍

PR #10857, opened 2025-10-15
Update RELEASE.md with more notes

mansona:
I just want to make it clear that we don't want to update ember-data in the beta phase 👍


mansona:
I cancelled CI because it's a markdown only change

PR #10870, opened 2025-11-11
Upgrade broccoli

kategengler:


NullVoxPopuli (approved):

bertdeblock (approved):

alechirsch:
Curious when this might get released, as it fixes a few security vulerabilities

kategengler:
Unless we backport it, it will go out in 6.10

If you use only Node >= 20.19, you can use overrides to use broccoli ^4

PR #10876, opened 2025-12-05
bump minimum node version to 20.19

mansona:
The thing that seems to be blocking me on https://github.com/ember-cli/ember-cli/pull/10873 is that warp-drive now **requires** you to be on a version of node that supports require(esm). This means that we have a new minimum minor version of node that we can support with the current dependencies as defined in the blueprints. 

I don't know if we need to update anything in our node support documentation for this 🤔

### DIFF
--- a/packages/app-blueprint/files/package.json
+++ b/packages/app-blueprint/files/package.json
@@ -99,7 +99,7 @@
     "webpack": "^5.101.3"
   },
   "engines": {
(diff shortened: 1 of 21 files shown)


kategengler:
This should be fine, since our minimum Node was already 20. The new version of broccoli also requires this minimum node version.

PR #10877, opened 2025-12-06
fix: ensure --lang option requires a value

AyushKashyapII:
### Description
The `new` command previously allowed the `--lang` flag to be passed without a value (e.g., `ember new my-app --lang`), resulting in the project being created with an empty language string.

This PR adds a check to ensure that if `--lang` is present, it must have a value. If not, it rejects with a `SilentError`.

### Reproduction
1. Run `ember new my-app --lang`
2. Observe that the app is created successfully (unexpected behavior).

### Solution
Added a validation check in `lib/commands/new.js` and a corresponding unit test in `tests/unit/commands/new-test.js`.


AyushKashyapII:
Hi @kategengler ,

I hope you are having a good week!

I just wanted to gently follow up on this PR regarding the fix for the `--lang` flag validation. I have verified that all CI checks are passing and the tests cover the reported issue.

I know this is a busy time of year, but whenever you have a moment, I would appreciate a review. Please let me know if there are any other changes required!

Thanks!

AyushKashyapII:
@kategengler r u available to check this

NullVoxPopuli (changes requested):
existing behavior, depeending on argument order:

```
❯ pnpm dlx ember-cli new my-lang-app --lang --pnpm
 WARN  7 deprecated subdependencies found: glob@10.5.0, glob@7.2.3, inflight@1.0.6, raw-body@1.1.7, rimraf@2.7.1, rimraf@3.0.2, source-map-url@0.3.0
Packages: +605
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 605, reused 590, downloaded 15, added 605, done
WARNING: An issue with the `--lang` flag returned the following message:
  Detected a `--lang` specification starting with command flag `-`.
  This issue is likely caused by using the `--lang` flag without a specification.`
WARNING: The human language of the application will NOT be set in
  the `<html>` element's `lang` attribute in `index.html`.
If this was not your intention, you may edit the `<html>` element's
  `lang` attribute in `index.html` manually after the process is complete.
Information about using the `--lang` flag:
  The `--lang` flag sets the base human language of an app or test app:
    - `app/index.html` (app)
    - `tests/dummy/app/index.html` (addon test app)
  If used, the lang option must specfify a valid language code.
  For default behavior, remove the flag.
  See `ember <command> help` for more information.
```

however, when the args are flipped, we get this:
```
❯ pnpm dlx ember-cli new my-lang-app2 --pnpm --lang 
installing app-blueprint
```

so, the fix here should make the lang print the same or similar message as above, I think.

Just eexpanding on the message, rather tahn just "requires a value"

AyushKashyapII:
Thank you for the detailed feedback on the argument order! You were exactly right—the behavior was inconsistent because nopt assigns true to the flag if it's at the end of the command, but assigns the next flag (e.g., --pnpm) as the value if it’s placed in the middle.
I have updated the logic in lib/commands/new.js to handle both cases:
1. Validation: It now rejects the command if --lang is:
- Empty ('')
- At the end of the command (Boolean true)
- Followed by another flag (value starts with -).
2. Consistency: I’ve expanded the SilentError message to include the detailed usage information you suggested. This ensures that regardless of the argument order, the user receives the same helpful information and the process 'fails fast' before generating a project with an incorrect lang attribute.

I also updated the unit tests to verify that the CLI rejects these invalid inputs with the expanded help text.

NullVoxPopuli:
What does the terminal print in each case?

AyushKashyapII:
Screenshot 2026-04-22 182731

lib/commands/new.js
NullVoxPopuli, line 1: I was expecting to at least see a removal somewhere. where does the error message come from today? I don't want to duplicate code

AyushKashyapII:
I have completed the refactor to eliminate the code duplication.You were right abt duplicate code !!!
Key changes:
Sourced from Truth: Replaced the hardcoded message in lib/commands/new.js with the centralized MSG_HELP exported from lib/utilities/get-lang-arg.js.
 - Cleaned Utility: Removed all ui side-effects (warnings/info logs) from the get-lang-arg utility. It is now a pure data utility that returns the language code, while the Command handles the UI 'fail-fast' logic.
 - Updated Tests: Rewrote the utility tests to verify it no longer emits UI logs and updated the command tests to assert the combined error message.
This removes the redundant logic while ensuring a consistent, high-visibility error message for the user.

NullVoxPopuli:
but why was existing behavior removed now?

AyushKashyapII:
I initially over-refactored: while removing duplicated --lang help text in new.js. I apologize for the hiccup i was not able to engineer the dependency well.

lib/utilities/get-lang-arg.js
```diff
@@ -259,10 +259,12 @@ function writeFullMsg(fullMsg, ui) {
...
+function getLangArg(langArg, ui) {
   let fullMsg = getFullMsg(langArg);
   if (fullMsg.body) {
     writeFullMsg(fullMsg, ui);
   }
   return getResult(langArg);
-};
+}
+module.exports = getLangArg;
+module.exports.MSG_HELP = MSG_HELP;
```
NullVoxPopuli, line 270: why does this need to be exported? it looks like if getLangArg is used, the message will print if the lang is invalid?

AyushKashyapII: I exported MSG_HELP so I could reuse the help text inside a SilentError in new.js.
While getLangArg prints the message, it does so as a warning and allows the command to continue. By exporting the string, I can throw a hard error in the new command to 'fail fast' while keeping the documentation in a single source of truth

PR #10878, opened 2025-12-08
stop using internal package cache for smoke-test-slow

mansona:
So I've been trying to track down the issue with https://github.com/ember-cli/ember-cli/pull/10873 for quite some time now and I am pretty sure there is something going on with the acceptance test PackageCache implementation 🙈 there is an interaction between tests that is causing eslint to hang ever since the new app blueprint started using the esm version of ember-template-compiler 🫠 

We really shouldn't be doing what the package cache is doing. It's not a great system to test that packages are working correctly so I have removed it for the test that was hanging and now we're relying on the pnpm store to do our caching for us. 

This might make our tests a bit slower but it's much easier to understand what is going on and should (finally) unblock the release 🎉 

Edit: I merged this branch into https://github.com/ember-cli/ember-cli/pull/10873 locally and it does indeed fix the problem 🎉 plz accept and merge 🙏

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,9 @@ jobs:
       - uses: ./.github/actions/setup-environment
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Work around windows short path alias # https://github.com/actions/runner-images/issues/712
+        if: runner.os == 'Windows'
+        run: new-item D:\temp -ItemType Directory; echo "TEMP=D:\temp" >> $env:GITHUB_ENV
       - run: pnpm test tests/acceptance/${{ matrix.test-file }}-test-slow.js
 
   feature-flags:


kategengler:
I'm curious to see how long the tests take. I tried removing the crazy PackageCache stuff awhile back and the tests took literal hours.

mansona:
Because we're now on pnpm it's much easier to rely on the global store to do all the caching for us.

Before: Mocha Tests Running Time: 3:00.981 (m:ss.mmm)
After: Mocha Tests Running Time: 5:49.536 (m:ss.mmm)

kategengler:
I don't think the speed was ever really an issue on Mac or linux.

mansona:
The Windows ones are quite bad now but not too problematic: 

Before: Mocha Tests Running Time: 6:19.937 (m:ss.mmm)
After: Mocha Tests Running Time: 10:55.878 (m:ss.mmm)

PR #10879, opened 2025-12-09
fix: load blueprint options before command validation (fixes #10872)

### LINKED ISSUE #10872: Blueprints should be allowed to have their own flags
Currently, I get warnings like this in all blueprints I help maintain
```
The option '--node-version' is not registered with the 'new' command. 
Run `ember new --help` for a list of supported options.
```

I believe the problem here is that the _command_ for blueprints is what defines the flags, and not the actual blueprints themselves.

This is a factoring problem, because it means if a blueprint wants to eliminate the warning, they must add it to ember-cli (or just not use ember-cli for blueprint generation (or just deal with the unprofesh warning 🙈  ))

Do we have a path to pushing flags into blueprints? or allowing blueprints to specify what flags they want to support?

AyushKashyapII:
### Description
Fixes #10872.

Currently, commands like `new` and `addon` validate their arguments **before** loading the options defined in the target blueprint. This causes the CLI to warn that "The option 'x' is not registered" even if the blueprint explicitly defines that option in its `availableOptions`.

### The Fix
1. **Execution Order:** Modified `lib/models/command.js` to execute the `beforeRun` hook (which calls `mergeBlueprintOptions`) inside `validateAndRun` **before** the argument validation/parsing occurs.
2. **Blueprint Lookup:** Updated `lib/utilities/merge-blueprint-options.js`. Since `nopt` parsing hasn't happened yet at this stage, the helper now manually parses `rawArgs` to find the blueprint name (handling `-b`, `--blueprint`, and defaulting to `app`/`addon`), ensuring the correct options are merged.
3. **Regression Fix:** Fixed a crash in `mergeDuplicateOption` that was exposed by this timing change. Previously, if a blueprint option had undefined aliases, `flatMap` would throw an error. Added a fallback `|| []` to prevent this.

### Reproduction
1. Create a blueprint that has a custom option (e.g., `--pizza` in `availableOptions`).
2. Run `ember new my-app --pizza=cheese`.
3. **Before:** The CLI warns: `The option '--pizza' is not registered with the 'new' command.`
4. **After:** The CLI accepts the option silently and passes it to the blueprint as expected.

### Testing
*   Added a unit test in `tests/unit/models/command-test.js` to verify `beforeRun` is called before argument parsing.
*   Verified manually that `ember new` and `ember addon` no longer emit warnings for valid blueprint flags.


NullVoxPopuli:
Needs some work <3
```
❯ node bin/ember new my-app --blueprint /✂️/emberjs/ember-app-blueprint --minimal
The option '--minimal' is not registered with the 'new' command. Run `ember new --help` for a list of supported options.
installing ember-app-blueprint
✨  Creating a new Ember app in <repo>/my-app:
The "path" argument must be of type string. Received an instance of Array


Stack Trace and Error Report: /tmp/error.dump.795192316a8a0de9cf025525534b229f.log

```

AyushKashyapII:
@NullVoxPopuli Thanks for the review!

I have addressed the feedback:
Fixed Path Resolution: I added normalizeBlueprint to the early lookup in merge-blueprint-options.js. This ensures that when passing a path (e.g., --blueprint ./local/path), it resolves correctly instead of crashing.
Fixed Alias Crash: I verified that mergeDuplicateOption could crash if a blueprint option had aliases: undefined. I updated command.js to handle this defensively (aliases || []).
I verified this locally using a test blueprint at a relative path (./blueprints/test) with custom flags, and confirmed both the warnings and the crash are gone. Ready for review!

AyushKashyapII:
@NullVoxPopuli r u still available to look into this

blueprints/addon-import/index.js
NullVoxPopuli, line 1: does this change need to be here? can you reduce the diff of this PR to just what it's changing?

lib/models/command.js
NullVoxPopuli, line 1: same here -- why do these changes exist?

lib/utilities/merge-blueprint-options.js
```diff
@@ -2,26 +2,43 @@
...
- */
 module.exports = function (rawArgs) {
   if (rawArgs.length === 0) {
     return;
   }
+
+  // Default behavior:The first argument is the blueprint (e.g., 'ember generate component')
+  let blueprintName = rawArgs[0];
+
+  // FIX: The 'new' and 'addon' commands use the first arg as the Project Name, not Blueprint.
```
NullVoxPopuli, line 15: comments like this don't make sense in code, and only make sense in PR descriptions

lib/utilities/merge-blueprint-options.js
```diff
@@ -2,26 +2,43 @@
...
+  // We need to find the actual blueprint name.
+  if (this.name === 'new' || this.name === 'addon') {
+    //Start with the default (usually 'app' or 'addon')
```
NullVoxPopuli, line 18: remove all comments explaining "what" instead of "why"

tests/unit/models/command-test.js
```diff
@@ -102,6 +102,42 @@ describe('models/command.js', function () {
     expect(new ServeCommand(options).parseArgs(['--port', '80'])).to.have.nested.property('options.port', 80);
   });
 
+  it('calls beforeRun before parsing args', async function () {
```
NullVoxPopuli, line 106: why is calling beforeRun important?

NullVoxPopuli, line 106: this test doesn't assert that the situation described in #10872 is resolved, as the only args used are ones that are defined

NullVoxPopuli (changes requested):

PR #10892, opened 2025-12-14
More dependency updates

bertdeblock:


kategengler (approved):

lib/utilities/execa.js
```diff
@@ -1,20 +0,0 @@
-'use strict';
```
bertdeblock: This was not used consistently, and we can get the same or more info by setting `NODE_DEBUG=execa` when needed.

PR #10907, opened 2025-12-20
Update `nopt` to v9

bertdeblock:


mansona:
@bertdeblock what is the status here? This is in draft so I didn't really pay much attention to it but it looks like it solves issues with https://github.com/ember-cli/ember-cli/pull/11010 and https://github.com/ember-cli/ember-cli/pull/11019

Should we rebase and merge?

PR #10921, opened 2026-01-17
[BUGFIX] ensure CSS instructions work out the box

billybonks:
Since Vite became the default, if you follow the instructions for plain CSS and create multiple CSS files they will not be compiled properly

this is because according to embroider spec the CSS file should have an import statement in order for the compilation to occur properly

https://embroider-build.github.io/embroider/docs/spec.html see under CSS header

Initial research done here https://github.com/embroider-build/embroider/issues/2657


NullVoxPopuli:
:wave: thank you for working on this!

The file you made a change to is actually the pre-vite blueprint.

The vite blueprint is over here: https://github.com/ember-cli/ember-app-blueprint/

billybonks:
Thanks, i moved the pr https://github.com/ember-cli/ember-app-blueprint/pull/188 let me know if its in the correct file :)

PR #10927, opened 2026-01-24
Ensure new projects do not start with a deprecation warning.

kategengler:
See https://deprecations.emberjs.com/v6.x#toc_using-amd-bundles

Not sure if we should do this because it will also change the default for upgraders and they won't know to check if it is safe to toggle.


kategengler:
No longer relevant.

PR #10945, opened 2026-02-07
update release-plan for OIDC

mansona:
Our last attempt to release an alpha failed because our npm token has now expired 😭 

I have already setup OIDC and this PR runs `pnpm create release-plan-setup@latest` to update our release-plan CI jobs to align with the new defaults and make it possible to release again

Note: we can no longer have multiple publish-branchname.yml CI jobs because of OIDC, I will temporarily go back to maintaining changes across branches on this file but I have a plan to make this easier to manage going forward


NullVoxPopuli (approved):
so close :P

PR #10990, opened 2026-04-03
bump node on publish.yml and stop updating npm

mansona:
This is forward-porting https://github.com/ember-cli/ember-cli/pull/10989 so that it can make it to all the other branches during the normal release workflow 👍


mansona:
I cancelled CI because it's not relevant to this change 👍

PR #10994, opened 2026-04-13
chore: migrate clean-remove and walk-up-path to TypeScript

JANS66:
This PR converts `lib/utilities/clean-remove` and `lib/utilities/walk-up-path` to TypeScript as part of the effort for **[RFC #0800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan/)**.

### Changes:
- Converted `.js` utilities to `.ts` with proper interfaces.
- Generated `.d.ts` declaration files for both utilities.
- Updated `lib/models/blueprint.js` to handle the ESM default export.
- Updated integration and unit tests (`clean-remove-test.js` and `walk-up-path-test.js`) to support the migration.
- Added `typescript`, `@types/node`, and `@types/fs-extra` to workspace devDependencies.

### Verification:
- All 1434 tests passed locally using `pnpm test`.


mansona:
What is the value of making this change? 🤔 the ember-cli codebase is very seldom consumed other than by internal modules, do you have a use case in mind?

JANS66:
> What is the value of making this change? 🤔 the ember-cli codebase is very seldom consumed other than by internal modules, do you have a use case in mind?

Honestly, I'm just looking to help out with **[RFC #0800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan/)** push for TS adoption. I know these are mostly internal, but i figured since the goal is to get the CLI codebase typed eventually, these utilities were a good place to start. Just trying to help knock out some of that incremental progress mentioned in the RFC

mansona:
> but i figured since the goal is to get the CLI codebase typed eventually

I don't think that is a goal that is stated anywhere 😂 

> Honestly, I'm just looking to help out with [RFC #0800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan/)

This RFC is already in the recommended stage https://github.com/emberjs/rfcs/blob/main/text/0800-ts-adoption-plan.md?plain=1#L2 which means the RFC is completed and there is no more work needed to do 🤷 

> push for TS adoption

If your goal is for TS adoption I would ask for more details about that, let's not churn internal tools that don't have much value for the end-user but instead let's identify a problem that you're trying to solve. Is there a thing that you've been trying to build that you thought would be better with more type information?

JANS66:
> > but i figured since the goal is to get the CLI codebase typed eventually
> 
> I don't think that is a goal that is stated anywhere 😂
> 
> > Honestly, I'm just looking to help out with [RFC #0800](https://rfcs.emberjs.com/id/0800-ts-adoption-plan/)
> 
> This RFC is already in the recommended stage https://github.com/emberjs/rfcs/blob/main/text/0800-ts-adoption-plan.md?plain=1#L2 which means the RFC is completed and there is no more work needed to do 🤷
> 
> > push for TS adoption
> 
> If your goal is for TS adoption I would ask for more details about that, let's not churn internal tools that don't have much value for the end-user but instead let's identify a problem that you're trying to solve. Is there a thing that you've been trying to build that you thought would be better with more type information?

My bad, i read the RFC as a push to migrate the internal source too, but i see what you mean about not wanting to churn stable internal tools without a specific reason. Im mostly looking to contribute and get more familiar with the codebase. Are there any specific areas where you would actually welcome some help or type coverage? Id rather spend my time on something you guys actually find valuable

PR #11000, opened 2026-04-22
Ensure blueprint options are registered before argument validation

### LINKED ISSUE #10872: Blueprints should be allowed to have their own flags

AyushKashyapII:
### Description
Fixes #10872.

Currently, commands like `new` and `addon` validate their arguments **before** loading the options defined in the target blueprint. This causes the CLI to warn that "The option 'x' is not registered" even if the blueprint explicitly defines that option in its `availableOptions`.

### The Fix
1. **Execution Order:** Modified `lib/models/command.js` to execute the `beforeRun` hook (which calls `mergeBlueprintOptions`) inside `validateAndRun` **before** the argument validation/parsing occurs.
2. **Blueprint Lookup:** Updated `lib/utilities/merge-blueprint-options.js`. Since `nopt` parsing hasn't happened yet at this stage, the helper now manually parses `rawArgs` to find the blueprint name (handling `-b`, `--blueprint`, and defaulting to `app`/`addon`), ensuring the correct options are merged.
3. **Regression Fix:** Fixed a crash in `mergeDuplicateOption` that was exposed by this timing change. Previously, if a blueprint option had undefined aliases, `flatMap` would throw an error. Added a fallback `|| []` to prevent this.

### Reproduction
1. Create a blueprint that has a custom option (e.g., `--pizza` in `availableOptions`).
2. Run `ember new my-app --pizza=cheese`.
3. **Before:** The CLI warns: `The option '--pizza' is not registered with the 'new' command.`
4. **After:** The CLI accepts the option silently and passes it to the blueprint as expected.

### Testing
*   Added a unit test in `tests/unit/models/command-test.js` to verify `beforeRun` is called before argument parsing.
*   Verified manually that `ember new` and `ember addon` no longer emit warnings for valid blueprint flags.


AyushKashyapII:
Calling beforeRun at this specific point is critical because of how Ember CLI handles dynamic options. Commands like new, addon, and generate don't know their full list of valid flags at compile-time; many flags are defined by the blueprint being used.
Currently, parseArgs (which triggers the 'unregistered option' warning) runs before the blueprint has a chance to register its unique options. By moving beforeRun to execute before parseArgs, we allow the merge-blueprint-options utility to look up the target blueprint and inject its availableOptions into the command. This ensures that when parseArgs eventually runs, it recognizes blueprint-specific flags (like --node-version or custom blueprint flags) as valid, thus preventing the confusing warning

NullVoxPopuli:
Can you show the terminal output while using custom flags that a blueprint provides? Do they show up in help now for the blueprint?

AyushKashyapII:
To verify the fix, I manually added a custom option { name: 'pizza', type: String } to the availableOptions of the default app blueprint on my local machine.
1. Execution without warnings:
As shown in the terminal output below, the CLI now correctly recognizes the blueprint-provided flag. The 'not registered' warning is no longer emitted, and the command proceeds to install the blueprint with the custom option.
Screenshot 2026-04-22 181316

2. Blueprint options in Help:
By moving the mergeBlueprintOptions call to the beforeRun hook, these dynamic options are now also registered in time for the help output. Running ember new --help now correctly lists --pizza as a supported option for the command.
This confirms that the change in execution order effectively allows blueprints to define and validate their own command-line interface flags.

NullVoxPopuli:
that is a case where we want to warn of unrecognized argument. as the default blueprint does not implement `--pizza`

AyushKashyapII:
To test the fix, I manually added --pizza to the availableOptions of the local app blueprint. My goal was to demonstrate that once a blueprint defines an option, the CLI correctly recognizes it without warning.

I completely agree that the CLI should still warn for truly unrecognized arguments. I have verified that this behavior is preserved. As shown in the screenshot below, if I pass an option like --taco (which is not defined in the command or the blueprint), the CLI still correctly emits the 'not registered' warning.
image

NullVoxPopuli:
I don't think it works: 
image



This is using an old blueprint https://github.com/embroider-build/addon-blueprint/blob/main/index.js

but it has many flags

AyushKashyapII:
Root cause was that custom blueprint options were not always registered early enough for parse-time validation when explicit external blueprints were used. I fixed registration in beforeRun flow, added external blueprint resolution fallback, and covered it with unit regressions for both new and addon. I re-ran relevant unit + acceptance suites and manual repros; all pass, and the original --addon-only registration issue is resolved.

NullVoxPopuli:
did you test with:
```bash
node ./bin/ember addon my-addon --blueprint @embroider/addon-blueprint --addon-only
```
?

AyushKashyapII:
Yes — I tested exactly this command:

node ./bin/ember addon my-addon --blueprint @embroider/addon-blueprint --addon-only

Result:
image


The previous warning (--addon-only is not registered) is no longer shown.
Command proceeds successfully with the blueprint.
I also verified no regressions by running:

pnpm test tests/unit/commands/addon-test.js tests/unit/commands/new-test.js tests/unit/models/command-test.js
pnpm test tests/acceptance/generate-test.js tests/acceptance/addon-generate-test.js
Both passed.

NullVoxPopuli:
now what about:
```bash
node ./bin/ember addon my-addon --blueprint @embroider/addon-blueprint --addon-only-i-made-this-up
```
this _should_ throw a warning

AyushKashyapII:
I validated both scenarios with the exact command flow you asked about.

Tested: node ./bin/ember addon my-addon --blueprint @embroider/addon-blueprint --addon-only --dry-run --skip-git --skip-npm Result: no --addon-only “not registered” warning.

Tested: node ./bin/ember addon my-addon --blueprint @embroider/addon-blueprint --addon-only --i-made-this-up --dry-run --skip-git --skip-npm Result: --i-made-this-up correctly warns as unregistered.

Implementation update: I replaced the earlier broad passthrough with a modular/generic pattern:

merge-blueprint-options now collects allowed explicit-blueprint passthrough options via command hook + blueprint source inference.
new/addon provide explicit passthrough options through getExplicitBlueprintPassthroughOptions(blueprintName).
This keeps the intended external blueprint flag working while preserving strict warnings for unknown flags.

PR #11010, opened 2026-04-27
update nopt to v9

mansona:
I have been updating dependencies recently to see which ones will be easy to change. It looks like this one had a change in behaviour so it's not an easy upgrade. 

I don't have the time to be looking at this right now so if anyone wants to take over and take a look please feel free 👍


2005rishabh:
I'm going to take a look at the basic-tests failure and see if I can align the code with the nopt v9 breaking changes.

PR #11018, opened 2026-05-08
Fix string option parsing with nopt v9

2005rishabh:
I investigated the `nopt@9` regression and narrowed the issue down to parsing string option values that begin with `--`.

`nopt@9` interprets those values as additional flags instead of string values, which breaks cases like:

`--options "--split 2 --random"`

The fix adds a small compatibility layer in `parseArgs()` that temporarily masks string option values beginning with `--` during parsing and restores them afterward.

Verified locally with:

* `tests/unit/models/command-test.js`
* relevant acceptance tests (`ember new uses yarn when blueprint has yarn.lock`)

Current local result:

* 173 passing
* 0 failing

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -569,17 +569,80 @@ let Command = CoreObject.extend({
       }
     };
 
+    let stringOptions = new Set();
+
     this.availableOptions.forEach((option) => {
-      if (typeof option.type !== 'string') {
-        knownOpts[option.name] = option.type;
-      } else if (option.type === 'Path') {
+      let type = option.type;
+
+      if (typeof type !== 'string' && type !== undefined) {
+        knownOpts[option.name] = type;
+      } else if (type === 'Path') {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
what's the regression? I'm not familiar

mansona:
@NullVoxPopuli I think regression is the wrong word here. I tried updating to the latest version and it was failing for some unknown reason https://github.com/ember-cli/ember-cli/pull/11010

@2005rishabh would you not be better to target the branch that updated the nopt version? 🤔 I don't see how making this change on main would help

2005rishabh:
You're right — this compatibility fix only makes sense alongside the existing `nopt@9` upgrade work.

I'll retarget the changes onto the `nopt` upgrade branch and update the PR.

2005rishabh:
Closing this in favor of a PR targeting the existing `nopt` upgrade branch.

PR #11019, opened 2026-05-09
Fix string option parsing with nopt v9

2005rishabh:
This builds on top of the existing `nopt@9` upgrade branch.

`nopt@9` treats string option values beginning with `--` as additional flags instead of string values, which breaks cases like:

`--options "--split 2 --random"`

This patch adds a small compatibility layer in `parseArgs()` that temporarily masks string option values beginning with `--` during parsing and restores them afterward.

Verified locally with:

* `tests/unit/models/command-test.js`
* relevant acceptance tests (`ember new uses yarn when blueprint has yarn.lock`)

Current local result:

* 173 passing
* 0 failing

### DIFF
--- a/lib/models/command.js
+++ b/lib/models/command.js
@@ -569,17 +569,73 @@ let Command = CoreObject.extend({
       }
     };
 
+    let stringOptions = new Set();
+
     this.availableOptions.forEach((option) => {
-      if (typeof option.type !== 'string') {
-        knownOpts[option.name] = option.type;
-      } else if (option.type === 'Path') {
+      let type = option.type;
+
+      if (typeof type !== 'string' && type !== undefined) {
+        knownOpts[option.name] = type;
+      } else if (type === 'Path') {
         knownOpts[option.name] = path;
+      } else if (typeof type === 'string' && type.toLowerCase() === 'boolean') {
+        knownOpts[option.name] = Boolean;
       } else {
-        knownOpts[option.name] = String;
+        knownOpts[option.name] = [String, null];
+      }
+
+      if (
+        knownOpts[option.name] === String ||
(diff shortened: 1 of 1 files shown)


mansona:
I haven't looked into this totally yet but the amount of change seems way overkill 🤔 was there any upgrade guide for nopt that suggested the kind of changes that we might need to do?

2005rishabh:
I couldn't find any upgrade or migration notes from nopt that covered this specific behavior change.

While debugging locally, it looked like nopt v9 was treating string option values that start with `--` as additional flags instead of plain string values. That was breaking cases like:

`--options "--split 2 --random"`

The masking/unmasking approach was my attempt to preserve the existing ember-cli behavior without changing the current command APIs or option definitions.

I do agree the patch ended up being a bit heavier than I'd ideally like though. If there's a cleaner way to handle this with nopt v9, I'm definitely happy to rework or simplify it.

mansona:
I noticed yesterday in @bertdeblock 's PR that attempted the same thing he actually removed the test as it could be wrong 🤷 https://github.com/ember-cli/ember-cli/pull/10907

any thoughts?

bertdeblock:
I _think_ the better fix is doing what's mentioned in https://github.com/ember-cli/ember-exam/issues/84#issuecomment-297858057, and avoid any additional "parsing" ourselves.

2005rishabh:
That context helps, thanks.

While debugging, it still looked like `nopt@9` was treating string values starting with `--` as extra flags, which is why I tried the masking approach to keep the existing ember-cli behavior working.

But I agree the fix ended up being heavier than ideal 😅

I’ll read through the ember-exam discussion more carefully and see if there’s a simpler way to handle this without adding extra parsing logic in ember-cli.

PR #11020, opened 2026-05-12
Update diff to latest v8.x

mkszepp:
This PR is fixing a low security issue in `@ember-tooling/blueprint-model` package (see https://github.com/advisories/GHSA-73rr-hh4g-fpgx)

The latest version of diff is v9, but it doesn't support anymore ES5... not sure if we can upgrade already to v9, but updating to latest v8 helps already to remove this security issue.

v8.x was already installed in `ember-cli` package, but i think there is no usage in `ember-cli` package (only in blueprint-model package there is use).

grafik

Let me know, if you want remove this i `ember-cli`

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "content-tag": "^4.0.0",
     "core-object": "^3.1.5",
     "dag-map": "^2.0.2",
-    "diff": "^8.0.2",
+    "diff": "^8.0.4",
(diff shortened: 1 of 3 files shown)


void-mAlex (approved):

void-mAlex:
test failure does seem to be related to the chage

PR #11025, opened 2026-05-21
[BUGFIX release] Refactor blueprint loading to support multiple file types (js, cjs, mjs)

NullVoxPopuli:
Unblocks https://github.com/emberjs/ember.js/pull/21419

In order to ember.js to try out being type=module, the blueprint loader needs this bugfix.

the bug: incorrectly assuming index.js is cjs, and then not respecting the other JS extensions.

thankfully, because our supported node matrix is _only_ the nodes that support require(esm), this is a super easy bugfix, and doesn't require anything invasive, like it would pre require(esm).


bertdeblock (approved):

lib/models/blueprint.js
```diff
@@ -1472,14 +1472,23 @@ Blueprint.load = function (blueprintPath, blueprintOptions) {
...
+
+        if (candidateSuffix.endsWith('.mjs')) {
+          blueprintModule = blueprintModule.default;
+        }
+
+        if (typeof blueprintModule === 'function') {
+          Constructor = blueprintModule;
+        } else {
+          Constructor = Blueprint.extend(blueprintModule);
+        }
```
mansona, line 1491: Shouldn't we return if we found something here? 🤔 this loop looks like it keeps going after it finds something, which means that it's going to have a preference for the last candidate and not the first

Is there an order of what Node looks up in this case? does it look for mjs then cjs then js?

mansona (dismissed):
this is a question but I would like it answered before we merge please 🙏

mansona:
Also why are we bugfixing this? you say that this unblocks a PR on ember-source but that PR is targeting main 🤔 I would be much more comfortable having this target main

NullVoxPopuli:
Ember sounce doesn't use ember-cli @ main tho, and i'm not waiting 3 months to resume that pr ;p

lib/models/blueprint.js
NullVoxPopuli: Early return is good. However, once you start using extensions, there is no resolving order 

https://nodejs.org/api/esm.html#features


It's .js, and then you must specify an extension

mansona (changes requested):
> Ember sounce doesn't use ember-cli @ main tho, and i'm not waiting 3 months to resume that pr ;p

I don't think that's a good enough reason to make this a bugfix release. It sounds like it should be using alpha on the ember-source main branch 🤷 

I'm happy to remove my objection if the other Tooling Team Members agree with you, but in the mean time can you please add this to the agenda for the next meeting 👍

NullVoxPopuli:
> It sounds like it should be using alpha on the ember-source main branch 🤷

why tho? your reasoning is vibes. :p the change is so simple, and _is_ a bugfix.
us not supporting even the cjs extension with our current node support _is a bug_.

kategengler:
I don't necessarily agree with the motivation being a PR on ember.js but otherwise this seems like a low-risk bugfix. I've actually already tried to use mjs for a blueprint and been surprised I could not.

NullVoxPopuli:
@kategengler, @mansona and I are chatting sync tomorrow to work through use cases to try to assess risk -- will post results here afterwards

kategengler:
I think we should do this bugfix sooner than later but
something to consider for that ember PR is that if ember-source 7.2 switches to cjs for blueprints there will be a minimum ember-cli version required to use ember-source 7.2.

NullVoxPopuli:
It's probably fine to try to keep all of ember-cli 7.x compatible with all of ember-source 7.x, and vice-a-versa

NullVoxPopuli:
A smaller fix is extracted to: https://github.com/ember-cli/ember-cli/pull/11028

NullVoxPopuli:
Closing because:
- we shipped support for ambiguous modules (__esModule gating) in 7.0.1
- too many conflicts

PR #11026, opened 2026-05-22
Prepare 7.1 beta

mansona:


tests/fixtures/deprecate-override/index.mjs
```diff
@@ -0,0 +1,13 @@
...
+  for: 'ember-cli', // this needs to be ember-cli so that it triggers our internal system
+  id: 'deprecate-override-test',
+  since: {
+    available: '4.0.0',
+    enabled: '4.0.0',
+  },
+  until: '15.0.0',
+});
+
+console.log('success');
```
void-mAlex, line 13: required?

mansona: yes this is required because we needed code that would execute after a potential throw. But also, this isn't relevant to the prepare 7.1 beta PR, it should have been a comment on https://github.com/ember-cli/ember-cli/pull/10610 🙈

mansona: we don't need to review all the code that goes into these PRs because they have already been reviewed, we're just checking for anything odd or stuff that looks broken with the dependencies

NullVoxPopuli (approved):

PR #11037, opened 2026-06-28
bring test fixtures in line with updates from addon readme

void-mAlex:
this fixes the three broken tests on main


mansona (approved):
nice catch 🎉

PR #11038, opened 2026-06-30
fix running tests on blueprint markdown changes

mansona:
There have been some changes recently to the blueprint markdown files that weren't updated in the fixtures, because my filter here was way too restrictive it wasn't actually running the tests 🙈 

Edit: the fact that CI is running on https://github.com/ember-cli/ember-cli/pull/11037 and not https://github.com/ember-cli/ember-cli/pull/11023 shows that this is working as expected 🎉 (note: I rebased both those PRs on this one and made this branch the target so it should be a proper test)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ on:
   pull_request:
     paths-ignore:
       - "**.md"
+      - "!packages/**.md"
+      - "!tests/**.md"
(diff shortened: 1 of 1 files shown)


void-mAlex (approved):

mansona:
I'm merging this through the failure because the fix for this failure is in https://github.com/ember-cli/ember-cli/pull/11037 👍

PR #11043, opened 2026-07-07
Update fixturify-project to ^7.1.3

NullVoxPopuli-ai-agent:
Updates `fixturify-project` from `^2.1.1` to `^7.1.3` (5 majors). The API changed substantially along the way, so the test helper and its consumers are updated to the new API. Net effect on the helper is a code *reduction* — the new upstream `write()` machinery replaces the helper's hand-rolled serialization.

## Upstream breaking changes handled

- **v3**: the class is now a named export (`const { Project } = require('fixturify-project')`), and `project.root` was removed — `project.baseDir` is now the project directory itself (fixture files are no longer nested under an extra `<name>/` folder inside the tmp dir).
- **v5/v6**: `writeSync()` and `toJSON()` were removed; writing is now the async `write()`.
- **v7**: `files` passed to the constructor are deep-merged (no impact on these tests).

### `tests/helpers/fixturify-project.js`

- `writeSync()` → async `write()`; `buildProjectModel()` / `buildProjectModelForInRepoAddon()` are now async and read the generated `package.json` back from disk.
- In-repo addons are no longer serialized into the parent's `files` (that relied on the removed `toJSON()`); the helper now tracks them and writes each one into `lib/<name>` using the regular `write()` machinery during the parent's `write()`. This deleted the old serialization code and, as a bonus, the file-moving workarounds in the per-bundle-addon-cache fixture helper.
- Dependencies created through `addDependency`/`addDevDependency` are constructed as `EmberCLIFixturifyProject` instances and handed to upstream via its public `addDependency(project)` overload (upstream's internal `addDep` hardcodes its own `Project` class since v5). This keeps nested usage like `addon.addAddon(...)` and `engine.addReferenceDependency(...)` working.
- 'Reference' (dev)dependencies — package.json entries whose files intentionally live elsewhere on disk — are injected via a small `pkgJSONWithDeps()` override. This overrides a private upstream method; there is currently no public hook for customizing the generated `package.json`.

### Tests

- All `writeSync()` / `buildProjectModel()` call sites are now `await`ed.
- `tests/helpers/per-bundle-addon-cache.js`: the standard cache fixture now defines the shared `test-addon-dep` / `test-engine-dep` directly in `PROJECT/lib` (and removes them from the project's `ember-addon.paths`) instead of creating them nested and moving their serialized files around — same disk layout, much simpler.
- `package-info-cache-test`: the `toJSON` monkeypatch that left an empty `node_modules/ember-resolver` folder is replaced by explicitly emptying that directory after `write()`. Path assertions using `fixturifyProject.root` now use `baseDir`.
- `proxy-test`: `expect(() => buildProjectModel()).to.throw(...)` became `await expect(...).to.be.rejectedWith(...)` since the model builder is now async.

## Behavioral change worth flagging

Upstream now writes `devDependencies` to `node_modules` **after** `dependencies` (v2's `toJSON` did the opposite), so when the same package name exists in both with different versions, the devDependency copy is what ends up on disk. The duplicated-addon expectations in `instantiate-addons-test` were updated from `1.0.0` to `2.0.0` accordingly (the addon *ordering* that test locks down is unchanged).

## Also included

- A one-line prettier fix to `RELEASE.md` (separate commit) — `prettier --check` has been failing on master because of it, which keeps the linting CI job red.

## Testing

- `node tests/runner tests/unit/**/*-test.js`: 979 passing, 0 failing
- `pnpm lint` (prettier + eslint): passing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -52,6 +52,7 @@ You can use [this saved search](https://github.com/ember-cli/ember-cli/pulls?q=i
   pnpm dlx update-blueprint-deps --filter '@ember-tooling/.*' --tag latest ./package.json ./packages/*-blueprint/**/*ackage.json
   pnpm dlx update-blueprint-deps --filter '.*' ./package.json ./packages/*-blueprint/**/*ackage.json ./tests/fixtures/*/package.json ./tests/fixtures/*/*/package.json
   ```
+
 - run `pnpm install` to update your pacakge.lock and download any new dependencies
 - run `pnpm lint:fix`
 - update the @ember/app-blueprint dependency `pnpm i -w @ember/app-blueprint@latest`
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
     "eslint-plugin-mocha": "^10.5.0",
     "eslint-plugin-n": "^17.24.0",
     "fixturify": "^3.0.0",
-    "fixturify-project": "^2.1.1",
+    "fixturify-project": "^7.1.3",
     "jsdom": "^21.1.2",
     "latest-version": "^9.0.0",
     "mocha": "^11.7.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -315,8 +315,8 @@ importers:
         specifier: ^3.0.0
         version: 3.0.0
       fixturify-project:
-        specifier: ^2.1.1
-        version: 2.1.1
+        specifier: ^7.1.3
+        version: 7.1.3
       jsdom:
         specifier: ^21.1.2
         version: 21.1.2(bufferutil@4.0.8)(utf-8-validate@5.0.10)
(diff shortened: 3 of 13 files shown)


mansona:
This is a perfect example of a PR that is significantly worse with AI. There were too many subtleties between the old and new ways of using fixturify project that this PR was just needlessly increasing complexity

The real Fix was this: https://github.com/ember-cli/ember-cli/pull/11048

NullVoxPopuli:
while I agree with you generally, in this case it's a bit funny, because it was close, and the merged PR is actually more diff :p

NullVoxPopuli:
(close be danger territory tho)

mansona:
not every line of code is equal, one line of code mulled over for weeks by a human is worth 1000s from a bot. 

I thought about every line that I changed and considered multiple options for each one, so no it wasn't close at all

NullVoxPopuli:
that's what I mean, the "closeness" is dangerous, because the _impact_ (esp long term, say for the next fixturify version) is _not_ close

PR #11044, opened 2026-07-14
add a notify discord webhook action for CI

mansona:
I tested this on this PR 👍 

Screenshot 2026-07-14 at 17 27 58

Note: I also fixed lint in this PR so that we can have CI green again

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,3 +139,24 @@ jobs:
         env:
           OVERRIDE_DEPRECATION_VERSION: "22.0.0"
         run: pnpm test:all
+
+  notify:
+    name: Notify Discord
+    runs-on: ubuntu-latest
+    needs:
+      - linting
+      - basic-tests
+      - basic-tests-matrix
+      - slow-tests
+      - feature-flags
+      - deprecations-broken-test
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -139,3 +139,24 @@ jobs:
...
+      - basic-tests-matrix
+      - slow-tests
+      - feature-flags
+      - deprecations-broken-test
+    if: ${{ !cancelled() && contains(github.ref, 'cron') == true }}
+    steps:
+      - uses: sarisia/actions-status-discord@v1
+        with:
+          webhook: ${{ secrets.CORE_TOOLING_WEBHOOK }}
+          status: "Failure"
```
NullVoxPopuli, line 158: is failure === `!cancelled`?

NullVoxPopuli (approved):
I don't actually know if this is valid, but it doesn't look harmful

.github/workflows/ci.yml
mansona: No that's a different thing. This won't run if the job is cancelled, and status is just an input to the `sarisia/actions-status-discord` action that picks what colour to show down the left of the notification (see the screenshot)

PR #11047, opened 2026-07-30
Update the classic app to Glint v2

### LINKED ISSUE #11046: Classic app blueprint (v6.11+) + typescript dependency issue
It seems that the classic app blueprint has some incompatible dependencies when the `--typescript` flag is used. It still contains a Glint v1 setup, which is not compatible with the bundled ember-cli-htmlbars v7 and npm throws an error during the installation process.

We should update the setup to Glint v2 and backport the fix to at least 6.12 LTS.

Discord discussion: https://discordapp.com/channels/480462759797063690/1532030733412077688

Steps to reproduce:
```
npx ember-cli@6.12 new dependency-issue-reproduction -b @ember-tooling/classic-build-app-blueprint --typescript
```

Windvis:
Closes #11046


Windvis:
I think we are missing fixture tests for the classic app blueprint. I only updated the addon fixtures, but the tests are all passing.

The app version is being skipped here:

https://github.com/ember-cli/ember-cli/blob/e2da7a951cb1ca69448f0ebd539a505984167548/tests/acceptance/new-test.js#L272-L274

That Vite experiment is always enabled at the moment:

https://github.com/ember-cli/ember-cli/blob/e2da7a951cb1ca69448f0ebd539a505984167548/packages/blueprint-model/utilities/experiments.js#L16

Do we need new fixture files for the classic app and update the new command so it uses the classic blueprint specifically? Or will that not work in the test setup?

packages/app-blueprint/files/types/global.d.ts
```diff
@@ -1 +1 @@
-import '@glint/environment-ember-loose';
+import '@glint/ember-tsc/types';
```
NullVoxPopuli: can we delete this file and move it to the tsconfig's compilerOptions#types?

NullVoxPopuli (changes requested):
thanks for doing this! this is really close -- just one change I'd like to see that brings it closer to the v2 app blueprint

Windvis:
> thanks for doing this! this is really close -- just one change I'd like to see that brings it closer to the v2 app blueprint

Done.

The TS app blueprint still isn't being tested though. Not all fixtures were updated and the tests pass. I poked at it briefly but don't really know the way forward there.


Probably best to get the RFC out to deprecate the classic blueprint it so it can be removed in v8  (or earlier since it's not really linked to ember-source) :smile: .

NullVoxPopuli:
> The TS app blueprint still isn't being tested though. Not all fixtures were updated and the tests pass. I poked at it briefly but don't really know the way forward there.

yeah, ember-cli doesn't fully test blueprints -- which was a huge motivation for pulling blueprints out to their own repos -- so we can spend more ci time on proper tests -- which we haven't had motivation to do with the blueprints we want to deprecate

> deprecate the classic blueprint

yea, agreed -- I need to remind myself where this is at -- (ie: ask @mansona )

NullVoxPopuli (approved):

PR #11051, opened 2026-08-13
fix release-plan

mansona:
now that we can approve PRs that are generated with the github bot we don't need to use a Personal Access Token any more 🎉 This means that the release plan PR won't look like it's coming from me any more 😂

### DIFF
--- a/.github/workflows/plan-alpha-release.yml
+++ b/.github/workflows/plan-alpha-release.yml
@@ -61,9 +61,6 @@ jobs:
           sign-commits: true
           branch: releaseplan-preview
           title: Prepare Alpha Release ${{ steps.explanation.outputs.new-version }}
-          # this doesn't use secrets.GITHUB_TOKEN because we want CI to run on the PR that it opens.
-          # See: https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
-          token: ${{ secrets.RELEASE_PLAN_GH_PAT }}
           body: |
             This PR is a preview of the release that [release-plan](https://github.com/embroider-build/release-plan) has prepared. To release you should just merge this PR 👍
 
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):

kategengler:
Just don't delete the PAT -- still used in publish.yml https://github.com/ember-cli/ember-cli/blob/master/.github/workflows/publish.yml#L50 so that the sync-output-repos workflow runs

mansona:
> Just don't delete the PAT

I didn't delete any PAT, it has expired. 

Also, nice catch on the publish. that would have been annoying to notice during the publish step 🙈 thanks 🙇

.github/workflows/publish.yml
```diff
@@ -47,4 +47,4 @@ jobs:
             pnpm release-plan publish --github-prerelease
           fi
         env:
-          GITHUB_AUTH: ${{ secrets.RELEASE_PLAN_GH_PAT }}
+          GITHUB_AUTH: ${{ secrets.GITHUB_TOKEN }}
```
kategengler: Without a PAT here, when this pushes, the workflow for sync-output-repos won't run because when one workflow pushes it doesn't trigger other workflows.

NullVoxPopuli: I think that's fine now because GH recently gave us an "Approve Workflows" button on workflow-generated PRs

kategengler: It doesn't generate a PR. It pushes the tag which should then be the trigger for the sync-output-repos workflow.

kategengler: cc @mansona With this change publishing a version will no longer kick off sync-output-repos which generates ember-new-output and ember-addon-output

kategengler (dismissed):

.github/workflows/publish.yml
mansona: good point, updated the PAT and undid that last commit.

It means that we should probably setup a notification for a failure since we might miss the fact that it's not working when it next expires 🤔

kategengler: That's tricky because an actual failure of sync-output-repos workflow _could_ notify just fine, but if it is never started I don't know how to trigger. 

I supposed we could change it to a cron and have it look for new tags every day.

